PR #338, opened 2017-02-23
ClientHello needs to fit in a single packet

martinthomson:
I'm marking this as editorial.  It touches both TLS and transport.  Anyone who thinks that this is OK can merge it.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -529,6 +529,29 @@ version of TLS.  An endpoint MUST terminate the connection if a version of TLS
 older than 1.3 is negotiated.
 
 
+## ClientHello Size
+
+QUIC requires that the initial handshake packet from a client fit within a
+single packet of 1280 octets.  With framing and packet overheads this value
+could be reduced.
+
+A TLS ClientHello can fit within this limit with ample space remaining.
+However, there are several variables that could cause this limit to be exceeded.
+Implementations are reminded that large session tickets or HelloRetryRequest
+cookies, multiple or large key shares, and long lists of supported ciphers,
+signature algorithms, versions, and other negotiable parameters and extensions
+could cause this message to grow.
+
+For servers, the size of the session tickets and HelloRetryRequest cookie
+extension can have an effect on a client's ability to connect.  Choosing a small
+value increases the probability that these values can be successfully used by a
+client.
+
+A TLS implementation does not need to enforce this size constraint.  QUIC
+padding can be used to reach this size, meaning that a TLS server is unlikely to
+receive a large ClientHello message.
+
+
 ## Peer Authentication
 
 The requirements for authentication depend on the application protocol that is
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-tls.md
```diff
@@ -487,6 +487,29 @@ version of TLS.  An endpoint MUST terminate the connection if a version of TLS
...
+QUIC requires that the initial handshake packet from a client fit within a
+single packet of 1280 octets.  With framing and packet overheads this value
+could be reduced.
+
+A TLS ClientHello can fit within this limit with ample space remaining.
+However, there are several variables that could cause this limit to be exceeded.
+Implementations are reminded that large session tickets or HelloRetryRequest
+cookies, multiple or large key shares, and long lists of supported ciphers,
+signature algorithms, versions, and other negotiable parameters and extensions
+could cause this message to grow.
```
janaiyengar: This also needs to inclde the transport params, right?

draft-ietf-quic-tls.md
```diff
@@ -1223,11 +1246,11 @@ by an attacker.
 Certificate caching {{?RFC7924}} can reduce the size of the server's handshake
 messages significantly.
 
-A client SHOULD also pad {{!RFC7685}} its ClientHello to at least 1024 octets.
+QUIC requires that the packet containing a ClientHello be padded to 1280 octets.
```
janaiyengar: "at least 1280 octets"

draft-ietf-quic-transport.md
```diff
@@ -697,6 +697,16 @@ the cryptographic handshake provides QUIC with:
   client can receive packets that are addressed with the transport address that
   is claimed by the client (see {{source-address-token}})
 
+The initial cryptographic handshake message needs to be sent in a single packet.
+This avoids having to reassemble a message from multiple packets. This would
+require that servers maintain state prior to establishing a connection, exposing
+servers to a denial of service risk.
```
janaiyengar: Phrasing unclear. How about "This avoids having to reassemble a message from multiple packets since doing so would require that a server maintain state prior to establishing a connection, exposing the server to a denial of service risk."

draft-ietf-quic-transport.md
```diff
@@ -697,6 +697,16 @@ the cryptographic handshake provides QUIC with:
...
+
+The first client packet of the cryptographic handshake protocol MUST fit within
+a 1280 octet QUIC packet.  this includes overheads that reduce the space
```
janaiyengar: I don't think we're requiring 1280 bytes, just that it fit within a packet, right? We talk about MTU elsewhere, how about just saying that this packet MUST fit within a single QUIC packet, assuming that the packet is MTU-sized or smaller?

martinthomson: Well, we can't exclude implementations that choose to only send 1280 octet packets.  That would be unwise, I think.  This is MORE strict, I agree, but I think that we want to specifically exclude handshake packets that exceed this or we will find that they aren't properly applicable across multiple networks and use cases.  

If a handshake packet grows too much, then we limit this to networks that can handle a larger MTU.  Given that we don't need the space right now, this seems like a constraint that won't be hard to meet.

draft-ietf-quic-tls.md
```diff
@@ -487,6 +487,29 @@ version of TLS.  An endpoint MUST terminate the connection if a version of TLS
 older than 1.3 is negotiated.
 
 
+## ClientHello Size
+
+QUIC requires that the initial handshake packet from a client fit within a
+single packet of 1280 octets.  With framing and packet overheads this value
```
janaiyengar: "at least 1280 octets."

draft-ietf-quic-transport.md
```diff
@@ -697,6 +697,17 @@ the cryptographic handshake provides QUIC with:
...
 
+The initial cryptographic handshake message MUST be sent in a single packet.
+Any second attempt that is triggered by address validation MUST also be sent
+within a single packet.  This avoids having to reassemble a message from
+multiple packets.  Reassembling messages requires that a server maintain state
+prior to establishing a connection, exposing the server to a denial of service
+risk.
+
+The first client packet of the cryptographic handshake protocol MUST fit within
+a 1280 octet QUIC packet.  This includes overheads that reduce the space
```
janaiyengar: change "1280 octet" to "MTU-sized"

PR #342, opened 2017-02-24
s/psuedo/pseudo

martinthomson:


seanturner:
right so this one looks "approvable" 💯

Issue #33, opened 2016-11-28
Decryption failure == drop packet


martinthomson:
We need to be absolutely explicit here: if unprotecting a packet fails, drop it.  

It's important that we have this rule to prevent trivial DoS attacks on connections.  Also, we need to make it clear that unauthenticated data does not get fed into the system unless we have very careful rules around how we handle that data.

martinthomson:
Closed by #345.

PR #345, opened 2017-02-24
Add a section on what to do when unprotecting a packet fails

### LINKED ISSUE #32: Reordering of really old packets
From @janaiyengar:

> We need rules using
> packet numbers, such that an old packet with a key phase matching the
> current expected should not cause the connection to blow up.

With multiple key updates, there is a possibility that packets from epoch n could collide with packets that were sent in epoch n+2.  This is unlikely, but still something implementations need to handle.  It's not a problem as such unless you insist that connections blow up on decryption failure (which we shouldn't, that's a DoS vector).

It's also not a really bad problem because we have packet numbers that should identify the reordered packet as belonging to the older epoch.  We know what the keys were at that time (if we chose to retain them), so it's not even the case that we need to discard the packet every time.

martinthomson:
Closes #32, #33.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1004,6 +1004,12 @@ confused for retransmissions of handshake messages.  A client cannot initiate a
 key update until all of its handshake messages have been acknowledged by the
 server.
 
+A packet that triggers a key update could arrive after successfully processing a
+packet with a higher packet number.  This is only possible if there is a key
+compromise and an attack, or if the peer is incorrectly reverting to use of old
+keys.  An endpoint MUST immediately terminate the connection if it detects this
+condition.
+
 
 # Client Address Validation {#client-address-validation}
 
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1007,8 +1007,8 @@ server.
 A packet that triggers a key update could arrive after successfully processing a
 packet with a higher packet number.  This is only possible if there is a key
 compromise and an attack, or if the peer is incorrectly reverting to use of old
-keys.  An endpoint MUST immediately terminate the connection if it detects this
-condition.
+keys.  Because the latter cannot be differentiated from an attack, an endpoint
+MUST immediately terminate the connection if it detects this condition.
 
 
 # Client Address Validation {#client-address-validation}
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -764,9 +764,9 @@ QUIC does not assume a reliable transport and is required to handle attacks
 where packets are dropped in other ways.  QUIC is therefore not affected by this
 form of truncation.
 
-The packet number is not reset and it is not permitted to go higher than its
-maximum value of 2^64-1.  This establishes a hard limit on the number of packets
-that can be sent.
+The QUIC packet number is not reset and it is not permitted to go higher than
+its maximum value of 2^64-1.  This establishes a hard limit on the number of
+packets that can be sent.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -778,6 +778,21 @@ the connection that is hosted on stream 1.  This sequence number is not visible
 to QUIC.
 
 
+## Receiving Protected Packets
+
+Once an endpoint successfully receives a packet with a given packet number, it
+MUST discard all packets with higher packet numbers if they cannot be
+successfully unprotected with either the same key, or - if there is a key update
+- the next packet protection key (see {{key-update}}).  Similarly, a packet that
```
marten-seemann: This can only work if these packets were queued previously (because they were undecryptable). Otherwise, they might already have been processed.

martinthomson: To what do you refer?  What works?  What is queued?  (Sorry, I'm being dense probably.)

martinthomson: Sorry, I get what you mean now.  Hmm, not sure what to recommend there.

ianswett: We could require closing the connection if a packet was already processed, under the theory it's either an implementation error or someone intentionally trying to corrupt the connection?

I believe it's as simple as checking if any packets with larger have been processed when a key change occurs?

draft-ietf-quic-tls.md
```diff
@@ -778,6 +778,21 @@ the connection that is hosted on stream 1.  This sequence number is not visible
...
+appears to trigger a key update, but cannot be unprotected successfully MUST be
+discarded.
+
+Failure to unprotect a packet does not necessarily indicate the existence of a
+protocol error in a peer or an attack.  The truncated packet number encoding
```
ianswett: I think it's more likely the internet corrupts the packet.  Also, incorrect truncation is a form of peer bug I'd argue.

draft-ietf-quic-tls.md
martinthomson: The problem there is that it's such a bizarre corner case that you wouldn't necessarily have code for it.  If the packets arrive in the other order, you will drop the higher packet number because you don't even try to decrypt it.  Maintaining the invariant that order of processing doesn't matter is hard here.  It's also the case that the obvious implementation of this doesn't necessarily couple the choice of advancing the keys with updating the last received packet number (I've implemented this for DTLS, and while the code is nearby, it's much easier not to entangle these concepts).

I think that I would rather say "MAY kill the connection".  That leaves the possibility open, and I don't have to explain how tricky it is to get this all lined up.

draft-ietf-quic-tls.md
martinthomson: I thought that too, but a few factors lead me to think that this is still worth saying.  Where the error comes from doesn't change the outcome.  And severe reordering does happen on occasional packets.  Also, your current scheme for picking the truncation length fails if there is a sudden large increase in send rate; packets before the jump that are delayed until after the jump in rate might have shorter packet numbers and thus could be decoded incorrectly.

martinthomson:
I added a small piece of text about the weird reordering Marten pointed out.  I flip-flopped on whether to use MUST, but I think that it is the right answer here.  You will observe that detection isn't guaranteed, so I decided that making the MUST conditional on detection gives implementations a tiny bit of wiggle room if they decide that detecting the condition is too tricky.

Note that it is not possible to generate packets that guarantee triggering this behaviour, so failure of a compliance test on this point is going to be hard to use as justification for a non-compliance label.

PR #346, opened 2017-02-24
Change default initial RTT to 100ms

ianswett:


martinthomson:
I assume that merging is preferable over mere approval.

PR #348, opened 2017-02-24
Define "httpq"

### LINKED ISSUE #253: HTTP/QUIC without Alt-Svc?
While HTTP/QUIC doesn't formally require a client to implement Alt-Svc, there's no discovery mechanism other than Alt-Svc provided, so you're not going to get very far without it.  That makes a full HTTP(/TLS)/TCP stack mandatory simply to open the connection in the first place.  For various reasons (e.g. an embedded device in a controlled environment), a client might want to dispense with TCP altogether when it knows that the endpoint will support QUIC.  It might also be useful in testing to be able to directly reference a QUIC endpoint.

Should we mint new scheme(s) that allows direct reference to a resource served exclusively over HTTP/QUIC?

(Note, in HTTP/2 the answer was "no," because HTTP/2 could be negotiated using the same TCP connection.  QUIC doesn't have that luxury.)

MikeBishop:
Closes #253.  I think there's value in having a way to specify a QUIC-only (or QUIC-first, though less so) endpoint for the reasons described in the issue.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1277,6 +1277,20 @@ it contains.
 
 # IANA Considerations
 
+## Registration of "httpq" URI scheme
+
+IANA maintains the registry of URI Schemes {{?RFC4395}} at
+<http://www.iana.org/assignments/uri-schemes/>.
+
+This document defines the following URI schemes, so the "Permanent
+URI Schemes" registry has been updated accordingly.
+
+|------------|---------------------------------------|---------------|
+| URI Scheme | Description                           | Reference     |
+|------------|---------------------------------------|---------------|
+| httpq      | Hypertext Transfer Protocol over QUIC | {{httpq}}     |
+|------------|---------------------------------------|---------------|
+
 ## Registration of HTTP/QUIC Identification String
 
 This document creates a new registration for the identification of HTTP/QUIC in
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -163,7 +163,7 @@ not support them), then attempt to connect to the alternative using the first
 version in the list which it does support.
 
 
-### QUIC-Only Origins
+### QUIC-Only Origins {#httpq}
 
 An HTTP endpoint where the authoritative endpoint is served over QUIC can be
 identified using the "httpq" URI scheme.
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -99,9 +99,11 @@ frames."  References without this preface refer to frames defined in {{frames}}.
 
 ## Discovering an HTTP/QUIC Endpoint
 
-An HTTP origin advertises the availability of an equivalent HTTP/QUIC endpoint
-via the Alt-Svc HTTP response header or the HTTP/2 ALTSVC frame ({{!RFC7838}}),
-using the ALPN token defined in {{connection-establishment}}.
+### QUIC Alternatives
+
+An HTTP/TCP origin advertises the availability of an equivalent HTTP/QUIC
+endpoint via the Alt-Svc HTTP response header or the HTTP/2 ALTSVC frame
+({{!RFC7838}}), using the ALPN token defined in {{connection-establishment}}.
 
 For example, an origin could indicate in an HTTP/1.1 or HTTP/2 response that
 HTTP/QUIC was available on UDP port 50781 at the same hostname by including the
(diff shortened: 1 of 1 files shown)


martinthomson:
Yes, needs lots of discussion (though we've been reserving that label to mean that it needs discussion at the next meeting, so it might be premature to label it now).  I don't see any way in which this could be deployed, at least in browsers.

MikeBishop:
I don't see any way in which this could be deployed to the WWW in any reasonable timescale -- which is not the same as saying it wouldn't work in a browser.  A browser accessing a dedicated service which runs over QUIC seems very feasible.

janaiyengar:
Why would a browser access a dedicated service running over QUIC? I think this PR touches on something that's not gotten a lot of airtime yet -- that using QUIC without a fallback may not be a good idea for the public Internet, specifically since UDP reachability is not universal. This definitely needs discussion.

ThomasSwindells:
Why (in the next 20 years) wouldn't a browser access a dedicated service running over QUIC as the first choice, hopefully by that time QUIC density will be such that every network does support UDP.

More concretely for the short term ALT-SVC isn't the only way to signal different ways to retrieve files. For instance DASH supports supplying multiple alternative base URLS within the manifest files, it would seem sensible to explicitly signal which servers can be used with httpq directly, otherwise all the benefits of 0RTT get lost. I don't think it is a reasonable assumption to assume that CDNs (or other large distributed systems) would necessarily be configured with all nodes supporting all services, it would be quite reasonable to deploy HTTP only and QUIC only nodes.

Another scenario a HTTPQ url would be sensible to support is when a QUIC service wants to generate a 302 redirect to cause the client to connect to a different node. The server can safely assume the client will be able to make another QUIC connection (particularly if the two server nodes are in the same network location). But when a client is presented with a redirect to a node it has never connected to it has no knowledge as to whether it should keep using HTTPQ or initially start with using HTTP. As I've already said I don't think it is a fair requirement to insist that all QUIC implementations must also be running an HTTP server.

zwoop:
Hmmm, not sure I'm buying this. Assuming QUIC wins, and becomes the future of HTTP, wouldn't UAs just switch behavior, such that we assume QUIC is the predominant protocol for https:// ? With an (optional?) fallback to HTTPS when needed (albeit slower)?

ThomasSwindells:
If you are running a CDN then these aren't origins, they are delivery nodes. Potentially having some delivery nodes which don't support legacy HTTPS, or nodes which haven't yet been upgraded to support QUIC are both realistic circumstances.

RyanAtGoogle:
Origin is defined to be the (scheme,host,port) of a URL. So if "httpq" is a
different scheme than "https", it's a different origin, by definition. If
we're merely talking about serving the same URL (https://) via both QUIC
and TLS, then we're not talking about a different scheme, merely an
alternative service (handled via the "hq" alpn token in an Alternative
Service advertisement.)

​Cheers,

Ryan​

mcmanus:
+1 to ryan's comment - minting new schemes creates un-necessary pressure on
same-origin policies and create operational complexity.

that being said, I know Mike doesn't think this is a web scheme... but the
conversation seems to indicate it would invariably leak into (at least) the
discussion.

I'm opposed to a new scheme

LPardue:
FWIW

The ["Is HTTP being used?" section](https://github.com/httpwg/http-extensions/blob/master/draft-ietf-httpbis-bcp56bis.md#is-http-being-used-used) in bcp56bis defines clauses that are used to determine if the protocol _uses_ HTTP or is _based_ on HTTP. 

It could be argued that HTTP/QUIC satisfies only one of those clauses, that a "http" or "https" scheme is used. Defining and using "httpq" could mean that we lose the tie and violate BCP.

MikeBishop:
Given that we're porting HTTP to a new transport, I would expect the definition in that draft to expand to include the "hq" ALPN token if this draft completes first.  Also, if we take this draft, the transport port in use would be :443.  And lastly, since Alt-Svc is the common case, the scheme will be http:// or https:// most of the time -- this simply provides another option.

LPardue:
All fair points, I was being a bit surreptitious.

Although I was under the impression that use of port 443 was no longer explicitly specified in HTTP/QUIC. It falls under the "any UDP port" statement.

It would be horrible if, formally speaking, the combination of clauses means sometimes HTTP/QUIC is using HTTP and sometimes its not. That might mean that bcp56bis needs some tightening up.

MikeBishop:
It's not, in the current document, since it's only used with Alt-Svc and Alt-Svc always gives you an explicit port.  This PR defines UDP port 443 as the default for the httpq:// scheme, though.

LPardue:
> This PR defines UDP port 443 as the default for the httpq:// scheme, though.

Quite right, forgot/missed that sorry!

MikeBishop:
Severely bitrotten, and the WG is adamantly not going this direction.

PR #350, opened 2017-02-28
Add stream offset limit of 2^64.

janaiyengar:


draft-ietf-quic-transport.md
```diff
@@ -1182,7 +1182,8 @@ The STREAM frame contains the following fields:
 
 * Offset: A variable-sized unsigned number specifying the byte offset in the
   stream for the data in this STREAM frame.  The first byte in the stream has an
-  offset of 0.
+  offset of 0. The largest offset delivered on a stream---the sum of the
+  re-constructed offset and data length---MUST be lesser than 2^64.
```
martinthomson: nit: don't use `---`

Issue #269, opened 2017-02-08
Public Packet Number Echo


martinduke:
In Mirja's Management draft:
[https://tools.ietf.org/html/draft-kuehlewind-quic-appman-00#section-3](url)

> [EDITOR'S NOTE: the addition of a simple packet number echo would
   allow passive RTT measurement and partial passive downstream loss/
   reordering measurement.  Packet number echo can be sampled at the
   echo-side (i.e. one-in-N packets or 1/p packets can carry an echo)
   for efficiency tradeoff, if necessary.]

I endorse this idea. Service providers are very interesting in measuring their users' quality of experience and this is one key thing we can expose to enable that. On the face of it, it seems that the privacy hit is marginal.

britram:
Adding packet-number echo is, IMO, much easier with Jana's alternate header format proposal. This could be done with just two additional codepoints. Propose:

`35 and b5: 1-RTT packet (4-byte packet number + 4-byte packet echo)`

```
 0                   1                   2                   3
 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+-+-+-+-+-+-+-+-+
|      Type     |             
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|                     Packet Number (4)                         |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|                  Packet Number Echo (4)                       |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|                            Payload                          ...
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```

`3d and bd: 1-RTT packet with connection ID(4-byte packet number + 4-byte packet echo)`

```
 0                   1                   2                   3
 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+-+-+-+-+-+-+-+-+
|      Type     |             
[...]

PR #351, opened 2017-03-01
Packet number echo

britram:
Initial suggestion to address #269, based on alternate header proposal from @janaiyengar

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -447,6 +447,8 @@ decryption. These packets contain:
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                     Packet Number (1/2/4)                     |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                  Packet Number Echo (optional)                |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                            Payload                          ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ```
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -470,7 +472,9 @@ this version, it contains:
   * 01: 1-RTT packet (packet number size = 1)
   * 02: 1-RTT packet (packet number size = 2)
   * 03: 1-RTT packet (packet number size = 4)
+  * 07: 1-RTT packet (packet number size = 4, with packet number echo)
```
britram: Can add two additional codepoints here:
```
* 05: 1-RTT packet (packet number size = 1, packet number echo size = 1)
* 06: 1-RTT packet (packet number size = 2, packet number echo size = 2)
```

(I remain skeptical of the tradeoff for varlen packet numbers, but if we're going to have them, we might as well have varlen echo as well.)

ianswett: Sadly, I think a 1 byte echo is commonly too small, whereas 2 bytes is almost always sufficient.  Given this is going to be used on the public internet, I'd define the 2 byte version but not the 1 byte version.

britram: Makes sense to me. Doesn't this mean we should undefine one-byte packet numbers everywhere, though? (That seems like a distinct PR)

martinthomson:
This needs to say how you decode a truncated packet number from the echoed field.  Decoding based on the largest acknowledged packet number that has already been seen is probably right.  

Al is right in observing that you could remove the number from the ACK frame to make the cost of this feature zero bytes, but it's really difficult to swallow the implementation complexity that introduces.  STOP_WAITING with its reliance on public flags is an abomination.  On the other hand, having packet numbers in both directions allows you to have an anchor point for referring to packets that travel in either direction.

You also need to change the guidance around picking a packet number size.  If you are sending an echo, then you need to use the greater of (the number of this packet - highest packet number from those you have sent and received an acknowledgment) and (the echoed packet number you intend to include - the highest acknowledged packet number that was in the most recent packet that contained an ACK) to determine how many octets to devote to the encoding so that you avoid decoding problems.  In other words, I am starting to agree more strongly with Brian that having variable length packet numbers hurts and I want that hurting to stop.

(The CI errors are in @janaiyengar's commit, which was never really intended to be complete.  Jana obviously hasn't bothered installing the commit hook.)

britram:
re: guidance on packet number size. ouch. yes, you're right. For me, this indicates that adding packet echo implies constant size (and probably 4-byte) packet numbers in version 1.

(re: CI errors; yep, I saw these, declined to fix them as I plan to merge @janaiyengar's `header` branch again, which will presumably fix them in a way that would conflict with mine...)

britram:
This has been replaced by #367 (with variable-length numbering) and #368 (with fixed-length, 32-bit numbering).

PR #354, opened 2017-03-02
GOAWAY needs a maximum stream number in both directions

### LINKED ISSUE #347: Clarify meaning/definition of GOAWAY
GOAWAY has a "Last Good Stream ID", which is "The last Stream ID which was accepted by the sender of the GOAWAY message. If no streams were replied to, this value must be set to 0."

The sender does not specify the last stream id it will use, which I think it should.

I also think the text in "6.6 Connection Termination" should specify that all streams <= the Last Good Stream ID will be completed(fin'd or reset), including delivering any outstanding data.

martinthomson:
I have also taken the liberty of updating the text that surrounds the frame type.

@janaiyengar, @mnot, @larseggert, I believe that this is basically a bugfix and it doesn't need any more discussion that it has already.  If you disagree, then we'll hold off on merging until we've had more discussion.

Closes #347.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1657,6 +1657,11 @@ data.  A GOAWAY frame indicates that any application layer actions on streams
 with higher numbers than those indicated can be safely retried because no data
 was exchanged.
 
+In addition to initiating a graceful shutdown of a connection, GOAWAY MAY be
+sent immediately prior to sending a CONNECTION_CLOSE frame that is sent as a
+result of detecting a fatal error.  Higher-numbered streams than those indicated
+in the GOAWAY frame can then be retried.
+
 For peer-initiated streams, an endpoint might indicate a lower value for the
 highest stream number than the value that might be sent by a peer.  After
 receiving a GOAWAY frame, an endpoint SHOULD send a GOAWAY frame in response to
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1668,14 +1668,6 @@ sent immediately prior to sending a CONNECTION_CLOSE frame that is sent as a
 result of detecting a fatal error.  Higher-numbered streams than those indicated
 in the GOAWAY frame can then be retried.
 
-For peer-initiated streams, an endpoint might indicate a lower value for the
-highest stream number than the value that might be sent by a peer.  After
-receiving a GOAWAY frame, an endpoint SHOULD send a GOAWAY frame in response and
-update the value for streams that it initiates, if the value is lower than the
-one it receives.  If an endpoint does send a GOAWAY in response to another
-GOAWAY, it SHOULD NOT update the value for peer-initiated streams, unless it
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1550,39 +1550,71 @@ The fields of a CONNECTION_CLOSE frame are as follows:
...
+  connection.
 
-* Reason Phrase Length: A 16-bit unsigned number specifying the length of the
-  reason phrase.  This may be zero if the sender chooses to not give details
-  beyond the error code.
+Last Client Stream ID:
+
+: The last client-initiated Stream ID which was accepted by the sender of the
+  GOAWAY frame.  All higher-numbered, client-initiated streams (that is, odd
+  numbered streams) are implicitly reset by sending or receiving the GOAWAY
```
MikeBishop: Nit:  "even-numbered" and "odd-numbered" should be hyphenated.

martinthomson:
From the commit comment:

I have also added text that captures something we learned late in the HTTP/2 experience.  GOAWAY pre-emptively closes streams, but for peer-initiated streams it might make sense to use a bigger number to allow any requests that the peer created in the last round trip to complete.  Some implementations find retrying quite difficult and allowing (for example) a client's last set of requests to complete can make the shutdown properly graceful (as opposed to toy-graceful).

The case where this matters is for a gateway to server, where the gateway can't reasonably retry the 10,000 requests that it initiated in the last second. To allow for that, the server can give the gateway a bit of extra leeway for those streams when it sends a GOAWAY frame.

mirjak:
Sorry if that's a little bit off topic but at this point I'm actually wondering if GOAWAY is really part of the transport semantics or should be an application layer signal...?

martinthomson:
@mirjak, I can't actually think of a strong reason why GOAWAY has to be a transport concept.  The ideas I came up with are:

1. It's nice to have the description of graceful shutdown in the transport document.  On the other hand, the pattern is easy to establish and copy.

2. GOAWAY affects stream state.  On higher numbered packets, you shouldn't see any mention of streams higher than the cutoff, and can generate errors accordingly.  But those could be application-layer errors.  

3. GOAWAY also removes the need to reset some streams.  However, any streams that are implicitly reset by GOAWAY don't need to be reset before CONNECTION_CLOSE is sent.  If those streams exist at the time that GOAWAY is sent, then there is no possible way to break the concurrent stream limit, which is the only other relevant transport-layer concern.  There's a potential for local resource management optimizations when you send GOAWAY (discarding flow control windows etc...), but the cases where that applies are few - it only applies if you are cancelling streams that you know to be in progress, and that usually only happens when you are then sending an immediate CONNECTION_CLOSE.

In other words, yeah, we can make this an application-layer concern.

MikeBishop:
I'm dubious that we'll land text on application-level clean shutdown before the draft deadline.  Let's merge this bugfix for consistency in -02, then start on that change separately.  If I'm wrong and we can close on it quickly, great!

PR #355, opened 2017-03-02
Remove error code and reason phrase from GOAWAY

### LINKED ISSUE #352: Does GOAWAY need an error code
In writing up a change for GOAWAY, I realize that it doesn't need to signal the error condition that triggered it.  An endpoint that wants to kill a connection can use CONNECTION_CLOSE and the error code that carries.  The purpose of GOAWAY is to signal graceful shutdown and to mark multiple streams as being safe to retry.

If it doesn't need an error code, then the reason phrase can go away as well.

(This might depend on the outcome of discussions that suggest that CONNECTION_CLOSE could be replaced by Public Reset.  Or, it might influence the outcome of that discussion...)

martinthomson:
In an error condition, CONNECTION_CLOSE is your friend.

Closes #352. Builds on #354.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1657,6 +1657,11 @@ data.  A GOAWAY frame indicates that any application layer actions on streams
 with higher numbers than those indicated can be safely retried because no data
 was exchanged.
 
+In addition to initiating a graceful shutdown of a connection, GOAWAY MAY be
+sent immediately prior to sending a CONNECTION_CLOSE frame that is sent as a
+result of detecting a fatal error.  Higher-numbered streams than those indicated
+in the GOAWAY frame can then be retried.
+
 For peer-initiated streams, an endpoint might indicate a lower value for the
 highest stream number than the value that might be sent by a peer.  After
 receiving a GOAWAY frame, an endpoint SHOULD send a GOAWAY frame in response to
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1668,14 +1668,6 @@ sent immediately prior to sending a CONNECTION_CLOSE frame that is sent as a
 result of detecting a fatal error.  Higher-numbered streams than those indicated
 in the GOAWAY frame can then be retried.
 
-For peer-initiated streams, an endpoint might indicate a lower value for the
-highest stream number than the value that might be sent by a peer.  After
-receiving a GOAWAY frame, an endpoint SHOULD send a GOAWAY frame in response and
-update the value for streams that it initiates, if the value is lower than the
-one it receives.  If an endpoint does send a GOAWAY in response to another
-GOAWAY, it SHOULD NOT update the value for peer-initiated streams, unless it
-wishes to cancel those streams.
-
 
 # Packetization and Reliability {#packetization}
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1600,6 +1600,7 @@ Reason Phrase:
 : An optional human-readable explanation for why the connection was closed.
 
 
+
 ## GOAWAY Frame {#frame-goaway}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1548,41 +1548,56 @@ The fields of a CONNECTION_CLOSE frame are as follows:
...
+  GOAWAY frame.  All higher-numbered, client-initiated streams (that is, odd
+  numbered streams) are implicitly reset by sending or receiving the GOAWAY
+  frame.
 
-* Reason Phrase Length: A 16-bit unsigned number specifying the length of the
-  reason phrase.  This may be zero if the sender chooses to not give details
-  beyond the error code.
+Last Server Stream ID:
+
+: The last server-initiated Stream ID which was accepted by the sender of the
```
LPardue: "The last server-initiated Stream ID which was accepted by the sender of the GOAWAY frame" sounds a bit wonky.

ianswett: Yes, how about was created or will be created?

MikeBishop: Because the sender isn't necessarily the server, and the client isn't "creating" server-initiated streams.  Maybe we could say "the highest ... ID on which the sender has processed or will process data"?  But that doesn't seem more compact, so I'm not sure it's an improvement.

But regardless, that's feedback on #354, not this change in particular.

martinthomson: I have taken this feedback in and will update #354 with this.  I found something else too.

draft-ietf-quic-transport.md
```diff
@@ -1548,41 +1548,61 @@ The fields of a CONNECTION_CLOSE frame are as follows:
...
+|                    Last Server Stream ID (32)                 |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ~~~
 
-The fields of a GOAWAY frame are as follows:
+The fields of a GOAWAY frame are:
 
-* Error Code: A 32-bit field error code which indicates the reason for closing
-  this connection.
+Last Client Stream ID:
```
ianswett: I might be biased towards Largest, instead of Last

draft-ietf-quic-transport.md
```diff
@@ -1548,41 +1548,61 @@ The fields of a CONNECTION_CLOSE frame are as follows:
...
+streams where packets creating new streams could be in transit; using a larger
+stream number allows those streams to complete.
+
+In addition to initiating a graceful shutdown of a connection, GOAWAY MAY be
+sent immediately prior to sending a CONNECTION_CLOSE frame that is sent as a
+result of detecting a fatal error.  Higher-numbered streams than those indicated
+in the GOAWAY frame can then be retried.
+
+For peer-initiated streams, an endpoint might indicate a lower value for the
+highest stream number than the value that might be sent by a peer.  After
```
ianswett: I don't think sending a GOAWAY in response does anything unless the peer receiving the GOAWAY wants to further decrease one or both of the Last Stream IDs, so typically I wouldn't expect this is necessary.  As such, I'd recommend changing this to MAY and explain it's only if the Last Stream IDs are too large.

martinthomson: I've kept this as a SHOULD but added the requested conditions.  It's important that "safe" retries are enabled as much as possible and failure to echo will create unnecessary uncertainty.  This is a big deal in HTTP.

ianswett: You have to ack the GOAWAY, so why echo it, that would seem redundant.

martinthomson: Actually, now that I think about it, the only reason why you would reduce the largest stream number is because you didn't use those streams, and if you don't want to use those streams, then you can just *not use them*.  Let's remove this paragraph.

PR #360, opened 2017-03-06
Clarify the Timestamps section

### LINKED ISSUE #308: "retransmitting" old timestamps in ACK frames
We need to include advice on the creation of ACK frames when an ACK frame was lost.  The advice from @ianswett is that new timing information isn't valuable.  Both endpoints only need to maintain timing information for individual packets for a short period, so retransmission of old timing information isn't valuable.

Thus, we should recommend that endpoints do just that: maintain a short list of packets that they want to provide timing for.  They should then not worry about re-sending any timestamps from lost packets.

ianswett:
Intends to resolve #308

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1221,6 +1221,13 @@ Unlike TCP SACKs, QUIC ACK blocks are cumulative and therefore irrevocable.
 Once a packet has been acknowledged, even if it does not appear in a future ACK
 frame, it is assumed to be acknowledged.
 
+QUIC ack frames contain a timestamp section with up to 255 timestamps.
+Timestamps enable better congestion control, but are not required for correct
+loss recovery, and old timestamps are less valuable, so it is not guaranteed
+every timestamp will be received by the sender.  A receiver SHOULD send a 
+timestamp exactly once for each retransmittable packet recevied. A receiver
+MAY send timestamps for non-retransmittable packets.
+
 A sender MAY intentionally skip packet numbers to introduce entropy into the
 connection, to avoid opportunistic acknowledgement attacks.  The sender MUST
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1221,6 +1221,13 @@ Unlike TCP SACKs, QUIC ACK blocks are cumulative and therefore irrevocable.
 Once a packet has been acknowledged, even if it does not appear in a future ACK
 frame, it is assumed to be acknowledged.
 
+QUIC ack frames contain a timestamp section with up to 255 timestamps.
```
martinthomson: ACK frames.

ianswett:
Added a section to further clarify that Delta Last Acknowledged 1..N" is relative to the last acknowledged

draft-ietf-quic-transport.md
```diff
@@ -1352,6 +1358,10 @@ receive times relative to the beginning of the connection.
 ~~~
 {: #timestamp-format title="Timestamp Section"}
 
+The timestamp section is time-ordered, so the 1 byte Delta Largest
+Acknowledged is always relative to the Largest Acknowleged, but the Time
+Since Previous Timestamp fields are relative.
```
janaiyengar: I would suggest moving this to below the descriptions of the fields. Suggested rewording: "The timestamp section lists packet receipt timestamps ordered by timestamp." I think you can drop the rest of this paragraph since it's redundant with text in the descriptions.

martinthomson:
Rebased and merged manually.

martinthomson:
@ianswett, can you try to remember to name your branches in future.  These branch names are going to collide with old patches (as this one already did, causing me some confusion).

ianswett:
Sure, I'm happy to name them.  I didn't realize it mattered.

Issue #35, opened 2016-11-28
Starting packet number


martinthomson:
We might be able to randomize the starting packet number and thereby avoid having it as a point of ossification.  It also increases the entropy for an off-path attacker who wishes to guess connection IDs because now they have to guess packet number as well.

ianswett:
This is possible today, but it should be required or highly recommended.

ianswett:
When fixing this, also fix the following inconsistency(noted by Mirja):
Section 4.2 (Regular Packets) page 10 says: "The first packet send by an endpoint MUST have a packet number of 1."
And section 7 (Packetization and Reliability) page 28 says: "That is, the first packet in both directions of the connection MUST have a packet number of 0."

JimRoskind:
Reading the motivation (re: off-path guessing of connection ID), I'm hopeful that a client would use a good Random Number Generator (RNG) and not produce a "guessable" connection ID. The presumption of a "good RNG" is indeed central to having many clients connecting to a single server (without collision).  If one would posit a poor RNG... then in addition to collisio problems, I'd be wary that a RNG used to select a "unguessable" initial packet number would be problematic.  Bottom line: I don't grok the security based motivation.

The other motivation is to prevent "ossification."  Can someone provide a plausible motivation for a "need" to start at a different packet number, and avoid "ossification" in this context?  I would have expected that standardization in this area simplifies code, and would not be a long-term problem.

mnot:
Discussed in Tokyo; agreement that it's a good idea. Randomise some number of the bottom bits.

martinthomson:
Currently, we aren't sure how many bits are going to be in initial packets.  When @ekr produces a proposal for the packet headers, I think that this should include a fixed-size slot for the packet number.  That will make this easier to write up.

martinthomson:
This was addressed in #283.

Issue #295, opened 2017-02-13
Connection ID on a version negotiation packet


martinthomson:
Given that we have opted to go with #119 universally, it might be reasonable to think that the server could make a proposal for the connection ID on a version negotiation packet.  However, it probably isn't a good idea.  Since version negotiation is stateless, there is less need to have stable routing for this packet.  Echoing the connection ID gives the client better assurance that its initial packet was what generated the response.

martinthomson:
Hmm, not sure if this is a duplicate of #244.

igorlord:
The first part ("should a version negotiation packet from the server -- a proposal to the client to change its version -- contain a Server-generated Connection ID?") is a new issue.  On one hand, I agree with you that this seems a bit like a kludge.  On the other hand, stable routing may actually be required by some systems.

Imagine a pop with a load balancer and a number of backends behind it.  One may want to roll out a new experimental version of QUIC on a subset of backends. To do that, the system should ensure that the same server that send a version negotiation packet proposing a new QUIC protocol version actually receives packets using the new version (if the client accepts that new version).  Our mechanism to make sure this happens is to encode the backend ID in the Server-proposed Connection ID.

The second part (proving that the sender of the version negotiation packet is on-path) is a dup of #244.

MikeBishop:
#361 purports to address this; please re-open or file a new issue if that's incorrect.

Issue #293, opened 2017-02-12
Does the connection id need to be in a consistent location?


ianswett:
@martinthomson created a header proposal that provides a fixed 8 byte connection id in a fixed location.  

We need to decide if this is a requirement or a nice to have.

I haven't found any environments it needed to be in a fixed location for all packets, as long as it's easy to determine where it will be from the first byte of the packet.

martinthomson:
I thought that it was nice to have, but not a deal-breaker.

ianswett:
Agreed.

MikeBishop:
#361 purports to address this; please re-open or file a new issue if that's incorrect.

Issue #244, opened 2017-01-26
Need a NONCE in version negotiation packets


igorlord:
Now that the client does not send a connection id in its version negotiation (see #119), client needs a public nonce in the packets. That prevents an off-path attacker guessing client source port and interfering with the client establishing connections with a given sever by spoofing version negotiation failure packets as if coming from that sever. That nonce should be returned by the sever in its version negotiation packets. Previously, client's connection id served such a purpose.

ianswett:
The public reset has the same issue.

I'm not sure how serious an issue this is, since the attacker would also have to guess the server IP and port as well as the client port.  In practice, this would be a very costly attack, because it only works during the handshake.  Typically handshakes complete within a second, so to prevent a client from connecting to a single IP/port, you'd have to send thousands of packets a second, with no guarantee or acknowledgement of whether your attack was a success?

If we think it's a serious issue, we could also reconsider requiring a connection id from client to server on the initial flight.

martinthomson:
Connection ID isn't the only field that contains entropy.  As you say, IP and port do contain some entropy (though not much).  Requiring proof that the server received a packet doesn't have to rely on those fields alone.  That gets easier if you don't worry about version negotiation packets because you can put things in the TLS handshake (something that we might want to do anyway).

There really isn't anything very strong in the TLS ServerHello.  The messages that follow have a *very* strong signal that the server received the ClientHello though.

mirjak:
The packet number and a potential packet number echo from the server would help here as well.

MikeBishop:
#361 purports to address this; please re-open or file a new issue if that's incorrect.

PR #361, opened 2017-03-07
Long and short packet header

janaiyengar:
This PR proposes an alternative header format for QUIC.  It is based on a previous proposal and incorporates feedback received on the list. This PR is unfortunately quite invasive.

Closes:
#35: Starting packet number
#56: Extending flags
#119: Server-proposed connection ID
#133: Connection ID in version negotiation
#135: DoS using Version Negotiation Packets
#147: Reflection Attack Resistance
#185: Reliable identification of the initial packet for a connection
#193: Flags section is kind of confusing
#244: Need a NONCE in version negotiation packets
#293: Does the connection ID need to be in a consistent location
#295: Connection ID on a version negotiation packet

(also addresses some of:
#148: QUIC packet header complexity
#170: Connection ID collisions)

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -511,7 +511,7 @@ A Public Reset packet contains:
 * Octets 1-8: Echoed data (octets 1-8 of received packet)
 * Octets 9-12: Echoed data (octets 9-12 of received packet)
 * Octets 13-16: Version (server version)
-* Octets 17+: Public Reset Proof (optional, TBD)
+* Octets 17+: Public Reset Proof
 
 For a client that sends a connection ID on every packet, a Public Reset
 packet received in response can be simply interpreted as:
@@ -521,22 +521,18 @@ packet received in response can be simply interpreted as:
   on the packet number size used by the client, followed by 0, 2, or 3
   subsequent bytes from the client packet.
 * Octets 13-16: Version (server version)
-* Octets 17+: Public Reset Proof (optional, TBD)
+* Octets 17+: Public Reset Proof
 
-Whether the Public Reset Proof field is included in a Public Reset packet
-depends on the entity that generates the packet.
-
-A Public Reset packet sent by an endpoint indicates that it does not have the
-state necessary to continue with a connection.  In this case, the endpoint will
+A Public Reset packet sent by a server indicates that it does not have the
+state necessary to continue with a connection.  In this case, the server will
 include the fields that prove that it originally participated in the connection
 (see {{public-reset-proof}} for details).
 
-Upon receipt of a Public Reset packet that contains a valid proof, an endpoint
-MUST tear down state associated with the connection.  The endpoint MUST then
+Upon receipt of a Public Reset packet that contains a valid proof, a client
+MUST tear down state associated with the connection.  The client MUST then
 cease sending packets on the connection and SHOULD discard any subsequent
 packets that arrive. A Public Reset that does not contain a valid proof MUST be
 ignored.
-
 
 ### Public Reset Proof
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -339,7 +339,7 @@ describing the value of fields.
 QUIC packet headers can be separated into long and short headers. Long headers
 are expected to be used for the initial phase of the connection and are used for
 version negotiation, connection establishment, 0-RTT, and public reset
-packets. Short headers are minimal version-specific headers, which MAY be used
+packets. Short headers are minimal version-specific headers, which can be used
 after version negotiation and 1-RTT keys are established, and MUST NOT be used
 earlier.
 
@@ -381,7 +381,8 @@ The first octet (octet 0) contains the following fields.
   {{cleartext-packet}}.)
   * 05: Other Server Cleartext packet (see {{cleartext-packet}}.)
   * 06: 0-RTT Encrypted packet (see {{encrypted-packet}}.)
-  * 07: 1-RTT Encrypted packet (see {{encrypted-packet}}.)
+  * 07: 1-RTT Encrypted packet with key phase 0 (see {{encrypted-packet}}.)
+  * 08: 1-RTT Encrypted packet with key phase 1 (see {{encrypted-packet}}.)
 
 The remainder of the packet layout is the same regardless of type, but the
 semantics of the fields are specific to each type (see corresponding sections
@@ -409,8 +410,8 @@ assigned a packet number by the sender, as described further in
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ```
 
-The short header is used after the version and 1-RTT keys are negotiated. This
-header has the following version-independent fields:
+The short header can be used after the version and 1-RTT keys are negotiated.
+This header has the following version-independent fields:
 * Octet 0: Flags
   * Bit 0 (0x80): HEADER_FORM, set to 0 for short headers.
   * Bit 1: CONNECTION_ID. Indicates presence of Connection ID field following
@@ -421,8 +422,8 @@ The remainder of the short header is defined to be specific to a version. In
 this version, it contains:
 * Octet 0: Flags
   * Bit 2: KEY_PHASE. Used by the QUIC packet protection to identify
-    the correct packet protection keys after the 1-RTT keys are initially
-    established. See {{QUIC-TLS}}.
+    the correct packet protection keys after the 1-RTT keys are established. 
+    See {{QUIC-TLS}} for more details.
   * Bit 3-7: Packet Type, indicating one of 32 packet types. The following types
     are currently defined.
   * 01: 1-RTT packet (packet number size = 1)
@@ -578,12 +579,17 @@ The payload of Cleartext packets contains frames, as described in {{frames}}.
 Packets encrypted with either 0-RTT or 1-RTT keys may be sent with long headers.
 Different packet types explicitly indicate the encryption level for ease of
 decryption. These packets contain:
-* Octet 0: 0x86 or 0x87 (Flags indicating long header and one of the two 
-  encrypted packet types)
+* Octet 0: 0x86, 0x87 or 0x88 (Flags indicating long header and an encrypted 
+  packet type)
 * Octets 1-8: Connection ID (client or server-selected, see {{connectionid}})
 * Octets 9-12: Packet Number (low 4 octets)
 * Octets 13-16: Version
 * Octets 17+: Encrypted Payload (see {{encrypted-payload}}.)
+
+A flags octet of 0x86 indicates a 0-RTT packet. Key phases are used by the QUIC
+packet protection to identify the correct packet protection keys after the 1-RTT
+keys are established. The default initial value key phase is 0. See {{QUIC-TLS}}
+for more details.
 
 The encrypted payload is both authenticated and encrypted using packet
 protection keys. {{QUIC-TLS}} describes packet protection in detail.  After
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -441,8 +441,7 @@ client packet of an unsupported version. It uses a long header and contains:
 
 * Octet 0: 0x81 (Flags indicating long header and Version Negotiation packet 
   type)
-* Octets 1-8: Connection ID (server-selected value, may be used in a subsequent
-  connection to reach the same server)
+* Octets 1-8: Connection ID (echoed)
 * Octets 9-12: Proof (first 4 octets of client-selected connection ID)
 * Octets 13-16: Version (echoed)
 * Octets 17+: Payload (version list, containing 0 or more acceptable versions)
@@ -498,10 +497,11 @@ outage). A server that wishes to indicate fatal errors with a connection MUST
 use a CONNECTION_CLOSE frame in preference to Public Reset if it has sufficient
 state to do so.
 
-A public reset packet contains:
+A Public Reset packet contains:
 
 * Octet 0: 0x82 (Flags indicating long header and Public Reset packet type) 
-* Octets 1-12: Echoed data (octets 1-12 of received packet)
+* Octets 1-8: Echoed data (octets 1-8 of received packet)
+* Octets 9-12: Echoed data (octets 9-12 of received packet)
 * Octets 13-16: Version (server version)
 * Octets 17+: Public Reset Proof (optional, TBD)
 
@@ -559,7 +559,7 @@ Other server Cleartext packets contain:
 
 * Octet 0: 0x85 (Flags indicating Long header and appropriate server Cleartext 
   packet type)
-* Octets 1-8: Connection ID (client-selected value)
+* Octets 1-8: Connection ID (echoed)
 * Octets 9-12: Packet Number (low 4 octets, random 31-bit initial value)
 * Octets 13-16: Version (echoed)
 * Octets 17+: Payload
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -336,114 +336,170 @@ all field sizes are in bits.  When discussing individual bits of fields, the
 least significant bit is referred to as bit 0.  Hexadecimal notation is used for
 describing the value of fields.
 
+QUIC packet headers can be separated into long and short headers. Long headers
+are expected to be used for the initial phase of the connection and are used for
+version negotiation, connection establishment, 0-RTT, and public reset
+packets. Short headers are minimal version-specific headers, which MAY be used
+after version negotiation and 1-RTT keys are established, and MUST NOT be used
+earlier.
 
-## Common Header
+## Long Header
 
-All QUIC packets begin with a QUIC Common header, as shown below.
-
-~~~
+```
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+
-|   Flags (8)   |
+|1|   Type      |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                                                               |
-+                     [Connection ID (64)]                      +
++                        Connection ID                          +
 |                                                               |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                   Type-Dependent Fields (*)                 ...
+|                  Packet Number / Proof                        |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-~~~
+|                           Version                             |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                            Payload                          ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+```
 
-The fields in the Common Header are the following:
+Long headers are used for packets that are sent prior to the completion of
+version negotiation or are not encrypted with 1-RTT keys. Once both conditions
+are met, a sender SHOULD switch to sending short-form headers. While
+inefficient, long headers MAY be used for 1-RTT packets. The long form allows
+for special packets, such as the version negotiation and the public reset
+packets to be represented in this uniform fixed-length packet format.
 
-* Flags:
+The first octet (octet 0) contains the following fields.
+* Bit 0 (0x80): Header form. Set to 1 for long headers.
+* Bits 1-7: Packet Type, indicating one of 128 packet types. The following types
+  are currently defined.
+  * 01: Version Negotiation packet
+  * 02: Public Reset packet
+  * 03: 0-RTT packet
+  * 04: Server cleartext packet
+  * 05: Client cleartext packet
+  * 06: 1-RTT packet with version field, connection ID,
+        and 2-byte packet number
 
-   * 0x01 = VERSION.  The semantics of this flag depends on whether the packet
-     is sent by the server or the client.  A client MAY set this flag and
-     include exactly one proposed version.  A server may set this flag when the
-     client-proposed version was unsupported, and may then provide a list (0 or
-     more) of acceptable versions as a part of version negotiation (described in
-     {{version-negotiation}}.)
+The remainder of the packet layout is the same regardless of type, but the
+semantics of the fields are specific to each type, as described next.
 
-   * 0x02 = PUBLIC_RESET.  Set to indicate that the packet is a Public Reset
-     packet.
+### Version Negotiation Packet
 
-   * 0x04 = KEY_PHASE.  This is used by the QUIC packet protection to identify
-     the correct packet protection keys, see {{QUIC-TLS}}.
+A Version Negotiation packet is sent by the server in response to a client
+packet of an unsupported version. It contains:
 
-   * 0x08 = CONNECTION_ID.  Indicates the Connection ID is present in the
-     packet.  This must be set in all packets until negotiated to a different
-     value for a given direction.  For instance, if a client indicates that the
-     5-tuple fully identifies the connection at the client, the connection ID is
-     optional in the server-to-client direction. The negotiation is described in
-     {{transport-parameter-definitions}}.
+* Octets 1-8: Connection ID (server-selected value, may be used in a subsequent
+  connection to reach the same server)
+* Octets 9-12: Proof (first 4 octets of client-selected connection ID)
+* Octets 13-16: Version (echoed)
+* Octets 17+: Payload (version list, containing 0 or more acceptable versions)
 
-   * 0x30 = PACKET_NUMBER_SIZE.  These two bits indicate the number of
-     low-order-bytes of the packet number that are present in each packet.
+See {{version-negotiation}} for a description of the version negotiation
+process.
 
-     + 11 indicates that 6 bytes of the packet number are present
-     + 10 indicates that 4 bytes of the packet number are present
-     + 01 indicates that 2 bytes of the packet number are present
-     + 00 indicates that 1 byte of the packet number is present
+### Public Reset Packet
 
-   * 0x40 = MULTIPATH.  This bit is reserved for multipath use.
+A Public Reset packet is sent when the server has no state for a received
+packet. A server may therefore have to respond to either a long-form or a
+short-form packet. A public reset packet contains:
 
-   * 0x80 is currently unused, and must be set to 0.
+* Octets 1-12: Proof (octets 1-12 of received packet)
+* Octets 13-16: Version (server version)
 
-* Connection ID: An unsigned 64-bit random number chosen by the client, used as
-  the identifier of the connection.  Connection ID is tied to a QUIC connection,
-  and remains consistent across client and/or server IP and port changes.
+For a client that sends a connection ID on every packet, a Public Reset
+packet received in response can be simply interpreted as:
 
+* Octets 1-8: Connection ID (echoed)
+* Octets 9-12: Proof (echoed packet number, which could be 1, 2, or 4 bytes
+  depending on the packet number size used by the client, followed by 0, 2, or 3
+  subsequent bytes from the client packet)
+* Octets 13-16: Version (server version)
 
-### Identifying Packet Types
+### Client Cleartext Packet
 
-While all QUIC packets have the same common header, there are three types of
-packets: Regular packets, Version Negotiation packets, and Public Reset packets.
-The flowchart below shows how a packet is classified into one of these three
-packet types:
+A client cleartext packet may be sent during the handshake. It contains:
+* Octets 1-8: Connection ID (randomly chosen)
+* Octets 9-12: Packet number (low 4 octets, starts at a random 31-bit value)
+* Octets 13-16: Version
+* Octets 17+: Payload
 
-~~~
-Check the flags in the common header
-              |
-              |
-              V
-        +--------------+
-        | PUBLIC_RESET |  YES
-        | flag set?    |-------> Public Reset packet
-        +--------------+
-              |
-              | NO
-              V
-        +------------+         +-------------+
-        | VERSION    |  YES    | Packet sent |  YES     Version
-        | flag set?  |-------->| by server?  |--------> Negotiation
-        +------------+         +-------------+          packet
-              |                       |
-              | NO                    | NO
-              V                       V
-      Regular packet with       Regular packet with
-  no QUIC Version in header    QUIC Version in header
-~~~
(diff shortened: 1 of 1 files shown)


alagoutte:
LGTM

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+## Cleartext Packets {#cleartext-packet}
+
+Cleartext packets are sent during the handshake prior to key negotiation. A
+Client Cleartext packet contains:
+
+* Octet 0: 0x83 (Flags indicating long header and client Cleartext packet type)
+* Octets 1-8: Connection ID (randomly chosen)
+* Octets 9-12: Packet number (low 4 octets, starts at a random 31-bit value)
+* Octets 13-16: Version
+* Octets 17+: Payload
```
ekr: I feel like cleartext would be better to go first. Lets do the happy path.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+* Octets 9-12: Packet Number (low 4 octets, random 31-bit initial value)
+* Octets 13-16: Version (echoed)
+* Octets 17+: Payload
+
+The client MUST choose a random value and use it as the Connection ID until the
+server replies with a server-selected Connection ID. Server-selected Connection
+IDs are used after a successful handshake, cleanly distinguishing packets that
+use them from packets using client-selected Connection IDs. All packets
+including and following a Server Cleartext End-of-Handshake packet use a
+server-selected Connection ID, as described in {{connection-id}}.
```
ekr: I would consider not interleaving the description of the semantics and the packet format. The packet format tells you which bytes are bytes, so you just need the semantics here.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+contain a Connection ID. Short headers indicate the presence of a Connection ID
+using the CONNECTION_ID flag. When present, the Connection ID is in the same
+location in all packet headers, making it straightforward for middleboxes, such
+as load balancers, to locate and use it.
+
+When a connection is initiated, the client MUST choose a random value and use it
+as the Connection ID in all Cleartext packets and all 0-RTT Encrypted
+packets. The client's Connection ID is a suggestion to the server. The server
+echoes this value in all packets until the handshake is successful (see
+{{QUIC-TLS}}. In its final successful handshake completion packet, the server
```
ekr: "handshake completion packet" does not seem to be a defined term.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
-Each Regular packet contains additional header fields followed by an encrypted
-payload, as shown below:
+: The third bit (0x20) of octet 0 indicates the key phase, which allows a
+  recipient of a packet to identify the packet protection keys that are used to
+  protect the packet.  See {{QUIC-TLS}} for details.
+
+Short Packet Type:
+
+: The remaining 5 bits of octet 0 include one of 32 packet types.
+  {{short-packet-types}} lists the types that are defined for short packets.
```
MikeBishop: We could unify the spaces, but there will be types that are only valid for long (public reset, version negotiation), and possibly types that are only valid for short.  If we did unify them, though, we could reclaim the key phase bit and just use the same 1-RTT (KP0) and 1-RTT (KP1) types the long header defines.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
 
-* the location and size of the Flags field,
+```
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+
+|0|C|K|  Type   |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                                                               |
++                   Connection ID (optional)                    +
```
MikeBishop: We have a notation convention for optional fields:  This would be [Connection ID (64)].

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+The packet type in a short header currently determines only the size of the
+packet number field.  Additional types can be used to signal the presence of
+other fields.
+
+| Type | Packet Number Size     |
+|:-----|:-----------------------|
+| 01   | 1 octet                |
+| 02   | 2 octets               |
+| 03   | 4 octets               |
+{: #short-packet-types title="Short Header Packet Types"}
```
MikeBishop: What bit are they burning?  This appears to only be indicated by the type byte.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+
+The header form, connection ID flag and connection ID of a short header packet
+are version independent.  The remaining fields are specific to the selected QUIC
```
MikeBishop: Hyphenate:  version-independent.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
-* Packet Number: The lower 8, 16, 32, or 48 bits of the packet number, based on
-  the PACKET_NUMBER_SIZE flag.  Each Regular packet is assigned a packet number
-  by the sender.  The first packet number is randomized (see
-  {{initial-packet-number}}).
+* Octet 0: 0x81 (Flags indicating long header and Version Negotiation packet
+  type)
+* Octets 1-8: Connection ID (echoed)
+* Octets 9-12: Packet Number (echoed)
+* Octets 13-16: Version (echoed)
+* Octets 17+: Payload (version list, containing 0 or more acceptable versions)
```
MikeBishop: Yeah, it seems a little odd to allow the server to respond "I speak just enough QUIC to say 'I don't speak QUIC.'"

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                          Frame N (*)                        ...
+|                    Supported Version N (32)                 ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ~~~
-{: #regular-packet-frames title="Contents of Encrypted Payload"}
+{: #version-negotiation-format title="Version Negotiation Packet"}
+
+A server that generates a version negotiation packet SHOULD include one of the
+reserved version numbers (those following the pattern 0x?a?a?a?a).  This ensures
```
MikeBishop: "helps ensure." Nothing can ensure well-written clients.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+that clients are able to correctly handle an unsupported version.
+
+The design of version negotiation permits a server to avoid maintaining state
+for packets that it rejects in this fashion.  However, when the server generates
+a version negotiation packet, it cannot randomly generate a reserved version
+number.  This is because the server is required to include the same value in its
+transport parameters (see {{version-validation}}).  To avoid the selected
+version number changing during connection establishment, the reserved version
+can be generated as a function of values that will be available to the server
```
MikeBishop: SHOULD?

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+
+A client MAY send a packet using a reserved version number.  This can be used to
+solicit a list of supported versions from a server.
+
+See {{version-negotiation}} for a description of the version negotiation
+process.
+
+## Public Reset Packet {#public-reset-packet}
+
+A Public Reset packet is sent by only a server and is used to abruptly terminate
```
MikeBishop: "only sent by servers"

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+Cleartext packets are sent during the handshake prior to key negotiation. A
+Client Cleartext packet contains:
+
+* Octet 0: 0x83 (Flags indicating long header and client Cleartext packet type)
+* Octets 1-8: Connection ID (randomly chosen)
+* Octets 9-12: Packet number (low 4 octets, starts at a random 31-bit value)
+* Octets 13-16: Version
+* Octets 17+: Payload
+
+A Server Cleartext End-of-Handshake packet contains:
```
MikeBishop: You called this Final above, not End-of-Handshake.

draft-ietf-quic-transport.md
MikeBishop: +1.  Define the fields, then describe how they're used by reference to those fields, not the underlying bytes.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+Different packet types explicitly indicate the encryption level for ease of
+decryption. These packets contain:
+* Octet 0: 0x86, 0x87 or 0x88 (Flags indicating long header and an encrypted
+  packet type)
+* Octets 1-8: Connection ID (client or server-selected, see {{connection-id}})
+* Octets 9-12: Packet Number (low 4 octets)
+* Octets 13-16: Version
+* Octets 17+: Encrypted Payload
+
+A flags octet of 0x86 indicates a 0-RTT packet. Key phases are used by the QUIC
```
MikeBishop: You're pivoting away from 0-RTT without lead-in.  I'd suggest "...indicates a 0-RTT packet.  After the 1-RTT keys are established, key phases...."

draft-ietf-quic-transport.md
MikeBishop: And are there multiple successful handshake completion packets?  Let's reference the specific packet type.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+using the CONNECTION_ID flag. When present, the Connection ID is in the same
+location in all packet headers, making it straightforward for middleboxes, such
+as load balancers, to locate and use it.
+
+When a connection is initiated, the client MUST choose a random value and use it
+as the Connection ID in all Cleartext packets and all 0-RTT Encrypted
+packets. The client's Connection ID is a suggestion to the server. The server
+echoes this value in all packets until the handshake is successful (see
+{{QUIC-TLS}}. In its final successful handshake completion packet, the server
+MUST use a server-selected Connection ID. All subsequent packets from the server
```
MikeBishop: Wait, if the Connection ID is a suggestion to the server, but the server MUST use a server-selected ID, what's the point of the suggestion?

Perhaps that packet type MUST contain the final Connection ID, which MAY be the one proposed by the client, or MAY be a new server-selected value?

draft-ietf-quic-transport.md
```diff
@@ -541,12 +721,54 @@ packet number.  Once any packet has been acknowledged, subsequent packets can
 use a shorter packet number encoding.
 
 
-## Frames and Frame Types {#frames}
+## Handling Packets from Different Versions {#version-specific}
+
+Between different versions the following things are guaranteed to remain
+constant:
+
+* the location and size of the Flags field in both header forms,
```
MikeBishop: You didn't originally call that field Flags in this version, though I see it has cropped up other places through the document.

draft-ietf-quic-transport.md
```diff
@@ -541,12 +721,54 @@ packet number.  Once any packet has been acknowledged, subsequent packets can
...
+
+* the location of the HEADER_FORM bit in the Flags field in both header forms,
+
+* the location of the CONNECTION_ID bit in the Flags field in short headers,
+
+* the location and size of the Connection ID field in both header forms,
+
+* the location and value of the Version field in long headers, and
```
MikeBishop: The value of the Version field can't change between versions?  That's really going to screw with version negotiation logic....  :-)  I suspect this should be location and meaning.

draft-ietf-quic-transport.md
```diff
@@ -541,12 +721,54 @@ packet number.  Once any packet has been acknowledged, subsequent packets can
...
+
+* the location and value of the Packet Number field in long headers.
```
MikeBishop: Likewise.

draft-ietf-quic-transport.md
```diff
@@ -541,12 +721,54 @@ packet number.  Once any packet has been acknowledged, subsequent packets can
...
+
+Implementations MUST assume that an unsupported version uses an unknown packet
+format. All other values MUST be ignored when processing a packet that contains
```
MikeBishop: Other fields, not just other values.

draft-ietf-quic-transport.md
```diff
@@ -541,12 +721,54 @@ packet number.  Once any packet has been acknowledged, subsequent packets can
...
+an unsupported version.
+
+
+# Frames and Frame Types {#frames}
+
+The payload of cleartext packets and the plaintext after decryption of encrypted
```
MikeBishop: After decryption of the payload -- you're not decrypting the entire packet.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+| Type | Name                   | Section                        |
+|:-----|:-----------------------|:-------------------------------|
+| 01   | Version Negotiation    | {{version-negotiation-packet}} |
+| 02   | Public Reset           | {{public-reset-packet}}        |
+| 03   | Client Cleartext       | {{cleartext-packet}}           |
+| 04   | Final Server Cleartext | {{cleartext-packet}}           |
+| 05   | Other Server Cleartext | {{cleartext-packet}}           |
+| 06   | 0-RTT Encrypted        | {{encrypted-packet}}           |
+| 07   | 1-RTT Encrypted (key phase 0) | {{encrypted-packet}}    |
+| 08   | 1-RTT Encrypted (key phase 1) | {{encrypted-packet}}    |
```
ekr: Oh, I see that I missed this in the short header. nm.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
 least significant bit is referred to as bit 0.  Hexadecimal notation is used for
 describing the value of fields.
 
+QUIC packet headers can be separated into long and short headers. Long headers
```
janaiyengar: Done

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+are expected to be used for the initial phase of the connection and are used for
```
martinthomson: s/expected to be used/used/

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
-~~~
+|                           Version                             |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                            Payload                          ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+```
 
-The fields in the Common Header are the following:
+Long headers are used for packets that are sent prior to the completion of
+version negotiation or are not encrypted with 1-RTT keys. Once both conditions
```
martinthomson: "prior to the completion of both version negotiation and the cryptographic handshake "

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+: The most significant bit (0x80) of the first octet (octet 0) of a packet is
+  the header form.  This bit is set to 0 for the short header.
 
-All other values MUST be ignored when processing a packet that contains an
-unsupported version.
+Connection ID Flag:
 
+: The second bit (0x40) of octet 0 indicates whether the connection ID field is
+  present.  If set to 1, then the connection ID field is present; if set to 0,
+  the connection ID field is zero length.
```
martinthomson: same end result, no?

draft-ietf-quic-transport.md
martinthomson: I thinking that this is just @janaiyengar trolling us all.

draft-ietf-quic-transport.md
martinthomson: Agreed, in spades, though this isn't new text for this PR.

draft-ietf-quic-transport.md
martinthomson: Not really.  The requirement here is that the bogus version is constant.  This is only suggesting one way that a server might be implemented to comply with that requirement.

draft-ietf-quic-transport.md
martinthomson: Hmm, I don't remember this being a restriction on its use.  Why only servers?

draft-ietf-quic-transport.md
martinthomson: This is an artifact of me partially fixing the formatting.

draft-ietf-quic-transport.md
martinthomson: I had the same question as @MikeBishop here.  I believe that the answer is that the server sends the "Final" long packet type on all of the packets after ServerHello.  It would only echo the client's connection ID if it needed to send a Version Negotiation packet or a HelloRetryReqest.

draft-ietf-quic-transport.md
martinthomson: Yeah, I would just say any value.  The fact that the server might pick the value that the client did isn't particularly interesting.  If anything, we might point out that the ID doesn't have to be *new*, just unique.

draft-ietf-quic-transport.md
martinthomson: I would simply refer to this as the first octet, which sorts the both the "location" and "size" aspects.  Maybe:

> New versions of QUIC can be specified as long as the invariants in this section are maintained.
>
> The first octet always indicates the type of packet:
>  * The header form bit (0x80) always indicates a long or short header.
>  * In the short header, the next bit (0x40) indicates whether the Connection ID field is present.
>  * The remaining bits of this octet are interpreted as determined by the protocol version.
>
> Between all protocol versions, the following is guaranteed to be constant:
> * The location, size, and semantics of the Connection ID in both header forms (conditional on the field being present in the short form),
> * The location, size, and semantics of the Version field in the long form header, and
> * The location and size (but not semantics) of the Packet Number field in the long form header.

draft-ietf-quic-transport.md
martinthomson: And let us talk about "(successful) removal of packet protection"

draft-ietf-quic-transport.md
MikeBishop: That was my initial reaction as well.  A client implementation on top of a UDP stack will never receive a packet on a closed connection, because there's no UDP socket open there.  The UDP stack will drop packets before they reach QUIC.  However, a server will have a listening socket open and receive all traffic that's sent to that port -- new connections, existing connections, and stale connections.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
 least significant bit is referred to as bit 0.  Hexadecimal notation is used for
 describing the value of fields.
 
+QUIC packet headers can be separated into long and short headers. Long headers
+are expected to be used for the initial phase of the connection and are used for
+version negotiation, connection establishment, 0-RTT, and public reset
+packets. Short headers are minimal version-specific headers, which can be used
+after version negotiation and 1-RTT keys are established, and MUST NOT be used
+earlier.
```
janaiyengar: Done

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
-     is sent by the server or the client.  A client MAY set this flag and
-     include exactly one proposed version.  A server may set this flag when the
-     client-proposed version was unsupported, and may then provide a list (0 or
-     more) of acceptable versions as a part of version negotiation (described in
-     {{version-negotiation}}.)
+Header Form:
 
-   * 0x02 = PUBLIC_RESET.  Set to indicate that the packet is a Public Reset
-     packet.
+: The most significant bit (0x80) of the first octet (octet 0) is set to 1 for
```
janaiyengar: Done

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
-   * 0x30 = PACKET_NUMBER_SIZE.  These two bits indicate the number of
-     low-order-bytes of the packet number that are present in each packet.
+Connection ID:
 
-     + 11 indicates that 6 bytes of the packet number are present
-     + 10 indicates that 4 bytes of the packet number are present
-     + 01 indicates that 2 bytes of the packet number are present
-     + 00 indicates that 1 byte of the packet number is present
+: Octets 1 through 8 include a connection ID.  {{connection-id}} describes the
+  use of this field in more details.
```
janaiyengar: changed to "contain the connection ID"

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
 
-   * 0x40 = MULTIPATH.  This bit is reserved for multipath use.
+Packet Number:
 
-   * 0x80 is currently unused, and must be set to 0.
+: Octets 9 to 12 contain a packet number.  {{packet-numbers} describes the use
+  of packet numbers.
```
janaiyengar: Done

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
 
-* Connection ID: An unsigned 64-bit random number chosen by the client, used as
-  the identifier of the connection.  Connection ID is tied to a QUIC connection,
-  and remains consistent across client and/or server IP and port changes.
+Version:
 
+: Octets 13 to 16 contain the selected protocol version.  This field indicates
+  which version of QUIC is in use and determines how the some protocol fields
+  are interpreted.
```
janaiyengar: done.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
 
-### Identifying Packet Types
+Payload:
 
-While all QUIC packets have the same common header, there are three types of
-packets: Regular packets, Version Negotiation packets, and Public Reset packets.
-The flowchart below shows how a packet is classified into one of these three
-packet types:
+: Octets from 17 onwards are the payload of the packet.
```
janaiyengar: I've added "(the rest of the QUIC packet)" for consistency since we don't really need to talk about UDP datagram anywhere else after QUIC packet is defined.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
-{: #packet-types title="Types of QUIC Packets"}
+The following packet types are defined:
 
+| Type | Name                   | Section                        |
+|:-----|:-----------------------|:-------------------------------|
+| 01   | Version Negotiation    | {{version-negotiation-packet}} |
+| 02   | Public Reset           | {{public-reset-packet}}        |
+| 03   | Client Cleartext       | {{cleartext-packet}}           |
+| 04   | Final Server Cleartext | {{cleartext-packet}}           |
+| 05   | Other Server Cleartext | {{cleartext-packet}}           |
```
janaiyengar: Changed consistently to Final and Non-Final

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+| 06   | 0-RTT Encrypted        | {{encrypted-packet}}           |
+| 07   | 1-RTT Encrypted (key phase 0) | {{encrypted-packet}}    |
+| 08   | 1-RTT Encrypted (key phase 1) | {{encrypted-packet}}    |
+{: #long-packet-types title="Long Header Packet Types"}
 
-### Handling Packets from Different Versions
+The header form, long packet type, connection ID, packet number and version
```
janaiyengar: I've added a sentence saying the packet types defined here are version-specific. I'm not sure about this though, so I've added a TODO as well.

draft-ietf-quic-transport.md
janaiyengar: As mentioned on the list, and in agreement with Mike, there's value in having the packet be self-descriptive for connection ID.

draft-ietf-quic-transport.md
janaiyengar: Yeah, there will be holes or overlap between the header types. I don't see the value in unifying since there's got to be a branch before reading the types about which bits to read from the packet anyways, and two small maps aren't really that expensive to maintain.

draft-ietf-quic-transport.md
janaiyengar: Good point. Changed to 1 or more.

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+a version negotiation packet, it cannot randomly generate a reserved version
+number.  This is because the server is required to include the same value in its
+transport parameters (see {{version-validation}}).  To avoid the selected
+version number changing during connection establishment, the reserved version
+can be generated as a function of values that will be available to the server
+when later generating its handshake packets.
+
+A pseudorandom function that takes client address information (IP and port) and
+the client selected version as input would ensure that there is sufficient
+variability in the values that a server uses.
```
janaiyengar: moved this down to version negotiation section in a subsection called "Using Reserved Versions"

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+
+A client MAY send a packet using a reserved version number.  This can be used to
+solicit a list of supported versions from a server.
```
janaiyengar: moved

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+
+Upon receipt of a Public Reset packet that contains a valid proof, a client
+MUST tear down state associated with the connection.  The client MUST then
+cease sending packets on the connection and SHOULD discard any subsequent
+packets that arrive. A Public Reset that does not contain a valid proof MUST be
+ignored.
+
+### Public Reset Proof
+
+Details to be added.
```
janaiyengar: done

draft-ietf-quic-transport.md
janaiyengar: I'm doing it in the order that the packet types are defined. I've moved the types and the sections to have happy path first.

draft-ietf-quic-transport.md
janaiyengar: Simplified all this significantly, and moved descriptions down to prose below the layouts.

draft-ietf-quic-transport.md
janaiyengar: clarified and changed text. Yes there can be several Final packets, and the server-selected value is used for all of them.

draft-ietf-quic-transport.md
janaiyengar: There's no way to send those packets otherwise, so this is definitional. I don't think we need to mandate behavior when no alternative is present.

draft-ietf-quic-transport.md
janaiyengar: changed to "omitted"

draft-ietf-quic-transport.md
janaiyengar: changing to "sent only by a server"

draft-ietf-quic-transport.md
```diff
@@ -322,175 +322,356 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+See {{version-negotiation}} for a description of the version negotiation
+process.
+
+## Public Reset Packet {#public-reset-packet}
+
+A Public Reset packet is sent by only a server and is used to abruptly terminate
+communications. Public Reset is provided as an option of last resort for a
+server that does not have access to the state of a connection.  This is intended
+for use by a server that has lost state (for example, through a crash or
+outage). A server that wishes to indicate fatal errors with a connection MUST
```
janaiyengar: done

draft-ietf-quic-transport.md
martinthomson: It's not definitional if you accept your example of out-of-band keying.

martinthomson:
I think that we should just land this now:

land it

draft-ietf-quic-transport.md
janaiyengar: Not sure what this comment means

martinthomson: Can't help you there.

janaiyengar:
I've addressed all comments, PTAL. I hope to land this Thu/Fri, and I hope it's state is better than mt's plane :-)

draft-ietf-quic-transport.md
```diff
@@ -336,175 +336,327 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
+
+## Connection ID {#connection-id}
+
+QUIC connections are identified by their 64-bit Connection ID. All long headers
+contain a Connection ID. Short headers indicate the presence of a Connection ID
+using the CONNECTION_ID flag. When present, the Connection ID is in the same
+location in all packet headers, making it straightforward for middleboxes, such
+as load balancers, to locate and use it.
 
+When a connection is initiated, the client MUST choose a random value and use it
```
ekr: What is the purpose of this being random?

draft-ietf-quic-transport.md
MikeBishop: Ekr had a comment (deleted?) that talked about both burning a bit _and_ using a frame type to communicate... something.  Packet number length, maybe?  My comment was that there no longer appears to be a dedicated bit, so I think his comment was invalid.  If he deleted it subsequently, apparently he thought so too.

draft-ietf-quic-transport.md
janaiyengar: The client then provides a suggestion that may be reasonably for servers to simply use. Only a server that cares about structure in the ID changes it, others use the clients suggested value as is.

janaiyengar:
Mike, you still have "Changes Requested".  I've addressed your comments though, so I'll go ahead and merge. Feel free to let me know if there's more you'd like me to do, happy to do it in a follow up PR.

Issue #359, opened 2017-03-06
IANA registries for error codes


MikeBishop:
Since we've split these up and they're composable, we probably need three registries here.  (Host-local doesn't get a registry, obviously.)

MikeBishop:
OBE.  The error handling changes have incorporated appropriate IANA registries and registrations.

PR #274, opened 2017-02-08
Restrict RST to data streams, require replay

MikeBishop:
Fixes #176, but I'm hoping we can find a cleaner way to do this.

PR #363, opened 2017-03-07
HTTP/2 Guidance and Extensions

### LINKED ISSUE #242: HTTP extension mechanisms
The HTTP/QUIC mapping document is missing consideration for HTTP/2 extension mechanisms. This is part of the charter text:

> This mapping will accommodate the extension mechanisms defined in the HTTP/2 specification.

RFC 7540 (https://tools.ietf.org/html/rfc7540#section-5.5) defines three extension mechanisms, lets order these as:

1. New frame types
2. New error codes
3. New settings

I can see how 1 and 2 might work out, following Mike's example of augmenting existing definitions and allowing for things that exist in the HTTP/2 and HTTP/QUIC spaces to be managed somewhat independently. A concrete use case I have is the ALTSVC frame (especially when used on stream 0), so I'm happy to contribute towards this issue with some guidance.

3 has some problems related to proposed SETTINGS changes.

Finally, extensions that could change of semantics are a bit problematic (does it change the HTTP or the QUIC semantic, if that either of those is even possible). Additionally,  the negotiation of these is trickier given the additional negotation concerns of QUIC (0-RTT etc).

MikeBishop:
Collates the HTTP/2 migration guidance from various places into a single section, and expands on details about how HTTP/2 extensions can be used/ported.  In the process, I added more IANA stuff, which should address the HTTP portion of #359.

Builds on #274; Closes #242.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -758,7 +758,6 @@ HTTP_MULTIPLE_SETTINGS (0x10):
 HTTP_RST_CONTROL_STREAM (0x11):
 : A message control stream closed abruptly.
 
-<<<<<<< HEAD
 
 # Considerations for Transitioning from HTTP/2
 
@@ -839,9 +838,6 @@ to the IANA registry in {{iana-settings}}.
 QUIC has the same concepts of "stream" and "connection" errors that HTTP/2
 provides. However, because the error code space is shared between multiple
 components, there is no direct portability of HTTP/2 error codes.
-=======
-## Mapping HTTP/2 Error Codes
->>>>>>> origin/master
 
 The HTTP/2 error codes defined in Section 7 of {{!RFC7540}} map to QUIC error
 codes as follows:
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1078,6 +1078,8 @@ The original authors of this specification were Robbie Shade and Mike Warres.
 
 - 0-RTT guidance added
 
+- Guide to differences from HTTP/2 and porting HTTP/2 extensions added
+
 ## Since draft-ietf-quic-http-00:
 
 - Changed "HTTP/2-over-QUIC" to "HTTP/QUIC" throughout
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -762,7 +762,7 @@ HTTP_RST_CONTROL_STREAM (0x11):
 # Considerations for Transitioning from HTTP/2
 
 HTTP/QUIC is strongly informed by HTTP/2, and bears many similarities.  This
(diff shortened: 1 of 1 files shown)


LPardue:
This looks pretty good, I have a comment about ALTSVC that I will make on #242.

draft-ietf-quic-http.md
```diff
@@ -782,7 +758,86 @@ HTTP_MULTIPLE_SETTINGS (0x10):
 HTTP_RST_CONTROL_STREAM (0x11):
 : A message control stream closed abruptly.
 
-## Mapping HTTP/2 Error Codes
+
+# Considerations for Transitioning from HTTP/2
+
+HTTP/QUIC is strongly informed by HTTP/2, and bears many similarities.  This
+section points out useful differences from HTTP/2 and describes how to map
```
martinthomson: Not all the differences are useful.  I would use the word "important".

draft-ietf-quic-http.md
```diff
@@ -782,7 +758,86 @@ HTTP_MULTIPLE_SETTINGS (0x10):
...
+not required.
+
+Frame payloads are largely drawn from {{!RFC7540}}. However, QUIC includes some
+features (e.g. flow control) which are also present in HTTP/2. In these cases,
+the HTTP mapping does not re-implement them. As a result, some frame types are
+not required in HTTP/QUIC. Where an HTTP/2-defined frame is no longer used, the
+frame ID has been reserved in order to maximize portability between HTTP/2 and
+HTTP/QUIC implementations. However, equivalent frames between the two mappings
+are not identical.
+
```
martinthomson: Another important difference is that while there is an absolute ordering between every frame in h2, this is not true for frames on different streams in QUIC.  You mention this below, but only obliquely.  I think that it's important enough for a paragraph.

draft-ietf-quic-http.md
```diff
@@ -782,7 +758,86 @@ HTTP_MULTIPLE_SETTINGS (0x10):
...
+references to the definition for each frame type in HTTP/2 and in HTTP/QUIC.
+Frames not defined as available in HTTP/QUIC SHOULD NOT be sent and SHOULD be
+ignored as unknown on receipt.
+
+## HTTP/2 SETTINGS Parameters
+
+Some transport-level options that HTTP/2 specifies via the SETTINGS frame are
+superseded by QUIC transport parameters in HTTP/QUIC. The HTTP-level options
+that are retained in HTTP/QUIC have the same value as in HTTP/2.
+
```
martinthomson: Another important difference: settings can't change.

MikeBishop:
Addressed Martin's feedback, and consolidated more text into the new section.  If we want to switch to a new IANA registry, it's a fairly small step from here.

PR #365, opened 2017-03-07
Count line length differently

martinthomson:
I need to check this on my mac.


martinthomson:
@janaiyengar, this should work for you.  I have also changed this so that it reports exactly which line is too long, including a line number.  That should make it easier to find problems.

MikeBishop:
I think this reverts the change I made to exempt tables from the length check, though.  Kramdown requires the entire table row to be on a single line, even if it winds up wrapping the table cell contents itself later.

PR #358, opened 2017-03-06
Remove STOP_WAITING frame

ianswett:
First pass at removing the STOP_WAITING frame.  Intends to resolve #66 .

PR #366, opened 2017-03-07
Don't ACK the ACK

### LINKED ISSUE #291: ACKing ACK
The text says:

> However, since it is possible that an endpoint might only send packets containing ACK frames (or other non-retransmittable frames), the receiving peer MAY send an ACK frame after a reasonable number (currently 20) of such packets have been received.

"currently 20" seems pretty arbitrary.  That smells like something from the Google QUIC implementation.  We should be more concrete about this, if that behaviour even makes sense.

If an endpoint is receiving nothing but ACKs but is sending itself, then acknowledging those ACKs is probably fine.

An endpoint with nothing to say can generate a lot of ACKs over time, but if both endpoints are being quiet other than ACKs, it's probably time to shut up entirely.  Maybe we should say that the peer SHOULD generate an ACK frame only if it has other frame types to send.

Limiting the number of PING frames that a peer will respond to is a separate issue, and may motivate a limit with a number attached to it.

martinthomson:
Closes #291.

Note that #358 contains advice on using PING to elicit an ACK if the endpoint wants to reduce the range of packet numbers that its acknowledgments cover.  We can rely on that same advice.

@ianswett, I don't know how you want to manage this and #358.  If you want to fold this into #358, that might be best.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1665,12 +1665,9 @@ The details of loss detection and congestion control are described in
 {{QUIC-RECOVERY}}.
 
 A receiver acknowledges receipt of a received packet by sending one or more ACK
-frames containing the packet number of the received packet.  To avoid perpetual
-acknowledgment between endpoints, a receiver MUST NOT generate an ACK frame in
-response to every packet containing only ACK frames.  However, since it is
-possible that an endpoint might only send packets containing ACK frames (or
-other non-retransmittable frames), the receiving peer MAY send an ACK frame
-after a reasonable number (currently 20) of such packets have been received.
+frames containing the packet number of the received packet.  To avoid creating
+an indefinite feedback loop, an endpoint MUST NOT generate an ACK frame in
+response to a packet containing only ACK and PADDING frames.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1665,12 +1665,9 @@ The details of loss detection and congestion control are described in
...
 A receiver acknowledges receipt of a received packet by sending one or more ACK
-frames containing the packet number of the received packet.  To avoid perpetual
-acknowledgment between endpoints, a receiver MUST NOT generate an ACK frame in
-response to every packet containing only ACK frames.  However, since it is
-possible that an endpoint might only send packets containing ACK frames (or
-other non-retransmittable frames), the receiving peer MAY send an ACK frame
-after a reasonable number (currently 20) of such packets have been received.
+frames containing the packet number of the received packet.  To avoid creating
+an indefinite feedback loop, an endpoint MUST NOT generate an ACK frame in
+response to a packet containing only ACK and PADDING frames.
```
ianswett: nit: ACK and/or PADDING frames?

PR #367, opened 2017-03-07
Packet number echo with variable-length numbering

britram:
This PR replaces #351, (manually) rebased on #361. It adds packet number echo with the echo taking the same size as the number, with correct guidance as to the packet number echo size (thanks @martinthomson).


britram:
Replaced by (manually-rebased) #391

PR #368, opened 2017-03-07
Packet number echo with fixed-length, 32-bit packet and echo numbers

britram:
This is a version of #367 (replacing #351 manually rebased on #361), that sets packet numbers and packet number echoes in short headers to 32-bits, mainly by cutting complexity out of #367.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -426,9 +426,9 @@ semantics are described in {{version-negotiation-packet}},
 +                        [Connection ID]                        +
 |                                                               |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                     Packet Number (1/2/4)                     |
+|                         Packet Number                         |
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -322,268 +322,192 @@ all field sizes are in bits.  When discussing individual bits of fields, the
...
-{{?RFC4086}} provides guidance on the generation of random values.
+Encrypted Payload:
 
-The first set of packets sent by an endpoint MUST include the low 32-bits of the
-packet number.  Once any packet has been acknowledged, subsequent packets can
-use a shorter packet number encoding.
+: Packets with a short header always include a 1-RTT protected payload.
 
+The packet type in a short header currently determines only the size of the
+packet number field.  Additional types can be used to signal the presence of
```
martinthomson: This first sentence no longer applies.  Probably just want to say:

> The packet type is used to determine the fields that are present in a short header and their length.

britram:
Replaced by #393.

martinthomson:
@britram, it's OK to force push.

PR #374, opened 2017-03-09
Send a TLS alert when TLS encounters an error

### LINKED ISSUE #272: Signaling TLS handshake failure
This needs to be written down.  I think that the best outcome would be to include BOTH a TLS alert and a QUIC error code (in a CONNECTION_CLOSE frame).

martinthomson:
Also rationalize the TLS error messages, which means removing most of them.

Closes #272


seanturner:
Since we got both TLS and QUIC connections, I figured it might better to be clear about which one the sentence is referring to.  Might be a bit over the to say add "TLS" before alert, but with it there's no ambiguity.

martinthomson:
Thanks Sean.

PR #376, opened 2017-03-09
Separate IANA registries

### LINKED ISSUE #81: Split from HTTP/2 framing?
Because of a stated goal from some implementers to reuse HTTP/2 framing as much as possible, the draft currently augments the HTTP/2 Frame Type Registry with a "Supported Protocol" column, allowing a frame to be defined for HTTP/2, HTTP/QUIC, or both.  A specification is required for each supported protocol.

However, in the current draft, all but one "Both" frames are *different* between HTTP/2 and HTTP/QUIC.  Small or large, nearly every frame has required some modification to work over QUIC.  If we accept #80 , *all* frames will be different between HTTP/2 and HTTP/QUIC.

At that point, is there really a point in continuing to define HTTP/QUIC in relation to HTTP/2, or should we just admit that it's a new, separate thing that's heavily inspired by HTTP/2?  Separate registry, etc., and future extensions can register themselves with both if appropriate?  Or do we keep the same registry anyway to enforce that you'll never wind up with totally unrelated frames sharing the same value across the two protocols?

MikeBishop:
Goes a step further than #363 by establishing separate IANA registries for HTTP/2 and HTTP/QUIC.

Fixes #81.


martinthomson:
LGTM.  Obviously others disagree.

MikeBishop:
Added a brief mention of principles -- similarity to HTTP/2 is a goal, joined-at-the-hip is not, differences should always be noted in the HTTP/2 Transition section.  However, I had to weasel-word for one thing:  We depart from HTTP/2 **primarily** where QUIC is different from TCP.

The other situation I expressed in the meeting was "when the HTTP WG tells us they wish HTTP/2 were different."  This is the case with the change to SETTINGS, which wasn't adopted by the HTTP WG as an extension, but was recommended for inclusion in HTTP/QUIC.  I have trouble describing "HTTP WG asked us to" without "shipping the org structure," as the saying goes.

Suggestions welcome on how to phrase that; otherwise it's staying "primarily".

draft-ietf-quic-http.md
```diff
@@ -794,10 +802,9 @@ CONTINUATION (0x9):
 : CONTINUATION frames do not exist; instead, larger HEADERS/PUSH_PROMISE
   frames than HTTP/2 are permitted, and HEADERS frames can be used in series.
 
-The IANA registry of frame types has been updated in {{iana-frames}} to include
-references to the definition for each frame type in HTTP/2 and in HTTP/QUIC.
-Frames not defined as available in HTTP/QUIC SHOULD NOT be sent and SHOULD be
-ignored as unknown on receipt.
+Frame types defined by extensions to HTTP/2 need to be re-registered for
```
martinthomson: s/re-registered/separately registered/

draft-ietf-quic-http.md
```diff
@@ -716,8 +716,16 @@ HTTP_RST_CONTROL_STREAM (0x11):
...
-section points out important differences from HTTP/2 and describes how to map
-HTTP/2 extensions into HTTP/QUIC.
+section describes the approach taken to design HTTP/QUIC, points out important
+differences from HTTP/2, and describes how to map HTTP/2 extensions into
+HTTP/QUIC.
+
+HTTP/QUIC begins from the premise that HTTP/2 code reuse is a useful feature,
+but not a hard requirement.  HTTP/QUIC departs from HTTP/2 primarily where
+necessary to accommodate the differences in behavior between QUIC and TCP (lack
+of ordering, support for streams).
```
martinthomson: I like the primarily here.  We don't need to make a hard commitment.

I think that you want to explicitly state that it is our intent to avoid gratuitous changes.  Especially those that might make it difficult or impossible to build extensions across both protocols that have the same semantics.

You might also observe that an extension that is valid in HTTP/QUIC is usually valid in HTTP/2, though not the converse.

draft-ietf-quic-http.md
```diff
@@ -834,10 +840,10 @@ SETTINGS_MAX_FRAME_SIZE:
 SETTINGS_MAX_HEADER_LIST_SIZE:
 : See {{settings-parameters}}.
 
-Settings defined by extensions to HTTP/2 MAY be expressed as integers with a
-maximum value of 2^32-1, if they are applicable to HTTP/QUIC, but SHOULD have a
-specification describing their usage.  Fields for this purpose have been added
-to the IANA registry in {{iana-settings}}.
+Settings defined by extensions to HTTP/2 need to be re-registered for HTTP/QUIC
+if still applicable.  The IDs of settings defined in {{!RFC7540}} have been
```
martinthomson: In the spirit of the above comment:

> Settings need to be defined for HTTP/QUIC and HTTP/2 separately.

Issue #340, opened 2017-02-23
Scale flow control offsets


martinthomson:
@ddragana observes that the 64-bit offset value in the WINDOW_UPDATE message is probably the first value that we will run out of (over packet numbers and stream numbers, which should last a bit longer in most cases).  We could avoid this by multiplying these values by a constant (she suggests 1024), which would reduce the granularity with which flow control credits could be granted, but increase the total lifetime of a long-lived connection.

So, trade-off time.  The scaling multiplier is now 1.  A larger value allows for longer-lived connections, a shorter value allows finer control over flow control.  I'd suggest that values in the order of 8 are probably easy to justify, can we go as high as 1024?

We could also choose a different value for the connection and stream flow control offsets on that basis that connections need more octets, whereas streams do not.

ddragana:
stream flow control does not need scaling. It is absolute offset.

martinthomson:
Well, both are absolute, but you are right, there isn't much value in increasing the amount of data available to individual streams.

martinthomson:
Closed by #378.

Issue #192, opened 2017-01-23
Stop pretending we are going to get to a 64-bit packet number


ekr:
S 5.2.1 contains a bunch of instructions for how to deal with packet number 2^64 but (for instance) the connection flow control window is 64-bits in absolute values, so given that we're mostly sending data, it seems improbable you will get to 2^64 packets.

martinthomson:
Should we increase the length of the flow control limits to allow for >2^64 then?

ekr:
That would also be fine.

martinthomson:
#378 would bring this limit back into play.

martinthomson:
Closes with #378.

PR #378, opened 2017-03-09
Flow control accounting uses the maximum data offset on each stream

### LINKED ISSUE #370: Out-of-order Flow Control
ILubashev's suggestion on #171 alludes to this; separating it out as a new issue.

Connection-level flow control is currently defined as a cumulative sum of all bytes _received_ on all streams.  However, if you receive STREAM frames out of order, you can know that more data is in flight, even if you haven't received it.  Currently, that data implicitly counts against the receiver's stream window (since it's a max offset), but not against the connection window.

This would allow the receiver of a DISINTEREST frame to silently stop retransmission if the STREAM w/ FIN had already been acknowledged (without generating a RST_STREAM) since the flow control states on both sides would be consistent.  However, it would also require more logic on STREAM receipt, since it's no longer just a running counter of the size of all received STREAM frames.

martinthomson:
Also, scales the connection-level flow control window by 1024

Closes #370, #340, #192.

Note: I was just going to fix these separately, but they both hit exactly the same text.  Seemed like I might as well kill two birds.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -865,8 +865,8 @@ connection_fc_offset (0x0001):
   of 1024 octets.  That is, the value here is multiplied by 1024 to determine
   the actual flow control offset.  The sender of this parameter sets the byte
   offset for connection level flow control to this value.  This is equivalent to
-  sending a WINDOW_UPDATE ({{frame-window-update}}) for stream 0 immediately
-  after completing the handshake.
+  sending a WINDOW_UPDATE ({{frame-window-update}}) for the connection
+  immediately after completing the handshake.
 
 concurrent_streams (0x0002):
 
@@ -1545,10 +1545,10 @@ The connection-level flow control offset is expressed in units of 1024 octets
 offset is determined by multiplying the encoded value by 1024.
 
 An endpoint accounts for the maximum offset of data that is sent or received on
-a stream.  Loss or reordering can mean that the size of STREAM frames can be
-less than the maximum data offset on a stream; similarly, STREAM frames might
-not cause the maximum offset on a stream.  A STREAM frame with a FIN bit set or
-RST_STREAM causes the final offset for a stream to be fixed.
+a stream.  Loss or reordering can mean that the maximum offset is greater than
+the total size of data received on a stream.  Similarly, receiving STREAM frames
+might not increase the maximum offset on a stream.  A STREAM frame with a FIN
+bit set or RST_STREAM causes the final offset for a stream to be fixed.
 
 The maximum data offset on a stream MUST NOT exceed the stream flow control
 offset advertised by the receiver.  The sum of the maximum data offsets of all
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -854,17 +854,19 @@ TransportParameters:
 stream_fc_offset (0x0000):
 
 : The initial stream level flow control offset parameter is encoded as an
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1530,12 +1532,31 @@ Stream ID:
...
+The flow control offset is expressed in units of octets for individual streams
+(for stream identifiers other than 0).
+
+The connection-level flow control offset is expressed in units of 1024 octets
+(for a stream identifier of 0).  That is, the connection-level flow control
+offset is determined by multiplying the encoded value by 1024.
+
+An endpoint accounts for the maximum offset of data that is sent or received on
+a stream.  Loss or reordering can mean that the size of STREAM frames can be
+less than the maximum data offset on a stream; similarly, STREAM frames might
```
marten-seemann: What does the size of a STREAM frame have to do with the maximum offset?

draft-ietf-quic-transport.md
```diff
@@ -854,17 +854,19 @@ TransportParameters:
...
-  connection flow control window encoded as an unsigned 32-bit integer.  The
-  sender of this parameter sets the byte offset for connection level flow
-  control to this value.  This is equivalent to sending a WINDOW_UPDATE
-  ({{frame-window-update}}) for stream 0 immediately after completing the
-  handshake.
+  connection flow control window encoded as an unsigned 32-bit integer in units
+  of 1024 octets.  That is, the value here is multiplied by 1024 to determine
+  the actual flow control offset.  The sender of this parameter sets the byte
+  offset for connection level flow control to this value.  This is equivalent to
+  sending a WINDOW_UPDATE ({{frame-window-update}}) for stream 0 immediately
```
marten-seemann: Better: for the connection. There is no stream 0.

draft-ietf-quic-transport.md
MikeBishop: "...can mean that the maximum offset is greater than the total size of data received on a stream."

draft-ietf-quic-transport.md
```diff
@@ -1530,12 +1532,31 @@ Stream ID:
...
+(for stream identifiers other than 0).
+
+The connection-level flow control offset is expressed in units of 1024 octets
+(for a stream identifier of 0).  That is, the connection-level flow control
+offset is determined by multiplying the encoded value by 1024.
+
+An endpoint accounts for the maximum offset of data that is sent or received on
+a stream.  Loss or reordering can mean that the size of STREAM frames can be
+less than the maximum data offset on a stream; similarly, STREAM frames might
+not cause the maximum offset on a stream.  A STREAM frame with a FIN bit set or
```
MikeBishop: "...might not cause the maximum offset on a stream **to increase**."

draft-ietf-quic-transport.md
```diff
@@ -1530,12 +1532,31 @@ Stream ID:
 : ID of the stream whose flow control windows is being updated, or 0 to specify
   the connection-level flow control window.
 
-Byte offset:
+Flow Control Offset:
+
+: A 64-bit unsigned integer indicating the flow control offset for the given
+  stream (for a stream ID other than 0) or the entire connection.
+
+The flow control offset is expressed in units of octets for individual streams
```
janaiyengar: Why not use a 1K multiplier for both stream and connection-level?

martinthomson: I considered it, but streams don't need as much space and they might benefit from a more granular expression of where they are up to.  We can refine this later.

PR #380, opened 2017-03-09
Mostly-minor header touchups

MikeBishop:
This should be almost all editorial -- wording, formatting, and figure consistency changes in the first commit.

While doing those, I noticed that the draft still talks about a "VERSION flag," but the header no longer has one, so I fixed those.  Please review this for technical accuracy.


martinthomson:
BTW, @MikeBishop, the commit hook doesn't check line length.  It's a failing of the way that the checks are added.  I can maybe restructure things, but it is a little complicated.

Issue #200, opened 2017-01-23
Race condition between stream creation and MSPC


ekr:
Consider the following sequence of events on a stream (assuming that MSPC allows exactly one application data stream in the C-S direction:

```
<---------- Stream Establishment ----------->
<----------------------------------------- FIN
FIN ----------------------------------------->
DATA (stream N+1) -------------------------->
```

If you get reordering between the FIN and the data then the server will see it as exceeding MSPC and send  QUIC_TOO_MANY_OPEN_STREAMS. Ugh.

Aron-Schats:
The client should not do that:

https://groups.google.com/a/chromium.org/forum/#!topic/proto-quic/27ee0fWisTc

ekr:
Well, that needs to go in the spec then.

martinthomson:
This was addressed in #421.  Reopen if you disagree.

PR #381, opened 2017-03-09
ACK means all frames processed; closed means all frames ACK'd

### LINKED ISSUE #341: What does it mean to acknowledge something?
We're starting to see a few cases where ACK frames are being used to close the loop on things.  But that raises questions like "if a peer acknowledges a packet containing a GOAWAY, does that mean I can assume that stream above the mentioned number are now closed at their end?"

I think that we need to state what the rule is, and it should be simple.  An acknowledgement for a packet implies that all QUIC-level control messages have been read and processed.

This includes PING, ACK, RST_STREAM, GOAWAY, etc...  I think that it also has to include FIN bits on STREAM frames.  That implies that the *transport* must have finished handling all the messages.

The application protocol might not have consumed the data from STREAM frames, but that is the only thing that is considered outstanding when a packet is acknowledged.  We have a special sort of acknowledgment for data: flow control credits.

If we add features that include special acknowledgement (like flow control), then those would be excluded, but that would be explicitly stated.  Otherwise, if it is in the transport and it is acknowledged, then it has been processed by the receiver.

MikeBishop:
Fixes #341, #200 .

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1364,9 +1364,10 @@ transmission efficiency to underfilled packets.
 
 ## ACK Frame {#frame-ack}
 
-Receivers send ACK frames to inform senders which packets they have received, as
-well as which packets are considered missing.  The ACK frame contains between 1
-and 256 ACK blocks.  ACK blocks are ranges of acknowledged packets.
+Receivers send ACK frames to inform senders which packets they have received and
+processed, as well as which packets are considered missing.  The ACK frame
+contains between 1 and 256 ACK blocks.  ACK blocks are ranges of acknowledged
+packets.
 
 To limit ACK blocks to those that have not yet been received by the sender, the
 receiver SHOULD track which ACK frames have been acknowledged by its peer.  Once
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2021,11 +2027,13 @@ frame with the FIN flag set can cause two state transitions.  When the FIN flag
...
-different when frames are in transit.  Endpoints do not coordinate the creation
-of streams; they are created unilaterally by either endpoint.  The negative
-consequences of a mismatch in states are limited to the "closed" state after
-sending RST_STREAM, where frames might be received for some time after closing.
+The recipient of a frame which changes stream state will have a delayed view of
+the state of a stream while the frame is in transit.  Endpoints do not
+coordinate the creation of streams; they are created unilaterally by either
+endpoint.  The negative consequences of a mismatch in states are limited to the
+"closed" state after sending RST_STREAM, where frames might be received for some
+time after closing.  Endpoints SHOULD use acknowledgments to understand the
```
martinthomson: s/SHOULD/can

draft-ietf-quic-transport.md
```diff
@@ -2061,8 +2069,9 @@ From this state, either endpoint can send a frame with the FIN flag set, which
 causes the stream to transition into one of the "half-closed" states.  An
 endpoint sending an FIN flag causes the stream state to become "half-closed
 (local)".  An endpoint receiving a FIN flag causes the stream state to become
-"half-closed (remote)"; the receiving endpoint MUST NOT process the FIN flag
-until all preceding data on the stream has been received.
+"half-closed (remote)" once all preceding data has arrived.  The receiving
+endpoint MUST NOT consider the stream state to have changed until all data has
+arrived.
```
martinthomson: Interesting question: what does the sending peer assume?  I think that they have to assume that the state isn't closed until all data has been acknowledged (or their RST_STREAM is acknowledged).  (Oh, I see that you caught that, awesome).

draft-ietf-quic-transport.md
```diff
@@ -2187,6 +2197,12 @@ count toward the maximum number of streams that an endpoint is permitted to
 open.  Streams in any of these three states count toward the limit advertised in
 the concurrent stream limit.
 
+A recently closed stream MUST also be considered to count toward this limit
+until packets containing all frames required to close the stream have been
+acknowledged. For a stream which closed cleanly, this means all STREAM frames
+have been acknowledged; for a stream closed abruptly, this means the RST_STREAM
```
martinthomson: for a stream *that* closed abruptly

draft-ietf-quic-transport.md
MikeBishop: Yes -- the sending peer considers it to have changed, but it continues to count against the limit until the change is ACK'd.

martinthomson:
I just modified the OP to note that this also closes #200.

PR #382, opened 2017-03-10
Remove QUIC_OVERLAPPING_STREAM_DATA

### LINKED ISSUE #146: STREAM frame boundaries
We had a [discussion](https://groups.google.com/a/chromium.org/forum/#!topic/proto-quic/j747iwug-Hg) about this earlier on the proto-quic mailing list, but as far as I remember, we never reached a conclusion or discussed this on the IETF mailing list.

There are two options:

1. STREAM frame boundaries MUST be preserved when retransmitting them, i.e. a STREAM frame that is being retransmitted MAY be split into further pieces, but MUST NOT be combined with other STREAM frames.
2. STREAM frames MAY be reassembled before retransmitting them, and STREAM frame boundaries need not be preserved.

Option 1. is what Chrome and quic-go are currently implementing. 2. is what the draft is specifying in the current version ([here](https://github.com/quicwg/base-drafts/blob/master/draft-ietf-quic-transport.md#sending-and-receiving-data)).

If we decide for 1., this needs to be reflected in the draft. If we decide for 2., we need to remove the QUIC_OVERLAPPING_STREAM_DATA (0x8057) error.

martinthomson:
Closes #146

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2361,9 +2361,6 @@ QUIC_MISSING_PAYLOAD (0x80000030):
 QUIC_INVALID_STREAM_DATA (0x8000002E):
 : STREAM frame data is malformed.
 
-QUIC_OVERLAPPING_STREAM_DATA (0x80000057):
-: STREAM frame data overlaps with buffered data.
-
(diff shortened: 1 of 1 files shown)


martinthomson:
I'm running a very short timer on this one.  I'll merge this very soon.

PR #384, opened 2017-03-10
Stream1 is not that special

### LINKED ISSUE #248: Exemption from congestion control
The current transport doc exempts stream 1 from *congestion control*.  That seems to be unnecessary, potentially hard to implement, and maybe even dangerous.

martinthomson:
This stream simply isn't that important.  We should definitely be able to function without the congestion control exemption.  I also removed the flow control exemption in line with other decisions we have made not to exempt stream 1 (it counts toward the concurrent stream limit, for example).  That was harder, but all it takes is realizing that we have default values for flow control that can be used prior to handshake completion.  Those defaults are so abundantly generous that only the most perverse certificate (or PQ key exchange) will get close to overflowing.

Closes #248.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1014,7 +1014,7 @@ truncate_connection_id (0x0004):
   connection ID being present in every packet.
 
 
-### Values of Transport Parameters for 0-RTT
+### Values of Transport Parameters for 0-RTT {#zerortt-parameters}
 
 Transport parameters from the server SHOULD be remembered by the client for use
 with 0-RTT data.  A client that doesn't remember values from a previous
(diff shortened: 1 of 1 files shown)


janaiyengar:
I think this is fine, since neither of these limits should really cause grevious harm to the connection or the handshake. But it should be discussed.

martinthomson:
Ahh, yes, I had forgotten that little tidbit.  I'll revise this to allow the exception for flow control.

That said, the server has the ability to send an authenticated WINDOW_UPDATE if it accepts 0-RTT (the only case where this matters, because otherwise 0-RTT data doesn't count*****).

** new issue inbound.

martinthomson:
We're not going to do this I think.

PR #397, opened 2017-03-13
Adds an overview section for loss detection and restructures a bit

janaiyengar:


draft-ietf-quic-recovery.md
```diff
@@ -166,12 +157,50 @@ measure and report the delay from when a packet was received by the OS kernel,
 which is useful in receivers which may incur delays such as context-switch
 latency before a userspace QUIC receiver processes a received packet.
 
+
 # Loss Detection
 
-We now describe QUIC's loss detection as functions that should be called on
-packet transmission, when a packet is acked, and timer expiration events.
+## Overview {#overview}
```
martinthomson: You don't need a specific overview section here.  The overview can go in the top-level section.  My technical writing teacher say that it is bad form to include empty sections, even if they have child sections, so you can avoid two problems.  Also, I assume that you need an overview for congestion control as well, so you avoid creating two sections with the label "overview".

draft-ietf-quic-recovery.md
```diff
@@ -166,12 +157,50 @@ measure and report the delay from when a packet was received by the OS kernel,
...
+    lost when a packet larger than it is acknowledged and a threshold amount of
+    time has passed since the packet was sent.
+
+  * Handshake packets are special in a number of ways, and a separate alarm
+    period is used for them.
+
+
+## Algorithm Details
+
+### Constants of interest
```
martinthomson: Try to maintain a consistent use of Title Case or Sentence case for headings.  Ideally we would be consistent with the [style guide](https://tools.ietf.org/html/rfc7322#section-3.4), which recommends the former.

draft-ietf-quic-recovery.md
```diff
@@ -166,12 +157,50 @@ measure and report the delay from when a packet was received by the OS kernel,
...
+
+  * If a packet is near the tail, where fewer than kReorderingThreshold packets
+    are sent after it, the sender cannot expect to detect loss based on the
+    previous mechanism. In this case, a sender uses both ack information and an
+    alarm to detect loss. Specifically, when the last sent packet is
+    acknowledged, the sender waits a short period of time to allow for
+    reordering and then marks any unacknowledged packets as lost. This mechanism
+    is based on the Linux implementation of TCP Early Retransmit.
+
+  * If a packet is sent at the tail, there are no packets sent after it, and the
```
ianswett: We don't mark any packets lost in this case until an ack is received.

draft-ietf-quic-recovery.md
```diff
@@ -166,12 +157,50 @@ measure and report the delay from when a packet was received by the OS kernel,
...
+    sender cannot use ack information to detect its loss. The sender therefore
+    relies on an alarm to detect such tail losses. This mechanism is based on
+    TCP's Tail Loss Probe.
 
-## Constants of interest
+  * If all else fails, a Retransmission Timeout (RTO) alarm is always set when
```
ianswett: That's not really true.  We only arm the RTO if we've exceeded max TLP and early retransmit timer and time loss detection don't apply.

draft-ietf-quic-recovery.md
```diff
@@ -166,12 +157,50 @@ measure and report the delay from when a packet was received by the OS kernel,
...
 # Loss Detection
 
-We now describe QUIC's loss detection as functions that should be called on
-packet transmission, when a packet is acked, and timer expiration events.
+## Overview {#overview}
+
+QUIC uses a combination of ack information and alarms to detect lost packets.
+An unacknowledged QUIC packet is marked as lost in one of the following ways:
+
+  * A packet is marked as lost if at least one packet that was sent a threshold
```
ianswett: I find this sentence a bit hard to read.  Is it possible to change these bullets to be structured in parallel and all begin with When?

ie: "When a packet sent more than a threshold number of packets(kReorderingThreshold) after an unackowledged packet is acknowledged."

PR #403, opened 2017-03-14
type the artwork

reschke:


MikeBishop:
Seems to make no difference in the HTML and add indents to the code blocks in the txt version.  Not opposed, but not seeing any particular benefit, either.  Does kramdown do something special with the types?

MikeBishop:
Note that prefixing with 4+ spaces makes it a code block in kramdown, but putting the ~~~/~~~ around it makes it take the spaces literally, which is why the indent shows up.

reschke:
The types propagate to type attributes on artwork, which in turn are used by rfc2629.xslt to use different CSS styles.

I did not intend to change anything in the TXT output.

Note that adding "type='abnf'" will trigger a correct warning in rfc2629.xslt, as the draft currently does not cite RFC 5234.

MikeBishop:
Ah, I see.  Okay, that's fine.  I'll probably remove the spaces, then -- they're not needed if we have the tildes.

PR #409, opened 2017-03-18
Incorporate F-RTO

ianswett:


draft-ietf-quic-recovery.md
```diff
@@ -248,6 +248,10 @@ tlp_count:
 rto_count:
 : The number of times an rto has been sent without receiving an ack.
 
+largest_sent_before_rto:
+: The last packet number sent prior to the first transmission due to
```
janaiyengar: nit: change phrase to "prior to the first retransmission timeout"

draft-ietf-quic-recovery.md
```diff
@@ -372,6 +377,12 @@ Pseudocode for OnPacketAcked follows:
 
 ~~~
    OnPacketAcked(acked_packet_number):
+     // If a packet sent prior to RTO was acked, then the RTO
+     // was spurious.  Otherwise, inform congestion control.
+     // Similar to the goal of F-RTO {{?RFC5682}}
```
janaiyengar: I would remove this last line from here and add it (now or later) to the text description.

draft-ietf-quic-recovery.md
```diff
@@ -634,32 +650,48 @@ are detected lost.
 ~~~
    OnPacketsLost(lost_packets):
      largest_lost_packet = lost_packets.last()
+     // Start a new recovery epoch if the lost packet is larger
+     // than the end of the previous recovery epoch.
      if (end_of_recovery < largest_lost_packet.packet_number):
        end_of_recovery = largest_sent_packet
        congestion_window *= kLossReductionFactor
+       congestion_window = max(congestion_window, kMinimumWindow)
```
janaiyengar: nit: combine this line with the previous one:
congestion_window = 
    max(congestion_window * kLossReductionFactor, kMinimumWindow)

draft-ietf-quic-recovery.md
```diff
@@ -634,32 +650,48 @@ are detected lost.
...
+sent immediately, and a time in the future if sending is pacing
+limited.
 
 ~~~
    TimeToSend(packet_size):
      if (bytes_in_flight + packet_size > congestion_window)
        return infinite
+     pacing_coefficient = 1.25
+     if (congestion_window < ssthresh)
+       pacing_coefficient = 2
```
janaiyengar: I don't think we need a pacing coeff of 2 here. I never understood why FQ did this, but I don't think we need to have this in the doc. In general I'm less certain that we want to describe the pacing impl as part of the congestion controller... I'm comfortable with definiing the TimeToSend interface without implementing it here. For now, I would at least remove the use of pacing_coeff here.

draft-ietf-quic-recovery.md
ianswett: I tried, but it ends up taking up 3 lines, so I just kept it as is.

PR #415, opened 2017-03-27
Add skipped packets detection to the OnAckReceived pseudocode

lucas-clemente:


draft-ietf-quic-recovery.md
```diff
@@ -332,6 +332,9 @@ Pseudocode for OnPacketSent follows:
 
 When an ack is received, it may acknowledge 0 or more packets.
 
+The sender MUST abort the connection if it receives an ACK for a packet it
+never sent, see {{QUIC-TRANSPORT}}.
```
martinthomson: This doesn't say what error to use, if any.

martinthomson:
Hmm, also, this is the wrong draft to put that text in.

Issue #142, opened 2017-01-13
get_delay() undefined in loss detection pseudocode


martinduke:
In draft-ietf-quic-recovery.md

`      if (rto_count = 0):
        alarm_duration = max(200ms, smoothed_rtt + 4 x rttvar);
      else:
        alarm_duration = loss_detection_alarm.get_delay() << 1;
      rto_count++;

The return value of get_delay() is not defined anywhere in the document.

martinthomson:
@ianswett, correct this if it isn't purely editorial.

ianswett:
It's editorial, as well as not defined.  I'll fix it.

PR #417, opened 2017-03-28
Simplify and clarify RTO alarm calculation

ianswett:


ianswett:
Intended to fix Issue #142

PR #418, opened 2017-03-29
Troubleshooting flags in public header

martinduke:
To move along #279, this PR provides a straw man proposal for both a Loss Detection Flag and a Blocked flag. I emphasize that these changes are separable, if one is palatable and the other is not.


martinthomson:
This is so badly bitrotten, that I think we need to start over if we get back to this.

PR #421, opened 2017-03-29
Add LIMIT_UPDATE and replaces concurrent stream limit

### LINKED ISSUE #419: Get rid of the concurrent stream limit by advertising a maximum stream ID
We've already had a lot of discussions about the concurrent streams limit, which all brought us back to the problem that at every given moment both peers have to agree about the number of open streams, which seems to require a non-trivial amount of bookkeeping (of ACKs received for all STREAM frames sent on a given stream).

I'd like to propose a different solution, which solves this problem, and has a couple of nice additional properties as well:

A receiver advertises the maximum Stream ID it is willing to accept. The sender is allowed to open all streams that have IDs not larger than this number. The receiver can advertise a higher maximum Stream ID whenever it likes to, by sending a STREAM_LIMIT_UPDATE frame. STREAM_LIMIT_UPDATE is a retransmittable frame that contains a single value, the maximum Stream ID.
Note that conceptually, this is similar to advertising a flow control offset (and the rules for STREAM_LIMIT_UPDATEs would be similar to those for WINDOW_UPDATEs).

An implementation is free to choose any algorithm to determine which maximum Stream ID it advertises. The number of open streams may be used, but I can think of a couple of other ways to limit the state a peer can create. It would even be possible to advertise different limits depending on the server load.

MikeBishop:
Fixes #419.  I largely copied the WINDOW_UPDATE language.  It might actually be cleaner to recast this as a third level of flow control:  Bytes on streams, bytes on connection, streams on connection.  That would let us consolidate some of the duplicated language.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -788,6 +788,7 @@ explained in more detail as they are referenced later in the document.
 | 0x04             |  WINDOW_UPDATE     | {{frame-window-update}}    |
 | 0x05             |  BLOCKED           | {{frame-blocked}}          |
 | 0x07             |  PING              | {{frame-ping}}             |
+| 0x08             |  LIMIT_UPDATE      | {{frame-limit-update}}     |
 | 0x40 - 0x7f      |  ACK               | {{frame-ack}}              |
 | 0x80 - 0xff      |  STREAM            | {{frame-stream}}           |
 {: #frame-types title="Frame Types"}
@@ -935,7 +936,7 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
    enum {
       stream_fc_offset(0),
       connection_fc_offset(1),
-      concurrent_streams(2),
+      initial_stream_limit(2),
       idle_timeout(3),
       truncate_connection_id(4),
       (65535)
@@ -997,10 +998,11 @@ connection_fc_offset (0x0001):
   sending a WINDOW_UPDATE ({{frame-window-update}}) for the connection
   immediately after completing the handshake.
 
-concurrent_streams (0x0002):
+initial_stream_limit (0x0002):
 
-: The maximum number of concurrent streams parameter is encoded as an unsigned
-  32-bit integer.
+: The initial stream number limit parameter contains the initial maximum stream
+  number the peer may initiate.  This is equivalent to sending a LIMIT_UPDATE
+  ({{frame-limit-update}}) immediately after completing the handshake.
 
 idle_timeout (0x0003):
 
@@ -1023,7 +1025,7 @@ truncate_connection_id (0x0004):
 Transport parameters from the server SHOULD be remembered by the client for use
 with 0-RTT data.  A client that doesn't remember values from a previous
 connection can instead assume the following values: stream_fc_offset (65535),
-connection_fc_offset (65535), concurrent_streams (10), idle_timeout (600),
+connection_fc_offset (65535), initial_stream_limit (20), idle_timeout (600),
 truncate_connection_id (absent).
 
 If assumed values change as a result of completing the handshake, the client is
@@ -1036,7 +1038,7 @@ particularly with respect to transport parameters that establish limits:
   increase to the affected flow control offsets is received, the client can
   recommence sending.
 
-* Similarly, a client might exceed the concurrent stream limit declared by the
+* Similarly, a client might exceed the initial stream limit declared by the
   server.  A client MUST reset any streams that exceed this limit.  A server
   SHOULD reset any streams it cannot handle with a code that allows the client
   to retry any application action bound to those streams.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2405,6 +2433,20 @@ A receiver MAY use an autotuning mechanism to tune the size of the offset
 increment to advertise based on a roundtrip time estimate and the rate at which
 the receiving application consumes data, similar to common TCP implementations.
 
+### Stream Limit Increment
+
+As with flow control, this document leaves when and how many streams to make
+available to a peer via LIMIT_UPDATE to the implementation, but offers a few
+considerations. LIMIT_UPDATE frames constitute overhead, and therefore, sending
+a LIMIT_UPDATE with small offset increments is undesirable.  At the same time,
```
marten-seemann: The only case where this is a noticeable overhead is when each stream is only used to transfer a small number of bytes, e.g. in the often cited case of video streaming where every frame consumes a stream. For HTTP (and probably all other use cases) sending the 5 bytes consumed by the LIMIT_UPDATE frame will be negligible.

draft-ietf-quic-transport.md
```diff
@@ -2405,6 +2433,20 @@ A receiver MAY use an autotuning mechanism to tune the size of the offset
...
+withholding LIMIT_UPDATES prevents the peer from fully utilizing the transport.
+Implementations must find the correct tradeoff between these sides to determine
+how large an offset increment to send in a WINDOW_UPDATE.
+
+A receiver MAY adjust the number of permitted streams not in the "closed" state
+based on current activity, system conditions, and other environmental factors.
```
marten-seemann: Better: the maximum StreamID.

MikeBishop:
Well, I'd typically say you add a flag, but QUIC embeds flags into the frame type, so a frame with one flag is equivalent to two frames.  We could pick another sentinel value for Stream ID (0xFFFFFFFF?), but that wastes quite a few bytes (since the window is 64 bits and you'll never need more than 32 to state a stream limit.

We could state the stream limit in each WINDOW_UPDATE frame, perhaps.  I don't have data on how quickly one changes versus the other.  Given that we have two levels of flow control and stream flow control seems likely to update more quickly than open streams, that seems wasteful as well.

What we _could_ perhaps do is remove the Stream 0 sentinel value from WINDOW_UPDATE and have a separate frame for connection flow control that gave you both a new max byte count and max stream ID in one shot.  That's a larger change, but if it's appealing to folks...?

What gets hairy there is that we say to ignore a WINDOW/LIMIT_UPDATE that doesn't increase what you're allowed; by having two values in one frame, you make it possible to send invalid combinations.

draft-ietf-quic-transport.md
```diff
@@ -788,6 +788,7 @@ explained in more detail as they are referenced later in the document.
 | 0x04             |  WINDOW_UPDATE     | {{frame-window-update}}    |
 | 0x05             |  BLOCKED           | {{frame-blocked}}          |
 | 0x07             |  PING              | {{frame-ping}}             |
+| 0x08             |  LIMIT_UPDATE      | {{frame-limit-update}}     |
```
martinthomson: I really don't like this name.  (I don't like WINDOW_UPDATE either).  How about `MAX_STREAM_ID`.

draft-ietf-quic-transport.md
```diff
@@ -2196,46 +2230,40 @@ new stream.
...
-receives a STREAM frame that causes its advertised concurrent stream limit to be
-exceeded MUST treat this as a stream error of type QUIC_TOO_MANY_OPEN_STREAMS
-({{error-handling}}).
-
+receives a STREAM frame with an ID greater than the limit it has sent MUST treat
+this as a stream error of type QUIC_TOO_MANY_OPEN_STREAMS ({{error-handling}}),
+unless this is a result of a change in the initial offsets (see
+{{zerortt-parameters}}).
+
+A Maximum Stream ID of 0 will cause a connection to be unusable.  Application
```
martinthomson: 1 also makes the connection unusable because stream 1 counts.

marten-seemann: 1 is a valid value. It allows the handshake to complete. The peers can always advertise a higher value later.

draft-ietf-quic-transport.md
```diff
@@ -2196,46 +2230,40 @@ new stream.
...
+protocols that use QUIC might require a certain minimum number of initial
+streams to function correctly.  If a peer advertises an maximum stream ID
+(initial_stream_limit) that is too small for the selected application protocol
```
marten-seemann: I don't it's necessary to define an error condition here.
If the server sets a too small initial_stream_limit, this just means that the client can't immediately open the streams requested by the application, but it has to wait until the server sends a LIMIT_UPDATE.

draft-ietf-quic-transport.md
martinthomson: Of course, that is the advantage of this design.

MikeBishop: 1 will likely make any application protocol unusable, either because it requires a control stream to open immediately (HTTP) or because the connection will be unable to serve any requests.  However, for something like the "connect and PING" test you proposed for first-interop (maybe ALPN token of "none"?), "1" would be an appropriate value to advertise.

draft-ietf-quic-transport.md
MikeBishop: If we have a BLOCKED analogue, that seems reasonable.  It effectively becomes a warning rather than a fatal error.

martinthomson:
Approving this; I have a plan to rename the frame type now and the other comments were addressed already.

MikeBishop:
Maybe MAX_STREAM_ID and MAX_OFFSET?  The names aren't really folded into the wire, so they're technically editorial.

martinthomson:
Pretty close to what I had in mind.  With #443 I think we can have: STREAM_LIMIT, DATA_LIMIT and STREAM_DATA_LIMIT.  Or as you suggest MAX_STREAM_ID, MAX_STREAM_DATA_OFFSET, and MAX_DATA_OFFSET.

martinthomson:
Merged manually.

draft-ietf-quic-transport.md
janaiyengar: Agreed that this surely doesn't seem worthy of a MUST terminate. I also don't think TOO_MANY_OPEN_STREAMS is the right error to use here.

mikkelfj: If the connection is not closed no a MAX_STREAM_ID violation, both sides need to update state to treat the violating stream as closed/reset and not to be used again, or at least some rules should clarify what should happen. Also, if such a violation happens, there would likely come many more because the sender is not keeping track. Alternatively, the sender could use a RST with error as control signal to back off, but that is just more complex state handling. So the simplest solution seems to be to close the entire connection and simply require sender to respect the limit.

martinthomson: We require that exceeding the limit is punished with an error closure.  But even if we didn't (see the 0-RTT exception), the connection still works when there are too many streams.  The specific error code that Marten is talking about here is the case where you don't have enough streams to operate: a condition that is easily rectified by increasing the maximum stream ID.

mikkelfj: My comment was aimed at @janaiyengar 's comment. I think I agree with Marten - I have not read up on 0-RTT.

PR #424, opened 2017-03-30
GREASE the server port number

bemasc:


draft-ietf-quic-http.md
```diff
@@ -105,6 +105,11 @@ Connectivity problems (e.g. firewall blocking UDP) can result in QUIC connection
 establishment failure, in which case the client SHOULD continue using the
 existing connection or try another alternative endpoint offered by the origin.
 
+Servers MAY serve HTTP/QUIC on any UDP port.  Servers MUST use the same port
+across all IP addresses that serve a single domain, and SHOULD NOT change this
+port. Servers SHOULD choose this port at random when first enabling HTTP/QUIC
```
MikeBishop: Discussion from the editors:  It's perfectly reasonable to point out that hq *can* be on any port and that's supported.  The "SHOULD randomize" gets into the middlebox debate and really should become a separate issue if we're going to recommend it.  (And it could legitimately become a best practice whether the spec recommends it or not.)

If you back this piece out (or I can do it), I think the rest is fine to merge.

bemasc: Should I just remove it, or switch to MAY?

MikeBishop: I'd just remove it -- if it's just a MAY, then it's duplicative of "MAY serve ... on any UDP port."

bemasc: OK, done!

PR #430, opened 2017-04-02
Reduce the number of offset lengths in a stream frame

### LINKED ISSUE #414: Stream frame can have an offset length equal 0
There should be a note saying what is the value of an offset if the offset length is 0. The implementation must assume that  the offset is 0.

ianswett:
Also fixes #414

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -788,8 +788,8 @@ explained in more detail as they are referenced later in the document.
 | 0x04             |  WINDOW_UPDATE     | {{frame-window-update}}    |
 | 0x05             |  BLOCKED           | {{frame-blocked}}          |
 | 0x07             |  PING              | {{frame-ping}}             |
-| 0x40 - 0x7f      |  ACK               | {{frame-ack}}              |
-| 0x80 - 0xff      |  STREAM            | {{frame-stream}}           |
+| 0xa0 - 0x7f      |  ACK               | {{frame-ack}}              |
+| 0xc0 - 0xff      |  STREAM            | {{frame-stream}}           |
 {: #frame-types title="Frame Types"}
 
 # Life of a Connection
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1292,10 +1292,10 @@ subsequent sections.
 ## STREAM Frame {#frame-stream}
 
 STREAM frames implicitly create a stream and carry stream data. The type byte
-for a STREAM frame contains embedded flags, and is formatted as `1FDOOOSS`.
+for a STREAM frame contains embedded flags, and is formatted as `11FDOOSS`.
 These bits are parsed as follows:
 
-* The leftmost bit must be set to 1, indicating that this is a STREAM frame.
+* The first two bit must be set to 11, indicating that this is a STREAM frame.
```
MikeBishop: bits, not bit

draft-ietf-quic-transport.md
```diff
@@ -1407,15 +1408,13 @@ entropy on demand, which should be adequate protection against most
 opportunistic acknowledgement attacks.
 
 The type byte for a ACK frame contains embedded flags, and is formatted as
-`01NULLMM`.  These bits are parsed as follows:
+`101NLLMM`.  These bits are parsed as follows:
 
-* The first two bits must be set to 01 indicating that this is an ACK frame.
+* The first two bits must be set to 101 indicating that this is an ACK frame.
```
MikeBishop: three bits

martinthomson:
I was a little surprised to see this change was merged.

@ianswett, this change to the transport doc wasn't reviewed by either @janaiyengar or myself; nor do I see any discussion on the list.  In this case, I think that it's a big improvement, so I've no objection, but I don't think we should be merging changes to documents without the approval of at least one editor.

(And sorry for not reviewing this, I don't know how I missed it.)

ianswett:
I remember Jana and I discussing it, but next time I'll make sure one of you approves it on Github.

martinthomson:
:) If Jana said it's OK, that's fine.  I also think that it's OK.

PR #437, opened 2017-04-13
Fix >80 character line so that Makefiles work

ekr:
This is breaking in the Makefile sanity checks.


MikeBishop:
Sorry, I fixed it directly before I saw you had this one.  Same change.

ekr:
No worries. It's important it be fixed, not that I bump up my contribution stats.

Issue #435, opened 2017-04-13
Stream ID sequential order may create head of queue blocking


huitema:
The transport draft says in Section 10.2 that "Streams MUST be created in sequential order." This is logical. Suppose now that an application sends the STREAM-DATA opening stream 5 in one packet number X, and then the opening of stream 7 in the next packet, number X+1. This is a fairly natural thing to do, for example, for QUIC over DNS, where each transaction opens its own stream. 

Suppose now that packet number X is lost. Packet X+1 will be processed. The receiver will see for the first time a reference to stream 7, without having seen stream 5. In a strict reading of the spec, since streams MUST be created in sequential order, the receiver should signal an error, since streams have been created out of order. But that's fiendish. Or maybe it should queue the stream 7 packet until stream 5 has been open, but that would cause head of queue blocking. Or maybe it should not care about the MUST, and just open stream 7.

Now, of course, if the receiver is willing to accept stream ID opened out of order, we can have different problems. How much out of order is too much? You may want to say "1", but consider that a client could easily pack a dozen DNS requests and as many stream opening in a single packet. If that packet is lost, there is a gap of a dozen stream ID.

Not sure what the right solution should be.

marten-seemann:
I don't see any problem here. Since streams are created sequentially, a receiver can open all streams with lower IDs when a such reordered packet arrives. By doing so, the concurrent steak limit is enforced implicitly.

This is how we implemented it in quic-go. Note that this also works out nicely if you expose an Accept() API for new streams.

huitema:
Marten, I agree that there is no real problem if the client does indeed create streams in order. But I am concerned with an adversarial client creating streams out of order. Suppose that the server, like you suggest, "open all streams with lower IDs". Suppose that the adversarial client opens stream 5, and then stream 1000005. Are you going to open 1 million streams? Will that not have a significant impact on server performance?

mikkelfj:
There are two solutions currently going as I understand. The current draft-01 with limit on concurrent active streams which the receiver announces to the sender. So the receiver will drop the stream or the connection if the limit is exceeded. However, there are race conditions. So the new proposal is for the receiver to annouce the maximum id accepted. It can increase this as older streams are being closed.

[...]

PR #445, opened 2017-04-19
Adding a description of stream commitment attacks in security section

huitema:
See discussion of issue #435: Stream ID sequential order may create head of queue blocking.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2706,13 +2706,13 @@ large number of connections, in a manner similar to
 SYN flooding attacks in TCP.
 
 Normally, clients will open streams sequentially,
-as explained in {{stream-identifiers}}. 
+as explained in {{stream-identifiers}}.
 However, when several streams are initiated at short intervals,
 transmission error may cause STREAM DATA frames opening streams to be
 received out of sequence. A receiver is obligated to open intervening
 streams if a higher-numbered stream ID is received. Thus, on a
 new connection, opening stream 2000001 opens 1 million streams,
-as required by the specification. 
+as required by the specification.
 
 The number of active streams is limited
 by the concurrent stream limit transport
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2697,6 +2697,45 @@ also be forward-secure encrypted.  Since the attacker will not have the forward
...
+However, when several streams are initiated at short intervals,
+transmission error may cause STREAM DATA frames opening streams to be
+received out of sequence. Some implementations will have provisions to
+open all intermediate streams in case of out-of-sequence arrival. For example,
+a server that receive STREAM DATA for streams 3 and then 9 may decide to
+open stream 5 and 7 immediately. An adversarial client
+would for example exploit that
+design by sending STREAM DATA for streams 3 and the 2,000,001, causing the
+server to open 1 million connections, and contributing to server resource
+exhaustion.
```
martinthomson: Actually, this is much easier: a receiver is obligated to open intervening streams if a higher-numbered stream ID is received.  Thus, on a new connection, opening stream 2000001 opens 1 million streams, *as required by the specification*.  The stream concurrency limit protects against this sort of attack.

draft-ietf-quic-transport.md
```diff
@@ -2697,6 +2697,45 @@ also be forward-secure encrypted.  Since the attacker will not have the forward
 secure key, the attacker will not be able to generate forward-secure encrypted
 packets with ACK frames.
 
+## Stream commitment attack
```
martinthomson: Title Case Headings are the Convention

marten-seemann:
The state allocated for one stream is pretty small: it's probably one map entry somewhere, and a couple of counter variables (for read and write position) in the stream. So, unless chosen absurdly large (O(10e5) or so), the memory commitment should be negligible.

huitema:
Following Martin's suggestion, I made the text somewhat shorter. I don't feel a compelling need to make it too short. I would like to make sure that future implementers understand the issue and do something about it.

huitema:
Martin, do the changes meet your requirements?

martinthomson:
Yes, this looks great.  I want to ensure that others get a chance to have a look before I merge this, but I will probably merge this tomorrow-ish.

PR #450, opened 2017-04-20
Split WINDOW_UPDATE into MAX_DATA and MAX_STREAM_DATA

### LINKED ISSUE #443: Split WINDOW_UPDATE
The current design has a special carve-out for stream 0, but special semantics for it.

If we split this into two different frame types, we avoid having special logic for the scaling of the connection-level window AND we can reclaim stream 0.

That leads to the possibility of using stream 0 for crypto, stream 1 for HTTP control.

martinthomson:
This also changes the name of the LIMIT_UPDATE frame to MAX_STREAM_ID to match.

A lot of the text talked about stream offsets and had flow control affect
maximum stream offsets.  This turned out to be unnecessarily obtuse.  The text
in this changeset simply says that there is a limit to the amount of data that
can be sent.  This turns out to be a lot of changes, but I think that it is
easier to understand as a result.

Closes #443.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2810,13 +2810,13 @@ merely aesthetically displeasing, or architecturally dubious).
 The initial contents of this registry are shown in
 {{iana-tp-table}}.
 
-| Value  | Parameter Name            | Specification                       |
-|:-------|:--------------------------|:------------------------------------|
-| 0x0000 | stream_fc_offset          | {{transport-parameter-definitions}} |
-| 0x0001 | connection_fc_offset      | {{transport-parameter-definitions}} |
-| 0x0002 | initial_stream_limit      | {{transport-parameter-definitions}} |
-| 0x0003 | idle_timeout              | {{transport-parameter-definitions}} |
-| 0x0004 | truncate_connection_id    | {{transport-parameter-definitions}} |
+| Value  | Parameter Name          | Specification                       |
+|:-------|:------------------------|:------------------------------------|
+| 0x0000 | initial_max_stream_data | {{transport-parameter-definitions}} |
+| 0x0001 | initial_max_data        | {{transport-parameter-definitions}} |
+| 0x0002 | initial_max_stream_id   | {{transport-parameter-definitions}} |
+| 0x0003 | idle_timeout            | {{transport-parameter-definitions}} |
+| 0x0004 | truncate_connection_id  | {{transport-parameter-definitions}} |
 {: #iana-tp-table title="Initial QUIC Transport Parameters Entries"}
 
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1685,7 +1685,7 @@ a result of a change in the initial limits (see {{zerortt-parameters}}).
 
 ## MAX_STREAM_DATA Frame {#frame-max-stream-data}
 
-The MAX_STREAM_DATA frame (type=0x04) is used in flow control to inform a peer
+The MAX_STREAM_DATA frame (type=0x05) is used in flow control to inform a peer
 of the maximum amount of data that can be sent on a stream.
 
 The frame is as follows:
@@ -1725,7 +1725,7 @@ this is a result of a change in the initial limits (see
 
 ## MAX_STREAM_ID Frame {#frame-max-stream-id}
 
-The MAX_STREAM_ID frame (type=0x08) informs the peer of the maximum stream ID
+The MAX_STREAM_ID frame (type=0x06) informs the peer of the maximum stream ID
 that they are permitted to open.
 
 The frame is as follows:
@@ -1757,7 +1757,7 @@ limits (see {{zerortt-parameters}}).
 
 ## BLOCKED Frame {#frame-blocked}
 
-A sender sends a BLOCKED frame (type=0x05) when it is ready to send data (and
+A sender sends a BLOCKED frame (type=0x08) when it is ready to send data (and
 has data to send), but is currently flow control blocked. BLOCKED frames are
 purely informational frames, but extremely useful for debugging purposes. A
 receiver of a BLOCKED frame should simply discard it (after possibly printing a
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1745,7 +1745,8 @@ Maximum Stream ID:
 
 Loss or reordering can mean that a MAX_STREAM_ID frame can be received which
 states a lower stream limit than the client has previously received.
-MAX_STREAM_ID frames which do not increase the maximum stream ID MUST be ignored.
+MAX_STREAM_ID frames which do not increase the maximum stream ID MUST be
+ignored.
 
 A peer MUST NOT initiate a stream with a higher stream ID than the greatest
 maximum stream ID it has received.  An endpoint MUST terminate a connection with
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1640,89 +1643,90 @@ by a client in protected packets, because it is certain that the server is able
...
+  connection-level data limit is determined by multiplying the encoded value by
+  1024.
 
-The connection-level flow control offset is expressed in units of 1024 octets
-(for a stream identifier of 0).  That is, the connection-level flow control
-offset is determined by multiplying the encoded value by 1024.
+When counting data toward this limit, an endpoint accounts for the maximum
+offset of data that is sent or received on every stream.  Loss or reordering can
+mean that the maximum offset is greater than the total size of data received on
+a stream.  Receiving STREAM frames might not increase the maximum offset on a
```
janaiyengar: "Loss or reordering can
 +mean that the maximum offset is greater than the total size of data received on
 +a stream." -- Not sure what this means.

janaiyengar: This sentence does not make sense: "Receiving STREAM frames might not increase the maximum offset on a stream if the maximum offset doesn't increase."

draft-ietf-quic-transport.md
```diff
@@ -1640,89 +1643,90 @@ by a client in protected packets, because it is certain that the server is able
...
+stream if the maximum offset doesn't increase.  A STREAM frame with a FIN bit
+set or RST_STREAM causes the final offset for a stream to be fixed.
```
janaiyengar: I don't think you need to describe how RST_STREAM or FIN affects STREAM offsets here. MAX_DATA is for Connection flow control, and this information is useful only for STREAM flow control.

draft-ietf-quic-transport.md
```diff
@@ -2344,54 +2376,50 @@ senders from exceeding a receiver's buffer capacity for the connection, and (ii)
...
+MAX_STREAM_DATA frames.  A receiver MUST NOT renege on an advertisement; that
+is, once a receiver advertises an offset, it MUST NOT subsequently advertise a
+smaller offset.  A sender could receive MAX_DATA or MAX_STREAM_DATA frames out
+of order; a sender MUST therefore ignore any flow control offset that does not
+move the window forward.
 
 A receiver MUST close the connection with a
 QUIC_FLOW_CONTROL_RECEIVED_TOO_MUCH_DATA error ({{error-handling}}) if the
-peer violates the advertised stream or connection flow control windows.
+peer violates the advertised limits on data on a connection or stream.
```
janaiyengar: suggestion: "violates the advertised connection or stream data limits."

draft-ietf-quic-transport.md
```diff
@@ -2419,44 +2447,53 @@ frame and has sent neither a FIN nor a RST_STREAM, it MUST send a RST_STREAM in
...
-considerations. LIMIT_UPDATE frames constitute minimal overhead, while
-withholding LIMIT_UPDATEs prevents the peer from fully utilizing the transport.
+available to a peer via MAX_STREAM_ID to implementations, but offers a few
+considerations.  MAX_STREAM_ID frames constitute minimal overhead, while
+withholding MAX_STREAM_ID frames can prevent the peer from using the available
+parallelism.
+
+Implementations will likely want to increase the maximum stream ID as
+peer-initiated streams close.  Sending a MAX_STREAM_ID frame along with ACK
+frames ensures a regular increase in the stream limit without generating excess
```
janaiyengar: Sending a MAX_STREAM_ID with each ACK makes no sense. Perhaps you meant to say: "... along with acknowledgements of frames that close a stream"?

draft-ietf-quic-transport.md
martinthomson: Hmm, that's old text, it says that for any given stream max_offset >= sum([x.len for x in all_stream_frames]) (as opposed to ==).

martinthomson: It's true, just a little repetitive.

draft-ietf-quic-transport.md
martinthomson: OK, dropping this (BTW, this is existing text).

draft-ietf-quic-transport.md
martinthomson: Not exactly.  Implies a particular strategy for increasing the maximum stream ID, which I want to avoid.  The point is that you might send any MAX_STREAM_ID frames you might want to send on the same schedule as ACK.  I'll try to improve it some.

martinthomson: I realize that the context is bad for that, so I'll try to fix that too.

draft-ietf-quic-transport.md
```diff
@@ -1671,10 +1671,10 @@ Maximum Data:
...
 When counting data toward this limit, an endpoint accounts for the maximum
 offset of data that is sent or received on every stream.  Loss or reordering can
-mean that the maximum offset is greater than the total size of data received on
-a stream.  Receiving STREAM frames might not increase the maximum offset on a
-stream if the maximum offset doesn't increase.  A STREAM frame with a FIN bit
-set or RST_STREAM causes the final offset for a stream to be fixed.
+mean that the maximum offset on a stream can be greater than the total size of
+data received on that stream.  Conversely, receiving STREAM frames might not
+increase the maximum offset on a stream if the maximum offset in the frame is
+lower than the maximum received for that stream.
```
janaiyengar: I understand now what this says, but it's confusing, since maximum offset and MAX_DATA both represent maxima. I would suggest using "largest received offset" instead of "maximum offset".

draft-ietf-quic-transport.md
```diff
@@ -1671,10 +1671,10 @@ Maximum Data:
 
 When counting data toward this limit, an endpoint accounts for the maximum
 offset of data that is sent or received on every stream.  Loss or reordering can
-mean that the maximum offset is greater than the total size of data received on
-a stream.  Receiving STREAM frames might not increase the maximum offset on a
-stream if the maximum offset doesn't increase.  A STREAM frame with a FIN bit
-set or RST_STREAM causes the final offset for a stream to be fixed.
+mean that the maximum offset on a stream can be greater than the total size of
+data received on that stream.  Conversely, receiving STREAM frames might not
+increase the maximum offset on a stream if the maximum offset in the frame is
```
janaiyengar: "might not" --> "will not". If the STREAM frame's largest offset is lower than the largest received so far, the largest received offset will not increase.

draft-ietf-quic-transport.md
```diff
@@ -2471,13 +2471,13 @@ considerations.  MAX_STREAM_ID frames constitute minimal overhead, while
 withholding MAX_STREAM_ID frames can prevent the peer from using the available
 parallelism.
 
-Implementations will likely want to increase the maximum stream ID as
-peer-initiated streams close.  Sending a MAX_STREAM_ID frame along with ACK
-frames ensures a regular increase in the stream limit without generating excess
+Sending any increase to the maximum stream ID in the same packet as an ACK frame
+ensures a regular increase in the stream limit without generating excess
```
janaiyengar: This advice is true for all frames, including MAX_DATA and MAX_STREAM_DATA frames. I'd remove this sentence or say something about *when* to send the MAX_STREAM_ID frame rather than which frame to bundle it with. It might make sense to send it immediately after/along with a STREAM frame that carries a FIN or a RST_STREAM frame on a peer-initiated stream, for instance.

draft-ietf-quic-transport.md
janaiyengar: Can you move this text to the MAX_STREAM_DATA section instead? It was written for WINDOW_UPDATE, which covered both connection and stream flow control.

draft-ietf-quic-transport.md
janaiyengar: An ACK is sent about every two packets received, which is not the schedule that a MAX_STREAM_ID frame would be sent. I've left a comment in my review of the new text.

draft-ietf-quic-transport.md
martinthomson: Yep, I think that I will just remove the paragraph.  The advice on when to send is still present in the next paragraph, which is what you are looking for.

PR #453, opened 2017-04-20
Update references with nicer labels

martinthomson:
The latest version of kramdown-rfc2629 has a few new features that make this much easier to manage.


martinthomson:
Anyone running tools from the command line might need to run this:

```sh
[sudo] gem install --update kramdown-rfc2629
make clean
```

PR #454, opened 2017-04-20
Split BLOCKED

martinthomson:
This builds on #450 and splits BLOCKED in the same way.

(For reviewers, #450 lands, the changes currently fit in 8d349ad966d9c4ab514b3ebd1e08a986b5da02b0)

PR #455, opened 2017-04-20
STREAM_ID_NEEDED frame

martinthomson:
Analogous to BLOCKED, this frame allows an endpoint to protest at not being able to open new streams.

I'm not convinced we need this (I'm also not especially enthusiastic about BLOCKED), but it was easy to write up.

Builds on #454, look at 3629948c1c575d12e0cbb974e666a56158cac6b3 if you want to see just these changes.


MikeBishop:
Maybe.  Alternatively, you could model beyond-the-max-ID streams as having a max offset of zero -- just send a STREAM_BLOCKED frame referencing the stream you'd like to open.  Has the advantage of explicitly telling the sender how many streams are throttled behind their max, rather than just that there is one.

martinthomson:
I think that I like that better.

PR #456, opened 2017-04-20
Move crypto to stream 0, HTTP control to stream 1

martinthomson:
The nasty overloading in WINDOW_UPDATE was the only reason that we couldn't use stream 0 for identifying streams.  That's gone now.

A nice benefit is that this also makes the special exception for `initial_max_stream_id` go away.  0 is now a completely valid initial maximum stream ID.  In some protocols, that might even make sense.

Builds on #454, look at 16dee05d236ecf27c2d069b94d142a70258ef57c for just the relevant changes here.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1670,9 +1670,9 @@ Maximum Data:
   1024.
 
 All data sent in STREAM frames counts toward this limit, with the exception of
-STREAM frames on stream 0.  The sum of the largest received offsets on all
-streams - including closed streams, but excluding stream 0 - MUST NOT exceed the
-value advertised by a receiver.  An endpoint MUST terminate a connection with a
+data on stream 0.  The sum of the largest received offsets on all streams -
+including closed streams, but excluding stream 0 - MUST NOT exceed the value
+advertised by a receiver.  An endpoint MUST terminate a connection with a
 QUIC_FLOW_CONTROL_RECEIVED_TOO_MUCH_DATA error if it receives more data than the
 maximum data value that it has sent, unless this is a result of a change in the
 initial limits (see {{zerortt-parameters}}).
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -207,11 +207,11 @@ setting up a secure transport connection.  QUIC connections are
 expected to commonly use 0-RTT handshakes, meaning that for most QUIC
 connections, data can be sent immediately following the client
 handshake packet, without waiting for a reply from the server.  QUIC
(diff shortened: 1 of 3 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2344,54 +2382,50 @@ senders from exceeding a receiver's buffer capacity for the connection, and (ii)
...
 Connection flow control is a limit to the total bytes of stream data sent in
-STREAM frames on all streams contributing to connection flow control.  A
-receiver advertises credit for a connection by sending a WINDOW_UPDATE frame
-with the Stream ID set to zero (0x00).  A receiver maintains a cumulative sum of
-bytes received on all streams contributing to connection-level flow control, to
-check for flow control violations. A receiver may maintain a cumulative sum of
-bytes consumed on all contributing streams to determine the connection-level
-flow control offset to be advertised.
+STREAM frames on all streams.  A receiver advertises credit for a connection by
+sending a MAX_DATA frame.  A receiver maintains a cumulative sum of bytes
```
mikkelfj: It is not clear if stream id 0 counts against this, even if it is not punished for violating the connection max_data limit.

martinthomson: The limiting mechanism is now a maximum number.  The smallest maximum that can be expressed is 0, so stream 0 is always OK.

martinthomson: My bad, the TLS spec says that stream 0 doesn't count and this obviously removes that exception.  That's a problem, I agree.  Not sure what the right answer is here.  We previously agreed that stream 0 was exempt from connection flow control, mainly due to problems with completing the handshake with 0-RTT, but with Victor's proposed changes, maybe we can avoid that.  I'll restore the exemption for stream 0 in the meantime.

martinthomson:
@janaiyengar, I'd like to merge this.  Please review; it's going to rot very quickly.

draft-ietf-quic-transport.md
```diff
@@ -1671,11 +1669,13 @@ Maximum Data:
   connection-level data limit is determined by multiplying the encoded value by
   1024.
 
-The sum of the largest received offsets on all streams - including closed
-streams - MUST NOT exceed the value advertised by a receiver.  An endpoint MUST
-terminate a connection with a QUIC_FLOW_CONTROL_RECEIVED_TOO_MUCH_DATA error if
-it receives more data than the maximum data value that it has sent, unless this
-is a result of a change in the initial limits (see {{zerortt-parameters}}).
+All data sent in STREAM frames counts toward this limit, with the exception of
+STREAM frames on stream 0.  The sum of the largest received offsets on all
```
janaiyengar: nit: data on stream 0

PR #457, opened 2017-04-20
Stream blocked future

martinthomson:
@MikeBishop suggested this in #455.  Unfortunately, it uncovered an issue with RST_STREAM in the "idle" state.


martinthomson:
I would like someone to pick between this, #455 and nothing pretty soon or this will all rot (as it already has).  @janaiyengar, that can be you if you can form an opinion.  If I don't get any input in the next week, I'm going to make a decision.

MikeBishop:
Obsoleted by #455.

Issue #179, opened 2017-01-19
Text for "half-closed (remote)"


ddragana:
> A stream that is "half-closed (remote)" is no longer being used by
>    the peer to send any data.  In this state, a sender is no longer
>    obligated to maintain a receiver stream-level flow-control window.
> 
>    If an endpoint receives any STREAM frames for a stream that is in
>    this state, it MUST close the connection with a
>    QUIC_STREAM_DATA_AFTER_TERMINATION error (Section 11).

If a packet sent before the packet with FIN bit set gets lost and is retransmitted is going to cause close of the connection.
The text needs to change to say if a frame with offset  larger than the offset + size of the frame with FIN bit set is received. (or a better wording)

(An editorial remark: In the "Lifecycle of a stream" diagram going from "half-closed(remote)" to "closed" should be cause by "send FIN" not "recv FIN" and from "half-closed(local)" to "closed" should be cause by "recv FIN")

martinthomson:
Yes, the correct formulation here is: if a STREAM frame *containing octets that are greater in offset than the last octet in a closed stream*, then the receiver MUST terminate ...  A receiver SHOULD retain a last octet value for closed streams for a short period of time after they close.  This value is the largest received octet offset if the stream is terminated by a STREAM frame containing a FIN bit; a peer will include this value in a RST_STREAM frame.

martinthomson:
I think that we can keep this editorial.

martinthomson:
We've a lot of [new text](https://quicwg.github.io/base-drafts/draft-ietf-quic-transport.html#final-offset) here.  If there are any remaining errors, please open a new issue.

PR #459, opened 2017-04-21
Change RST_STREAM impact on state

### LINKED ISSUE #458: RST_STREAM has no semantics in idle state
This appears to be a bug with the state machine.  An endpoint is allowed to send RST_STREAM, but the draft doesn't actually define what happens.  There is no text and no transitions in the picture.

martinthomson:
The behaviour of RST_STREAM was fine, but it's description created several problems.  Most of these problems manifested in a convoluted description of the closed state.  The same problem existed in HTTP/2, but it was less of a problem there because of the strict ordering provided by TCP.

This refactors the state machine to recognize that RST_STREAM is really unidirectional.  Both sides need to send it or flow control gets screwy.

So people can understand, here's the old state machine picture:
```
                            +--------+
                            |        |
                            |  idle  |
                            |        |
                            +--------+
                                 |
                                 | send data/
                                 | recv data/
                                 | recv higher stream
                                 |
                                 v
                            +--------+
                recv FIN    |        |    send FIN
                  ,---------|  open  |-----------.
                 /          |        |            \
                v           +--------+             v
         +----------+            |             +----------+
         |   half   |            |             |   half   |
         |  closed  |            | send RST/   |  closed  |
         | (remote) |            | recv RST    |  (local) |
         +----------+            |             +----------+
             |                   |                    |
             | send FIN/         |          recv FIN/ |
             | send RST/         v          send RST/ |
             | recv RST     +--------+      recv RST  |
             `------------->|        |<---------------'
                            | closed |
                            |        |
                            +--------+
```

The new one is a little simpler:
```
                            +--------+
                 recv RST   |        |    send RST
              ,-------------|  idle  |---------------.
             /              |        |                \
            |               +--------+                 |
            |                    |                     |
            |        send STREAM / recv STREAM         |
            |                    |                     |
            |                    v                     |
            |    recv FIN/  +--------+    send FIN/    |
            |    recv RST   |        |    send RST     |
            |     ,---------|  open  |-----------.     |
            |    /          |        |            \    |
            v   v           +--------+             v   v
         +----------+                          +----------+
         |   half   |                          |   half   |
         |  closed  |                          |  closed  |
         | (remote) |                          |  (local) |
         +----------+                          +----------+
             |                                        |
             |   send FIN/  +--------+    recv FIN/   |
              \  send RST   |        |    recv RST   /
               `----------->| closed |<-------------'
                            |        |
                            +--------+
```

I've built this on top of #450 to make my life easier in the case that #450 is accepted (in other words, I'm betting on it being accepted).

Closes #458, #179.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2235,21 +2235,8 @@ closes a stream SHOULD be discarded.  An endpoint MAY choose to limit the period
 over which it ignores frames and treat frames that arrive after this time as
 being in error.
 
-An endpoint will learn the final offset of the data it receives on a stream when
-it enters the "half-closed (remote)".  The final offset is carried explicitly in
-the RST_STREAM frame; otherwise, the final offset is the offset of the end of
-the data carried in STREAM frame marked with a FIN flag.
-
-An endpoint MUST NOT send data on a stream at or beyond the final offset.
-
-Once a final offset for a stream is known, it cannot change.  If a RST_STREAM or
-STREAM frame causes the final offset to change for a stream, an endpoint SHOULD
-respond with a QUIC_STREAM_DATA_AFTER_TERMINATION error (see
-{{error-handling}}).  A receiver SHOULD treat receipt of data at or beyond the
-final offset as a QUIC_STREAM_DATA_AFTER_TERMINATION error.  Generating these
-errors is not mandatory, but only because requiring that an endpoint generate
-these errors also means that the endpoint needs to maintain the final offset
-state for closed streams, which could mean a significant state commitment.
+An endpoint will know the final offset of the data it receives on a stream when
+it reaches the "half-closed (remote)" state, see {{final-offset}} for details.
 
 A stream in this state can be used by the endpoint to send any frame that
 mentions a stream ID.  In this state, the endpoint MUST observe advertised
@@ -2514,6 +2501,31 @@ best to not let the sender go into quiescence if avoidable.  To avoid blocking a
 sender, and to reasonably account for the possibiity of loss, a receiver should
 send a MAX_DATA or MAX_STREAM_DATA frame at least two roundtrips before it
 expects the sender to get blocked.
+
+
+## Stream Final Offset {#final-offset}
+
+The final offset is the count of the number of octets that are transmitted on a
+stream.  For a stream that is reset, the final offset is carried explicitly in
+the RST_STREAM frame.  Otherwise, the final offset is the offset of the end of
+the data carried in STREAM frame marked with a FIN flag.
+
+An endpoint will know the final offset for a stream when the stream enters the
+"half-closed (remote)" state.  However, if there is reordering or loss, an
+endpoint might learn the final offset prior to entering this state if it is
+carried on a STREAM frame.
+
+An endpoint MUST NOT send data on a stream at or beyond the final offset.
+
+Once a final offset for a stream is known, it cannot change.  If a RST_STREAM or
+STREAM frame causes the final offset to change for a stream, an endpoint SHOULD
+respond with a QUIC_STREAM_DATA_AFTER_TERMINATION error (see
+{{error-handling}}).  A receiver SHOULD treat receipt of data at or beyond the
+final offset as a QUIC_STREAM_DATA_AFTER_TERMINATION error, even after a stream
+is closed.  Generating these errors is not mandatory, but only because
+requiring that an endpoint generate these errors also means that the endpoint
+needs to maintain the final offset state for closed streams, which could mean a
+significant state commitment.
 
 
 # Error Handling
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2178,49 +2181,54 @@ A RST_STREAM frame on an "open" stream causes the stream to become
...
 
 ### half-closed (local)
 
 A stream that is in the "half-closed (local)" state MUST NOT be used for sending
-STREAM frames; MAX_STREAM_DATA and RST_STREAM MAY be sent in this state.
-
-A stream transitions from this state to "closed" when a STREAM frame that
-contains a FIN flag is received and all prior data has arrived, or when either
-peer sends a RST_STREAM frame.
+of new STREAM frames.  Retransmission of data that has already been sent on
```
MikeBishop: Nit:  Do we send data "on" STREAM frames?

hardie:
While I agree with Mike that this design could be the fist step toward a unidirectional stream model, it seems to met that it doesn't make sense to switch to that model yet.  Unless/until we get to a point where we stream types that are not always matched by a stream in the "reverse" direction, I think "half-closed" will be more accessible terminology.  The unidirectional stream approach implies capabilities/characteristics that I don't think we're using (yet).

MikeBishop:
@hardie, maybe a separate discussion, but there are stages, and we decide how far we want to go.

1. Stream are bidirectional bytestreams with a single state (idle, open, half-closed, closed)
2. Stream are pairs of unidirectional bytestreams, each of which has a state (idle, open, closed)
3. Streams are unidirectional bytestreams, each of which has a state, and any association between streams in opposite directions needs to be defined and communicated by the app protocol.

I agree that (3) would be a huge departure.  Potentially useful -- but we're not there yet, if ever.  (2), I'm thinking it might be time for.  I don't think that's actually a harder mental model -- if anything, it simplifies the state machine.  The current state machine is really just a cross-product of the two.

hardie:
@MikeBishop  The key question seems to be the extent to which one stream's state can be independent of its peer's.  In model 1, they are tightly coupled.  In model 2, an open stream may be paired with a related stream that is open or idle, but it may not be paired with one that is closed (or you have a unidirectional stream with no way to reset it at the transport layer).  In model 3, a stream may be linked with nothing or to one or more streams and is up to the application to sort it all out.

I think we are still at model 1, even with this description change.  I'm not actually sure I see the advantage of switching to model two unless we believe either that it is a stepping stone to 3 or that there are applications where open streams paired with idle streams makes sense.  I don't have a personal example of that, so it's coming down to "stepping stone to 3" for me.  That's why I'm at "not yet".

I'm hapy to hear if I'm missing something, though.

mikkelfj:
I strongly favor uni-directional streams as in model 3. It may be hard to argue for model 3 when considering a classic request reply model, but if implementing an application protocol agnostic async API for QUIC, the API becomes much simpler in model 3 and will consequently be much less error prone and have a smaller attack surface. If sending a request, the stream can be closed immediately and torn down after sending FIN without waiting for the peer - state is then only needed for retransmission.

There are many use cases where there is no direct request/reply pattern, for example the pub/sub pattern where you subscribe on one write stream, handle errors on a read stream, and have a separate read stream for each subscribed topic. Logging and system monitoring also falls into this category.

I also much approve of the new RST_STREAM behavior, but it has the downside that it is not possible to unilaterally force a stream to be fully closed because DISINTEREST is only advisory.

It is possible to implement to implement silent close at the api level if the receiver accepts a peer stream as either R/W or RO and an initiator creates a stream as R/W or WO. Then a RST or FIN is not needed nor expected on the one half of a stream and consequently resources can be freed earlier, and an error can be issued if the expectations are violated. However, this is much more complicated to manage that uni-directional streams.

If a uni-directional approach is chosen, as in model 3, I believe the receiver should have the power to force the stream to be closed and be able to error the connection after a timeout.

The main problem with handling RPC request/reply over unilateral streams is that the request cannot specify which stream the reply should appear on, unless multiplexing replies on known streams. A requester therefore needs to send a request context and look up that context on any new peer initiated reply stream. Since a new peer stream may have several purposes, this further complicates the lookup since both the stream purpose and request context must now be identified. This is an argument for keeping bi-directional streams. I just don't think it is worth the complication.

Model 2 might have some merit in solving the above RPC problem, but I haven't given this a lot of thought. My intuition tells me that model 2 is largely the same as model 1 with the unidiretional RST_STREAM behaviour, and still has a lot of state complexity.

mikkelfj:
More thoughts on uni-directional RPC pattern:

There could be a REPLY frame type which is similar to a STREAM frame with offset 0 but instead of storing an offset it stores the STREAM ID of the request stream it is replying to. The request stream is an ordinary stream. The replying endpoint must have observed request stream in OPEN state, but the request may be open or closed when replying, and the requesting endpoint need not maintain any state for the request stream. In other words, the link between the two is very weak and unknown to the request stream.

The same could have been obtained at the application layer by carrying a context id for the request, but having the support at transport level has the benefit of automatic ID assignment and no special framing requirements. A REPLY is trivially identified as such.

It is possible to have many logical replies to the same request stream, and there could be many logical request coming in on the request stream. But these are ordinary stream frames with offsets. (This somewhat ties into partially reliable streams with message boundaries, but that is a separate concern).

martinthomson:
Fair warning: I plan to merge this in about 24 hours.

@RyanAtGoogle, I think that @janaiyengar had you on the hook for double-checking this.

martinthomson:
This also closes #179 by virtue of correcting the errors in describing the final offset.

PR #20, opened 2016-11-23
Add support for a hash-based Public Reset packet (#215)

ekr:
The model here is that neither on-path nor off-path attackers should be able to forge one, but
that on-path elements should be able to validate the public reset.

PR #460, opened 2017-04-21
Authenticate public reset with a hash (#215)

### LINKED ISSUE #215: Public reset
There is some sadness about public reset. See PR #20

martinthomson:
This takes @ekr's design from #20 and expands on it quite a bit.  There are a few little wrinkles that I think we might want to discuss a little.

First, this is a one-time password. That means that the combination of key and connection ID can't ever be repeated for a given server instance.  Our 64-bit connection ID space isn't really enough to provide this.  How a server moves to a new static key during operation will be challenging; it probably needs to partition the connection ID space.

Second, transport parameters are encrypted.  #20 made a point of having the verifier in the clear so that intermediaries could validate the Public Reset.  I don't want to create a second TLS extension, but I guess that is the best option we have available to us if this is what is required.  I don't think we should move transport parameters into the clear either.

Closes #215, #20.


martinthomson:
Closing in favour of #574.

PR #96, opened 2016-12-22
Error space

MikeBishop:
Breaks the error space into different regions (#74) for the different components and starts defining error codes in each document.  (Note:  I didn't move anything into the loss-detection document, because I don't know what belongs there versus core transport.  Please help.)

Definitely need review and discussion here -- I may have deleted or moved error codes inappropriately.  Feel free to make changes to this branch.

Issue #211, opened 2017-01-24
Error codes


martinthomson:
There are LOTS of error codes.  The only sensible thing to do is use those code to steer the receiver in the direction of the right logic for recovering from the error.  There is a different piece of information that might be provided to an application for diagnostic purposes (the reason phrase helps if it goes across the wire) and there might be other needs for an API, but across the wire we can probably cut this down a lot.

martinthomson:
Note that #374 fixes this for TLS.

martinthomson:
Moving discussion from #467 here.

@janaiyengar proposed retaining the following error codes.  Here are my reasons for removing them:

If the intent is to signal to the other side, then I can't see how HANDSHAKE_TIMEOUT or NETWORK_IDLE_TIMEOUT would be sent, so I removed it.

Similar to the timeouts, UNENCRYPTED_STREAM_DATA, INVALID_VERSION_NEGOTIATION_PACKET, and ADDRESS_VALIDATION_FAILURE can't be sent in anything.  Maybe if we managed to come to some conclusion about how to signal handshake errors (#608), then maybe this sort of error can be restored.

I have no idea how to signal PUBLIC_RESET_RECEIVED either.

TOO_MANY_RTOS seems to fall into the same category as the timeouts, but I can't tell because it isn't referenced anywhere.

All of the above seem to be best suited to passing across the interface between a QUIC stack and the application that uses it.

I removed CRYPTO_STREAM_CLOSED and PAYLOAD_MISSING on the basis that these were too specific, but I'm happy to retain them.  I'd prefer to open issues tracking their restoration so that we can add text on when they are generated (see below).

Finally, none of these error codes was referenced by text in the document.  If the document doesn't describe the conditions for generating the code, then they clearly weren't that important.

ianswett:
Thanks for enumerating the removed error codes.

I suspect we'll want to reinstate some of those, but none of them seem pressing at the moment.

To clarify, the replacement for HANDSHAKE_TIMEOUT and NETWORK_IDLE_TIMEOUT is NO_ERROR?

MikeBishop:
- CRYPTO_STREAM_CLOSED is duplicative of QUIC_CLOSED_CRITICAL_STREAM, and when you generate it is already covered in the transport doc.
- UNENCRYPTED_STREAM_DATA *might* be sent in something, depending when we receive it.  This actually seems like a worthwhile one to get out:  "You sent stream X != 0 in an unencrypted packet."

[...]

Issue #184, opened 2017-01-20
Error code constraints


LPardue:
QUIC-TRANSPORT defines the syntax of QUIC error codes and defines some values for the QUIC transport. The section says:

> This section lists the defined QUIC transport error codes that may be used in a CONNECTION_CLOSE or RST_STREAM frame. Error codes share a common code space. Some error codes apply only to either streams or the entire connection and have no defined semantics in the other context.

Since the error space is shared, it is not clear what error codes are appropriate to use with which frame types. There is the possibility for endpoints to send a semantically incorrect error code in a frame. How should a receiver handle such an occurrence?

This possibility exists only for the Application-specific, QUIC transport or Cryptographic error codes. Host-local error codes must not be sent.  

As an example, a client sends a `RST_STREAM` frame with the error code `QUIC_TOO_MANY_SESSIONS_ON_SERVER`.  Is this actually allowed and what should the server do (close the connection, igore it, ...)?

MikeBishop:
A convention we used in HTTP/2 with some success is that an endpoint MAY always treat a stream error as connection-fatal if the implementation so chooses.  So any error code is acceptable in a CONNECTION_CLOSE.  I'd advocate for that, at least.

And why isn't GOAWAY mentioned?

MikeBishop:
Overtaken by events.  RST_STREAM no longer contains transport-level error codes (other than STOPPING), and application versus transport _CLOSE frames are separate.

Issue #177, opened 2017-01-19
CONNECTION_CLOSE reliability


ddragana:
>  If the total number of packets transmitted in this connection reaches 2^64 - 1, the sender MUST close the connection by sending a CONNECTION_CLOSE frame with the error code QUIC_SEQUENCE_NUMBER_LIMIT_REACHED

If this packet is lost there is not packet number available for its retransmission. CONNECTION_CLOSE should be reliably transmitted.

marten-seemann:
Related: #192

martinthomson:
Closed by #724.

martinthomson:
I think that I meant #721.

martinthomson:
Fixed in #721.

PR #467, opened 2017-04-27
Prune transport error codes

martinthomson:
This prunes the transport error codes to the short list I shared on the [mailing list](https://mailarchive.ietf.org/arch/msg/quic/xJ5Rr1_N6JI1DT1S7z77G0nZG2Y).

@janaiyengar this rots quickly, I want to close on before we next chat.

Closes #96, #177, #184, #211.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2337,7 +2337,7 @@ Stream IDs are usually encoded as a 32-bit integer, though the STREAM frame
 stream ID are zero.
 
 
-## Life of a Stream
+## Life of a Stream {#stream-states}
 
 The semantics of QUIC streams is based on HTTP/2 streams, and the lifecycle of a
 QUIC stream therefore closely follows that of an HTTP/2 stream {{?RFC7540}},
@@ -2807,6 +2807,10 @@ risks a peer missing the first such packet.  The only mechanism available to an
 endpoint that continues to receive data for a terminated connection is to send a
 Public Reset packet.
 
+An endpoint that receives an invalid error code in a CONNECTION_CLOSE frame MUST
+NOT signal the existence of the error to its peer.  It MAY treat the error as an
+INTERNAL_ERROR on the basis that there is some fault with the peer.
+
 
 ## Stream Errors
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2575,7 +2576,7 @@ affected stream.
 
 Stream 1 is critical to the functioning of the entire connection.  If stream 1
 is closed with either a RST_STREAM or STREAM frame bearing the FIN flag, an
-endpoint MUST generate a connection error of type QUIC_CLOSED_CRITICAL_STREAM.
+endpoint MUST generate a connection error of type PROTOCOL_VIOLATION.
```
janaiyengar: Why remove this error? It's strictly more information, and is surely useful for debugging on the side that's likely messing up.

draft-ietf-quic-transport.md
```diff
@@ -2611,173 +2612,74 @@ CONNECTION_CLOSE or RST_STREAM frame. Error codes share a common code space.
...
+INTERNAL_ERROR (0x80000001):
 
-QUIC_DECOMPRESSION_FAILURE (0x80000018):
-: There was an error decompressing data.
+: The endpoint encountered an internal error and cannot continue with the
+  connection.
 
-QUIC_NETWORK_IDLE_TIMEOUT (0x80000019):
-: The connection timed out due to no network activity.
+ENHANCE_YOUR_CALM (0x80000002):
```
janaiyengar: Where's this intended to be used? It's not clear to me what exactly the endpoint wants the peer to reduce generation of -- Data? Acks? Window updates? -- and what the peer is expected to do when this error code is received. Is this a stream or connection error? It would be an odd time to say reduce generation  of something if used as a connection error... the connection is now closed. Does this error code mean that the peer SHOULD NOT reconnect?

draft-ietf-quic-transport.md
```diff
@@ -2611,173 +2612,74 @@ CONNECTION_CLOSE or RST_STREAM frame. Error codes share a common code space.
...
 
-QUIC_TIMEOUTS_WITH_OPEN_STREAMS (0x8000004a):
-: Disabled QUIC because of too many timeouts with streams open.
+FRAME_FORMAT_ERROR (0x80000009):
 
-QUIC_TOO_MANY_RTOS (0x80000055):
-: QUIC timed out after too many RTOs.
+: An endpoint received a frame that was badly formatted.  For instance, an empty
+  STREAM frame that omitted the FIN flag, or an ACK frame that has more
+  acknowledgment ranges than the remainder of the packet could carry.
```
janaiyengar: These are very different errors in terms of what a peer would do with this information. I don't think collapsing these two examples is a good idea, since the peer can actually do something concrete with this information -- separating STREAM_FRAME_FORMAT_ERROR from ACK_FRAME_FORMAT_ERROR is a useful for monitoring and debugging.

draft-ietf-quic-transport.md
```diff
@@ -2522,13 +2520,12 @@ An endpoint MUST NOT send data on a stream at or beyond the final offset.
 
 Once a final offset for a stream is known, it cannot change.  If a RST_STREAM or
 STREAM frame causes the final offset to change for a stream, an endpoint SHOULD
```
martinthomson: Let's fix/discuss that separately.

draft-ietf-quic-transport.md
martinthomson: See comment below.

draft-ietf-quic-transport.md
```diff
@@ -2565,6 +2562,10 @@ risks a peer missing the first such packet.  The only mechanism available to an
 endpoint that continues to receive data for a terminated connection is to send a
 Public Reset packet.
 
+An endpoint that receives an invalid error code in a CONNECTION_CLOSE frame MUST
+NOT signal the existence of the error to its peer.  It MAY treat the error as an
+INTERNAL_ERROR on the basis that there is some fault with the peer.
```
martinthomson: Yeah, this needs work.  It's trying to say that you can't respond to a CONNECTION_CLOSE with a PROTOCOL_VIOLATION.  Give me another shot at this.

martinthomson:
Gah, this is already rotten :(

I think that you might have misunderstood the approach that I've taken.  The point is to identify classes of errors that can have operational practices associated with them.  Finer grained errors are certainly useful during the development of a protocol, but I don't know how someone would deploy an operational practice to distinguish between a cases where it sent a badly formatted STREAM frame vs. a badly formatted ACK frame.  Those errors are going to need someone to sit down with traces and work out what is going on.

I'm not suggesting that an implementation couldn't generate different error codes for local consumption.  That's great for debugging.  But we're talking protocol signals here.  And when we use a different code point in a protocol, we have to have something in those signals that can be acted upon differently.

draft-ietf-quic-transport.md
```diff
@@ -675,9 +675,9 @@ increase by at least one after sending any packet.
 
 A QUIC endpoint MUST NOT reuse a packet number within the same connection (that
 is, under the same cryptographic keys).  If the packet number for sending
-reaches 2^64 - 1, the sender MUST close the connection by sending a
-CONNECTION_CLOSE frame with the error code QUIC_SEQUENCE_NUMBER_LIMIT_REACHED
-(connection termination is described in {{termination}}.)
+reaches 2^64 - 1, the sender MUST close the connection without sending a
+CONNECTION_CLOSE frame or any further packets; the sender MAY send a Public
+Reset packet in response to further packets that it receives.
```
ianswett: Shouldn't it use the last packet number to send a connection close?

draft-ietf-quic-transport.md
ianswett: Agreed, this does not seem helpful.

draft-ietf-quic-transport.md
ianswett: +1

ianswett:
Generating different error codes for local consumption is not a direction I'd like to go in if I can avoid it.

We have a huge error code space available, so as long as it's clear when one should use an error code, I'm not sure why we wouldn't want a relatively complete set?

draft-ietf-quic-transport.md
martinthomson: See the analogous code in h2.  There isn't a whole lot of text in QUIC for this, but you can find the sorts of applications we use this for in h2.

draft-ietf-quic-transport.md
martinthomson: Well, that is a possibility, but think about what it will take to reach this number.  We could put lots of effort into dealing with this case, but you have to ask yourself if that's worth it.

draft-ietf-quic-transport.md
ianswett: If it's already in H2, should it be in the http mapping?  I can't find a good link for H2's usage, the best I found was it's informal use by twitter: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

And why is it not error code 420?

martinthomson: My understanding is that it is part of the transport in h2, not the application.  If you disagree with that assessment, let me know.

It was 420, for a while, but then we decided that a contiguous error space was more valuable.  However, it's *not* a status code.

draft-ietf-quic-transport.md
janaiyengar: Ah, ok, that makes sense.

draft-ietf-quic-transport.md
janaiyengar: If this has a specific meaning, I would want to name it as such. "ENHANCE_YOUR_CALM" is not a helpful name... but I still don't understand what it stands for. What is this used for in HTTP/2?

janaiyengar:
It is very helpful operationally to have these codes available, since a server, for instance, can monitor the different error codes that are received. A spike in FRAME_ERROR for a STREAM frame is much easier to diagnose than a general FRAME_ERROR. As to why this may be useful later and not just during development -- any time you touch code, you can inadvertently introduce bugs. A server that does not want to monitor does not have to, but having the error codes makes it possible for interested parties to do so.

draft-ietf-quic-transport.md
martinthomson: Better that you search the RFC for the code and read it yourself.

janaiyengar: I don't think I said anything to warrant a rude response. Perhaps you didn't mean it to be rude, but it sure sounds it.

I know of the error code from HTTP/2, and I looked it up before my first comment on this PR. RFC 7540 is not helpful in answering the questions I'm asking.

martinthomson: I apologize, I just think that me trying to restate its purpose wouldn't really help because it's got a lot of context attached to it.

draft-ietf-quic-transport.md
ianswett: Ok, fair enough.  I'm happy handling that edge case in whatever way is easiest.

draft-ietf-quic-transport.md
ianswett: link?

draft-ietf-quic-transport.md
janaiyengar: Yeah, I agree. I also realized that this max packet number is known to the receiver, and the receiver knows that the sender is near the max, so clearly it knows that the sender would close the connection.

draft-ietf-quic-transport.md
```diff
@@ -2858,173 +2858,74 @@ CONNECTION_CLOSE or RST_STREAM frame. Error codes share a common code space.
 Some error codes apply only to either streams or the entire connection and have
 no defined semantics in the other context.
 
-QUIC_INTERNAL_ERROR (0x80000001):
```
ianswett: you'll want this or some equivalent.

draft-ietf-quic-transport.md
```diff
@@ -2858,173 +2858,74 @@ CONNECTION_CLOSE or RST_STREAM frame. Error codes share a common code space.
...
-QUIC_CLOSED_CRITICAL_STREAM (0x80000007):
-: A stream that is critical to the protocol was closed.
-
-QUIC_MISSING_PAYLOAD (0x80000030):
-: The packet contained no payload.
-
-QUIC_INVALID_STREAM_DATA (0x8000002E):
-: STREAM frame data is malformed.
-
-QUIC_UNENCRYPTED_STREAM_DATA (0x8000003D):
```
ianswett: We definitely want this.

draft-ietf-quic-transport.md
```diff
@@ -2858,173 +2858,74 @@ CONNECTION_CLOSE or RST_STREAM frame. Error codes share a common code space.
...
-QUIC_MAYBE_CORRUPTED_MEMORY (0x80000059):
-: Received a frame which is likely the result of memory corruption.
-
-QUIC_INVALID_RST_STREAM_DATA (0x80000006):
-: RST_STREAM frame data is malformed.
-
-QUIC_INVALID_CONNECTION_CLOSE_DATA (0x80000007):
-: CONNECTION_CLOSE frame data is malformed.
-
-QUIC_INVALID_GOAWAY_DATA (0x80000008):
```
ianswett: I do want an error code per frame type.  It really helps narrow down issues.

draft-ietf-quic-transport.md
ianswett: Still confusing, particularly on a transport level.

And if we're going to add it, we're going to need to restate it's purpose, we can't just assume readers know what it means.

seanturner: From RFC 7540:

    ENHANCE_YOUR_CALM (0xb):  The endpoint detected that its peer is
    exhibiting a behavior that might be generating excessive load.

martinthomson:
Discussed in Paris; we're going to add the frame-specific codes in *addition* to the generic one.

We'll remove ENHANCE_YOUR_CALM from this one.

draft-ietf-quic-transport.md
```diff
@@ -1213,7 +1213,7 @@ server MUST check that it would have sent a version negotiation packet if it had
 received a packet with the indicated initial_version.  If a server would have
 accepted the version included in the initial_version and the value differs from
 the value of negotiated_version, the server MUST terminate the connection with a
-QUIC_VERSION_NEGOTIATION_MISMATCH error.
+VERSION_NEGOTIATION_ERROR error.
```
mikkelfj: I am having some trouble understanding how to send this error https://github.com/quicwg/base-drafts/issues/597 - but it might be resolved with updated PUBLIC_RESET to replace CONNECTION_CLOSE.

martinthomson:
@janaiyengar, I've rebased and updated this with what we agreed at the interim.  There is now a generic frame format error, and a set of frame-type-specific errors.  I removed ENHANCE_YOUR_CALM.

I went over the error codes that you added back in #521 and I'm going to suggest that we separate those out.  Some depend on the conclusion of #608, and I *think* that the others aren't useful to signal.  I'm sure that I'm wrong about at least one, but we should discuss them separately.

janaiyengar:
I'm not sure which ones from #521 are not added in here. I wouldn't remove ones that we need discussion on... If you want to discuss ones that are in #521, please add them back in and have a separate issue on removing them.

PR #468, opened 2017-04-27
Make TLS track changes in transport

martinthomson:
This is a small change to follow #421 and friends.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1186,16 +1186,22 @@ obtained from protected messages, information obtained from less reliable
 sources can be discarded.
 
 
-### WINDOW_UPDATE Frames
+### Updates to Data and Stream Limits
 
-`WINDOW_UPDATE` frames MUST NOT be sent unprotected.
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-tls.md
```diff
@@ -1186,17 +1186,23 @@ obtained from protected messages, information obtained from less reliable
...
+frames MUST NOT be sent unprotected.
 
-Though data is exchanged on stream 1, the initial flow control window is
-sufficiently large to allow the TLS handshake to complete.  This limits the
-maximum size of the TLS handshake and would prevent a server or client from
-using an abnormally large certificate chain.
+Though data is exchanged on stream 1, the initial flow control window on that
+stream is sufficiently large to allow the TLS handshake to complete.  This
+limits the maximum size of the TLS handshake and would prevent a server or
+client from using an abnormally large certificate chain.
```
ekr: Isn't this technically just a restriction on a single RT?

martinthomson: Well, if the server wants to send a 1M cert, then the client can't send MAX_STREAM_DATA to allow that to happen.  It will run into the flow control limit (which I assume is 64k, but we don't actually say that).

I know that there are large certs, and there are probably certs with OCSP and associated gunk that will exceed this limit.  Those certs won't work with QUIC.

PR #470, opened 2017-04-27
Small editorial fixes to stream identifiers

### LINKED ISSUE #209: Fix stream identifier size
The doc defines stream identifiers as "variable-length".  But are they really?

martinthomson:
Closes #209.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1342,7 +1342,7 @@ Data Length:
 
 Stream ID:
 
-: The stream ID of the stream.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1347,7 +1347,7 @@ Data Length:
 
 Stream ID:
 
-: A variable-sized unsigned ID unique to this stream.
+: The stream ID of the stream.
```
janaiyengar: Add a reference to #stream-identifiers.

PR #472, opened 2017-04-27
Define the payload size of the initial packet rather than the total packet size

### LINKED ISSUE #267: Server enforcement of 1280 octet packet size
PR #106 added the following requirement on servers:

> Servers MUST reject the first plaintext packet received from a client if it its total size is less than 1280 octets, to mitigate amplification attacks.

However, this runs into an issue in the case where there are IP header extensions that the client knows about but the server doesn't.  If the client is aware of the IP header extensions and constructs a packet with a smaller payload in order to comply with our recommendation to create a packet of exactly 1280 octets, the server will receive a packet that it thinks is too small.

I see a few options:
1. loosen the enforcement constraint on the server
2. specify that the server only enforce the constraint if it sees the entire IP packet
3. specify the minimum size and its enforcement in terms of payload size (we could set different values here to allow for header extensions)

martinthomson:
The entity that creates the packet is not always aware of the existence of header extensions.  That makes hitting the 1280 octet limit hard.  Worse, if they are aware of extensions, they might reduce the size and then encounter a server that enforces the limit.  That server will discard the packet.  That makes enforcement risky.

This change places the limit on the payload.  It mandates 1232/1252 for v6/v4.

This is the uncomplicated version.  The complicated version that I didn't write up allows a client that knows more about overheads to reduce the payload size as long as they keep above some floor value (e.g., 1180).  We might consider doing that later if we want to allow QUIC to operate on paths with a much tighter MTU overhead.  Evidence suggests that we don't need this though (open an issue if you disagree).

Closes #267

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -533,9 +533,8 @@ older than 1.3 is negotiated.
 ## ClientHello Size
 
 QUIC requires that the initial handshake packet from a client fit within the
-payload of a single packet.  A minimum-sized initial packet in IPv6 has space
-for a single TLS record of 1197 octets once the QUIC header and minimal framing
-is added.
+payload of a single packet.  The size limits on QUIC packets mean that a record
+containing a ClientHello needs to fit within 1197 octets.
 
 A TLS ClientHello can fit within this limit with ample space remaining.
 However, there are several variables that could cause this limit to be exceeded.
@@ -549,9 +548,9 @@ extension can have an effect on a client's ability to connect.  Choosing a small
 value increases the probability that these values can be successfully used by a
 client.
 
-A TLS implementation does not need to enforce this size constraint.  QUIC
-padding can be used to reach this size, meaning that a TLS server is unlikely to
-receive a large ClientHello message.
+The TLS implementation does not need to ensure that the ClientHello is
+sufficiently large.  QUIC PADDING frames are added to increase the size of the
+packet as necessary.
 
 
 ## Peer Authentication
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -532,9 +532,10 @@ older than 1.3 is negotiated.
 
 ## ClientHello Size
 
-QUIC requires that the initial handshake packet from a client fit within a
-single packet of at least 1280 octets.  With framing and packet overheads this
-value could be reduced.
+QUIC requires that the initial handshake packet from a client fit within the
+payload of a single packet.  A minimum-sized initial packet in IPv6 has space
+for a single TLS record of 1197 octets once the QUIC header and minimal framing
+is added.
 
 A TLS ClientHello can fit within this limit with ample space remaining.
 However, there are several variables that could cause this limit to be exceeded.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1137,7 +1137,7 @@ on its own.
 
 Several methods are used in QUIC to mitigate this attack.  Firstly, the initial
 handshake packet is padded to at least 1280 octets.  This allows a server to
-send a similar amount of data without risking causing an amplication attack
+send a similar amount of data without risking causing an amplification attack
 toward an unproven remote address.
 
 A server eventually confirms that a client has received its messages when the
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1957,7 +1957,7 @@ appropriately, and storing the result of previous PMTU determinations.
 
 In the absence of these mechanisms, QUIC endpoints SHOULD NOT send IP packets
 larger than 1280 octets. Assuming the minimum IP header size, this results in
-a UDP payload length of 1232 octets for IPv6 and 1252 octets for IPv4.
+a QUIC packet size of 1232 octets for IPv6 and 1252 octets for IPv4.
 
 QUIC endpoints that implement any kind of PMTU discovery SHOULD maintain an
 estimate for each combination of local and remote IP addresses (as each pairing
@@ -1969,10 +1969,10 @@ An endpoint MUST NOT reduce their MTU below this number, even if it receives
 signals that indicate a smaller limit might exist.
 
 Clients MUST ensure that the first packet in a connection, and any
-retransmissions of those octets, has a payload of least 1232 octets for an IPv6
-datagram and 1252 octets for an IPv4 datagram.  In the absence of extensions to
-the IP header, padding to exactly these values will result in a packet that is
-1280 octets.
+retransmissions of those octets, has a QUIC packet size of least 1232 octets for
+an IPv6 packet and 1252 octets for an IPv4 packet.  In the absence of extensions
+to the IP header, padding to exactly these values will result in an IP packet
+that is 1280 octets.
 
 The initial client packet SHOULD be padded to exactly these values unless the
 client has a reasonable assurance that the PMTU is larger.  Sending a packet of
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -913,7 +913,7 @@ prior to establishing a connection, exposing the server to a denial of service
 risk.
 
 The first client packet of the cryptographic handshake protocol MUST fit within
-a 1280 octet QUIC packet.  This includes overheads that reduce the space
+a 1200 octet QUIC packet.  This includes overheads that reduce the space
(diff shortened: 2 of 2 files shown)


martinthomson:
@janaiyengar, this has been discussed and I'm just enacting the agreed solution.  I'd like to merge this before 2017-05-02.

draft-ietf-quic-tls.md
```diff
@@ -532,9 +532,10 @@ older than 1.3 is negotiated.
 
 ## ClientHello Size
 
-QUIC requires that the initial handshake packet from a client fit within a
-single packet of at least 1280 octets.  With framing and packet overheads this
-value could be reduced.
+QUIC requires that the initial handshake packet from a client fit within the
+payload of a single packet.  A minimum-sized initial packet in IPv6 has space
+for a single TLS record of 1197 octets once the QUIC header and minimal framing
+is added.
```
ekr: This is phrased awkwardly, and also seems to semi-conflict with {{#packetization}}.

I would consolidate these, and you also need to fix the requirement in {{#packetization}} that things be exactly 1280 because the client has no control over that.

I think rather what we need to say is:

1. The client needs to ensure that the initial packet has at least X bytes of UDP payload
2. When you add the QUIC header, that means CH can be up to Y bytes.
3. The rest of the space MUST be used by PADDING frames.

Based on the text, I believe that X == 1232 (I wouldn't bother to adjust between v4 and v6).
How do you get 1197? If we have 1232 bytes for UDP payload and then a 17-byte minimal QUIC header and a 3 byte STREAM frame header....
   
   
And I would put all of this in this space and leave the general guidance in {{#packetization}}.

martinthomson: A STREAM frame header can be two octets.  And then there is the FNV-1a hash, which I know isn't in the current draft, but I'm trying to avoid having to update this once that lands.

I really only wanted to say point 2 here.  I'll take another stab at it.

draft-ietf-quic-transport.md
```diff
@@ -1138,9 +1138,9 @@ use the server to send more data toward the victim than it would be able to send
 on its own.
 
 Several methods are used in QUIC to mitigate this attack.  Firstly, the initial
-handshake packet from a client is padded to at least 1280 octets.  This allows a
-server to send a similar amount of data without risking causing an amplication
-attack toward an unproven remote address.
+handshake packet is padded to at least 1280 octets.  This allows a server to
+send a similar amount of data without risking causing an amplication attack
```
janaiyengar: typo: amplication --> amplification

draft-ietf-quic-transport.md
```diff
@@ -1952,16 +1952,19 @@ An endpoint MUST NOT reduce their MTU below this number, even if it receives
...
-a reasonable assurance that the PMTU is larger. Sending a packet of this size
-ensures that the network path supports an MTU of this size and helps mitigate
-amplification attacks caused by server responses toward an unverified client
-address.
-
-Servers MUST reject the first plaintext packet received from a client if it its
-total size is less than 1280 octets, to mitigate amplification attacks.
+retransmissions of those octets, has a payload of least 1232 octets for an IPv6
+datagram and 1252 octets for an IPv4 datagram.  In the absence of extensions to
+the IP or UDP header, padding to exactly these values will result in a packet
```
janaiyengar: Do UDP header extensions exist?

draft-ietf-quic-tls.md
janaiyengar: Ekr: There's 16 octets of FNV hash. I think it's off too though, since adding 16 bytes to your number (which I believe is correct otherwise) gives 1233 bytes.

I don't think it's necessary to articulate how many bytes of QUIC payload, since it just gets confusing. I think it's enough to say 1252 and 1232 bytes of UDP payload (or QUIC packet size) for v4 and v6. I think it's useful and important to separate v4 and v6 -- 20 bytes in the common v4 case is a fair bit.

draft-ietf-quic-transport.md
ekr: No, I think not

ekr:
Well, as I understood it, MT was trying to say how big the ClientHello could be, and that's the thing that fits in the frame. I could live with either set of calculations as long as the numbers are correct.

With regard to v4 or v6, do you plan to do happy eyeballs by sending the QUIC packet 0 over both paths? In that case, you need a CH which will fit in either versio, as you can't change it later.

draft-ietf-quic-transport.md
janaiyengar: Then I would drop "or UDP" form the sentence

janaiyengar:
I understand what MT was trying to say. I prefer the simpler 1252/1232 version since it requires less description, you don't have to make assumptions about packet number size, and we don't have to keep updating if the QUIC header changes.

Good point about v4/v6 -- I think let's leave this description as it is, and, if/when we describe happy eyeballs, we can discuss this point. FWIW, happy eyeballs is not trivial with QUIC because of 0-RTT data. We should probably describe this somewhere.

martinthomson:
There's a potential simplification here.  Don't try to hit a particular target, just specify a single target.  1232 would work, as would any value in that territory.  The lower we go, the greater the exposure to amplification attack, but the difference between 1232 and 1200 is pretty academic here.

janaiyengar:
Perhaps I'm missing something, so let me know if so... but what needs
simplifying here? I don't think it's particularly complicated to say that
the QUIC packet carrying the CH MUST fit within 1232 octets for IPv6 and
1252 octets for IPv4.

martinthomson:
Compare: "The initial packet MUST fit within 1232 octets." with "The initial packets MUST fit within 1232 octets if carried in IPv6 and 1252 octets if carried in IPv4."  The former is much easier, especially with the comment about happy eyeballs.

janaiyengar:
I don't think this complexity warrants simplification... I don't think
you're gaining anything here. The simplification doesn't resolve happy
eyeballs, since happy eyeballs was written for a 1-RTT handshake world...
you don't want to simply send all 0-RTT data on both the v6 and the v4
paths.

martinthomson:
We still have a 1-RTT handshake for initial connections (return connections are also highly likely not to need happy eyeballs).

draft-ietf-quic-transport.md
```diff
@@ -1952,16 +1952,19 @@ An endpoint MUST NOT reduce their MTU below this number, even if it receives
...
-ensures that the network path supports an MTU of this size and helps mitigate
-amplification attacks caused by server responses toward an unverified client
-address.
-
-Servers MUST reject the first plaintext packet received from a client if it its
-total size is less than 1280 octets, to mitigate amplification attacks.
+retransmissions of those octets, has a payload of least 1232 octets for an IPv6
+datagram and 1252 octets for an IPv4 datagram.  In the absence of extensions to
+the IP header, padding to exactly these values will result in a packet that is
+1280 octets.
```
janaiyengar: "packet" --> "UDP datagram"

draft-ietf-quic-transport.md
```diff
@@ -1952,16 +1952,19 @@ An endpoint MUST NOT reduce their MTU below this number, even if it receives
...
-RECOMMENDED that a packet be padded to exactly 1280 octets unless the client has
-a reasonable assurance that the PMTU is larger. Sending a packet of this size
-ensures that the network path supports an MTU of this size and helps mitigate
-amplification attacks caused by server responses toward an unverified client
-address.
-
-Servers MUST reject the first plaintext packet received from a client if it its
-total size is less than 1280 octets, to mitigate amplification attacks.
+retransmissions of those octets, has a payload of least 1232 octets for an IPv6
+datagram and 1252 octets for an IPv4 datagram.  In the absence of extensions to
```
janaiyengar: use of "payload" is confusing. Suggest changing to "QUIC packet size" or "UDP payload"

janaiyengar:
Why not? If your client's local network changes, you'd want to do happy eyeballs again since you don't know if v6 is supported. But you'd still want to do 0-RTT since the server's the same. My point however was to say that happy eyeballs is not trivial, and needs its own discussion. CH size is the easiest of the issues to address in that discussion.

I don't feel very strongly, so I'll go with what you have.

draft-ietf-quic-transport.md
martinthomson: I'll use "IP packet", since UDP implies that it doesn't include the IP header and MTU calculations generally talk about IP packet sizes.

martinthomson:
I don't really care about the simplification, I just keep trying to find ways to simplify.  Anything that reduces cognitive load.

PR #476, opened 2017-04-27
Don't exempt stream 1 from congestion control

### LINKED ISSUE #248: Exemption from congestion control

martinthomson:
Closes #248


martinthomson:
@janaiyengar, also trivial.  2017-05-02 is my target date.

PR #478, opened 2017-04-28
MAX_DATA/MAX_STREAM_DATA will also solicit an ACK

### LINKED ISSUE #385: Soliciting ACKs
We now forbid the generation of an ACK in response to an ACK (lest the two endpoints talk each other to death over nothing), rather than recommend acknowledging only a fraction of ACKs.  That's great, and much easier to implement.

However, when removing STOP_WAITING, @ianswett wanted to have a way to solicit an ACK so that a receiver could stop sending acknowledgments for some packets.  He suggested using PING, which is great.  I have another suggestion that I think is superior: WINDOW_UPDATE.  Like PING, WINDOW_UPDATE causes an ACK to be generated, albeit with less urgency and a receiver is highly like to need to send one occasionally.  I think that PING is still viable, but WINDOW_UPDATE is likely to happen naturally anyway.

martinthomson:
Closes #385


draft-ietf-quic-transport.md
```diff
@@ -1387,9 +1387,13 @@ packets.
...
 an ACK frame has been acknowledged, the packets it acknowledges SHOULD not be
-acknowledged again.  To handle cases where the receiver is only sending ACK
-frames, and hence will not receive acknowledgments for its packets, it MAY send
-a PING frame at most once per RTT to explicitly request acknowledgment.
+acknowledged again.
+
+A receiver that is only sending ACK frames will not receive acknowledgments for
+its packets.  Sending an occasional MAX_DATA or MAX_STREAM_DATA frame as data is
+received will ensure that acknowledgements are generated by a peer.  Otherwise,
+an endpoint MAY send a PING frame at most once per RTT to solicit
```
janaiyengar: This is in the old text, but it's odd to have "MAY send .. at most once per RTT." I'd suggest removing the "at most" and changing this to "MAY send ... once per RTT."

PR #479, opened 2017-04-28
Don't require the client to reset excess streams

### LINKED ISSUE #425: Do not require clients to reset initial streams
The draft currently requires clients to reset initial streams beyond the value declared by the server.
> Similarly, a client might exceed the initial stream limit declared by the server.  A client MUST reset any streams that exceed this limit.

This can be problematic for applications where stream IDs are meaningful, since the client will never be able to reopen a previously-reset stream ID. It is also wasteful of bandwidth, in case the server is willing to accept the extra data.

Also, the paragraph in its entirety is either unclear or self-contradictory:
>A client MUST reset any streams that exceed this limit. A server SHOULD reset any streams it cannot handle [...]

If the client MUST reset extra streams, whether server can or cannot handle the extra streams becomes irrelevant -- they are being reset!

Instead, we should require that the client stops sending any additional STREAM frames for the extra streams (even retransmissions) until the server indicates that it is ready for those streams via LIMIT_UPDATE (#419).  The server should be free to reset what it cannot handle (but for an application with meaningful stream IDs this would render the connection useless).

martinthomson:
In 0-RTT, the client might overrun the server's stream limit.  Let the server
work out what to do about that.

Closes #425

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1045,9 +1045,8 @@ particularly with respect to transport parameters that establish limits:
   limit is received, the client can recommence sending.
 
 * Similarly, a client might exceed the initial stream limit declared by the
-  server.  A client MUST reset any streams that exceed this limit.  A server
-  SHOULD reset any streams it cannot handle with a code that allows the client
-  to retry any application action bound to those streams.
+  server.  A server MAY reset any streams it cannot handle with a code that
+  allows the client to retry any application action bound to those streams.
 
 A server MAY close a connection if remembered or assumed 0-RTT transport
 parameters cannot be supported, using an error code that is appropriate to the


draft-ietf-quic-transport.md
```diff
@@ -1045,9 +1045,8 @@ particularly with respect to transport parameters that establish limits:
   limit is received, the client can recommence sending.
 
 * Similarly, a client might exceed the initial stream limit declared by the
-  server.  A client MUST reset any streams that exceed this limit.  A server
-  SHOULD reset any streams it cannot handle with a code that allows the client
-  to retry any application action bound to those streams.
+  server.  A server MAY reset any streams it cannot handle with a code that
+  allows the client to retry any application action bound to those streams.
```
ekr: Which code?

martinthomson: Oh, I'm waiting for #467 to land before doing that (it's CANCELLED).

PR #480, opened 2017-04-28
Add a description of the FNV-1a algorithm

### LINKED ISSUE #167: Add a hash for unencrypted packets
QUIC originally specified a 12 byte hash for unencrypted packets.  This was dropped when QUIC crypto was swapped for the TLS 1.3 mapping.

We should add it back, since it is useful in eliminating corrupted packets and random internet junk.  And possibly in preventing reflection attacks mentioned in #147

martinthomson:
I've added this in the TLS document, even though this would have been easier to
add to the transport document in some ways.  The current plan (if we ever clear
our backlog) is to move the details of the packet protection and key schedule
to the transport.  If that ever happens, it will be easy enough to move this
new text at the same time as the packet protection.

Two choices that I made here:

1. all packets are protected, including Public Reset - this means that the
packet construction code can have very simple logic for determining whether to
add an integrity check, namely is there an AEAD?; the cost being that public
reset is a little bigger

2. the integrity check mechanism can change in a new version - this avoids
expanding the list of things that we promise that won't change between
versions; the cost being one extra trivial check on receiving an unprotected
packet

Closes #167.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -83,15 +83,6 @@ informative:
         org: Google
         role: editor
 
-  FNV:
-    title: "FNV Hash"
-    date: 2015-01-07
-    author:
-      -
-        ins: L. Noll
-        name: Landon Curt Noll
-    target: "http://www.isthe.com/chongo/tech/comp/fnv/"
-
 
 --- abstract
 
@@ -853,24 +844,19 @@ indicate that the packet will be protected (that is, 0-RTT Encrypted (0x05),
 1-RTT Encrypted (key phase 0) (0x06), or 1-RTT Encrypted (key phase 1) (0x07))
 first constructs the packet that it sends without the integrity check.
 
-In constructing the packet, the endpoint has to consider the extra length that
-the integrity check will add relative to the path MTU.  In particular, the
-integrity check length needs to be subtracted from available space when padding
-the initial client packet.
-
 The sender then calculates the integrity check over the entire packet, starting
 from the type field.  The output of the hash is appended to the packet.
 
 A receiver that receives an unprotected packet first checks that the version is
-correct, then removes the trailing octets.  It calculates the integrity check
+correct, then removes the trailing 16 octets.  It calculates the integrity check
 over the remainder of the packet.  Unprotected packets that do not contain a
 valid integrity check MUST be discarded.
 
 
 ## The 128-bit FNV-1a Algorithm {#fnv1a}
 
-QUIC uses the 128-bit version of the alternative Fowler/Noll/Vo hash, or FNV-1a
-{{FNV}}.
+QUIC uses the 128-bit version of the alternative Fowler/Noll/Vo hash (FNV-1a)
+{{?FNV=I-D.eastlake-fnv}}.
 
 FNV-1a can be expressed in pseudocode as:
 
@@ -891,7 +877,7 @@ The offset basis for the 128-bit FNV-1a is the decimal value
 (in hex, 0x1000000000000000000013b; or as an expression 2^88 + 2^8 + 0x3b).
 
 Once all octets have been processed in this fashion, the final integer value is
-encoded on 16 octets in network byte order.
+encoded as 16 octets in network byte order.
 
 
 # Key Phases
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -83,6 +83,15 @@ informative:
         org: Google
         role: editor
 
+  FNV:
+    title: "FNV Hash"
+    date: 2015-01-07
+    author:
+      -
+        ins: L. Noll
+        name: Landon Curt Noll
+    target: "http://www.isthe.com/chongo/tech/comp/fnv/"
+
 
 --- abstract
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -821,6 +830,73 @@ used in QUIC can cause packet numbers to be decoded incorrectly if they are
...
+A receiver that receives an unprotected packet first checks that the version is
+correct, then removes the trailing octets.  It calculates the integrity check
+over the remainder of the packet.  Unprotected packets that do not contain a
+valid integrity check MUST be discarded.
+
+An implementation could choose to treat the addition and removal of the
+integrity check as though it were an AEAD.
+
+
+## The 128-bit FNV-1a Algorithm {#fnv1a}
```
janaiyengar: Remove this entirely, and leave a reference to the FNV IETF draft (https://datatracker.ietf.org/doc/html/draft-eastlake-fnv).

draft-ietf-quic-tls.md
```diff
@@ -821,6 +830,73 @@ used in QUIC can cause packet numbers to be decoded incorrectly if they are
...
+## Integrity Check Processing
+
+An endpoint sending a packet that has a long header and a type that does not
+indicate that the packet will be protected (that is, 0-RTT Encrypted (0x05),
+1-RTT Encrypted (key phase 0) (0x06), or 1-RTT Encrypted (key phase 1) (0x07))
+first constructs the packet that it sends without the integrity check.
+
+In constructing the packet, the endpoint has to consider the extra length that
+the integrity check will add relative to the path MTU.  In particular, the
+integrity check length needs to be subtracted from available space when padding
```
janaiyengar: "when padding the initial client packet" --> "when generating unprotected client packets".

draft-ietf-quic-tls.md
```diff
@@ -821,6 +830,73 @@ used in QUIC can cause packet numbers to be decoded incorrectly if they are
...
+The sender then calculates the integrity check over the entire packet, starting
+from the type field.  The output of the hash is appended to the packet.
+
+A receiver that receives an unprotected packet first checks that the version is
+correct, then removes the trailing octets.  It calculates the integrity check
+over the remainder of the packet.  Unprotected packets that do not contain a
+valid integrity check MUST be discarded.
+
+An implementation could choose to treat the addition and removal of the
+integrity check as though it were an AEAD.
```
janaiyengar: Unnecessary -- remove this sentence

draft-ietf-quic-tls.md
```diff
@@ -821,6 +830,73 @@ used in QUIC can cause packet numbers to be decoded incorrectly if they are
...
+1-RTT Encrypted (key phase 0) (0x06), or 1-RTT Encrypted (key phase 1) (0x07))
+first constructs the packet that it sends without the integrity check.
+
+In constructing the packet, the endpoint has to consider the extra length that
+the integrity check will add relative to the path MTU.  In particular, the
+integrity check length needs to be subtracted from available space when padding
+the initial client packet.
+
+The sender then calculates the integrity check over the entire packet, starting
+from the type field.  The output of the hash is appended to the packet.
```
janaiyengar: Google QUIC adds this hash right after the packet header. This is similar to how AEAD does it, where the hash appears just before the encrypted payload, and it is indeed how we generate these packets (null encrypter with an FNV hash.) Can you change this to have the hash after the packet header?

draft-ietf-quic-tls.md
martinthomson: Nope, other client packets don't have strong requirements about length, but that one needs to be handled specially (the second flight of handshake packets doesn't have a size constraint).

draft-ietf-quic-tls.md
martinthomson: I'd prefer to have it at the end.  Then it's not wedged in the middle so that you have to hash 17 octets, jump 16 and then hash the rest.

draft-ietf-quic-tls.md
martinthomson: That draft has an uncertain future.  It's not even an IETF draft.  I don't want to hitch our wagon to something like that.

It's even more uncertain given that the market is now overwhelmingly using Murmur variants.

draft-ietf-quic-tls.md
martinthomson: Note also that "AEAD" doesn't do it any particular way.  AES-GCM adds the authentication tag at the end.

ekr: I agree with MT here. While there's no rules about where authentication tags for AEAD, as a practical matte everyone thinks they go last for the reason MT indicates. Is there any technical reason it should go first?

draft-ietf-quic-tls.md
```diff
@@ -821,6 +830,70 @@ used in QUIC can cause packet numbers to be decoded incorrectly if they are
...
+
+A receiver that receives an unprotected packet first checks that the version is
+correct, then removes the trailing octets.  It calculates the integrity check
```
ekr: trailing 16 octets.

draft-ietf-quic-tls.md
```diff
@@ -821,6 +830,70 @@ used in QUIC can cause packet numbers to be decoded incorrectly if they are
...
+over the remainder of the packet.  Unprotected packets that do not contain a
+valid integrity check MUST be discarded.
+
+
+## The 128-bit FNV-1a Algorithm {#fnv1a}
+
+QUIC uses the 128-bit version of the alternative Fowler/Noll/Vo hash, or FNV-1a
+{{FNV}}.
```
ekr: (FNV-1a), not "or" because these are the same thing, not alternatives

draft-ietf-quic-tls.md
```diff
@@ -821,6 +830,70 @@ used in QUIC can cause packet numbers to be decoded incorrectly if they are
...
+for each octet of the input, the exclusive binary OR of the value is taken, then
+multiplied by a prime.  Any overflow from multiplication is discarded.
+
+The offset basis for the 128-bit FNV-1a is the decimal value
+144066263297769815596495629667062367629 (in hex,
+0x6c62272e07bb014262b821756295c58d).  The prime is 309485009821345068724781371
+(in hex, 0x1000000000000000000013b; or as an expression 2^88 + 2^8 + 0x3b).
+
+Once all octets have been processed in this fashion, the final integer value is
+encoded on 16 octets in network byte order.
```
ekr: encoded as.

ekr:
I agree we need to document this here. Having a core piece of QUIC farmed out to some individual draft is not good. If someone wants to resurrect draft-eastlake as a QUIC document, then that would be OK I guess.

draft-ietf-quic-tls.md
janaiyengar: My bad -- I had assumed that AEAD convention was to do it at the beginning, but I was wrong. Trailing hash SGTM.

draft-ietf-quic-tls.md
janaiyengar: I see what you were trying to describe. This process generally applies to *all* packets though, which an implementer needs to be careful about if the packet is to fit within the max packet size.

janaiyengar:
Ok with documenting FNV here, but I'd prefer the draft as a reference instead of the website.

draft-ietf-quic-tls.md
martinthomson: Good point.

draft-ietf-quic-tls.md
```diff
@@ -821,6 +821,65 @@ used in QUIC can cause packet numbers to be decoded incorrectly if they are
 delayed significantly.
 
 
+# Unprotected Packets
+
+QUIC adds an integrity check to all unprotected packets.  Any packet that is not
+protected by the negotiated AEAD (see {{packet-protection}}), includes a
+integrity check.  This check does not prevent the packet from being altered, it
```
mikkelfj: an

PR #482, opened 2017-04-28
Address validation is stateless reject

### LINKED ISSUE #60: Stateless Reject mechanism needs description
Also, Martin Duke observes that it's not clear how QUIC would determine that the crypto handshake was incomplete after one message.

I note that this might be aided by #35.

martinthomson:
We already have enough text on how to tell a client to go away and come back.
But we don't really say how a server might do that statelessly.  Add a small
bit of text on that and we're done.

Closes #60

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1058,6 +1058,10 @@ validation was not requested originally.  In such cases, the cookie extension
 could either be absent or it could indicate that an address validation token is
 not present.
 
+A server can use the cookie extension to store all state necessary to continue
+the connection.  This allows a server to avoid committing state for clients that
+have unvalidated source addresses.
+
 
 ## NewSessionTicket Address Validation
 
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1058,9 +1058,26 @@ validation was not requested originally.  In such cases, the cookie extension
 could either be absent or it could indicate that an address validation token is
 not present.
 
+
+### Stateless Address Validation
+
 A server can use the cookie extension to store all state necessary to continue
 the connection.  This allows a server to avoid committing state for clients that
 have unvalidated source addresses.
+
+For instance, a server could use a statically-configured key to encrypt the
+information that it requires and include that information in the cookie.  In
+addition to address validation information, a server that uses encryption also
+needs to be able recover the hash of the ClientHello and its length, plus any
+information it needs in order to reconstruct the HelloRetryRequest.
+
+A client proceeds as though the server were stateful.  It constructs a second
+ClientHello and sends it following the initial ClientHello.  Note that this
+means a server that doesn't maintain state between the intial and second
+ClientHello will receive the second ClientHello in STREAM frames with a non-zero
+offset.  The resulting gap in the received stream data needs to be ignored and
+data passed to TLS; if a valid cookie is found, the size of the gap SHOULD be
+validated prior to generating a response.
 
 
 ## NewSessionTicket Address Validation


draft-ietf-quic-tls.md
```diff
@@ -1058,6 +1058,10 @@ validation was not requested originally.  In such cases, the cookie extension
 could either be absent or it could indicate that an address validation token is
 not present.
 
+A server can use the cookie extension to store all state necessary to continue
+the connection.  This allows a server to avoid committing state for clients that
+have unvalidated source addresses.
+
```
ekr: This works, though I think you should be clearer about how you pickle that information (self-encrypted AEAD).

However, if we're in the business of telling people what to do, we actually should document both this technique and the CH...SFIN, NST, RST technique that davidben and victor described.

martinthomson:
Offline, Jānis Čoders pointed out that we aren't very clear about what stream offset you expect on the second ClientHello.  I need to update to include a reminder about that too.

draft-ietf-quic-tls.md
martinthomson: I think that I'll leave the tricks for another PR.  I'm not convinced that the trick will work.  It will always require at least two packets because NST and CONNECTION_CLOSE require packet protection (I suppose you might argue that is more reason to add a new cleartext+encrypted packet type).  Then there is the size of the certificate.  It's more computation, more amplification factor, and all we get is one round trip less delay.

draft-ietf-quic-tls.md
```diff
@@ -1059,6 +1059,27 @@ could either be absent or it could indicate that an address validation token is
 not present.
 
 
+### Stateless Address Validation
+
+A server can use the cookie extension to store all state necessary to continue
```
janaiyengar: add citation for "the cookie extension"

draft-ietf-quic-tls.md
```diff
@@ -1059,6 +1059,27 @@ could either be absent or it could indicate that an address validation token is
...
+information that it requires and include that information in the cookie.  In
+addition to address validation information, a server that uses encryption also
+needs to be able recover the hash of the ClientHello and its length, plus any
+information it needs in order to reconstruct the HelloRetryRequest.
+
+A client proceeds as though the server were stateful.  It constructs a second
+ClientHello and sends it following the initial ClientHello.  Note that this
+means a server that doesn't maintain state between the intial and second
+ClientHello will receive the second ClientHello in STREAM frames with a non-zero
+offset.  The resulting gap in the received stream data needs to be ignored and
```
janaiyengar: I don't think this hack of delivering bytes out of order works... how is a server to know that there aren't lost packets?

Alternatively, a server can send a new connection ID for the client to use for the next leg of the connection. That would be cleaner, and wouldn't need this hack of delivering bytes out of order to TLS.

draft-ietf-quic-tls.md
martinthomson: It's in TLS.

draft-ietf-quic-tls.md
martinthomson: This isn't a hack, it's just physics.  The first ClientHello takes N bytes, the second takes M.  The offset for the first is 0, the second is N because it follows immediately after.  The point here is that the client proceed as if this were the same connection; the server - if it does a stateless reject - has to account for that and deal with the fact that after the reject it won't ever receive the first N bytes - they were already received.

mikkelfj: Why is the stream abstraction important during handshake?

mikkelfj: Here is a proposal that might help simplify handshake without interfering with 0-RTT stream semantics. For 1-rtt it could be required that streams are not permitted before handshake is complete. https://github.com/quicwg/base-drafts/issues/489

RyanAtGoogle: Is this predicated on the idea that a ClientHello must fit into a single packet? If a ClientHello can span multiple packets, then I don't think I understand how a server would distinguish between a packet which contains part 2 of a ClientHello from a packet which contains a post-stateless-reject ClientHello.

martinthomson: We already require that ClientHello fits in a single packet.  Even the possibility of a two-part ClientHello makes DoS mitigation very difficult.

janaiyengar: What's the reason for making it transparent to the client? Why can the client not try with a new connection ID?

martinthomson: The server is the one that cares, so it takes on the extra effort of dealing with the complexity.  Why would the client retry with a new connection ID?

Also, logically, TLS maintains a handshake transcript.  Having the transcript appear to be on the same connection is less surprising than the odd gap you get with a stateless reject.  Besides, it's not a gap to the client, just the server.

janaiyengar: I'm trying to understand how this would work at the server and it's not clear to me. So, when the server receives a packet that basically has one stream frame with stream ID 1 and offset 1050, what should it do? Should it deliver it up to TLS, or should it buffer? Since this retry logic is only known to the TLS part, how is QUIC to determine that this is a retry and not just a reordered packet?

martinthomson: Yes, deliver it to TLS.  When TLS asks whether the address validation is correct, it will provide the address validation token.  That token should include the length of the original ClientHello.  If TLS doesn't provide a token, or the length doesn't match, then it's bad and you can abort at that point.

abraud: What about the packet number on the SHLO ? 
From what i gathered, the one on the retryRequest was randomly generated. If we generate one randomly again then the client will most likely have issues.

abraud: What about the packet number on the SHLO ? 
From what i gathered, the one on the retryRequest was randomly generated. If we generate one randomly again then the client will most likely have issues.

mikkelfj: And what are the rules for dealing with other streams and retransmittable frames such as ping in the initial packet?

RyanAtGoogle: I can understand how this could be made to work, but I worry a bit about the state of the crypto stream being different between the two endpoints. The client believes it has send 2 ClientHellos as bytes [0,N] on stream 0. If the server is required to process pre-handshake stream frames even with non-0 offsets, then it seems like it might be possible for an active attacker to rewrite the second ClientHello to start at, say, offset 2N. This would mean that any post-handshake messages from the client to the server would be ignored as duplicates.

I presume the solution to this is for the server to include in the data it sends to the client (and the client sends back to the server) the byte offset that the follow-up ClientHello is expected to start at, and to validate that when it process the follow-up ClientHello.

Does that sound right?

martinthomson: @abraud is correct.  What makes this hard is that the server might have generated multiple HelloRetryRequest packets, each with their own randomized packet number.  That's a real issue because it means that it can't save that value without changing the HelloRetryRequest.  Then the client will receive two conflicting bits of stream data.  That's bad because we allow endpoints to reject connections if that happens.  Separately, @janaiyengar suggests that we treat a stateless reject as a clean break.  I will see if I can use that; it will need a lot more text :(

@mikkelfj, you can't send anything other than a STREAM frame.

@RyanAtGoogle, you got it right.  The server saves the offset, then pretends that it received those missing bytes once it has validated the token.  If we take @janaiyengar's suggestion though, that's won't be necessary.

janaiyengar: Yeah, this doesn't need QUIC connection continuity, just TLS continuity. A new connection with a new connection ID should achieve that. (The connection doesn't need to go back to the same server, does it?)

martinthomson: The whole point of a stateless reject is to avoid creating state, so if it had to go back to the same server, something is wrong :)

janaiyengar: Good point.

PR #491, opened 2017-05-02
Add a NEW_CONNECTION_ID frame

### LINKED ISSUE #232: New connection ID message
In order to break linkability across events that are known to an endpoint, we should explore the possibility of having each endpoint provide the other with a new connection identifier AND a number of packet numbers that should be skipped.  This allows an endpoint to continue using the same connection on a new path without linking use of that new path to use of the old path.

You need to change connection ID and packet numbers or they can be used for linking.  That doesn't preclude the use of other signals for linking, but it removes obvious ones.

This wouldn't address all of the potential linkability issues we identified, but would help in cases where an endpoint is aware of the need - or even potential need - for a gap in linkability.

martinthomson:
This frame is used by servers to support client unlinkability across network
migrations.

This assumes that the client is aware of the need for unlinkability.  I haven't
gone into a lot of detail on rationale, why this might be needed and the like.
That stuff does need expansion, but we don't have a lot of agreement on the
finer details.  For the moment having the basic mechanism in place will allow
us to make progress.

Closes #232.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1279,8 +1279,8 @@ A client might need to send packets on multiple networks without receiving any
 response from the server.  To ensure that the client is not linkable across each
 of these changes, a new connection ID and packet number gap are needed for each
 network.  To support this, a server sends multiple NEW_CONNECTION_ID messages.
-Each NEW_CONNECTION_ID is marked with a sequence number, which determines the
-order in which they are used.
+Each NEW_CONNECTION_ID is marked with a sequence number.  Connection IDs MUST be
+used in the order in which they are numbered.
 
 Packet number gaps are cumulative.  A client might skip connection IDs, but it
 MUST ensure that it applies the associated packet number gaps in addition to the
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -791,8 +791,9 @@ explained in more detail as they are referenced later in the document.
 | 0x08        | BLOCKED          | {{frame-blocked}}          |
 | 0x09        | STREAM_BLOCKED   | {{frame-stream-blocked}}   |
 | 0x0a        | STREAM_ID_NEEDED | {{frame-stream-id-needed}} |
-| 0xa0 - 0x7f | ACK              | {{frame-ack}}              |
-| 0xc0 - 0xff | STREAM           | {{frame-stream}}           |
+| 0x0b        | NEW_CONNECTION_ID | {{frame-new-connection-id}} |
+| 0xa0 - 0x7f | ACK               | {{frame-ack}}               |
+| 0xc0 - 0xff | STREAM            | {{frame-stream}}            |
 {: #frame-types title="Frame Types"}
 
 # Life of a Connection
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1239,12 +1240,59 @@ error code.
...
+used to link activity across networks.  A server MUST provide a 32-bit value for
+the packet number gap field that is indistinguishable from random to a passive
+observer.
+
+A client might need to send packets on multiple networks without receiving any
+response from the server.  To ensure that the client is not linkable across each
+of these changes, a new connection ID and packet number gap are needed for each
+network.  To support this, a server sends multiple NEW_CONNECTION_ID messages.
+Each NEW_CONNECTION_ID is marked with a sequence number, which determines the
+order in which they are used.
```
janaiyengar: nit: "they are used." -> "they must be used."

draft-ietf-quic-transport.md
```diff
@@ -1854,6 +1902,49 @@ alive when a stream is open. The default is to send a PING frame after 15
...
+connection IDs that can be used to break linkability when migrating connections
+(see {{migration-linkability}}).
+
+The NEW_CONNECTION_ID is as follows:
+
+~~~
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|       Sequence (16)           |
```
janaiyengar: 8 bits ought to be enough - we aren't expecting > 255 migrations

... and this can wrap! I suppose you're trying to support a general scheme that supports up to 1 connection ID per packet. 16 bits would limit the BDP to 32K packets in this case (or 32MB, or a ~2Gbps x 100ms path), which is probably fine.

marten-seemann:
Can we make this symmetric (write peer instead of server / client)?
In a p2p setting, it doesn't really make a difference which peer is which, and both of them might want to do a connection migration at some point.

martinthomson:
@marten-seemann I would love to do that, but we'd want to change the underlying design so that the server sends one connection ID and the client sends another.  That's a valid design, but the current design assumes a single identifier that is selected by the server.

Can you open an issue so that we can track/discuss/reject that idea separately?  I don't think that we can just sneak this in here.

draft-ietf-quic-transport.md
martinthomson: Yes, this was part of the reasoning, the other was that at 255 it's much harder to get the wrapping right.  Of course, if we decided to do that we really would run out of packet numbers.  The fix, make the packet number much bigger, the 64-bit limit is pretty arbitrary anyway and the nonce for most AEADs is longer than that, meaning that we could go to as much as 96-bit without a redesign.

ekr:
I think it's premature to land this; some of the obvious designs for changing conn-id frequently don't require attaching a PN gap. It's also worth noting that you don't need to explicitly carry that; you can generate it using an exported key and the new conn ID.

martinthomson:
It might have been a little premature, but maybe you could phrase your concerns as deltas to the current design.  Removing the packet number field and replacing it with an exported value (is this a valid use for the context?) would be a valid enhancement.

ekr:
yes, I will do that.

PR #496, opened 2017-05-03
Remove the explicit packet number gap in favor of one computed

ekr:
based on a shared secret.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1257,22 +1257,21 @@ way to retry them.
 
 ### Privacy Implications of Connection Migration {#migration-linkability}
 
-Using a stable connection ID on multiple network paths allows a passive observer
-to correlate activity between those paths.  A client that moves between networks
-might not wish to have their activity correlated by any entity other than a
-server.
-
-The NEW_CONNECTION_ID message can be sent by a server to provide an unlinkable
-connection ID for use in case the client wishes to explicitly break linkability
+Using a stable connection ID on multiple network paths allows a
+passive observer to correlate activity between those paths.  A client
+that moves between networks might not wish to have their activity
+correlated by any entity other than a server. The NEW_CONNECTION_ID
+message can be sent by a server to provide an unlinkable connection ID
+for use in case the client wishes to explicitly break linkability
 between two points of network attachment.
 
-The NEW_CONNECTION_ID message includes a connection ID that the client can use
-for packets on the current connection.  This message also includes packet number
-gap, which is the number of packet numbers that the client skips when using the
-new connection ID.  Without this, the packet number field in the header might be
-used to link activity across networks.  A server MUST provide a 32-bit value for
-the packet number gap field that is indistinguishable from random to a passive
-observer.
+A client which wishes to break linkability upon changing networks MUST
+use the NEW_CONNECTION_ID as well as incrementing the packet sequence
+number by an externally unpredictable value computed as described in
+{{packet-number-gap}}. Packet number gaps are cumulative.  A client
+might skip connection IDs, but it MUST ensure that it applies the
+associated packet number gaps in addition to the packet number gap
+associated with the connection ID that it does use.
 
 A client might need to send packets on multiple networks without receiving any
 response from the server.  To ensure that the client is not linkable across each
@@ -1280,10 +1279,6 @@ of these changes, a new connection ID and packet number gap are needed for each
 network.  To support this, a server sends multiple NEW_CONNECTION_ID messages.
 Each NEW_CONNECTION_ID is marked with a sequence number.  Connection IDs MUST be
 used in the order in which they are numbered.
-
-Packet number gaps are cumulative.  A client might skip connection IDs, but it
-MUST ensure that it applies the associated packet number gaps in addition to the
-packet number gap associated with the connection ID that it does use.
 
 A server that receives a packet that is marked with a new connection ID recovers
 the packet number by adding the cumulative packet number gap to its expected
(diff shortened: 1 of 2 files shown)


ekr:
@martinthomson: I note that this change means that these don't need to be delivered to the client where they did before.

Also, I wonder if we should allow multiple conn_ids to be delivered in the same frame, thus saving some bits.

draft-ietf-quic-transport.md
```diff
@@ -1257,22 +1257,21 @@ way to retry them.
...
 
-The NEW_CONNECTION_ID message includes a connection ID that the client can use
-for packets on the current connection.  This message also includes packet number
-gap, which is the number of packet numbers that the client skips when using the
-new connection ID.  Without this, the packet number field in the header might be
-used to link activity across networks.  A server MUST provide a 32-bit value for
-the packet number gap field that is indistinguishable from random to a passive
-observer.
+A client which wishes to break linkability upon changing networks MUST
+use the NEW_CONNECTION_ID as well as incrementing the packet sequence
```
martinthomson: `MUST use the NEW_CONNECTION_ID` implies sending, it's the value from the message that the client uses.

I'm not sure about the MUST-requirement.  Isn't this a simple "can"?

draft-ietf-quic-transport.md
```diff
@@ -1296,6 +1291,22 @@ connection ID, it should expect packets on the new connection ID to start at 18.
 A packet with the new connection ID and a packet number of 17 is discarded as
 being in error.
 
+#### Packet Number Gap
+
+In order to avoid linkage, the packet number gap MUST be externally
+indistinguishable from random. The packet number gap for a connection
+ID with sequence number is computed by encoding the sequence number
+as a 32-bit integer in big-endian format, and then computing:
```
martinthomson: Since packet numbers are 64-bit, might as well make this the same type.  HKDF won't care.

draft-ietf-quic-transport.md
ekr: Well I think my claim is that if you want it unlinkable you have to

draft-ietf-quic-transport.md
ekr: Yeah, agreed.

draft-ietf-quic-transport.md
janaiyengar: The MUST is unclear. I'm not sure how to read it -- is this a requirement of the spec? Seems not, since according to the text, a client may choose to not break linkability. Perhaps using lowercase "must" is what you want here?

draft-ietf-quic-transport.md
```diff
@@ -1296,6 +1291,22 @@ connection ID, it should expect packets on the new connection ID to start at 18.
...
+
+~~~
+Gap = HKDF-Expand-Label(packet_number_secret,
+                        "QUIC packet sequence gap", sequence, 4)
```
janaiyengar: HKDF-Expand-Label is not defined in the transport draft. This should be one more addition to the QUIC-TLS API.

janaiyengar:
Nice -- I like this quite a bit! On compressing, as martinthomson@ points out we haven't settled yet.  I would suggest leaving it as is, and open an issue for (later) looking into > 1 IDs per NEW_CONNECTION_ID frame.

draft-ietf-quic-transport.md
martinthomson: I don't think you want that.  The API we agreed on previously was that TLS provides an exporter and the transport uses it.  If you want this to be in TLS, then we end up with a much tighter coupling than you would like.  (Me, I don't care, I don't see any point in disentangling these two parts, but we did have consensus to separate them.)

ekr: @janaiyengar, is what you want here:

"The TLS-QUIC API must expose a function QUIC-Packet-Gap(sequence) -> Gap which returns a 32-bit integer indicating the packet gap for a given sequence number"?

janaiyengar: Hm -- good point martinthomson@ -- I hadn't thought about that. Is it adequate to do some sort of simple hash here then?

martinthomson: @ekr's design is best practice here.  I would not be that happy about a design that was less robust than this.  HKDF has a bunch of features that we might not need, but the alternatives would require considerably more analysis before we could be sure that they are OK.

janaiyengar: Why wouldn't a simple keyed-hash be adequate? Of course, organization of text shouldn't determine our design; do you have a suggestion for organizing the text if HKDF is to be used?

martinthomson: If we move packet protection to the transport, as planned, then we'll need HKDF, so there's no real issue with taking this change as it is.

ekr: "keyed hash" isn't really a well-formed cryptographic construct. What you want here is a PRF, and these days that typically means using HMAC (which is what HKDF is under the covers). So, if you really prefer, we could inline the HMAC expansion of HKDF-Expand-Label, but I'm not really seeing what the benefit of that is in this instance, given that, as Martin says, if we move packet protection into this document, we'll need HKDF anyway.

Jana, it might help me if it were a bit clearer about what you were optimizing for here. Any implementation of this document is going to need HKDF (and even if you were to swap out TLS, it's quite likely you would end up using HKDF), so it's not really implementation complexity, so is it just that you don't like having a citation to some construct exposed by TLS? If so, that seems like something we can deal with much later in the document cycle if we still feel that way.

janaiyengar: Yeah, both of your responses make sense. Thanks for helping me understand.

PR #498, opened 2017-05-04
Version validation fields can't move or change

### LINKED ISSUE #121: Version downgrade vulnerability
If we do this badly, there might be two versions of QUIC for which the transport parameters are confusable in a way that confounds the validation of version negotiation because the transport parameters are interpreted differently for different QUIC versions.

We need to write that into the security considerations.  We might recommend preserving the initial part of the transport parameters extension, or - better yet - using a new extension codepoint when the format needs to change.

martinthomson:
Closes #121.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1232,6 +1232,13 @@ client MUST terminate with a VERSION_NEGOTIATION_ERROR error code if
 version negotiation occurred but it would have selected a different version
 based on the value of the supported_versions list.
 
+The position and size of the version validation fields in the transport
+parameters MUST NOT change between different versions of QUIC.  If a version of
+QUIC needs to define a new format for transport parameters, it MUST define and
+use a new TLS extension rather than redefine the layout of the existing
+extension.  Confusion about the format of transport pararameters, especially the
+version negotiation fields, could lead to version downgrade attacks.
+
 
 ## Stateless Retries {#stateless-retry}
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1234,12 +1234,12 @@ based on the value of the supported_versions list.
 
 When an endpoint accepts multiple QUIC versions, it can potentially interpret
 transport parameters as they are defined by any of the QUIC versions it
-supports. Since the QUIC protocol relies on transport parameters for
-authenticating the version, the position and the format of the version fields in
-transport parameters MUST be consistent across all QUIC versions supported by
-the client. If a version of QUIC needs to define a new format for transport
-parameters, it MUST define and use a new TLS extension for transport parameters
-rather than redefining the layout of the existing extension.
+supports.  The version field in the QUIC packet header are only authenticated
+using the transport parameters.  The position and the format of the version
+fields in transport parameters MUST either be identical across different QUIC
+versions, or be unambiguously different in a way that ensures that there is no
+possibility for confusion about their interpretation.  One way that a new format
+could be introduced is to define a TLS extension with a different codepoint.
 
 
 ## Stateless Retries {#stateless-retry}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1232,12 +1232,14 @@ client MUST terminate with a VERSION_NEGOTIATION_ERROR error code if
 version negotiation occurred but it would have selected a different version
 based on the value of the supported_versions list.
 
-The position and size of the version validation fields in the transport
-parameters MUST NOT change between different versions of QUIC.  If a version of
-QUIC needs to define a new format for transport parameters, it MUST define and
-use a new TLS extension rather than redefine the layout of the existing
-extension.  Confusion about the format of transport pararameters, especially the
-version negotiation fields, could lead to version downgrade attacks.
+When an endpoint accepts multiple QUIC versions, it can potentially interpret
+transport parameters as they are defined by any of the QUIC versions it
+supports. Since the QUIC protocol relies on transport parameters for
+authenticating the version, the position and the format of the version fields in
+transport parameters MUST be consistent across all QUIC versions supported by
+the client. If a version of QUIC needs to define a new format for transport
+parameters, it MUST define and use a new TLS extension for transport parameters
+rather than redefining the layout of the existing extension.
 
 
 ## Stateless Retries {#stateless-retry}


draft-ietf-quic-transport.md
```diff
@@ -1131,6 +1131,13 @@ client MUST terminate with a QUIC_VERSION_NEGOTIATION_MISMATCH error code if
 version negotiation occurred but it would have selected a different version
 based on the value of the supported_versions list.
 
+The position and size of the version validation fields in the transport
+parameters MUST NOT change between different versions of QUIC.  If a version of
+QUIC needs to define a new format for transport parameters, it MUST define and
+use a new TLS extension rather than redefine the layout of the existing
```
ianswett: This is a bizarre solution to this problem.  Why not say that it must define a version that indicates a change of format and base new formats on that?  Or something of that sort?

At the point we're doing version negotiation, TLS hasn't come into play, so I don't think this works.

martinthomson: @ianswett, I've read this suggestion several times and I'm sorry, but I don't understand what you are trying to say.  Can you explain your alternative design in more detail?

ianswett: I misunderstood what this text was attempting to accomplish, I was thinking of transport parameters as connection ID/packet number/etc.

That being said, I'm not clear on why the transport parameter format can't be tied to QUIC version.  Can you explain more about how this attack works?

mikkelfj: If a QUIC version is defined that does not rely on TLS, how would it define transport parameters?

martinthomson: @ianswett, let's say that we don't change the format of the transport parameters for version 1.  We also have a server that supports version 1 and some hypothetical version 4.

With the current format, a server that selects version 1 will send advertise transport parameters starting with the bytes 0x0008, 0x00000004, 0x00000001 (a length, followed by version 4 and version 1).  After that, it will have some lengths and TLVs for parameters.

However, if version 4 decides to change the transport parameters, then the client that is subject to spoofing of a version negotiation packet might find that they interpret transport parameters differently.  If version 4 is sufficiently similar to version 1 in other ways, you could even end up in a situation where the client and server disagree on what version they are speaking.

For instance, the above value could be interpreted as being valid against a different schema.

ianswett: I believe I now understand what you're saying.  I keep thinking the entire packet is protected, but it's not.

The current wording is still a bit confusing to me, I think Victor had some suggestions on how to make it clearer.

martinthomson: @vasilvv, can you help us out here?

vasilvv: I feel like it would be clearer if you expand on the attack you’re trying to mitigate.  Something like:
“When a peer accepts multiple versions of the handshake, it can potentially interpret the handshake as defined by any of the versions it supports.  Since the QUIC protocol relies on security of the handshake for authenticating the version, the position and the format of the version field in the handshake MUST be consistent across all QUIC versions supported by the client.”

martinthomson:
Thanks to @vasilvv, I have some more text.  @ianswett, can you take another look?

draft-ietf-quic-transport.md
```diff
@@ -1232,6 +1232,15 @@ client MUST terminate with a VERSION_NEGOTIATION_ERROR error code if
 version negotiation occurred but it would have selected a different version
 based on the value of the supported_versions list.
 
+When an endpoint accepts multiple QUIC versions, it can potentially interpret
+transport parameters as they are defined by any of the QUIC versions it
+supports. Since the QUIC protocol relies on transport parameters for
+authenticating the version, the position and the format of the version fields in
+transport parameters MUST be consistent across all QUIC versions supported by
+the client. If a version of QUIC needs to define a new format for transport
+parameters, it MUST define and use a new TLS extension for transport parameters
```
ianswett: One request: The rest of the section on transport parameters doesn't mention TLS.  Can we change this to "a new handshake extension" or something similar to avoid adding new references to it.

MikeBishop:
It took me a couple times through the issue and the PR to figure out the actual attack here, and I don't think the current text actually captures it.  What finally clicked for me was that, although the version number is explicitly given in the packet, an on-path attacker who tried to force a version negotiation mismatch could also undetectably modify the contents of the version field in the packet header.  (Because the integrity protection / encryption on the packets at this point is trivially recalculated by an attacker.)  It's the contents of the handshake that can't be modified without detection, but we have to guard against getting just the right set of transport parameters in the future that looks like the version confirmation one would expect from the current version.

So what we're actually saying is that, if the version validation fields are in a different location in a different version of QUIC, they MUST be unmistakably different from a valid confirmation of having selected the current version or any future version that uses this handshake and transport settings format.

ianswett:
Thank you Mike for elegantly summarizing my previous confusion.

draft-ietf-quic-transport.md
janaiyengar: This isn't defined as a TLS extension in this document, but the format is specified as the layout for in the TransportParameters field in the quic_transport_parameters TLS extension (se line 1076 above.) I think this text probably belongs there, along with the description of the wire layout, instead of here with version negotiation. I think the text is fine, but I would move it up there.

martinthomson: This is still the right place for the text, because it is talking specifically about version downgrade and how those are detected. Ian's comment is easy to address.

draft-ietf-quic-transport.md
```diff
@@ -1232,6 +1232,15 @@ client MUST terminate with a VERSION_NEGOTIATION_ERROR error code if
 version negotiation occurred but it would have selected a different version
 based on the value of the supported_versions list.
 
+When an endpoint accepts multiple QUIC versions, it can potentially interpret
+transport parameters as they are defined by any of the QUIC versions it
+supports.  The version field in the QUIC packet header are only authenticated
```
janaiyengar: nit: are -> is
also, "is only authenticated using" -> "is authenticated using"

draft-ietf-quic-transport.md
```diff
@@ -1232,6 +1232,15 @@ client MUST terminate with a VERSION_NEGOTIATION_ERROR error code if
...
+using the transport parameters.  The position and the format of the version
+fields in transport parameters MUST either be identical across different QUIC
+versions, or be unambiguously different in a way that ensures that there is no
+possibility for confusion about their interpretation.  One way that a new format
```
janaiyengar: suggested rephrase: "be unambiguously different in a way that ensures that there is no possibility for confusion about their interpretation" --> "be unambiguously different to ensure no confusion about their interpretation"

PR #499, opened 2017-05-04
Mention GQUIC versions

### LINKED ISSUE #51: QUIC version number scheme
There have been proposals on the list. We need to finalize a version number scheme for QUIC versions.

martinthomson:
This should help avoid the case where people reuse the numbers that Google are
currently using.

I would like to be able to say "Prior to <some date>" here, but we can add that
if/once Google switches to implementing the IETF protocol.

Closes #51.  Again.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -316,6 +316,9 @@ MAY use one of these version numbers with the expectation that the server will
 initiate version negotiation; a server MAY advertise support for one of these
 versions and can expect that clients ignore the value.
 
+Version numbers with the lower 24 bits clearted were used to identify
+pre-standards versions of QUIC.
+
 \[\[RFC editor: please remove the remainder of this section before
 publication.]]
 


janaiyengar:
That's incorrect. Google QUIC versions are in the "experimental space", and basically occupy the space covered by the ASCII value of the four chars "QXXX", where X is a digit. This is a non-contiguous space, but it squarely falls within experimental, so I don't think we should document it in the spec, but only on the wiki.

I just checked the wiki, and that's incorrect too. I'm fixing that.

janaiyengar:
Fixed the wiki.

MikeBishop:
1) Clearted?

2) I don't think the QXXX space and "lower 24 bits cleared" reconcile.  Isn't it "upper 8 bits are 'Q'"?

janaiyengar:
It's weird. Upper 8-bits are 'Q' is waaay too much space, as I've described in the Wiki.

martinthomson:
Clearly this isn't going anywhere.  If someone wants a change, they are going to have to propose text.

PR #500, opened 2017-05-04
Cover some corner cases in version negotiation

### LINKED ISSUE #497: Packets that shouldn't generate Version Negotiation
There are a bunch of things that might arrive at a server that aren't QUIC.  Not all of these need to generate a Version Negotiation packet in response.

We should insist that short-form packets are ignored.

We might also insist that a packet that isn't obvious an initial handshake packet be dropped.  That means that it needs to be at least 1232/1252 octets long.

These both might happen if routing is bad, if connection migration happens (and the client isn't using connection ID properly) or the server receives internet-chaff.  Better to ignore chaff than to waste effort on it.

martinthomson:
In dealing with #497, I expected that it would be possible to reject initial
packets that are too small.  I also expected that you could reject packets that
weren't of the right type.  However, in order to be able to do that we would
have to create a minimum size requirement for all QUIC versions, and make the
packet types (or at least some of them) version independent also.

That seemed unwise so instead I just described a couple of corner cases.  Short
header packets get thrown away, as do packets that are too small.

Closes #497.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -819,7 +819,10 @@ initial packets sent from the client to the server MUST use the long header
 format and MUST specify the version of the protocol being used.
 
 When the server receives a packet from a client with the long header format, it
-compares the client's version to the versions it supports.
+compares the client's version to the versions it supports.  Packets MUST be
+discarded if they appear to use the short header format, or if they are too
+short to contain a complete packet header and at least one octet of payload.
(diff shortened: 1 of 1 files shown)


janaiyengar:
All we need to say is that a Version Negotiation packet is sent in response to only a Long Form packet. The rest of it is unnecessary.

martinthomson:
I'm not attached to this.

PR #501, opened 2017-05-04
Limit the size of packets

### LINKED ISSUE #383: Do we need to define a maximum packet size?
In the error codes there's a QUIC_PACKET_TOO_LARGE error, but there's no maximum packet size specified anywhere in the document.

Does it make sense to define a maximum packet size or must a QUIC implementation be capable of handling packets with the maximum UDP packet size? Google QUIC currently uses 1452 as the maximum packet size (= Ethernet MTU - minimum UDP header size).

martinthomson:
This is primarily for constrained implementations.  Based on experience with
TLS, there can be real burden for those implementations just implementing an
AEAD correctly over large records.  This setting would allow an absurdly
constrained device some defense against peers who might otherwise send jumbo
packets.  After all, jumbo packets can be more efficient.

No real consensus for this yet, just a proof that it is possible.

Closes #383.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1052,6 +1052,7 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
       initial_max_stream_id(2),
       idle_timeout(3),
       truncate_connection_id(4),
+      max_packet_size(5),
       (65535)
    } TransportParameterId;
 
@@ -1132,6 +1133,14 @@ truncate_connection_id (0x0004):
   the 5-tuple is sufficient to identify a connection.  This parameter is zero
   length.  Omitting the parameter indicates that the endpoint relies on the
   connection ID being present in every packet.
+
+max_packet_size (0x0005):
+
+: The maximum packet size parameter places a limit on the size of packets that
+  the endpoint is willing to receive, encoded as an unsigned 16-bit integer.
+  The default for this parameter is the maximum permitted UDP payload of 65527.
+  Values below 256 are invalid.  The initial packet from a client is exempt from
+  this limit.
 
 
 ### Values of Transport Parameters for 0-RTT {#zerortt-parameters}
@@ -3149,6 +3158,7 @@ The initial contents of this registry are shown in
 | 0x0002 | initial_max_stream_id   | {{transport-parameter-definitions}} |
 | 0x0003 | idle_timeout            | {{transport-parameter-definitions}} |
 | 0x0004 | truncate_connection_id  | {{transport-parameter-definitions}} |
+| 0x0005 | max_packet_size         | {{transport-parameter-definitions}} |
 {: #iana-tp-table title="Initial QUIC Transport Parameters Entries"}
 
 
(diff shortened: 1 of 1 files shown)


janaiyengar:
I'm not convinced there's a need. QUIC is pretty complex; I would argue that severely constrained implementations that can't do AEAD over 1280 bytes are unlikely to do QUIC. 

TLS doesn't have the luxury of saying this, since it has to work for every lightbulb.

martinthomson:
hq is certainly quite complex, but is QUIC really that complex if you pare it down?  It's not that much more complicated than TLS and TCP.  I wouldn't be so presumptive about what can and cannot be done.  Also, @hardie was so kind as to point me at [RFC 2675](https://tools.ietf.org/html/rfc2675), which suggests that 64k isn't the outer limit on packet size.

janaiyengar:
Yes, just TLS + QUIC is more complex than TLS + TCP. Having worked on the SCTP implementation in BSD which had no hq but only transport, I can say that adding multistreaming alone adds a fair bit of complexity. (I'm talking about prod code, not a lab prototype.)

It surely is possible to pare it down to the point that much of the flexibility is lost, to the point that the endpoint can't do AEAD over 1232 bytes (not 1280 bytes.) But why would one care to use QUIC at that point and not just do TCP (or even TCP over UDP)?

martinthomson:
Careful, you stand to risk undermining some of your benefits.  The limit I'm more concerned about here is 64k, not 1232/1252.  The protocol doesn't prevent an endpoint from sending 64k packets (or greater if you accept the possibility that someone implemented RFC 2675, which I understand is unlikely).

janaiyengar:
I don't understand what you mean by "Careful, you stand to risk undermining some of your benefits." -- can you clarify?

Yeah, I guess the only other outcome would be that the endpoint would drop a large packet, which is ... less than awesome. Ok.

martinthomson:
Even with approval, I want to leave this until after the interim before landing, so that we can discuss it in the wider group.  It's not urgent.

mcmanus:
I'm concerned about this because I think the most likely outcome is an
implementation trying to use this to deal with PMTU issues via endpoint
configuration and accidentally creating long term ossification as the
network evolves. Packet size is more of a network property than it is a
resource scaling property (though this issue clearly shows that it is
both!) and configuring it strictly as resource limit might be a mistake.

martinthomson:
Yes, this isn't a stand-in for PMTU.  Using it for that would be quite inappropriate.  I can add an admonition on that point if you think that would help at all.

janaiyengar:
Discussing at interim SGTM.

draft-ietf-quic-transport.md
```diff
@@ -1133,6 +1134,14 @@ truncate_connection_id (0x0004):
   length.  Omitting the parameter indicates that the endpoint relies on the
   connection ID being present in every packet.
 
+max_packet_size (0x0005):
+
+: The maximum packet size parameter places a limit on the size of packets that
+  the endpoint is willing to receive, encoded as an unsigned 16-bit integer.
+  The default for this parameter is the maximum permitted UDP payload of 65527.
+  Values below 256 are invalid.  The initial packet from a client is exempt from
```
ianswett: 1280 not 256?  Also, do we want to mention that the peer is indicating it will drop packets larger than the limit?

martinthomson:
@ianswett, I forgot to commit the change to make this 1252 (I blame jet lag), but you said two comments and only had one.  Just making sure that you didn't have another one.

PR #512, opened 2017-05-09
Force the server and client to remember transport parameters for 0-RTT

martinthomson:
This includes Victor's recommendation to avoid the client violating server
constraints on operation.  This means that the server that accepts 0-RTT is
required to provide new transport parameters that are compatible with its
previously advertised value, or reject 0-RTT.

I made this easier on servers, by allowing them to advertise different
transport parameters that only apply to 0-RTT.  They include these values in
the TLS NewSessionTicket message, alongside the ticket that is used to resume.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1037,33 +1037,29 @@ truncate_connection_id (0x0004):
 
 ### Values of Transport Parameters for 0-RTT {#zerortt-parameters}
 
-Transport parameters from the server SHOULD be remembered by the client for use
-with 0-RTT data.  A client that doesn't remember values from a previous
-connection can instead assume the following values: initial_max_stream_data
-(65535), initial_max_data (65535), initial_max_stream_id (20), idle_timeout
-(600), truncate_connection_id (absent).
+Transport parameters from the server MUST be remembered by the client for use
+with 0-RTT data.  If the TLS NewSessionTicket message includes the
+quic_transport_parameters extension, then those values are used for the server
+values when establishing a new connection using that ticket.  Otherwise, the
+transport parameters that the server advertises during connection establishment
+are used.
 
-If assumed values change as a result of completing the handshake, the client is
-expected to respect the new values.  This introduces some potential problems,
-particularly with respect to transport parameters that establish limits:
+A server can remember the transport parameters that it advertised, or store an
+integrity-protected copy of the values in the ticket and recover the information
+when accepting 0-RTT data.  A server uses the transport parameters in
+determining whether to accept 0-RTT data.
 
-* A client might exceed a newly declared initial value for the connection or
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1037,33 +1037,29 @@ truncate_connection_id (0x0004):
...
+determining whether to accept 0-RTT data.
+
+A server MAY accept 0-RTT and subsequently provide different values for
+transport parameters for use in the new connection.  If 0-RTT data is accepted
+by the server, the server MUST NOT reduce any limits or alter any values that
+might be violated by the client with its 0-RTT data.  In particular, a server
+that accepts 0-RTT data MUST NOT set values for initial_max_data or
+initial_max_stream_data that are smaller than the remembered value of those
+parameters.  Similarly, a server MUST NOT reduce the value of
+initial_max_stream_id.
```
ekr: Why is this required, actually? It seems like you could say "this is what I want next time" without fear.

MikeBishop: There's no problem with a server that allows 100KB after 1 RTT advertising 10KB initial for next time.  The problem is if the client has *remembered* 25KB for the 0-RTT flight, and the server wants to turn it down to 10KB after 1 RTT.  The goal is to say turning numbers down on a given connection is never allowed, only up.

ekr: I got that, but my question is: why?

martinthomson: The problem is that the client might have already violated that limit.  It could send 25KB of 0-RTT data, only to learn that the limit on the data it can send is 10KB.

This rule prevents the server from creating a situation where a limit has been violated.

martinthomson:
Note that this doesn't remove the need for Karatsuba multiplication in Javascript.  That's a little unfortunate, but JS is very much in the minority when it comes to 64-bit integer support.

draft-ietf-quic-transport.md
mikkelfj: If it is not a security issue, why force the client to ensure the server complies with the overhead this involves rather than just require that server to comply and have the connection eventually fail if the server doesn't. There are endless other ways a server might fail to comply that will only eventually result in failure.

martinthomson: Let's take the maximum stream number as an example.  

We have two choices: allow the server to set any value, or force the server to provide a value that is >= the value is knows that the client might use.

Given that the client might have used up to the previously-agreed limit, if we allow a smaller limit in the handshake, that creates an unusual condition on the server.  It sets a limit that it can't enforce uniformly.  It has to use two limits when deciding whether the client is misbehaving.  That's a bunch of special code.

In this design, we chose the simpler option.  The only cost is that if the server wants to both *accept* 0-RTT and reduce limits, then it can't.  That little bit of logic is the only place where we have special code.  It can still do things like reset streams and withhold increases to any limit in order to manage its costs.

PR #518, opened 2017-05-10
Update TLS draft to track recent changes

martinthomson:
This tracks the changes in #493 and so forth to ensure that the TLS doc makes sense.  Up until now it still contained discussion of the VERSION bit, which has been gone for some time.

This sits on top of #493.  Any comments with that I'll deal with over there.


martinthomson:
Bonus change here: use "protected" rather than "encrypted" to match TLS terminology and the terms used elsewhere.

janaiyengar:
I fixed a couple of tiny nits.

martinthomson:
Thanks Jana.

PR #521, opened 2017-05-12
Restored a few error codes and redefined FRAME_ERROR.

janaiyengar:
I'd like to go over this again to ensure we don't leave anything out. In the meanwhile, take a look at the direction I'm pushing in.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2842,45 +2842,83 @@ CONNECTION_CLOSE or RST_STREAM frame. Error codes share a common code space.
 Some error codes apply only to either streams or the entire connection and have
 no defined semantics in the other context.
 
-INTERNAL_ERROR (0x80000001):
+FRAME_ERROR (0x800000TT):
+
+: An endpoint received a frame that was badly formatted.  For instance, an empty
+  STREAM frame that omitted the FIN flag, or an ACK frame that has more
+  acknowledgment ranges than the remainder of the packet could carry. The final
+  octet of the error code carries the Type field in the offending frame.
+
+INTERNAL_ERROR (0x80000100):
 
 : The endpoint encountered an internal error and cannot continue with the
   connection.
 
-ENHANCE_YOUR_CALM (0x80000002):
+CRYPTO_STREAM_CLOSED (0x80000101): 
 
-: The endpoint detected that its peer is exhibiting a behavior that might be
-  generating excessive load.
+: A fin or a RST_STREAM was received for stream 0.
 
-NO_ERROR (0x80000003):
+PAYLOAD_MISSING (0x80000102):
+
+: Received packet contained no payload.
+
+HANDSHAKE_TIMEOUT (0x80000103):
+
+: Connection timed out waiting for the handshake to complete.
+
+NETWORK_IDLE_TIMEOUT (0x80000104):
+
+: Connection timed out due to no network activity.
+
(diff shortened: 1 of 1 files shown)


martinthomson:
It sounds like we're going to be discussing this at the interim.  I think that this misses the point of doing a consolidation.

For instance, why would anyone actually *send* HANDSHAKE_TIMEOUT?  For that matter, how would they send it?

janaiyengar:
Good point -- we can remove HANDSHAKE_TIMEOUT. Unless there's a good reason, I'd like to keep the others. I think the errors are useful, and I don't see an advantage to removing the ones that I've listed.

martinthomson:
That was an example, there are quite a number more that fit the same criterion.

martinthomson:
This might help understand where I'm coming from here.  In going through the existing codes, I asked myself the question "what code would I write to handle a CONNECTION_CLOSE/RST_STREAM that contained this error code?"  If the code is the same for two error conditions, the separate codes were redundant.  The other question to ask might be "why would I send a CONNECTION_CLOSE/RST_STREAM containing this error code?"

janaiyengar:
I think I understand where you're coming from. That doesn't cover monitoring though; Google monitors these error codes, and we find it useful operationally. I expect that other server will find it useful too. Happy to air this out at the interim for others to weigh in on.

martinthomson:
@janaiyengar, I had to rebase the `transport_errors` branch, which means that you probably want to rebase this branch too.  `git fetch origin;git checkout revised_errors;git rebase origin/transport_errors` should do the trick.

janaiyengar:
Thanks for the heads-up; rebased.

draft-ietf-quic-transport.md
```diff
@@ -2612,71 +2612,104 @@ CONNECTION_CLOSE or RST_STREAM frame. Error codes share a common code space.
 Some error codes apply only to either streams or the entire connection and have
 no defined semantics in the other context.
 
-INTERNAL_ERROR (0x80000001):
+FRAME_ERROR (0x800000TT):
+
+: An endpoint received a frame that was badly formatted.  For instance, an empty
```
ianswett: nit: badly -> incorrectly

draft-ietf-quic-transport.md
```diff
@@ -2612,71 +2612,104 @@ CONNECTION_CLOSE or RST_STREAM frame. Error codes share a common code space.
...
+  STREAM frame that omitted the FIN flag, or an ACK frame that has more
+  acknowledgment ranges than the remainder of the packet could carry. The final
```
ianswett: A better ack example would be an ack packet that indicates previously acked packets are now missing.

draft-ietf-quic-transport.md
```diff
@@ -2612,71 +2612,104 @@ CONNECTION_CLOSE or RST_STREAM frame. Error codes share a common code space.
...
-NO_ERROR (0x80000003):
+INVALID_VERSION_NEGOTIATION_PACKET (0x80000106):
+
+: Version negotiation packet is malformed.
+
+PUBLIC_RESET_RECEIVED (0x80000107):
+
+: A public reset packet was received for this connection post-handshake.
+
+TOO_MANY_RTOS (0x80000108):
```
ianswett: This may be useful, but it does assume one implements a too many RTOs feature.

ianswett:
Is this PR obsolete?

PR #524, opened 2017-05-12
Increase resilience to spoofed Version Negotiation packets

### LINKED ISSUE #523: Version Negotiation validation
@janaiyengar observes,

> We haven't nailed down rules for when an endpoint receives a VN with a packet number that does not match any sent packets.

This is true.  We should write that down.

martinthomson:
There's two layers of defense here:

1. Version Negotiation echoes 31-bits of entropy (maybe more if you consider the version to be at all unpredictable, or the client's source address to contain any entropy).

2. If validation fails and version negotiation happened, encourage the client to try again ignoring any Version Negotiation packets that match the failed profile.

Closes #523.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -927,13 +927,20 @@ If the packet contains a version that is acceptable to the server, the server
 proceeds with the handshake ({{handshake}}).  This commits the server to the
 version that the client selected.
 
-When the client receives a Version Negotiation packet from the server, it should
-select an acceptable protocol version.  If the server lists an acceptable
-version, the client selects that version and reattempts to create a connection
-using that version.  Though the contents of a packet might not change in
-response to version negotiation, a client MUST increase the packet number it
-uses on every packet it sends.  Packets MUST continue to use long headers and
-MUST include the new negotiated protocol version.
+A client MUST discard a Version Negotiation packet that does not contain Packet
+Number and Version fields that match those fields in a packet that the client
+previously sent.  This doesn't guarantee that the Version Negotiation packet is
+genuine, but it reduces the chances that the packet is spoofed.
+
+A client that receives a valid Version Negotiation packet selects an acceptable
+protocol version from those listed by the server.  The client then reattempts to
+create a connection using that version.  Though the contents of a packet might
+not change in response to version negotiation, a client MUST choose a new packet
+number it uses on every packet it sends.
+
+If the server does not list an acceptable version, the client MAY ignore the
+Version Negotiation packet.  This might reduce the likelihood that a spoofed
+Version Negotiation packet can be used to disrupt connection establishment.
 
 The client MUST use the long header format and include its selected version on
 all packets until it has 1-RTT keys and it has received a packet from the server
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1228,6 +1235,18 @@ client MUST terminate with a QUIC_VERSION_NEGOTIATION_MISMATCH error code if
...
 
+If the client receives a Version Negotiation packet and these validation checks
+subsequently fail, it is likely that the client received a spoofed Version
+Negotiation packet.  A client MAY attempt to create a new connection and ignore
+any Version Negotiation packets that match those that caused the connection to
+fail.
+
+Note:
+
+: The client cannot rely on the version list from the transport parameters.
```
MikeBishop: I'd rephrase this -- the client certainly *can* rely on that list -- once the handshake has completed.

draft-ietf-quic-transport.md
```diff
@@ -931,13 +931,20 @@ If the packet contains a version that is acceptable to the server, the server
...
+
+A client that receives a valid Version Negotiation packet selects an acceptable
+protocol version from those listed by the server.  The client then reattempts to
+create a connection using that version.  Though the contents of a packet might
+not change in response to version negotiation, a client MUST choose a new packet
+number it uses on every packet it sends.
+
+If the server does not list an acceptable version, the client MAY ignore the
+Version Negotiation packet.  This might reduce the likelihood that a spoofed
+Version Negotiation packet can be used to disrupt connection establishment.
```
MikeBishop: This seems very painful if there is a real mismatch.  We've already blocked out injected packets, and I suspect that legitimate failures will be more common.

janaiyengar: +1. I would expect the common case to be mismatch, since an attacker can easily generate a VN packet that supports a large number of known versions (minus the client's chosen).

draft-ietf-quic-transport.md
janaiyengar: The client has to rely on the list; it just cannot validate the list until the end of the handshake.

draft-ietf-quic-transport.md
```diff
@@ -1228,6 +1235,18 @@ client MUST terminate with a QUIC_VERSION_NEGOTIATION_MISMATCH error code if
 version negotiation occurred but it would have selected a different version
 based on the value of the supported_versions list.
 
+If the client receives a Version Negotiation packet and these validation checks
+subsequently fail, it is likely that the client received a spoofed Version
+Negotiation packet.  A client MAY attempt to create a new connection and ignore
```
janaiyengar: I don't think we know this to be likely a result of spoofing, so I wouldn't say it yet.

draft-ietf-quic-transport.md
martinthomson: I'll cut this paragraph.

draft-ietf-quic-transport.md
janaiyengar: I completely agree that something is broken in this case, but the text doesn't say that. It says "it is likely that the client received a spoofed Version Negotiation packet."

martinthomson:
I ended up dropping this for some reason, but I've updated it now.

mikkelfj:
What makes version negotiation special compared to all other first packets from a server?

martinthomson:
OBE: we decided not to worry about spoofed version negotiation too much.  #724 covers the bit we want to keep here.

Issue #527, opened 2017-05-16
type octet language and public reset value


mcmanus:
transport commonly refers to a type octet when referring to long header packet types.. (e.g. 0x88) but the type field is defined as 7 bits and the long header packet type table lists values without the high bit (e.g. 0x08). I think we're better off without the type octet concept.

Also, public reset is defined in the table as having value 9, but in 5.6 as having a type octet of 0x88 which is already used by keyphase 1 protected packets. I think 9/0x89 is the intention. I would open a separate issue, but the merge conflict is inevitable :)

janaiyengar:
"type field", then? 
or we can go with "type septet" :-)

janaiyengar:
I think this change sounds reasonable.

janaiyengar:
Addressed in #528

PR #528, opened 2017-05-16
issue with type octect and public reset value

mcmanus:
for issue #527


janaiyengar:
I've changed slightly to say "uses long headers with a type value of X" since type values are different for the different header types.

PR #531, opened 2017-05-16
Remove use of initial_rtt variable

### LINKED ISSUE #525: Nits in SetLossDetectionAlarm pseudo code
A few things:

* I think the `initial_rtt` variable can be removed, since it is only ever assigned the `kDefaultInitialRtt` default, so it can be replaced by that constant in the pseudo code for `SetLossDetectionAlarm`.

* There are two places where `alarm_duration` is left-shifted in the pseudo code for `SetLossDetectionAlarm`. That operation assumes an integer representation of alarm_duration; suggest to replace that with multiplication.

janaiyengar:
Closes #525

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -408,8 +408,7 @@ below shows how the single timer is set based on the alarm mode.
 The initial flight has no prior RTT sample.  A client SHOULD remember
 the previous RTT it observed when resumption is attempted and use that for an
 initial RTT value.  If no previous RTT is available, the initial RTT defaults
-to 200ms.  Once an RTT measurement is taken, it MUST replace the initial RTT
-value.
+to 200ms.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -412,7 +408,8 @@ below shows how the single timer is set based on the alarm mode.
 The initial flight has no prior RTT sample.  A client SHOULD remember
 the previous RTT it observed when resumption is attempted and use that for an
 initial RTT value.  If no previous RTT is available, the initial RTT defaults
-to 200ms.  Once an RTT measurement is taken, it MUST replace initial_rtt.
+to 200ms.  Once an RTT measurement is taken, it MUST replace the initial RTT
+value.
```
ianswett: Reading this now, I think we can remove this last sentence.

Issue #509, opened 2017-05-08
Proof of receipt of packets in ACK


ekr:
The "skip a packet number" thing as a defense against ack spoofing  is kinda hacky and seems like it is going to make life difficult for us if we want to do other stuff with packet numbers. As a concrete example, if we want to use server-issued PN tokens, then the server doesn't know what you're going to skip as a client. In addition, it's a different mode the sender has to be in.

I suggest a simpler design: every ACK frame contains a 32-bit field containing the XOR of the final 4 bytes of each packet being ACKed. In the current design, these are pseudorandom as long as the content (or basically anything at all about the packetization) is at all unpredictable, because it's either ciphertext or the authentication tag. If we want, we can always add an "entropy" frame that randomizes the packet; it can be relatively short (one byte) because the combinatorics add up very quickly.

martinthomson:
Even the existence of an "entropy" frame would work.  Also, you can randomize MAX_[STREAM_]DATA to get the same effect.

ekr:
Or randomize the frames in the packet.

igorlord:
> Or randomize the order of the frames in the packet.
Let's not do this or rely on this. Since endpoints will usually process frames in order, the sender may wish to put these frames in a particular order (order of priority, for example).

martinthomson:
At the interim, we discussed this design and decided that it was too similar to the one-bit version of this design - the entropy bit - that was part of earlier versions of the protocol.  People were very happy to have the entropy bit removed because it was quite onerous to implement.  Instead, we are now looking at defining an explicit proof of receipt packet.  See https://github.com/quicwg/base-drafts/issues/161#issuecomment-307387364 for some more details on that.

Issue #423, opened 2017-03-29
Multiple ACK frames


ianswett:
Currently the QUIC spec does not discuss sending multiple ack frames in one packet, and the way it's structured implies largest acked must always increase.  There are plans to use a new initial packet number and a new connection ID when migrating, making multiple ack frames valuable, if not strictly necessary, since otherwise you won't be able to continue acking pre-migration packets.

Additionally, this could allow the ack frame to be simpler by making all ack block lengths 1 byte long, instead of the current 1/2/4/6.

martinthomson:
@ianswett, I like the idea that this could be made simpler.  I had thought of sending multiple frames, but not the fact that it could be a way to make ACK frames smaller and less complicated.

mikkelfj:
What about the case where you have a long fat pipe and more acks to send than fits within an ack frame, or even a single MTU? Do you always only send the earliest acks? And when do you decide to no longer send these and progress with larger acks. You might not receive ack acks because both ends stick to old ranges. The now defunct STOP_WAITING set a clear lower bound that appear to be less obvious now.

mikkelfj:
Related: If we ignore timestamp section of ack frames, the ACK Delay field requires a timestamp stored for every packet, or at least makes it difficult to handle without it. This is because we might not be able to sent up to largest received packet. I propose allowing the ACK Delay field to be 1) absent, or 2) set to zero, or 3) or set to 0xffff when delay information is not readily available. The value ACK Delay in this case is limited anyway. It could be required that ACK frames SHOULD provide ack delay when the largest acked is also the largest received.

As to the ignored timestamp section, this section can selectively drop some timestamps and only delilver a representative subset. Further, it will be drained quickly while Ack frames must repeat over packet ranges and retain more information. Therefore the timestamp section is not a good source of information for ACK Delay.

ianswett:
I agree that if we want to be able to ack an entire flight(ie: up to 2^32) packets, we can't move to a single byte format.  Which makes me question the idea of simplifying the ack frame to only have 1 byte ranges.  

STOP_WAITING doesn't help us in that case, because we wouldn't expect to receive a STOP_WAITING for a packet until at least an RTT later, and I believe the concern is having more packets in flight than fit into an ack frame?

[...]

Issue #161, opened 2017-01-17
Address change and consent to send


martinthomson:
When a client changes IP addresses (for any reason, including NAT rebinding), the server has no way of gaining an assurance that the client has actually seen its packets.  That opens an attack where the attacker establishes a connection, then starts spoofing source address.  If the server starts to send to that source address without verifying that the client can receive packets, that's potentially a powerful packet amplification attack.

It's not sufficient to look for ACKs or other such things, because no client message includes proof that the client has actually received a packet from the server.  A client can just generate ACK and WINDOW_UPDATE frames as necessary, even new requests.  The only risk a client runs is that it overestimates the number of packets that the server has sent and sends too many spurious ACK frames, which the server will then recognize and react to.

marten-seemann:
This would be the equivalent of an Optimistic ACK attack, after completing the crypto handshake. If the server randomly skips packet numbers, the server will notice the attack **eventually** and close the connection.
The question is, will that happen soon enough to make this attack unpractical. In quic-go, we randomly skip one packet out of 500, so just ACKing random packets would make the server send about 800 kB of data before noticing the attack.

martinthomson:
Packet skipping would seem to be a possible answer, though the client only needs to ack a few packets to keep things moving along.  If it is eventually unlucky enough to hit a skipped packet the attack ends.  I'd need to do the math, but 800k sounds right for an expected value, which is pretty high when you compare it to the cost of a handshake (around 1400 octets and some crypto for clients).

mirjak:
You could maybe send a ping frame with a random number if you see a new IP address...?

ianswett:
In the case of port or IP address change, skipping a larger number of packets makes it much harder to guess.

martinthomson:
FWIW, PING doesn't work here.  The response to a PING is an ACK, which contains no information from the PING.

mirjak:
Yes, I meant to actually extend the ping mechanism here to make it possible to reflect a random number chosen by the ping sender. Having this as an explicit check rather than doing some implicit guess-work might make sense given it's simple and the cost is low.

ianswett:
The idea of a PING packet with an random number and a PONG in response seems reasonable, though if this is the only use case, I think skipping packet numbers is sufficient, given the 5-tuple is changing anyway.

martinthomson:
[...]

PR #565, opened 2017-06-01
Loosen the MUST requirement to close if an unsent packet is acknowledged

### LINKED ISSUE #564: loosen MUST requirement around close-on-ack-of-unsent
>  sender MUST close the connection if an unsent packet number is acknowledged

I suggest s/MUST/SHOULD .. doing this as MUST requires keeping the state of every unused packet number for the entire duration of the connection to separate the ack of concern from a duplicated ack.

martinthomson:
Closes #564

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1590,7 +1590,7 @@ frames. A receiver MAY send timestamps for non-retransmittable packets.
 A receiver MUST not send timestamps in unprotected packets.
 
 A sender MAY intentionally skip packet numbers to introduce entropy into the
-connection, to avoid opportunistic acknowledgement attacks.  The sender MUST
+connection, to avoid opportunistic acknowledgement attacks.  The sender SHOULD
 close the connection if an unsent packet number is acknowledged.  The format of
 the ACK frame is efficient at expressing blocks of missing packets; skipping
 packet numbers between 1 and 255 effectively provides up to 8 bits of efficient


draft-ietf-quic-transport.md
```diff
@@ -1590,7 +1590,7 @@ frames. A receiver MAY send timestamps for non-retransmittable packets.
 A receiver MUST not send timestamps in unprotected packets.
 
 A sender MAY intentionally skip packet numbers to introduce entropy into the
-connection, to avoid opportunistic acknowledgement attacks.  The sender MUST
+connection, to avoid opportunistic acknowledgement attacks.  The sender SHOULD
```
ianswett: It could also say "The sender MUST close the connection if a packet it knows is unsent is acknowledged."  Which keeps the must, but does not require unlimited state.

martinthomson: @mcmanus, @siyengar, does Ian's suggestion work better?

ekr:
Before we spend a lot of time bikeshedding this, maybe we should consider whether we want to adopt #509

martinthomson:
The same general question applies even if we adopt #509.  Do we require that the sender track what it has sent in order to be able to correctly validate an ACK?

mikkelfj:
More generally, I believe an earlier draft had a concept of the oldest known package number which got replaced by ACK'ing acks periodically.

I think it would be good if the protocol maintains a concept of what is the oldest packet number still of interest to the sender such that the receiver can promptly drop state older than this value for dupe rejection etc., and also such that received acks pertaining packets older than this value can promptly be dropped (or at least the older part of the ACK frame).

Any packet number newer than this lower bound could then be required to be remembered and invalid acks SHOULD/MUST be rejected.

An adversarial sender could fail to update the lower bound forcing more state at receiver. The current behavior is safe in the sense that there is no lower bound other than heuristics anyway, but if a lower bound was placed, this issue must be considered.

ianswett:
@mikkelfj I don't think the least unacked value that was in the stop waiting helps here, except as a convenient lower bound on the range the MUST applies to.  Even without the frame, there must be a lower bound on the range of tracked packets sender side.

For the record, I'm fine with my suggested wording or Martin's PR, I just was trying to provide an option if we want to keep the MUST.

mikkelfj:
An explicit lower bound might significantly reduce the state the receiver needs to maintain and perhaps the frequency of both acks and ack acks - at least that is my intuition. But it is correct that some bound must be maintained regardless.

mikkelfj:
But that discussion is perhaps more relevant to https://github.com/quicwg/base-drafts/issues/423

draft-ietf-quic-transport.md
mcmanus: "it knows" isn't really verifiable in any normative sense - it pretty much means SHOULD. prefer SHOULD.

ekr:
Not exactly. You won't be able to validate the ACK at all without keeping exact track, so either you are implementing the defense or you're not, but the implications are clear either way. In any case, you won't need it for basic ACK functionality

mcmanus:
if you get a duplicated ack from long ago, how do you know if it is (dup) acking a packet you sent or not? (assuming you skipped some numbers non deterministically)

mikkelfj:
> if you get a duplicated ack from long ago, how do you know if it is (dup) acking a packet you sent or not? (assuming you skipped some numbers non deterministically)

The sender maintains an oldest relevant packet number regardless of whether it is transmitted or not (it currently isn't). If the ack refers to an older packet it is ignored - even if it could be by malicious intent, it cannot really do any harm anymore. If the ack refers to a more recent packet number, it could be used to confuse the bandwidth estimation algorithm and causing the sender to DoS the receiver, which harms the ack sender, but also any other connection to the server.

draft-ietf-quic-transport.md
siyengar: Generally I worry with a SHOULD no-one will end up implementing it because people will realize that it requires maintaining a lot of state. I like the spirit of Ian's wording but it's super confusing. SHOULD seems like the right thing here. I think if we acknowledge why it is a SHOULD it might convince more people to implement it along with a suggestion of how to implement it. 

For example a sender could implement this by keeping track of all the packets sent and by advancing a smallest ackable packet periodically.

mikkelfj: The wording could also be "SHOULD close a connection when receiving an ACK for a packet that has not been sent unless the packet is older that oldest retransmittable packet in which case the ackowledgement of that packet should be ignored."

siyengar:
I was thinking that when you advance the ACK lower bound you would call getting an ACK for a packet below that bound would be an error. There are several triggers that could be used to advance the lower bound of ACKs. 

* If you keep track of the ACK of the ACK of a packet sequence number you could maintain a variable called
mininum_acceptable_packet = smallest packet that we're keeping state for, for which we have received the ACK of the ACK for the receipt of the packet. This puts a lower bound on the state while tolerating 1 RTT of reordering

* You could just use RTTs, i.e after getting an ACK for a packet and then a few RTTs have elapsed just advance the minimum_acceptable_packet

a reasonable receiver would not send ACks in such a delayed manner which could be assumed.

mikkelfj:
> I was thinking that when you advance the ACK lower bound you would call getting an ACK for a packet below that bound would be an error. 

I believe @ianswett mentioned that GQUIC through live testing found it better to acknowledge latest packets over earliest acknowledge packets at the possible cost of retransmission. That is, when you have to make a choice due to ACK frame size constraints. But, it is not clear exactly how to handle ACKS that do not cover the full range of received packets, as discussed in  #423
The sender also cannot trust the ACK

Rant: I am personally not too happy with the current ACK concept that seems open to many interpretations and risk holding a lot of excessive state both for dupe detection and retransmission of ACK frames. I would much rather have a concept such as STOP_WAITING, or a similar simple range of fully handled packet number ranges without needing the details of which ACK frames have been sent or in need of transmission. Other retransmissions can be stored by reference to external data (mmap etc.).

But attack vectors must be also be considered:

> a reasonable receiver would not send ACks in such a delayed manner which could be assumed.

A major part of the excercise is dealing with unreasonable receivers.  Also, for low powered devices or for avoiding unnecessary network activity that triggers context switches every time a packet is received, slowing down busy systems, ACKs could quite well be rather delayed. Only, this generates more state that must be kept. As long as this state are only packet ranges acked or not, this is not a big problem.

draft-ietf-quic-transport.md
MikeBishop: But the retransmission queue may be sparse -- you don't necessarily know whether a packet no longer there was never sent or was simply already acknowledged.

SHOULD is supposed to have an explanation of why you might not.  You SHOULD do this if you detect it, but you're not required to maintain enough state to detect it 100% of the time.

mikkelfj: @MikeBishop I think I agree but isn't 

> But the retransmission queue may be sparse -- you don't necessarily know whether a packet no longer there was never sent or was simply already acknowledged.

a good reason for SHOULD?

martinthomson:
Great discussion everyone.  I don't think that I can recognize this as a bike shed given the many layers of paint it now sits under.

We discussed this at the interim and concluded that we would remove the ability to have gaps (see https://github.com/quicwg/base-drafts/issues/161#issuecomment-307387364), but this uncertainty still exists to some extent due to the NEW_CONNECTION_ID frame and the use of packet number skipping in that context.  This will remove one source of gaps in the packet number sequence and make detecting ACKs for unsent packets easier, but not guaranteed.

Thus, I think that we have a SHOULD here and we can point people who ask at this wonderful thread if they want to understand all the gory details.

PR #589, opened 2017-06-06
Move connection ID change to only Server Cleartext

### LINKED ISSUE #588: Server chosen connection ID design
Summarizing discussion in Paris:
- Client Initial, 0-RTT carries Client-chosen Connection ID (CCID)
- Stateless Retry / Version Negotiation carry CCID, no SCID recommendations here
- Server Cleartext carries Server-chosen Connection ID (SCID)
- Client Cleartext, 1-RTT carry SCID

martinthomson:
Closes #588

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -727,7 +727,7 @@ as load balancers, to locate and use it.
 The client MUST choose a random connection ID and use it in Client Initial
 packets ({{packet-client-initial}}) and 0-RTT packets ({{packet-protected}}).
 If the client has received any packet from the server, it uses the connection ID
-it received from the server.
+it received from the server for all packets other than 0-RTT packets.
 
 When the server receives a Client Initial packet and decides to proceed with the
 handshake, it chooses a new value for the connection ID and sends that in a
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -592,8 +592,8 @@ It carries cryptographic handshake messages and acknowledgments.  It is used
 by a server that wishes to perform a stateless retry (see
 {{stateless-retry}}).
 
-The packet number and connection ID fields echo the packet number of the
-triggering client packet.  This allows a client to verify that the server
+The packet number and connection ID fields echo the corresponding fields from
+the triggering client packet.  This allows a client to verify that the server
 received its packet.
 
 After receiving a Server Stateless Retry packet, the client uses a new Client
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -595,11 +592,9 @@ It carries cryptographic handshake messages and acknowledgments.  It is used
 by a server that wishes to perform a stateless retry (see
 {{stateless-retry}}).
 
-The connection ID field in a Server Stateless Retry packet contains a server
-selected value, see {{connection-id}}.
-
-The packet number field echoes the packet number of the triggering client
-packet.  This allows a client to verify that the server received its packet.
+The packet number and connection ID fields echo the packet number of the
```
MikeBishop: This says the connection ID field contain's the original packet's packet number.  Don't think that's correct.

martinthomson: I believe that it is correct.

igorlord:
So if I have 50% of the backend servers upgraded to a newer QUIC version, how do I make sure that client comes back to the same server after Version Negotiation?

draft-ietf-quic-transport.md
MikeBishop: No, the packet number field should contain the client's packet number, and the connection ID field should contain the client's connection ID.  Otherwise, this is just... weird.

chocis:
Why such design changes? I want to selectively decide which server will load balance asymmetric cryptography (heavy processing) of TLS handshake. With this design its not possible.

martinthomson:
@chocis, That was discussed here at the interim as being a primary advantage of the current design.  No one spoke in favour of retaining that.

draft-ietf-quic-transport.md
```diff
@@ -730,20 +725,18 @@ location in all packet headers, making it straightforward for middleboxes, such
...
-({{packet-server-stateless}}), and the first Server Cleartext packet
-({{packet-server-cleartext}}).  The server MAY choose to use the value that the
-client initially selects.
-
-A server MUST NOT propose a different connection ID in response to a Client
-Cleartext packet ({{packet-client-cleartext}}).  A Client Cleartext packet is
-only sent after the server has committed to maintaining connection state.
+packets ({{packet-client-initial}}) and 0-RTT packets ({{packet-protected}}).
+If the client has received any packet from the server, it uses the connection ID
+it received from the server.
```
janaiyengar: This sentence is slightly misleading, since 0-RTT packets will continue using the client Connection ID. I would remove this sentence since paras below cover everything else.

MikeBishop:
@huitema also spoke *against* that, since it becomes easy for an attacker to forge.

draft-ietf-quic-transport.md
martinthomson: "... for all packets other than 0-RTT packets." ?

chocis:
Well then I would like to propose to reconsider that. Is there any reason why Server Stateless Retry couldn't change CID?

1) Firstly, in my opininon server without using Server Stateless Retry (without verifying src address/port) is simply asking to be DOS attacked. I could fill server's RAM with Raspberry Pi by simply spoofing src address. Thats why I am talking about situation where that check is enabled.

2) Secondly, imageine that you want to use CID for load balancing your clients. In current solution you would have to support accepting TLS handshake on all backends and when initial request is validated the partial TLS and QUIC state must be serialized and somehow transported to decided loadbalanced server (with the new CID) - and thats not easy AT ALL. I think, that this can be solved by simply returning new CID in Server Stateless Retry without doing any work or state transportation on backend.

igorlord:
The way I would implement the load balancer:

1. If the packet has CCID, consistent hash it into the pool of backend servers for this dest IP or hostname based on the 6-tuple (5+tuple + CCID) and forward the packet.
1a. The backend server replies with a Version Negotiation packet (too bad cannot use SCID here). or
1b. The backend server replies with a SCID to continue the handshake.

2. If the packet has SCID, validate that SCID seems ok and forward the packet to the identified server.

Issue #554, opened 2017-05-23
Reconsider the use of FNV-1a for "cleartext" integrity


ekr:
We had consensus to use FNV-1a to provide integrity for packets in non-adversarial settings. However as observed in #522, we still have to potentially worry about packet injection from off-path attackers. A natural design is to have the pre-handshake integrity check be somehow keyed by information in the ClientHello (e.g., the conn id). In this case, we would probably want to use something more like HMAC, though it's not necessarily required.

martinthomson:
Is this just a more complicated way of providing proof of receipt?

ekr:
I don't think so.... It's rather a way of avoiding injection of bogus handshake packets by off-path attackers.

ianswett:
While we're reconsidering this, @martinduke and I were talking again about the idea of encrypting using AES-GCM with a version specific key.  That re-convinced me it was a good idea.

1) AES-GCM hardware offload is a common feature, and being able to use it as much as possible is nice.
2) Forces middleboxes that want to do TLS inspection to stay up to date with the latest versions, otherwise they won't be able in inspect SNI.  This makes future changes to TLS less likely to ossify, since middleboxes that aren't frequently updated won't be able to see the new version of TLS.

Given recent deployment issues with TLS 1.3, I think point 2 is more valuable than I previously believed.

mikkelfj:
While not against the idea, 

>It would greatly simplify the permutations if packets were always
encrypted, and it was just a matter of switching from an insecure key to a
secure one.

does not hold when encryption is not necssarily AES-GCM after handshake or in a new version. Further, some low-end IoT devices do not have AES accelleration, and will not have, and may want to use ChaCha20 instead. Such devices could use a generic implementation since it is only for handshake, but it does kill the simplicity argument.

I have previously in a private mail to some authors suggested signing the cleartext headers uses a CMAC algorithm with a version specific key. This is much simpler than AES-GCM and better suited for non-version specific implementation. CMAC will not encrypt, but AES can still be used for that, if the concern is ossification.

Regarding ossification - isn't there a risk that middleboxes will only work if they understand the protocol, and thus will reject newer QUIC versions with a key they don't have? This could make the problem worse. I'm just speculating as I have no experience with these issues.

mikkelfj:
For reference here is the CMAC proposal from March 15 I sent:

> The approach discussed below can be extended to unauthenticated headers:
> 
[...]

PR #593, opened 2017-06-06
Require the client to detect partly bogus version negotiation packets

ekr:
in which the server sends back the client's version, as servers
aren't supposed to do this.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -933,7 +933,10 @@ proceeds with the handshake ({{handshake}}).  This commits the server to the
 version that the client selected.
 
 When the client receives a Version Negotiation packet from the server, it should
-select an acceptable protocol version.  If the server lists an acceptable
+select an acceptable protocol version. If the packet contains the version that
+the client initially offered, the client MUST terminate the connection
+using a QUIC_INVALID_VERSION_NEGOTIATION_PACKET error.
+If the server lists an acceptable
 version, the client selects that version and reattempts to create a connection
 using that version.  Though the contents of a packet might not change in
 response to version negotiation, a client MUST increase the packet number it


draft-ietf-quic-transport.md
```diff
@@ -933,7 +933,10 @@ proceeds with the handshake ({{handshake}}).  This commits the server to the
 version that the client selected.
 
 When the client receives a Version Negotiation packet from the server, it should
-select an acceptable protocol version.  If the server lists an acceptable
+select an acceptable protocol version. If the packet contains the version that
+the client initially offered, the client MUST terminate the connection
+using a QUIC_INVALID_VERSION_NEGOTIATION_PACKET error.
```
martinthomson: It's not currently possible to send that error code.  (See #467.)

It might be better to just consider QUIC to have failed in this case.

MikeBishop:
I'm not clear why this is an improvement.  We currently have:

> A client MUST ignore a Version Negotiation packet that lists the client’s chosen version.

If we define a packet which causes us to just bail, it becomes very easy to disrupt handshakes by packet injection.  Whereas ignoring the potentially-malicious packet and waiting for a timeout gives a chance to succeed.

ekr:
I had missed that line.

With that said, I think we want a general defense against packet injection
(cf. https://github.com/quicwg/base-drafts/issues/554)

huitema:
Agree with Mike. It is better to ignore obviously bogus packets than to react to them.

MikeBishop:
Is this PR still active?  It doesn't appear to have a correlated issue, and the comments have all argued against merging.  @ekr, if you still believe this change needs to happen, can you open an issue and articulate why?  If not, let's close this.

janaiyengar:
I'm closing this PR. ekr@: If you think this is necessary, please open an issue where we can discuss this.

PR #600, opened 2017-06-07
Split QUIC core from QUIC v1

MikeBishop:
First pass at the proposed idea to pull the things that are version-invariant into a separate doc.  Creates a "core" document which describes packet headers, version negotiation, and stateless reset.  Leaves to individual versions to define additional packet types and their fields.  Transport doc does that, along with all frame definitions.

I'll note that it feels a little odd to have "core QUIC" not be multiplexed and have no concept of a crypto handshake.  We've previously said that version defines what you expect to find on Stream 0, but by the split mentioned at the interim, the very existence of Stream 0 and its use for the handshake is itself a version-local concept.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -100,10 +100,9 @@ informative:
 --- abstract
 
 This document defines the initial version of the QUIC transport protocol.
-
-Building on the framework in {{QUIC-CORE}}, this
-document describes packet format, multiplexing and reliability. Accompanying
-documents describe the cryptographic handshake and loss detection.
+Building on the framework in {{QUIC-CORE}}, this document describes packet
+format, multiplexing and reliability. Accompanying documents describe the
+cryptographic handshake and loss detection.
 
 
 --- note_Note_to_Readers
@@ -358,7 +357,7 @@ the following sections.
 ## Short Header
 
 The short header can be used after the version and 1-RTT keys are negotiated.
-The format of the long-form header is described in {{QUIC-CORE}}.
+The format of the short-form header is described in {{QUIC-CORE}}.
 
 In this version, the packet type in a short header currently determines only the
 size of the packet number field.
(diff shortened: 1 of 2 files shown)


mikkelfj:
> I'll note that it feels a little odd to have "core QUIC" not be multiplexed and have no concept of a crypto handshake.

Was wondering the same with the current draft text: there needs to be some requirements a custom crypto handshake must comply to. Such as as error handling, verifying the size of Client Initial Packet, how and when to accept clear text packets and how to transition to encryption. QUIC-TLS and TLS-1.3 covers this, but some information should probably be destilled in a general purpose handshake discussion.

ekr:
Hmm... I think I would trim this down to QUIC invariants which *just* defines what has to be in the initial packet and the VN packet.

MikeBishop:
That's basically what I was going for.  But the contents of the initial packet are already version-dependent, and would lock in a lot of details that we may want to vary in future versions.  So we need to leave the packet payload in the hand of the specific version.

What I attempted to do was just take the list of things that can't change between versions and lift those into the separate doc.

draft-ietf-quic-transport.md
```diff
@@ -335,169 +325,43 @@ are using for private experimentation on the
...
-{: #fig-short-header title="Short Header Format"}
-
 The short header can be used after the version and 1-RTT keys are negotiated.
-This header form has the following fields:
-
-Header Form:
-
-: The most significant bit (0x80) of the first octet of a packet is the header
-  form.  This bit is set to 0 for the short header.
+The format of the long-form header is described in {{QUIC-CORE}}.
```
LPardue: nit: short-form header

MikeBishop:
This is seriously bit-rotten and will need to be re-done if we agree to do this.  Closing the old pull request, though I might inherit some text from it later.

janaiyengar:
Per discussion at editors' meeting, the issue is parked for now, so closing this PR since it's already bit-rotten

PR #609, opened 2017-06-08
Latency Spin Bit

huitema:
Adding text to the transport spec to describe the "latency spin bit".

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -198,7 +198,7 @@ strengths of QUIC include:
 
 * Version negotiation
 
-* Support for monitoring 
+* Support for monitoring
 
 
 ## Low-Latency Connection Establishment
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -863,6 +863,10 @@ the network path. This bit is set as follow:
 Observation points can estimate the network latency by monitoring the
 changes in the latency spin bit.
 
+The latency spin bit handling is similar to the
+"Alternate Marking method for passive performance monitoring"
+described in [@?I-D.draft-ietf-ippm-alt-mark].
+
 # Frames and Frame Types {#frames}
 
 The payload of cleartext packets and the plaintext after decryption of protected
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -304,7 +304,7 @@ flow of encrypted packets, but the usual tools for monitoring
 latency require observing packet numbers and acknowledgements, which
 in QUIC are encrypted.
 QUIC packets enable monitoring of latency by including a clear text
-latency spin bit that is reflected between peers.
+latency spin bit that is reflected between client and server.
 
 # Versions {#versions}
 
@@ -853,15 +853,19 @@ an unsupported version.
 The latency spin bit enables latency monitoring from observation points on
 the network path. This bit is set as follow:
 
-* The connection responder sets the spin bit value to the value of the
-  spin bit in the last packet received from the connection initiator.
+* The server sets the spin bit value to the value of the
+  spin bit in the last packet received from the client.
 
-* The connection initiator sets the spin bit value to the opposite
-  of the last value received from the connection responder, or to 0
+* The client sets the spin bit value to the opposite
+  of the last value received from the server, or to 0
   if no packet as been received yet.
 
-Observation points can estimate the network latency by monitoring the
+Observation points can estimate the network latency by monitoring these
 changes in the latency spin bit.
+If packets are delivered in sequence, this procedure will cause the spin
+bit to change value in each direction once per round trip. Out of
+sequence deliveries can cause some spurious transitions, which
+monitoring agents may have to filter.
 
 The latency spin bit handling is similar to the
 "Alternate Marking method for passive performance monitoring"
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -198,6 +198,8 @@ strengths of QUIC include:
 
 * Version negotiation
 
+* Support for monitoring 
+
 
 ## Low-Latency Connection Establishment
 
@@ -294,6 +296,15 @@ QUIC version negotiation allows for multiple versions of the protocol to be
 deployed and used concurrently. Version negotiation is described in
 {{version-negotiation}}.
 
+## Support for Monitoring
+
+Network administrators often find difficult to monitor the behavior
+of encrypted protocols. Bandwidth can be monitored by observing the
+flow of encrypted packets, but the usal tools for monitoring
+latency require observing packet numbers and acknowledgements, which
+in QUIC are encrypted.
+QUIC packets include a clear text latency spin bit that enables
+monitoring of latency by observing packets in transit.
 
 # Versions {#versions}
 
@@ -355,7 +366,7 @@ version negotiation and 1-RTT keys are established.
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+
-|1|   Type (7)  |
+|1|S|  Type (6) |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                                                               |
 +                       Connection ID (64)                      +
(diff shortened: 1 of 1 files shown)


britram:
I like this approach, but would like to dig into how it breaks in corner cases.

It seems to me to be quite related to "coloring" based approaches, see e.g. https://datatracker.ietf.org/doc/draft-ietf-ippm-alt-mark/

hardie:
I also like this approach, and I think the one key corner case (out-of-order packet arrival) can be handled by description of how to smooth the base RTT signal.  That would allow out of color packets to be seen as out-of-order indications rather than screwing up the RTT calculation.

draft-ietf-quic-transport.md
```diff
@@ -198,6 +198,8 @@ strengths of QUIC include:
 
 * Version negotiation
 
+* Support for monitoring
```
martinthomson: I would rather that we not include this section.

britram: in any case "support for monitoring" is much broader than what this is.

mikkelfj:
I would suggest to make it optional. There are some privacy issues associated with this allowing to assocate packets with a connection. Though UDP also associates packets, alternate transport layers can obfuscate routing.

mirjak:
Instead of using a 'whole' bit, you can also have another packet type for protected packets. And if you define a well know patter that is used, you can even measure loss with it.

draft-ietf-quic-transport.md
huitema: If you completely remove the question, then the bit description just appears out of nowhere. But then, yes, we could be specific about latency monitoring, rather than monitoring in general.

mirjak:
Or just have the flag on short packets...

draft-ietf-quic-transport.md
```diff
@@ -294,6 +296,15 @@ QUIC version negotiation allows for multiple versions of the protocol to be
 deployed and used concurrently. Version negotiation is described in
 {{version-negotiation}}.
 
+## Support for Latency Monitoring
+
+Network administrators often find difficult to monitor the behavior
```
ianswett: nit: how about "Network administrators have difficulty monitoring ..."?

draft-ietf-quic-transport.md
```diff
@@ -294,6 +296,15 @@ QUIC version negotiation allows for multiple versions of the protocol to be
...
+of encrypted protocols. Bandwidth can be monitored by observing the
+flow of encrypted packets, but the usal tools for monitoring
```
ianswett: usal/usual

draft-ietf-quic-transport.md
```diff
@@ -294,6 +296,15 @@ QUIC version negotiation allows for multiple versions of the protocol to be
...
+latency require observing packet numbers and acknowledgements, which
+in QUIC are encrypted.
+QUIC packets include a clear text latency spin bit that enables
```
ianswett: nit: "QUIC packets enable monitoring of latency by including a clear text latency spin bit that is reflected between peers."

huitema:
I get Martin's point about per path signal, which it indeed is. But then, we don't really have an end-to-end vs path distinction yet. In the current transport draft, there seems to be just one path at a time. I think the distinction will only become apparent when we start specifying QUIC multipath.

draft-ietf-quic-transport.md
```diff
@@ -827,6 +848,24 @@ Implementations MUST assume that an unsupported version uses an unknown packet
 format. All other fields MUST be ignored when processing a packet that contains
 an unsupported version.
 
+## Latency Spin Bit
+
+The latency spin bit enables latency monitoring from observation points on
+the network path. This bit is set as follow:
+
+* The connection responder sets the spin bit value to the value of the
+  spin bit in the last packet received from the connection initiator.
```
ianswett: What if two packets were received and the first had the spin bit set and the second did not?  I would think the spin bit would still be set by the responder in that case?

huitema:
Mirja, I hesitated between the "packet type" design and the "spin bit". I like having Red/Blue packet types, but I am concerned with the interaction with issue 311, "GREASE the packet type octet". If we do grease the packet type, we will need to leave the spin bit in clear text. This is much easier to do with a reserved bit than with the packet type.

mirjak:
Yes, there is dependency on greasing. We could also just grease the long header because the short header has a bunch of flags that can anyway change all the time. I even thought about having only flags with the short header (and no type anymore) because that's nearly what we have already and that makes sense to me (and also looks slightly clearer to me) because all control packets are long headers.  For me the point of the short header is basically a packet where you want to save some of the header bits and use it for payload instead and the flags simply define which part of the header bits you omit.

mikkelfj:
There is also the option to grease the type but reserve a single ungreased value for monitoring packets.

draft-ietf-quic-transport.md
ianswett: Also, can you add clarifying text about what to do when exiting quiescence?

martinthomson:
> There is also the option to grease the type but reserve a single ungreased value for monitoring packets.

This is unaffected by my favourite form of greasing (xor with the last octet of the unencrypted frame).  We have a choice as to whether or not this is greased, of course.

martinthomson:
Thinking about this some more, this bit will flip lots in ways that aren't immediately obvious.  I think that we can safely not grease this.  (We could also grease it, but that makes it a teensy bit less accessible to the devices who we intend to consume it, so not greasing is probably best.)  Either way, that's not a question we need to resolve with the change as it stands.

I still strongly think that we should avoid adding this to the list of transport features, and I think that you should address @ianswett's comments about dropping an entire flight (the answer being just "oh well" here).

martinthomson:
Oh, and I really think that this needs to be crisp about being per-path.  I think that we need to explicitly sasy that we reset the signal when the IP or port at either end changes.

draft-ietf-quic-transport.md
huitema: Ian, there are two references to quiescence in the current draft. One about the Ping frame, "The default is to send a PING frame after 15 seconds of quiescence." The other about flow control, "it is generally considered best to not let the sender go into quiescence if avoidable." But the concept of quiescence is not well defined. Is it just "has not sent packet for a long time"? 
What would you suggest to do?

mikkelfj: > What if two packets were received and the first had the spin bit set and the second did not? I would think the spin bit would still be set by the responder in that case?

If an old packet race ahead this could add noise to signal. I think it would be good to explain how this would work, or alternatively filter out the noise by only flipping when a new highest packet number is received. Noise could probably be used effectively if understood by analyser.

ianswett: Scratch my comment, I think if we define this correctly, we don't need to talk about quiescence, it'll just naturally work out.

draft-ietf-quic-transport.md
```diff
@@ -827,6 +848,24 @@ Implementations MUST assume that an unsupported version uses an unknown packet
...
+
+* The connection initiator sets the spin bit value to the opposite
```
ianswett: I'm confused about why it's the opposite?  My mental model is that the bit would be set to 1 on a packet, and then it would be bounced back and forth, so there would only be a single packet marked with the 1 spin bit at any given point in time.

ianswett: Ok, ekr explained to me out of band that I misunderstood how the spin bit worked.  I think the design here is more robust to loss than what I was thinking of, so I'm happy with it.

However, an overview sentence explaining what it's doing (ie: mark the entire flight with the spin bit, etc) would have prevented my confusion.

janaiyengar:
Can we please move discussions to issues, and not have them on PRs?

britram:
@janaiyengar I filed #631 for that purpose.

britram:
And the loss-detection inference ability of this proposal can be discussed in #632.

draft-ietf-quic-transport.md
ianswett: One more comment: I think we need text on packet reordering here.  Possibly this should be the largest packet received instead of the last?

janaiyengar:
I've moved discussion to the issue (thanks for filing it, @britram).

ianswett:
@janaiyengar To clarify, I think my comments were editorial, and so those still belong here, not in the issue?

janaiyengar:
Ian -- I wasn't responding to your comments specifically, but I wanted to
move general discussion over to the issue.

ianswett:
Great, thanks for clarifying.

draft-ietf-quic-transport.md
mikkelfj: The terminology elsewhere appears to be Client and Server, not Connecton Initiator and Connection Responder, although I prefer the latter since the protocol is not necessarily used in a client-server setup.

martinthomson: "connection initiator" and "connection responder"  are just pretentious ways of saying client and server.  We happily use client-server protocols in non-client-server arrangements all the time.

huitema: Martin, initiator and responder are certainly equivalent to client and server when the application is HTTP. Not necessarily so is other applications, e.g. DNS or Web RTC. But as you seem to feel strongly about it, lets use Client and Server in the PR to make it consistent with the rest of the document.

draft-ietf-quic-transport.md
```diff
@@ -854,10 +854,12 @@ The latency spin bit enables latency monitoring from observation points on
...
 
 * The server sets the spin bit value to the value of the
-  spin bit in the last packet received from the client.
+  spin bit in the packet received from the client with
+  the largest sequence number.
 
 * The client sets the spin bit value to the opposite
-  of the last value received from the server, or to 0
+  of the value set in the packet received from the server with the
+  largest sequence number, or to 0
```
ianswett: nit: I think these two lines may fit on one line?

martinthomson:
Aside from being bitrotten, #1046 covers this now.

PR #619, opened 2017-06-09
Frame type word

ekr:


martinthomson:
@janaiyengar, I think that this is a strict improvement.  The current layout has bits in the order DOS and the corresponding fields in the order DSO.  While DOS is a great acronym, I think that the order of the fields makes more sense, so @ekr's suggestion here is quite reasonable.

Now that I look at it, perhaps the most logical order is SOD.  That way you can say "this is a frame for stream S, starting at offset O and of length D". In particular, frames that mention streams would then have the stream ID in the same place, mostly (I'd want to fix RST_STREAM at the same time).

Either way, I think that a consistent order is best.  Also, it would be nice to have this in the implementation draft.

ekr:
I do agree that SOD would be superior. I was just trying to make the minimal change.

Issue #571, opened 2017-06-04
Transition to half closed local should probably happen on FIN ack


siyengar:
in 10.2.2. 
" An endpoint sending an FIN flag causes the stream state to become “half-closed (local)”. An endpoint receiving a FIN flag causes the stream state to become “half-closed (remote)” once all preceding data has arrived"

this seems to suggest that a sender will transition to half closed immediately after sending a FIN. Is that intentional? 

It seems that it would be better for the sender to transition to FIN only after getting the ACK otherwise you'd get into a case where:

Sender sends FIN -> transitions to half closed local
Receiver races with a RST -> sender has sent the FIN, so doesn't send a RST stream is response
Sender transitions to closed
FIN gets lost

However now that the sender is in closed it might still have to retransmit the FIN. Making the sender wait until all bytes until FIN are ACKed to transition to half closed local seems better, that way if the FIN (or any pending bytes before FIN) is not ACKed, the sender could just go into half closed remote like normal when it gets the RST.

I like the idea of closed state being a terminal state where no additional work is needed from the sender.

martinthomson:
This is just an inconsistency.  The text already says that streams close when you have no more to send (or no more to receive), with emphasis:

> A stream transitions from this state to “closed” when a STREAM frame that contains a FIN flag is received **and all prior data has arrived**, or when a RST_STREAM frame is received.

We should try to be more explicit.

janaiyengar:
The *stream* doesn't need to wait for the FIN to be acked. It's a reliable frame, so the connection will ensure that the FIN bit gets retransmitted, but I think the stream should close in the case you mention. The stream state machine does not include reliability states (it does not include acking of any data).

mikkelfj:
> The stream doesn't need to wait for the FIN to be acked. It's a reliable frame

Does this consider state complexity as a function of races with RST?

PR #623, opened 2017-06-10
Fix stream state machine

### LINKED ISSUE #572: Should we transition to Open state when recv even MAX_STREAM_DATA or RST on idle
While thinking about implementation of the handling of idle state, it seems like we'd practically allocate the stream state object anyway if we get a MAX_STREAM_DATA or STREAM_BLOCKED (to keep track of flow control).

It seems better then to treat these streams as equivalent to Open streams because it'd simplify the logic of deciding the Stream limit count to advertise. If the reason we want to allow this transition is to deal with lost packets, we'd practically allocate the stream data eventually when the packets are recovered. 

Proposing a change where:

Idle -> Open

transition occurs not only on STREAM, but also BLOCKED, MAX_STREAM_DATA as well.

martinthomson:
This corrects a few infidelities about the way that stream states
update.  This includes Subodh's excellent suggestion to limit
transitions from idle to only go through open.

Fixes #572, #571.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2428,11 +2428,11 @@ From the "open" state, either endpoint can send a frame with the FIN flag set,
 which causes the stream to transition into one of the "half-closed" states.
 This flag can be set on the frame that opens the stream, which causes the stream
 to immediately become "half-closed".  Once an endpoint has completed sending all
-stream data and a STREAM frame with a FIN flag, including any retransmissions,
-the stream state becomes "half-closed (local)".  When an endpoint receives all
-stream data a FIN flag the stream state becomes "half-closed (remote)".  An
-endpoint MUST NOT consider the stream state to have changed until all data has
-been acknowledged, received or discarded.
+stream data and a STREAM frame with a FIN flag, the stream state becomes
+"half-closed (local)".  When an endpoint receives all stream data a FIN flag the
+stream state becomes "half-closed (remote)".  An endpoint MUST NOT consider the
+stream state to have changed until all data has been sent, received or
+discarded.
 
 A RST_STREAM frame on an "open" stream causes the stream to become
 "half-closed".  A stream that becomes "open" as a result of sending or receiving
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2428,18 +2417,28 @@ A stream in the "open" state may be used by both peers to send frames of any
...
+From the "open" state, either endpoint can send a frame with the FIN flag set,
+which causes the stream to transition into one of the "half-closed" states.
+This flag can be set on the frame that opens the stream, which causes the stream
+to immediately become "half-closed".  Once an endpoint has completed sending all
+stream data and a STREAM frame with a FIN flag, including any retransmissions,
+the stream state becomes "half-closed (local)".  When an endpoint receives all
+stream data a FIN flag the stream state becomes "half-closed (remote)".  An
 endpoint MUST NOT consider the stream state to have changed until all data has
-arrived.
+either been acknowledged or received.
```
mikkelfj: I'm not sure if it is obvious that a RST is also considered acknowledged data here.

mikkelfj: But from the following it seems this also not required.

martinthomson: Thanks for picking up on that.

draft-ietf-quic-transport.md
```diff
@@ -2428,18 +2417,28 @@ A stream in the "open" state may be used by both peers to send frames of any
...
+From the "open" state, either endpoint can send a frame with the FIN flag set,
+which causes the stream to transition into one of the "half-closed" states.
+This flag can be set on the frame that opens the stream, which causes the stream
+to immediately become "half-closed".  Once an endpoint has completed sending all
+stream data and a STREAM frame with a FIN flag, including any retransmissions,
+the stream state becomes "half-closed (local)".  When an endpoint receives all
+stream data a FIN flag the stream state becomes "half-closed (remote)".  An
 endpoint MUST NOT consider the stream state to have changed until all data has
-arrived.
+been acknowledged, received or discarded.
```
janaiyengar: The "arrived" was deliberate -- the stream does not need to hang around in open state until all data is acked. That's a matter of reliability, which should not have anything to do with stream states. I would like to keep stream states decoupled from reliability requirements.

janaiyengar: I would suggest that you revert the changes that require receipt of acks for sent data for streams to transition to closed. The other changes LGTM.

PR #633, opened 2017-06-15
    correct the definition of the ACK FRAME N bit

mcmanus:
* it does not indicate if there is more than 1 range in the frame as
     even if present it may be 0

    * the num blocks field is not in the ack block section at all as stated


mikkelfj:
That doesn't make sense to me. Why would it be optional if the count is 0. That just adds ambiguity. Then it should always be present and the N-bit removed, or it should not be permitted when the count is 0.

ianswett:
From an editorial perspective, I can see this as an improvement, even though I was initially confused.

I would like to only have one way to represent a single range, so I opened issue #635

PR #641, opened 2017-06-19
Update note well in CONTRIBUTING.md for 8179

britram:
With the publication of RFC 8179, the Note Well has been updated to refer to it.


martinthomson:
Such is the virtue of using BCP links.  I don't have to update the template, it's just QUIC that has the busted link.  Of course, the BCP link is one click further away from the real content.

britram:
Right. This is perhaps feedback we should give for the IETF website (https://www.ietf.org/about/note-well.html gives RFC numbers and links by RFC number as well...)

PR #653, opened 2017-06-21
reword 'cumulative ack' language to avoid tcp confusion

mcmanus:
I have had the experience of being confused by the "cumulative ack"
langauge in the current draft and then explaining it to others
confused about the same thing.

cumulative' is not used in the tcp sense where
the largest-ack is cumulative for all of the packets previous to
it. It tends to mean non-revokable here (which the draft also uses -
imo to more clarity.) While future acks can incorprate the data from
old ack frames (being in that sense cumulative) they do not have to
and in some cases will not be able to be so for space reasons.

In this case I think it is just the clarifying text that is confusing
- so less is more.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1581,9 +1581,9 @@ to unnecessarily retransmit some data.  When this is necessary, the receiver
 SHOULD acknowledge newly received packets and stop acknowledging packets
 received in the past.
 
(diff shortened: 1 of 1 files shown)


siyengar:
+1 this, I also did a back-take when I first read this.

janaiyengar:
heh, fair enough.

PR #657, opened 2017-06-23
A STREAM frame MUST NOT be empty

### LINKED ISSUE #652: Stream frame data length wording.
In current draft there is sentence: "A STREAM frame MUST have either non-zero data length or the FIN bit set."
At least for me this was confusing if it refers to Data Length field + D bit. Wouldn't a better wording be something similar to:  "A STREAM frame MUST have either non-empty Stream Data payload or the FIN bit set." or "A STREAM frame MUST have either some Stream Data payload or the FIN bit set without payload"?

martinthomson:
Closes #652.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1544,9 +1544,9 @@ Stream Data:
 
 : The bytes from the designated stream to be delivered.
 
-A STREAM frame MUST have either non-zero data length or the FIN bit set.  When
-the FIN flag is sent on an empty STREAM frame, the offset in the STREAM frame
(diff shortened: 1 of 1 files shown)


martinthomson:
@ianswett, see the referenced issue.  I had no issue with the original text, but Mike's text seemed like a (minor) improvement.

ianswett:
Agreed, it's a minor improvement, I just wanted to ensure I wasn't missing something.

MikeBishop:
The previous wording implied, contrary to what i think we all meant, that non-zero length *and* FIN on the same data frame was an invalid combination.

PR #668, opened 2017-06-27
[transport] disambiguate stream frame 'payload'

phluid61:
"payload" already means something else, and this seems to be the only instance of it being used this way.

Better words may be available.


martinthomson:
Merged manually.

PR #671, opened 2017-06-28
Clarify how connection ID can be omitted

### LINKED ISSUE #659: When can the connection ID be omitted?
The current draft doesn't really talk much about when the connection ID can be omitted. Is this a unilateral unidirectional decision by a sender? Is it negotiated somehow? How are connections being identified in the absence of a connection ID? Are there cases when connection IDs need to be sent again after having been omitted? Etc., etc.

martinthomson:
The definition of the truncate_connection_id was correct, but the name
was poor (it implied that part of the connection ID was included) and it
wasn't referenced from the critical part of the document (the short
header definition).  This corrects both of those oversights.

Closes #659.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -468,7 +468,10 @@ Connection ID Flag:
 
 : The second bit (0x40) of the first octet indicates whether the Connection ID
   field is present.  If set to 1, then the Connection ID field is present; if
-  set to 0, the Connection ID field is omitted.
+  set to 0, the Connection ID field is omitted.  The Connection ID field can
+  only be omitted if the omit_connection_id transport parameter
+  ({{transport-parameter-definitions}}) is used by the intended recipient of the
+  packet.
 
 Key Phase Bit:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -468,7 +468,10 @@ Connection ID Flag:
 
 : The second bit (0x40) of the first octet indicates whether the Connection ID
   field is present.  If set to 1, then the Connection ID field is present; if
-  set to 0, the Connection ID field is omitted.
+  set to 0, the Connection ID field is omitted.  The Connection ID field can
+  only be omitted if the omit_connection_id transport parameter
+  ({{transport-parameter-definitions}}) is used by the intended recipient of the
```
ianswett: nit: By used, do you mean negotiated?

martinthomson: This isn't really a negotiation.  Each endpoint indicates whether it will accept a packet that omits connection ID.  Nothing that its peer says will change that.  "Used" is perhaps a little too generic, but would "includes" work?

ianswett: Includes is ok, but maybe "supplied" or "specified"?

PR #672, opened 2017-06-28
Stream correlation in Transport

MikeBishop:
This is an augmentation of @martinthomson's PR to add stream correlation, as promised.

# Major changes

Leveraging @igorlord's insight that OO=00 only occurs on the first STREAM frame of a stream, I used that as the trigger for a Stream Properties byte.  Two bits of that byte describe the directionality of the stream:

- Unidirectional (no response expected)
- Initial bidirectional (one response expected)
- Initial multi-response (one or more responses expected; needs a better name)
- Response

If the type is Response, there's an Associated Stream ID field, length given by two more bits following the same pattern as the SS bits in the STREAM frame ID.

# Personal Opinion

On the plus side, these stream types seem to cover the abstractions I can envision for most applications.  You can unilaterally send something (unidirectional), do request/response (bidirectional), or pub/sub (single subscription stream, series of update streams).

I don't care for the fact that I still need the stream type header in HTTP after putting this in the transport.  That will be ameliorated if we go back to one stream per request, since all unidirectional streams will be push streams.  (As a side-note, I considered using the multiple-response option in the HTTP mapping, but then I need a stream header again to indicate which is the response and which the pushes.)

I particularly don't like that you now have to look at the frame type header to find out whether a field exists which tells you the length of something else in the header.  I'd like to simplify that.  I went with this model over a CREATE_STREAM frame because of @mikkelfj's use-case of very small messages -- this adds only one byte to the first frame on a stream in one direction and 2-5 bytes to the first frame of response streams.  A separate frame type would be somewhat larger, but could be cleaner in that respect.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -251,11 +251,19 @@ it MUST terminate the connection with an HTTP_INVALID_STREAM_HEADER error.
 ### Request Streams {#stream-request}
 
 Request streams are opened by the client.  Request streams carry the headers and
-optional trailers of requests.
+any trailers of requests.
 
-The stream header for a request stream contains only the type octet,
-which is set to 0x00.  After the stream header, request streams contain a
-sequence of frames (see {{http-framing-layer}}).
+~~~~~
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+
+|   Type (8)    |
++-+-+-+-+-+-+-+-+
+~~~~~
+
+The stream header for a request stream contains only the type octet, which is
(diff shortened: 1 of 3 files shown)


mikkelfj:
I have so far only looked at transport.md - I think the exact encoding is not too important for now - notably var length integers might replace bits in the type byte.

One thought is that it might not be desirable to check for association errors because the initial stream might be closed and gone before a response arrives. Checking validity would require state to be retained - in principle indefinitely. But the association type may still be useful to and API layer.

MikeBishop:
Good point.  Most things fall quickly into the bucket of “associating to these is not valid” – all unidirectional streams, response streams, and bidirectional streams that have seen a response are the same. But every pub/sub stream remains valid forever, as does a bidirectional stream that never gets answered.  There are some possible memory attacks there we’d need to iron out if we went this way.

martinthomson:
@MikeBishop, do you think that #720 might be better and we could close this?

MikeBishop:
Closed in favor of #720.

PR #681, opened 2017-07-12
Try to clarify the situation around sending ACKs in response to

ekr:
ACK-only packets.

This does not seem entirely clear in the text.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1831,7 +1831,9 @@ Connection ID:
 Receivers send ACK frames to inform senders which packets they have received and
 processed, as well as which packets are considered missing.  The ACK frame
 contains between 1 and 256 ACK blocks.  ACK blocks are ranges of acknowledged
-packets.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1831,7 +1831,9 @@ Connection ID:
 Receivers send ACK frames to inform senders which packets they have received and
 processed, as well as which packets are considered missing.  The ACK frame
 contains between 1 and 256 ACK blocks.  ACK blocks are ranges of acknowledged
-packets.
+packets. Implementations SHOULD NOT generate ACK packets in response to packets
```
ianswett: I'd try for MUST NOT and MUST ACK, though I admit it's a bit difficult to enforce either.

ianswett: I'd go for MUST NOT and MUST ACK, though I admit it's a bit difficult to enforce either.

ekr:
I would be fine with that. I only used SHOULD because the later guidance on ACK tracking is SHOULD.

PR #685, opened 2017-07-15
Change Client Initial size to 1200 flat

### LINKED ISSUE #684: Simplify Client Initial size
The consensus at the hackathon was that the rules around size of the initial packet were overly onerous.  We agreed at the table that a flat 1200 floor was easiest and equally effective.

martinthomson:
Closes #684.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -541,7 +541,7 @@ older than 1.3 is negotiated.
 
 QUIC requires that the initial handshake packet from a client fit within the
 payload of a single packet.  The size limits on QUIC packets mean that a record
-containing a ClientHello needs to fit within 1165 octets.
+containing a ClientHello needs to fit within 1171 octets.
 
 A TLS ClientHello can fit within this limit with ample space remaining.
 However, there are several variables that could cause this limit to be exceeded.
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -541,7 +541,7 @@ older than 1.3 is negotiated.
 
 QUIC requires that the initial handshake packet from a client fit within the
 payload of a single packet.  The size limits on QUIC packets mean that a record
-containing a ClientHello needs to fit within 1197 octets.
+containing a ClientHello needs to fit within 1165 octets.
 
 A TLS ClientHello can fit within this limit with ample space remaining.
 However, there are several variables that could cause this limit to be exceeded.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -573,11 +573,11 @@ the packet contents increment the packet number by one, see
 ({{packet-numbers}}).
 
 The payload of a Client Initial packet consists of a STREAM frame (or frames)
-for stream 0 containing a cryptographic handshake message, plus any PADDING
-frames necessary to ensure that the packet is at least the minimum PMTU size
-(see {{packetization}}).  The stream in this packet always starts at an offset
-of 0 (see {{stateless-retry}}) and the complete cyptographic handshake message
-MUST fit in a single packet (see {{handshake}}).
+for stream 0 containing a cryptographic handshake message, with enough PADDING
+frames that the packet is at least 1200 octets (see {{packetization}}).  The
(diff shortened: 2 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2206,19 +2206,16 @@ An endpoint MUST NOT reduce their MTU below this number, even if it receives
...
 
 Clients MUST ensure that the first packet in a connection, and any
-retransmissions of those octets, has a QUIC packet size of least 1232 octets for
-an IPv6 packet and 1252 octets for an IPv4 packet.  In the absence of extensions
-to the IP header, padding to exactly these values will result in an IP packet
-that is 1280 octets.
+retransmissions of those octets, has a QUIC packet size of least 1200 octets.
 
-The initial client packet SHOULD be padded to exactly these values unless the
+The initial client packet SHOULD be padded to exactly 1200 octets unless the
```
ekr: Can you make clear that this is just the QUIC packet, not the headers.

martinthomson: Done.

draft-ietf-quic-tls.md
```diff
@@ -541,7 +541,7 @@ older than 1.3 is negotiated.
 
 QUIC requires that the initial handshake packet from a client fit within the
 payload of a single packet.  The size limits on QUIC packets mean that a record
-containing a ClientHello needs to fit within 1197 octets.
+containing a ClientHello needs to fit within 1165 octets.
```
ianswett: To be clear, this 1165 number may change if we increase or decrease the amount of framing overhead QUIC adds.  Are we ok about that?

martinthomson: I'm OK with that.  As long as we stop changing the overhead when we ship the spec.

draft-ietf-quic-transport.md
```diff
@@ -2206,19 +2206,17 @@ An endpoint MUST NOT reduce their MTU below this number, even if it receives
...
-reduce the amplitude of amplification attacks caused by server responses toward
-an unverified client address.
+The initial client packet SHOULD be padded to exactly 1200 octets unless the
+client has a reasonable assurance that the PMTU is larger.  This size includes
+the QUIC header and integrity check, but not the UDP or IP header. Sending a
+packet of this size ensures that the network path supports an MTU of this size
+and helps reduce the amplitude of amplification attacks caused by server
+responses toward an unverified client address.
 
 Servers MUST ignore an initial plaintext packet from a client if its total size
```
ianswett: It'd be nice to come up with a consistent way of referring to this "total size" which does not include IP+UDP.

draft-ietf-quic-tls.md
ianswett: One q: How are you arriving at 1165 if the 1200 doesn't include IP + UDP?  1(type byte) + 8(conn id) + 4(packet number) + stream frame type byte(1) + stream id(1 byte) + ?

draft-ietf-quic-transport.md
martinthomson: "QUIC packet size" or "packet size" should do.

draft-ietf-quic-transport.md
```diff
@@ -1140,7 +1140,7 @@ max_packet_size (0x0005):
   the endpoint is willing to receive, encoded as an unsigned 16-bit integer.
   This indicates that packets larger than this limit will be dropped.  The
   default for this parameter is the maximum permitted UDP payload of 65527.
-  Values below 1252 are invalid.  This limit only applies to protected packets
+  Values below 1200 are invalid.  This limit only applies to protected packets
```
ianswett: Can we somehow clarify this size is also without UDP + IP headers?

ianswett: Payload size?

martinthomson:
@ianswett,

> One q: How are you arriving at 1165 if the 1200 doesn't include IP + UDP? 1(type byte) + 8(conn id) + 4(packet number) + stream frame type byte(1) + stream id(1 byte) + ?

1200- all of that(15) - version(4) - fnv-1a(8), which I realize now is 29 and not 35 because I calculated the original number with FNV-1a 128.

PR #688, opened 2017-07-16
missing cross ref

reschke:


martinthomson:
I think that the intent here is not to reference h2, but to use CONNECTION_CLOSE.  @MikeBishop?

MikeBishop:
Correct; this should be a reference to 6.1, not RFC 7540.

reschke:
I see. I thought it was because the problem goes back to early versions, and that section number apparently never existed...

MikeBishop:
Merged manually, fixed to local ref.

PR #692, opened 2017-07-19
HTTP requests on one stream

### LINKED ISSUE #245: Single stream for headers and body

MikeBishop:
Fixes #245,#557  as agreed in Prague.  Keeps HPACK from blowing up on RST_STREAM by fixing the dynamic table size at zero until we replace HPACK.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -197,10 +197,10 @@ stream 5, and so on. The server's first push consumes stream 2.
 This stream carries frames related to the request/response (see {{frames}}).
 When a stream terminates cleanly, if the last frame on the stream was truncated,
 this MUST be treated as a connection error (see HTTP_MALFORMED_* in
-{{http-error-codes}}).  Streams which terminate abruptly may do so at any point
-in the frame.
+{{http-error-codes}}).  Streams which terminate abruptly may be reset at any
+point in the frame.
 
-Streams SHOULD be utilized sequentially, with no gaps.  Streams used for pushed
+Streams SHOULD be used sequentially, with no gaps.  Streams used for pushed
 resources MAY be initiated out-of-order, but stream IDs SHOULD be allocated to
 promised resources sequentially.
 
@@ -587,8 +587,8 @@ The payload consists of:
 
 # Error Handling {#errors}
 
-QUIC allows the application to abruptly terminate individual streams or the
-entire connection when an error is encountered.  These are referred to as
+QUIC allows the application to abruptly terminate (reset) individual streams or
+the entire connection when an error is encountered.  These are referred to as
 "stream errors" or "connection errors" and are described in more detail in
 [QUIC-TRANSPORT].
 
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -237,8 +237,7 @@ be used.
 Trailing header fields are carried in an additional header block following the
 body. Such a header block is a sequence of HEADERS frames with End Header Block
 set on the last frame. Senders MUST send only one header block in the trailers
-section; receivers MUST decode any subsequent header blocks in order to maintain
-HPACK decoder state, but the resulting output MUST be discarded.
+section; receivers MUST discard any subsequent header blocks.
 
 An HTTP request/response exchange fully consumes a QUIC stream. After sending a
 request, a client closes the stream for sending; after sending a response, the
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -181,93 +181,67 @@ application.
...
+means that the client's first request occurs on QUIC stream 3, the second on
+stream 5, and so on. The server's first push consumes stream 2.
+
+This stream carries frames related to the request/response (see {{frames}}).
+When a stream terminates cleanly, if the last frame on the stream was truncated,
+this MUST be treated as a connection error (see HTTP_MALFORMED_* in
+{{http-error-codes}}).  Streams which terminate abruptly may do so at any point
+in the frame.
+
+Streams must be utilized sequentially, with no gaps.  HTTP does not need to do
```
MikeBishop: Regardless, it's a lower-case MUST, so I need to change this to something.  I'll take a look.

draft-ietf-quic-http.md
```diff
@@ -181,93 +181,67 @@ application.
...
-##  Stream 1: Connection Control Stream
+the stream management. An HTTP request/response consumes a single stream: This
+means that the client's first request occurs on QUIC stream 3, the second on
+stream 5, and so on. The server's first push consumes stream 2.
+
+This stream carries frames related to the request/response (see {{frames}}).
+When a stream terminates cleanly, if the last frame on the stream was truncated,
+this MUST be treated as a connection error (see HTTP_MALFORMED_* in
+{{http-error-codes}}).  Streams which terminate abruptly may do so at any point
+in the frame.
```
MikeBishop: From the transport draft, RST_STREAM is used "to abruptly terminate a stream."  Basically, I'm saying don't worry about truncated frames in the event of resets.

draft-ietf-quic-http.md
```diff
@@ -181,93 +181,70 @@ application.
...
+
+Streams used for requests MUST be utilized sequentially, with no gaps.  Streams
```
siyengar: is this a sender side only restriction, because there doesn't seem to be a way to enforce this on the receiver

ianswett: Only senders create streams, so by definition, I think that's a yes?  But I agree there's no good way to enforce this.

draft-ietf-quic-http.md
```diff
@@ -181,93 +181,70 @@ application.
...
+used for pushed resources MAY be initiated out-of-order, but stream IDs SHOULD
```
ianswett: I don't have a problem with it, but Is there a reason to allow this?

MikeBishop: Artifact of PUSH_PROMISE including a Stream ID.  Servers can fulfill pushes in a different order than they actually promise them, based on its best estimate of priority.  So when you've promised something on streams 2,4,6,8 based on the requested resource but you decide that stream 6's resource is the most important, you actually start sending data on 6 first.

If we pulled out Martin's idea that PUSH_PROMISE specifies an ID rather than a stream, and the push tells you what promise it's answering, then you can tighten up this side since the first stream will just go to the first thing the server wants to send.

ianswett: I'm happy as is, thanks for the clarification.

draft-ietf-quic-http.md
MikeBishop: Do you think it should be a SHOULD, since it's unenforceable?

ianswett: I think a SHOULD is appropriate, given it's unenforceable.

draft-ietf-quic-http.md
martinthomson: use the word "reset" then

draft-ietf-quic-http.md
```diff
@@ -184,93 +184,70 @@ application.
...
+means that the client's first request occurs on QUIC stream 3, the second on
+stream 5, and so on. The server's first push consumes stream 2.
+
+This stream carries frames related to the request/response (see {{frames}}).
+When a stream terminates cleanly, if the last frame on the stream was truncated,
+this MUST be treated as a connection error (see HTTP_MALFORMED_* in
+{{http-error-codes}}).  Streams which terminate abruptly may do so at any point
+in the frame.
+
+Streams SHOULD be utilized sequentially, with no gaps.  Streams used for pushed
```
martinthomson: https://grammarpartyblog.com/2012/01/17/use-versus-utilize/

MikeBishop: Not new in this PR, but sure.

PR #694, opened 2017-07-24
ENABLE_PUSH instead of DISABLE_PUSH to match HTTP/2

### LINKED ISSUE #477: Use positive phrasing instead of double negative
The setting `SETTINGS_DISABLE_PUSH` is set to `false` by default. This double negation seems needlessly confusing. (*"Setting the push setting to true turns off push?"*)

Suggestion: Renaming it to `SETTINGS_ENABLE_PUSH` and default to `true`. This is easier to understand and matches the parameter defined by HTTP/2.

MikeBishop:
Fixes #477.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -363,7 +363,7 @@ stream of the dependent request, not the data stream.
 HTTP/QUIC supports server push as described in {{!RFC7540}}. During connection
 establishment, the client indicates whether it is willing to receive server
 pushes via the SETTINGS_ENABLE_PUSH setting in the SETTINGS frame (see
-{{connection-establishment}}), which defaults to 0 (false).
+{{connection-establishment}}), which is disabled by default.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -362,8 +362,8 @@ stream of the dependent request, not the data stream.
 
 HTTP/QUIC supports server push as described in {{!RFC7540}}. During connection
 establishment, the client indicates whether it is willing to receive server
-pushes via the SETTINGS_DISABLE_PUSH setting in the SETTINGS frame (see
-{{connection-establishment}}), which defaults to 1 (true).
+pushes via the SETTINGS_ENABLE_PUSH setting in the SETTINGS frame (see
+{{connection-establishment}}), which defaults to 0 (false).
```
martinthomson: I would remove the value and just say ", which is disabled by default".  The way that we express Boolean values is somewhat special and suggesting a concrete value invites misinterpretation.

PR #695, opened 2017-07-24
Cite RFC5234 for ABNF

### LINKED ISSUE #404: ABNF doesn't cite RFC 5234
@reschke points out that we can't use ABNF without citing RFC 5234.  Note to self to add reference in appropriate place.

MikeBishop:
Fixes #404.  (And appropriate that issue #404 is to indicate that something is missing.)


reschke:
Looks good to me.

PR #697, opened 2017-07-26
Move GOAWAY from transport to HTTP

### LINKED ISSUE #696: Move GOAWAY to HTTP draft
We've discussed doing this multiple times, but I can't find an issue on it.  Opening one; if I missed an obvious duplicate, feel free to resolve.

MikeBishop:
Fixes #696.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -585,6 +585,116 @@ The payload consists of:
   : HPACK-compressed request headers for the promised response.
 
 
+### GOAWAY {#frame-goaway}
+
+The GOAWAY frame (type=0x7) is used to initiate graceful shutdown of a
+connection.  GOAWAY allows an endpoint to stop accepting new streams while still
+finishing processing of previously established streams.  This enables
+administrative actions, like server maintenance.
+
+There is an inherent race condition between an endpoint starting new streams and
+the remote sending a GOAWAY frame.  To deal with this case, the GOAWAY contains
+the stream identifier of the last peer- initiated stream that was or might be
+processed on the sending endpoint in this connection.  For instance, if the
+server sends a GOAWAY frame, the identified stream is the highest-numbered
+stream initiated by the client.  This MAY be lower than the stream limit
+identified by a QUIC MAX_STREAM_ID frame.
+
+Once sent, the sender will refuse requests sent on streams initiated by the
+receiver if the stream has an identifier higher than the included last stream
+identifier.  Receivers of a GOAWAY frame MUST NOT send new requests on the
+connection, although a new connection can be established for new streams.
+
+If the receiver of the GOAWAY has sent requests on streams with a higher stream
+identifier than what is indicated in the GOAWAY frame, those streams are not or
+will not be processed.  The receiver of the GOAWAY frame can treat the requests
+as though they had never been sent at all, thereby allowing those requests to be
+retried later on a new connection.  Servers SHOULD treat requests on streams
+above this ID as stream errors of type HTTP_REQUEST_CANCELLED.
+
+Endpoints SHOULD send a GOAWAY frame when the closing of a connection is known
+in advance, even if the window is small, so that the remote peer can know
+whether a stream has been partially processed or not.  For example, if an HTTP
+client sends a POST at the same time that a server closes a QUIC connection, the
+client cannot know if the server started to process that POST request if the
+server does not send a GOAWAY frame to indicate what streams it might have acted
+on.
+
+For unexpected closures caused by error conditions, transport-level mechanisms
+SHOULD be used.  However, a GOAWAY MAY still be sent to provide additional
+detail to clients.
+
+~~~~~~~~~~  drawing
+ +---------------------------------------------------------------+
+ |                   Last-Stream-ID (32)                         |
+ +---------------------------------------------------------------+
+ |                     [Error Code (32)]                         |
+ +---------------------------------------------------------------+
+ |                 [Additional Debug Data (*)]                 ...
+ +---------------------------------------------------------------+
+~~~~~~~~~~
+ {: #fig-goaway title="GOAWAY frame payload"}
+
+The GOAWAY frame does not define any flags.
+
+The GOAWAY frame applies to the connection, not a specific stream. An endpoint
+MUST treat a GOAWAY frame on a stream other than the control stream as a
+connection error ({{errors}}) of type HTTP_WRONG_STREAM.
+
+The last stream identifier in the GOAWAY frame contains the highest- numbered
+stream identifier for which the sender of the GOAWAY frame might have taken some
+action on or might yet take action on.  All streams up to and including the
+identified stream might have been processed in some way.  The last stream
+identifier can be set to 0 if no streams were processed.
+
+> Note: In this context, "processed" means that some data from the stream was
+> passed to some higher layer of software that might have taken some action as a
+> result.
+
+If a connection terminates without a GOAWAY frame, the last stream identifier is
+effectively the highest possible stream identifier.
+
+Reattempting requests on streams with lower- or equal-numbered identifiers is
+not possible, with the exception of idempotent actions like HTTP GET, PUT, or
+DELETE.  Any protocol activity that used higher-numbered streams can be safely
+retried using a new connection.
+
+Activity on streams numbered lower or equal to the last stream identifier might
+still complete successfully.  The sender of a GOAWAY frame might gracefully shut
+down a connection by sending a GOAWAY frame, maintaining the connection in an
+"open" state until all in- progress streams complete.
+
+An endpoint MAY send multiple GOAWAY frames if circumstances change. For
+instance, an endpoint that sends GOAWAY without an error code during graceful
+shutdown could subsequently encounter an error condition.  The last stream
+identifier from the last GOAWAY frame received indicates which streams could
+have been acted upon.  Endpoints MUST NOT increase the value they send in the
+last stream identifier, since the peers might already have retried unprocessed
+requests on another connection.
+
+A client that is unable to retry requests loses all requests that are in flight
+when the server closes the connection.  A server that is attempting to
+gracefully shut down a connection SHOULD send an initial GOAWAY frame with the
+last stream identifier set to 2^32-1.  This signals to the client that a
+shutdown is imminent and that initiating further requests is prohibited.  After
+allowing time for any in-flight requests (at least one round-trip time), the
+server MAY send another GOAWAY frame with an updated last stream identifier.
+This ensures that a connection can be cleanly shut down without losing requests.
+
+After sending a GOAWAY frame, the sender MUST reset any existing or subsequently
+opened streams with an HTTP_REQUEST_CANCELLED error code.
+
+The GOAWAY frame contains an optional 32-bit error code ({{http-error-codes}})
+that contains the reason for closing the connection.  If no error has been
+encountered, this field MAY be omitted.
+
+Endpoints MAY append opaque data to the payload of any GOAWAY frame. Additional
+debug data is intended for diagnostic purposes only and carries no semantic
+value.  Debug information could contain security- or privacy-sensitive data.
+Logged or otherwise persistently stored debug data MUST have adequate safeguards
+to prevent unauthorized access.
+
+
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-http.md
```diff
@@ -585,6 +585,116 @@ The payload consists of:
...
+> Note: In this context, "processed" means that some data from the stream was
+> passed to some higher layer of software that might have taken some action as a
+> result.
+
+If a connection terminates without a GOAWAY frame, the last stream identifier is
+effectively the highest possible stream identifier.
+
+Reattempting requests on streams with lower- or equal-numbered identifiers is
+not possible, with the exception of idempotent actions like HTTP GET, PUT, or
+DELETE.  Any protocol activity that used higher-numbered streams can be safely
```
martinthomson: This is probably cribbed from RFC 7450 (as is most of this), however it could be improved.  I think that you want to concentrate first on what this guarantees (requests on higher stream numbers can be retried automatically), then note that other streams cannot be automatically retried unless it is otherwise permitted (for example, a request with an idempotent method).

draft-ietf-quic-http.md
```diff
@@ -585,6 +585,116 @@ The payload consists of:
...
+retried using a new connection.
+
+Activity on streams numbered lower or equal to the last stream identifier might
+still complete successfully.  The sender of a GOAWAY frame might gracefully shut
+down a connection by sending a GOAWAY frame, maintaining the connection in an
+"open" state until all in-progress streams complete.
```
martinthomson: We can be more concrete about this.  Say that GOAWAY doesn't cause the connection to close, and the connection will remain open until requests on streams up to the last stream ID have completed.  After that, the connection can be closed (let's avoid making a statement about silent close vs. a signaled close at this point by remaining silent on exactly *how* it is closed).

draft-ietf-quic-http.md
```diff
@@ -585,6 +585,116 @@ The payload consists of:
...
+shutdown could subsequently encounter an error condition.  The last stream
+identifier from the last GOAWAY frame received indicates which streams could
+have been acted upon.  Endpoints MUST NOT increase the value they send in the
+last stream identifier, since the peers might already have retried unprocessed
+requests on another connection.
+
+A client that is unable to retry requests loses all requests that are in flight
+when the server closes the connection.  A server that is attempting to
+gracefully shut down a connection SHOULD send an initial GOAWAY frame with the
+last stream identifier set to 2^32-1.  This signals to the client that a
```
martinthomson: This requirement is no longer correct because we have MAX_STREAM_ID to govern this.  You can send a GOAWAY with the value of MAX_STREAM_ID and be reasonably sure that all in-flight requests are covered.

draft-ietf-quic-http.md
```diff
@@ -585,6 +585,116 @@ The payload consists of:
...
+shutdown is imminent and that initiating further requests is prohibited.  After
+allowing time for any in-flight requests (at least one round-trip time), the
+server MAY send another GOAWAY frame with an updated last stream identifier.
+This ensures that a connection can be cleanly shut down without losing requests.
+
+After sending a GOAWAY frame, the sender MUST reset any existing or subsequently
+opened streams with an HTTP_REQUEST_CANCELLED error code.
```
martinthomson: I think that this paragraph is incorrect.  I think that there is a missing condition here: it only does this if it wants to allow those requests to be retried.

draft-ietf-quic-http.md
```diff
@@ -585,6 +585,116 @@ The payload consists of:
...
+last stream identifier, since the peers might already have retried unprocessed
+requests on another connection.
+
+A client that is unable to retry requests loses all requests that are in flight
+when the server closes the connection.  A server that is attempting to
+gracefully shut down a connection SHOULD send an initial GOAWAY frame with the
+last stream identifier set to 2^32-1.  This signals to the client that a
+shutdown is imminent and that initiating further requests is prohibited.  After
+allowing time for any in-flight requests (at least one round-trip time), the
+server MAY send another GOAWAY frame with an updated last stream identifier.
```
martinthomson: There's an open question here about whether a recipient of GOAWAY can initiate any more requests.  I don't think that we were sufficient clear about that in h2; this is similarly unclear.  I think that we should allow for new requests to be initiated here; after all, we have no clear demarcation point that can be used.  I think that we might advise against increasing MAX_STREAM_ID after sending GOAWAY (at least on the server side).  But that goes back to my other comments, which you will soon see on the issue.

draft-ietf-quic-http.md
```diff
@@ -585,6 +585,116 @@ The payload consists of:
...
+
+If the receiver of the GOAWAY has sent requests on streams with a higher stream
+identifier than what is indicated in the GOAWAY frame, those streams are not or
+will not be processed.  The receiver of the GOAWAY frame can treat the requests
+as though they had never been sent at all, thereby allowing those requests to be
+retried later on a new connection.  Servers SHOULD treat requests on streams
+above this ID as stream errors of type HTTP_REQUEST_CANCELLED.
+
+Endpoints SHOULD send a GOAWAY frame when the closing of a connection is known
+in advance, even if the window is small, so that the remote peer can know
```
MikeBishop: If the server knows that it's going to close a connection, it should send GOAWAY to warn the client.  Errors that warrant CONNECTION_CLOSE aren't generally known in advance, but if they are, then I agree that this should include them.

draft-ietf-quic-http.md
```diff
@@ -585,6 +585,116 @@ The payload consists of:
...
+whether a stream has been partially processed or not.  For example, if an HTTP
+client sends a POST at the same time that a server closes a QUIC connection, the
+client cannot know if the server started to process that POST request if the
+server does not send a GOAWAY frame to indicate what streams it might have acted
+on.
+
+For unexpected closures caused by error conditions, transport-level mechanisms
+SHOULD be used.  However, a GOAWAY MAY still be sent to provide additional
+detail to clients.
```
MikeBishop: Fair enough.

draft-ietf-quic-http.md
MikeBishop: I think 7540 (and therefore this text) is fairly explicit on that point:

> Receivers of a GOAWAY frame MUST NOT send new requests on the connection, although a new connection can be established for new streams.

I agree that the server can't actually enforce that precisely, however, just as it can't in h2.

draft-ietf-quic-http.md
MikeBishop: IIRC, that text was targeted (in H2) at the disjoint case -- you supply a horizon (the GOAWAY) and then RST the individual streams below that.  I've tried to clarify that.

draft-ietf-quic-http.md
martinthomson: This is probably something that needs to be said explicitly here.

draft-ietf-quic-http.md
```diff
@@ -591,6 +591,83 @@ The payload consists of:
...
+administrative actions, like server maintenance.  GOAWAY by itself does not
+close a connection.
+
+The GOAWAY frame does not define any flags, and the payload is a QUIC stream
+identifier.  The GOAWAY frame applies to the connection, not a specific stream.
+An endpoint MUST treat a GOAWAY frame on a stream other than the control stream
+as a connection error ({{errors}}) of type HTTP_WRONG_STREAM.
+
+There is an inherent race condition between a client sending new requests and
+the server sending a GOAWAY frame.  To deal with this case, the GOAWAY contains
```
martinthomson: We never explained the nature of that race condition in 7540, and now it's bugging me.  Should we open an issue to track this?

Maybe the right way to fix this is to change the description.  Say that this frame exists to allow client and server to agree on what requests were accepted and which were rejected.  GOAWAY identifies the last request that the client can make and the server will accept.  All requests that are sent on higher-numbered streams are implicitly cancelled.

Described that way, there isn't a need to talk about the race condition.

draft-ietf-quic-http.md
```diff
@@ -591,6 +591,83 @@ The payload consists of:
...
+
+Once sent, the server will refuse requests sent on streams with an identifier
+higher than the included last stream identifier.  Clients MUST NOT send new
+requests on the connection after receiving GOAWAY, although requests might
+already be in transit. A new connection can be established for new requests.
+
+If the client has sent requests on streams with a higher stream identifier than
+indicated in the GOAWAY frame, those requests were not and will not be
+processed.  Servers SHOULD reset new streams above this ID with the error code
+HTTP_REQUEST_CANCELLED.  Servers MAY also reset streams below the indicated ID
```
martinthomson: s/Servers SHOULD reset new streams/Endpoints SHOULD reset client-initiated streams/

It's not especially graceful, but the server might see those streams as open, and cancelling them avoids problems.  Also, this isn't about the server, because clients should also reset their requests and avoid retransmissions.  


Separately, I wonder: HTTP_REQUEST_CANCELLED from the client doesn't have the effect that we want: it doesn't guarantee that the request wasn't processed by the server; it can't.  Do we need some text on that.

draft-ietf-quic-http.md
```diff
@@ -591,6 +591,83 @@ The payload consists of:
...
+with HTTP_REQUEST_CANCELLED if the associated requests were not processed.
+
+The client can treat these requests as though they had never been sent at all,
```
martinthomson: A client can treat requests that are cancelled by the server as though they have never been sent

If the client cancels them, then they still might be live.  (Yeah, this is in the context of a server-sent thing, but it's better to have an accurate context-free statement.)

draft-ietf-quic-http.md
```diff
@@ -591,6 +591,83 @@ The payload consists of:
   : HPACK-compressed request headers for the promised response.
 
 
+### GOAWAY {#frame-goaway}
+
+The GOAWAY frame (type=0x7) is used to initiate graceful shutdown of a
+connection by a server.  GOAWAY allows a server to stop accepting new requests
+while still finishing processing of previously received requests.  This enables
+administrative actions, like server maintenance.  GOAWAY by itself does not
+close a connection.
```
martinthomson: I think that we should explicitly say that a client doesn't need to send GOAWAY: it just stops sending requests.

draft-ietf-quic-http.md
```diff
@@ -591,6 +591,83 @@ The payload consists of:
...
+The GOAWAY frame does not define any flags, and the payload is a QUIC stream
+identifier.  The GOAWAY frame applies to the connection, not a specific stream.
+An endpoint MUST treat a GOAWAY frame on a stream other than the control stream
+as a connection error ({{errors}}) of type HTTP_WRONG_STREAM.
+
+There is an inherent race condition between a client sending new requests and
+the server sending a GOAWAY frame.  To deal with this case, the GOAWAY contains
+the stream identifier of the last client-initiated request that was or might be
+processed in this connection.  This MAY be lower than the stream limit
+identified by a QUIC MAX_STREAM_ID frame, and MAY be zero if no requsets were
```
martinthomson: typo: requsets

draft-ietf-quic-http.md
```diff
@@ -591,6 +591,83 @@ The payload consists of:
...
+processed.  Servers SHOULD reset new streams above this ID with the error code
+HTTP_REQUEST_CANCELLED.  Servers MAY also reset streams below the indicated ID
+with HTTP_REQUEST_CANCELLED if the associated requests were not processed.
+
+The client can treat these requests as though they had never been sent at all,
+thereby allowing them to be retried later on a new connection. Automatically
+retrying other requests is not possible, unless this is otherwise permitted
+(e.g. idempotent actions like GET, PUT, or DELETE).  Requests on stream IDs less
+than or equals to the stream ID in the GOAWAY frame remain open until they are
+completed successfully, reset, or the connection terminates.
```
martinthomson: The "remain open" clause here is potentially confusing.  The problem here is that stream state can't be affected by these HTTP frames.  I think that you might remove this last sentence.

MikeBishop: I think the sentence itself is worth emphasizing rather than dropping, but I agree that it's a little awkward to say that the request "remains open," since that sounds more like the stream this text refers to in 7540.  Hopefully a more appropriate phrase for a request will occur to me when it's not half past bedtime.

martinthomson:
Merge it!

PR #701, opened 2017-07-27
Push ID

### LINKED ISSUE #702: Concurrent server pushes for the same resource
@kazuho pointed out that because there is no ordering between streams, you can have the need to trigger a server push on multiple streams.  If two concurrent requests identify the same subresource, then you end up having to push that subresource twice.  That's inefficient.

This is less of a problem in h2 where the server could push for the first response and avoid pushing on the second.  Strict ordering means that it knows that the client has at least seen the PUSH_PROMISE for the first response.  Of course, it's not a perfect design: routing of the responses at the client might mean that the they might be consumed by different entities, leading to the possibility that the second doesn't get a chance to see the push.

martinthomson:
As requested, I split this out of the unidirectional PR (#643).

There is a change here, that needs to be highlighted.  @kazuho points out that without inter-stream ordering guarantees, requests might be triggered on multiple streams.  The server then has an awkward choice: either push multiple times, or risk the client triggering requests when streams arrive out of order.  This contains a fix for that bug.

The fix is to allow the same push ID to be used in multiple PUSH_PROMISE frames.  Then, fulfillment happens at most once.

I think that this is a neat fix, but it creates two more problems:

1. How does a client know when the server will stop referencing a push ID in PUSH_PROMISE?  With reordering, the pushed response might arrive and be consumed before a PUSH_PROMISE that references it.  I think that it might be possible to add a counter to the push stream header that says how many times the server promised this response.  The problem with that is that if a stream that carries the PUSH_PROMISE is reset, then that counter will be too high.  I've decided to just explain the issue and add some weasel wording rather than provide a direct fix.

2. What if the server provides different headers for the same Push ID?  I've chosen to fix this by simply forbidding variations in the content of the header block.  I considered only allowing the headers to appear once, but then you have to deal with that stream being reset.  Repetition should be rare enough, and header compression might be able to help with the extra bytes.

I think that those problems are both less serious than the bug this fixes.  Besides, the first is really only an generalized form of a problem we have already: the stream carrying a PUSH_PROMISE can be reset.  What currently happens is that the server more or less has to send a push stream anyway, even though it can't know whether the client got the PUSH_PROMISE.  With this, a server can use CANCEL_REQUEST.

Closes #702, #281.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -368,14 +368,14 @@ same Push ID can be used in multiple PUSH_PROMISE frames (see
 {{frame-push-promise}}).
 
 After the push stream header, a push contains a response ({{request-response}}),
-with response headers and (if present) trailers carried by HEADERS frames sent
-on the control stream, and response body (if any) carried by DATA frames.
+with response headers, a response body (if any) carried by DATA frames, then
+trailers (if any) carried by HEADERS frames.
 
-If a promised push stream is not needed by the client, the client SHOULD send a
-QUIC STOP_SENDING on the promised stream with an appropriate error code (e.g.
-HTTP_PUSH_REFUSED, HTTP_PUSH_ALREADY_IN_CACHE; see {{errors}}).  This asks the
-server not to transfer the data and indicates that it will be discarded upon
-receipt.
+If a promised server push is not needed by the client, the client SHOULD send a
+CANCEL_PUSH frame; if the push stream is already open, a QUIC STOP_SENDING frame
+with an appropriate error code can be used instead (e.g., HTTP_PUSH_REFUSED,
+HTTP_PUSH_ALREADY_IN_CACHE; see {{errors}}).  This asks the server not to
+transfer the data and indicates that it will be discarded upon receipt.
 
 
 # HTTP Framing Layer {#http-framing-layer}
@@ -452,13 +452,10 @@ The flags defined are:
 
   PUSH_PRIORITIZED (0x04):
   : Indicates that the Prioritized Stream is a server push rather than a
-    request.  If set, this flag indicates that the Prioritized Stream identifies
-    a response stream and the Promise Index field is present.
+    request.
 
   PUSH_DEPENDENT (0x02):
-  : Indicates a dependency on a pushed request.  If set, this flag indicates
-    that the Dependent Stream identifies a response stream and that the
-    Dependent Promise Index field is present.
+  : Indicates a dependency on a server push.
 
   E (0x01):
   : Indicates that the stream dependency is exclusive (see {{!RFC7540}}, Section
@@ -481,15 +478,15 @@ The PRIORITY frame payload has the following fields:
 
   Prioritized Request:
   : A 32-bit identifier for a request.  This contains the stream ID of a request
-    stream when the PUSH_PRIORITIZED flag is clear, or a Push ID when the PUSH
-    flag is set.
+    stream when the PUSH_PRIORITIZED flag is clear, or a Push ID when the
+    PUSH_PRIORITIZED flag is set.
 
   Stream Dependency:
   : A 32-bit stream identifier for a dependent request.  This contains the
     stream ID of a request stream when the PUSH_DEPENDENT flag is clear, or a
     Push ID when the PUSH_DEPENDENT flag is set.  A request stream ID of 0
     indicates a dependency on the root stream. For details of dependencies,
-    see {{priority}} and {!RFC7540}}, Section 5.3).
+    see {{priority}} and {!RFC7540}}, Section 5.3.
 
   Weight:
   : An unsigned 8-bit integer representing a priority weight for the stream (see
@@ -502,9 +499,9 @@ when the corresponding PUSH_PRIORITIZED or PUSH_DEPENDENT flag is not set.
 Setting the PUSH or PUSH_DEPENDENT flag causes the frame to identify a
 PUSH_PROMISE using a Push ID (see {{frame-push-promise}} for details).
 
-A PRIORITY frame MAY identify no request by using a stream ID of 0; as in
-{{!RFC7540}}, this makes the request dependent on the root of the dependency
-tree.
+A PRIORITY frame MAY identify no request in the Prioritized Request field by
+using a stream ID of 0; as in {{!RFC7540}}, this makes the request dependent on
+the root of the dependency tree.
 
 A PRIORITY frame MAY identify a dependent stream with a stream ID of 0; as in
 {{!RFC7540}}, this makes the request dependent on the root of the dependency
@@ -521,50 +518,37 @@ The length of a PRIORITY frame is 9 octets.  A PRIORITY frame with any other
 length MUST be treated as a connection error of type HTTP_MALFORMED_PRIORITY.
 
 
-### CANCEL_REQUEST {#frame-cancel-request}
+### CANCEL_PUSH {#frame-cancel-push}
 
-The CANCEL_REQUEST frame (type=0x3) is used to request cancellation of a request
-prior to the response stream being created.  This frame can be used to cancel
-server push requests that are initiated with a PUSH_PROMISE.
-
-The CANCEL_REQUEST frame identifies a server push request by Push ID (see
-{{frame-push-promise}}).
+The CANCEL_PUSH frame (type=0x3) is used to request cancellation of server push
+prior to the push stream being created.  The CANCEL_REQUEST frame identifies a
+server push request by Push ID (see {{frame-push-promise}}).
 
 When a server receives this frame, it aborts sending the response for the
 identified server push.  If the server has not yet started to send the server
-push, it can use the receipt of a CANCEL_REQUEST frame to avoid opening a
+push, it can use the receipt of a CANCEL_PUSH frame to avoid opening a
 stream.  If the push stream has been opened by the server, the server SHOULD
 sent a QUIC RST_STREAM frame on those streams and cease transmission of the
 response.
 
 A server can send this frame to indicate that it won't be sending a response
-prior to creation of a push stream.  Once the push stream has been created, a
-RST_STREAM with a CANCELLED code can be used instead.
+prior to creation of a push stream.  Once the push stream has been created,
+sending CANCEL_PUSH has no effect on the state of the push stream.  A QUIC
+RST_STREAM frame SHOULD be used instead to cancel transmission of the server
+push response.
 
-A CANCEL_REQUEST frame is sent on the control stream.  Sending a CANCEL_REQUEST
-frame on a stream other than the control stream MUST be treated as a
-HTTP_WRONG_STREAM error.
+A CANCEL_PUSH frame is sent on the control stream.  Sending a CANCEL_PUSH frame
+on a stream other than the control stream MUST be treated as a stream error of
+type HTTP_WRONG_STREAM.
 
-The CANCEL_REQUEST frame has no defined flags.
+The CANCEL_PUSH frame has no defined flags.
 
-~~~~~~~~~~  drawing
-    0                   1                   2                   3
-    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
-   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-   |                         Push ID (32)                          |
-   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-~~~~~~~~~~
-{: #fig-cancel-request title="CANCEL_REQUEST frame payload"}
+The CANCEL_PUSH frame carries a 32-bit Push ID that identifies the server push
+that is being cancelled (see {{frame-push-promise}}).
 
-The CANCEL_REQUEST frame payload has the following fields:
-
-  Push ID:
-  : A 32-bit Push ID that identifies the request (see {{frame-push-promise}}).
-
-A CANCEL_REQUEST frame always contains a four octet payload.  A server MUST
-treat a CANCEL_REQUEST frame payload that is other than 4 octets in length, or a
-CANCEL_REQUEST frame that identifies an unknown Push ID as an
-HTTP_MALFORMED_CANCEL_REQUEST error.
+A server MUST treat a CANCEL_PUSH frame payload that is other than 4 octets in
+length, or a CANCEL_PUSH frame that identifies an unknown Push ID as a
+connection error of type HTTP_MALFORMED_CANCEL_PUSH.
 
 
 ### SETTINGS {#frame-settings}
@@ -699,10 +683,9 @@ server to client, as in HTTP/2.  The PUSH_PROMISE frame defines no flags.
 The payload consists of:
 
 Push ID:
-: A 32-bit identifier for the server push request.  This identifier is used in
-  push stream header ({{server-push}}), CANCEL_REQUEST frames
-  ({{frame-cancel-request}}), and PRIORITY frames ({{frame-priority}}) to
-  identify this request.
+: A 32-bit identifier for the server push request.  A push ID is used in push
+  stream header ({{server-push}}), CANCEL_PUSH frames ({{frame-cancel-push}}),
+  and PRIORITY frames ({{frame-priority}}).
 
 Header Block:
 : HPACK-compressed request headers for the promised response.
@@ -952,8 +935,8 @@ PRIORITY (0x2):
 
 RST_STREAM (0x3):
 : RST_STREAM frames do not exist, since QUIC provides stream lifecycle
-  management.  The same code point is used for the CANCEL_REQUEST frame
-  ({{frame-cancel-request}}).
+  management.  The same code point is used for the CANCEL_PUSH frame
+  ({{frame-cancel-push}}).
 
 SETTINGS (0x4):
 : SETTINGS frames are sent only at the beginning of the connection.  See
@@ -1156,7 +1139,7 @@ The entries in the following table are registered by this document.
 | DATA           | 0x0  | {{frame-data}}           |
 | HEADERS        | 0x1  | {{frame-headers}}        |
 | PRIORITY       | 0x2  | {{frame-priority}}       |
-| CANCEL_REQUEST | 0x3  | {{frame-cancel-request}} |
+| CANCEL_PUSH    | 0x3  | {{frame-cancel-push}}    |
 | SETTINGS       | 0x4  | {{frame-settings}}       |
 | PUSH_PROMISE   | 0x5  | {{frame-push-promise}}   |
 | Reserved       | 0x6  | N/A                      |
@@ -1273,6 +1256,7 @@ The original authors of this specification were Robbie Shade and Mike Warres.
 - Use separate frame type and settings registries from HTTP/2 (#81)
 - SETTINGS_ENABLE_PUSH instead of SETTINGS_DISABLE_PUSH (#477)
 - Restored GOAWAY (#696)
+- Identify server push using Push ID rather than a stream ID (#702,#281)
 
 ## Since draft-ietf-quic-http-03
 
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -329,7 +329,8 @@ either by identifying the stream that carries a request or by using a Push ID
 ({{frame-push-promise}}).  Other than the means of identifying requests, the
 prioritization system is identical to that in HTTP/2.
 
-Only a client can prioritize requests.  A server MUST NOT send a PRIORITY frame.
+Only a client can send PRIORITY frames.  A server MUST NOT send a PRIORITY
+frame.
 
 
 ## Server Push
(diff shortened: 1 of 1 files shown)


MikeBishop:
I'll note that, with unidirectional streams, we could use a stream from the server for the request headers as well, and then only have them once.  The problem with that is that, until they're processed, the client doesn't know what request it is that it shouldn't be making as it processes the original resource.

draft-ietf-quic-http.md
```diff
@@ -430,22 +473,86 @@ The flags defined are:
...
+length MUST be treated as a connection error of type HTTP_MALFORMED_PRIORITY.
+
+
+### CANCEL_REQUEST {#frame-cancel-request}
+
+The CANCEL_REQUEST frame (type=0x3) is used to request cancellation of a request
+prior to the response stream being created.  This frame can be used to cancel
+server push requests that are initiated with a PUSH_PROMISE.
+
+The frame identifies a server push request is identified by Push ID (see
```
mcmanus: There is a typo in here somewhere.. "request _that_ is identified by" perhaps?

martinthomson:
@MikeBishop, I considered that.  Getting the ordering guarantees right here is tricky and I think that it's unfair on clients to say "I'm going to promise you something", but not tell them exactly what.

draft-ietf-quic-http.md
```diff
@@ -331,18 +339,40 @@ pushes via the SETTINGS_ENABLE_PUSH setting in the SETTINGS frame (see
...
+client MUST treat this as a connection error of type HTTP_DUPLICATE_PUSH.  The
+same Push ID can be used in multiple PUSH_PROMISE frames (see
+{{frame-push-promise}}).
 
-The server push response is conveyed in the same way as a non-server-push
-response, with response headers and (if present) trailers carried by HEADERS
-frames, and response body (if any) carried by DATA frames.
+After the push stream header, a push contains a response ({{request-response}}),
+with response headers and (if present) trailers carried by HEADERS frames sent
+on the control stream, and response body (if any) carried by DATA frames.
```
MikeBishop: Not on the control stream.

draft-ietf-quic-http.md
```diff
@@ -402,15 +432,28 @@ without EHB set, followed by a HEADERS frame with EHB set.
...
 The flags defined are:
 
+  PUSH_PRIORITIZED (0x04):
+  : Indicates that the Prioritized Stream is a server push rather than a
+    request.  If set, this flag indicates that the Prioritized Stream identifies
+    a response stream and the Promise Index field is present.
+
+  PUSH_DEPENDENT (0x02):
+  : Indicates a dependency on a pushed request.  If set, this flag indicates
+    that the Dependent Stream identifies a response stream and that the
```
MikeBishop: Saying that it "identifies a response stream" in both these places feels a little odd.  If it's the stream ID, then just say it's the stream ID and you don't even need the flags.  If it's the Push ID, use the term here for clarity.

draft-ietf-quic-http.md
```diff
@@ -430,22 +473,87 @@ The flags defined are:
 
 The PRIORITY frame payload has the following fields:
 
-  Prioritized Stream:
-  : A 32-bit stream identifier for the request stream whose priority is being
-    updated.
+  Prioritized Request:
+  : A 32-bit identifier for a request.  This contains the stream ID of a request
+    stream when the PUSH_PRIORITIZED flag is clear, or a Push ID when the PUSH
```
MikeBishop: PUSH => PUSH_PRIORITIZED?

draft-ietf-quic-http.md
```diff
@@ -430,22 +473,87 @@ The flags defined are:
...
-A PRIORITY frame MUST have a payload length of nine octets.  A PRIORITY frame
-of any other length MUST be treated as a connection error of type
-HTTP_MALFORMED_PRIORITY.
+A PRIORITY frame identifies a request to priotize, and a request upon which that
+request is dependent.  A request is identified by the stream ID of a request
+when the corresponding PUSH_PRIORITIZED or PUSH_DEPENDENT flag is not set.
+Setting the PUSH or PUSH_DEPENDENT flag causes the frame to identify a
+PUSH_PROMISE using a Push ID (see {{frame-push-promise}} for details).
+
+A PRIORITY frame MAY identify no request by using a stream ID of 0; as in
```
MikeBishop: Since there are two references in a PRIORITY frame, and setting one of them to zero is prohibited by your other pull request, perhaps be more specific than "identify" here.

draft-ietf-quic-http.md
```diff
@@ -430,22 +473,87 @@ The flags defined are:
...
+non-existent Push ID MUST be treated as a HTTP_MALFORMED_PRIORITY error.
+
+The length of a PRIORITY frame is 9 octets.  A PRIORITY frame with any other
+length MUST be treated as a connection error of type HTTP_MALFORMED_PRIORITY.
+
+
+### CANCEL_REQUEST {#frame-cancel-request}
+
+The CANCEL_REQUEST frame (type=0x3) is used to request cancellation of a request
+prior to the response stream being created.  This frame can be used to cancel
```
MikeBishop: Given that there aren't (currently 😉) separate request/response streams, this basically can only be used to cancel a push.  Let's call it CANCEL_PUSH for now and not talk about requests in this section.

draft-ietf-quic-http.md
```diff
@@ -430,22 +473,87 @@ The flags defined are:
...
+When a server receives this frame, it aborts sending the response for the
+identified server push.  If the server has not yet started to send the server
+push, it can use the receipt of a CANCEL_REQUEST frame to avoid opening a
+stream.  If the push stream has been opened by the server, the server SHOULD
+sent a QUIC RST_STREAM frame on those streams and cease transmission of the
+response.
+
+A server can send this frame to indicate that it won't be sending a response
+prior to creation of a push stream.  Once the push stream has been created, a
+RST_STREAM with a CANCELLED code can be used instead.
```
MikeBishop: "can" feels like it needs an RFC2119 WORD, but I don't know which.

draft-ietf-quic-http.md
```diff
@@ -430,22 +473,87 @@ The flags defined are:
...
+
+A CANCEL_REQUEST frame is sent on the control stream.  Sending a CANCEL_REQUEST
+frame on a stream other than the control stream MUST be treated as a
```
MikeBishop: "stream error of type X" avoids the argument of whether to do a/an before HTTP.

draft-ietf-quic-http.md
```diff
@@ -430,22 +473,87 @@ The flags defined are:
...
+The CANCEL_REQUEST frame has no defined flags.
+
+~~~~~~~~~~  drawing
+    0                   1                   2                   3
+    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+   |                         Push ID (32)                          |
+   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+~~~~~~~~~~
+{: #fig-cancel-request title="CANCEL_REQUEST frame payload"}
```
MikeBishop: I've been trying to avoid drawings of things that have only a single field and call that single thing the payload of the frame.

draft-ietf-quic-http.md
```diff
@@ -577,12 +686,35 @@ server to client, as in HTTP/2.  It defines no flags.
...
 
-  Promised Stream ID:
-  : A 32-bit Stream ID indicating the QUIC stream on which the response will be
-    sent
-
-  Payload:
-  : HPACK-compressed request headers for the promised response.
+Push ID:
+: A 32-bit identifier for the server push request.  This identifier is used in
+  push stream header ({{server-push}}), CANCEL_REQUEST frames
```
MikeBishop: "in the" or "in ... headers".  "In ... header" sounds odd.

MikeBishop:
Oh, and fill in the changelog.

draft-ietf-quic-http.md
martinthomson: Heh, you noticed that, but then you didn't notice that the ordering was completely wrong :)

draft-ietf-quic-http.md
martinthomson: I think that I can remove the second sentence and rely on the description of the fields below.

draft-ietf-quic-http.md
martinthomson: I've gone with SHOULD.  There isn't any point in MUST here, because it's not a correctness issue, it's just pointless to use CANCEL_PUSH at this point.

draft-ietf-quic-http.md
```diff
@@ -430,29 +469,81 @@ The flags defined are:
...
+    PUSH_PRIORITIZED flag is set.
 
   Stream Dependency:
-  : A 32-bit stream identifier for the request stream that this stream depends
-    on (see {{priority}} and {{!RFC7540}}, Section 5.3).
+  : A 32-bit stream identifier for a dependent request.  This contains the
+    stream ID of a request stream when the PUSH_DEPENDENT flag is clear, or a
+    Push ID when the PUSH_DEPENDENT flag is set.  A request stream ID of 0
+    indicates a dependency on the root stream. For details of dependencies,
+    see {{priority}} and {!RFC7540}}, Section 5.3.
```
MikeBishop: Unbalanced braces

draft-ietf-quic-http.md
```diff
@@ -430,29 +469,81 @@ The flags defined are:
...
 
+A PRIORITY frame identifies a request to priotize, and a request upon which that
+request is dependent.  A request is identified by the stream ID of a request
+when the corresponding PUSH_PRIORITIZED or PUSH_DEPENDENT flag is not set.
+Setting the PUSH or PUSH_DEPENDENT flag causes the frame to identify a
+PUSH_PROMISE using a Push ID (see {{frame-push-promise}} for details).
+
+A PRIORITY frame MAY identify no request in the Prioritized Request field by
+using a stream ID of 0; as in {{!RFC7540}}, this makes the request dependent on
+the root of the dependency tree.
```
MikeBishop: "Prioritized Request" is the request whose priority is being modified.  Per the following paragraph, this cannot be zero (or one).  I think you can delete this paragraph / merge it with the following one.

draft-ietf-quic-http.md
```diff
@@ -430,29 +469,81 @@ The flags defined are:
...
+non-existent Push ID MUST be treated as a HTTP_MALFORMED_PRIORITY error.
+
 The length of a PRIORITY frame is 9 octets.  A PRIORITY frame with any other
 length MUST be treated as a connection error of type HTTP_MALFORMED_PRIORITY.
 
 
+### CANCEL_PUSH {#frame-cancel-push}
+
+The CANCEL_PUSH frame (type=0x3) is used to request cancellation of server push
+prior to the push stream being created.  The CANCEL_REQUEST frame identifies a
```
MikeBishop: CANCEL_REQUEST => CANCEL_PUSH

draft-ietf-quic-http.md
```diff
@@ -430,29 +469,79 @@ The flags defined are:
...
+type HTTP_WRONG_STREAM.
+
+The CANCEL_PUSH frame has no defined flags.
+
+The CANCEL_PUSH frame carries a 32-bit Push ID that identifies the server push
+that is being cancelled (see {{frame-push-promise}}).
+
+A server MUST treat a CANCEL_PUSH frame payload that is other than 4 octets in
+length, or a CANCEL_PUSH frame that identifies an unknown Push ID as a
+connection error of type HTTP_MALFORMED_CANCEL_PUSH.
```
janaiyengar: A CANCEL_PUSH from a server could be reordered with a PUSH_PROMISE... I'm not sure what to do about this but I wanted to raise it.

draft-ietf-quic-http.md
```diff
@@ -583,12 +673,34 @@ server to client, as in HTTP/2.  It defines no flags.
 
 The payload consists of:
 
-  Promised Stream ID:
-  : A 32-bit Stream ID indicating the QUIC stream on which the response will be
-    sent
+Push ID:
+: A 32-bit identifier for the server push request.  A push ID is used in push
```
janaiyengar: It really sounds odd to say "server push request". It's not a request... how about "promised resource"?

draft-ietf-quic-http.md
```diff
@@ -583,12 +673,34 @@ server to client, as in HTTP/2.  It defines no flags.
...
+  and PRIORITY frames ({{frame-priority}}).
+
+Header Block:
+: HPACK-compressed request headers for the promised response.
 
-  Payload:
-  : HPACK-compressed request headers for the promised response.
+A server MAY use the same Push ID in multiple PUSH_PROMISE frames.  This allows
+the server to use the same server push in response to multiple concurrent
+requests.  Referencing the same server push ensures that a PUSH_PROMISE can be
```
janaiyengar: "Referencing the same promised resource"

draft-ietf-quic-http.md
```diff
@@ -583,12 +673,34 @@ server to client, as in HTTP/2.  It defines no flags.
...
+made in relation to every response in which server push might be needed without
+causing duplicating pushes.
```
janaiyengar: "causing duplicating pushes" --> "duplicating pushes"

draft-ietf-quic-http.md
```diff
@@ -583,12 +673,34 @@ server to client, as in HTTP/2.  It defines no flags.
...
+A server that uses the same Push ID in multiple PUSH_PROMISE frames MUST include
+the same header fields each time.  The octets of the header block MAY be
+different due to differing encoding, but the header fields and their values MUST
+be identical.  Note that ordering of header fields is significant.  A client
+MUST treat receipt of a PUSH_PROMISE with conflicting header field values for
+the same Push ID as a connection error of type HTTP_MALFORMED_PUSH_PROMISE.
+
+A server SHOULD avoid sending a PUSH_PROMISE that includes a Push ID that was
+fulfilled prior to the request being made.  Though a client needs to handle
+receiving a promised response prior to it being promised due to reordering of
```
janaiyengar: Suggested rewording: "Due to the possibility of network reordering, a client must be able to receive a promised resource prior to the corresponding PUSH_PROMISE."

draft-ietf-quic-http.md
```diff
@@ -583,12 +673,34 @@ server to client, as in HTTP/2.  It defines no flags.
...
+stream delivery, clients might be unable to reuse a pushed response that was
+long since consumed. \[Editor's Note: should we include a count of the number of
```
janaiyengar: Not sure what this phrase means "clients might be unable to reuse a pushed response that was long since consumed." Can you clarify?

draft-ietf-quic-http.md
```diff
@@ -811,14 +926,17 @@ PRIORITY (0x2):
...
+  ({{frame-cancel-push}}).
 
 SETTINGS (0x4):
 : SETTINGS frames are sent only at the beginning of the connection.  See
   {{frame-settings}} and {{h2-settings}}.
 
 PUSH_PROMISE (0x5):
-: See {{frame-push-promise}}.
+: The PUSH_PROMISE does not reference the stream that carries the response to
+  the pushed request; instead the push stream references the PUSH_PROMISE frame.
```
janaiyengar: reword. suggestion: "The PUSH_PROMISE does not reference the stream carrying the promised resource; instead the push stream references the PUSH_PROMISE frame."

draft-ietf-quic-http.md
```diff
@@ -430,29 +469,79 @@ The flags defined are:
...
 
-  Prioritized Stream:
-  : A 32-bit stream identifier for the request stream whose priority is being
-    updated.
+  Prioritized Request:
+  : A 32-bit identifier for a request.  This contains the stream ID of a request
+    stream when the PUSH_PRIORITIZED flag is clear, or a Push ID when the
+    PUSH_PRIORITIZED flag is set.
 
   Stream Dependency:
```
martinthomson: Hmm, I think that "dependent request" means the stream that is dependent, so I changed the name.  I'll fix the picture.

draft-ietf-quic-http.md
```diff
@@ -430,29 +469,79 @@ The flags defined are:
...
 
   Weight:
   : An unsigned 8-bit integer representing a priority weight for the stream (see
     {{!RFC7540}}, Section 5.3). Add one to the value to obtain a weight between
     1 and 256.
 
-A PRIORITY frame MAY identify a dependent stream with a stream ID of 0; as in
+A PRIORITY frame identifies a request to priotize, and a request upon which that
+request is dependent.  A request is identified by the stream ID of a request
+when the corresponding PUSH_PRIORITIZED or PUSH_DEPENDENT flag is not set.
```
martinthomson: I've reworded a little.  I've used the field names more explicitly.  I didn't want to use "prioritized request" because that is too close to "Prioritized Request ID".

draft-ietf-quic-http.md
```diff
@@ -430,29 +469,79 @@ The flags defined are:
...
+Setting the PUSH or PUSH_DEPENDENT flag causes the frame to identify a
+PUSH_PROMISE using a Push ID (see {{frame-push-promise}} for details).
+
+A PRIORITY frame MAY identify a Stream Dependency with a stream ID of 0; as in
```
martinthomson: That would only work if we prohibit a Push ID of 0.

draft-ietf-quic-http.md
```diff
@@ -430,29 +469,79 @@ The flags defined are:
...
+
 The length of a PRIORITY frame is 9 octets.  A PRIORITY frame with any other
 length MUST be treated as a connection error of type HTTP_MALFORMED_PRIORITY.
 
 
+### CANCEL_PUSH {#frame-cancel-push}
+
+The CANCEL_PUSH frame (type=0x3) is used to request cancellation of server push
+prior to the push stream being created.  The CANCEL_PUSH frame identifies a
+server push request by Push ID (see {{frame-push-promise}}).
```
martinthomson: The resource is the thing identified by a URL.  I'm trying to be consistent here by always identifying requests, either those made by the client, or those synthesized/predicted by the server.

draft-ietf-quic-http.md
martinthomson: That's a completely different statement.

draft-ietf-quic-http.md
martinthomson: A server might promise with Push ID X and fulfill that promise.  Later, it could send the same PUSH_PROMISE for X and assume that the client already has the response.  A client can't be expected to retain the response.  I have a different proposal for handling this now that I've had time away from the problem, I'll write that down.

draft-ietf-quic-http.md
```diff
@@ -402,15 +431,25 @@ without EHB set, followed by a HEADERS frame with EHB set.
 ### PRIORITY {#frame-priority}
 
 The PRIORITY (type=0x02) frame specifies the sender-advised priority of a stream
-and is substantially different from {{!RFC7540}}. In order to support ordering,
-it MUST be sent only on the control stream. The format has been modified to
-accommodate not being sent on-stream and the larger stream ID space of QUIC.
+and is substantially different in format from {{!RFC7540}}.  In order to ensure
```
MikeBishop: This is an echo of language elsewhere in the HTTP draft, indicating which frames are mostly-direct copies of the HTTP/2 equivalent, and which ones are fundamentally different.  Happy to discuss rephrasing that all-up in a separate editorial issue, but I think following the existing pattern in this PR is entirely appropriate.

draft-ietf-quic-http.md
MikeBishop: It is a request -- the PUSH_PROMISE contains a request body that the server is telling the client it expects the client will want to make shortly.

draft-ietf-quic-http.md
MikeBishop: At least one other place, you refer simply to "the push" -- after some initial definition, that might be less awkward, though it feels somewhat informal.

draft-ietf-quic-http.md
MikeBishop: Technically, there's nothing stopping you from having two different pushes for the same resource in flight at once.  I'm drawing a blank on why you would do that, but I'm not sure we want to foreclose it here.

draft-ietf-quic-http.md
```diff
@@ -494,17 +495,19 @@ The PRIORITY frame payload has the following fields:
...
 A PRIORITY frame identifies a request to priotize, and a request upon which that
-request is dependent.  A request is identified by the stream ID of a request
-when the corresponding PUSH_PRIORITIZED or PUSH_DEPENDENT flag is not set.
-Setting the PUSH or PUSH_DEPENDENT flag causes the frame to identify a
-PUSH_PROMISE using a Push ID (see {{frame-push-promise}} for details).
-
-A PRIORITY frame MAY identify a Stream Dependency with a stream ID of 0; as in
-{{!RFC7540}}, this makes the request dependent on the root of the dependency
+request is dependent.  A Prioritized Request ID or Stream Dependency ID
+identifies a client initiated request using the corresponding stream ID when the
```
MikeBishop: I believe client-initiated should be hyphenated.

MikeBishop:
Random question while reading -- is '0' a valid Push ID?  The answer isn't entirely obvious, and I could see implementations differing painfully.

draft-ietf-quic-http.md
martinthomson: That's another reason to talk about "requests" rather than resources.

martinthomson:
I hope to fix the 0 thing in a follow-on.  See my response to Jana on the mailing list.  Right now, the entire 32-bit space is open for business.

draft-ietf-quic-http.md
janaiyengar: +1

PR #703, opened 2017-07-27
Don't allow reprioritization of stream 0

martinthomson:
I can't believe we didn't do this in h2, but it's not something we should allow.


MikeBishop:
Reprioritizing Stream 1 wouldn't have any practical impact, but sure.

Issue #328, opened 2017-02-20
TIME_WAIT


mcmanus:
6.4 currently has
`TODO: Connections that are terminated are added to a TIME_WAIT list at the server, so as to absorb any straggler packets in the network. Discuss TIME_WAIT list.`

this is an open issue for the TODO.

Some initial thoughts: is it always the server that is in TIME_WAIT as opposed to the TCP active close behavior? (seems fine at first blush, but asking for a friend who is less sure :)).

what happens in the TIME_WAIT state.. retransmissions? What happens when you receive packets (ignored - trigger retrrans, RST?)

RyanAtGoogle:
In the current Google QUIC implementation, for each connection in time
wait, the server saves the serialized "close" packet for that connection.
When new packets are received for that connection, the server retransmits
the close packet (subject to back-off). This is to handle the case where
the close packet is lost and so the client is unaware that the connection
has been closed. I can imagine that there might be better alternatives,
though.

mcmanus:
on the list @janaiyengar notes

> On the point of what to do in TIME_WAIT, the server should hang on to some
> state. At the least, hanging on to the connection ID protects the server by
> preventing late packets from causing creation of new connection state, and
> the server can simply discard incoming packets for connection IDs in
> TIME_WAIT. (If a server wanted to hang on to more state, it could do so,
> allowing it to retransmit CONNECTION_CLOSE frames later, but as you note,
> this is more state and I wouldn't require it.)

```

MikeBishop:
But why is that the server specifically?  It seems like either peer, having sent a CONNECTION_CLOSE and not received one, needs to be in this state.

martinthomson:
I think that @MikeBishop is right.

Rather than just fixing this one small thing, I think that we need a more comprehensive rework of the shutdown logic.  @janaiyengar sent an email to the list outlining one extreme in the design space.  I want to write up a proposal that is close to that.  That will necessarily include a roughly 2xRTO TIME_WAIT2-like state on both peers.  Then we can discuss whether that sort of design is desirable.

martinthomson:
Fixed by #721, I hope.

PR #705, opened 2017-07-27
Specify RST-like behavior for CONNECTION_CLOSE. Fixes #330, #328

### LINKED ISSUE #330: CONNECTION_CLOSE semantics
From discussion on the list, we could clarify the meaning of CONNECTION_CLOSE. Specifically, it would be good to clarify if it acts more like a TCP FIN or a TCP RST. I would suggest that the analog of a TCP FIN is a FIN on a QUIC stream as they both gracefully half-close a single bidirectional communication channel. After a FIN, retransmission of data may still occur to deal with packet loss.

A QUIC CONNECTION_CLOSE, on the other hand, is an immediate termination of the connection. No more data is retransmitted. As such it is more like a TCP RST.

ekr:


siyengar:
lgtm, maybe could use some guidance on how exactly the server could implement such a mechanism, however we can always add that later once we have more implementation experience.

mikkelfj:
This is not entirely clear to me. If there is a requirement for backing off, but not requirement for replying to a connection close (is there?), then state will just be tied up for no good reason. If you have 100K connections per minute, and you need to hold a connection for 10 minutes to make sure you can back off, well ...

ekr:
See my email on this topic to the list where I lay out the tradeoffs.

draft-ietf-quic-transport.md
```diff
@@ -1469,9 +1469,17 @@ of three ways:
...
-TODO: Connections that are terminated are added to a TIME_WAIT list at the
-server, so as to absorb any straggler packets in the network.  Discuss TIME_WAIT
-list.
+After receiving either a CONNECTION_CLOSE frame or a Public Reset, an
+endpoint MUST NOT send additional packets on that connection. After
+sending either a CONNECTION_CLOSE frame or a Public Reset packet,
+implementations MUST NOT send any non-closing packets on that connection. If
+additional packets are received after this time, implementations
+SHOULD respond to them by sending either a CONNECTION_CLOSE frame or a
+Public Reset packet, either of which may just be a duplicate of a
```
martinthomson: I recommend that you use CONNECTION_CLOSE here and instead note that an endpoint MAY send a Public Reset instead if it doesn't want to save the state necessary to generate CONNECTION_CLOSE.

There's a sliding scale here:

1. maintain full state and generate a new CONNECTION_CLOSE in a new packet
2. only keep the packet that holds CONNECTION_CLOSE and retransmit that unmodified
3. drop state and rely on Public/Stateless Reset

draft-ietf-quic-transport.md
```diff
@@ -1469,9 +1469,17 @@ of three ways:
...
+previous packet. Implementations SHOULD throttle these responses, for
+instance by exponentially backing off the number of packets which must
+be received before sending a response.
```
martinthomson: s/must be/are

draft-ietf-quic-transport.md
ekr: I'm happy to revise this if that's the consensus. I was trying to be flexible.

ianswett: I agree that 2 and 3 are a sliding scale and that's why SHOULD makes sense to me.  I haven't come up with any reasons why option 1 is preferable to 2, so I would not suggest it, but maybe there are some reasons I haven't come up with?

janaiyengar: I agree with Ian that we don't need the endpoint to maintain complete state, so 2 and 3 make sense. I'd suggest that we add a suggestion about how long a peer resends CONNECTION_CLOSE after sending it once (basically how long state should be kept in option (2)), and I propose idle_timeout as a time period that makes sense.

janaiyengar:
lgtm. I have a non-blocking suggestion in my inline comment.

draft-ietf-quic-transport.md
martinthomson: I think that we should allow all three, though gently discourage 3.

ekr:
How's this look?

draft-ietf-quic-transport.md
janaiyengar: Whats the argument for allowing 1?

janaiyengar:
LGTM. Not a major issue, but the line breaks are at odd places. If you are able to, please format them to 80 cols.

janaiyengar:
I'd like to merge this. martinthomson@, if you think we should make the text more accommodating, let's do that separately.

PR #711, opened 2017-08-08
Limit pushes using MAX_PUSH_ID

### LINKED ISSUE #709: Make Push ID sequential
And add a MAX_PUSH_ID message to hq

martinthomson:
This repurposes SETTINGS_ENABLE_PUSH as the initial limit.

A client can use this to limit the number of outstanding pushes.  I've included
advice on increasing the limit, not as the pushes are first used, but as the
pushes are completed (or cancelled).

There wasn't a rule for handling a truncated push stream header, so I added
that as well.  The HTTP_DUPLICATE_PUSH error was reclaimed for signaling all
three types of error with a push stream header: truncation, duplicate Push ID,
and exceeding MAX_PUSH_ID.

Closes #709.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -362,9 +362,20 @@ This header consists of a 32-bit Push ID, which identifies a server push (see
 ~~~~~
 {: #fig-push-stream-header title="Push Stream Header"}
 
+A push stream always starts with a 32-bit Push ID.  A client MUST treat
+receiving a push stream that contains fewer than 4 octets as a connection error
+of type HTTP_MALFORMED_PUSH.
+
+A server SHOULD use Push IDs sequentially, starting at 0.  A client uses the
+MAX_PUSH_ID frame ({{frame-max-push-id}}) and SETTINGS_MAX_PUSH_ID
+({{settings-parameters}}) to limit the number of pushes that a server can
+promise initiate.  A client MUST treat receipt of a push stream with a Push ID
+that is greater than the maximum Push ID as a connection error of type
+HTTP_MALFORMED_PUSH.
+
 Each Push ID MUST only be used once in a push stream header.  If a push stream
 header includes a Push ID that was used in another push stream header, the
-client MUST treat this as a connection error of type HTTP_DUPLICATE_PUSH.  The
+client MUST treat this as a connection error of type HTTP_MALFORMED_PUSH.  The
 same Push ID can be used in multiple PUSH_PROMISE frames (see
 {{frame-push-promise}}).
 
@@ -631,8 +642,12 @@ The following settings are defined in HTTP/QUIC:
   SETTINGS_HEADER_TABLE_SIZE (0x1):
   : An integer with a maximum value of 2^32 - 1.  This value MUST be zero.
 
-  SETTINGS_ENABLE_PUSH (0x2):
-  : Transmitted as a Boolean
+  SETTINGS_MAX_PUSH_ID (0x2):
+  : An integer with a maximum value of 2^32 - 1.  This determines the initial
+    maximum Push ID that a server can use.  This setting only has meaning when
+    set by a client; a server MUST NOT provide a value for this setting.  A
+    client MUST treat receipt of this setting from a server as a connection
+    error of type HTTP_MALFORMED_SETTINGS.
 
   SETTINGS_MAX_HEADER_LIST_SIZE (0x6):
   : An integer with a maximum value of 2^32 - 1
@@ -691,6 +706,12 @@ Push ID:
 
 Header Block:
 : HPACK-compressed request headers for the promised response.
+
+A server MUST NOT use a Push ID that is larger than the client has provided in a
+MAX_PUSH_ID frame ({{frame-max-push-id}}) or the SETTINGS_MAX_PUSH_ID setting
+({{settings-parameters}}).  A client that receives a PUSH_PROMISE that contains
+a larger Push ID than is allowed MUST be treated as a connection error of type
+HTTP_MALFORMED_PUSH_PROMISE.
 
 A server MAY use the same Push ID in multiple PUSH_PROMISE frames.  This allows
 the server to use the same server push in response to multiple concurrent
@@ -793,6 +814,30 @@ the server MAY send another GOAWAY frame with an updated last stream identifier.
 This ensures that a connection can be cleanly shut down without losing requests.
 
 
+# MAX_PUSH_ID {#frame-max-push-id}
+
+The MAX_PUSH_ID frame (type=0xA) is used by clients to control the number of
+server pushes that the server can initiate.  This sets the maximum value for a
+Push ID that the server can use in a PUSH_PROMISE frame.  Consequently, this
+also limits the number of push streams that the server can initiate.
+
+The initial value for the maximum Push ID is determined by the
+SETTINGS_MAX_PUSH_ID setting ({{settings-parameters}}).  A client that wishes to
+manage the number of promised server pushes can increase the maximum Push ID by
+sending a MAX_PUSH_ID frame as the server fulfills or cancels server pushes.
+
+The MAX_PUSH_ID frame has no defined flags.
+
+The MAX_PUSH_ID frame carries a 32-bit Push ID that identifies the maximum value
+for a Push ID that the server can use (see {{frame-push-promise}}).
+
+A server MUST NOT send a MAX_PUSH_ID frame.  A client MUST treat the receipt of
+a MAX_PUSH_ID frame as a connection error of type HTTP_MALFORMED_MAX_PUSH_ID.
+
+A server MUST treat a MAX_PUSH_ID frame payload that is other than 4 octets in
+length as a connection error of type HTTP_MALFORMED_MAX_PUSH_ID.
+
+
 # Error Handling {#errors}
 
 QUIC allows the application to abruptly terminate (reset) individual streams or
@@ -862,8 +907,8 @@ HTTP_WRONG_STREAM (0x0F):
 HTTP_MULTIPLE_SETTINGS (0x10):
 : More than one SETTINGS frame was received.
 
-HTTP_DUPLICATE_PUSH (0x11):
-: Multiple push streams used the same Push ID.
+HTTP_MALFORMED_PUSH (0x11):
+: A push stream header was malformed or included an invalid Push ID.
 
 
 # Considerations for Transitioning from HTTP/2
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -794,6 +815,37 @@ the server MAY send another GOAWAY frame with an updated last stream identifier.
 This ensures that a connection can be cleanly shut down without losing requests.
 
 
+# MAX_PUSH_ID {#frame-max-push-id}
+
+The MAX_PUSH_ID frame (type=0xA) is used by clients to control the number of
+server pushes that the server can initiate.  This sets the maximum value for a
+Push ID that the server can use in a PUSH_PROMISE frame.  Consequently, this
+also limits the number of push streams that the server can initiate.
```
RyanAtGoogle: I'm not sure I'm following this completely. Clearly MAX_PUSH_ID limits the number of *promises* that the server can make. But isn't the number of actual push streams limited by the MAX_STREAM_ID? 

(This makes me wonder if "Promise ID" might be a better name than "Push ID" but I may well just be confused.)

MikeBishop: You can't deliver a push that hasn't been promised, so the actual limit is whichever is smaller.

draft-ietf-quic-http.md
```diff
@@ -362,9 +362,20 @@ This header consists of a 32-bit Push ID, which identifies a server push (see
 ~~~~~
 {: #fig-push-stream-header title="Push Stream Header"}
 
+A push stream always starts with a 32-bit Push ID.  A client MUST treat
+receiving a push stream that contains fewer than 4 octets as a connection error
+of type HTTP_MALFORMED_PUSH.
+
+A server SHOULD use Push IDs sequentially, starting at 0.  A client uses the
+MAX_PUSH_ID frame ({{frame-max-push-id}}) and SETTINGS_MAX_PUSH_ID
```
MikeBishop: Nit:  I'd mention these in the order they're used; the setting to provide an initial value and the frame to increment it later.

draft-ietf-quic-http.md
```diff
@@ -693,6 +708,12 @@ Push ID:
 Header Block:
 : HPACK-compressed request headers for the promised response.
 
+A server MUST NOT use a Push ID that is larger than the client has provided in a
+MAX_PUSH_ID frame ({{frame-max-push-id}}) or the SETTINGS_MAX_PUSH_ID setting
+({{settings-parameters}}).  A client that receives a PUSH_PROMISE that contains
+a larger Push ID than is allowed MUST be treated as a connection error of type
+HTTP_MALFORMED_PUSH_PROMISE.
```
MikeBishop: "is allowed" makes it sound like an absolute limit; might be worth rewording slightly.

draft-ietf-quic-http.md
```diff
@@ -794,6 +815,37 @@ the server MAY send another GOAWAY frame with an updated last stream identifier.
 This ensures that a connection can be cleanly shut down without losing requests.
 
 
+# MAX_PUSH_ID {#frame-max-push-id}
+
+The MAX_PUSH_ID frame (type=0xA) is used by clients to control the number of
```
MikeBishop: Note that you're now conflicting with the frame type for the ALTSVC frame in the HTTP/2 registry.  ORIGIN is using 0xC, which suggests someone has 0xB but I can't think of who off the top of my head right now.  Might want to jump out to at least 0xD to be safe, or (ab)use 0x8 instead.

draft-ietf-quic-http.md
```diff
@@ -794,6 +815,37 @@ the server MAY send another GOAWAY frame with an updated last stream identifier.
...
+manage the number of promised server pushes can increase the maximum Push ID by
+sending a MAX_PUSH_ID frame as the server fulfills or cancels server pushes.
+
+The MAX_PUSH_ID frame has no defined flags.
+
+The MAX_PUSH_ID frame carries a 32-bit Push ID that identifies the maximum value
+for a Push ID that the server can use (see {{frame-push-promise}}).  A
+MAX_PUSH_ID frame cannot reduce the maximum Push ID; receipt of a MAX_PUSH_ID
+that contains a smaller value than previously received MUST be treated as a
+connection error of type HTTP_MALFORMED_MAX_PUSH_ID.
```
MikeBishop: General comment:  Because these are on the control stream, which means guaranteed in-order delivery, we *could* use the HTTP/2 model of sending an increment rather than an absolute number.  I'm not sold that the few bytes saved are worth it, though.

draft-ietf-quic-http.md
RyanAtGoogle: Good point. (Though I think it common practice we expect the promise limit to be higher than the stream limit so we would not expect this to be the lower of the two.) Perhaps the text could say:

Consequently, this also limits the number of push streams that the server can initiate in conjunction with MAX_STREAM_ID.

(or some such)

draft-ietf-quic-http.md
```diff
@@ -857,14 +909,17 @@ HTTP_INTERRUPTED_HEADERS (0x0E):
 : A HEADERS frame without the End Header Block flag was followed by a frame
   other than HEADERS.
 
-HTTP_SETTINGS_ON_WRONG_STREAM (0x0F):
-: A SETTINGS frame was received on a request control stream.
+HTTP_WRONG_STREAM (0x0F):
+: A frame was received on a different stream to where it is permitted.
```
MikeBishop: "different... to" is British; "different... than" is American.  "Different... from" is neutral.  Maybe we should use that, or just sidestep the issue with "on a stream where it is not permitted."

draft-ietf-quic-http.md
martinthomson: I think that we can make a bigger saving here by not sending the frame as often, noting that MAX_STREAM_ID covers the resource exhaustion issue.

draft-ietf-quic-http.md
martinthomson: I learn something every day.  I say that in its most Lovecraftian meaning, of course.

draft-ietf-quic-http.md
```diff
@@ -362,9 +362,20 @@ This header consists of a 32-bit Push ID, which identifies a server push (see
...
 {: #fig-push-stream-header title="Push Stream Header"}
 
+A push stream always starts with a 32-bit Push ID.  A client MUST treat
+receiving a push stream that contains fewer than 4 octets as a connection error
+of type HTTP_MALFORMED_PUSH.
+
+A server SHOULD use Push IDs sequentially, starting at 0.  A client uses the
+SETTINGS_MAX_PUSH_ID ({{settings-parameters}}) setting and the MAX_PUSH_ID frame
+({{frame-max-push-id}}) to limit the number of pushes that a server can promise
+initiate.  A client MUST treat receipt of a push stream with a Push ID that is
```
MikeBishop: "promise initiate" needs to have only one word or a conjunction.

draft-ietf-quic-http.md
janaiyengar: I don't think the savings are worth it, since this shouldn't be sent that often really as it is. 

On resource exhaustion, I think both MAX_STREAM_ID and MAX_PUSH_ID both cover that issue, just at different levels. QUIC decides MAX_STREAM_ID based on its resource constraints for streams, and the HTTP "layer" decides MAX_PUSH_ID based on its constraints for handling pushed resources. I think these are independent.

draft-ietf-quic-http.md
```diff
@@ -857,14 +910,17 @@ HTTP_INTERRUPTED_HEADERS (0x0E):
 : A HEADERS frame without the End Header Block flag was followed by a frame
   other than HEADERS.
 
-HTTP_SETTINGS_ON_WRONG_STREAM (0x0F):
-: A SETTINGS frame was received on a request control stream.
+HTTP_WRONG_STREAM (0x0F):
+: A frame was received on a different stream from where it is permitted.
```
janaiyengar: from -> than ? I mostly don't care because it might be about local idioms, but it took me a moment to parse :-)
Offline, Ryan suggested "A frame was received on stream where it is not permitted", which SGTM as well (though it is semantically slightly different).

draft-ietf-quic-http.md
```diff
@@ -632,8 +643,12 @@ The following settings are defined in HTTP/QUIC:
...
   : An integer with a maximum value of 2^32 - 1.  This value MUST be zero.
 
-  SETTINGS_ENABLE_PUSH (0x2):
-  : Transmitted as a Boolean
+  SETTINGS_MAX_PUSH_ID (0x2):
+  : An integer with a maximum value of 2^32 - 1.  This determines the initial
+    maximum Push ID that a server can use.  This setting only has meaning when
+    set by a client; a server MUST NOT provide a value for this setting.  A
+    client MUST treat receipt of this setting from a server as a connection
+    error of type HTTP_MALFORMED_SETTINGS.
```
janaiyengar: Can this be 0? I assume so, since there's no other way to disable PUSH now. Can you state this explicitly here (that a client can disable push by setting this value to 0)?

draft-ietf-quic-http.md
```diff
@@ -794,6 +815,38 @@ the server MAY send another GOAWAY frame with an updated last stream identifier.
 This ensures that a connection can be cleanly shut down without losing requests.
 
 
+# MAX_PUSH_ID {#frame-max-push-id}
+
+The MAX_PUSH_ID frame (type=0xD) is used by clients to control the number of
+server pushes that the server can initiate.  This sets the maximum value for a
+Push ID that the server can use in a PUSH_PROMISE frame.  Consequently, this
+also limits the number of push streams that the server can initiate in addition
+to the limit set by the QUIC MAX_STREAM_ID frame.
```
janaiyengar: It seems silly but possible to send this with 0. Can you add a sentence that says this MUST be > 0?

draft-ietf-quic-http.md
```diff
@@ -794,6 +815,38 @@ the server MAY send another GOAWAY frame with an updated last stream identifier.
...
+SETTINGS_MAX_PUSH_ID setting ({{settings-parameters}}).  A client that wishes to
+manage the number of promised server pushes can increase the maximum Push ID by
+sending a MAX_PUSH_ID frame as the server fulfills or cancels server pushes.
+
+The MAX_PUSH_ID frame has no defined flags.
+
+The MAX_PUSH_ID frame carries a 32-bit Push ID that identifies the maximum value
+for a Push ID that the server can use (see {{frame-push-promise}}).  A
+MAX_PUSH_ID frame cannot reduce the maximum Push ID; receipt of a MAX_PUSH_ID
+that contains a smaller value than previously received MUST be treated as a
```
janaiyengar: I would explicitly state that this previous value may have been received via MAX_PUSH_ID or SETTINGS_MAX_PUSH_ID. Perhaps "a smaller value than that received via a previous MAX_PUSH_ID or SETTINGS_MAX_PUSH_ID ..." ?

draft-ietf-quic-http.md
martinthomson: I took Ryan's text, but I moved that to #712.  This PR will appear to be include the change, but that's just because I rebased on top of #712 on the assumption that it will be merged.

janaiyengar:
Heh -- thanks for rephrasing, it's one of those trivial yet annoying things.

draft-ietf-quic-http.md
MikeBishop: Sending a frame with a value less than you provided in transport settings is a clear error, actually, since you know the transport settings necessarily came before the frame.

PR #712, opened 2017-08-08
Fix error code definitions

martinthomson:
HTTP_MALFORMED_DATA description mentioned HEADERS and it wasn't in the IANA table.

HTTP_WRONG_STREAM was defined as HTTP_SETTINGS_ON_WRONG_STREAM.

@MikeBishop, I hope that this is properly editorial, the HTTP_SETTINGS_ON_WRONG_STREAM thing isn't clear, though WRONG_STREAM is now used in a few other places.


MikeBishop:
"Different... from" versus "different... to" versus "different... than" here as well.

PR #717, opened 2017-08-10
editorial nits

janaiyengar:


martinthomson:
Yikes, I kinda screwed that one up :)

PR #719, opened 2017-08-10
Revise text on ACK frames prior to handshake completion

### LINKED ISSUE #221: 0RTT ACKs
The quic-tls draft says:

```
  What about 0-RTT data?  Should we allow acknowledgment of 0-RTT with
  unprotected frames?  If we don't, then 0-RTT data will be unacknowledged until
  the handshake completes.  This isn't a problem if the handshake completes
  without loss, but it could mean that 0-RTT stalls when a handshake packet
  disappears for any reason.
```

I assume what you're talking about here is the case where the server's first flight of packets are lost, because the server can only find itself in one of two positions because the CH must fit in a single packet.

- It has no ClientHello, in which case it must buffer any packets
- It has a ClientHello, in which case it can can complete the handshake by sending its first flight.

My sense is that having stalling of 0-RTT in the case where parts of the server's first flight are lost is not that big a deal. If you really don't want stalling, you should retransmit faster :)

martinthomson:
This extensively used language like "protected ACK frames", which is incorrect.
I fixed that here, recognizing that more extensive fixes are probably needed.

I also just made the observation that the server might acknowledge 0-RTT, but
the client might be unable to read those ACK frames, which closes #221.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1262,25 +1262,25 @@ able to inject these packets.  Timing and packet retransmission information from
 `ACK` frames is critical to the functioning of the protocol, but these frames
 might be spoofed or altered.
 
-Endpoints MUST NOT use an unprotected `ACK` frame to acknowledge data that was
-protected by 0-RTT or 1-RTT keys.  An endpoint MUST ignore an unprotected `ACK`
-frame if it claims to acknowledge data that was sent in a protected packet.
-Such an acknowledgement can only serve as a denial of service, since an endpoint
-that can read protected data is always able to send protected data.
+Endpoints MUST NOT use an `ACK` frame in an unprotected packet to acknowledge
+packets that were protected by 0-RTT or 1-RTT keys.  An endpoint MUST ignore an
+`ACK` frame in an unprotected packet if it claims to acknowledge data that was
+sent in a protected packet.  Such an acknowledgement can only serve as a denial
+of service, since an endpoint that can read protected data is always able to
+send protected data.
 
-ISSUE:
+Note:
 
-: What about 0-RTT data?  Should we allow acknowledgment of 0-RTT with
-  unprotected frames?  If we don't, then 0-RTT data will be unacknowledged until
-  the handshake completes.  This isn't a problem if the handshake completes
-  without loss, but it could mean that 0-RTT stalls when a handshake packet
-  disappears for any reason.
+: 0-RTT data can be acknowledged by the server as it receives it, but any
+  packets containing acknowledgments of 0-RTT data cannot have packet protection
+  removed by the client until the entire server handshake is received by the
+  client.
 
-An endpoint SHOULD use data from unprotected or 0-RTT-protected `ACK` frames
-only during the initial handshake and while they have insufficient information
-from 1-RTT-protected `ACK` frames.  Once sufficient information has been
-obtained from protected messages, information obtained from less reliable
-sources can be discarded.
+An endpoint SHOULD use data from `ACK` frames carried in unprotected or
+0-RTT-protected packets only during the initial handshake and while they have
+insufficient information from `ACK` frames in 1-RTT-protected packets.  Once
+sufficient information has been obtained from protected messages, information
+obtained from less reliable sources can be discarded.
 
 
 ### Updates to Data and Stream Limits
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1276,11 +1276,13 @@ Note:
   removed by the client until the entire server handshake is received by the
   client.
 
-An endpoint SHOULD use data from `ACK` frames carried in unprotected or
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -1262,25 +1262,25 @@ able to inject these packets.  Timing and packet retransmission information from
...
-  disappears for any reason.
-
-An endpoint SHOULD use data from unprotected or 0-RTT-protected `ACK` frames
-only during the initial handshake and while they have insufficient information
-from 1-RTT-protected `ACK` frames.  Once sufficient information has been
-obtained from protected messages, information obtained from less reliable
-sources can be discarded.
+Endpoints MUST NOT use an `ACK` frame in an unprotected packet to acknowledge
+packets that were protected by 0-RTT or 1-RTT keys.  An endpoint MUST ignore an
+`ACK` frame in an unprotected packet if it claims to acknowledge data that was
```
ianswett: So the endpoint needs to ignore the entire ack frame if any packets acknowledged were protected?

draft-ietf-quic-tls.md
```diff
@@ -1262,25 +1262,25 @@ able to inject these packets.  Timing and packet retransmission information from
...
+Note:
+
+: 0-RTT data can be acknowledged by the server as it receives it, but any
+  packets containing acknowledgments of 0-RTT data cannot have packet protection
+  removed by the client until the entire server handshake is received by the
+  client.
+
+An endpoint SHOULD use data from `ACK` frames carried in unprotected or
+0-RTT-protected packets only during the initial handshake and while they have
+insufficient information from `ACK` frames in 1-RTT-protected packets.  Once
```
ianswett: I'd like to make this stronger if possible.  How about "only before they have received an 'ACK' frame in a 1-RTT protected packet.  Once they have received an ack frame in a 1-RTT protected packet, all ack frames with less protection MUST be ignored."

draft-ietf-quic-tls.md
```diff
@@ -1262,25 +1262,25 @@ able to inject these packets.  Timing and packet retransmission information from
...
+Endpoints MUST NOT use an `ACK` frame in an unprotected packet to acknowledge
+packets that were protected by 0-RTT or 1-RTT keys.  An endpoint MUST ignore an
+`ACK` frame in an unprotected packet if it claims to acknowledge data that was
+sent in a protected packet.  Such an acknowledgement can only serve as a denial
+of service, since an endpoint that can read protected data is always able to
+send protected data.
+
+Note:
+
+: 0-RTT data can be acknowledged by the server as it receives it, but any
```
ianswett: I'm not sure how this note is helpful.  The server has to send all acks either unprotected or in 1RTT packets, correct?  If so, it seems simpler to say "0-RTT data can be acknowledged by the server as it is received, but it must be send with 1-RTT protection."?

draft-ietf-quic-tls.md
martinthomson: I'm applying principles we agreed: if it looks like trash, throw it out.  This ensures that a disruptive presence has to guess everything correctly in order for their disruption to be successful.  I'd say throw out the entire packet, but that might not be possible (other stuff might have been processed, and we don't want to ask implementations to rewind their TLS state machine, for instance).

draft-ietf-quic-tls.md
martinthomson: The point here is to highlight a problem with this requirement, not to restate the requirement to protect ACK frames.  The problem is that the server handshake might not be complete from the perspective of the client.  That leads to the client not having 1-RTT keys.  Which then leads to the client being unable to read those ACK frames.

draft-ietf-quic-tls.md
martinthomson: I think that we could be even stronger.  See revision.

draft-ietf-quic-tls.md
ianswett: Another option is to close the connection if we think this is something we want to enforce.  I think I'd prefer that, since partially consumed packets creates edge cases I don't want to consider if I don't need to.

draft-ietf-quic-tls.md
ianswett: Ah, makes sense.  How about "will not" instead of "cannot" and possibly a mention of why, such as the server sending acks with 1RTT keys and the client not having them?  I thought you were telling the client what it must do, not describing the issue.

draft-ietf-quic-tls.md
martinthomson: That would work, and would be consistent with the other principle (don't do too much about interference prior to agreeing on keys).

draft-ietf-quic-tls.md
martinthomson: "will not" implies lack of will, whereas cannot is correct - it's impossible (or a cryptographically-enforced equivalent thereof).  I will explain why though.

martinthomson:
This is no longer strictly editorial.  @janaiyengar, can you take a quick look?

PR #720, opened 2017-08-11
Add bidirectional streams on top of unidirectional

martinthomson:
This is based on the PR in #643.  It adds the concept of "related streams", which allows for pairing of streams.  It recommends against using the concept for anything else, but notes that any semantics are the responsibility of the application protocol.  This modifies the HTTP draft to prohibit use of this capability - that draft has a superior mechanism for request-response association.

I've kept this separate from #643 so that it can be assessed independently.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2641,7 +2641,7 @@ stream ID of the first in the Related Stream ID of the first stream.  This
 enables straightforward use of QUIC for request-response protocols or protocols
 that rely on bidirectional channels.
 
-This document doesn't proscribe the use of the Related Stream ID field, leaving
+This document doesn't define how to use the Related Stream ID field, leaving
 decisions about the semantics of related streams to protocols that use QUIC.
 Protocols that use QUIC MUST either define the semantics of related streams or
 prohibit the use of the Related Stream ID field.  Protocols MAY define other
@@ -2652,11 +2652,17 @@ invalid Related Stream ID field values, unless the protocol defines a more
 specific error code for the circumstances.
 
 The Related Stream ID for a stream only needs to be sent once for each stream
-(allowing for retransmissions due to loss).  Sending the field on all stream
-frames with an offset of 0 ensures that the value is received without
-duplicating the field unnecessarily.  If the value of the Related Stream ID
-changes for a stream, the endpoint that receives the conflicting value MUST
-treat this as a connection error of type INVALID_RELATED_STREAM.
+(allowing for retransmissions due to loss).  If a stream is related to another
+stream, then all stream frames with an offset of 0 MUST include the Related
+Stream ID field.  This is generally sufficient to ensure that the relationship
+between streams is correctly received.  However, the field MAY be included on
+other STREAM frames, which ensures that the stream relationship can be
+established even if the packet containing the first octets of a stream is lost.
+If the Related Stream ID is present in a STREAM frame and the value is different
+to the value included in other STREAM frames or a STREAM frame with a non-zero
+offset contains the value and a STREAM frame with offset zero does not, the
+recipient of the conflicting information MUST treat this as a connection error
+of type INVALID_RELATED_STREAM.
 
 A related stream MUST NOT refer to a stream in the "idle" state
 ({{stream-state-idle}}).  Receipt of a STREAM frame that includes a Related
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2223,12 +2223,10 @@ Offset:
 Related Stream ID:
 
 : A 32-bit value containing the stream ID of a stream in the opposite direction.
-  This field is only present if the Offset field is zero length.  Using a
+  This field is only present if the R bit in the frame type is set.  Using a
   Related Stream ID allows an application protocol to create bidirectional
-  streams by having a stream in one direction refer to another.  The identified
-  stream MUST NOT be in the "idle" state; receipt of a STREAM frame that
-  contains a Related Stream ID for an "idle" stream MUST be treated as a
-  connection error of type PROTOCOL_VIOLATION.
+  streams by having a stream in one direction refer to another (see
+  {{related-streams}}).
 
 Data Length:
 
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2618,6 +2633,39 @@ advertise a smaller maximum ID.  A sender may receive MAX_STREAM_ID frames out
...
+INVALID_RELATED_STREAM error code can be used by protocols to signal receipt of
+invalid Related Stream ID field values, unless the protocol defines a more
+specific error code for the circumstances.
+
+The Related Stream ID for a stream only needs to be sent once for each (allowing
+for retransmissions due to loss).  Sending the field on all stream frames with
+an offset of 0 ensures that the value is received without duplicating the field
+unnecessarily.  If the value of the Related Stream ID changes for a stream, the
+endpoint that receives the conflicting value MUST treat this as a connection
+error of type INVALID_RELATED_STREAM.
```
mikkelfj: First part of paragraph could be shortened a bit to just state the related stream id is only sent in the first frame with offset zero (allowing for retransmission).

Second part of the paragraph is unclear - what does it mean to receive different ID's? It can only happen during retransmission because offset 0 can only be sent once. If due to retransmission, the protocol logic should only see one version and discard duplicates without interpretation (general protocol logic, though this might be not be explicit).

The second part could also be interpreted as if the original referenced stream can only be associated with a single stream and using multiple would be an error. But I do not believe this is the intention, and if it is, I don't think it is reasonable to place any restrictions here. But - it might be good to add a note stating the a field may be referenced 0, 1, or more times depending on application semantics.

draft-ietf-quic-transport.md
```diff
@@ -2618,6 +2633,39 @@ advertise a smaller maximum ID.  A sender may receive MAX_STREAM_ID frames out
...
+
+A related stream MUST NOT refer to a stream in the "idle" state
+({{stream-state-idle}}).  Receipt of a STREAM frame that includes a Related
+Stream ID field that identifies an "idle" stream MUST be treated as a connection
+error of type INVALID_RELATED_STREAM.
+
+
 ## Sending and Receiving Data
```
mikkelfj: Isn't this redundant from earlier paragraph line 2229?

draft-ietf-quic-transport.md
mikkelfj: My initial comment was thinking about packet duplicates which is not reasonable since frames are retransmitted independently of packets. However, each steam has its own retransmission filter and here  I believe the same applies: already received content should be ignored - it is not reasonable to check all received retransmission for inconsistencies, and related stream ID would just be a special case - unless I am missing something?

draft-ietf-quic-transport.md
```diff
@@ -2215,6 +2220,16 @@ Offset:
   on a stream - the sum of the re-constructed offset and data length - MUST be
   less than 2^64.
 
+Related Stream ID:
+
+: A 32-bit value containing the stream ID of a stream in the opposite direction.
+  This field is only present if the Offset field is zero length.  Using a
```
MikeBishop: Well, it's actually only present if the `R` bit is set.  But what I think you mean is that the `R` bit MUST NOT be set if the `OO` bits aren't zero, which doesn't belong in the definition of this field.

(Side-note:  This means these three bits can encode eight states, of which only five are defined valid.)

draft-ietf-quic-transport.md
```diff
@@ -2618,6 +2633,39 @@ advertise a smaller maximum ID.  A sender may receive MAX_STREAM_ID frames out
...
+decisions about the semantics of related streams to protocols that use QUIC.
+Protocols that use QUIC MUST either define the semantics of related streams or
+prohibit the use of the Related Stream ID field.  Protocols MAY define other
+uses for the Related Stream ID field, but are advised that complex relationships
+between streams could be poorly supported in generic QUIC implementations.  The
+INVALID_RELATED_STREAM error code can be used by protocols to signal receipt of
+invalid Related Stream ID field values, unless the protocol defines a more
+specific error code for the circumstances.
+
+The Related Stream ID for a stream only needs to be sent once for each (allowing
```
MikeBishop: Probably want a noun after "for each".

draft-ietf-quic-transport.md
MikeBishop: Actually, I read this part differently than I'd initially thought you meant.  It says you don't _need_ to include it on any but the first frame, but doesn't actually prohibit it and only prescribes an error condition when there's a mismatch.  I could envision an application choosing to stick this value on one frame per packet until something gets ACK'd to make sure it arrives ASAP -- but since the remote QUIC stack can't deliver data out of order (unless #252?), that's largely moot.

draft-ietf-quic-transport.md
MikeBishop: Not only redundant, conflicts on which error code to use.

draft-ietf-quic-transport.md
mikkelfj: Ahh - I see. I then think the R bit MUST be set on offset 0 frames if the stream is related and then either only once or for all frames but not not anyhting else. I.e. leaving out the R bit is equivalent to changing the id which is an error. However, if the receiver is able to generate an error, it means it already knows the state, and then it is pointless to send it in other frames, or, it is helpful to include it in every frame but then an error cannot be generated. So overall, to reduce confusion, I would say the ID MUST only be be sent in the first frame if at all present.

mikkelfj: OTOH - I can see the benefit of including the ID on every fram if you have a many to one relation because the recevier don't need to store a lot of context. But then, as I said before, it becomes impractical to generate an error. So alternatively a) don't set the R bit at all, b) set it only on the first frame (if app procotol says so), or c) set on all frames (if app protocol says so). In all cases the use of frame ids must be consistent and receiver SHOULD generate an error if it is able to do so.

MikeBishop: I don't think the updated text explicitly says whether **NOT** setting the `R` bit on subsequent frames constitutes a change.  I think the simplest way to scope it is that if there's a related stream, `R` MUST be set on the zero-offset frame and MAY be set on other frames, but the value of the Related Stream ID MUST NOT change on subsequent frames if `R` is set.

If you receive frames out-of-order and `R` is set, you have the info early; if it's *not* set then you can't do anything with that yet because it might still have been set on some other frame.  You still have to wait for the zero-offset frame to conclude that it will never be set.

martinthomson:
After thinking about the last comment, I think that it is prudent to mandate the inclusion of Related Stream ID on STREAM frames with an offset of 0.  That way, recipients have some certainty about where to look for the information.  I don't want to end up in a situation where you have a long stream that you start delivering to an application, only to find out right at the end that it was supposed to be related to something else.

mikkelfj:
Yes, but I believe it should be 0, 1, or all frames that carry related stream id, starting at offset zero. Anything else is rather messy. I'm also a bit concerned about MUST treat this as an error, because it is either not enforceable or it implies implementation overhead that is not desirable.

Issue #61, opened 2016-12-01
Silent close


martinthomson:
I think that this is one or other peer walking away from a conversation without saying goodbye.  If we're going to talk about silent closes, we need to explain what it is, why a peer might choose to do it, and the consequences for everyone involved if they exist.

martinthomson:
It's probably timely that we have the "how do you close a connection" discussion.

@janaiyengar and I were discussing one possibility.  In that model, the transport only has an explicit message for error-related closes.  The transport also has an idle timer that starts as soon as there are no open streams (note that this means nothing to send, nothing to receive).  The connection silently closes if that timer runs down without a new stream being opened.

The application protocol could define a message that could negotiate a graceful close to trigger this.  HTTP would need this, otherwise stream 3 would hold the connection open indefinitely.  To speed things up, the application protocol could force the timer to be shortened to basically FIN_WAIT2 (a couple of RTOs) by telling the transport that it was done.  (You need to wait around a little for retransmissions and reordering.)

lucas-clemente:
Would PING frames reset this idle timer?

ianswett:
@lucas-clemente PING frames reset the idle timer today in GQUIC, and that has been useful.

@martinthomson GQUIC currently closes silently by default and does close connections if there are open streams, but only after a PING has already been sent, but not acked.  It's possible one can design something better, but I don't think closing the headers(or crypto) streams should be necessary.

martinthomson:
@ianswett, I suggested closing stream 3 because otherwise the transport has no information about whether the connection is still "in use".  That's a consequence of a logical separation of transport and HTTP.  If we wanted to avoid that, we would need to identify "static" streams.

gloinul:
From today's WG meeting: Martin concluded that he better understand that part of the issue is that Silent Close exists at all.

martinthomson:
@janaiyengar, do you think that you could take lead on writing up a proposal for this?  I think that you understand the TIME_WAIT/drain issues better than I do.  I will talk to @MikeBishop about how we might better manage graceful shutdown for HTTP.

janaiyengar:
We should plan to discuss this in Seattle. I'll spin up some text and slides before then so we can kick off discussion on the list.

martinthomson:
With #721, I believe that we have the ground work in place.  It's not perfect, but we should use new issues to discuss any additional changes.  (There are already a couple of open issues.)

PR #721, opened 2017-08-11
Refactor the section on connection termination

### LINKED ISSUE #733: Acknowledging *_CLOSE
The current text suggests that the sender of CONNECTION_CLOSE keep sending the final packet in response to subsequent incoming packets with a backoff, until you wait long enough that you figure the other side probably figured it out and stopped sending.

@nibanks points out that if you have a QUIC object that represents the connection, this means that the object and its data can't actually go away until the timeout is done.  Is there value in having a short-circuit by which the sender of *_CLOSE can be sure it was received and go away early?

The obvious solution is to suggest responding to a *_CLOSE with a matching *_CLOSE, but that just moves the timeout to the other side.  Given that servers are likely to have a persistent connection management layer, that might be acceptable, but isn't so perfect for peer-to-peer applications.

martinthomson:
This makes progress toward #61, but doesn't actually commit to a particular design.  It still needs fleshing out with details of how long to wait around and so forth, but I hope that it's a reasonable start.

Closes #733, #748, #328, #177.  Maybe more.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1425,10 +1425,11 @@ from its peer.  These packets might have been sent prior to receiving any close
 signal, or they might be retransmissions of packets for which acknowledgments
 were lost.
 
-The draining period persists for three times the maximum of the RTO or the RTT
-(see {{QUIC-RECOVERY}} for descriptions of these values).  During this period,
-new packets can be attributed to the correct connection and handled
-appropriately, but the connection is no longer considered active and usable.
+The draining period persists for three times the maximum of minimum RTO
+(kMinRTOTimeout) or the round trip time (smoothed_rtt), see {{QUIC-RECOVERY}}
+for descriptions of these values.  During this period, new packets can be
+attributed to the correct connection and acknowledged, but the connection is no
+longer considered active and usable.
 
 Different treatment is given to packets that are received while a connection is
 in the draining period depending on how the connection was closed.  In all
@@ -1452,9 +1453,9 @@ connection, after which the application requests that the connection be closed.
 A negotiated shutdown might not result in exchanging messages that are visible
 to the transport.
 
-In the draining period ({{draining}}), an endpoint that has been closed by an
-application SHOULD generate and send ACK frames as normal.  This allows the peer
-to receive acknowledgements where previous acknowledgements were lost.
+In the draining period, an endpoint that has been closed by an application
+SHOULD generate and send ACK frames as normal.  This allows the peer to receive
+acknowledgements where previous acknowledgements were lost.
 
 
 ### Idle Timeout
@@ -1478,21 +1479,21 @@ closed; open streams can be assumed to be implicitly reset.  After receiving a
 CONNECTION_CLOSE frame, endpoints MUST NOT send additional packets on that
 connection.
 
-An endpoint SHOULD respond to any packet that it receives in the draining period
-({{draining}}) with another CONNECTION_CLOSE message.  To reduce the state that
-an endpoint maintains in this case, they MAY send the exact same
-CONNECTION_CLOSE packet.  However, endpoints SHOULD limit the number of
-CONNECTION_CLOSE messages they generate.  For instance, an endpoint could
-progressively increase the number of packets that it receives before sending
-additional CONNECTION_CLOSE frames.
+An endpoint that sends a CONNECTION_CLOSE frame SHOULD respond to any packet
+that it receives in the draining period with another packet containing a
+CONNECTION_CLOSE frame.  To reduce the state that an endpoint maintains in this
+case, they MAY send the exact same packet.  However, endpoints SHOULD limit the
+number of CONNECTION_CLOSE messages they generate.  For instance, an endpoint
+could progressively increase the number of packets that it receives before
+sending additional CONNECTION_CLOSE frames.
 
 Note:
 
-: This intentionally contradicts other advice in this document that recommends
-  the creation of new packet numbers for every packet.  Sending new packet
-  numbers is primarily of advantage to loss recovery and congestion control,
-  which are not expected to be relevant for a closed connection.  Retransmitting
-  the final packet requires less state.
+: Allowing retransmision of a packet contradicts other advice in this document
+  that recommends the creation of new packet numbers for every packet.  Sending
+  new packet numbers is primarily of advantage to loss recovery and congestion
+  control, which are not expected to be relevant for a closed connection.
+  Retransmitting the final packet requires less state.
 
 
 ### Stateless Reset {#stateless-reset}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1415,31 +1415,59 @@ TODO: see issue #161
 Connections should remain open until they become idle for a pre-negotiated
 period of time.  A QUIC connection, once established, can be terminated in one
 of four ways: negotiated shutdown, idle timeout, immediate close, and a
-connection reset.
+stateless reset.
 
 
-### Negotiated Shutdown
+### Draining Period {#draining}
 
-An application protocol might arrange to abandon a connection after negotiating
-a graceful shutdown.  The application protocol exchanges whatever messages that
-are needed to cause both endpoints to agree to close the connection, after which
-the connection is closed.  A negotiated shutdown might not result in exchanging
-messages that are visible to the transport.
+After a connection is closed for any reason, an endpoint might receive packets
+from its peer.  These packets might have been sent prior to receiving any close
+signal, or they might be retransmissions of packets for which acknowledgments
+were lost.
+
+The draining period persists for three times the maximum of the RTO or the RTT
+(see {{QUIC-RECOVERY}} for descriptions of these values).  During this period,
+new packets can be attributed to the correct connection and handled
+appropriately, but the connection is no longer considered active and usable.
+
+Different treatment is given to packets that are received while a connection is
+in the draining period depending on how the connection was closed.  In all
+cases, it is possible to acknowledge packets that are received as normal, but
+other reactions might be preferrable depending on how the connection was closed.
+
+Once the draining period has ended, an endpoint SHOULD discard per-connection
+state.  This results in new packets on the connection being discarded.  An
+endpoint MAY use a stateless reset in response to any further incoming packets.
+
+The draining period does not apply when a stateless reset ({{stateless-reset}})
+is used.
+
+
+### Application Close
+
+An application protocol can arrange to close a connection.  This might be after
+negotiating a graceful shutdown.  The application protocol exchanges whatever
+messages that are needed to cause both endpoints to agree to close the
+connection, after which the application requests that the connection be closed.
+A negotiated shutdown might not result in exchanging messages that are visible
+to the transport.
+
+In the draining period ({{draining}}), an endpoint that has been closed by an
+application SHOULD generate and send ACK frames as normal.  This allows the peer
+to receive acknowledgements where previous acknowledgements were lost.
 
 
 ### Idle Timeout
 
 A connection that remains idle for longer than the idle timeout (see
-{{transport-parameter-definitions}} becomes closed.  Either peer removes
+{{transport-parameter-definitions}}) becomes closed.  Either peer removes
 connection state if they have neither sent nor received a packet for this time.
 
 The time at which an idle timeout takes effect won't be perfectly synchronized
-on peers.  Endpoints might allow for the possibility that the remote side might
-attempt to send packets before the timeout.  In this case, an endpoint might
-choose to retain enough information to generate a packet containing
-CONNECTION_CLOSE (see {{immediate-close}}).  Endpoints MAY instead rely on
-sending Stateless Reset in response to packets that arrive after an idle
-timeout.
+on peers.  A connection enters the draining period when the idle timeout
+expires.  During this time, an endpoint that receives new packets MAY choose to
+restore the connection.  Alternatively, an endpoint that receives packets MAY
+signal the timeout using an immediate close.
 
 
 ### Immediate Close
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,65 @@ TODO: see issue #161
...
+connection reset.
+
+
+### Negotiated Shutdown
+
+An application protocol might arrange to abandon a connection after negotiating
+a graceful shutdown.  The application protocol exchanges whatever messages that
+are needed to cause both endpoints to agree to close the connection, after which
+the connection is closed.  A negotiated shutdown might not result in exchanging
+messages that are visible to the transport.
```
janaiyengar: I don't understand what QUIC mechanism this is discussing -- specifically, why is this a QUIC connection termination mechanism? It seems to suggest that the app can use its own signaling to determine when to shutdown, but what QUIC mechanism is used for shutdown?

draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,65 @@ TODO: see issue #161
...
+A connection that remains idle for longer than the idle timeout (see
+{{transport-parameter-definitions}} becomes closed.  Either peer removes
+connection state if they have neither sent nor received a packet for this time.
+
+The time at which an idle timeout takes effect won't be perfectly synchronized
+on peers.  Endpoints might allow for the possibility that the remote side might
+attempt to send packets before the timeout.  In this case, an endpoint might
+choose to retain enough information to generate a CONNECTION_CLOSE.  Endpoints
+MAY instead rely on sending Public Reset in response to packets that arrive
+after an idle timeout.
```
janaiyengar: I don't think this advice is correct or adequate -- the time can be synchronized reasonably well. I would leave this subsection as is, if it needs to change in the future anyways.

draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,65 @@ TODO: see issue #161
...
+  the creation of new packet numbers for every packet.  Sending new packet
+  numbers is primarily of advantage to loss recovery and congestion control,
+  which are not expected to be relevant for a closed connection.  Retransmitting
+  the final packet requires less state at the server.
+
+Implementations SHOULD limit the number of CONNECTION_CLOSE messages they
+generate.  For instance, an implementation could exponentially increase the
+number of packets that it receives before sending the packet containing
+CONNECTION_CLOSE.  Once enough time has passed to allow a peer to receive the
+CONNECTION_CLOSE, an endpoint SHOULD discard per-connection state and MAY
```
janaiyengar: The current text says idle_timeout period of time, since that's the amount of time after which we expect the peer to drop state anyways. Is there a reason to change that?

SImilarly, the current text recommends throttling responses, which is really important. I would argue for retaining that.

draft-ietf-quic-transport.md
martinthomson: There isn't a mechanism, but the QUIC connection goes away.

Do I need to say that explicitly?

draft-ietf-quic-transport.md
martinthomson: I disagree about synchronization.  Say that I set a timeout of 10s.  At 9.8s, one endpoint sends a packet.  That packet gets lost somehow.  Now we have a difference of opinion about when the connection started being "idle" that is almost as long as the timeout duration.

I agree that we might want to have a better or clearer description of how to synchronize these things, but I was aiming to just make it clear that there is a potential problem in synchronization and that a stateless reset is the fallback.

draft-ietf-quic-transport.md
martinthomson: If the idle timeout is 2 minutes, then that is far larger than the time you might wait to have the other side receive the close.  Unreasonably so in my view.  2xRTT is probably adequate.

As for retaining text on throttling, that's right at the start of this paragraph: 

> Implementations SHOULD limit the number of CONNECTION_CLOSE messages they generate.

martinthomson:
Notes from editors chat: make the negotiated shutdown an "application shutdown" and explain that the application tells QUIC to die and then QUIC dies.

When the application shutdown happens, then we need to talk about what happens to streams.

Tie the CONNECTION_CLOSE case to the application shutdown case: you need to hang around for TIME_WAIT during which you acknowledge packets and nothing else really.  3*max(MIN_RTO, RTT) should be OK as a conservative start.

martinthomson:
I've introduced the notion of a "draining period" after close.  This is like the TCP TIME_WAIT period, but what happens depends on how the connection was closed.  On Jana's advice, I've used 3*max(MIN_RTO, RTT), recognizing that this is probably just a starting point and we can refine the exact value later.

During the draining period, the baseline here is that the endpoint generates ACK frames, but nothing else.  That's all we need for an application close (the success case).  For an idle timeout, the text permits reviving of the connection if new packets are received.  In immediate close, the text requires that the CONNECTION_CLOSE be sent, and permits re-sending the last packet to save on state space.

@janaiyengar, can you check to see that I've captured the intent well-enough here?

draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,89 @@ TODO: see issue #161
...
+signal the timeout using an immediate close.
+
+
+### Immediate Close
+
+An endpoint sends a CONNECTION_CLOSE frame to terminate the connection
+immediately.  A CONNECTION_CLOSE causes all open streams to immediately become
+closed; open streams can be assumed to be implicitly reset.  After receiving a
+CONNECTION_CLOSE frame, endpoints MUST NOT send additional packets on that
+connection.
```
nibanks: What would be the harm in allowing the endpoint to ACK a CONNECTION_CLOSE frame/packet? It would allow for the side that initiated the close to safely, immediately end its draining period. The ACK wouldn't be required, but a useful signal if used.

draft-ietf-quic-transport.md
```diff
@@ -1415,31 +1415,59 @@ TODO: see issue #161
 Connections should remain open until they become idle for a pre-negotiated
 period of time.  A QUIC connection, once established, can be terminated in one
 of four ways: negotiated shutdown, idle timeout, immediate close, and a
```
MikeBishop: This list still refers to "negotiated shutdown" while the subsections refer to "application close."  Might want to align them.

draft-ietf-quic-transport.md
MikeBishop: This probably falls under the above clause:

> In all cases, it is possible to acknowledge packets that are received as normal, but other reactions might be preferrable depending on how the connection was closed.

However, that's in conflict with the MUST NOT in this paragraph.

draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,89 @@ TODO: see issue #161
...
+The draining period persists for three times the maximum of minimum RTO
+(kMinRTOTimeout) or the round trip time (smoothed_rtt), see {{QUIC-RECOVERY}}
+for descriptions of these values.  During this period, new packets can be
+attributed to the correct connection and acknowledged, but the connection is no
+longer considered active and usable.
+
+Different treatment is given to packets that are received while a connection is
+in the draining period depending on how the connection was closed.  In all
+cases, it is possible to acknowledge packets that are received as normal, but
+other reactions might be preferrable depending on how the connection was closed.
```
MikeBishop: "preferable", not "preferrable"

draft-ietf-quic-transport.md
martinthomson: Yeah, that's an error.  I think that the right answer here is to enter the draining period and to separate prohibit frames other than ACK, PADDING, and CONNECTION_CLOSE while in that state.  I think that addresses the concerned @nibanks has raised about acknowledging a CONNECTION_CLOSE.

martinthomson:
This is what you get for trying to clean things up: they end up being substantial.

mikkelfj:
Is there a scenario where the draining period can be truncated by sending ACK in response to CONNECTION_CLOSE and where CONNECTION_CLOSE is retransmitted if ACK is not received?

martinthomson:
If an endpoint receives an ACK for a CONNECTION_CLOSE, it still needs to handle more packets (it remains in its draining period) because of the chance that packets sent before the ACK are still in transit.  Thus, the shortcut isn't really that useful here.  Obviously the endpoint that acknowledges the CONNECTION_CLOSE won't send any more packets, but if its own packets are still in flight, then it might get a few more CONNECTION_CLOSE frames (as a result of packets sent before it acknowledged the CONNECTION_CLOSE). 

The only real trimming we might do is to have the endpoint that sends the CONNECTION_CLOSE stop sending more CONNECTION_CLOSE frames in response to receiving more packets.

draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,92 @@ TODO: see issue #161
...
-to them by sending a CONNECTION_CLOSE (which MAY just be a duplicate
-of the previous CONNECTION_CLOSE packet) but MAY also send a Public
-Reset packet.  Implementations SHOULD throttle these responses, for
-instance by exponentially backing off the number of packets which are
-received before sending a response.  After this time, implementations
-SHOULD respond to unexpected packets with a Public Reset packet.
-
-
-## Stateless Reset {#stateless-reset}
+of four ways: negotiated shutdown, idle timeout, immediate close, and a
```
MikeBishop: I think this comment got lost somewhere amongst the commits, so I'll re-add it:  You have "negotiated shutdown" on this line, but "application close" on 1449.  I think consistency would help.

draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,92 @@ TODO: see issue #161
...
+for descriptions of these values.  During this period, new packets can be
+attributed to the correct connection and acknowledged, but the connection is no
+longer considered active and usable.
+
+Different treatment is given to packets that are received while a connection is
+in the draining period depending on how the connection was closed.  In all
+cases, it is possible to acknowledge packets that are received as normal, but
+other reactions might be preferable depending on how the connection was closed.
+An endpoint that is in a draining period MUST NOT send packets other than ACK,
+PADDING, or CONNECTION_CLOSE.
```
MikeBishop: Nit:  These aren't packets, they're frames.

draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,92 @@ TODO: see issue #161
...
+
+Once the draining period has ended, an endpoint SHOULD discard per-connection
+state.  This results in new packets on the connection being discarded.  An
+endpoint MAY use a stateless reset in response to any further incoming packets.
```
MikeBishop: use => send

draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,92 @@ TODO: see issue #161
...
+restore the connection.  Alternatively, an endpoint that receives packets MAY
+signal the timeout using an immediate close.
+
+
+### Immediate Close
+
+An endpoint sends a CONNECTION_CLOSE frame to terminate the connection
+immediately.  A CONNECTION_CLOSE causes all open streams to immediately become
+closed; open streams can be assumed to be implicitly reset.  After receiving a
+CONNECTION_CLOSE frame, endpoints immediately enter a draining period.
```
MikeBishop: Endpoints presumably enter a draining period after sending a CONNECTION_CLOSE as well, no?

draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,92 @@ TODO: see issue #161
...
+
+An endpoint that sends a CONNECTION_CLOSE frame SHOULD respond to any packet
+that it receives in the draining period with another packet containing a
+CONNECTION_CLOSE frame.  To reduce the state that an endpoint maintains in this
+case, they MAY send the exact same packet.  However, endpoints SHOULD limit the
```
MikeBishop: they => it, since it's a single endpoint

martinthomson:
@janaiyengar, this really needs your review.

ronieven:
section 5.7 "A QUIC endpoint MUST NOT reuse a packet number within the same
   connection (that is, under the same cryptographic keys).  If the
   packet number for sending reaches 2^64 - 1, the sender MUST close the
   connection without sending a CONNECTION_CLOSE frame or any further
   packets;"
 1. I think the first sentence contradicts the Immediate close which allows same packet.
2. Which of the terminating cases  is described in the above text.

ronieven:
Another question for stateless reset "After the Stateless Reset Token, the endpoint pads the message with an arbitrary number of octets containing random values." is it the endpoint or the server?

mcmanus:
the endpoint is the server in this case as only the server generates
stateless reset and could pad it.

ronieven:
the text should say server since the term endpoint is client or server

draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,99 @@ TODO: see issue #161
...
+
+### Draining Period {#draining}
+
+After a connection is closed for any reason, an endpoint might receive packets
+from its peer.  These packets might have been sent prior to receiving any close
+signal, or they might be retransmissions of packets for which acknowledgments
+were lost.
+
+The draining period persists for three times the maximum of minimum RTO
+(kMinRTOTimeout) or the round trip time (smoothed_rtt), see {{QUIC-RECOVERY}}
```
janaiyengar: I'm sorry if I mis-spoke, I would simply say "three times the RTO, see {{QUIC-RECOVERY}}." The RTO value is basically a max(minRTO, f(smoothed_rtt, rtt_variance)), and 3 times the RTO value allows for two consecutive timeout periods.

draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,99 @@ TODO: see issue #161
...
+for descriptions of these values.  During this period, new packets can be
+attributed to the correct connection and acknowledged, but the connection is no
```
janaiyengar: nit: remove phrase "attributed to the correct connection and". Seems unnecessary.

draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,99 @@ TODO: see issue #161
...
+longer considered active and usable.
```
janaiyengar: nit: instead of "the connection is no longer considered active and usable", how about "no new data or retransmittable packets may be sent on the connection"? The allowable frames are elucidated below, so I'd be happy with just "no new data may be sent on the connection."

draft-ietf-quic-transport.md
```diff
@@ -1414,42 +1414,99 @@ TODO: see issue #161
...
+SHOULD respond to any subsequent packet that it receives with another packet
+containing a CONNECTION_CLOSE frame.  To reduce the state that an endpoint
+maintains in this case, it MAY send the exact same packet.  However, endpoints
+SHOULD limit the number of CONNECTION_CLOSE messages they generate.  For
+instance, an endpoint could progressively increase the number of packets that it
+receives before sending additional CONNECTION_CLOSE frames.
+
+Note:
+
+: Allowing retransmision of a packet contradicts other advice in this document
```
janaiyengar: typo: retransmision --> retransmission

draft-ietf-quic-transport.md
martinthomson: I went with the latter.  I'm trying *really* hard not to say that anything is retransmittable, packets or frames.  I want to keep reinforcing the idea that it's the *contents* of frames (the data or signal) that is retransmitted.

PR #722, opened 2017-08-11
Split error code space

### LINKED ISSUE #485: Stream cancellation by transport
Should we prohibit the transport from resetting streams?

---
@ekr asked which code a server might use to cancel streams that exceed a revised limit after 0-RTT in #479.

My initial response was CANCELLED, which I'm adding in #467.  However on reflection I think that might be a category error.  The application protocol should be the one to cancel things and we should use the HTTP-defined HTTP_REQUEST_CANCELLED instead.

The reason being twofold:

* On general principles, the transport shouldn't be cancelling streams without first consulting the application layer; what if the transport cancelled stream 3 (or maybe stream 1)?  To the transport that's not special

* Signaling the ability to retry is easier at the application layer.  At the transport layer the only way to distinguish between retryable and not is to ask whether any bytes were passed to the application layer.  At the application layer, even if bytes were delivered, the application might still be able to cancel without side effects.

Errors at the transport layer should probably be fatal to the connection as a whole.  If you look at the error codes, there are few that aren't as a result of some sort of fatal error.

That suggests a split.  Errors in TLS or the transport go in CONNECTION_CLOSE.  Errors in the application layer go in RST_STREAM.  The nice partitioning that @MikeBishop did can be thrown out and we can fold transport and TLS into the same space (or split that, I care not).

martinthomson:
This creates two orthogonal spaces for error codes.  Application error codes
can be used for both connection and stream errors and are under the control of
the application protocol.  Transport error codes are QUIC-controlled, but can
only terminate the connection.

To fix this, I had to add IANA considerations for error codes, plus a few extra
tweaks.  I think that the error code space could be narrowed to 16 bits after
this, if only to keep things sane.

Closes #485, #184.

(I've stacked this on top of #721 because I don't like managing conflicts and this touches the same general area.)


draft-ietf-quic-tls.md
```diff
@@ -1503,42 +1503,54 @@ SHOULD track redundant packets and treat excessive volumes of any non-productive
...
 
 
-# Error codes {#errors}
+# Error Codes {#errors}
 
-The portion of the QUIC error code space allocated for the crypto handshake is
-0xC0000000-0xFFFFFFFF. The following error codes are defined when TLS is used
-for the crypto handshake:
+This document defines error codes from the error code space used in
+{{QUIC-TRANSPORT}}.
```
janaiyengar: This document --> This section.
Also, editorial nit: I like the mention of "crypto handshake when TLS is used for it".

draft-ietf-quic-transport.md
```diff
@@ -1610,43 +1636,51 @@ Stream ID:
...
 
-The fields of a CONNECTION_CLOSE frame are as follows:
+The fields of a TRANSPORT_CLOSE frame are as follows:
 
 Error Code:
 
 : A 32-bit error code which indicates the reason for closing this connection.
+  TRANSPORT_CLOSE uses codes from the space defined in {{error-codes}};
+  APPLICATION_CLOSE uses codes from the application protocol error code space
+  ({{app-error-codes}}).
```
janaiyengar: Seems unnecessary to mention APPLICATION_CLOSE in the definitoion of TRANSPORT_CLOSE.

draft-ietf-quic-transport.md
```diff
@@ -2608,11 +2654,10 @@ discarded upon receipt.  This avoids potential ambiguity about which STREAM
 frames count toward flow control.
 
 Upon receipt of a STOP_SENDING frame on a stream in the "open" or "half-closed
-(remote)" states, an endpoint MUST send a RST_STREAM with an error code of
-QUIC_RECEIVED_RST.  If the STOP_SENDING frame is received on a stream that is
-already in the "half-closed (local)" or "closed" states, a RST_STREAM frame MAY
-still be sent in order to cancel retransmission of previously-sent STREAM
-frames.
+(remote)" states, an endpoint MUST send a RST_STREAM frame.  If the STOP_SENDING
```
janaiyengar: What's the proposed error code in the RST_STREAM now? If there's a requirement for QUIC to generate and send this frame, the error code has to be specified.

draft-ietf-quic-transport.md
```diff
@@ -2884,32 +2929,39 @@ frame that signals the error.  Where this specification identifies error
...
+number of retransmissions and the time over which this final packet is sent
+limits the effort expended on terminated connections.
 
-An endpoint that receives an invalid CONNECTION_CLOSE frame MUST NOT signal the
-existence of the error to its peer.
+An endpoint that chooses not to retransmit packets containing TRANSPORT_CLOSE or
+APPLICATION_CLOSE risks a peer missing the first such packet.  The only
+mechanism available to an endpoint that continues to receive data for a
+terminated connection is to use the stateless reset process
+({{stateless-reset}}).
```
janaiyengar: editorial nit: I would merge this para with the one above, since this one follows from the earlier.

draft-ietf-quic-transport.md
```diff
@@ -2884,32 +2929,39 @@ frame that signals the error.  Where this specification identifies error
...
+A TRANSPORT_CLOSE or APPLICATION_CLOSE frame could be sent in a packet that is
+lost.  An endpoint SHOULD be prepared to retransmit a packet containing either
+frame type if it receives more packets on a terminated connection.  Limiting the
+number of retransmissions and the time over which this final packet is sent
+limits the effort expended on terminated connections.
 
-An endpoint that receives an invalid CONNECTION_CLOSE frame MUST NOT signal the
-existence of the error to its peer.
+An endpoint that chooses not to retransmit packets containing TRANSPORT_CLOSE or
+APPLICATION_CLOSE risks a peer missing the first such packet.  The only
```
janaiyengar: Not just the first packet, but any packets lost in the network.

draft-ietf-quic-transport.md
```diff
@@ -2923,115 +2975,102 @@ Stream 0 is critical to the functioning of the entire connection.  If stream 0
...
-application protocol does not need to inform the transport that a stream is
-critical; it can instead generate appropriate errors in response to being
-notified that the critical stream is closed.
-
-An endpoint MAY send a RST_STREAM frame in the same packet as a CONNECTION_CLOSE
-frame.
-
-
-## Error Codes
+A stream error is always an application-layer construct.  A RST_STREAM MUST NOT
```
janaiyengar: I would drop this first sentence -- it's a bit imprecise and does not add value.

draft-ietf-quic-transport.md
```diff
@@ -2923,115 +2975,102 @@ Stream 0 is critical to the functioning of the entire connection.  If stream 0
...
+be generated by the QUIC transport layer.  Resetting a stream at the transport
+layer could cause unrecoverable loss of application protocol state.  Application
+protocols might require certain streams to be reliably delivered in order to
+guarantee consistent state between endpoints.  For this reason, all errors that
+can be detected at the transport layer result in connection errors.
```
janaiyengar: This last sentence does not belong in this section, and is unnecessary. I would drop it.

draft-ietf-quic-transport.md
```diff
@@ -2923,115 +2975,102 @@ Stream 0 is critical to the functioning of the entire connection.  If stream 0
...
 
-Error codes are 32 bits long, with the first two bits indicating the source of
-the error code:
+An endpoint that detects a stream error MAY choose to treat the error as a
+connection error and send an APPLICATION_CLOSE frame in place of RST_STREAM.
```
janaiyengar: This is confusing. What is the "endpoint"? QUIC, or HTTP/QUIC? If QUIC is no longer generating RST_STREAM, then which stream errors are detected by QUIC? I suspect the answer is that the endpoint is HTTP/QUIC, in which case I would remove this sentence from here entirely.

draft-ietf-quic-transport.md
```diff
@@ -2923,115 +2975,102 @@ Stream 0 is critical to the functioning of the entire connection.  If stream 0
...
+
+
+## Application Protocol Error Codes {#app-error-codes}
+
+Application protocol error codes are 32-bits long, but the management of
+application error codes are left to application protocols.  Application protocol
+error codes are used for the RST_STREAM ({{frame-rst-stream}}) and
+APPLICATION_CLOSE ({{frame-application-close}}) frames.
+
+Application protocols SHOULD define an error codes for indicating no error and
```
janaiyengar: typo: remove "an"

draft-ietf-quic-transport.md
```diff
@@ -2923,115 +2975,102 @@ Stream 0 is critical to the functioning of the entire connection.  If stream 0
...
   included as the last octet of the error code.  For example, an error in a
-  MAX_STREAM_ID frame would be indicated with the code (0x80000106).
+  MAX_STREAM_ID frame would be indicated with the code (0x106).
+
+See {{iana-error-codes}} for details of registering new error codes.
+
+
+## Application Protocol Error Codes {#app-error-codes}
+
+Application protocol error codes are 32-bits long, but the management of
```
janaiyengar: 16-bits

draft-ietf-quic-transport.md
```diff
@@ -1596,7 +1622,7 @@ The RST_STREAM frame is as follows:
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                        Stream ID (32)                         |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                        Error Code (32)                        |
+| App Protocol Error Code (16)  |
```
janaiyengar: This is still 32 bits in this PR, right?

draft-ietf-quic-transport.md
```diff
@@ -2923,115 +2975,102 @@ Stream 0 is critical to the functioning of the entire connection.  If stream 0
...
-critical; it can instead generate appropriate errors in response to being
-notified that the critical stream is closed.
-
-An endpoint MAY send a RST_STREAM frame in the same packet as a CONNECTION_CLOSE
-frame.
-
-
-## Error Codes
+A stream error is always an application-layer construct.  A RST_STREAM MUST NOT
+be generated by the QUIC transport layer.  Resetting a stream at the transport
```
janaiyengar: drop "transport layer". 

Also, it's odd to say QUIC MUST not generate a RST_STREAM, since it is a QUIC frame and QUIC does in fact generate it. How about "RST_STREAM MUST be instigated by the application and MUST carry an application error code."?

draft-ietf-quic-transport.md
martinthomson: Yes, I must have split this from the other PR poorly.

draft-ietf-quic-transport.md
martinthomson: I considered not doing that, but it seemed more helpful to highlight the distinction here given that I wasn't repeating the format stuff in the description of APPLICATION_CLOSE.

draft-ietf-quic-transport.md
martinthomson: Actually, the point here is that the *application* sends the RST_STREAM.  The transport can't send RST_STREAM.  The transport can't do that without risking putting the application into a bad state.  I will change this to make it clearer.

Actually, I'm going to raise this on the list.  It seems like STOP_SENDING could be made an application-layer construct and we could avoid that complexity at the transport layer entirely.  I'm trying to see how visibility at the transport layer would be of value, but I'm drawing a blank.

draft-ietf-quic-tls.md
martinthomson: The crypto handshake bit is on the very next line.  The point of separation is that the codes are consumed even if TLS isn't used.

draft-ietf-quic-transport.md
martinthomson: I'll remove this sentence, reword the remaining sentence, and merge.

mnot:
@martinthomson I think CI needs to be re-run; seems to have been a transient error from GH.

martinthomson:
Rebased it.  I've changed the config so that intermittent error won't cause failures (it's the issue sync, which isn't critical).

draft-ietf-quic-transport.md
```diff
@@ -844,7 +844,8 @@ explained in more detail as they are referenced later in the document.
 |:------------|:------------------|:----------------------------|
 | 0x00        | PADDING           | {{frame-padding}}           |
 | 0x01        | RST_STREAM        | {{frame-rst-stream}}        |
-| 0x02        | CONNECTION_CLOSE  | {{frame-connection-close}}  |
+| 0x02        | TRANSPORT_CLOSE   | {{frame-transport-close}}   |
```
janaiyengar: I prefer connection close -- a connection is  a transport notion anyways. CONNECTION_CLOSE and APPLICATION_CLOSE ought to be clear enough, IMO.

draft-ietf-quic-transport.md
```diff
@@ -1609,7 +1635,7 @@ The RST_STREAM frame is as follows:
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                        Stream ID (32)                         |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                        Error Code (32)                        |
+|                  App Protocol Error Code (32)                 |
```
janaiyengar: nit: spell out "Application"

draft-ietf-quic-transport.md
```diff
@@ -1623,43 +1649,51 @@ Stream ID:
 
 : The 32-bit Stream ID of the stream being terminated.
 
-Error code:
+App Protocol Error Code:
```
janaiyengar: ditto

draft-ietf-quic-transport.md
janaiyengar: I would at least parenthesize it then.

draft-ietf-quic-transport.md
janaiyengar: I understand, but there's an issue here. If the transport receives a STOP_SENDING and MUST send a RST_STREAM, there's nothing in here that requires the application to do something in between the receiving of the STOP_SENDING and sending of the RST_STREAM. If it's not a transport requirement to send a RST_STREAM, then the "MUST" doesn't belong here. So I would remove any mention of RST_STREAM being sent here; that's an application's prerogative. The transport does not need to ensure this.

That said, I think I'm with you on STOP_SENDING. It's an HTTP-ism, so it probably belongs in the application. I'll chime in on the list.

martinthomson: I agree.  I think that I'd rather do the move of STOP_SENDING, but I'll tweak the text here in case someone objects to the other PR.

draft-ietf-quic-http.md
```diff
@@ -1067,7 +1070,7 @@ The HTTP/2 error codes defined in Section 7 of {{!RFC7540}} map to QUIC error
 codes as follows:
 
 NO_ERROR (0x0):
-: QUIC_NO_ERROR
+: HTTP_NO_ERROR
```
MikeBishop: "...in {{http-error-codes}}." for consistency with the others.

martinthomson:
Yeah, for that though I would prefer to just merge #759.

janaiyengar:
I agree, but I don't think we've reached a resolution there yet. In the meanwhile, this PR could remove the requirement within QUIC to send a RST_STREAM in response to a STOP_SENDING, since it doesn't make sense anyways.

martinthomson:
Busywork like that is low priority, but you are welcome to make a change to the branch if you feel strongly about it.

janaiyengar:
Then let's wait for #759 to land first.

martinthomson:
@janaiyengar, #759 actually depends on this, so merging that first would be hard.  Though we could wait until we agree on that, it seems like we're not going to agree without some discussion.

In the meantime, I've looked at the text around STOP_SENDING and I think that I have addressed your concerns.  Actually, the rebase for this seems to have picked up other changes that did most of the work.  The most recent commit just shuffles things around slightly and makes a few editorial corrections.

I think that this is ready, but am happy to defer until Monday to discuss if you have any lingering reservations.

PR #723, opened 2017-08-11
Make error codes smaller

martinthomson:
With #722, the error code space is much more tightly defined.  We don't need the enormous space that we have.  A 16-bit value is adequate.


MikeBishop:
LGTM.  I did the conflict resolution just to keep our PRs clean.

martinthomson:
Ahh, I see the problem now.  @janaiyengar I will create a new PR for this one.

Issue #568, opened 2017-06-03
Cleartext integrity as version independent


mcmanus:
making the cleartext integrity checksum version independent would prevent a lot of version negotiation packets from being generated based on receipt of non-quic (and make it harder to use that as an amplification vector?)

martinthomson:
If version negotiation is amplifying, something is badly wrong.

I believe that version negotiation should only be generated for packets that exceed a certain size (because the only packet that it makes sense to send version negotiation in response to is an initial client packet and those are guaranteed* to be pretty big*).  That is, for some version of a guarantee and some version of big.  Note however that we don't have that written down (or even agreed).

As for being able to distinguish noise from QUIC, that's a good reason.  We should discuss that.

mcmanus:
re: amp - that's a versioning guarantee thing again.. a v1 initial client packet is pretty big.. but a >1 packet received by a v1 server? we have very few rules about the future.. maybe we need a few.

martinthomson:
Yes, some constraints on future flexibility seem wise, as long as we can keep it light.

martinthomson:
I think that we resolved this with invariants and #724.  @mcmanus, explain and reopen if you disagree.

PR #724, opened 2017-08-11
Define requirements for handling packets

### LINKED ISSUE #570: Limit what Version Negotiation can be sent for
In #568, I noted that Version Negotiation doesn't really need to be sent in response to anything other than a client initial packet.  Those currently have two properties that we could hook into if we wanted to fix those properties across all QUIC versions: the packet type octet, and the size.  Either or both of those could be used as a way to filter non-QUIC noise from QUIC so that servers don't generate Version Negotiation unnecessarily.

I don't think that doing this represents a particularly big commitment, but we should be careful.  The minimum size seems like it might be a constant across versions.  The packet type might be a little rough, particularly in light of techniques like that suggested in #311.

martinthomson:
We never really said how to handle incoming packets.  This adds a section on
that.  This was originally motivated by #570, where the rules for what to
generate a Version Negotiation packet for were a little unclear.

Now though, this is a far more comprehensive description of what to do with
a packet that is received.

For #570, I've chosen to limit packet size.  If the packet is too small, then
this requires the server to drop the packet.  In response to the potential for
a future version of QUIC to define a smaller initial packet (@mikkelfj raised
this point), I've included advice on what might be done to avoid having packets
discarded.

Closes #570, #523, #568.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -870,44 +870,97 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
 described in {{migration}}.  Finally a connection may be terminated by either
 endpoint, as described in {{termination}}.
 
+
+## Matching Packets to Connections {#packet-handling}
+
+Incoming packets are classified on receipt.  Packets can either be associated
+with an existing connection, be discarded, or - for servers - potentially create
+a new connection.
+
+A packet that can be identified as belonging to an existing connection is
+handled according to the current state of that connection.  Packets are
+associated with established connections using connection ID if it is present;
+this might include connection IDs that were advertised using NEW_CONNECTION_ID
+({{frame-new-connection-id}}).  Packets without connection IDs and long-form
+packets for connections that have incomplete cryptographic handshakes are
+associated using the tuple of source and destination IP addresses and ports.
+
+A packet that uses the short header form could be associated with an existing
+connection with an incomplete cryptographic handshake.  Such a packet could be a
+valid packet that has been reordered with respect to the long-form packets that
+will complete the cryptographic handshake.  These packets SHOULD be buffered in
+anticipation of the handshake completing.
+
+Buffering ensures that data is not lost, which improves performance; conversely,
+discarding these packets could create false loss signals for the congestion
+controllers.  However, limiting the number and size of buffered packets might be
+needed to prevent exposure to denial of service.
+
+For clients, all packets that cannot be associated with a connection MUST be
+discarded.  Discarded packets MAY be logged for diagnostic or security purposes.
+
+For servers, packets that aren't associated with a connection potentially create
+a new connection.  However, only packets that use the long packet header and
+packets that meet the minimum size defined for the protocol version can be
+initial packets.  A server MUST discard packets that use the short header form,
+or packets that are smaller than the smallest minimum size for any version that
+the server supports.
+
+This version of QUIC defines a minimum size for initial packets of 1200 octets.
+Versions of QUIC that define smaller minimum initial packet sizes need to be
+aware that initial packets will be discarded by servers that only support
+versions with larger minimums.  Clients that support multiple QUIC versions can
+avoid this problem by ensuring that they increase the size of their initial
+packets to the largest minimum size across all of the QUIC versions they
+support.
+
+
 ## Version Negotiation {#version-negotiation}
 
 QUIC's connection establishment begins with version negotiation, since all
 communication between the endpoints, including packet and frame formats, relies
 on the two endpoints agreeing on a version.
 
-A QUIC connection begins with a client sending a handshake packet. The details
-of the handshake mechanisms are described in {{handshake}}, but all of the
-initial packets sent from the client to the server MUST use the long header
-format and MUST specify the version of the protocol being used.
+A QUIC connection begins with a client sending a client initial packet
+({{packet-client-initial}}). The details of the handshake mechanisms are
+described in {{handshake}}, but all of the initial packets sent from the client
+to the server MUST use the long header format - which includes the version of
+the protocol being used - and they MUST be padded to at least 1200 octets.
 
-When the server receives a packet from a client with the long header format, it
-compares the client's version to the versions it supports.
-
-If the version selected by the client is not acceptable to the server, the
-server discards the incoming packet and responds with a Version Negotiation
-packet ({{packet-version}}).  This includes a list of versions that the server
-will accept.
-
-To avoid packet amplification attacks a server MUST NOT send a Version
-Negotiation packet that is larger than the packet it responds to.  It is
-anticipated that this is ample space for all QUIC versions that a single server
-might need to advertise.
-
-A server sends a Version Negotiation packet for every packet that it receives
-with an unacceptable version.  This allows a server to process packets with
-unsupported versions without retaining state.  Though either the initial client
-packet or the version negotiation packet that is sent in response could be lost,
-the client will send new packets until it successfully receives a response.
+The server receives this packet and determines whether it potentially creates a
+new connection (see {{packet-handling}}).  If the packet might generate a new
+connection, the server then checks whether it understands the version that the
+client has selected.
 
 If the packet contains a version that is acceptable to the server, the server
 proceeds with the handshake ({{handshake}}).  This commits the server to the
 version that the client selected.
 
-When the client receives a Version Negotiation packet from the server, it should
-select an acceptable protocol version.  If the server lists an acceptable
-version, the client selects that version and reattempts to create a connection
-using that version.  Though the contents of a packet might not change in
+
+### Sending Version Negotiation Packets {#send-vn}
+
+If the version selected by the client is not acceptable to the server, the
+server responds with a Version Negotiation packet ({{packet-version}}).  This
+includes a list of versions that the server will accept.
+
+A server sends a Version Negotiation packet for any packet with an unaccepable
+version if that packet could create a new connection.  This allows a server to
+process packets with unsupported versions without retaining state.  Though
+either the client initial packet or the version negotiation packet that is sent
+in response could be lost, the client will send new packets until it
+successfully receives a response.
+
+
+### Handling Version Negotiation Packets {#handle-vn}
+
+When the client receives a Version Negotiation packet, it first checks that the
+packet number and connection ID match the values it sent in a client initial
+packet.  If this check fails, the packet MUST be discarded.
+
+Once the Version Negotiation packet is determined to be valid, the client then
+selects an acceptable protocol version from the list provided by the server.
+The client then attempts to create a connection using that version.  Though the
+contents of the client initial packet that the client sends might not change in
 response to version negotiation, a client MUST increase the packet number it
 uses on every packet it sends.  Packets MUST continue to use long headers and
 MUST include the new negotiated protocol version.
@@ -918,7 +971,7 @@ which is not a Version Negotiation packet.
 
 A client MUST NOT change the version it uses unless it is in response to a
 Version Negotiation packet from the server.  Once a client receives a packet
-from the server which is not a Version Negotiation packet, it MUST ignore other
+from the server which is not a Version Negotiation packet, it MUST discard other
 Version Negotiation packets on the same connection.  Similarly, a client MUST
 ignore a Version Negotiation packet if it has already received and acted on a
 Version Negotiation packet.
@@ -928,6 +981,7 @@ version.
 
 Version negotiation uses unprotected data. The result of the negotiation MUST be
 revalidated as part of the cryptographic handshake (see {{version-validation}}).
+
 
 ### Using Reserved Versions
 
@@ -950,6 +1004,7 @@ variability in the values that a server uses.
 
 A client MAY send a packet using a reserved version number.  This can be used to
 solicit a list of supported versions from a server.
+
 
 ## Cryptographic and Transport Handshake {#handshake}
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -910,7 +910,7 @@ diagnostic or security purposes.
 For servers, packets that aren't associated with a connection potentially create
 a new connection.  However, only packets that use the long packet header and
 that are at least the minimum size defined for the protocol version can be
-initial packets.  Unless the server is buffering 0-RTT packets, a server MUST
+initial packets.  Unless the server is buffering 0-RTT packets, a server MAY
 discard packets with a short header or packets that are smaller than the
 smallest minimum size for any version that the server supports.  A server that
 discards a packet that cannot be associated with a connection MAY also generate
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -911,9 +911,10 @@ For servers, packets that aren't associated with a connection potentially create
 a new connection.  However, only packets that use the long packet header and
 that are at least the minimum size defined for the protocol version can be
 initial packets.  Unless the server is buffering 0-RTT packets, a server MUST
-discard packets that cannot be associated with a connection if they use the
-short header form, or they are smaller than the smallest minimum size for any
-version that the server supports.
+either discard a packet or generate a stateless reset ({{stateless-reset}}) if
+the packet cannot be associated with a connection if they use the short header
+form, or they are smaller than the smallest minimum size for any version that
+the server supports.
 
 This version of QUIC defines a minimum size for initial packets of 1200 octets.
 Versions of QUIC that define smaller minimum initial packet sizes need to be
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -965,8 +965,9 @@ successfully receives a response or it abandons the connection attempt.
 ### Handling Version Negotiation Packets {#handle-vn}
 
 When the client receives a Version Negotiation packet, it first checks that the
-packet number and connection ID match the values it sent in a Client Initial
-packet.  If this check fails, the packet MUST be discarded.
+packet number and connection ID match the values the client sent in a previous
+packet on the same connection.  If this check fails, the packet MUST be
+discarded.
 
 Once the Version Negotiation packet is determined to be valid, the client then
 selects an acceptable protocol version from the list provided by the server.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -923,7 +923,7 @@ QUIC's connection establishment begins with version negotiation, since all
 communication between the endpoints, including packet and frame formats, relies
 on the two endpoints agreeing on a version.
 
-A QUIC connection begins with a client sending a client initial packet
+A QUIC connection begins with a client sending a Client Initial packet
 ({{packet-client-initial}}). The details of the handshake mechanisms are
 described in {{handshake}}, but all of the initial packets sent from the client
 to the server MUST use the long header format - which includes the version of
@@ -948,7 +948,7 @@ includes a list of versions that the server will accept.
 A server sends a Version Negotiation packet for any packet with an unaccepable
 version if that packet could create a new connection.  This allows a server to
 process packets with unsupported versions without retaining state.  Though
-either the client initial packet or the version negotiation packet that is sent
+either the Client Initial packet or the version negotiation packet that is sent
 in response could be lost, the client will send new packets until it
 successfully receives a response or it abandons the connection attempt.
 
@@ -956,13 +956,13 @@ successfully receives a response or it abandons the connection attempt.
 ### Handling Version Negotiation Packets {#handle-vn}
 
 When the client receives a Version Negotiation packet, it first checks that the
-packet number and connection ID match the values it sent in a client initial
+packet number and connection ID match the values it sent in a Client Initial
 packet.  If this check fails, the packet MUST be discarded.
 
 Once the Version Negotiation packet is determined to be valid, the client then
 selects an acceptable protocol version from the list provided by the server.
 The client then attempts to create a connection using that version.  Though the
-contents of the client initial packet the client sends might not change in
+contents of the Client Initial packet the client sends might not change in
 response to version negotiation, a client MUST increase the packet number it
 uses on every packet it sends.  Packets MUST continue to use long headers and
 MUST include the new negotiated protocol version.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -910,11 +910,10 @@ diagnostic or security purposes.
 For servers, packets that aren't associated with a connection potentially create
 a new connection.  However, only packets that use the long packet header and
 that are at least the minimum size defined for the protocol version can be
-initial packets.  Unless the server is buffering 0-RTT packets, a server MAY
-discard packets with a short header or packets that are smaller than the
-smallest minimum size for any version that the server supports.  A server that
-discards a packet that cannot be associated with a connection MAY also generate
-a stateless reset ({{stateless-reset}}).
+initial packets.  A server MAY discard packets with a short header or packets
+that are smaller than the smallest minimum size for any version that the server
+supports.  A server that discards a packet that cannot be associated with a
+connection MAY also generate a stateless reset ({{stateless-reset}}).
 
 This version of QUIC defines a minimum size for initial packets of 1200 octets
 (see {{packetization}}).  Versions of QUIC that define smaller minimum initial
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -348,7 +348,7 @@ version negotiation and establishment of 1-RTT keys.  Short headers are minimal
 version-specific headers, which can be used after version negotiation and 1-RTT
 keys are established.
 
-## Long Header
+## Long Header {#long-header}
 
 ~~~~~
  0                   1                   2                   3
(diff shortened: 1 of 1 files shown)


mcmanus:
I suppose fixing the long-header-bit and the min-size is enough for closing #568 in practice.

I really appreciate this PR btw - good stuff.

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,93 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+controllers.  However, limiting the number and size of buffered packets might be
+needed to prevent exposure to denial of service.
+
+For clients, all packets that cannot be associated with a connection MUST be
+discarded.  Discarded packets MAY be logged for diagnostic or security purposes.
+
+For servers, packets that aren't associated with a connection potentially create
+a new connection.  However, only packets that use the long packet header and
+packets that meet the minimum size defined for the protocol version can be
+initial packets.  A server MUST discard packets that use the short header form,
```
ianswett: nit: "A server MUST discard un-associable packets ..." to clarify what's being discarded.

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,93 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+discarding these packets could create false loss signals for the congestion
+controllers.  However, limiting the number and size of buffered packets might be
+needed to prevent exposure to denial of service.
+
+For clients, all packets that cannot be associated with a connection MUST be
+discarded.  Discarded packets MAY be logged for diagnostic or security purposes.
+
+For servers, packets that aren't associated with a connection potentially create
+a new connection.  However, only packets that use the long packet header and
+packets that meet the minimum size defined for the protocol version can be
```
ianswett: nit: "and packets that meet" -> "and meet" for brevity

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,93 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+
+Buffering ensures that data is not lost, which improves performance; conversely,
+discarding these packets could create false loss signals for the congestion
+controllers.  However, limiting the number and size of buffered packets might be
+needed to prevent exposure to denial of service.
+
+For clients, all packets that cannot be associated with a connection MUST be
+discarded.  Discarded packets MAY be logged for diagnostic or security purposes.
+
+For servers, packets that aren't associated with a connection potentially create
```
ianswett: nit: What about "aren't associated with an existing connection"

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,93 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
-using that version.  Though the contents of a packet might not change in
+
+### Sending and Handling Version Negotiation Packets {#send-vn}
+
+If the version selected by the client is not acceptable to the server, the
+server responds with a Version Negotiation packet ({{packet-version}}).  This
+includes a list of versions that the server will accept.
+
+A server sends a Version Negotiation packet for any packet with an unaccepable
+version if that packet could create a new connection.  This allows a server to
```
ianswett: Does it make sense to specify only sending VN in response to client initial or are there other types you have in mind for "that packet could create a new connection."?

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,93 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+process packets with unsupported versions without retaining state.  Though
+either the client initial packet or the version negotiation packet that is sent
+in response could be lost, the client will send new packets until it
+successfully receives a response.
```
ianswett: At some point, the handshake might timeout.

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,93 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+
+When the client receives a Version Negotiation packet, it first checks that the
+packet number and connection ID match the values it sent in a client initial
+packet.  If this check fails, the packet MUST be discarded.
+
+Once the Version Negotiation packet is determined to be valid, the client then
+selects an acceptable protocol version from the list provided by the server.
+The client then attempts to create a connection using that version.  Though the
+contents of the client initial packet that the client sends might not change in
```
ianswett: nit: that the -> the

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,93 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+
+## Matching Packets to Connections
+
+Incoming packets are classified on receipt.  Packets can either be associated
+with an existing connection, be discarded, or - for servers - potentially create
+a new connection.
+
+A packet that can be identified as belonging to an existing connection is
+handled according to the current state of that connection.  Packets are
+associated with established connections using connection ID if it is present;
```
ianswett: nit: A switch from existing to established here.  Is that intentional?

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,93 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+this might include connection IDs that were advertised using NEW_CONNECTION_ID
+({{frame-new-connection-id}}).  Packets without connection IDs and long-form
+packets for connections that have incomplete cryptographic handshakes are
+associated using the tuple of source and destination IP addresses and ports.
+
+A packet that uses the short header form could be associated with an existing
```
ianswett: It'd be nice to just say "Short header packets" here and elsewhere, since the extra passive voice makes these passages longer and a harder to read.  That's probably a larger change than just this new section, so I can do a followup editorial change for that if you think it'll work?

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,93 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
 endpoint, as described in {{termination}}.
 
+
+## Matching Packets to Connections
+
+Incoming packets are classified on receipt.  Packets can either be associated
+with an existing connection, be discarded, or - for servers - potentially create
+a new connection.
+
+A packet that can be identified as belonging to an existing connection is
```
ekr: You say "discarded", but many discarded packets trigger behavior (VN, stateless reset)

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,93 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+For servers, packets that aren't associated with a connection potentially create
+a new connection.  However, only packets that use the long packet header and
+packets that meet the minimum size defined for the protocol version can be
+initial packets.  A server MUST discard packets that use the short header form,
+or packets that are smaller than the smallest minimum size for any version that
+the server supports.
+
+This version of QUIC defines a minimum size for initial packets of 1200 octets.
+Versions of QUIC that define smaller minimum initial packet sizes need to be
+aware that initial packets will be discarded by servers that only support
```
ekr: It's not that they will be discarded but rather that they will be *silently* discarded rather than eliciting VN

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,93 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+versions with larger minimums.  Clients that support multiple QUIC versions can
+avoid this problem by ensuring that they increase the size of their initial
+packets to the largest minimum size across all of the QUIC versions they
+support.
```
ekr: You need to provide the appropriate guidance for servers of accepting the minimum size, regardless of version.

Suppose we have versions 1, 2, 3 with minimum sizes 1200, 1201 and 1202 respectively. A client which supports (1, 2) sends a packet of size 1201. A server which supports (2, 3) and doesn't recognize 1 needs to accept [1, 1201] and send VN even though [3, 1201] would not be acceptable

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,93 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
-When the server receives a packet from a client with the long header format, it
-compares the client's version to the versions it supports.
-
-If the version selected by the client is not acceptable to the server, the
-server discards the incoming packet and responds with a Version Negotiation
-packet ({{packet-version}}).  This includes a list of versions that the server
-will accept.
+A QUIC connection begins with a client sending a client initial packet
+({{packet-client-initial}}). The details of the handshake mechanisms are
+described in {{handshake}}, but all of the initial packets sent from the client
```
ekr: s/initial/first flight/

draft-ietf-quic-transport.md
MikeBishop: I think those fall under existing (stateless reset is for connections that have reached a terminal state) or new (VN).

draft-ietf-quic-transport.md
MikeBishop: Packet types are version-specific.  If the server doesn't speak the version, it doesn't know it's a Client Initial.

ianswett: Ok, that wasn't obvious to me.  Can we clarify when to send VN then?

draft-ietf-quic-transport.md
martinthomson: they mean the same thing, is that a problem?

draft-ietf-quic-transport.md
martinthomson: I'd be ok with a PR that fixed that.  It's just that we don't have clearly established definitions that would allow for the shorter usage.

draft-ietf-quic-transport.md
martinthomson: I was trying to save words here.  I figured that this was used enough to be implicit by now.

draft-ietf-quic-transport.md
martinthomson: In what way is this text not providing that clarification?  `if (might_create_a_connection && !acceptable(version)) { send_vn() }`

draft-ietf-quic-transport.md
martinthomson: Pro tip: when reviewing something and you have a new comment, don't reply to another comment, it makes it hard to respond.

draft-ietf-quic-transport.md
martinthomson: Initial is correct

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,95 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+will complete the cryptographic handshake.  These packets SHOULD be buffered in
+anticipation of the handshake completing.
+
+Buffering ensures that data is not lost, which improves performance; conversely,
+discarding these packets could create false loss signals for the congestion
+controllers.  However, limiting the number and size of buffered packets might be
+needed to prevent exposure to denial of service.
+
+For clients, all packets that cannot be associated with an existing connection
+MUST be discarded.  Discarded packets MAY be logged for diagnostic or security
```
janaiyengar: MUST -> SHOULD, since there's no reason to require this.
Separately, if the handshake is not complete, there's no existing connection, and this statement contradicts the advice on buffering short header packets for such connections. I would suggest starting this para with "In all other cases, packets SHOULD be discarded."

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,95 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+
+A packet that can be identified as belonging to an existing connection is
+handled according to the current state of that connection.  Packets are
+associated with existing connections using connection ID if it is present; this
+might include connection IDs that were advertised using NEW_CONNECTION_ID
+({{frame-new-connection-id}}).  Packets without connection IDs and long-form
+packets for connections that have incomplete cryptographic handshakes are
+associated using the tuple of source and destination IP addresses and ports.
+
+A packet that uses the short header form could be associated with an existing
```
janaiyengar: This isn't accurate. A server could use the long header for 1-RTT encypted packets, and 0-RTT encrypted packets are long header.

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,95 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
-format and MUST specify the version of the protocol being used.
-
-When the server receives a packet from a client with the long header format, it
-compares the client's version to the versions it supports.
-
-If the version selected by the client is not acceptable to the server, the
-server discards the incoming packet and responds with a Version Negotiation
-packet ({{packet-version}}).  This includes a list of versions that the server
-will accept.
+A QUIC connection begins with a client sending a client initial packet
```
janaiyengar: capitalize Client Initial

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,95 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
 version that the client selected.
 
-When the client receives a Version Negotiation packet from the server, it should
-select an acceptable protocol version.  If the server lists an acceptable
-version, the client selects that version and reattempts to create a connection
-using that version.  Though the contents of a packet might not change in
+
+### Sending and Handling Version Negotiation Packets {#send-vn}
+
+If the version selected by the client is not acceptable to the server, the
```
janaiyengar: With the section break now, it's not obvious that this flows from the previous, especially if you start reading at this section. I would  remove the section break.

draft-ietf-quic-transport.md
janaiyengar: I think Ian's point is that it's not clear what "might_create_a_connection" means. I am not clear about this either, TBH. How does a server know that a received packet might create a connection if the version is unspoken? I think all you need is if (!acceptable(version)) { send_vn() }

draft-ietf-quic-transport.md
```diff
@@ -860,39 +860,95 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+successfully receives a response or it abandons the connection attempt.
+
+When the client receives a Version Negotiation packet, it first checks that the
+packet number and connection ID match the values it sent in a client initial
+packet.  If this check fails, the packet MUST be discarded.
+
+Once the Version Negotiation packet is determined to be valid, the client then
+selects an acceptable protocol version from the list provided by the server.
+The client then attempts to create a connection using that version.  Though the
+contents of the client initial packet the client sends might not change in
```
janaiyengar: capitalize Client Initial

draft-ietf-quic-transport.md
MikeBishop: No short packet can start a connection.  Long packets with connection IDs that are currently active inherently can't start a different connection.  So, sticking with the pseudo-code, `might_create_a_connection := !ActiveConnectionIDs.Contains(packet.connection_id) && (packet.header.IsLongFormat)`

janaiyengar: If there's an active connection with this ID, you will match the connection earlier, attempt to decrypt the packet with the stored key and fail, and drop the packet on the floor. Even if you check the version before you try to decrypt you'd send a VN packet in response.

So this boils down to `if ((flags & 0x80) && !acceptable(version)) { send_vn() }`, since whether there's an active connection or not, we still send a VN packet. In which case, I think the long header check is unnecessary.

Simply checking the version seems adequate, no?

draft-ietf-quic-transport.md
martinthomson: I have revised this so that it allows for the case of buffering.

draft-ietf-quic-transport.md
martinthomson: I like the section breaks.  In fact, so much that I would like to add another.  The text is cleanly split into "what the server does" and "what the client does", so I would like to take advantage of that.

draft-ietf-quic-transport.md
martinthomson: The text (that this should have referred to) boils down to almost exactly what you describe, but with a little more nuance.  "Might have created a connection" is defined further up (in this self-same PR), but I neglected to cross-reference it properly.

In short, the condition is expanded to include one more check: that the packet is big enough to generate a new connection on any version that the server supports.  This means that if you are doing 0-RTT, then you might get a Version Negotiation packet only in response to the Client Initial and any 0-RTT packet that is large enough.  This avoids the server sending Version Negotiation in response to smaller packets (and thereby acting as an amplifier).

draft-ietf-quic-transport.md
martinthomson: Actually, it is correct, but it wasn't clear what it was referring to.  It refers to the long-form packets and short-form headers sent by the server as it completes the handshake.  It sends a long header for the handshake, then short headers for packets that follow that.  The packets with a short header can arrive before the packet containing the handshake.

I will expand on this a little.

draft-ietf-quic-transport.md
```diff
@@ -870,44 +870,107 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+For clients, any packet that cannot be associated with an existing connection
+SHOULD be discarded if it not buffered.  Discarded packets MAY be logged for
+diagnostic or security purposes.
+
+For servers, packets that aren't associated with a connection potentially create
+a new connection.  However, only packets that use the long packet header and
+that are at least the minimum size defined for the protocol version can be
+initial packets.  Unless the server is buffering 0-RTT packets, a server MUST
+discard packets that cannot be associated with a connection if they use the
+short header form, or they are smaller than the smallest minimum size for any
```
MikeBishop: Doesn't this rule out the very case Stateless Reset is intended for?  Can't be associated with an active connection, short header, but includes the Connection ID.

draft-ietf-quic-transport.md
```diff
@@ -870,44 +870,107 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+version that the server supports.
+
+This version of QUIC defines a minimum size for initial packets of 1200 octets.
+Versions of QUIC that define smaller minimum initial packet sizes need to be
+aware that initial packets will be discarded without action by servers that only
+support versions with larger minimums.  Clients that support multiple QUIC
+versions can avoid this problem by ensuring that they increase the size of their
+initial packets to the largest minimum size across all of the QUIC versions they
```
MikeBishop: Side-note which may not require a change:  If the server and client have any overlap between versions, this will work.  If they don't have any overlap between versions, however, this potentially leads to a connection time-out rather than a VN exchange and immediate conclusion that there's no overlap.

Consider the case where the minima of the versions supported by the client are (1040,680,256), and the minima of the version supported by the server are (1280,1280,1600).  Even following this advice, the server will always ignore the client's Client Initial packets, and the client will ultimately time out.

I'm not sure we need to fix this, given that otherwise we risk spewing VN packets in response to random noise.  However, we might want to be clear that this isn't guaranteed to avoid the problem.

draft-ietf-quic-transport.md
```diff
@@ -870,44 +870,107 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
-version that the client selected.
+### Handling Version Negotiation Packets {#handle-vn}
 
-When the client receives a Version Negotiation packet from the server, it should
-select an acceptable protocol version.  If the server lists an acceptable
-version, the client selects that version and reattempts to create a connection
-using that version.  Though the contents of a packet might not change in
+When the client receives a Version Negotiation packet, it first checks that the
+packet number and connection ID match the values it sent in a Client Initial
+packet.  If this check fails, the packet MUST be discarded.
```
MikeBishop: Another potential timing issue:  Client sends Client Initial + 0-RTT data.  Client Initial is lost in transit.  Version Negotiation comes back responding to 0-RTT packets because the server only received those.  Per the current text, it MUST ignore the server's VN packets until it has retransmitted the CI and gotten *that* VN packet.  Hopefully rare, since if you're doing 0-RTT you're unlikely to use an unsupported version -- but not impossible.

Maybe we should simply say "match the values it sent in a packet on this connection"?

draft-ietf-quic-transport.md
```diff
@@ -870,44 +870,109 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+needed to prevent exposure to denial of service.
+
+For clients, any packet that cannot be associated with an existing connection
+SHOULD be discarded if it not buffered.  Discarded packets MAY be logged for
+diagnostic or security purposes.
+
+For servers, packets that aren't associated with a connection potentially create
+a new connection.  However, only packets that use the long packet header and
+that are at least the minimum size defined for the protocol version can be
+initial packets.  Unless the server is buffering 0-RTT packets, a server MUST
```
ianswett: It seems worth mentioning packet type as well here, since there are only two packet types that carry client hellos.

draft-ietf-quic-transport.md
```diff
@@ -870,44 +870,109 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+either discard a packet or generate a stateless reset ({{stateless-reset}}) if
+the packet cannot be associated with a connection if they use the short header
+form, or they are smaller than the smallest minimum size for any version that
+the server supports.
+
+This version of QUIC defines a minimum size for initial packets of 1200 octets.
```
ianswett: What has to be 1200 bytes?  Is this including or excluding IP + UDP overhead?

martinthomson: That is defined elsewhere.  I'm starting to think that we need to consolidate that text, it's all over the place.  (Another patch though.)

draft-ietf-quic-transport.md
martinthomson: There is only one type of packet that could be an initial packet.  However... this allows for the possibility of defining a new version of the protocol that has more than one type.  We decided not to make the initial packet type something that is invariant across versions though.

If you want to make it an invariant (and potentially prevent new versions from defining multiple opening packet types), then it's probably best to open an issue for that.

ianswett: Fair point, I'm happy with this as is.

draft-ietf-quic-transport.md

draft-ietf-quic-transport.md
```diff
@@ -870,44 +870,109 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+valid packet that has been reordered with respect to the long-form packets that
+will complete the cryptographic handshake.  This might happen after the final
+set of cryptographic handshake messages from either peer.  These packets are
+expected to be correlated with a connection using the tuple of IP addresses and
+ports.  Packets that might be reordered in this fashion SHOULD be buffered in
+anticipation of the handshake completing.
+
+0-RTT packets might be received prior to a Client Initial packet at a server.
+If the version of these packets is acceptable to the server, it MAY buffer these
+packets in anticipation of receiving a reordered Client Initial packet.
```
janaiyengar: This is redundant and a bit confusing. The earlier para says "Such a packet could be a valid packet that has been reordered with respect to the long-form packets that will complete the cryptographic handshake", which applies to 0-RTT packets as well. I think the para above is adequate, and this one about 0-RTT is redundant.

The para above says reordered packets SHOULD be buffered, this one says MAY.  Either is fine with me, I'd lean towards a SHOULD.

draft-ietf-quic-transport.md
```diff
@@ -870,44 +870,109 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+
+Buffering ensures that data is not lost, which improves performance; conversely,
+discarding these packets could create false loss signals for the congestion
+controllers.  However, limiting the number and size of buffered packets might be
+needed to prevent exposure to denial of service.
+
+For clients, any packet that cannot be associated with an existing connection
+SHOULD be discarded if it not buffered.  Discarded packets MAY be logged for
```
janaiyengar: nit: missing "is" in "if it not buffered"

draft-ietf-quic-transport.md
janaiyengar: The server could be buffering 1-RTT packets as well (consider out-of-band key negotiation).

draft-ietf-quic-transport.md
```diff
@@ -870,44 +870,109 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+diagnostic or security purposes.
+
+For servers, packets that aren't associated with a connection potentially create
+a new connection.  However, only packets that use the long packet header and
+that are at least the minimum size defined for the protocol version can be
+initial packets.  Unless the server is buffering 0-RTT packets, a server MUST
+either discard a packet or generate a stateless reset ({{stateless-reset}}) if
+the packet cannot be associated with a connection if they use the short header
+form, or they are smaller than the smallest minimum size for any version that
+the server supports.
```
janaiyengar: This sentence needs work. It's a run-on and confusing.

draft-ietf-quic-transport.md
```diff
@@ -870,44 +870,109 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
-packet ({{packet-version}}).  This includes a list of versions that the server
-will accept.
+server responds with a Version Negotiation packet ({{packet-version}}).  This
+includes a list of versions that the server will accept.
 
-To avoid packet amplification attacks a server MUST NOT send a Version
-Negotiation packet that is larger than the packet it responds to.  It is
-anticipated that this is ample space for all QUIC versions that a single server
-might need to advertise.
+A server sends a Version Negotiation packet for any packet with an unaccepable
```
janaiyengar: typo: "unacceptable"

draft-ietf-quic-transport.md
martinthomson: 0-RTT uses the long header.  The last paragraph talks about short headers.  Though they are related, these are completely different conditions.

The requirements have to be different.  For packets that are reordered over the last flight of the handshake, you already have an assumption that the connection is going to be made and the corresponding state commitment.  So buffering isn't a huge burden.

For 0-RTT, you receive random 0-RTT packets from some unspecified endpoints prior to any handshake packets.  Buffering of those packets is risky - they might never be followed by a handshake packet.

janaiyengar:
Please remove the restriction that a server MUST discard packets unless they're 0-RTT packets before merging.

martinthomson:
I'll downgrade to MAY

draft-ietf-quic-transport.md
```diff
@@ -910,7 +910,7 @@ diagnostic or security purposes.
 For servers, packets that aren't associated with a connection potentially create
 a new connection.  However, only packets that use the long packet header and
 that are at least the minimum size defined for the protocol version can be
-initial packets.  Unless the server is buffering 0-RTT packets, a server MUST
+initial packets.  Unless the server is buffering 0-RTT packets, a server MAY
```
janaiyengar: This is an odd sentence now. Remove the phrase "Unless the server is buffering 0-RTT packets"

PR #730, opened 2017-08-14
Change design and and FIN to flow control

siyengar:
This PR has 1 editorial change and 2 design changes to flow control:

* [editorial] Clarifies the language around stream and connection flow control and makes it match the flow control section. 

* [design] Defines flow control in terms of offsets instead of max data. This means something different, because previously, maxData = 10 indicated that you could only send offset till 9, however with the change, sending maxOffset = 10 means that you can send offset until 10, i.e. 11 bytes. We could alternatively change the language of maxOffset to be MUST be LESS than instead of MUST NOT EXCEED to match maximumData.

* [design]  Makes FIN bit to be flow controllable. This would reduce one edge case where a packet scheduler does not have to do a special check for whether it only needs to send a FIN when it checks for flow control. This has the advantage of being able to define flow control on a stream in terms of offsets rather than max data.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1727,11 +1727,11 @@ reordering can mean that the largest received offset on a stream can be greater
 than the total size of data received on that stream.  Receiving STREAM frames
 might not increase the largest received offset.
 
-The largest offset sent on a stream MUST NOT exceed the largest maximum stream offset
-value advertised by the receiver.  For consistency, the FIN is also subject to this limit.
-An endpoint MUST terminate a connection with a FLOW_CONTROL_ERROR error if it receives more 
-data than the largest maximum stream data that it has sent for the affected stream, unless 
-this is a result of a change in the initial limits (see {{zerortt-parameters}}).
+The largest offset sent on a stream MUST be less than the largest maximum stream offset
+value advertised by the receiver.  An endpoint MUST terminate a connection with a
+FLOW_CONTROL_ERROR error if it receives more data than the largest maximum stream
+data that it has sent for the affected stream, unless this is a result of a change
+in the initial limits (see {{zerortt-parameters}}).
 
 
 ## MAX_STREAM_ID Frame {#frame-max-stream-id}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1680,7 +1680,8 @@ Maximum Offset:
 
 : A 64-bit unsigned integer indicating the sum of maximum offset of data that
   can be sent across all streams, in units of 1024 octets.  That is, the updated
-  connection-level data limit is determined by multiplying the encoded value by 1024.
+  connection-level data limit is determined by multiplying the encoded value by
+  1024.
 
 All data sent in STREAM frames counts toward this limit, with the exception of
 data on stream 0.  The sum of the largest received offsets on all streams -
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1660,7 +1660,7 @@ Reason Phrase:
 ## MAX_DATA Frame {#frame-max-data}
 
 The MAX_DATA frame (type=0x04) is used in flow control to inform the peer of
-the maximum amount of data that can be sent on the connection as a whole.
+the maximum offset of data that can be sent on the connection as a whole.
```
ianswett: I don't think the concept of an offset for a connection makes sense in QUIC.  It's not TCP, so what we're really talking about is the sum of all consumed stream flow controls.  So I'd suggest: "the maximum consumed stream flow control across all streams so far in the connection." or something similar?

martinthomson:
I don't think that we should take either design change.

1. MAX_STREAM_DATA of 0 should mean that you can't send any data.  This prevents that.

2. The FIN bit doesn't take any space.  Having it take space is unnecessary.  See [my email to the list on this point](https://mailarchive.ietf.org/arch/msg/quic/zxDYecCajBYnGEKdh9mNLAk1ea0).

An editorial change might have some value.

siyengar:
@martinthomson 

1. could be solved by changing the condition to be MUST be LESS than as suggested in the summary. 
2. Ya let's discuss this point on the list.

draft-ietf-quic-transport.md
siyengar: The text after this describes this in more detail.

I'm not a huge fan of "he maximum consumed stream flow control across all streams", This could be misinterpreted as 

offset = for all streams: max(stream offset)

Maybe "sum of maximum offsets of data that can be sent on all streams". However that seems a bit repetitive with the text below.

ianswett: Good point.  In that case, I'm happy with the existing text.

siyengar:
This PR should now be editorial

draft-ietf-quic-transport.md
```diff
@@ -1728,11 +1727,11 @@ reordering can mean that the largest received offset on a stream can be greater
 than the total size of data received on that stream.  Receiving STREAM frames
 might not increase the largest received offset.
 
-The data sent on a stream MUST NOT exceed the largest maximum stream data value
-advertised by the receiver.  An endpoint MUST terminate a connection with a
-FLOW_CONTROL_ERROR error if it receives more data than the largest maximum
-stream data that it has sent for the affected stream, unless this is a result of
-a change in the initial limits (see {{zerortt-parameters}}).
+The largest offset sent on a stream MUST be less than the largest maximum stream offset
```
martinthomson: So I get MAX_STREAM_DATA with Maximum Stream Offset of 10.  I then send 10 octets (0..9) in one STREAM frame.  The maximum offset is therefore 9.  The empty STREAM frame I then send has an offset of 10, which would appear to violate this rule.

siyengar: My understanding from our discussion on the mailing list is that that the intention of final offset is that it is not actually an offset, it just signals the number of bytes written so far.

draft-ietf-quic-transport.md
```diff
@@ -1728,11 +1728,11 @@ reordering can mean that the largest received offset on a stream can be greater
 than the total size of data received on that stream.  Receiving STREAM frames
 might not increase the largest received offset.
 
-The data sent on a stream MUST NOT exceed the largest maximum stream data value
-advertised by the receiver.  An endpoint MUST terminate a connection with a
-FLOW_CONTROL_ERROR error if it receives more data than the largest maximum
-stream data that it has sent for the affected stream, unless this is a result of
-a change in the initial limits (see {{zerortt-parameters}}).
+The largest offset sent on a stream MUST be less than the largest maximum
+stream offset value advertised by the receiver.  An endpoint MUST terminate a
```
marten-seemann: I find it counterintuitive to say that the largest offset that can be sent must be **less** than the maximum offset advertised.

janaiyengar:
I don't think we want to flow control the FIN, which works out since the FIN does not consume an offset. The text actually seems clear on this point... if you keep everything in terms of bytes, and not in terms of offsets, the text is consistent. Subodh: Do you still see a need for clarification?

siyengar:
@janaiyengar  Sounds fair, I didn't see anyone on the list who really spoke up that this text is a significant improvement, and people generally seem to be ambivalent, so this PR doesn't seem worth it. I could reduce this to minor changes instead which doesn't rename the fields.

martinthomson:
If you do have time to make those changes, a new PR would be best.

PR #732, opened 2017-08-15
Address validation for connection migration

### LINKED ISSUE #161: Address change and consent to send

martinthomson:
This has been much-discussed, and it's a relatively isolated change, so I did it.

This modifies PING to have an optional payload and adds a PONG frame to echo
the PING.  An empty PING generates an ACK; a PING with a payload demands a
PONG.

Generating an unguessable PING is the basis of mid-connection address
validation.  If the PING is sent on the new path, and the PONG comes back, then
the remote address is probably OK to use.

I've taken the discussion in the issue into consideration here.  There's a lot
of potential nuance to capture in terms of how an endpoint might reduce and
restore send rates, but I've done what I can to thread the gap between allowing
unbounded sending along new and untested paths and allowing connections to get
back to doing business.

It's annoying that this makes PING and PONG so disparate.  I think that we have
a re-ordering of frames in our near future to correct minor infidelities like
this.  I didn't want to do that here and pollute this PR though.

Closes #161.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1533,7 +1533,8 @@ sufficient to ensure that it is easier to receive the packet than it is to guess
 the value correctly.
 
 If the PING frame is determined to be lost, a new PING frame SHOULD be
-generated.  This frame MUST include a new payload that is difficult to guess.
+generated.  This PING frame MUST include a new Data field that is similarly
+difficult to guess.
 
 If validation of the new remote address fails, after allowing enough time for
 possible loss and recovery of packets carrying PING and PONG frames, the
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1435,8 +1435,9 @@ recover such as outstanding requests, which might otherwise be lost with no easy
 way to retry them.
 
 An endpoint that receives packets that contain a source IP address and port that
-has not yet been used MUST start sending new packets with those as a destination
-IP address and port.  Packets exchanged between endpoints now follow a new path.
+has not yet been used can start sending new packets with those as a destination
+IP address and port.  Packets exchanged between endpoints can then follow the
+new path.
 
 Due to variations in path latency or packet reordering, packets from different
 source addresses might be reordered.  The packet with the highest packet number
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1503,9 +1503,9 @@ as described in Section 5.6 of {{QUIC-TLS}}.
 
 ### Address Validation for Migrated Connections {#migrate-validate}
 
-An endpoint that see a new source IP address and port (or just a new source
-port) on packets from its peer is likely seeing a connection migration at the
-peer.
+An endpoint that receives a packet from a new remote IP address and port (or
+just a new remote port) on packets from its peer is likely seeing a connection
+migration at the peer.
 
 However, it is also possible that the peer is spoofing its source address in
 order to cause the endpoint to send excessive amounts of data to an unwilling
@@ -1515,7 +1515,7 @@ generate toward a victim.
 
 Thus, when seeing a new remote transport address, an endpoint MUST verify that
 its peer can receive and respond to packets at that new address.  By providing
-copies of the frames that it receives, the peer proves that it is receiving
+copies of the data that it receives, the peer proves that it is receiving
 packets at the new address and consents to receive data.
 
 Prior to validating the new remote address, and endpoint MUST limit the amount
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1532,6 +1532,10 @@ return.  A PING frame containing 12 randomly generated {{?RFC4086}} octets is
 sufficient to ensure that it is easier to receive the packet than it is to guess
 the value correctly.
 
+Note:
+
+: Retransmissions of the PING frame MUST also use the same remote address.
+
 If validation of the new remote address fails, after allowing enough time for
 possible loss and recovery of packets carrying PING and PONG frames, the
 endpoint MUST terminate the connection.  When setting this timer,
@@ -1558,6 +1562,49 @@ longer valid based on the changed address.
 Address validation using the PING frame MAY be used at any time by either peer.
 For instance, an endpoint might check that a peer is still in possession of its
 address after a period of quiescence.
+
+Upon seeing a connection migration, an endpoint that sees a new address MUST
+abandon any address validation it is performing with other addresses on the
+expectation that the validation is likely to fail.  Abandoning address
+validation primarily means not closing the connection when a PONG frame is not
+received, but it could also mean ceasing retransmissions of the PING frame.  An
+endpoint that doesn't retransmit a PING frame might receive a PONG frame, which
+it MUST ignore.
+
+
+## Spurious Connection Migrations
+
+A connection migration could be triggered by an attacker that is able to capture
+and forward a packet such that it arrives before the legitimate copy of that
+packet.  Such a packet will appear to be a legitimate connection migration and
+the legitimate copy will be dropped as a duplicate.
+
+After a spurious migration, validation of the source address will fail because
+the entity at the source address does not have the necessary cryptographic keys
+to read or respond to the PING frame that is sent to it, even if it wanted to.
+Such a spurious connection migration could result in the connection being
+dropped when the source address validation fails.  This grants an attacker the
+ability to terminate the connection.
+
+Receipt of packets with higher packet numbers from the legitimate address will
+trigger another connection migration.  This will cause the validation of the
+address of the spurious migration to be abandoned.
+
+To ensure that a peer sends packets from the legitimate address before the
+validation of the new address can fail, an endpoint SHOULD attempt to validate
+the old remote address before attempting to validate the new address.  If the
+connection migration is spurious, then the legitimate address will be used to
+respond and the connection will migrate back to the old address.
+
+As with any address validation, packets containing retransmissions of the PING
+frame validating an address MUST be sent to the address being validated.
+Consequently, during a migration of a peer, an endpoint could be sending to
+multiple remote addresses.
+
+An endpoint MAY abandon address validation for an address that it considers to
+be already valid.  That is, if successive connection migrations occur in quick
+succession with the final remote address being identical to the initial remote
+address, the endpoint MAY abandon address validation for that address.
 
 
 ## Connection Termination {#termination}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1532,9 +1532,8 @@ return.  A PING frame containing 12 randomly generated {{?RFC4086}} octets is
 sufficient to ensure that it is easier to receive the packet than it is to guess
 the value correctly.
 
-Note:
-
-: Retransmissions of the PING frame MUST also use the same remote address.
+If the PING frame is determined to be lost, a new PING frame SHOULD be
+generated.  This frame MUST include a new payload that is difficult to guess.
 
 If validation of the new remote address fails, after allowing enough time for
 possible loss and recovery of packets carrying PING and PONG frames, the
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1343,6 +1343,9 @@ Source address validation is therefore performed during the establishment of a
 connection.  TLS provides the tools that support the feature, but basic
 validation is performed by the core transport protocol.
 
+A different type of source address validation is performed after a connection
+migration, see {{migrate-validate}}.
+
 
 ### Client Address Validation Procedure
 
@@ -1431,6 +1434,19 @@ connection when they move networks.  This includes state that can be hard to
 recover such as outstanding requests, which might otherwise be lost with no easy
 way to retry them.
 
+An endpoint that receives packets that contain a source IP address and port that
+has not yet been used MUST start sending new packets with those as a destination
+IP address and port.  Packets exchanged between endpoints now follow a new path.
+
+Due to variations in path latency or packet reordering, packets from different
+source addresses might be reordered.  The packet with the highest packet number
+MUST be used to determine which path to use.  Endpoints also need to be prepared
+to receive packets from an older source address.
+
+An endpoint MUST validate that its peer can receive packets at the new address
+before sending any significant quantity of data to that address, or it risks
+being used for denial of service.  See {{migrate-validate}} for details.
+
 
 ### Privacy Implications of Connection Migration {#migration-linkability}
 
@@ -1485,9 +1501,63 @@ number. "packet_number_secret" is derived from the TLS key exchange,
 as described in Section 5.6 of {{QUIC-TLS}}.
 
 
-### Address Validation for Migrated Connections
+### Address Validation for Migrated Connections {#migrate-validate}
 
-TODO: see issue #161
+An endpoint that see a new source IP address and port (or just a new source
+port) on packets from its peer is likely seeing a connection migration at the
+peer.
+
+However, it is also possible that the peer is spoofing its source address in
+order to cause the endpoint to send excessive amounts of data to an unwilling
+host.  If the endpoint sends significantly more data than the peer, connection
+migration might be used to amplify the volume of data that an attacker can
+generate toward a victim.
+
+Thus, when seeing a new remote transport address, an endpoint MUST verify that
+its peer can receive and respond to packets at that new address.  By providing
+copies of the frames that it receives, the peer proves that it is receiving
+packets at the new address and consents to receive data.
+
+Prior to validating the new remote address, and endpoint MUST limit the amount
+of data and packets that it sends to its peer.  At a minimum, this needs to
+consider the possibility that packets are sent without congestion feedback.
+
+Once a connection is established, address validation is relatively simple (see
+{{address-validation}} for the process that is used during the handshake).  An
+endpoint validates a remote address by sending a PING frame containing a payload
+that is hard to guess.  This frame MUST be sent in a packet that is sent to the
+new address.  Once a PONG frame containing the same payload is received, the
+address is considered to be valid.  The PONG frame can use any path on its
+return.  A PING frame containing 12 randomly generated {{?RFC4086}} octets is
+sufficient to ensure that it is easier to receive the packet than it is to guess
+the value correctly.
+
+If validation of the new remote address fails, after allowing enough time for
+possible loss and recovery of packets carrying PING and PONG frames, the
+endpoint MUST terminate the connection.  When setting this timer,
+implementations are cautioned that the new path could have a longer round trip
+time than the original.  The endpoint MUST NOT send a CONNECTION_CLOSE frame in
+this case; it has to assume that the remote peer does not want to receive any
+more packets.
+
+If the remote address is validated successfully, the endpoint MAY increase the
+rate that it sends on the new path using the state from the previous path.  The
+capacity available on the new path might not be the same as the old path.  An
+endpoint MUST NOT restore its send rate unless it is reasonably sure that the
+path is the same as the previous path.  For instance, a change in only port
+number is likely indicative of a rebinding in a middlebox and not a complete
+change in path.  This determination likely depends on heuristics, which could be
+imperfect; if the new path capacity is significantly reduced, ultimately this
+relies on the congestion controller responding to congestion signals and reduce
+send rates appropriately.
+
+After verifying an address, the endpoint SHOULD update any address validation
+tokens ({{address-validation}}) that it has issued to its peer if those are no
+longer valid based on the changed address.
+
+Address validation using the PING frame MAY be used at any time by either peer.
+For instance, an endpoint might check that a peer is still in possession of its
+address after a period of quiescence.
 
 
 ## Connection Termination {#termination}
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ issues.json
 pulls.json
 lib
 .targets.mk
+*.upload


draft-ietf-quic-transport.md
```diff
@@ -846,6 +846,7 @@ explained in more detail as they are referenced later in the document.
 | 0x0a        | STREAM_ID_NEEDED  | {{frame-stream-id-needed}}  |
 | 0x0b        | NEW_CONNECTION_ID | {{frame-new-connection-id}} |
 | 0x0c        | STOP_SENDING      | {{frame-stop-sending}}      |
+| 0x07        | PONG              | {{frame-pong}}              |
```
marten-seemann: This should be `0x0d`.

draft-ietf-quic-transport.md
```diff
@@ -1351,6 +1355,14 @@ connection when they move networks.  This includes state that can be hard to
 recover such as outstanding requests, which might otherwise be lost with no easy
 way to retry them.
 
+An endpoint that receives packets that contain a source IP address and port that
+has not yet been used MUST start sending new packets with those values as a
+destination IP address and port.  Packets exchanged between endpoints now
+follows a new path.  An endpoint MUST validate that its peer can receive packets
```
ianswett: nit: follows -> follow

draft-ietf-quic-transport.md
```diff
@@ -1351,6 +1355,14 @@ connection when they move networks.  This includes state that can be hard to
 recover such as outstanding requests, which might otherwise be lost with no easy
 way to retry them.
 
+An endpoint that receives packets that contain a source IP address and port that
```
ianswett: This is reasonable, but I'm not sure it's what we always want.  I think we should have a special case for PING/PONG to allow proactive path validation.  If you receive only a PING, only send the PONG down that path.  If you receive other packets, then send all traffic down the new path?

draft-ietf-quic-transport.md
```diff
@@ -1405,9 +1417,63 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+host.  If the endpoint sends significantly more data than the peer, connection
+migration might be used to amplify the volume of data that an attacker can
+generate toward a victim.
+
+Thus, when seeing a new remote transport address, an endpoint MUST verify that
+its peer can receive and respond to packets at that new address.  By providing
+copies of the frames that it receives, the peer proves that it is receiving
+packets at the new address and consents to receive data.
+
+Prior to validating the new remote address, and endpoint MUST limit the rate of
```
ianswett: Nit: I think amount matters more than rate for amplification attacks.

draft-ietf-quic-transport.md
ianswett: Also, reordering can easily cause one to receive a packet from the new path, then then old path, then the new path.  I think it'd be helpful to provide some advice about sending in response to the packet with the largest packet number?

draft-ietf-quic-transport.md
```diff
@@ -1405,9 +1417,63 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+path is the same as the previous path.  For instance, a change in only port
+number is likely indicative of a rebinding in a middlebox and not a complete
+change in path.  This determination likely depends on heuristics, which could be
+imperfect; if the new path capacity is significantly reduced, ultimately this
+relies on the congestion controller responding to congestion signals and reduce
+send rates appropriately.
+
+After verifying an address, the endpoint might wish to update address validation
+tokens ({{address-validation}}) that it has issued to its peer.  Previous tokens
+might have been invalidated by the migration.
```
MikeBishop: Is this an RFC6919 MAY WISH TO?  Why not SHOULD update ... if the tokens were invalidated?

draft-ietf-quic-transport.md
martinthomson: I didn't want to accidentally add multipath here.  The preemptive thing is a fine idea, which I encourage you to put in an issue, though I suspect that we'll want to park it.

BTW, I do want to add multipath.  I'm itching to do multipath, but I hear that there are dragons and we're not done here yet.

The point about reordering is certainly valid.  The new path could be shorter, for instance.  Changes inbound.

draft-ietf-quic-transport.md
martinthomson: I think that we can limit both here.

draft-ietf-quic-transport.md
martinthomson: Well, not all tokens are invalidated.  I'll reword to be more in the form of "do X if Y".

draft-ietf-quic-transport.md
ianswett: Ok, so this is explicitly to handle unintentional port/IP migration and intentional port/IP migration without changing paths(ie: privacy preservation)?  I wouldn't call that connection migration, which I think of as explicitly moving from one path to the other, but not using both paths at once.  I'll re-read this with that in mind.

RyanAtGoogle:
It seems like sending a PING and waiting for an ACK of that packet would be roughly equivalent to sending a PING with a token and waiting to receive a PONG with that token. Would PING + ACK be sufficient? (I'm probably not thinking it through clearly) If not, can you elaborate what PING + PONG mitigates that PING + ACK does not?

MikeBishop:
Yes, since the response is allowed to be from the old IP, this is where a false ACK would cause pain.  If you were to do a packet number jump so the token were the new packet number, maybe -- but then you get into weird states where you don't want to disclose the higher packet number to the old IP, so you're maintaining parallel last-send packet numbers, and....  Ugh.

But I agree, PING with payload feels a bit weird.

draft-ietf-quic-transport.md
martinthomson: It's intended to be applicable to both.

martinthomson:
As Mike says, PING + ACK isn't sufficient.  The attack this defends against is where the attacker uses their real address to make a connection and open the congestion window, then they spoof a "migration" and cause the data that the other side is sending to be sent to a victim.  In that case, the attacker can usually guess what packet to ACK and just pretend to receive the PING.

I don't know how PING with payload is weird.  [h2 does it](http://httpwg.org/specs/rfc7540.html#PING).  And [so does ICMP](https://tools.ietf.org/html/rfc792#page-14) - though it's called "Echo" and "Echo Reply" there (which is so easy to forget).

mikkelfj:
How is it possible to spoof anything once 1RTT encryption is in place? If it is not in place, why would you want to allow connection migration when you can use stateless retry? It seems to be two separate problems.

martinthomson:
Yes, in this attack, the attacker has the keys, but not the address.

draft-ietf-quic-transport.md
```diff
@@ -1405,9 +1422,63 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+endpoint validates a remote address by sending a PING frame containing a payload
+that is hard to guess.  This frame MUST be sent in a packet that is sent to the
+new address.  Once a PONG frame containing the same payload is received, the
+address is considered to be valid.  The PONG frame can use any path on its
+return.  A PING frame containing 12 randomly generated {{?RFC4086}} octets is
+sufficient to ensure that it is easier to receive the packet than it is to guess
+the value correctly.
+
+If validation of the new remote address fails, after allowing enough time for
+possible loss and recovery of packets carrying PING and PONG frames, the
```
huitema: This opens a man-on-the-side denial of service attack on established connections. Obtain a valid packet from peer A, and race a copy to peer B from a different IP address. B receives the packet, the decryption works, B thus sends a PING to the new address, the PONG of course never comes back, and B drop the connection. Not good.

martinthomson: Firstly, this isn't new, but it suggests some more work (more on this below).

I considered this, but didn't give it enough thought.  The reason I dismissed this was that the very next packet from the victim will come from a genuine address and it will appear to be yet another migration, correcting the problem.  As long as that happens, the attacker only causes a few packets to be sent to it.  But then the next packet might take too long to arrive to be relevant.

That suggests some changes:

1. Abandon any tests on a previous address if a packet from another new address is seen.  (Use packet numbers to ensure that you can detect the most recent packet.)
2. Require that the endpoint that observes the migration ping the old address as well.

The first ensures that that the raced packet can't cause a flip to stick.  Unless the attacker can reliably capture and win a race on every packet, thereby denying service.

The second ensures that if this attack happens and the old address remains valid, a packet will be seen on the old address, causing the migration to switch back.

Does that work?

Or, we could design multipath...

huitema: I replied to your PR, Martin. But I think the key is to break the possibility for third parties to forward PING and PONG. I don't think we can use crypto to verify the source address, but we might request that packets knowingly sent from a new IP be sent with a new connection ID. Of course, there are all kinds of dragons hiding behind the word knowingly.

martinthomson:
Rebased and - I believe - ready.

draft-ietf-quic-transport.md
```diff
@@ -1425,9 +1442,110 @@ number. "packet_number_secret" is derived from the TLS key exchange,
 as described in Section 5.6 of {{QUIC-TLS}}.
 
 
-### Address Validation for Migrated Connections
+### Address Validation for Migrated Connections {#migrate-validate}
 
-TODO: see issue #161
+An endpoint that see a new source IP address and port (or just a new source
```
ianswett: see/sees

draft-ietf-quic-transport.md
```diff
@@ -1425,9 +1442,110 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+
+However, it is also possible that the peer is spoofing its source address in
+order to cause the endpoint to send excessive amounts of data to an unwilling
+host.  If the endpoint sends significantly more data than the peer, connection
+migration might be used to amplify the volume of data that an attacker can
+generate toward a victim.
+
+Thus, when seeing a new remote transport address, an endpoint MUST verify that
+its peer can receive and respond to packets at that new address.  By providing
+copies of the frames that it receives, the peer proves that it is receiving
```
ianswett: nit: Technically, they're not copies of the frames.

draft-ietf-quic-transport.md
```diff
@@ -1425,9 +1442,110 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+that is hard to guess.  This frame MUST be sent in a packet that is sent to the
+new address.  Once a PONG frame containing the same payload is received, the
+address is considered to be valid.  The PONG frame can use any path on its
+return.  A PING frame containing 12 randomly generated {{?RFC4086}} octets is
+sufficient to ensure that it is easier to receive the packet than it is to guess
+the value correctly.
+
+Note:
+
+: Retransmissions of the PING frame MUST also use the same remote address.
```
ianswett: Should we be retransmitting the same PING frame or a different PING frame?

draft-ietf-quic-transport.md
```diff
@@ -1425,9 +1442,110 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+
+If validation of the new remote address fails, after allowing enough time for
```
ianswett: What if other packets continue to arrive from the old address?  Shouldn't the connection just stay on the old address?

draft-ietf-quic-transport.md
```diff
@@ -1425,9 +1442,110 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+
+After a spurious migration, validation of the source address will fail because
+the entity at the source address does not have the necessary cryptographic keys
+to read or respond to the PING frame that is sent to it, even if it wanted to.
+Such a spurious connection migration could result in the connection being
+dropped when the source address validation fails.  This grants an attacker the
+ability to terminate the connection.
+
+Receipt of packets with higher packet numbers from the legitimate address will
+trigger another connection migration.  This will cause the validation of the
```
ianswett: Thanks, this makes sense.  Possibly allude to it earlier in the text?

draft-ietf-quic-transport.md
janaiyengar: We're definitely learning that having the ability to probe is quite useful even for connection migration. I've been working on redesigning connection migration, and our experience has shown that there needs to be more nuance around connection migration than simply switching away on the first packet. I don't want normative language here yet. I would suggest, for the sake of getting this PR in, to rephrase the MUST below to be non-normative ("should start sending", not capitalized.) We can then do an issue about what to do around connection migration separately, and figure out what should be normative.

draft-ietf-quic-transport.md
janaiyengar: Yes, agreed. We don't want to terminate the connection because a new path that we want to switch to doesn't work.

draft-ietf-quic-transport.md
```diff
@@ -1425,9 +1442,110 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+If validation of the new remote address fails, after allowing enough time for
+possible loss and recovery of packets carrying PING and PONG frames, the
+endpoint MUST terminate the connection.  When setting this timer,
+implementations are cautioned that the new path could have a longer round trip
+time than the original.  The endpoint MUST NOT send a CONNECTION_CLOSE frame in
+this case; it has to assume that the remote peer does not want to receive any
+more packets.
+
+If the remote address is validated successfully, the endpoint MAY increase the
+rate that it sends on the new path using the state from the previous path.  The
```
janaiyengar: I would remove the entire paragraph here and simply replace with "If the remote address is validated successfully, the endpoint SHOULD reset the congestion controller to its initial state."

draft-ietf-quic-transport.md
```diff
@@ -1425,9 +1442,110 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+For instance, an endpoint might check that a peer is still in possession of its
+address after a period of quiescence.
+
+Upon seeing a connection migration, an endpoint that sees a new address MUST
+abandon any address validation it is performing with other addresses on the
+expectation that the validation is likely to fail.  Abandoning address
+validation primarily means not closing the connection when a PONG frame is not
+received, but it could also mean ceasing retransmissions of the PING frame.  An
+endpoint that doesn't retransmit a PING frame might receive a PONG frame, which
+it MUST ignore.
```
janaiyengar: This paragraph is unnecessary. Using PINGs on alt paths is fine, you don't want to start sending data on all of them.

draft-ietf-quic-transport.md
```diff
@@ -1841,17 +1959,37 @@ than it has sent, unless this is a result of a change in the initial limits (see
...
-A PING frame has no additional fields.
+The PING frame contains a variable-length payload.
 
-The PING frame can be used to keep a connection alive when an application or
-application protocol wishes to prevent the connection from timing out.  An
+~~~
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| Pong Length(8)|              Pong Data (*)                  ...
```
janaiyengar: Calling this data length and PING data avoids confusion.

draft-ietf-quic-transport.md
```diff
@@ -1841,17 +1959,37 @@ than it has sent, unless this is a result of a change in the initial limits (see
...
+
+Pong Data:
+
+: This variable-length field contains arbitrary data.
+
+The receiver of a PING frame with a Pong Data field that is empty simply needs
+to acknowledge the packet containing this frame.
+
+If the Pong Data is not empty, the recipient of this frame MUST generate a PONG
+frame ({{frame-pong}}) containing the same Pong Data.
```
janaiyengar: It's easiest if the PING also generates an ack. Separately, the PONG packet will require an ack.

draft-ietf-quic-transport.md
```diff
@@ -1984,6 +2122,18 @@ Stream ID:
...
+
+## PONG Frame {#frame-pong}
+
+The PONG frame (type=0x0d) is sent in response to a PING frame that contains
+data.  Its format is identical to the PING frame ({{frame-ping}}).
+
+An endpoint that receives an unsolicited PONG frame - that is, a PONG frame
+containing a payload that is either empty or not identical to the content of a
+PING frame that the endpoint previously sent - MUST generate a connection error
+of type UNSOLICITED_PONG.
```
janaiyengar: A PONG frame needs to be acked.

janaiyengar:
BTW, thanks for all the text -- it's great to have this to kick off discussion.

draft-ietf-quic-transport.md
martinthomson: That wasn't the conclusion of the discussion on #161.  FWIW, I would prefer that, but I am trying to capture the outcome of the discussion faithfully.

draft-ietf-quic-transport.md
martinthomson: The logic here is such that if the old address continues to be used (and this is not just due to old packets being delayed, because it checks that the packet number increases) then the check on the new path is abandoned and the old path is OK.

draft-ietf-quic-transport.md
martinthomson: Correct.  That's the default.

draft-ietf-quic-transport.md
martinthomson: I removed all qualifiers on the names.

draft-ietf-quic-transport.md
martinthomson: How far up?

draft-ietf-quic-transport.md
martinthomson: ICE sends a new random number, we should too.

draft-ietf-quic-transport.md
martinthomson: It sounds like you have a design in mind.  Removing this paragraph breaks the guarantee from above.

I can't write about a design that I don't know about.  Are you describing a case where each endpoint is able to probe paths, but not cause the peer to use those paths until it decides that the path is OK?  (ICE does something similar in its make-before-break scenario, it's a reasonable model.)

If so, you need signals so that the poor peer isn't forced to guess which of these paths is the one that is preferred.  You also need some sort of system to ensure that both peers don't get different ideas about which path is "best".  Without that you get uncontrolled path oscillations.  We don't have either of those mechanisms right now, and the only signal we have is receipt of packets.  Unless we add the better design, this design at least works, even if it is suboptimal.  It's easy enough to back out or qualify a MUST or two once we have that better design complete.

draft-ietf-quic-transport.md
janaiyengar: (What does "too woolly" mean?) We as a wg haven't decided anything about connection migration yet, so I don't think we should add any normative language until we agree on what is necessary. There's a bigger design issue question here, and I think this PR is unnecessarily rushing to resolve the design issue.

martinthomson: I agree, we don't know and we should discuss.  But this PR addresses problems with use of new paths that can't otherwise be addressed.  I agree with you that this isn't a great design, it's probably even terrible, but a good design will take time.  You aren't committing to this design though.  I would prefer to see this merged, while recognizing that it is bad in certain specific ways (I don't agree that this is so bad that the protocol will break, however).  I will even open the issue that explains the shortcoming of this design if we merge this.

draft-ietf-quic-transport.md
janaiyengar: This is about behavior after the path is validated ... conservative behavior would be to reset the congestion controller to initial state. I say SHOULD because there may be good reasons for using other state (such as state stored about the new network). I'm just suggesting a shortening that is adequate, not a different statement here. My suggestion here is to not say anything about what to do with NAT rebinding... if current practice is useful, I'm happy to help document what we observe in the wild. Otherwise, let's just recommend the conservative thing.

draft-ietf-quic-transport.md
janaiyengar: I don't think this is too bad, it's just one design that I don't think is adequate. Which is why I don't think there should be normative behavior here yet. I would like to build on this PR, and add normative behavior that _we all agree_ should be normative. At the moment I don't think what you're stating as normative is useful. You're requiring that an endpoint switch immediately, which is unnecessary.

FWIW, I've opened #880.

draft-ietf-quic-transport.md
```diff
@@ -1484,9 +1501,110 @@ number. "packet_number_secret" is derived from the TLS key exchange,
 as described in Section 5.6 of {{QUIC-TLS}}.
 
 
-### Address Validation for Migrated Connections
-
-TODO: see issue #161
+### Address Validation for Migrated Connections {#migrate-validate}
```
janaiyengar: I think this section is useful, and does capture the resolution of #161.

martinthomson:
> Mostly, I'll be fine if you remove the normative text about requiring switching to the new path, since this fundamentally prevents probing. 

I don't think that it does.  It makes probing awkward, certainly, and if you probe then expect to receive a probe in return, but you can try out other paths if you want to test them out.  The problem I have with a non-immediate switch is that I don't know how to write an endpoint if you don't switch at some point.  Do you have a proposal for what we would tell the recipient of packets to do?  It doesn't have to be great (we all agree that we're going to nuke this design at the first opportunity), but I feel like we should say *something*.

janaiyengar:
Hm, yes, you're right. So, I do have articulated rules about when to switch, but it'll take a lot more to explain, and it is the "more detailed model" of switching. So, IIUC, in this simple case, client swithces and sends packet from new IP/port, server sees packet from new address and starts validation, during and after validation completes continues sending data to new IP/port. Is this correct? This is fine with me. (FWIW, this is exactly what GQUIC used to do/does right now, and we're working through its inadequacies.)

martinthomson:
I've removed the normative and made this more descriptive than prescriptive.  I hope that gives us the wiggle room we need.

BTW, I look forward to having the more detailed rules.  I hope that they are more comprehensible than RFC 5245 though :)

janaiyengar:
Ah! Sorry I read 5246, but you said 5245. Hah! Yes, indeed :-)

martinthomson:
Heh, well, either one.  5246 is much, much better, but I think we might be able to improve on that some too.

BTW, if you were tempted to look at 5245 for clues, I would probably try the -bis document, which is nearing completion.

janaiyengar:
*shudder* I'll do that.

draft-ietf-quic-transport.md
```diff
@@ -1485,9 +1502,110 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+
+Receipt of packets with higher packet numbers from the legitimate address will
+trigger another connection migration.  This will cause the validation of the
+address of the spurious migration to be abandoned.
+
+To ensure that a peer sends packets from the legitimate address before the
+validation of the new address can fail, an endpoint SHOULD attempt to validate
+the old remote address before attempting to validate the new address.  If the
+connection migration is spurious, then the legitimate address will be used to
+respond and the connection will migrate back to the old address.
```
igorlord: We are certainly not designing multi-path here, but even today you have devices that hold multiple addresses, while migrating networks.  These devices can receive on both old and new addresses during the migration process.

You do not want to probe the old address and, upon finding that the device responds (possibly from the new address), abandon the migration.

PR #737, opened 2017-08-16
supported_versions field of transport parameters has min size of 4 (1…

mcmanus:
… version)


martinthomson:
This TLS grammar is annoying.  ASN.1 is so much better.

Issue #736, opened 2017-08-16
Clarify Acks of Acks


mcmanus:
currently section 9 says
> To avoid creating an indefinite feedback loop, an endpoint MUST NOT generate an ACK frame in response to a packet containing only ACK or PADDING frames.

There is some implementation confusion around it.

Consensus seems to be that this means the packet containing only ACK/PADDING does end up referenced in an ACK frame eventually, but the generation of the ACK frame should not be triggered by the receipt of the ACK/PADDING-only packet.

however an alternative explanation, don't include the packet number of the ACK/PADDING-only packet in any ACK frame, is possible. It would be good to be clear.

I've noticed that the phrases like "don't ack acks" contribute to the confusion. Is it true that every packet except VN and Stateless-Retry needs to eventually be acknowledged? If so - that might be a useful clarifier.

ianswett:
The consensus is correct.  The packet should be acknowledged in the next ack frame sent, but it should not cause a packet containing an ack frame to be sent.

martinthomson:
Closed by #739, though Ian's comment wasn't in the precise form that GitHub wanted.

PR #739, opened 2017-08-16
Clarify when and what to ACK

ianswett:
Closes Issue #736

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2358,8 +2358,10 @@ by STREAM or RST_STREAM frames have occurred. Once the packet has been fully
 processed, a receiver acknowledges receipt by sending one or more ACK frames
 containing the packet number of the received packet.
 
-To avoid creating an indefinite feedback loop, an endpoint MUST NOT generate an
-ACK frame in response to a packet containing only ACK or PADDING frames.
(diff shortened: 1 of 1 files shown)


martinthomson:
I tweaked it a little, but this is an improvement.  The concern that remains is that we do have a need to send ACK frames even if you only receive ACK frames, but this would still seem to prohibit it.

tatsuhiro-t:
> The endpoint MUST acknowledge packets containing only ACK or PADDING frames in the next ACK frame that it sends.

Does this mean that, when we received 2nd packet which contains only ACK, it triggers generation of ACK only packet which contains ACKs for 1st and 2nd ACK only packets?  Or should we wait for reception of non-ACK packet?

ianswett:
@martinthomson If you need an ack of an ack frame for some reason, you can simply include a PING or other small frame with the ack.

@tatsuhiro-t I think the first sentence should now make this clear? "To avoid creating an indefinite feedback loop, an endpoint MUST NOT send an ACK frame in response to a packet containing only ACK or PADDING frames, even if there are packet gaps which precede the received packet."

tatsuhiro-t:
Thank you.  So it precludes generation of ACK packet which only acks one or more ACK only packets.

PR #744, opened 2017-08-22
Keep-alive requirements

### LINKED ISSUE #729: Keep alive advice
Providing the right sort of keep-alive mechanisms is hard.  TCP failed at this.

QUIC is currently following in august footsteps.  The PING frame description has this text:

> The PING frame SHOULD be used to keep a connection alive when a stream is open. 

This is bad advice: HTTP will have a stream open always.

What this wants to say is that a connection should be kept alive based on some application-driven signal that the connection is needed.  In HTTP, this might be reduced to if you have open requests or unfulfilled server push promises.  Other protocols have different keep-alive semantics.

I propose that we do this:

1. Change the description of the PING frame to say that if the application wishes to keep the connection live, then a PING frame can be sent periodically (along multiple path if maintaining multiple paths is desired).  This ensures that middleboxes don't drop state while the connection is still actively in use.  We might also say that experience with middleboxes suggests that a number between 15 and 30 seconds ensures that the greatest number of middleboxes will retain state (we can cite REQ-5 of RFC 4787 here and note that the internet sometimes doesn't pay attention to RFCs).

2. Add to that description a requirement on application protocols.  If the protocol requires keep-alive at any time, it should specify the conditions under which a keep-alive is required *and the peer that is expected to maintain that keep-alive*.  Experience with SIP has shown that the server often knows more about the keep-alive timers, so negotiating the time can be more effective than having the client drive things.

3. Change the HTTP draft to include specific advice on when to maintain the connection and say that the client should drive this.

martinthomson:
This removes the bad advice in PING about keeping connections alive and goes
for a more nuanced description.  The transport contains generic advice and a
recommendation that application protocols provide further guidance.  The HTTP
mapping specifies that the client is the one that performs keep-alives and that
it should only do so if it is waiting on responses from the server.

Closes #729.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -848,9 +848,9 @@ QUIC connections are persistent.  All of the considerations in Section 9.1 of
 
 HTTP clients are expected to use QUIC PING frames to keep connections open.
 Servers SHOULD NOT use PING frames to keep a connection open.  A client SHOULD
-NOT use PING frames for this purpose unless there are no responses outstanding
-for requests or server pushes.  If the client is not expecting a response from
-the server, allowing an idle connection to time out (based on the idle_timeout
+NOT use PING frames for this purpose unless there are responses outstanding for
+requests or server pushes.  If the client is not expecting a response from the
+server, allowing an idle connection to time out (based on the idle_timeout
 transport parameter) is preferred over expending effort maintaining a connection
 that might not be needed.  A gateway MAY use PING to maintain connections in
 anticipation of need rather than incur the latency cost of connection
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -841,6 +841,22 @@ A server MUST treat a MAX_PUSH_ID frame payload that is other than 4 octets in
 length as a connection error of type HTTP_MALFORMED_MAX_PUSH_ID.
 
 
+# Connection Management
+
+QUIC connections are persistent.  All of the considerations in Section 9.1 of
+{{?RFC7540}} apply to the management of QUIC connections.
+
+HTTP clients are expected to use QUIC PING frames to keep connections open.
+Servers SHOULD NOT use PING frames to keep a connection open.  A client SHOULD
+NOT use PING frames for this purpose unless there are no responses outstanding
+for requests or server pushes.  If the client is not expecting a response from
+the server, allowing an idle connection to time out (based on the idle_timeout
+transport parameter) is preferred over expending effort maintaining a connection
+that might not be needed.  A gateway MAY use PING to maintain connections in
+anticipation of need rather than incur the latency cost of connection
+establishment to servers.
+
+
 # Error Handling {#errors}
 
 QUIC allows the application to abruptly terminate (reset) individual streams or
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1786,9 +1786,24 @@ than it has sent, unless this is a result of a change in the initial limits (see
...
+
+The PING frame can be used to keep a connection alive when an application or
+application protocol wishes to avoid the connection from timing out.  An
+application protocol SHOULD provide guidance about the conditions under which
+generating a PING is recommended.  This guidance SHOULD indicate whether it is
+the client or the server that is expected to send the PING.  Having both
+endpoints send PING frames without coordination can cause excessive number of
+packets and poor performance.
+
+A connection will time out if no packets are for a period longer than the time
```
nibanks: I believe you meant to say something like `if no packets are sent or received for a period longer` or `if there are no packets for a period longer`.

draft-ietf-quic-http.md
```diff
@@ -841,6 +841,22 @@ A server MUST treat a MAX_PUSH_ID frame payload that is other than 4 octets in
...
 
 
+# Connection Management
+
+QUIC connections are persistent.  All of the considerations in Section 9.1 of
+{{?RFC7540}} apply to the management of QUIC connections.
+
+HTTP clients are expected to use QUIC PING frames to keep connections open.
+Servers SHOULD NOT use PING frames to keep a connection open.  A client SHOULD
+NOT use PING frames for this purpose unless there are no responses outstanding
```
MikeBishop: I think you have one more negative here than you meant.  A client SHOULD NOT use PING frames for this purpose when there are no responses outstanding.  If you're waiting on a server response, that's exactly when it's critical to keep the connection alive and the NAT bindings fresh, so that you're sure you can get the server response when it finally arrives.  If you're not waiting on anything, that's when it's fine to let the connection idle and potentially close.

draft-ietf-quic-http.md
```diff
@@ -841,6 +841,22 @@ A server MUST treat a MAX_PUSH_ID frame payload that is other than 4 octets in
...
+for requests or server pushes.  If the client is not expecting a response from
+the server, allowing an idle connection to time out (based on the idle_timeout
+transport parameter) is preferred over expending effort maintaining a connection
+that might not be needed.  A gateway MAY use PING to maintain connections in
+anticipation of need rather than incur the latency cost of connection
+establishment to servers.
```
MikeBishop: Why does the 0-RTT argument not apply to gateways?  Why does the logic that applies to gateways not apply to clients?  I'd just generically say that letting the connection idle out is preferred, but you MAY keep a connection open if you think you'll need it again soon.

draft-ietf-quic-transport.md
```diff
@@ -1786,9 +1786,24 @@ than it has sent, unless this is a result of a change in the initial limits (see
...
 additional fields. The receiver of a PING frame simply needs to acknowledge the
-packet containing this frame. The PING frame SHOULD be used to keep a connection
-alive when a stream is open. The default is to send a PING frame after 15
-seconds of quiescence. A PING frame has no additional fields.
+packet containing this frame.
+
+A PING frame has no additional fields.
+
+The PING frame can be used to keep a connection alive when an application or
+application protocol wishes to avoid the connection from timing out.  An
```
MikeBishop: Possible dialect difference:  "avoid ... from" doesn't sound right to me.  I'd have expected to see "keep ... from" or "prevent a connection time-out."

draft-ietf-quic-transport.md
martinthomson: Whoops :)

martinthomson:
Build failure here is benign, it can be ignored.

PR #171, opened 2017-01-18
Independent states and STOP_SENDING

MikeBishop:
Fixes #165.  Key changes:

- RST_STREAM kills only the sender's direction, not both.  (Retains the existing requirement to respond with RST_STREAM with another RST_STREAM if the stream isn't already closed, which eventually leads to the same result.)
- New REQUEST_RST frame allows peer to solicit a RST_STREAM; text states that this reflects an app recv_close, and that received data is being discarded

This should solve the scenarios outlined in #165; a server can ask a client to RST_STREAM its own direction after having sent an early response.  Feedback welcome, of course -- please point out anything I've missed.

PR #750, opened 2017-08-24
RST_STREAM doesn't demand a RST_STREAM in response.

martinthomson:
We missed this in #171.


MikeBishop:
Yes, this is just a leftover.

PR #754, opened 2017-08-31
Size limit for Version Negotiation

### LINKED ISSUE #585: Version Negotiation and MTU
In Section 5.3 of -03 of the draft:

  The payload of the Version Negotiation packet is a list of 32-bit
   versions which the server supports, as shown below.

As this is an open ended number of versions out of the version number space that actually is very large. Considering that this is an packet sent when no state may exist about MTU, this needs to be limited to keep into safe levels. The general recommendation in later section is to limit full packet to below 1280 bytes. That is 1232 bytes for QUIC headers, and there are 16 bytes before this list. Thus, there is a maximum of 304 versions that fits. Not an immediate problem, but some hint to that there are a limit to N, might be valuable.

martinthomson:
As discussed, the limits are largely practical, and there should be enough
incentive to avoid inflation, but it pays to be a little more precise when
defining protocols or you can get odd downstream effects.  I've chosen to limit
the size to the CI.  That's 1200 currently (or more) and should be plenty og
space for many versions.

Closes #585.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -887,12 +887,10 @@ server discards the incoming packet and responds with a Version Negotiation
 packet ({{packet-version}}).  This includes a list of versions that the server
 will accept.
 
-There is no limit to the size of the Version Negotiation packet, other than the
-number of versions that the server supports and the estimated path MTU.  To
-avoid packet amplification attacks a server MUST NOT send a Version Negotiation
-packet that is larger than the packet it responds to.  It is anticipated that
-this is ample space for all QUIC versions that a single server might need to
-advertise.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -887,6 +887,13 @@ server discards the incoming packet and responds with a Version Negotiation
 packet ({{packet-version}}).  This includes a list of versions that the server
 will accept.
 
+There is no limit to the size of the Version Negotiation packet, other than the
+number of versions that the server supports and the estimated path MTU.  To
```
MikeBishop: Since the point of this change is to impose a limit, this first sentence seems contradictory.  I'd just drop it.

PR #755, opened 2017-09-04
There was an extra mention of Public Reset

martinthomson:
Roni noticed this, and some in the text that ekr added about it being RST-like.
That text I haven't changed on the expectation that we will merge #721.


MikeBishop:
Addressed in #721.

PR #759, opened 2017-09-04
Move STOP_SENDING to HTTP/QUIC

### LINKED ISSUE #758: STOP_SENDING is an application protocol concept
Objectively, it makes very little sense for the frame to be defined at the transport layer.  There are minor advantages to the current arrangement in terms of generality (if we can agree that the concept is properly generic), and in performance (it can be sent at the transport layer without being subject to HOLB, though this might be removed by having it sent on a new stream).  However, neither of those benefits are significant enough to justify the cost of it being out of place.

Raised as a consequence of discussion on #722. This being just my opinion.

martinthomson:
This is really an application-layer concept because it deals with the
cancellation of streams and the transport should never be doing that.

Closes #758.


draft-ietf-quic-http.md
```diff
@@ -389,6 +389,23 @@ HTTP_PUSH_ALREADY_IN_CACHE; see {{errors}}).  This asks the server not to
...
 
+## Solicited Stream Cancellation
+
+If an endpoint is no longer interested in the data being received, it MAY send a
+STOP_SENDING frame to request prompt closure of the stream in the opposite
+direction.  This typically indicates that the receiving application is no longer
+reading from the stream, but does not guarantee that incoming data will be
+ignored.
+
+Upon receipt of a STOP_SENDING frame on a stream in the "open" or "half-closed
```
MikeBishop: I wouldn't say "on a stream," since the frame is delivered on the control channel.

draft-ietf-quic-http.md
```diff
@@ -389,6 +389,23 @@ HTTP_PUSH_ALREADY_IN_CACHE; see {{errors}}).  This asks the server not to
...
+(remote)" states, an endpoint MUST send a QUIC RST_STREAM frame.  If the
+STOP_SENDING frame is received on a stream that is already in the "half-closed
+(local)" or "closed" states, a RST_STREAM frame MAY still be sent in order to
+cancel retransmission of previously-sent STREAM frames.  Sending a STOP_SENDING
```
MikeBishop: Moving this piece to the HTTP layer adds an implicit new feature to QUIC API surfaces -- the ability to RST a closed stream.

draft-ietf-quic-http.md
```diff
@@ -841,6 +858,38 @@ A server MUST treat a MAX_PUSH_ID frame payload that is other than 4 octets in
 length as a connection error of type HTTP_MALFORMED_MAX_PUSH_ID.
 
 
+## STOP_SENDING Frame {#frame-stop-sending}
+
+An endpoint may use a STOP_SENDING frame (type=0x0E) to communicate that
```
MikeBishop: non-normative "may".  Let's make this either "MAY use" or "uses".

MikeBishop:
We agreed in Seattle not to remove STOP_SENDING from the transport.  There may be a separate proposal to have an application-layer mechanism *in addition* to STOP_SENDING.

martinthomson:
We didn't really conclude that STOP_SENDING or an analogue wasn't needed at the hq layer, but we did decide that the transport needs the facility.  We need a new proposal if anything is going to change.

PR #760, opened 2017-09-05
Clarify the In Effect a new Connection language in Stateless Retry

mcmanus:
Regarding client processing of Server Stateless Retry:

Remove language of "In effect, the next cryptographic handshake message is sent on a new
connection" as its pretty much not true if for no other reason that the crypto state is not the same as a new connection.

Of particular concern here was an interop issue where the second client hello used a new connection ID because of the "new connection" language, but the server rejected the HRR cookie because the address validation within it was bound to the old connection ID which seems like a legitimate thing to do when making those tokens opaque. 2 other implementations have chosen to keep the connID constant across CI.


mcmanus:
cc @tatsuhiro-t

martinthomson:
@mcmanus, I tweaked it.  Is that OK?

mcmanus:
lgtm; thanks

PR #767, opened 2017-09-07
Fixes incorrect text and renames STREAM_ID_NEEDED

janaiyengar:
Roni pointed out that this text was incorrectly placed, which I've fixed here. Additionally, I'm also proposing a change from STREAM_ID_NEEDED to STREAM_ID_BLOCKED, keeping in line with BLOCKED and STREAM_BLOCKED.


martinthomson:
Yep, managed to mess that one up...

PR #770, opened 2017-09-11
Servers SHOULD NOT misuse the CANCELLED state

### LINKED ISSUE #757: GOAWAY review
In reviewing GOAWAY I noticed a few things.  I don't think that these are controversial, but they probably need clarification.

* The QUIC stream ID MUST be a client-initiated frame.

* The QUIC stream ID can be lower than the ID of streams that carried requests, even completed requests.  If the stream carried a request that was successfully completed, then the fact that the stream was "cancelled" by the GOAWAY doesn't retroactively cancel that request.  Critically, completed streams can't be safely retried. See also #756.

MikeBishop:
Fixes #757. Servers shouldn't cancel / send GOAWAY with a value less than a stream that started getting a response.  Clients SHOULD discard responses when servers do this, unless the response was complete.  (Blowing up seems kind of pointless when the connection is winding down anyway.)

This really feels like a case of "If it hurts, don't do it."  Suggestions on tightening the text appreciated.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -767,6 +767,8 @@ indicated in the GOAWAY frame, those requests were not and will not be
 processed.  Endpoints SHOULD reset any streams above this ID with the error code
 HTTP_REQUEST_CANCELLED.  Servers MAY also reset streams below the indicated ID
 with HTTP_REQUEST_CANCELLED if the associated requests were not processed.
+Servers MUST NOT use the HTTP_REQUEST_CANCELLED status for requests which were
+partially or fully processed.
 
 The client can treat requests cancelled by the server as though they had never
 been sent at all, thereby allowing them to be retried later on a new connection.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -767,14 +767,20 @@ indicated in the GOAWAY frame, those requests were not and will not be
 processed.  Endpoints SHOULD reset any streams above this ID with the error code
 HTTP_REQUEST_CANCELLED.  Servers MAY also reset streams below the indicated ID
 with HTTP_REQUEST_CANCELLED if the associated requests were not processed.
+Servers MUST NOT use the HTTP_REQUEST_CANCELLED status for requests which were
+partially or fully processed.
```
martinthomson: Clients can detect this by observing that it received part of a response. Do you want to recommend a reaction for clients?

MikeBishop: Following paragraph discusses what clients do with cancelled requests and this PR adds, "However, if a stream is cancelled after receiving a partial response, the response SHOULD NOT be used and the request cannot be safely retried."

Though rereading it, it might be clearer to swap the order of the last two sentences.

PR #771, opened 2017-09-11
Push during message exchanges

### LINKED ISSUE #741: Description of "HTTP Message Exchanges" should explicitly PUSH_PROMISE headers.
[HTTP Message Exchanges](https://tools.ietf.org/id/draft-ietf-quic-http-05.html#rfc.section.4.2) should have text to describe PUSH_PROMISES.  Notably the case of promises interleaved with body data frames *after* the stream's own response headers, since this was a prime motivation for reverting from the previous draft's scheme to the present one.

MikeBishop:
Fixes #741.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -237,6 +237,11 @@ response may contain zero or more header blocks containing the message headers
 of informational (1xx) HTTP responses (see {{!RFC7230}}, Section 3.2 and
 {{!RFC7231}}, Section 6.2).
 
+At any point during a response message, PUSH_PROMISE frames MAY also be present
+indicating a pushed resource related to the response. These PUSH_PROMISE frames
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -237,6 +237,11 @@ response may contain zero or more header blocks containing the message headers
 of informational (1xx) HTTP responses (see {{!RFC7230}}, Section 3.2 and
 {{!RFC7231}}, Section 6.2).
 
+At any point during a response message, PUSH_PROMISE frames MAY also be present
```
janaiyengar: nit: It might be clearer to use the word interleaved in there somewhere. Suggested rewording: "A response message MAY be interleaved with PUSH_PROMISE frames that indicate a pushed resource related to the response." Otherwise, LGTM.

PR #772, opened 2017-09-11
Larger packet number, not larger packet

janaiyengar:
Editorial nit, so submitting without review.


MikeBishop:
FWIW, I typically don't even create a PR for editorial changes like this when I don't expect to request input.

janaiyengar:
Yeah, I've directly committed to master in the past, but I just went about my usual workflow this time... probably an overkill.

PR #780, opened 2017-09-19
Add OnPacketSentCC

ianswett:
Also improve pacing text and remove pacing pseudocode.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -730,7 +730,7 @@ acked_packet from sent_packets.
      else:
        // Congestion avoidance.
        congestion_window +=
-           acked_packets.bytes / congestion_window
+         kDefaultMss * acked_packets.bytes / congestion_window
 ~~~
 
 ### On Packets Lost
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -636,7 +640,9 @@ The pacing rate is a function of the mode, the congestion window, and
 the smoothed rtt.  Specifically, the pacing rate is 2 times the
 congestion window divided by the smoothed RTT during slow start
 and 1.25 times the congestion window divided by the smoothed RTT during
-slow start.
+slow start.  In order to fairly compete with flows that are not pacing,
+it is recommended to not pace the first 10 sent packets when exiting
+quiescence.
```
janaiyengar: I wonder if this entire section should be removed and left to implementers. But this is fine for now, let's discuss that separately.

draft-ietf-quic-recovery.md
```diff
@@ -693,22 +702,35 @@ follows:
    ssthresh = infinite
 ~~~
 
+### On Packet Sent
+
+Whenever a packet is sent, and is contains non-ACK frames,
```
nibanks: Instead of `and is contains non-ACK frames,` I assume you mean `and it contains non-ACK frames,`?

ianswett:
Thanks for the comments Nick, I'll try to make sure those end up in my next PR.

PR #781, opened 2017-09-19
Allow RST_STREAM to be sent and received after FIN

### LINKED ISSUE #779: Is it permitted to send both FIN and RST
@mcmanus argues that this is useful because you might send FIN and then get tired of waiting for ACKs and that it's processed on the receiver side where first one that comes ready (RST received, FIN in-order) is processed and the other ignored. My reading of the text is that this is implicitly allowed.

janaiyengar:
Addresses #779.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2635,8 +2635,10 @@ Any frame type that mentions a stream ID can be sent in this state.
 
 A stream that is in the "half-closed (local)" state MUST NOT be used for sending
 on new STREAM frames.  Retransmission of data that has already been sent on
-STREAM frames is permitted.  An endpoint MAY also send MAX_STREAM_DATA and
-STOP_SENDING in this state.
+STREAM frames is permitted.  If the stream is canceled or reset while in this
+state, a RST_STREAM frame MAY be sent. A RST_STREAM frame sent in this state
+MUST carry the stream's final offset. An endpoint MAY also send MAX_STREAM_DATA
+and STOP_SENDING in this state.
 
 An endpoint that closes a stream MUST NOT send data beyond the final offset that
 it has chosen, see {{state-closed}} for details.
@@ -2667,6 +2669,10 @@ even after the stream is closed for sending.  Frames received after a peer
 closes a stream SHOULD be discarded.  An endpoint MAY choose to limit the period
 over which it ignores frames and treat frames that arrive after this time as
 being in error.
+
+An endpoint may receive a RST_STREAM after a FIN if the peer resets the stream
+after sending a FIN on it. In this case, the endpoint MAY discard any data that
+it already received on that stream.
 
 An endpoint will know the final offset of the data it receives on a stream when
 it reaches the "half-closed (remote)" state, see {{final-offset}} for details.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2638,9 +2638,10 @@ on new STREAM frames.  Retransmission of data that has already been sent on
 STREAM frames is permitted.  An endpoint MAY also send MAX_STREAM_DATA and
 STOP_SENDING in this state.
 
-If application resets a stream in this state, a RST_STREAM frame MAY be sent.
-The final offset carried in a RST_STREAM frame sent in this state MUST be the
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2648,6 +2650,10 @@ closes a stream SHOULD be discarded.  An endpoint MAY choose to limit the period
 over which it ignores frames and treat frames that arrive after this time as
 being in error.
 
+An endpoint may receive a RST_STREAM after a FIN if the peer resets the stream
+after sending a FIN on it. In this case, the endpoint MAY discard any data that
+it already received on that stream.
```
ianswett: How about adding a sentence that the endpoing MAY close the connection if a stream is reset with a different offset from the previously received fin.

draft-ietf-quic-transport.md
```diff
@@ -2615,8 +2615,10 @@ Any frame type that mentions a stream ID can be sent in this state.
 
 A stream that is in the "half-closed (local)" state MUST NOT be used for sending
 on new STREAM frames.  Retransmission of data that has already been sent on
-STREAM frames is permitted.  An endpoint MAY also send MAX_STREAM_DATA and
-STOP_SENDING in this state.
+STREAM frames is permitted.  If the stream is canceled or reset while in this
+state, a RST_STREAM frame MAY be sent. A RST_STREAM frame sent in this state
```
ekr: Are cancelled or reset different?

draft-ietf-quic-transport.md
ekr: I would think should

draft-ietf-quic-transport.md
ekr: It's probably useful to say why you would do this.

draft-ietf-quic-transport.md
```diff
@@ -2648,6 +2650,10 @@ closes a stream SHOULD be discarded.  An endpoint MAY choose to limit the period
 over which it ignores frames and treat frames that arrive after this time as
 being in error.
 
+An endpoint may receive a RST_STREAM after a FIN if the peer resets the stream
+after sending a FIN on it. In this case, the endpoint MAY discard any data that
```
ekr: I would put the "if first" so it reads "If the peer resets the stream...."

draft-ietf-quic-transport.md
ekr: So I can also send multiple RST_STREAMs?

janaiyengar: It can, but they'd be indistinguishable from a retransmission

janaiyengar:
ekr: github is acting weird, so responding to one of your comments here. You suggested SHOULD for "the endpoint MAY discard any data that it already received on that stream." I'm sticking with MAY, since that's consistent with text around what to do with a received RST_STREAM. Perhaps we should change this in both places.

MikeBishop:
This PR covers half-closed, but you can also send/receive RST_STREAM in closed to avert retransmissions.  If we're being explicit, we should cover that case as well.

janaiyengar:
Thanks, Mike -- good point! I've added some text.

MikeBishop:
👍

janaiyengar:
Since this is simply stating explicitly what is already implicitly said, I am committing this. I am not closing the issue with this PR.

janaiyengar:
Hm. I spoke too soon - I don't have any approvals on this PR yet, and I don't want to submit this with approval, so I'll wait for @martinthomson to review this.

draft-ietf-quic-transport.md
```diff
@@ -2618,6 +2618,11 @@ on new STREAM frames.  Retransmission of data that has already been sent on
 STREAM frames is permitted.  An endpoint MAY also send MAX_STREAM_DATA and
 STOP_SENDING in this state.
 
+An application with a stream in this state may no longer care about data sent on
+the stream and may choose to reset the stream. To permit this behavior, a
```
martinthomson: s/To permit this behavior, a/A/

Or just replace the first two sentences with "An application can decide to abandon a stream in this state.  An endpoint can send RST_STREAM for a stream that was closed with the FIN flag."  Keep the final offset piece.

martinthomson:
Thanks Jana.

PR #784, opened 2017-09-20
Clarify Connection ID text

igorlord:
Clarify Connection ID use for 0-RTT and 1-RTT packets.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -667,10 +667,11 @@ that are used to remove packet protection.
 Packets protected with 0-RTT keys use a type value of 0x06.  The connection ID
 field for a 0-RTT packet is selected by the client.
 
-The client can send 0-RTT packets after having received a packet from the server
-if that packet does not complete the handshake.  Even if the client receives a
-different connection ID from the server, it MUST NOT update the connection ID it
-uses for 0-RTT packets.  This enables consistent routing for all 0-RTT packets.
+The client can send 0-RTT packets after receiving a Server Cleartext packet
+({{packet-server-cleartext}}), if that packet does not complete the handshake.
+Even if the client receives a different connection ID in the Server Cleartext
+packet, it MUST continue to use the connection ID selected by the client for
+0-RTT packets, see {{connection-id}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -670,7 +670,8 @@ field for a 0-RTT packet is selected by the client.
 The client can send 0-RTT packets after having received a packet from the server
 if that packet does not complete the handshake.  Even if the client receives a
 different connection ID from the server, it MUST NOT update the connection ID it
-uses for 0-RTT packets.  This enables consistent routing for all 0-RTT packets.
+uses for 0-RTT packets.  This ensures a predictable origin of connection IDs for
+all 0-RTT packets, see {{connection-id}}.
```
martinthomson: This phrasing doesn't make sense to me.

The point is that all 0-RTT are labelled with the same connection ID, therefore they (should) end up in the same place.

igorlord: What I meant to say is that we are trying to make all packets of a particular type to always have CCID or SCID.  Mentioning "routing" here seems out of place -- network elements can use CID for whatever purpose they like, including routing.

I'll reword this.

draft-ietf-quic-transport.md
```diff
@@ -697,17 +698,15 @@ as load balancers, to locate and use it.
...
 When the server receives a Client Initial packet and decides to proceed with the
 handshake, it chooses a new value for the connection ID and sends that in a
 Server Cleartext packet.  The server MAY choose to use the value that the client
 initially selects.
 
 Once the client receives the connection ID that the server has chosen, it uses
-this for all subsequent packets that it sends, except for any 0-RTT packets,
-which all have the same connection ID.
+this for all subsequent 1-RTT packets ({{packet-protected}}) and only 1-RTT
+packets that it sends.
```
igorlord: Actually, this conflicts with # 558, which states that "Client Cleartext" also has SCID.  I'll fix that, too, and repost.

igorlord:
Proposed connection ID clarifications, consistent with #588.

PR #787, opened 2017-09-21
HTTP Integration Error Nits

mcmanus:
* error code space is 30 bits
  * error codes can be used with stop_sending
  * genericized CANCEL to indicate it can be used bidirectionally


MikeBishop:
Merged the non-obsolete bit manually.

PR #788, opened 2017-09-21
Specify connection ID and version a unsigned numbers

### LINKED ISSUE #786: Clarify Connection ID and Version number endianness
It should be obvious, but we should make it clear since endianness matters, especially when we feed connection ID into the en/decrypters.

Also, version number should be treated as a uint. We should be explicit about this as well.

RyanAtGoogle:
Closes #786


martinthomson:
I like the number bit for versions.  I'm less enthusiastic about the connection ID because we never actually treat it as a number anywhere.  That said, this is fine.

PR #807, opened 2017-09-28
Don't retransmit old MAX_STREAM_DATA and MAX_DATA

### LINKED ISSUE #806: Don't retransmit old MAX_STREAM_DATA
The current text says that one ought to retransmit MAX_STREAM_DATA:

```
   o  All application data sent in STREAM frames MUST be retransmitted,
      unless the endpoint has sent a RST_STREAM for that stream.  When
      an endpoint sends a RST_STREAM frame, data outstanding on that
      stream SHOULD NOT be retransmitted, since subsequent data on this
      stream is expected to not be delivered by the receiver.

   o  ACK and PADDING frames MUST NOT be retransmitted.  ACK frames
      containing updated information will be sent as described in
      Section 8.13.

   o  STOP_SENDING frames MUST be retransmitted, unless the stream has
      become closed in the appropriate direction.  See Section 10.3.
   o  All other frames MUST be retransmitted.
```

But this is silly. Say I transmitted MAX_STREAM_DATA=100, MAX_STREAM_DATA=200, MAX_STREAM_DATA=300 and then none of them were acknowledged. Why am I retransmitting frames which must be ignored.

janaiyengar:
Closes #806

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2420,6 +2420,11 @@ When a packet is detected as lost, the sender re-sends any frames as necessary:
 * STOP_SENDING frames MUST be retransmitted, unless the stream has become closed
   in the appropriate direction.  See {{solicited-state-transitions}}.
 
+* The most recent MAX_STREAM_DATA and MAX_DATA frames MUST be retransmitted
+  until acked. Any previous unacknowledged MAX_STREAM_DATA and MAX_DATA frames
+  SHOULD NOT be retransmitted since the newer frames obviate the need for
+  delivering the older ones.
+
 * All other frames MUST be retransmitted.
 
 Upon detecting losses, a sender MUST take appropriate congestion control action.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2420,10 +2420,11 @@ When a packet is detected as lost, the sender re-sends any frames as necessary:
 * STOP_SENDING frames MUST be retransmitted, unless the stream has become closed
   in the appropriate direction.  See {{solicited-state-transitions}}.
 
-* The most recent MAX_STREAM_DATA and MAX_DATA frames MUST be retransmitted
-  until acked. Any previous unacknowledged MAX_STREAM_DATA and MAX_DATA frames
-  SHOULD NOT be retransmitted since the newer frames obviate the need for
-  delivering the older ones.
+* The most recent MAX_STREAM_DATA frame for a stream MUST be retransmitted. Any
+  previous unacknowledged MAX_STREAM_DATA frame for the same stream SHOULD NOT
+  be retransmitted since a newer MAX_STREAM_DATA frame for a stream obviates the
+  need for delivering older ones. Similarly, the most recent MAX_DATA frame MUST
+  be retransmitted; previous unacknowledged ones SHOULD NOT be retransmitted.
 
 * All other frames MUST be retransmitted.
 


draft-ietf-quic-transport.md
```diff
@@ -2420,6 +2420,12 @@ When a packet is detected as lost, the sender re-sends any frames as necessary:
 * STOP_SENDING frames MUST be retransmitted, unless the stream has become closed
   in the appropriate direction.  See {{solicited-state-transitions}}.
 
+* The most recent MAX_STREAM_DATA frame for a stream MUST be retransmitted. Any
+  previous unacknowledged MAX_STREAM_DATA frame for the same stream SHOULD NOT
+  be retransmitted since a newer MAX_STREAM_DATA frame for a stream obviates the
+  need for delivering older ones. Similarly, the most recent MAX_DATA frame MUST
+  be retransmitted; previous unacknowledged ones SHOULD NOT be retransmitted.
```
ekr: I am starting to think it's a mistake to phrase this in terms of frames at all. It's the information you want to retransmit. It seems like there are lots of cases where you would be constantly updating your willingness to receive (if, say, you want to continue offering a given buffer size) but you only transmit occasionally. If I have to retransmit, why would I want to advertise stale data rather than the most current data.

ianswett: Agreed.  Anytime the words "retransmit frame" appear in the text, it's probably an error.  And the phrase 'retransmit packet' should occur twice: Once to tell you not to do it and once to tell you can do it for connection close packets only.

mikkelfj:
What if the connection close packet has other content?

Kind Regards,
Mikkel Fahnøe Jørgensen


On 29 September 2017 at 14.37.26, ianswett (notifications@github.com) wrote:

*@ianswett* commented on this pull request.
------------------------------

In draft-ietf-quic-transport.md
<https://github.com/quicwg/base-drafts/pull/807#discussion_r141857547>:

> @@ -2420,6 +2420,12 @@ When a packet is detected as lost, the sender re-sends any frames as necessary:
 * STOP_SENDING frames MUST be retransmitted, unless the stream has
become closed
   in the appropriate direction.  See {{solicited-state-transitions}}.

+* The most recent MAX_STREAM_DATA frame for a stream MUST be retransmitted. Any
+  previous unacknowledged MAX_STREAM_DATA frame for the same stream SHOULD NOT
+  be retransmitted since a newer MAX_STREAM_DATA frame for a stream
obviates the
+  need for delivering older ones. Similarly, the most recent MAX_DATA
frame MUST
+  be retransmitted; previous unacknowledged ones SHOULD NOT be retransmitted.

Agreed. Anytime the words "retransmit frame" appear in the text, it's
probably an error. And the phrase 'retransmit packet' should occur twice:
Once to tell you not to do it and once to tell you can do it for connection
close packets only.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/quicwg/base-drafts/pull/807#discussion_r141857547>, or mute
the thread

ianswett:
I would recommend putting other content with a connection close, but I
don't think it'd cause issues to retransmit a connection close packet that
happened to have other content in it.

martinthomson:
As Ian says, don't put anything else in the packet with CONNECTION_CLOSE.  There are a few things that I would put in, but only maybe: RST_STREAM (so that the receiver can identify the stream that triggered the error), and PADDING.

martinthomson:
Taking this for now, on the understanding that we want to properly fix this later.

PR #809, opened 2017-09-30
Fix references to companion QUIC drafts

### LINKED ISSUE #808: submitted drafts reference unspecific sibling specs
For instance:

https://tools.ietf.org/html/draft-ietf-quic-http-06#section-11.1

~~~
   [QUIC-TRANSPORT]
              Iyengar, J., Ed. and M. Thomson, Ed., "QUIC: A UDP-Based
              Multiplexed and Secure Transport", draft-ietf-quic-
              transport (work in progress), September 2017.
~~~

This should say:

~~~
   [QUIC-TRANSPORT]
              Iyengar, J., Ed. and M. Thomson, Ed., "QUIC: A UDP-Based
              Multiplexed and Secure Transport", draft-ietf-quic-
              transport-06 (work in progress), September 2017.
~~~

janaiyengar:
Closes #808

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -19,27 +19,6 @@ author:
     email: Michael.Bishop@microsoft.com
     role: editor
 
-normative:
-
-  QUIC-TRANSPORT:
-    title: "QUIC: A UDP-Based Multiplexed and Secure Transport"
-    date: {DATE}
-    seriesinfo:
-      Internet-Draft: draft-ietf-quic-transport-latest
-    author:
-      -
-        ins: J. Iyengar
-        name: Jana Iyengar
-        org: Google
-        role: editor
-      -
-        ins: M. Thomson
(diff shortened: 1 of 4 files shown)


martinthomson:
This won't work.  It will cause quic-http-05 to reference quic-transport-04.

The template has the ability to generate the right anchors as necessary for submission, but I haven't checked that it works properly.  I'll double-check when I get the chance.

BTW, I have a new submission process, which should remove some of the human error factors.  The code that generates the right tags has been recently updated (since the -06 drafts were submitted).

martinthomson:
I am fairly confident now that my new code correctly changes any instances of `draft-foo-latest` when building for submission.

I didn't include a link before, but [here are the new instructions](https://github.com/martinthomson/i-d-template/blob/master/doc/SUBMITTING.md). Be careful to note the use of `git tag -a`, which creates an annotated tag.

PR #813, opened 2017-10-02
Retransmit stream data, not stream frames

### LINKED ISSUE #763: ACK-only "frames"
4.4 references "Ack only frames" -- I suspect this should be ACK-only *packets*, or perhaps more generally, packets which don't contain anything retransmittable.

I'm also a little confused as to why this is, but that could just be my lack of familiarity with the congestion control space.

ianswett:
Motivated by #763


martinthomson:
This is a fine change, but I don't think that it closes #763.

ianswett:
Agreed on #763, there are a lot more errors than I thought.

martinthomson:
:)  Luckily this didn't actually close #763, even though it seemed like it might.

PR #815, opened 2017-10-02
Must not ack acks

ianswett:
Change SHOULD to MUST in "Implementations MUST NOT generate ACK packets in response to packets which only contain ACKs."

The current text is an error.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1992,7 +1992,7 @@ Error Code:
 Receivers send ACK frames to inform senders which packets they have received and
 processed, as well as which packets are considered missing.  The ACK frame
 contains between 1 and 256 ACK blocks.  ACK blocks are ranges of acknowledged
-packets. Implementations SHOULD NOT generate ACK packets in response to packets
+packets. Implementations MUST NOT generate ACK packets in response to packets
 which only contain ACKs. However, they SHOULD ACK those packets when sending
 ACKs for other packets.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1992,7 +1992,7 @@ Error Code:
 Receivers send ACK frames to inform senders which packets they have received and
 processed, as well as which packets are considered missing.  The ACK frame
 contains between 1 and 256 ACK blocks.  ACK blocks are ranges of acknowledged
-packets. Implementations SHOULD NOT generate ACK packets in response to packets
+packets. Implementations MUST NOT generate ACK packets in response to packets
```
martinthomson: Hmm, this is a fine change in principle, but the sentence is nonsensical.  "ACK packets"?

Isn't the change we want something like: "MUST NOT generate a packet containing only ACK frames if all they have received is packets containing ACK frames."

ianswett: Yes, of course.  That packet text is everywhere.

ianswett:
Thanks Martin, take another look at the updated text and see if that is correct.

draft-ietf-quic-transport.md
```diff
@@ -1992,9 +1992,10 @@ Error Code:
 Receivers send ACK frames to inform senders which packets they have received and
 processed, as well as which packets are considered missing.  The ACK frame
 contains between 1 and 256 ACK blocks.  ACK blocks are ranges of acknowledged
-packets. Implementations SHOULD NOT generate ACK packets in response to packets
-which only contain ACKs. However, they SHOULD ACK those packets when sending
-ACKs for other packets.
+packets. Implementations MUST NOT generate packets that only contain ACK frames
+in response to packets which only contain ACK frames. However, they SHOULD ACK
+packets containing only ack frames when sending ACK frames in response to other
```
martinthomson: "they SHOULD *acknowledge* packets containing only *ACK* frames when sending ACK frames in response to other packets."

PR #817, opened 2017-10-02
Simplify stateless retry

ekr:


draft-ietf-quic-transport.md
```diff
@@ -618,10 +618,11 @@ any version negotiation that occurred (see {{version-negotiation}}).  The client
...
 
-The payload of the Server Stateless Retry packet contains STREAM frames and
-could contain PADDING and ACK frames.  A server can only send a single Server
-Stateless Retry packet in response to each Client Initial packet that is
-receives.
+The payload of the Server Stateless Retry packet contains a single
+STREAM frame on stream 0 with offset 0 containing the server's
+cryptographic stateless retry material. It MUST NOT contain any other
+frames. Any future stream frames from the server will also start at
+stream offset 0.
```
martinthomson: Nit: STREAM frames.  Also,  this somewhat implies that ALL STREAM frames will have an offset of 0.  I think that you mean that the NEXT STREAM frame starts at an offset of 0.

marten-seemann:
Why am I not allowed to include an ACK for the Client Initial here? This would provide the Client with a first RTT measurement.
Furthermore, I’d like to avoid special code that makes sure that this single packet doesn’t contain any other frames.

ianswett:
Marten brings up a good point.  Having an early RTT signal is really nice both to avoid tail latency and spurious retransmits.  If you've sent multiple client initials, you don't know which this is in response to.

mcmanus:
the packetnumber of the SSR is the packetnumber of the CI that triggered it.. and now that we don't do re-randomization that should be deterministic. so that's essentially an ACK, right?

ianswett:
I didn't realize that's true, but if so, that would be sufficient for an RTT sample.  It is a non-standard path for RTT measurement, but SSR is a sort of special case, since you're essentially creating a whole new connection, so I'm ok with that in this case.

marten-seemann:
But why do we need to create a special case here? Why is it explicitly forbidden to include an ACK?

martinthomson:
Let's open an issue on this.  I had assumed that the Retry would be sent immediately, so the ACK delay wouldn't factor into the RTT calculation.

PR #824, opened 2017-10-03
Remove packet number re-randomization requirement

### LINKED ISSUE #823: Re-randomization for packet numbers
The draft is pretty inconsistent about what packet number to use in response to version negotiation or stateless retry.  Since no one implements re-randomization, just move to an increment-by-one.

martinthomson:
I don't believe that anyone actually implemented this, and it is
unnecessary.

Closes #823.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -569,10 +569,9 @@ unless it has received a packet from the server.  If the client has received a
 packet from the server, the connection ID field uses the value provided by the
 server.
 
-The packet number used for Client Initial packets is initialized with a random
-value each time the new contents are created for the packet.  Retransmissions of
-the packet contents increment the packet number by one, see
-({{packet-numbers}}).
+The first Client Initial packet that is sent by a client contains a random
+31-bit value.  All subsequent packets contain a packet number that is
+incremented by one, see ({{packet-numbers}}).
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -907,9 +896,9 @@ When the client receives a Version Negotiation packet from the server, it should
 select an acceptable protocol version.  If the server lists an acceptable
 version, the client selects that version and reattempts to create a connection
 using that version.  Though the contents of a packet might not change in
-response to version negotiation, a client MUST increase the packet number it
-uses on every packet it sends.  Packets MUST continue to use long headers and
-MUST include the new negotiated protocol version.
+response to version negotiation, a client MUST increase the packet number by one
```
janaiyengar: This requirement of increment by 1 is broken around migration. This should probably make an exception for that, with a pointer to Section 7.6... but that's probably an overkill. How about "a client MUST increase the packet number by one for every packet it sends in response to version negotiation."

martinthomson: Right, though you won't be migrating during version negotiation, so your fix works.

janaiyengar:
This is fine, but I would just close it. We've agreed that we don't need randomization.

martinthomson:
Since #850 isn't resolving, I'm going to merge this and then we can continue to fight about whether 0 is special or not at our leisure.

PR #833, opened 2017-10-03
Connection ID is actually opaque

### LINKED ISSUE #834: Connection ID is not a number
The resolution to #786 wasn't editorial.  It changed connection ID to a number, which is unnecessary.  The value is constructed by the server and echoed by the client.  If it is stored internally by clients and server, that's not really observable.

martinthomson:
Closes #834, #881.


martinduke:
I have no metaphysical position on this question, but it is not purely academic. Connection ID is an input to cleartext key generation and we need to make this clear one way or the other.

PR #842, opened 2017-10-09
Move stateless reset token to the end

### LINKED ISSUE #820: Stateless Reset Format is hard to implement
The issue is that part of the token is in what you would otherwise interpret as the sequence number field, which means that you need special serializers and deserializers. It would be better to just have a random sequence number (which is what this effectively is anyway) and then the token in the body.  In fact, that's so obviously better that when I implemented this, I totally missed it.

martinthomson:
This makes the location of the token fixed with respect to the end of
packets, which makes it easier to find.  It also expands the conditions
for checking the token, which now include cases where packets appear
to be duplicated.

Closes #820.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1554,6 +1554,10 @@ following layout:
 +                     [Connection ID (64)]                      +
 |                                                               |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                     Packet Number (8/16/32)                   |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                        Random Octets (*)                    ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                                                               |
 +                                                               +
 |                                                               |
@@ -1562,27 +1566,28 @@ following layout:
 +                                                               +
 |                                                               |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                        Random Octets (*)                    ...
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ~~~
-
-This packet SHOULD use the short header form with the shortest possible packet
-number encoding.  This minimizes the perceived gap between the last packet that
-the server sent and this packet.  The leading octet of the Stateless Reset Token
-will be interpreted as a packet number.  A server MAY use a different short
-header type, indicating a different packet number length, but this allows for
-the message to be identified as a stateless reset more easily using heuristics.
 
 A server copies the connection ID field from the packet that triggers the
 stateless reset.  A server omits the connection ID if explicitly configured to
 do so, or if the client packet did not include a connection ID.
 
+The Packet Number field is set to a randomized value.  This packet SHOULD use
+the short header form with the shortest possible packet number encoding.  This
+minimizes the perceived gap between the last packet that the server sent and
+this packet.  A server MAY use a different short header type, indicating a
+different packet number length, but this allows for the message to be identified
+as a stateless reset more easily using heuristics.
+
 After the first short header octet and optional connection ID, the server
 includes the value of the Stateless Reset Token that it included in its
 transport parameters.
 
-After the Stateless Reset Token, the server pads the message with an arbitrary
+After the Packet Number, the server pads the message with an arbitrary
 number of octets containing random values.
+
+Finally, the last 16 octets of the packet are set to the value of the Stateless
+Reset Token.
 
 This design ensures that a stateless reset packet is - to the extent possible -
 indistinguishable from a regular packet.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1562,28 +1566,29 @@ following layout:
...
-the server sent and this packet.  The leading octet of the Stateless Reset Token
-will be interpreted as a packet number.  A server MAY use a different short
-header type, indicating a different packet number length, but this allows for
-the message to be identified as a stateless reset more easily using heuristics.
-
 A server copies the connection ID field from the packet that triggers the
 stateless reset.  A server omits the connection ID if explicitly configured to
 do so, or if the client packet did not include a connection ID.
 
+The Packet Number field is set to a randomized value.  This packet SHOULD use
```
ekr: Nit: I would say "The server SHOULD use..."

draft-ietf-quic-transport.md
```diff
@@ -1562,28 +1566,29 @@ following layout:
...
+the short header form with the shortest possible packet number encoding.  This
```
ekr: I think I would phrase this as "the packet type should be X, which represents the shortest packet encoding", because you aren't allowed to use long format under any conditions.

draft-ietf-quic-transport.md
```diff
@@ -1595,11 +1600,11 @@ CONNECTION_CLOSE or APPLICATION_CLOSE frame if it has sufficient state to do so.
...
 A client detects a potential stateless reset when a packet with a short header
-cannot be decrypted.  The client then performs a constant-time comparison of the
-16 octets that follow the Connection ID with the Stateless Reset Token provided
-by the server in its transport parameters.  If this comparison is successful,
-the connection MUST be terminated immediately.  Otherwise, the packet can be
-discarded.
+either cannot be decrypted or is marked as a potential duplicate.  The client
+then performs a constant-time comparison of the last 16 octets of the packet
+with the Stateless Reset Token provided by the server in its transport
+parameters.  If this comparison is successful, the connection MUST be terminated
```
ekr: Can you say what "terminated" means? Is it the same as having received CONNECTION_CLOSE?

draft-ietf-quic-transport.md
```diff
@@ -1562,28 +1566,30 @@ following layout:
...
 A server copies the connection ID field from the packet that triggers the
 stateless reset.  A server omits the connection ID if explicitly configured to
 do so, or if the client packet did not include a connection ID.
 
+The Packet Number field is set to a randomized value.  The server SHOULD send a
+packet with a short header and a type of 0x01.  This produces the shortest
+possible packet number encoding, which minimizes the perceived gap between the
+last packet that the server sent and this packet.  A server MAY use a different
+short header type, indicating a different packet number length, but this allows
+for the message to be identified as a stateless reset more easily using
```
janaiyengar: nit: "this allows for" - I think you mean to say that using a 1-byte packet number makes life easier, but "this" is ambiguous here. Maybe something like: "A server MAY use a different short header type, indicating a different packet number length, but using 0x01 allows for the message to be identified as a stateless reset more easily using heuristics."

draft-ietf-quic-transport.md
janaiyengar: Perhaps "all connection state MUST be discarded."

draft-ietf-quic-transport.md
MikeBishop: It's not the identification we care about, but the fact that an observer could identify the SR by seeing a packet number jump.  If the number is only eight bits long, then it's harder for an observer to infer a large jump.

draft-ietf-quic-transport.md
MikeBishop: Any of the above seem fine for now; we'll want to tie this in with @janaiyengar's Unified Theory of Connection Closure when that lands.

draft-ietf-quic-transport.md
martinthomson: Yeah, but I take the comment- I will revise along the lines suggested.

draft-ietf-quic-transport.md
martinthomson: You can't discard state if you are going to enter the draining period.  But we don't need draining, we just need to stop.  (Any subsequent stateless resets we get will get discarded according to our new packet handling rules.)

draft-ietf-quic-transport.md
```diff
@@ -1595,11 +1601,12 @@ CONNECTION_CLOSE or APPLICATION_CLOSE frame if it has sufficient state to do so.
 #### Detecting a Stateless Reset
 
 A client detects a potential stateless reset when a packet with a short header
-cannot be decrypted.  The client then performs a constant-time comparison of the
-16 octets that follow the Connection ID with the Stateless Reset Token provided
-by the server in its transport parameters.  If this comparison is successful,
-the connection MUST be terminated immediately.  Otherwise, the packet can be
-discarded.
+either cannot be decrypted or is marked as a potential duplicate.  The client
```
ianswett: nit: I'd go with "is marked as a duplicate packet."

draft-ietf-quic-transport.md
```diff
@@ -1595,11 +1601,12 @@ CONNECTION_CLOSE or APPLICATION_CLOSE frame if it has sufficient state to do so.
...
+then performs a constant-time comparison of the last 16 octets of the packet
```
ianswett: Is it necessary to mention it's constant time?

draft-ietf-quic-transport.md
```diff
@@ -1595,11 +1601,12 @@ CONNECTION_CLOSE or APPLICATION_CLOSE frame if it has sufficient state to do so.
...
+with the Stateless Reset Token provided by the server in its transport
+parameters.  If this comparison is successful, the client MUST discard all
```
ianswett: Instead of successful, how about "If they are identical" ?

draft-ietf-quic-transport.md
```diff
@@ -1595,11 +1601,12 @@ CONNECTION_CLOSE or APPLICATION_CLOSE frame if it has sufficient state to do so.
...
+connection state and not send any further packets on this connection. If the
+comparison is unsuccessful, the packet can be discarded.
```
ianswett: is unsuccessful => fails?

PR #844, opened 2017-10-09
AEAD all packets. Fixed #693, Fixes #840.

### LINKED ISSUE #693: Specify AEAD-based protection of cleartext packets
Based on discussion in Prague:

- (Provisionally) algorithm is AES-GCM
- Key should be derived from (1) client connection ID and (2) a fixed key baked into the QUIC version
- Otherwise standard AEAD

### LINKED ISSUE #840: Version negotiation responses and FNV-1a protection
Reading the specs, it wasn't clear to me that version negotiation responses are not protected by FNV-1a. This should IMO be explicitly called out.

draft-tls says

> QUIC adds an integrity check to all cleartext packets.

and draft-transport says

> Cleartext packets are sent during the handshake prior to key negotiation.

Neither if which made it obvious to me that a version negotiation response was excluded.

(And out of curiosity - why isn't there an FNV-1a? Seems like an odd special case?)

ekr:
This makes the change we discussed in Seattle where all Cleartext
packets are encrypted with AES-GCM using a key derived from the
client connection ID and a random per-version value.

Note: I generated this value using randomness.org.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -632,6 +632,33 @@ QUIC uses HKDF with the same hash function negotiated by TLS for
 key derivation.  For example, if TLS is using the TLS_AES_128_GCM_SHA256, the
 SHA-256 hash function is used.
 
+### Cleartext Packet Secrets {#cleartext-secrets}
+
+Cleartext packets are protected with secrets derived from the client's
+connection ID. Specifically:
+
+~~~
+   quic_version_1_salt = ed 1d a2 c6 0b ae 80 03 eb db 17 d8 92 5d 95 92
+                         4a 6e 87 57 cd 48 d9 bf d2 41 c7 19 cb 80 81 ef
+
+   cleartext_secret = HKDF-Extract(quic_version_1_salt,
+                                   client_connection_id)
+
+   client_cleartext_secret =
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-tls.md
```diff
@@ -758,8 +785,11 @@ used for QUIC packet protection is AEAD that is negotiated for use with the TLS
 connection.  For example, if TLS is using the TLS_AES_128_GCM_SHA256, the
 AEAD_AES_128_GCM function is used.
 
-Regular QUIC packets are protected by an AEAD algorithm {{!RFC5116}}.  Version
-negotiation and public reset packets are not protected.
+All QUIC packets other than version negotiation and public reset packets are
+protected with an AEAD algorithm {{!RFC5116}}. Cleartext packets are protected
+with the AEAD_AES_128_GCM and a key derived from the client's connection ID.
+This provides protection against off-path attackers and robustness against
+QUIC version unaware middleboxes, but not against on-path attackers.
```
huitema: I assume that there will be at some point a request to support "QUIC for IOT" and use AES 128 with CCM 8. But we are not there yet, and the IOT folks would just have to create their own version...

mikkelfj:
Why the case inconsistency in HKDF "QUIC client cleartext Secret" and ditto server?

Issue #822, opened 2017-10-03
multiple cleartext types


jpinner-lyft:
There is no value and additional validation cost in prescribing cleartext packet types for each direction of communication.

ianswett:
One benefit is that if the network routes packets incorrectly, it's easy for the server to drop the client's packets and client to drop the server's.

jpinner-lyft:
If there is a concern about directionality I think it would be preferable to mirror the directionality inferred in the encrypted packets by using hashes based on the packet direction similar to what's done for the encryption keys today.

ianswett:
I'm happy with that as well.

PR #847, opened 2017-10-10
Rename and consolidate packet types

### LINKED ISSUE #846: Consolidate Client and Server Cleartext
Now that we have an AEAD for handshake packets, and we have different key derivation for server and client, the different packet type for server and client messages is redundant.  We can remove the extra type and simplify.

martinthomson:
The two cleartext types are no longer needed to distinguish between client and
server.  Server packets are protected with a different key to that used for
client packets, so a server will be unable to process a packet from another
server as though it were a client packet (not that this was possible anyway
given that we have a different type for the initial packet).

Consequently, I have renamed packets to use shorter names and to better fit
their roles:

* Client Initial -> Initial
* Server Stateless Retry -> Retry (also avoids confusion with Stateless Reset)
* Server Cleartext -> Handshake
* Client Cleartext -> Handshake

I've rephrased the text in the TLS spec to match.

The only functional change here is to consolidate Server and Client Cleartext
into a single codepoint.  I did renumber the 0-RTT and 1-RTT protected packets
as a result.

Closes #846, #822.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -318,7 +318,7 @@ ensures that TLS handshake messages are delivered in the correct order.
    Replayable QUIC Frames
                             -------->
 
-                                      QUIC STREAM Frame <0>: @C
+                                      QUIC STREAM Frame <0>: @H
                                                ServerHello
                                   {TLS Handshake Messages}
                             <--------
@@ -326,7 +326,7 @@ ensures that TLS handshake messages are delivered in the correct order.
 
                                            QUIC Frames <any> @1
                             <--------
-@C QUIC STREAM Frame(s) <0>:
+@H QUIC STREAM Frame(s) <0>:
      (EndOfEarlyData)
      {Finished}
                             -------->
@@ -339,8 +339,8 @@ In {{quic-tls-handshake}}, symbols mean:
 
 * "<" and ">" enclose stream numbers.
 
-* "@" indicates the keys that are used for protecting the QUIC packet (C =
-  cleartext, with integrity only; 0 = 0-RTT keys; 1 = 1-RTT keys).
+* "@" indicates the keys that are used for protecting the QUIC packet (H =
+  handshake, with integrity only; 0 = 0-RTT keys; 1 = 1-RTT keys).
 
 * "(" and ")" enclose messages that are protected with TLS 0-RTT handshake or
   application keys.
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -549,70 +548,70 @@ See {{version-negotiation}} for a description of the version negotiation
...
+All these packets use the long header and contain the current QUIC version in
+the version field.
 
-In order to prevent tampering by version-unaware middleboxes, Cleartext
-packets are protected with a connection and version specific key, as
-described in {{QUIC-TLS}}. This protection does not provide confidentiality
-or integrity against on-path attackers, but provides some level of
-protection against off-path attackers.
+In order to prevent tampering by version-unaware middleboxes, handshake packets
+are protected with a connection and version specific key, as described in
```
MikeBishop: "connection- and version-specific"

draft-ietf-quic-transport.md
```diff
@@ -549,70 +548,70 @@ See {{version-negotiation}} for a description of the version negotiation
...
 and the complete cryptographic handshake message MUST fit in a single packet
 (see {{handshake}}).
 
-The client uses the Client Initial Packet type for any packet that contains an
-initial cryptographic handshake message.  This includes all cases where a new
-packet containing the initial cryptographic message needs to be created, this
-includes the packets sent after receiving a Version Negotiation
-({{packet-version}}) or Server Stateless Retry packet
-({{packet-server-stateless}}).
+The client uses the Initial Packet type for any packet that contains an initial
```
MikeBishop: Some capitalization inconsistency between "Initial Packet" and "Initial packet".  Might be worth a quick pass to decide which system you intend to use.

draft-ietf-quic-transport.md
```diff
@@ -805,8 +787,8 @@ an unsupported version.
 
 # Frames and Frame Types {#frames}
 
-The payload of cleartext packets and the plaintext after decryption of protected
-payloads consists of a sequence of frames, as shown in {{packet-frames}}.
+The payload of Initial, Handshake and all protected packets consists of a
```
MikeBishop: Since Initial and Handshake packets are also semi-protected, we might be able to get the specific list out entirely.  The payload of all packet types, after removing packet protection, consists of..."

draft-ietf-quic-tls.md
```diff
@@ -610,11 +612,11 @@ As TLS reports the availability of keying material, the packet protection keys
...
 selection of AEAD function is also updated to match the AEAD negotiated by TLS.
 
-For packets other than any unprotected handshake packets (see {{cleartext-hs}}),
-once a change of keys has been made, packets with higher packet numbers MUST be
-sent with the new keying material.  The KEY_PHASE bit on these packets is
-inverted each time new keys are installed to signal the use of the new keys to
-the recipient (see {{key-phases}} for details).
+For packets other than any handshake packets (see {{hs-protection}}), once a
+change of keys has been made, packets with higher packet numbers MUST be sent
+with the new keying material.  The KEY_PHASE bit on these packets is inverted
```
ianswett: Why would one want to send more handshake packets without 1RTT keys once 1RTT keys have been established?

martinthomson: Not sure what you mean.  This says that if you change keys, then you need to use the new keys and not the old ones.

ianswett: Nevermind, I'd been up since 3AM and ran out of brain.

ianswett:
I think this also closes #822 

Is there any reason not to merge this, since I thought we already had?

martinthomson:
No reason in particular, I just wanted to give it more "soak" time.

Issue #614, opened 2017-06-08
PMTUD details


gloinul:
In draft-ietf-quic-transport-03, Section 9:

   All QUIC packets SHOULD be sized to fit within the estimated PMTU to
   avoid IP fragmentation or packet drops.  To optimize bandwidth
   efficiency, endpoints SHOULD use Packetization Layer PMTU Discovery
   ([RFC4821]) and MAY use PMTU Discovery ([RFC1191], [RFC1981]) for
   detecting the PMTU, setting the PMTU appropriately, and storing the
   result of previous PMTU determinations.

There are details in the method described in RFC 4821 that needs to have explicit description of how the method is applied in QUIC. This includes, format of probing packets, loss response, additional challenge of ICMP feedback.

martinthomson:
@martinduke, do you think that you could take a look at this?

martinduke:
Yes.

martinduke:
Hi Magnus,

I agree that some language about probe packets would be useful and there will be a PR shortly. This will also incorporate some advice about loss response. I would very much like @ianswett to consolidate the various actions to take upon loss detection in that document, but that's another issue.

As for "challenges with ICMP feedback", the section "Special Considerations for PMTU Discovery" is specifically meant to address this. What is that missing?

Is there anything else you'd like to see here?

martinduke:
I had another look at my RFC4821 notes. Here are some observations:
- In my opinion, someone familiar with other aspects of QUIC can read RFC 4821 and implement PLPMTUD with no more difficulty than they would with TCP or SCTP. So strictly speaking, no additional text in the QUIC spec is necessary.
- Section 6 discusses necessary protocol properties, and QUIC has them. Section 6 specifically mentions padding as a nice-to-have. PR #849, in its current state, makes this explicit.
- It would be possible to state explictly that search_low (7.2) should be the IPv6 minimum packet size, as that is the minimum size in QUIC. However, I personally find this redundant.
- It is possible to recommend an aggressive probe sizing strategy (7.3) since QUIC probe retransmissions are unnecessary. Again, I find this redundant.

I believe the current PR adequately cues the reader on relevant considerations without laboriously repeating RFC 4821. I am also persuadable that the original text is fine and the PR is unnecessary. However, I am interested if others find the current language, in conjunction with 4821, to be insufficient, and if so what pieces are ambiguous.

gloinul:
[...]

PR #849, opened 2017-10-10
Add some suggestions for RFC 4821 probe packets.

martinduke:
This is in response to #614 .

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2405,12 +2405,12 @@ discussed in more detail in {{QUIC-RECOVERY}}.
 
 # Special Considerations for Packetization Layer PMTU Discovery
 
-The PADDING frame provides a useful option for PMTU probe packets that does not
-exist in other transports. QUIC Packets with no frames other than PADDING must be
-acknowledged, but need not be retransmitted if lost. PADDING probe packets, if
-lost, also do not delay delivery of application data. Their only impact on
-application data is the extent that they consume the congestion window and
-connection flow control credits.
+The PADDING frame provides a useful option for PMTU probe packets that does
+not exist in other transports. PADDING frames generate acknowledgements, but
+their content need not be delivered reliably. PADDING frames may delay the
+delivery of application data, as they consume the congestion window. However,
+by definition their likely loss in a probe packet does not require delay-
+inducing retransmission of application data.
 
 ## Special Considerations for PMTU Discovery
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2412,6 +2412,16 @@ delivery of application data, as they consume the congestion window. However,
 by definition their likely loss in a probe packet does not require delay-
 inducing retransmission of application data.
 
+When implementing the algorithm in Section 7.2 of {{!RFC4821}}, the initial
+value of search_low SHOULD be consistent with the IPv6 minimum packet size.
+Paths that do not support this size cannot deliver Client Initial packets,
+and therefore are not QUIC-compliant.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2403,6 +2403,15 @@ that it sends.
 Strategies and implications of the frequency of generating acknowledgments are
 discussed in more detail in {{QUIC-RECOVERY}}.
 
+# Special Considerations for Packetization Layer PMTU Discovery
+
+The PADDING frame provides a useful option for PMTU probe packets that does not
+exist in other transports. QUIC Packets with no frames other than PADDING must be
+acknowledged, but need not be retransmitted if lost. PADDING probe packets, if
```
martinthomson: We're trying to avoid talking about retransmission of packets, instead concentrating on the information content.  In this case, you could say "QUIC packets that only include PADDING frames cause the receiver to generate an acknowledgment, but PADDING frames don't contain information that needs to be reliably delivered."  Or something like that.

draft-ietf-quic-transport.md
```diff
@@ -2403,6 +2403,15 @@ that it sends.
...
+lost, also do not delay delivery of application data. Their only impact on
```
martinthomson: I would remove the "do not delay clause" on the basis that they could consume congestion control bytes.

draft-ietf-quic-transport.md
```diff
@@ -2403,6 +2403,15 @@ that it sends.
...
+application data is the extent that they consume the congestion window and
+connection flow control credits.
```
martinthomson: Padding doesn't consume flow control (only the bytes delivered in STREAM frames do).

draft-ietf-quic-transport.md
```diff
@@ -2403,6 +2403,15 @@ that it sends.
 Strategies and implications of the frequency of generating acknowledgments are
 discussed in more detail in {{QUIC-RECOVERY}}.
 
+# Special Considerations for Packetization Layer PMTU Discovery
+
+The PADDING frame provides a useful option for PMTU probe packets that does not
+exist in other transports. QUIC Packets with no frames other than PADDING must be
```
martinthomson: This line is a tiny bit too long.

draft-ietf-quic-transport.md
martinduke: Perhaps poorly worded, but the subject of the setting is "the <b>loss</b> of PADDING probe packets". Sending them in the first place consumes cwnd, I'll reorder the sentences.

martinduke:
@martinthomson I re-read the RFC, but I've asked @gloinul to comment more on his concerns with the existing text, rather than try to guess. This PR addresses the one objection I wholeheartedly agree with.

martinthomson:
I need to confirm that PADDING frames generate acknowledgments.  I think that they do, but the text needs to be clearer.  This is the best I could find:

> A receiver that is only sending ACK frames will not receive acknowledgments for its packets. Sending an occasional MAX_DATA or MAX_STREAM_DATA frame as data is received will ensure that acknowledgements are generated by a peer. Otherwise, an endpoint MAY send a PING frame once per RTT to solicit an acknowledgment.

Is it possible to rephrase this so that it mentions frame types that we know generate acknowledgments (like PING)?

ianswett:
@martinthomson The current text intends to say ALL frames generate acknowledgements except ACK frames.  Hence only packets that only contain ACK_FRAMEs do not generate acknowledgements.  

I'd rather no call out every frame that is expected to cause an ACK_FRAME to be sent, since it's all minus 1.

I need to clarify this and other text after discussion in Seattle, so I sent out PR #907

martinthomson:
@martinduke, this has a few structural issues that I'd like to resolve.  Can you check #985?

PR #851, opened 2017-10-10
Make error codes smaller

### LINKED ISSUE #211: Error codes

martinthomson:
See #723 for a full discussion.

Closes #211 (finally).

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1723,7 +1723,7 @@ The CONNECTION_CLOSE frame is as follows:
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |           Error Code (16)     |   Reason Phrase Length (16)   |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                      Reason Phrase Length (*)               ...
+|                        Reason Phrase (*)                    ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ~~~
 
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1721,17 +1721,17 @@ The CONNECTION_CLOSE frame is as follows:
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                         Error Code (32)                       |
+|           Error Code (16)     |   Reason Phrase Length (16)   |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|   Reason Phrase Length (16)   |      [Reason Phrase (*)]    ...
+|                      Reason Phrase Length (*)               ...
```
MikeBishop: Two length fields, no value field.

draft-ietf-quic-transport.md
```diff
@@ -3099,10 +3099,11 @@ See {{iana-error-codes}} for details of registering new error codes.
 
 ## Application Protocol Error Codes {#app-error-codes}
 
-Application protocol error codes are 32-bits long, but the management of
-application error codes are left to application protocols.  Application protocol
-error codes are used for the RST_STREAM ({{frame-rst-stream}}) and
-APPLICATION_CLOSE ({{frame-application-close}}) frames.
+Application protocol error codes are 16 bit unsigned integers, but the
```
MikeBishop: "16-bit," no?

martinthomson:
Thanks Mike.

PR #853, opened 2017-10-10
Remove ACKs from draining period

### LINKED ISSUE #852: Acknowledgments during draining period
At the Seattle interim, we agreed to simplify connection shutdown by removing the ability to send acknowledgments during the draining period.  An endpoint is only able to send packets that include CONNECTION_CLOSE/APPLICATION_CLOSE.

martinthomson:
As agreed at the interim, this removes the ability to acknowledge during the
draining period.  An endpoint can use the APPLICATION_CLOSE/CONNECTION_CLOSE
frame to exit this state early (it's a kind of ACK).

Closes #852.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1461,14 +1461,10 @@ The draining period persists for three times the current Retransmission Timeout
 can be acknowledged, but no new application data can be sent on the connection.
 
 Different treatment is given to packets that are received while a connection is
-in the draining period depending on how the connection was closed.  An endpoint
-that is in a draining period MUST NOT send packets unless they contain a
-CONNECTION_CLOSE or APPLICATION_CLOSE frame.
+in the draining period depending on how the connection was closed.
 
-An endpoint that receives either CONNECTION_CLOSE or APPLICATION_CLOSE while in
-the draining period MAY terminate the draining period immediately.  An endpoint
-that receives either of these frames can send a packet containing a
-CONNECTION_CLOSE to allow its peer to exit the draining period more rapidly.
+An endpoint that is in a draining period MUST NOT send packets unless they
+contain a CONNECTION_CLOSE or APPLICATION_CLOSE frame.
 
 Once the draining period has ended, an endpoint SHOULD discard per-connection
 state.  This results in new packets on the connection being discarded.  An
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1461,18 +1461,21 @@ The draining period persists for three times the current Retransmission Timeout
 can be acknowledged, but no new application data can be sent on the connection.
 
 Different treatment is given to packets that are received while a connection is
-in the draining period depending on how the connection was closed.  In all
-cases, it is possible to acknowledge packets that are received as normal, but
-other reactions might be preferable depending on how the connection was closed.
-An endpoint that is in a draining period MUST NOT send packets containing frames
-other than ACK, PADDING, or CONNECTION_CLOSE.
+in the draining period depending on how the connection was closed.  An endpoint
+that is in a draining period MUST NOT send packets unless they contain a
+CONNECTION_CLOSE or APPLICATION_CLOSE frame.
+
+An endpoint that receives either CONNECTION_CLOSE or APPLICATION_CLOSE while in
+the draining period can terminate the draining period immediately.  An endpoint
+that receives either of these frames can send a packet containing a
+CONNECTION_CLOSE to allow its peer to exit the draining period more rapidly.
 
 Once the draining period has ended, an endpoint SHOULD discard per-connection
 state.  This results in new packets on the connection being discarded.  An
 endpoint MAY send a stateless reset in response to any further incoming packets.
 
 The draining period does not apply when a stateless reset ({{stateless-reset}})
-is used.
+is sent.
 
 
 ### Idle Timeout
@@ -1511,10 +1514,6 @@ Note:
   new packet numbers is primarily of advantage to loss recovery and congestion
   control, which are not expected to be relevant for a closed connection.
   Retransmitting the final packet requires less state.
-
-An endpoint can cease sending CONNECTION_CLOSE or APPLICATION_CLOSE frames if it
-receives either a CONNECTION_CLOSE, APPLICATION_CLOSE or an acknowledgement for
-a packet that contained a either close frame.
 
 An immediate close can be used after an application protocol has arranged to
 close a connection.  This might be after the application protocols negotiates a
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1462,11 +1462,14 @@ The draining period persists for three times the current Retransmission Timeout
...
 
 Different treatment is given to packets that are received while a connection is
-in the draining period depending on how the connection was closed.  In all
-cases, it is possible to acknowledge packets that are received as normal, but
-other reactions might be preferable depending on how the connection was closed.
-An endpoint that is in a draining period MUST NOT send packets containing frames
-other than ACK, PADDING, or CONNECTION_CLOSE.
+in the draining period depending on how the connection was closed.  An endpoint
+that is in a draining period MUST NOT send packets unless they contain a
+CONNECTION_CLOSE or APPLICATION_CLOSE frame.
```
ianswett: This looks fine, but what's APPLICATION_CLOSE.  Did I miss that from another PR?

martinthomson: See #722.

ianswett: I saw that PR go by, but the description didn't mention a new frame type.  I'm not super opposed, but I definitely wasn't expecting it.

draft-ietf-quic-transport.md
```diff
@@ -1462,11 +1462,14 @@ The draining period persists for three times the current Retransmission Timeout
...
+
+An endpoint that receives either CONNECTION_CLOSE or APPLICATION_CLOSE while in
+the draining period can terminate the draining period immediately.  An endpoint
+that receives either of these frames can send a packet containing a
```
janaiyengar: can send --> MAY send. That said, I don't get the value of sending a CONNECTION_CLOSE here. Why is this useful?

martinthomson: I had a chat with @nibanks at the interim and he was concerned about finding a way to quickly terminate the draining period.  Lingering timers are annoying to manage in testing.

janaiyengar: I don't follow. This says the endpoint terminates draining immediately. Are you thinking the peer might be in draining? Can you give an example?

martinthomson: On error client sends CONNECTION_CLOSE.  It has to wait for 3RTO (or whatever) to cleanup state.  But if the server sends a reciprocal CONNECTION_CLOSE, there is far less need to hang around.

I guess that the concern is that you might get reordering around that close.  Maybe that's the problem and maybe @nibanks will have to work out another way to get test cases to end quickly.

nibanks: I just was asking for some way to (optionally) acknowledge that the connection was closed, so that the draining period could be short circuited.

Specifically, in my case, I have client and server running on the same machine, and the tests will usually connect them, do something and then try to cleanly disconnect. With >100 test cases and running them locally every time I build, I want them to be as fast as possible.

martinthomson: Yeah, though as a practical matter, terminating the draining period won't work in general.  Maybe - with the allowance for a reciprocal close - you can make some assumptions in your test cases that will allow you to terminate early.

The alternative is to do as I suggested: find a way to shift timers.  I do that for every test framework I build because I hate waiting.  It will make your recovery tests run faster as well.

For now, I'm going to remove the termination clause from this PR.

nibanks: I'm still unclear why it wouldn't work. What is the reordering problem? Why can't you just use ACKs then (which is what I was doing originally before this PR)?

martinthomson: Server sends CONNECTION_CLOSE, client sends CONNECTION_CLOSE.  Server throws away state.  Client packet from before the CONNECTION_CLOSE arrives at server.

The point of the draining period is to have a catcher for the delayed packets that might arrive.  Now, if you don't care (because you stopped listening on the socket, for instance), then that's great and you shorten the draining period to whatever you like.  But if you are writing test cases, then you can skew the clock artificially and retain the benefits of the draining period.

(Maybe we need to make it clear that the draining period is somewhat discretionary on the part of endpoints.  Happy to do that in a followup PR.)

nibanks: Ok, one follow up question then:

What are you supposed to do with incoming packets while in the draining period? Do you throw them away or indicate them up to the application?

If you are throwing them away, what is the harm it interpreting either an ACK for your CONNECTION_CLOSE or a separate CONNECTION_CLOSE as reason enough to end the local draining period? What other state would be lost?

If you are supposed to indicate them to the application, how is that supposed to work, assuming they are the one that initiated the close in the first place? They would likely be in a blocking close() call.

martinthomson: Dump them.

There is no harm in doing whatever you like at this point - you have decided that the connection is no more.  You either terminated the connection gracefully (because you decided you had everything you needed) or you detected a fatal error of some kind (including stateless reset).

You can - if you like - end the draining period early, but you have to be willing to accept the consequences.  You might still receive packets for a while.  Meaning that you no longer have something that is allocated to handle them properly, though that might not be a particular problem for your architecture.  In the case where you sent the first CONNECTION_CLOSE, it might be the case that your CONNECTION_CLOSE was lost and needs retransmission.  Not being able to generate another CONNECTION_CLOSE means that you might get a LOT more packets.  A server might have to resort to a stateless reset in that case.

janaiyengar: So the point of draining is to hang around to respond to packets that the peer doesn't know have been acknowledged. If we go into a draining period after sending a CONNECTION_CLOSE, then we have to wait until we hear back from the peer. When we hear back from the peer, they've heard us, we've heard them, and it's all good. 

However the side that sends the last CONNECTION_CLOSE (in response to a received CONNECTION_CLOSE) cannot leave because it cannot be sure that the peer heard the last CONNECTION_CLOSE.

There more text that surely needs to be added here, so I say let's get this in and we can work out the details of how to finesse this later. I'll merge this.

nibanks: Thanks a lot for walking me through the issues. The main issue is that if you exit draining too early, you might receive packets destined for the connection afterwards and they might be hard to handle, depending on the implementation, correct? Specifically, that is mostly a server side problem, as on the client side, you likely close the socket and don't have to worry about it any more?

Is that reason enough to remove the optional mechanism for exiting the draining early? Shouldn't implementations that know they can't handle it well, just not exit early?

janaiyengar: I think we can have text here that says something like the endpoint SHOULD hang around, which leaves a bit of wiggle room for folks who know what they're doing. I'd rather stipulate the thing that keeps folks safe and have the ones who know what they're doing circumvent the stipulation. Do you want to file an issue for leaving draining early?

nibanks: Sure, I will open an issue.

PR #855, opened 2017-10-10
Remove application close

### LINKED ISSUE #854: Use immediate close for application close
At the Seattle interim we concluded that it was important to signal a close to intermediaries (those that have the keys, that is).  This allows application endpoints to reach an agreed close state and then terminate the transport without leaving the middlebox on a timer.

In short, once the application protocol has reached a terminal state, it will signal using APPLICATION_CLOSE (some code, maybe NO_ERROR).

martinthomson:
As agreed, this moves the graceful shutdown under the banner of an immediate
close.  That is, the application protocol negotiates a shutdown.  Once the
agreed state has been reached, endpoints send APPLICATION_CLOSE and we are
done.

Closes #854.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1442,9 +1442,8 @@ TODO: see issue #161
 
 Connections should remain open until they become idle for a pre-negotiated
 period of time.  A QUIC connection, once established, can be terminated in one
-of four ways:
+of three ways:
 
-* application close ({{application-close}})
 * idle timeout ({{idle-timeout}})
 * immediate close ({{immediate-close}})
 * stateless reset ({{stateless-reset}})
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1531,6 +1516,13 @@ An endpoint can cease sending CONNECTION_CLOSE or APPLICATION_CLOSE frames if it
 receives either a CONNECTION_CLOSE, APPLICATION_CLOSE or an acknowledgement for
 a packet that contained a either close frame.
 
+An immediate close can be used after an application protocol has arranged to
+close a connection.  This might be after negotiating a graceful shutdown.  The
```
janaiyengar: I would clarify that this negotiation is in the application, not in QUIC.

martinthomson:
Yeah, the work never ends.  But we can only move forward in the hope that it will someday be less urgent to fix anything (at which point we declare victory).

Issue #832, opened 2017-10-03
don't allow 0 as a packet number


marten-seemann:
This would make implementations a lot easier (especially ACK generation).

mikkelfj:
I agree

nibanks:
Personally, I don't think it makes it easier. If we add this restriction, we would have to update our implementation to add **extra special case logic** for packet number randomization to not allow 0, and then on the receive side of things, always check to see if packet number was 0 and drop it.

marten-seemann:
That's a little extra logic (one if for generating and one if for receiving). By allowing 0 with the probability of 2^-31 we lose the ability to check that `highest received packet number != 0`, which will already is problem for several implementations, and will produce some really nasty corner cases in the future.

nibanks:
But why do you need this? Are you just trying to track if you have received any packet? If so, wouldn't a separate flag for that case work just as easily? Or a counter for number of packets received?

> the ability to check that the `highest received packet number != 0`

marten-seemann:
Correct, the alternative would be creating bools to track if the first packet was already received, just to cover an extremely improbable corner case in the packet number randomisation.

nibanks:
If you don't like using a Boolean, keep a counter, which could have other uses, such as a statistical measurement.

ianswett:
Stats are useful, but it'd nice to keep my stats code separate from my framing code.  Marten's right, this is definitely annoying to deal with, probably on the send side as well, since our code initializes packet numbers to 0 everywhere under the thought that 0 is an invalid packet number.  

It's really nice to have an invalid packet number constant, and using 0 as an invalid packet number is a lot better than using some other value in my opinion(ie: 2^64-1)

nibanks:
If we really must have an invalid packet number I would also prefer 2^64-1, as that would never be initialized to, since we use 32 random bits, and we would essentially limit the maximum number allowed by 1. Really, no additional logic required.

ekr:
I agree with @nibanks, but it's worth noting that it's also easy to produce
a value that is nonzero:

33s | <30 random bits> | 1

RyanAtGoogle:
I also completely agree with the idea that 0 is an invalid packet number to
simplify tracking.

janaiyengar:
Zero should not be allowed. Zero is a special number, it should be treated as such.

pravb:
Any special packet number precludes packet number wraparound in future.

RyanAtGoogle:
[...]

PR #864, opened 2017-10-12
intial packet number should be 0.

### LINKED ISSUE #850: Remove initial packet number randomization
With the resolution of #693 we no longer need initial packet numbers to be randomized.

mcmanus:
Closes #850

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -573,9 +573,9 @@ unless it has received a packet from the server.  If the client has received a
 packet from the server, the connection ID field uses the value provided by the
 server.
 
-The first Initial packet that is sent by a client contains a random 31-bit
-value.  All subsequent packets contain a packet number that is incremented by
-one, see ({{packet-numbers}}).
+The first Initial packet that is sent by a client contains packet number 0.  All
+subsequent packets contain a packet number that is incremented by one, see
+({{packet-numbers}}).
 
 The payload of a Initial packet consists of a STREAM frame (or frames)
 for stream 0 containing a cryptographic handshake message, with enough PADDING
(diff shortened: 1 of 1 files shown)


mikkelfj:
It is not clear to me if this increment by one is only in the clear text or it applies to all packet numbers, including 1-RTT?

RyanAtGoogle:
I don't want to pile on, but please, 1 instead of 0. It's convenient in code to have a sentinel value for an invalid packet number and the obvious choices are 0 and MAX_INT64. 0 has the nice virtual of being obvious when printed in base10, whereas MAX_INT64 is just some random big number. In addition, 0 has the property of being less than valid packet number. So this allows you to write code like:

  if current_packet.packet_number > largest_received_packet_:
    UpdateNewLargestPacket()

Without needing to explicitly handle the case where largest_received_packet_ is MAX_INT64.

ianswett:
Agreed, let's make this 1.

ekr:
Note that if we are starting at 0, you can just use int64_t, tear the
connection down at 2^{63} and use -1 as a sentinel.

-Ekr

nibanks:
We already have an issue tracking whether or not to allow 0: #832. I feel the discussion is better served if we keep it in that one place.

mcmanus:
packet number 0 is already legal (the random range is between 0 and 2^31-1). @nibanks is right, if we want to change that do it in another issue - this PR is to remove the randomness.

mcmanus:
@mikkelfj all packet numbers

mcmanus:
`if current_packet.packet_number > largest_received_packet_:
UpdateNewLargestPacket()`

??

if current_packet.packet_number >= next_expected_packet 
 Update()

RyanAtGoogle:
"if current_packet.packet_number >= next_expected_packet 
  Update()"

Ok fine, this is what I get for not using a real example :) >= works for the particular case I described. But there are cases in the code where we may see the same packet number multiple times but only want to do an operation when we've seen a new largest number and not if we're merely seeing the same largest number against. In those cases >= does not work.

marten-seemann:
We would have the same problems in quic-go. Excluding 0 makes the code so much easier, please make it 1.

MikeBishop:
Regardless of the outcome of the 0 vs. 1 debate, this is stale from the updated Packet names from #847.

PR #868, opened 2017-10-13
Enter draining after receiving a stateless reset

### LINKED ISSUE #867: Clients don't drain after stateless reset
I think that this was just an error.  A stateless reset isn't necessary singular: the server might send multiple as it receives multiple packets.  So it makes sense for the client to stick around and absorb packets for a while.

martinthomson:
If a client receives a stateless reset, it makes sense for it to stick around
in the draining period so that you can absorb any further stateless reset
packets that might have been sent by the server.  The server might have
received multiple packets, so it might also generate multiple resets.

Closes #867.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1472,7 +1472,7 @@ state.  This results in new packets on the connection being discarded.  An
 endpoint MAY send a stateless reset in response to any further incoming packets.
 
 The draining period does not apply when a stateless reset ({{stateless-reset}})
-is used.
+is sent.
 
 
 ### Idle Timeout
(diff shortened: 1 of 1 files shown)


ekr:
Hmm... I'm not following this. If the server is sending a stateless reset, that means that the server has lost state. The draining period seems to contemplate that the client would send ACKs and the like (https://github.com/quicwg/base-drafts/pull/868), but that's not useful to the server, because even if it has in-flight packets, it cannot process them. So, all that is necessary is that the client discard further stateless reset packets. Is your concern that the client will be sending ICMP unreachables rather than discarding them silently?

janaiyengar:
The client may have sent an entire RTT's worth of data, and will likely receive more than just one Stateless Reset in response.

janaiyengar:
Sorry -- that wasn't a complete response.

As you say, this just means that the client hangs around gathering any dangling stateless resets in the network. It's avoids strange side-effects, like the ICMP unreachables that you indicate.

ekr:
But you *also* drop packets during the idle period, whereas the draining period is clearly intended per the text to have you send packets.

@janaiyengar @martinthomson: I disagree with this change and I don't think it should have been merged (it's not just within editor's discretion). How would you like to handle this? I can file a new issue or mail the list, or...?

martinthomson:
Please file an issue.  I don't understand your concern, so perhaps you should expand on it some more.

PR #870, opened 2017-10-13
 Tweaks for the draining period

### LINKED ISSUE #869: Exit Draining Period Early
If a client closes the connection (via a CONNECTION_CLOSE or APPLICATION_CLOSE), it immediately enters the draining period. If it receives either an ACK for that packet or a CONNECTION_CLOSE in response, it can assume the server no longer needs any additional information and can immediately exit the draining period and clean up all remaining state (close the socket).

I believe the same would be true if the server originated the close (though reordered packets may arrive afterwards still). In other words, if one side initiates the close, it should be able to immediately clean up local state in response to receiving some acknowledgement that the other side received the close, so long as it can correctly handle any packets that come in afterwards.

The peer that receives the close frame must still go through the draining period though.

martinthomson:
This makes the 3RTO timer advisory (SHOULD).

It also adds a note about leaving early.  If you can be sure that the other
side isn't going to send you more packets indefinitely, then you can end early
if you are OK with receiving reordered or delayed packets.

I also removed a residual mention of sending ACK frames in this state.

Closes #869.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1541,10 +1541,10 @@ peer is also draining.  Receiving a CONNECTION_CLOSE or APPLICATION_CLOSE frame
 is sufficient confirmation, as is receiving a stateless reset.  However,
 disposing of connection state could result in delayed or reordered packets to be
 handled poorly.  For endpoints that have some alternative means to ensure that
-that late arriving packets on the connection are discarded, such as closing the
-UDP port, an abbreviated draining period can allow for faster resource recovery.
-Servers that retain an open port for accepting new connections SHOULD NOT exit
-the draining period early.
+late-arriving packets on the connection do not create QUIC state, such as by
+closing the UDP socket, an abbreviated draining period can allow for faster
+resource recovery.  Servers that retain an open port for accepting new
+connections SHOULD NOT exit the draining period early.
 
 The draining period does not apply when a stateless reset ({{stateless-reset}})
 is sent.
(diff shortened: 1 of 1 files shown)


ekr:
This just seems like a hack for failing to acknowledge that there are times when draining is required and times when it's not. I'd like to see this whole section revisited rather than continuing to add epicycles.

martinthomson:
It sounds like you have higher-level concerns than this.  If you could write them up in an issue, that would be appreciated.

draft-ietf-quic-transport.md
```diff
@@ -1522,20 +1522,30 @@ from its peer.  These packets might have been sent prior to receiving any close
...
 state.  This results in new packets on the connection being discarded.  An
 endpoint MAY send a stateless reset in response to any further incoming packets.
 
+An endpoint MAY exit the draining period earlier if it can guarantee that its
+peer is also draining.  Receiving a CONNECTION_CLOSE or APPLICATION_CLOSE frame
+is sufficient confirmation, as is receiving a stateless reset.  However,
+disposing of connection state could result in delayed or reordered packets to be
+handled poorly.  For endpoints that have some alternative means to ensure that
+that late arriving packets on the connection are discarded, such as closing the
+UDP port, an abbreviated draining period can allow for faster resource recovery.
```
janaiyengar: (i) Closing the "port" makes no sense, you probably meant to say "socket".
(ii) Closing the port doesn't ensure that they;re discarded -- this can result in an ICMP message being generated in response. Perhaps say something like "to ensure that late-arriving packets on the connection do not create QUIC state, such as by closing the UDP socket"

draft-ietf-quic-transport.md
```diff
@@ -1522,20 +1522,30 @@ from its peer.  These packets might have been sent prior to receiving any close
...
+Servers that retain an open port for accepting new connections SHOULD NOT exit
+the draining period early.
```
janaiyengar: This sentence is good.

PR #872, opened 2017-10-13
Identify streams unidirectional vs bidirectional based on stream ID

RyanAtGoogle:
Identify streams unidirectional vs bidirectional based on stream ID, much like how client vs server created is identified.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2542,7 +2542,7 @@ shown in the following figure and described below.
    recv:   endpoint receives this frame
 
    STREAM: a STREAM frame
-   UNI:    a unidirectional stream
+   UNI:    a STREAM frame with a unidirectional stream ID
    FIN:    FIN flag in a STREAM frame
    RST:    RST_STREAM frame
    MSD:    MAX_STREAM_DATA frame
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2630,7 +2630,8 @@ A stream that is in the "half-closed (local)" state MUST NOT be used for sending
 on new STREAM frames.  Retransmission of data that has already been sent on
 STREAM frames is permitted.  An endpoint MAY also send MAX_STREAM_DATA and
 STOP_SENDING in this state. Unidirectional streams created by the peer are
-immediately half-closed (local) to the receiver.
+immediately half-closed (local) to the receiver and are subsequently treated
+identically to a half-closed bidirectional stream.
 
 An application can decide to abandon a stream in this state. An endpoint can
 send RST_STREAM for a stream that was closed with the FIN flag. The final offset
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1857,8 +1857,10 @@ The frame is as follows:
 The fields in the MAX_STREAM_ID frame are as follows:
 
 Maximum Stream ID:
-: ID of the maximum unidirectional or bidirectional (based on the second least
-  signification bit of the stream id) peer-initiated stream ID for the connection.
+: ID of the maximum unidirectional or bidirectional peer-initiated stream ID
+  for the connection. The limit applies to unidirectional steams if the second
+  least signification bit of the stream id is 0, and applies to bidirectional
+  streams if it is 1.
 
 Loss or reordering can mean that a MAX_STREAM_ID frame can be received which
 states a lower stream limit than the client has previously received.
@@ -2442,7 +2444,7 @@ Streams in QUIC provide a lightweight, ordered, and bidirectional byte-stream
 abstraction modeled closely on HTTP/2 streams {{?RFC7540}}.
 
 Streams can be created either by the client or the server, can concurrently send
-data interleaved with other streams, and can be cancelled, and can be opened in
+data interleaved with other streams, can be cancelled, and can be opened in
 unidirectional mode.
 
 Data that is received on a stream is delivered in order within that stream, but
@@ -2654,7 +2656,7 @@ after a frame bearing the FIN flag is sent.
 A stream is "half-closed (remote)" when the stream is no longer being used by
 the peer to send any data.  An endpoint will have either received all data that
 a peer has sent or will have received a RST_STREAM frame and discarded any
-received data. Unidirectional streams created by the locally are immediately
+received data. Unidirectional streams created locally are immediately
 half-closed (remote) to the creator.
 
 Once all data has been either received or discarded, a sender is no longer
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1859,8 +1859,8 @@ The fields in the MAX_STREAM_ID frame are as follows:
 Maximum Stream ID:
 : ID of the maximum unidirectional or bidirectional peer-initiated stream ID
   for the connection. The limit applies to unidirectional steams if the second
-  least signification bit of the stream id is 0, and applies to bidirectional
-  streams if it is 1.
+  least signification bit of the stream id is 1, and applies to bidirectional
+  streams if it is 0.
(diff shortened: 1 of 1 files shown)


RyanAtGoogle:
This identifies unidirectional vs bidirectional streams by a bit in the stream ID, just like we do for client vs server. This avoids the need to pass around directionality meta-data in the various places we use stream id (STREAM frames, MAX_STREAM_ID, MAX_STREAM_DATA, etc) and means that stream id is always a unique identifier of a single stream.

draft-ietf-quic-transport.md
```diff
@@ -2525,6 +2542,7 @@ shown in the following figure and described below.
    recv:   endpoint receives this frame
 
    STREAM: a STREAM frame
+   UNI:    a unidirectional stream
```
janaiyengar: nit: a STREAM frame with a unidirectional stream ID

draft-ietf-quic-transport.md
```diff
@@ -2544,6 +2562,12 @@ coordinate the creation of streams; they are created unilaterally by either
 endpoint.  Endpoints can use acknowledgments to understand the peer's subjective
 view of stream state at any given time.
 
+Only the initiator of a stream may send a STREAM frame without receiving one
+unless otherwise specified by the application.  All implicitly opened streams
```
janaiyengar: This sentence is a bit confusing. I don't think we want to allow applications to override stream ID semantics, do we?

RyanAtGoogle: Not sure, I copied this language from Ian's PR on this topic. Perhaps I should just nuke the whole paragraph?

draft-ietf-quic-transport.md
RyanAtGoogle: Done.

draft-ietf-quic-transport.md
ianswett: Yes, this was an effort to fix a problem with my design that you do not have, so I'd nuke it.

RyanAtGoogle: Done.

martinthomson:
Firstly, thanks for this PR, I think that it is almost exactly what we need.  I have two concerns, the first of which we should resolve before merging this.

The encoding of type into the low bits of the identifier is problematic.  I would prefer that we encode an explicit type field into the stream ID encoding.  The number of bits doesn't change, but the way that we identify the streams could.  I suspect that this is going to generate a ton of angry emails, in the same way that suggesting that 0 being special did, but I find the encoding of the type into the low bits to make certain operations more difficult than is ideal.

If you look at the proposed change in #877, I would define a modified stream ID encoding that steals two bits before the length bits.  It's less uniform in terms of the integer encoding, but you have a predictable location to look for the stream type.

The state machine here needs some serious work, but that can be deferred.  I would prefer that we wait for someone to transcribe the state machine that Mike shared in the Seattle interim.

mikkelfj:
While I am also concerned with the complexity of adding stream type into the ID (it adds a lot of special cases, but likely most alternatives would too), I don't think it is a good idea to have several ways of decoding varlen integers. It is best to have a single function to deal with endianness and unaligned read access and decoding, rather than several implementations.

If the bits are to be relocated, they could be placed in the frame type which can the be placed in the upper bits after decoding. But unfortunately that adds a lot of sub-frame types as Ian pointed out. Still, I think I prefer this. It makes it possible to immediately branch out on stream type.

martinthomson:
@mikkelfj, I originally wanted to move the bits to the frame type; but you would need three of each type of frame that mentions streams and there are 6 frame types to modify in that way.  On balance, moving the identification into the same slot that we encode the stream ID as @RyanAtGoogle proposes might be better. That said, the HTTP mapping only uses certain specific types of streams, so maybe it could work out.

I forgot one comment: this PR doesn't address the initial values for MAX_STREAM_ID and MAX_STREAM_DATA.  I think that we should have separate configuration for those.  That suggests that maybe a separate frame type isn't so bad an idea.

janaiyengar:
@martinthomson On using the high-order bits instead of the low-order ones, I don't have a strong opinion, but it's not complicated to use low-order ones. We've done that for client/server streams, this merely extends it.

I'm happy with the PR (modulo adding default MAX_STREAM_ID and MAX_STREAM_DATA separately for uni and bi streams), and I think we can make further changes after we get this one in.

mikkelfj:
I also believe it is simpler to keep bits low rather than having a non-standard varlen integer format.

ekr:
I'm generally OK with this strategy, but I agree with MT that I'd rather use the high-order bits of the stream ID as control bits (for both the direction and for the bi/uni switch). This just gives you four spaces that you can talk about sensibly.

janaiyengar:
You get four spaces either way -- you can use the high-order bits or the low-order ones. ... and we'd lose the high-order bits to size if we went the direction of PR 877. In the context of that PR, I definitely prefer low-order bits.

MikeBishop:
Yes, if we go with the direction of #877, the type bits need to be held separately from the Stream ID.  I agree that's challenging, but doesn't seem insurmountable, and feels cleaner than a mod-4 space.  I never liked the even-odd approach to begin with, I didn't like the mod-4 structure when it was in the HTTP/QUIC draft, and I really don't want to see mod-4 invading the QUIC transport.

That said, it's a bikeshed.  The bits exist, and exactly where they land in the frame doesn't ultimately change the state model of the protocol.

martinthomson:
So that I don't forget this... This PR omits two very important things:

1. It doesn't actually say what a unidirectional stream *is*.

2. It doesn't say anything about stream usage in HTTP.

As we discussed on the editor's call today, we are going to merge this, then fix it.  I'll prioritize fixing the first issue though.

Given the HTTP usage, I'm going to push harder for a separation of the type bits, because that will make the HTTP usage of the identifiers easier (it has implicit type identification).  That will have to wait until #877 resolves.  That also needs some HTTP text.

draft-ietf-quic-transport.md
```diff
@@ -2472,6 +2476,18 @@ corresponds to a stream which is allocated to it (i.e., odd-numbered for
...
+and bidirectional streams. Unidirectional streams always have this bit set to
+1 and bidirectional streams have this bit set to 0. As a result the initial
+stream ID for various stream types is listed below:
+
+| Stream ID   | Type                          |
+|:------------|:------------------------------|
+| 0x01        | Client Bidirectional          |
+| 0x02        | Server Bidirectional          |
+| 0x03        | Client Unidirectional         |
+| 0x04        | Server Unidirectional         |
```
martinthomson: Hmm, this table is wrong too.

RyanAtGoogle:
On Tue, Oct 17, 2017 at 4:33 PM, Martin Thomson <notifications@github.com>
wrote:

> So that I don't forget this... This PR omits two very important things:
>
>    1.
>
>    It doesn't actually say what a unidirectional stream *is*.
>
> ​Doh.​
​ I'm not awesome at spec language. I tried to define this implicitly by
saying that "

​"​
Unidirectional streams created by the peer are
​ ​
immediately half-closed (local) to the receiver and are subsequently treated
​ ​
identically to a half-closed bidirectional stream.
​"​
​

But that's probably not terribly useful. Have a suggestion on test?

>
>    1.
>    2.
>
>    It doesn't say anything about stream usage in HTTP.
>
> As we discussed on the editor's call today, we are going to merge this,
> then fix it. I'll prioritize fixing the first issue though.
>

​Are you planning to merge this as is, or shall I address some of the other
feedback (the table, for example).​

martinthomson:
Hi Ryan, maybe you could provide feedback on #885.  If we can't agree on that, then maybe we need to defer the change until we do agree.

janaiyengar:
I think this is good to go. We have agreement on #885, so let's land this and then land that one. I'll wait 24 hours and then merge.

PR #874, opened 2017-10-13
Express MAX_DATA in octets rather than KB

### LINKED ISSUE #873: Limit MAX_DATA to the integer range (64 or 62 depending..)
dimitry and @martinthomson  had a bit of a discussion on this in August: https://www.ietf.org/mail-archive/web/quic/current/msg02117.html

tl;dr - max-data is effectively a 74 bit integer.. plausibly much bigger than required, extremely hard to do integration testing with so presumably buggy as its > common native types.. at terrabit rates it would take years to transfer 64/62 bits (the obvious alternative) worth of flow controlled bytes.

I think this is an architectural issue because the bug risk for interop failure in long deployed systems is significant.

The existing arrangement falls short of accommodating the full sized packets across the whole packet number space anyhow. (to say nothing of dreaming of bigger MTUs) If we're going to keep it supersized then +16 bits makes a lot more sense to me than +10.

mcmanus:
closes #873

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1829,9 +1829,7 @@ The fields in the MAX_DATA frame are as follows:
 Maximum Data:
 
 : A 64-bit unsigned integer indicating the maximum amount of data that can be
-  sent on the entire connection, in units of 1024 octets.  That is, the updated
-  connection-level data limit is determined by multiplying the encoded value by
-  1024.
+  sent on the entire connection, in units octets.
 
 All data sent in STREAM frames counts toward this limit, with the exception of
 data on stream 0.  The sum of the largest received offsets on all streams -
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1829,9 +1829,7 @@ The fields in the MAX_DATA frame are as follows:
 Maximum Data:
 
 : A 64-bit unsigned integer indicating the maximum amount of data that can be
-  sent on the entire connection, in units of 1024 octets.  That is, the updated
-  connection-level data limit is determined by multiplying the encoded value by
-  1024.
+  sent on the entire connection, in units octets.
```
martinthomson: of ?

mcmanus: imagine how many mistakes I can make in a diff more than 1 line long.

siyengar:
wohoo

siyengar:
What about initial_max_data in the extension?

mcmanus:
are you suggesting it needs to be more than 32 as an initial value? That
sort of stretches the credulity of flow control...

siyengar:
no i'm suggesting it doesn't need to have a 1024 multiplier.

RyanAtGoogle:
+1! Let's just use octets.

PR #881, opened 2017-10-17
Connection ID is big endian

ekr:


martinthomson:
How does this fit with #833?  It seems to presume an outcome.

ekr:
@martinthomson my intent here was just to clarify the wire encoding while we fought over #833. I am still in favor of merging #833.

Issue #627, opened 2017-06-13
Implicit acknowledgment during handshake


ekr:
The motivation for this issue is: What happens if you receive a ServerCleartext packet that contains SH but *doesn't* contain the acknowledgement for CI? Should you eventually retransmit CI? This doesn't make sense, but the protocol suggests you should.

We could just make it an exception, but I think we might want to go bigger. Specifically, I propose we take the data in ClientInitial out of stream 0 and make it special data. It is anyway because it can't be fragmented, has to be padded, etc. Then the client's second flight would go in stream 0 at offset 0.

mikkelfj:
Related question - why does the client send end of handshake in a cleartext packets after receiving server handshake, except for retransmission of earlier packets no longer needed.

QUIC TLS draft section 4.1:

>  @C QUIC STREAM Frame(s) <0>:
        (EndOfEarlyData)
        {Finished}

ekr:
It avoids double-encryption of the TLS handshake packets.

mikkelfj:
> It avoids double-encryption of the TLS handshake packets.

I may get the following wrong, as I haven't studied everything in detail yet, but my take is, ignoring 0RTT:

If the server was required to only respond with a single packet, and double encryption was acceptible (which I think is not problem at all), then the handshake could have the form:

@c ClientInitial ->
<-c ServerInitial
@1-> stream 0{EndOfEarlyData, Finished}, Optionally other streams ->

No ACKS in server to client, no acks in client to server.

If these initial packets are made special, as you suggest with ClientInitial, we can also drop packet numbering of these.

This also avoids the issue with ACKS of untrusted packets during handshake: https://github.com/quicwg/base-drafts/issues/624

mikkelfj:
In addition, I suggest that ServerInitial sends reflects the Client Connection ID (CCID) and ALSO sends a new Server Connection ID (SCID). This uniquely links the client request to the server response, and also handles Stateless Retry and Version Negotiation packet responses if these do the same thing, mirroring CCID except these other packets won't have a SCID. In addition, this would avoid having to rely on the uniqueness of a 5-tuple for setting up and maintaining connection state.

ekr:
"If the server was required to only respond with a single packet..."

This is not a reasonable assumption. Certificates regularly push you over the MTU.

mikkelfj:
> This is not a reasonable assumption. Certificates regularly push you over the MTU.

Could this not also happen to client certificates when server requires auth?

[...]

PR #882, opened 2017-10-17
require ACKs and allow PADDING in the Server Stateless Retry packet

### LINKED ISSUE #879: RTT Measurement for HelloRetryRequest
In #817, @marten-seemann asks why the server can't send an ACK with its HelloRetryRequest.

I had assumed that this ACK would be pointless.  Obviously, the server can only acknowledge a single packet (the Initial packet), and this is implicitly acknowledged.  The only information that the packet might now contain is the ACK Delay field now that we removed the timestamps from the default ACK format.  That assumes that the server is delaying its response in any way or that its processing is somehow significant enough to be worth measuring.  I don't think that the server has any reason to delay the generation of a HelloRetryRequest, and the entire design depends on the processing time required to generate it is minimal.

Did I miss something?

### LINKED ISSUE #1067: Retry packet and PADDING
-08 in Section 5.4.2. says:

>   The payload of the Retry packet contains a single STREAM frame on
>   stream 0 with offset 0 containing the server's cryptographic
>   stateless retry material.  It MUST NOT contain any other frames.  The
>   next STREAM frame sent by the server will also start at stream offset
>   0.

Any reason why PADDING is not allowed? (Insert usual stuff about zero-copy and fixed offsets.)

marten-seemann:
Closes #1067. Updates #627.


ekr:
I'm not in favor of this for the reasons discussed in #879, so we should discuss before merging.

draft-ietf-quic-transport.md
```diff
@@ -627,10 +627,11 @@ any version negotiation that occurred (see {{version-negotiation}}).  The client
 MAY also retain any observed RTT or congestion state that it has accumulated for
 the flow, but other transport state MUST be discarded.
 
-The payload of the Retry packet contains a single STREAM frame
-on stream 0 with offset 0 containing the server's cryptographic stateless retry
-material. It MUST NOT contain any other frames. The next STREAM frame sent by
-the server will also start at stream offset 0.
+The payload of the Retry packet contains a STREAM frame on
```
nibanks: I don't think the first two lines here should be changed. Maybe this is just out of date?

MikeBishop:
This PR needs to be re-targeted to cover the new scope of #627, but it's otherwise good.

janaiyengar:
Ugh, sorry @marten-seemann -- I meant to send a patch to your commit, but ended up committing to your branch instead. Take a look, and respond if you think it's fine. I'm happy to merge if you agree.

draft-ietf-quic-transport.md
```diff
@@ -627,10 +627,11 @@ any version negotiation that occurred (see {{version-negotiation}}).  The client
...
 the flow, but other transport state MUST be discarded.
 
-The payload of the Retry packet contains a single STREAM frame
-on stream 0 with offset 0 containing the server's cryptographic stateless retry
-material. It MUST NOT contain any other frames. The next STREAM frame sent by
-the server will also start at stream offset 0.
+The payload of the Retry packet contains at least two frames. It MUST include a
+STREAM frame on stream 0 with offset 0 containing the server's cryptographic
+stateless retry material. It MUST also include an ACK frame to acknowledge the
+client's Initial packet. It MAY additionally include one ore more PADDING
```
marten-seemann: s/ore/or

marten-seemann:
LGTM.

PR #884, opened 2017-10-18
Fix nits

### LINKED ISSUE #883: Editorial nits in quic-tls draft
Some things that were unclear when I tried to cleartext encryption from the -07 draft. I am more or less a TLS novice, so YMMV:

1) Section 5.2.1: For the cleartext secret, it would be useful to state quite clearly here that the connection ID is the "wire image" or "network order" version of connection ID; implementers should not change to host order, etc before calling HKDF-Extract.

2) Section 5.2.4: "the IV length is the larger of 8 or N_MIN (see section 4 of RFC 5116)".  This is super user-unfriendly. At a minimum, we should state that the N_MIN for AES_128_GCM (needed for cleartext) is 12 bytes. At best, we should just list the IV size for supported ciphers.

3) Section 5.3: "Prior to TLS providing keys, no record protection is performed and the plaintext, P, is transmitted unmodified." I believe this paragraph is obsolete.

martinthomson:
Remove text about no protection.
Describe the value for N_MIN.

Closes #883.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -113,7 +113,7 @@ code and issues list for this draft can be found at
 
 
 This document describes how QUIC {{QUIC-TRANSPORT}} is secured using
-Transport Layer Security (TLS) version 1.3 {{!I-D.ietf-tls-tls13}}.  TLS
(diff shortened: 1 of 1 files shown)


martinthomson:
I agree, we need to make sure that it is perfectly clear.

Issue #175, opened 2017-01-18
Unidirectional streams


martinthomson:
QUIC borrows the HTTP/2 state machine, or tries to replicate it as much as it can.  Part of that replication is the idea that a stream is bidirectional.  The state related to outgoing data and the state related to inbound data are tied.

#174 suggests a simplification that could open up a more dramatic simplification: independent state for each direction.  #171 moves in that direction also.

The state machine that results from that is dramatically simpler: idle -> open -> closed.  That is, before packets, packets and after packets.

The consequence for the HTTP mapping is largely that it needs a different method for correlating request and response.

## Correlating request and response

Currently, request and response are correlated by matching the stream identifier.  However, unidirectional streams would allow for cases where streams are not created symmetrically.  For instance, server push would only require streams from the server.  Similarly, the server would not be obligated to respond to requests in the order in which they were received (if at all, since it could reset request streams).

That implies that this would add a requirement for the server to identify which request a given stream is answering.  This is an extension of the requirement in #174, which only required that the server identify a push promise (actually, it's less onerous, because you only have one request per stream).

I can imagine an arrangement where a stream containing headers identifies the stream containing the request headers from the client.  Then the stream containing the body identifies the stream that contained the headers.

## Saving stream identifiers

This is a minor point, though it could be relevant for very long-lived connections.

QUIC currently insists on a strictly monotonically increasing stream identifier: stream N is followed by stream N+2 always, no skipping of streams is permitted.  Aside from wasting every other stream identifier, other streams are wasted by the mapping. 

GET requests don't include a body, but nonetheless require that the client send a HEADERS on stream N, then a FIN on stream N+2.  Including a reference to the initiating stream at the start of the body of a message would use some octets, but potentially fewer than would be used otherwise (particularly if the reference was a varlen encoding of the difference in stream numbers, which would rarely be anything other than 1).

Obviously, this is an optimization for the case where there are no bodies, but that is a case where the optimization is particularly valuable.  Making more requests (which more frequently have no bodies) during the early stages of a connection is critical to performance.

mnot:
Discussed in Tokyo; people need to see a more concrete proposal. @martinthomson to noodle on it.

RyanAtGoogle:
[...]

PR #885, opened 2017-10-18
 Unidirectional and bidirectional streams

martinthomson:
This is #872 with some additional exposition and one significant change.

Client-initiated streams are now EVEN.

Server initiated streams are now ODD.

This is a pretty noticeable change, because it inverts the current sign, and
might cause some disruption in implementations.  However, it makes stream 0
work within the framework better.  It wasn't clear what Ryan's intent was with
respect to numbering, but I wouldn't have noticed this until I noticed a
conflict between prose and the table.

Closes #643, #656, #720, #872, #175.

This does not close the other issues that were addressed by #643, because this
doesn't include any changes in the HTTP draft.  I'll create a follow-up PR for
that.

(We can argue about bit placement later, I'm just trying to make this as good as it can be.)

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3178,8 +3178,8 @@ is increased.
 
 The final offset is the count of the number of octets that are transmitted on a
 stream.  For a stream that is reset, the final offset is carried explicitly in
-the RST_STREAM frame.  Otherwise, the final offset is the offset of the end of
-the data carried in a STREAM frame marked with a FIN flag, or 0 in the case of
+a RST_STREAM frame.  Otherwise, the final offset is the offset of the end of the
+data carried in a STREAM frame marked with a FIN flag, or 0 in the case of
 incoming unidirectional streams.
 
 An endpoint will know the final offset for a stream when the stream enters the
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2739,7 +2739,7 @@ shown in the following figure and described below.
                                  v
                  recv FIN/  +--------+    send FIN/
                  recv RST   |        |    send RST
-                 recv UNI   |        |    send UNI
+                 send UNI   |        |    recv UNI
                   ,---------|  open  |-----------.
                  /          |        |            \
                 v           +--------+             v
@@ -3179,7 +3179,7 @@ is increased.
 The final offset is the count of the number of octets that are transmitted on a
 stream.  For a stream that is reset, the final offset is carried explicitly in
 the RST_STREAM frame.  Otherwise, the final offset is the offset of the end of
-the data carried in STREAM frame marked with a FIN flag, or 0 in the case of
+the data carried in a STREAM frame marked with a FIN flag, or 0 in the case of
 incoming unidirectional streams.
 
 An endpoint will know the final offset for a stream when the stream enters the
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2687,8 +2687,7 @@ identifier.
 
 If an endpoint receives a frame for a stream that it expects to initiate (i.e.,
 odd-numbered for the client or even-numbered for the server), but which it has
-not yet opened but which it has not yet created, it MUST close the connection
-with error code STREAM_STATE_ERROR.
+not yet opened, it MUST close the connection with error code STREAM_STATE_ERROR.
 
 The second least significant bit (0x2) of the Stream ID differentiates between
 unidirectional streams and bidirectional streams. Unidirectional streams always
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2715,7 +2715,7 @@ same direction being counted as open.
 
 Stream IDs are usually encoded as a 32-bit integer, though the STREAM frame
 ({{frame-stream}}) permits a shorter encoding when the leading bits of the
-stream ID are zero.
+Stream ID are zero.
 
 
 ## Life of a Stream {#stream-states}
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2635,12 +2638,18 @@ Too Big messages.
...
+
+There are two basic types of stream in QUIC.  Unidirectional streams carry data
+in one direction only; bidirectional streams allow for data to be sent in both
+directions.  Different stream identifiers are used to distinguish between
+unidirectional and bidirectional streams, as well as to create a separation
+between streams that are initiated by the client and server (see {{stream-id}}).
 
-Streams can be created either by the client or the server, can concurrently send
+Either type of stream can be created by either endpoint, can concurrently send
 data interleaved with other streams, and can be cancelled.
```
mikkelfj: Not entirely clear which endpoint can cancel.

draft-ietf-quic-transport.md
```diff
@@ -2665,15 +2674,39 @@ for some applications.
...
+| 0x0      | Client-Initiated, Bidirectional  |
+| 0x1      | Server-Initiated, Bidirectional  |
+| 0x2      | Client-Initiated, Unidirectional |
+| 0x3      | Server-Initiated, Unidirectional |
+{: #stream-id-types title="Stream ID Types"}
 
-Stream ID 0 (0x0) is reserved for the cryptographic handshake.  Stream 0 MUST
-NOT be used for application data, and is the first client-initiated stream.
+Stream ID 0 (0x0) is a client-initiated, bidirectional stream that is used for
+the cryptographic handshake.  Stream 0 MUST NOT be used for application data.
```
mikkelfj: handshake and subsequent key exchanges?

draft-ietf-quic-transport.md
```diff
@@ -2706,6 +2739,7 @@ shown in the following figure and described below.
                                  v
                  recv FIN/  +--------+    send FIN/
                  recv RST   |        |    send RST
+                 recv UNI   |        |    send UNI
                   ,---------|  open  |-----------.
```
mikkelfj: Isn't send and recv UNI swapped?

draft-ietf-quic-transport.md
```diff
@@ -3141,7 +3179,8 @@ is increased.
 The final offset is the count of the number of octets that are transmitted on a
 stream.  For a stream that is reset, the final offset is carried explicitly in
 the RST_STREAM frame.  Otherwise, the final offset is the offset of the end of
-the data carried in STREAM frame marked with a FIN flag.
+the data carried in STREAM frame marked with a FIN flag, or 0 in the case of
+incoming unidirectional streams.
```
mikkelfj: in _the_ STREAM?

draft-ietf-quic-transport.md
martinthomson: That's not important at this point.

draft-ietf-quic-transport.md
martinthomson: As above.

draft-ietf-quic-transport.md
mikkelfj: Detail, but the RST_STREAM just above is referenced as _the_ RST_STREAM - is that intentional?

martinthomson:
@ianswett:
> I'm a bit worried this change will create a lot of confusion for people familiar with HTTP/2. The alternative would be to make the crypto stream 1 again.(Yes, I know you probably hate that, but it is simple).

I realize that, but then we're already there.  Rather than requests on every other stream, we have every fourth stream.  Better to have the clean break in my opinion.  Also, the number of uses for stream IDs in HTTP is limited, and could even be eliminated entirely, depending on how we manage header compression.

RyanAtGoogle:
If we're revisiting stream numbering I'll make a plea to release stream id 0 from being the crypto stream and instead make it an invalid stream id. It's a tad annoying that sometimes stream_id=0 means crypto stream (in a STREAM frame) and sometimes it mean the whole connection (MAX_STREAM_DATA). So if we made the crypto stream 1, then we'd be able to keep the current even/odd ness and eliminate the minor wart around stream_id 0.

What do you think?

mikkelfj:
@RyanAtGoogle In general I prefer 0 to be invalid and as such I am in line with your thinking. However, I did find crypto 0 to be helpful because any stream that is not crypto is handled via normal life time requirements whereas the crypto stream is sent to dedicated handlers. Unlike largest received packet number, the crypto stream always exists and any inbound stream will either be stream 0, or something to be validated against stream limits etc.

martinthomson:
MAX_STREAM_DATA with a Stream ID of 0 is perfectly valid.  It sets the flow control for stream 0.  I think that you are confusing that with MAX_DATA.  We dispensed with 0 as a special value for that a long time ago.

The special-ness for stream 0 punches right through any issues of frames or encoding; those are all perfectly consistent with other frames.  It's the packet protection rules that are crazy for stream 0.

RyanAtGoogle:
Ah, yes good point. Seems reasonable to me then.

janaiyengar:
Thanks for writing this up -- this PR LGTM.

draft-ietf-quic-transport.md
```diff
@@ -2665,15 +2674,39 @@ for some applications.
...
+The least significant bit (0x1) of the Stream ID identifies the initiator of the
+stream.  Clients initiate even-numbered streams (those with the least
+significant bit set to 0); servers initiate even-numbered streams (with the bit
+set to 1).  Separation of the stream identifiers ensures that client and server
+are able to open streams without the latency imposed by negotiating for an
+identifier.
+
+If an endpoint receives a frame for a stream that it expects to initiate (i.e.,
+odd-numbered for the client or even-numbered for the server), but which it has
+not yet opened but which it has not yet created, it MUST close the connection
```
afrind: Some duplicate text here 'but which it has not yet opened but which it has not yet created'

PR #877, opened 2017-10-16
Integers

martinthomson:
This isn't quite ready - I have to make some changes in the HTTP doc, but I thought that I'd share it now so that we have a better location for discussion.

This proposal replaces all integer encodings in frames with a unified variable-length encoding format.  It simplifies the encoding considerably, most noticeably in the ACK frame, which I have largely rewritten.

Closes #595, #109.

PR #887, opened 2017-10-18
Minimal change for HTTP integers

martinthomson:
I don't especially like this version, but it is the easiest change to HTTP after we take #877.


MikeBishop:
Replaced by #888.

PR #888, opened 2017-10-18
Maximal change for HTTP integer encoding

martinthomson:
This didn't turn out to be that disruptive.  I'm somewhat worried that I missed
something actually.  Given that the changes are small, I think that I would
prefer this over #887 for #877.

The most interesting change here is that extension frame types that mention
Stream IDs will have to change.  I am not aware of any proposed thus far, so
maybe it's not too much of a cost to bear.  It further confirms my view that we
shouldn't be using Stream IDs at this layer anyway.

Based on this, I don't see any need for a middle-ground change.  The biggest
cost here is in taking the larger Stream ID space, not in taking the integer
encoding.  Worse, I just noticed that PRIORITY would need a lot more work to
split if we only took the change for Stream ID, so the difference would be tiny.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -358,8 +358,8 @@ allows a server to fulfill promises in the order that best suits its needs.
 The server push response is conveyed on a push stream.  A push stream is a
 server-initiated stream.  A push stream includes a header (see
 {{fig-push-stream-header}}) that identifies the PUSH_PROMISE that it fulfills.
-This header consists of a 32-bit Push ID, which identifies a server push (see
-{{frame-push-promise}}).
+This header consists of a Push ID, encoded as a variable-length integer.  The
+Push ID identifies a server push (see {{frame-push-promise}}).
 
 ~~~~~
  0                   1                   2                   3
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -569,8 +578,9 @@ that is being cancelled (see {{frame-push-promise}}).
 If the client receives a CANCEL_PUSH frame, that frame might identify a Push ID
```
ianswett: There's a reference to a 32 bit Push ID three lines above this one.  I think that needs to be updated as well?

draft-ietf-quic-http.md
```diff
@@ -966,6 +976,13 @@ applicable to both protocols at once.
 
 These departures are noted in this section.
 
+## Streams {#h2-streams}
+
+HTTP/QUIC permits use of a larger number of streams (2^62-1) then HTTP/2.  The
+considerations about exhaustion of stream identifier space apply, though the
+space is significantly larger such that it is likely that other limits in QUIC
+are reached first, such asthe limit on the connection flow control window.
```
ianswett: asthe -> as the

draft-ietf-quic-http.md
```diff
@@ -701,9 +708,9 @@ server to client, as in HTTP/2.  The PUSH_PROMISE frame defines no flags.
 The payload consists of:
 
 Push ID:
-: A 32-bit identifier for the server push request.  A push ID is used in push
-  stream header ({{server-push}}), CANCEL_PUSH frames ({{frame-cancel-push}}),
-  and PRIORITY frames ({{frame-priority}}).
+: A variable-length integer that identifies the server push request.  A push ID
```
ianswett: It'd be nice if we didn't have to call it a variable length integer everywhere.  The encoding is variable length, but I think of the integer being encoded as having a fixed length of 62 bits.  I'm not sure it's better to say a 62 bit integer or not.

draft-ietf-quic-http.md
martinthomson: Actually, there were two, thanks.

draft-ietf-quic-http.md
martinthomson: I think that this is OK.  XXX-bit and variable-length occupy the same mental space and the later is more correct.  Substitute 62-bit if that works for you, but I don't want to have to explain that 62-bit means the variable-length encoding.

draft-ietf-quic-http.md
martinthomson: Or three; I missed your example, but fixed two others...

PR #891, opened 2017-10-19
Allow handshake errors to be reported with CONNECTION_CLOSE

### LINKED ISSUE #608: Handshake errors
If the handshake fails, how do we signal that?  We could send CONNECTION_CLOSE in the clear...

If version negotiation fails, how do we signal that?  I don't think that we can, though we currently require that.

martinthomson:
Closes #608.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1289,9 +1289,10 @@ handshake completes.
 
 ### Handshake Failures
 
-The `CONNECTION_CLOSE` frame MAY be sent by either endpoint during the handshake
-to signal a fatal error with connection establishment.  A `STREAM` frame
-carrying a TLS alert MAY be included in the same packet.
+The `CONNECTION_CLOSE` frame MAY be sent by either endpoint in a Handshake
+packet.  This allows an endpoint to signal a fatal error with connection
+establishment.  A `STREAM` frame carrying a TLS alert MAY be included in the
+same packet.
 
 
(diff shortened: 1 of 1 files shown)


RyanAtGoogle:
Nice. LGTM

draft-ietf-quic-tls.md
```diff
@@ -1287,6 +1287,13 @@ Similarly, there is no need to increase the number of allowed streams until the
 handshake completes.
 
 
+### Handshake Failures
+
+The `CONNECTION_CLOSE` frame MAY be sent by either endpoint during the handshake
+to signal a fatal error with connection establishment.  A `STREAM` frame
+carrying a TLS alert MAY be included in the same packet.
+
```
huitema: Maybe we should add a note in the next section, "Denial of Service with Unprotected Packets". Sending "connection close" frames in Server Clear Text or Client Clear Text packets would be another way to deny connection.

martinthomson: I'll do that separately: see https://github.com/quicwg/base-drafts/pull/908/files

martinthomson:
@seanturner, PTAL

MikeBishop:
Why is this in the TLS draft?  If we're allowing CONNECTION_CLOSE in Handshake packets, shouldn't that be in the Transport doc?

martinthomson:
The TLS doc contains all the packet protection stuff, which includes rules about what can be sent and when.  The plan is to move this text to -transport eventually.

PR #894, opened 2017-10-20
More detailed stream states

### LINKED ISSUE #634: Stream creation order
Dmitri Tikhonov writes on the mailing list:

"Section 10.1 of draft-ietf-quic-transport-03 states that “Streams MUST be created in sequential order.”  My question is: why mandate stream creation order, since the other side may receive out-of-order packets?  Let the implementation do what it wants with Stream IDs – as long as it does not go over the limit, all should be well.
 
This phrase is a holdover from the previous version; this requirement seems to be unnecessary since issue #435 has been resolved."

I think this is right. Now that we have also replaced concurrent streams, this seems quite unnecessary. This will clean up some more text, but I wanted to air this issue here before turning it into a PR.

martinthomson:
This takes the state machines that @MikeBishop showed at the Seattle interim
and turns them into text and ASCII art.

I made a few changes in the process.  The state where all data is available
didn't turn out to be any different to the state preceding it, so I removed it.
I also found it valuable to recognize that streams might be created for sending
before any frames are sent (I considered doing the same for receive streams,
but it complicated the transitions more than I liked, I could be convinced to
add this still, it just seems less valuable).

I added a section on composite states, showing how this mess all maps to stuff
you already know (h2).

This is still work-in-progress.  I have changed the main chunk of text on
states, but I haven't changed any of the rest of the document, which still
talks about "idle" and all that.  I'll do that next if people are generally
happy with the direction this is headed.

There are probably a few minor regressions here.  In particular, I need to look at
what retransmissions can be received after they are no longer needed.  But it's
late enough already.

Closes #634.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2032,11 +2032,11 @@ Maximum Data:
 
 All data sent in STREAM frames counts toward this limit, with the exception of
 data on stream 0.  The sum of the largest received offsets on all streams -
-including closed streams, but excluding stream 0 - MUST NOT exceed the value
-advertised by a receiver.  An endpoint MUST terminate a connection with a
-QUIC_FLOW_CONTROL_RECEIVED_TOO_MUCH_DATA error if it receives more data than the
-maximum data value that it has sent, unless this is a result of a change in the
-initial limits (see {{zerortt-parameters}}).
+including streams in terminal states, but excluding stream 0 - MUST NOT exceed
+the value advertised by a receiver.  An endpoint MUST terminate a connection
+with a QUIC_FLOW_CONTROL_RECEIVED_TOO_MUCH_DATA error if it receives more data
+than the maximum data value that it has sent, unless this is a result of a
+change in the initial limits (see {{zerortt-parameters}}).
 
 
 ## MAX_STREAM_DATA Frame {#frame-max-stream-data}
@@ -2681,15 +2681,16 @@ When a packet is detected as lost, the sender re-sends any frames as necessary:
 * ACK and PADDING frames MUST NOT be retransmitted.  ACK frames
   containing updated information will be sent as described in {{frame-ack}}.
 
-* STOP_SENDING frames MUST be retransmitted, unless the stream has become closed
-  in the appropriate direction.  See {{solicited-state-transitions}}.
+* STOP_SENDING frames MUST be retransmitted until the receive stream enters
+  either a "Data Recvd" or "Reset Recvd" state.  See
+  {{solicited-state-transitions}}.
 
-* The most recent MAX_STREAM_DATA frame for a stream MUST be retransmitted. Any
-  previous unacknowledged MAX_STREAM_DATA frame for the same stream SHOULD NOT
-  be retransmitted since a newer MAX_STREAM_DATA frame for a stream obviates the
-  need for delivering older ones. Similarly, the most recent MAX_DATA and
-  MAX_STREAM_ID frames MUST be retransmitted; previous unacknowledged ones
-  SHOULD NOT be retransmitted.
+* The most recent MAX_STREAM_DATA frame for a stream MUST be retransmitted until
+  the receive stream enters a "Size Known" state. Any previous unacknowledged
+  MAX_STREAM_DATA frame for the same stream SHOULD NOT be retransmitted since a
+  newer MAX_STREAM_DATA frame for a stream obviates the need for delivering
+  older ones. Similarly, the most recent MAX_DATA frame MUST be retransmitted;
+  previous unacknowledged ones SHOULD NOT be retransmitted.
 
 * BLOCKED, STREAM_BLOCKED, and STREAM_ID_BLOCKED frames SHOULD be retransmitted
   if the sender is still blocked on the same limit.  If the limit has been
@@ -2838,10 +2839,10 @@ in both directions.
 Opening a stream causes all lower-numbered streams of the same type to
 implicitly open.  This includes both send and receive streams if the stream is
 bidirectional.  For bidirectional streams, an endpoint can send data on an
-implicitly opened a stream.  On both unidirectional and bidirectional streams,
-an endpoint MAY send MAX_STREAM_DATA or STOP_SENDING on implicitly opened
-streams.  An endpoint SHOULD NOT implicitly open streams that it initiates,
-instead opening streams in order.
+implicitly opened stream.  On both unidirectional and bidirectional streams, an
+endpoint MAY send MAX_STREAM_DATA or STOP_SENDING on implicitly opened streams.
+An endpoint SHOULD NOT implicitly open streams that it initiates, instead
+opening streams in order.
 
 Note:
 
@@ -3067,8 +3068,8 @@ the states of the pair of streams.  The simplest model presents the stream as
 states that loosely correspond to the stream states in HTTP/2
 {{?HTTP2=RFC7540}}.  This shows that multiple states on send or receive streams
 are mapped to the same composite state.  Note that this is just one possibility
-for a mapping, one that requires that data is acknowledged before the transition
-to a closed or half-closed state.
+for such a mapping; thi mapping requires that data is acknowledged before the
+transition to a "closed" or "half-closed" state.
 
 | Send Stream            | Receive Stream         | Composite State      |
 |:-----------------------|:-----------------------|:---------------------|
@@ -3088,7 +3089,7 @@ to a closed or half-closed state.
 Note (*1):
 
 : A stream is considered "idle" if it has not yet been created, or if the
-  receive stream is in the Recv state without yet having received any frames.
+  receive stream is in the "Recv" state without yet having received any frames.
 
 
 ## Solicited State Transitions
@@ -3104,19 +3105,21 @@ connection and stream flow-control windows, even though these frames will be
 discarded upon receipt.  This avoids potential ambiguity about which STREAM
 frames count toward flow control.
 
-STOP_SENDING can only be sent for any stream that is not "idle", however it is
-mostly useful for streams in the "open" or "half-closed (local)" states.  A
-STOP_SENDING frame requests that the receiving endpoint send a RST_STREAM frame.
-An endpoint that receives a STOP_SENDING frame MUST send a RST_STREAM frame for
-that stream with an error code of STOPPING.  If the STOP_SENDING frame is
-received on a stream that is already in the "half-closed (local)" or "closed"
-states, a RST_STREAM frame MAY still be sent in order to cancel retransmission
-of previously-sent STREAM frames.
+A STOP_SENDING frame requests that the receiving endpoint send a RST_STREAM
+frame.  An endpoint that receives a STOP_SENDING frame MUST send a RST_STREAM
+frame for that stream, and can use an error code of STOPPING.  If the
+STOP_SENDING frame is received on a send stream that is already in the "Data
+Sent" state, a RST_STREAM frame MAY still be sent in order to cancel
+retransmission of previously-sent STREAM frames.
 
-While STOP_SENDING frames are retransmittable, an implementation MAY choose not
-to retransmit a lost STOP_SENDING frame if the stream has already been closed
-in the appropriate direction since the frame was first generated.
-See {{packetization}}.
+STOP_SENDING can only be sent for a receive stream that has not been
+reset. STOP_SENDING is most useful for streams in the "Recv" or "Size Known"
+states.
+
+An endpoint is expected to send STOP_SENDING frames if packets containing the
+frame are lost.  However, once either all stream data has been received or a
+RST_STREAM frame - that is, any receive stream state other than "Recv" or "Size
+Known" - sending the frame is not necessary.
 
 
 ## Stream Concurrency {#stream-concurrency}
@@ -3363,10 +3366,8 @@ a RST_STREAM frame.  Otherwise, the final offset is the offset of the end of the
 data carried in a STREAM frame marked with a FIN flag, or 0 in the case of
 incoming unidirectional streams.
 
-An endpoint will know the final offset for a stream when the stream enters the
-"half-closed (remote)" state.  However, if there is reordering or loss, an
-endpoint might learn the final offset prior to entering this state if it is
-carried on a STREAM frame.
+An endpoint will know the final offset for a stream when the receive stream
+enters the "Size Known" or "Reset Recvd" state.
 
 An endpoint MUST NOT send data on a stream at or beyond the final offset.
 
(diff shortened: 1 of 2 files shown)


mikkelfj:
Just a quick note from a glance: It woud be useful to have a clear overview of which frames can be sent from which states. In its current form self transitions are not included. It could be in a separate transition table to reduce clutter.

martinthomson:
This is now essentially complete - at least in my view.  I've checked that I didn't miss anything from the original state machine and that the remainder of the document (and documents) don't refer to the old stream states.

draft-ietf-quic-transport.md
```diff
@@ -2966,19 +2968,21 @@ connection and stream flow-control windows, even though these frames will be
...
-in the appropriate direction since the frame was first generated.
-See {{packetization}}.
+STOP_SENDING can only be sent for a receive stream that has not been
+reset. STOP_SENDING is most useful for streams in the "Recv" or "Size Known"
+states.
+
+An endpoint is expected to send STOP_SENDING frames if packets containing the
+frame are lost.  However, once either all stream data has been received or a
+RST_STREAM frame - that is, any receive stream state other than "Recv" or "Size
+Known" - sending the frame is not necessary.
```
mikkelfj: Can this be formulated more clearly? I think the paragraph is discussing retransmission while avoiding using the loaded term. E.g. what does "the frame is lost" mean? Which frame is lost? Does it mean something like: A STOP_SENDING frame should be retransmitted in case of packet loss until acknowledged or a FIN or RST_STREAM signal has been received.

draft-ietf-quic-transport.md
```diff
@@ -2675,201 +2676,289 @@ created, it MUST close the connection with error code STREAM_STATE_ERROR.
 Stream ID 0 (0x0) is reserved for the cryptographic handshake.  Stream 0 MUST
 NOT be used for application data, and is the first client-initiated stream.
 
-A QUIC endpoint MUST NOT reuse a Stream ID.  Streams MUST be created
-in sequential order.  Open streams can be used in any order.  Streams
-that are used out of order result in lower-numbered streams in the
-same direction being counted as open.
+A QUIC endpoint MUST NOT reuse a Stream ID.  Open streams can be used in any
+order.  Streams that are used out of order result in lower-numbered streams in
+the same direction being counted as open.
```
ianswett: There is both direction and type(uni/bidi).  You use 'type' below, which seems fine here as well?

nit: "being opened." not being counted as opened.

draft-ietf-quic-transport.md
```diff
@@ -2675,201 +2676,289 @@ created, it MUST close the connection with error code STREAM_STATE_ERROR.
...
-the peer to send any data.  An endpoint will have either received all data that
-a peer has sent or will have received a RST_STREAM frame and discarded any
-received data.
+The sending part of a bidirectional stream initiated by a peer (type 0 for a
+server, type 1 for a client) enters the "Open" state if the receiving part
+enters the "Recv" state.
 
-Once all data has been either received or discarded, a sender is no longer
-obligated to update the maximum received data for the connection.
+Sending the first STREAM or STREAM_BLOCKED frame to be sent causes a send stream
```
ianswett: nit: remove "to be sent"

draft-ietf-quic-transport.md
```diff
@@ -2675,201 +2676,289 @@ created, it MUST close the connection with error code STREAM_STATE_ERROR.
...
+which can allow for better stream prioritization.
 
-Due to reordering, an endpoint could continue receiving frames for the stream
-even after the stream is closed for sending.  Frames received after a peer
-closes a stream SHOULD be discarded.  An endpoint MAY choose to limit the period
-over which it ignores frames and treat frames that arrive after this time as
-being in error.
+In the "Send" state, an endpoint transmits - and retransmits as necessary - data
+in STREAM frames.  The endpoint respects the flow control limits of its peer,
+accepting MAX_STREAM_DATA frames.  It generates STREAM_BLOCKED frames if it
```
ianswett: You talk about flow control a fair amount here.  Is it worth adding a 'blocked' state to the state machine?  I'm also happy with this as is and adding it later if we think it's useful.  Just a thought.

draft-ietf-quic-transport.md
```diff
@@ -2675,201 +2676,289 @@ created, it MUST close the connection with error code STREAM_STATE_ERROR.
...
+send stream enters the "Reset Recvd" state, which is a terminal state.
 
 
-### closed {#state-closed}
+### Receive Stream States {#stream-recv-states}
 
-The "closed" state is the terminal state for a stream.  Reordering might cause
-frames to be received after closing, see {{state-hc-remote}}.
+{{fig-stream-recv-states}} shows the states for the part of a stream that
+receives data from a peer.  The states for a receive stream mirror only some the
```
ianswett: nit: "only some of"

draft-ietf-quic-transport.md
```diff
@@ -2675,201 +2676,289 @@ created, it MUST close the connection with error code STREAM_STATE_ERROR.
...
+
+A bidirectional stream also opens when a MAX_STREAM_DATA frame is received.
+Receiving a MAX_STREAM_DATA frame implies that the remote peer has opened the
+stream and is providing flow control credit.  A MAX_STREAM_DATA frame might
+arrive before a STREAM or STREAM_BLOCKED frame if packets are lost or reordered.
+
+In the "Recv" state, the endpoint receives STREAM and STREAM_BLOCKED frames.
+Incoming data is buffered and reassembled into the correct order for delivery to
+the application.  As data is consumed by the application and buffer space
+becomes available, the endpoint sends MAX_STREAM_DATA frames to allow the peer
```
ianswett: MAY send?

draft-ietf-quic-transport.md
```diff
@@ -2675,201 +2676,289 @@ created, it MUST close the connection with error code STREAM_STATE_ERROR.
...
+or 0 and 2 for a server) are created when the first STREAM, STREAM_BLOCKED, or
+RST_STREAM is received for that stream.  The initial state for a receive stream
+is "Recv".  Receiving a RST_STREAM frame causes the receive stream to
+immediately transition to the "Reset Recvd".
+
+The receive stream enters the "Recv" state when the sending part of a
+bidirectional stream initiated by the endpoint (type 0 for a client, type 1 for
+a server) enters the "Open" state.
+
+A bidirectional stream also opens when a MAX_STREAM_DATA frame is received.
```
ianswett: I would move this up a paragraph so it's next to the other frames.

draft-ietf-quic-transport.md
```diff
@@ -2885,19 +2974,21 @@ connection and stream flow-control windows, even though these frames will be
...
+states.
 
-While STOP_SENDING frames are retransmittable, an implementation MAY choose not
-to retransmit a lost STOP_SENDING frame if the stream has already been closed
-in the appropriate direction since the frame was first generated.
-See {{packetization}}.
+An endpoint is expected to send STOP_SENDING frames if packets containing the
+frame are lost.  However, once either all stream data has been received or a
+RST_STREAM frame - that is, any receive stream state other than "Recv" or "Size
+Known" - sending the frame is not necessary.
```
erickinnear: Minor nit: This sentence doesn't parse for me. 
> However, once either all stream data has been received or a RST_STREAM frame sending the frame is not necessary.

Is the idea as below?  

> An endpoint is expected to only retransmit STOP_SENDING frames for receive stream states "Recv" and "Size Known". For example, if all stream data, or a RST_STREAM frame, has been received, retransmitting STOP_SENDING frames is not necessary.

draft-ietf-quic-transport.md
MikeBishop: You'd have to have it sit to the side of Send, though, with the ability to transition from Open to Blocked (no credit to start with) and the ability to oscillate between Blocked and Send, *and* the ability to close or reset from there as well.  It feels like it would complicate the concept of the state machine much more than it would clarify.

Just say that, while in the Send state, you MUST obey flow control, sending STREAM_BLOCKED if you're stuck.

draft-ietf-quic-transport.md
```diff
@@ -2675,201 +2676,289 @@ created, it MUST close the connection with error code STREAM_STATE_ERROR.
...
+the "Data Recvd" state, which is a terminal state.
 
-A stream in this state can be used by the endpoint to send any frame that
-mentions a stream ID.  In this state, the endpoint MUST observe advertised
-stream and connection data limits (see {{flow-control}}).
+From any of the "Open", "Send", or "Data Sent" states, an application can signal
+that it wishes to abandon transmission of stream data.  Similarly, the endpoint
+might receive a STOP_SENDING frame from its peer.  This causes the endpoint to
+send a RST_STREAM frame, which causes the stream to enter the "Reset Sent"
+state.
```
MikeBishop: Pedant alert:  The wording here doesn't make it blatantly obvious that "signal ... abandon" is done by RST_STREAM.  Maybe replace "This causes the endpoint to send..." with "In either case, the endpoint sends..."?

draft-ietf-quic-transport.md
```diff
@@ -2675,201 +2676,289 @@ created, it MUST close the connection with error code STREAM_STATE_ERROR.
...
 
-A stream transitions from this state to "closed" by completing transmission of
-all data.  This includes sending all data carried in STREAM frames including
-the terminal STREAM frame that contains a FIN flag.
+An endpoint MAY send a RST_STREAM to open a send stream; this causes the send
+stream to open and immediately transition to the "Reset Sent" state.
```
MikeBishop: Given that the stream was in the "Open" state before you send anything, describing the RST as opening the stream feels a little awkward.  Maybe say "...as the first frame on a send stream.  This causes the send stream to immediately transition...".

draft-ietf-quic-transport.md
```diff
@@ -2675,201 +2676,289 @@ created, it MUST close the connection with error code STREAM_STATE_ERROR.
...
 
-A stream also becomes "closed" when the endpoint sends a RST_STREAM frame.
+Once a packet containing a RST_STREAM has been successfully acknowledged, the
```
MikeBishop: Nit:  From the receiver's side of the ACK, there are no unsuccessful acknowledgements.

draft-ietf-quic-transport.md
```diff
@@ -2675,201 +2676,289 @@ created, it MUST close the connection with error code STREAM_STATE_ERROR.
...
+### Receive Stream States {#stream-recv-states}
 
-The "closed" state is the terminal state for a stream.  Reordering might cause
-frames to be received after closing, see {{state-hc-remote}}.
+{{fig-stream-recv-states}} shows the states for the part of a stream that
+receives data from a peer.  The states for a receive stream mirror only some the
+states of the send stream at the peer.  A receive stream doesn't track states on
+the send stream that can't be observed, such as the "Open" state; instead,
+receive streams track the delivery of data to the application or application
+protocol.
```
MikeBishop: ...some of which cannot be observed by the sender.

draft-ietf-quic-transport.md
```diff
@@ -2675,201 +2676,289 @@ created, it MUST close the connection with error code STREAM_STATE_ERROR.
...
 
-If the application resets a stream that is already in the "closed" state, a
-RST_STREAM frame MAY still be sent in order to cancel retransmissions of
-previously-sent STREAM frames.
+~~~
+       o
+       | Recv STREAM / STREAM_BLOCKED / RST_STREAM
```
MikeBishop: This makes it look like two RST_STREAMs are needed to enter "RST received."

draft-ietf-quic-transport.md
```diff
@@ -2675,201 +2676,289 @@ created, it MUST close the connection with error code STREAM_STATE_ERROR.
...
+       | Open Paired Stream (bidirectional)
+       | Recv MAX_STREAM_DATA
+       v
```
MikeBishop: On the send side, you have separate states for "I know the stream is open" and "data is flowing."  On the receive side, you go immediately to the "Recv" state.  Is there value in splitting those?

draft-ietf-quic-transport.md
```diff
@@ -2885,19 +2974,21 @@ connection and stream flow-control windows, even though these frames will be
...
+
+STOP_SENDING can only be sent for a receive stream that has not been
+reset. STOP_SENDING is most useful for streams in the "Recv" or "Size Known"
+states.
 
-While STOP_SENDING frames are retransmittable, an implementation MAY choose not
-to retransmit a lost STOP_SENDING frame if the stream has already been closed
-in the appropriate direction since the frame was first generated.
-See {{packetization}}.
+An endpoint is expected to send STOP_SENDING frames if packets containing the
```
MikeBishop: send => resend, send another, etc.

draft-ietf-quic-transport.md
```diff
@@ -2885,19 +2974,21 @@ connection and stream flow-control windows, even though these frames will be
...
+frame are lost.  However, once either all stream data has been received or a
```
MikeBishop: "all stream data or a RST_STREAM has been received"

draft-ietf-quic-transport.md
martinthomson: Yeah, blocked is a sub-state that doesn't really add much at this level of the description.

draft-ietf-quic-transport.md
martinthomson: Probably not, honestly - it's a MUST in a sense as well.  Best to leave the normative language about that to other parts of the spec.

draft-ietf-quic-transport.md
ianswett: I don't think so?

draft-ietf-quic-transport.md
```diff
@@ -2675,201 +2676,291 @@ created, it MUST close the connection with error code STREAM_STATE_ERROR.
 Stream ID 0 (0x0) is reserved for the cryptographic handshake.  Stream 0 MUST
 NOT be used for application data, and is the first client-initiated stream.
 
-A QUIC endpoint MUST NOT reuse a Stream ID.  Streams MUST be created
-in sequential order.  Open streams can be used in any order.  Streams
-that are used out of order result in lower-numbered streams in the
-same direction being counted as open.
+A QUIC endpoint MUST NOT reuse a Stream ID.  Open streams can be used in any
```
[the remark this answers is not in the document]
martinthomson: I forgot that I did that.  Since it's unenforceable, can't we just concede that open-in-order isn't strictly an interoperability requirement?  I could restore text with weasel words, but I've been convinced that strictly legislating here isn't constructive.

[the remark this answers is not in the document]
ianswett: Fair enough, make sure to find and close the issue this was discussed in then.

martinthomson:
@MikeBishop, what do you think is remaining here?  I'm happy to consider splitting frames into two categories, but that's something I'd prefer to do as a separate change.

PR #897, opened 2017-10-24
Update text on reaction to version negotiation

### LINKED ISSUE #826: clients always must regenerate client hello
The text in Sec 7.1 that details client handling of a version negotiation packet is not accurate:

"Though the contents of a packet might not change in response to version negotiation, a client MUST increase the packet number it uses on every packet it sends."

In particular because the transport parameters must be updated to reflect the version negotiation, the contents of the client initial packet must change. Also the first packet number will never be acknowledged so increasing the packet number by one to signify a retransmission does not make sense.

In particular the packet number should be reinitialized as specified in Sec 5.4.1 since as stated above, new contents must be created.

martinthomson:
Critically, to observe that the transport parameters need to change, even if nothing else does.

Closes #826.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -959,11 +959,12 @@ discarded.
 
 Once the Version Negotiation packet is determined to be valid, the client then
 selects an acceptable protocol version from the list provided by the server.
-The client then attempts to create a connection using that version.  Though the
-contents of the Client Initial packet the client sends might not change in
-response to version negotiation, a client MUST increase the packet number it
-uses on every packet it sends.  Packets MUST continue to use long headers and
-MUST include the new negotiated protocol version.
+The client then attempts to create a connection using that version.  The client
+generates a new Client Initial packet with a new packet number.  Even if the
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -959,11 +959,12 @@ discarded.
...
-contents of the Client Initial packet the client sends might not change in
-response to version negotiation, a client MUST increase the packet number it
-uses on every packet it sends.  Packets MUST continue to use long headers and
-MUST include the new negotiated protocol version.
+The client then attempts to create a connection using that version.  The client
+generates a new Client Initial packet with a new packet number.  Even if the
+cryptographic protocol used does not change as a result of a change in version,
+the cryptographic handshake message MUST be updated to include updated transport
+parameters that reflect the change of version.  This packet and any subsequent
+packets that use a long header MUST include the selected protocol version.
```
MikeBishop: Careful here -- the transport parameters shouldn't reflect the new version as its initially-proposed one, and an unwary reader could read it that way.

PR #899, opened 2017-10-25
Rework draining period into two (or three) states

### LINKED ISSUE #871: You don't always need the draining period
PR#868 doubled down on the draining and the proposal for PR#870 is about creating an exception to the draining period.

The basic problem here is spec confusion about the draining period, specifically, whether it's some sort "do nothing state" or whether it's a state where you're supposed to send packets. So, we have:

```
The draining period persists for three times the current Retransmission Timeout
(RTO) interval as defined in {{QUIC-RECOVERY}}.  During this period, new packets
can be acknowledged, but no new application data can be sent on the connection.

Different treatment is given to packets that are received while a connection is
in the draining period depending on how the connection was closed.

An endpoint that is in a draining period MUST NOT send packets unless they
contain a CONNECTION_CLOSE or APPLICATION_CLOSE frame.

Once the draining period has ended, an endpoint SHOULD discard per-connection
state.  This results in new packets on the connection being discarded.  An
endpoint MAY send a stateless reset in response to any further incoming packets.
```

However, in the specific case of stateless reset, the only thing that the connection needs to do is the thing in the fourth paragraph, which is *after* the draining period, namely discard packets (note that it can't send a stateless reset because that's forbidden).

The right fix for this is to have *two* states:

1. A state where you're expected to occasionally respond to new packets with something that says "go away" (e.g., CONNECTION_CLOSE) and where we're prescriptive about you doing that.

2. A state where you're expected to silently discard packets (perhaps with a stateless reset).

martinthomson:
As @ekr points out, there was definitely more than one closing state.  I've
modified this to the following three:

1. draining - when you know that the other side is closing, this is just to
   absorb stray packets

2. closing - when you have initiated the close and you need to re-send the
   close in addition to absorbing stray packets (this could be longer than the
   3RTO period if we like, but I've opted for keeping a shared timer)

3. timed out - when you timed out, and you are willing to revive the connection
   if new packets arrive; this one being optional

In the process I realized that we needed greater clarity about how to respond
to a closing frame (thanks to @huitema for the prompting).  I allow the
recipient to generate one closing frame, but no more than that.  They enter the
draining period thereafter and can't send any more packets.  Having two peers
in the closing state could result in a constant back-and-forth.

Editorial: I added the term "closing frame" to cover CONNECTION_CLOSE and
APPLICATION_CLOSE collectively, which were getting cumbersome.

Closes #871.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1618,69 +1618,84 @@ of three ways:
 * stateless reset ({{stateless-reset}})
 
 
-### Draining Period {#draining}
+### Draining and Closing Periods {#draining}
 
-After a connection is closed for any reason, an endpoint might receive packets
-from its peer.  These packets might have been sent prior to receiving any close
-signal, or they might be retransmissions of packets for which acknowledgments
-were lost.
+After a connection is closed, an endpoint might receive packets from its peer.
+These packets might have been sent prior to receiving any close signal, or they
+might be retransmissions of packets for which acknowledgments were lost.
 
 The draining period exists to ensure that delayed or reordered packets are
 properly discarded.  This state SHOULD persist for three times the current
-Retransmission Timeout (RTO) interval as defined in {{QUIC-RECOVERY}}.  During
-this period no new application data can be sent on the connection.
+Retransmission Timeout (RTO) interval as defined in {{QUIC-RECOVERY}}.  No
+packets can be sent during this time.
 
-Different treatment is given to packets that are received while a connection is
-in the draining period depending on how the connection was closed.  An endpoint
-that is in a draining period MUST NOT send packets unless they contain a
-CONNECTION_CLOSE or APPLICATION_CLOSE frame.
+Disposing of connection state prior to the end of the draining period could
+result in delayed or reordered packets to be handled poorly.  Endpoints that
+have some alternative means to ensure that late-arriving packets on the
+connection do not create QUIC state, such as by closing the UDP socket, MAY use
+an abbreviated draining period which can allow for faster resource recovery.
+Servers that retain an open socket for accepting new connections SHOULD NOT exit
+the draining period early.
 
 Once the draining period has ended, an endpoint SHOULD discard per-connection
-state.  This results in new packets on the connection being discarded.  An
-endpoint MAY send a stateless reset in response to any further incoming packets.
+state.  This results in new packets on the connection being handled generically.
+For instance, an endpoint MAY send a stateless reset in response to any further
+incoming packets.
 
-An endpoint MAY exit the draining period earlier if it can guarantee that its
-peer is also draining.  Receiving a CONNECTION_CLOSE or APPLICATION_CLOSE frame
-is sufficient confirmation, as is receiving a stateless reset.  However,
-disposing of connection state could result in delayed or reordered packets to be
-handled poorly.  For endpoints that have some alternative means to ensure that
-late-arriving packets on the connection do not create QUIC state, such as by
-closing the UDP socket, an abbreviated draining period can allow for faster
-resource recovery.  Servers that retain an open port for accepting new
-connections SHOULD NOT exit the draining period early.
+An endpoint enters a closing period after initiating an immediate close
+({{immediate-close}}) and optionally after an idle timeout ({{idle-timeout}}).
+This state runs for the same duration as the draining period.  While in the
+closing period, an endpoint MUST NOT send packets unless they contain a
+CONNECTION_CLOSE or APPLICATION_CLOSE frame (see {{immediate-close}} for
+details).
 
-The draining period does not apply when a stateless reset ({{stateless-reset}})
-is sent.
+The closing period differs from the draining period only in that closing frames
+can be retransmitted.
+
+An endpoint MAY transition from the closing period to the draining period if it
+can confirm that its peer is also closing or draining.  Receiving a closing
+frame is sufficient confirmation, as is receiving a stateless reset.  The
+draining period SHOULD end when the closing period would have ended.  In other
+words, the endpoint can use the same end time, but cease retransmission of the
+closing packet.
+
+The draining and closing periods do not apply when a stateless reset
+({{stateless-reset}}) is sent.
 
 
 ### Idle Timeout
 
 A connection that remains idle for longer than the idle timeout (see
-{{transport-parameter-definitions}}) becomes closed.  Either peer removes
-connection state if they have neither sent nor received a packet for this time.
+{{transport-parameter-definitions}}) is closed.  At implementation discretion, a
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1618,69 +1618,84 @@ of three ways:
...
-in the draining period depending on how the connection was closed.  An endpoint
-that is in a draining period MUST NOT send packets unless they contain a
-CONNECTION_CLOSE or APPLICATION_CLOSE frame.
+Disposing of connection state prior to the end of the draining period could
+result in delayed or reordered packets to be handled poorly.  Endpoints that
+have some alternative means to ensure that late-arriving packets on the
+connection do not create QUIC state, such as by closing the UDP socket, MAY use
+an abbreviated draining period which can allow for faster resource recovery.
+Servers that retain an open socket for accepting new connections SHOULD NOT exit
+the draining period early.
```
huitema: We may want to be a bit more specific about what connection state can be disposed of when entering the closing period. Implementations can probably just drop all information about unacknowledged packets, received packets or streams, but they need to be able to receive packets and decrypt them. If they don't do that, they won't be able to parse the incoming packet and notice arrival of a closing frame.

As for the draining period, state management becomes simpler with the new definition. The implementation just needs to remember the old connection ID, and guarantee that it won't send any packet for that connection ID until the end of the draining -- including not sending any stateless reset packet.

draft-ietf-quic-transport.md
```diff
@@ -1618,69 +1618,84 @@ of three ways:
...
+connection enters either the draining or closing period when the idle timeout
+expires.
 
 The time at which an idle timeout takes effect won't be perfectly synchronized
-on peers.  A connection enters the draining period when the idle timeout
-expires.  During this time, an endpoint that receives new packets MAY choose to
-restore the connection.  Alternatively, an endpoint that receives packets MAY
-signal the timeout using an immediate close.
+on peers.  An endpoint MAY enter a third state where receipt of new packets
+causes the connection to be restored to normal operation.
```
MikeBishop: "A third state" in a second subsection at the same level feels a little off.  If you want to talk about three closing states, I'd make closing/draining/idle three subsections in one place (or just merge this text into the same section above and don't rely on section headers to demarcate).

Also, these sections no longer line up with the three bullet points immediately before them.  Maybe we need to talk about "closing states" and "ways connections close."

draft-ietf-quic-transport.md
```diff
@@ -1618,69 +1618,84 @@ of three ways:
...
-
-During the draining period, an endpoint that sends a CONNECTION_CLOSE or
-APPLICATION_CLOSE frame SHOULD respond to any subsequent packet that it receives
-with another packet containing either close frame.  To reduce the state that an
-endpoint maintains in this case, it MAY send the exact same packet.  However,
-endpoints SHOULD limit the number of packets they generate containing either
-close frame.  For instance, an endpoint could progressively increase the number
-of packets that it receives before sending additional packets.
+An endpoint sends a closing frame, either CONNECTION_CLOSE or APPLICATION_CLOSE,
+to terminate the connection immediately.  Either closing frame causes all open
```
MikeBishop: Nit:  You don't need "open" here.  All streams are immediately closed; open streams get reset.

draft-ietf-quic-transport.md
```diff
@@ -1618,69 +1618,84 @@ of three ways:
...
+packets can be sent during this time.
 
-Different treatment is given to packets that are received while a connection is
-in the draining period depending on how the connection was closed.  An endpoint
-that is in a draining period MUST NOT send packets unless they contain a
-CONNECTION_CLOSE or APPLICATION_CLOSE frame.
+Disposing of connection state prior to the end of the draining period could
+result in delayed or reordered packets to be handled poorly.  Endpoints that
+have some alternative means to ensure that late-arriving packets on the
+connection do not create QUIC state, such as by closing the UDP socket, MAY use
```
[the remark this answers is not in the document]
martinthomson: You would have to assume that that is acceptable.  Probably not worth mentioning though (to avoid a serious digression).

draft-ietf-quic-transport.md
martinthomson: I've taken a stab at this.  connection ID is pretty much it.  I will use MAY for the packet protection keys.  I'll observe here (but not in the text) that a key update will be difficult to manage, as much as a path migration is difficult.  Let's not over-optimize for those cases.

draft-ietf-quic-transport.md
```diff
@@ -1618,69 +1618,84 @@ of three ways:
...
+an abbreviated draining period which can allow for faster resource recovery.
+Servers that retain an open socket for accepting new connections SHOULD NOT exit
+the draining period early.
 
 Once the draining period has ended, an endpoint SHOULD discard per-connection
-state.  This results in new packets on the connection being discarded.  An
-endpoint MAY send a stateless reset in response to any further incoming packets.
+state.  This results in new packets on the connection being handled generically.
+For instance, an endpoint MAY send a stateless reset in response to any further
+incoming packets.
```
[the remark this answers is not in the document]
martinthomson: This is a great suggestion.  I've taken the suggestion.

draft-ietf-quic-transport.md
martinthomson: I think that I might remove this third state thing and not mention revivification.

draft-ietf-quic-transport.md
```diff
@@ -1618,69 +1618,84 @@ of three ways:
...
+streams to immediately become closed; open streams can be assumed to be
+implicitly reset.
+
+After sending a closing frame, endpoints immediately enter a closing period.
+After receiving a closing frame, endpoints enter a draining period.  An endpoint
+that receives a closing frame MAY send a single closing frame before entering a
+draining period, but MUST NOT send further packets, which could result in a
+constant exchange of closing frames until the closing period on either peer
+ended.
+
```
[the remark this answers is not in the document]
martinthomson: I considered that, but a strict prohibition is easier.  It would be possible to reserve a codepoint for this, but we would also have to mandate decryption and I don't want to do that.

draft-ietf-quic-transport.md
```diff
@@ -1618,69 +1618,84 @@ of three ways:
...
+During the draining period, an endpoint that sends a closing frame SHOULD
+respond to any subsequent packet that it receives with another packet containing
+either closing frame.  To reduce the state that an endpoint maintains in this
+case, it MAY send the exact same packet.  However, endpoints SHOULD limit the
+number of packets they generate containing a closing frame.  For instance, an
+endpoint could progressively increase the number of packets that it receives
+before sending additional packets.
```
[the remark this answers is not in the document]
martinthomson: Yep, that's a straight error on my part.  I've rearranged this a little so that the distinction is clearer.

PR #902, opened 2017-10-27
fix KEY_PHASE bit on long header

marten-seemann:
The long header doesn't have any option for the KEY_PHASE.
I guess removing this was forgotten in #857 (but was wrong already, the corresponding packets types would have been 0x07 and 0x08, not 0x06 and 0x07).


martinthomson:
Thanks!

PR #908, opened 2017-11-08
Rework denial of service on unprotected packets

martinthomson:
This builds on #891 and applies the agreed principles to the question of denial of service prior to handshake completion.

Note that this does not address the concern @huitema raised about the second client flight.  Someone needs to come up with a proposal for that.  Right now, that flight is also subject to denial of service.


martinthomson:
Definitely another PR for the other one.  That would need an idea on how we might do that protection.

huitema:
I agree that the second flight DOS is another issue.

huitema:
Issue #1018 documents the second flight DOS attack.

PR #911, opened 2017-11-11
Make HTTP use bidirectional and unidirectional stream

### LINKED ISSUE #515: A server can't send SETTINGS first

MikeBishop:
The changes in #885 don't include HTTP, which leaves that draft somewhat
broken.  This corrects the omission.

There is an obvious optimization for HTTP, which is to not send the low two
bits of the Stream ID whenever it needs to identify a stream.  Right now, the
protocol only ever mentions client-initiated, bidirectional streams in frames,
so those two bits are wasted.  I'll save that change for later though.

I have chosen to take the control stream and use two unidirectional streams, which
addresses #515.

Closes #515, #240, #281.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -185,27 +185,32 @@ the HTTP framing layer. A QUIC receiver buffers and orders received STREAM
 frames, exposing the data contained within as a reliable byte stream to the
 application.
 
-QUIC reserves Stream 0 for crypto operations (the handshake, crypto config
-updates). Stream 1 is reserved for sending and receiving HTTP control frames,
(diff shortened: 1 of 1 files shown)


MikeBishop:
This is a duplicate of #886, created to resolve the fact that it got merged into the wrong branch.

PR #913, opened 2017-11-12
Clarify the length of the authentication tag

ekr:


marten-seemann:
Why is that? Doesn't this depend on the ciphersuite negotiated by TLS?

ekr:
@marten-seemann: here is what we specify

```
| TLS_AES_128_GCM_SHA256          | {0x13,0x01} |
| TLS_AES_256_GCM_SHA384          | {0x13,0x02} |
| TLS_CHACHA20_POLY1305_SHA256    | {0x13,0x03} |
| TLS_AES_128_CCM_SHA256          | {0x13,0x04} |
| TLS_AES_128_CCM_8_SHA256        | {0x13,0x05} |
```

We're not going to do CCM and it just happens that the others use 16-byte tags.

martinduke:
If in practice all cipher suites use the 16-byte tag, it saves the reader sifting through an endless cascade of documents to dig out the answer.

PR #915, opened 2017-11-12
clarify connection id for handshake secret

mcmanus:
Clarify that the Handshake Secret Connection ID comes from the client Initial packet as opposed to any client packet.


draft-ietf-quic-tls.md
```diff
@@ -644,7 +644,8 @@ SHA-256 hash function is used.
 ### Handshake Secrets {#handshake-secrets}
 
 Packets that carry the TLS handshake (Initial, Retry, and Handshake) are
-protected with secrets derived from the client's connection ID. Specifically:
+protected with secrets derived from the connection ID used in the
+client's Inital packet. Specifically:
```
marten-seemann: s/Inital/Initial/

mcmanus:
thanks!

Issue #917, opened 2017-11-12
Allow removal of ACK blocks that are in flight


martinthomson:
The discussion in #907 seemed to conclude that it was OK to not send all acknowledgments all the time.  That PR only mentions removal of ACK blocks based on them having been acknowledged, but did not include the other option, which is to not include ACK blocks that are in flight.

There's value in sending acknowledgments more often than once per round trip (timeliness of ACK delivery is tantamount), but there also could be some value in reducing the size of ACK frames if the same ACK block has been sent multiple times already in outstanding packets.

mikkelfj:
I think the multiple ack frames issue was closed, but there is still no clarification on how to handle a situation where ACK's will not fit into a single frame, or even MTU. Even if unlikely, implementations need guidelines. How, and when, to not ack certain packets applies here.

ianswett:
I think the existing text is mostly sufficient(though I realize it's a bit redundant at the moment, which I'll fix).  I really do not advise some clever scheme where you send an ack block once and then then stop sending it for some arbitrary period of time and then start again, so I don't want to write text implying I'd recommend it.

There is still work to be done to add more discussion of when to send acknowledgements and what to include in them, but it was decided in Seattle that the transport draft should focus on the boundaries of what the receiver should do and the recovery draft can discuss the nuances and optimizations that are possible with considerations about their pros and cons.  That work is tracked in #230

"To limit receiver state or the size of ACK frames, a receiver MAY limit the number of ACK blocks it sends. A receiver can do this even without receiving acknowledgment of its ACK frames, with the knowledge this could cause the sender to unnecessarily retransmit some data. When this is necessary, the receiver SHOULD acknowledge newly received packets and stop acknowledging packets received in the past. Removing ack blocks before the peer has received them increases the chances the peer never received an acknowledgement, will declare the packet lost, and spuriously retransmit information."

MikeBishop:
Can you help me understand the pathological case when something is ACK’d, omitted, then ACK’d again?  If the original ACK gets through, the sender should ignore the second one.  If the original ACK gets lost, it looks to the sender like a severely delayed packet and it will realize the retransmission was spurious (if it still has data), but might update the congestion controller with the un-loss event.  That doesn’t seem so bad at first glance.

mikkelfj:
> Can you help me understand the pathological case when something is ACK’d, omitted, then ACK’d again? 

[...]

PR #918, opened 2017-11-12
Remove a redundant sentence

ianswett:
Also add text on why stopping and then resuming sending older ack blocks is not recommended in response to #917


martinthomson:
This closes #917.

janaiyengar:
LGTM, merging.

janaiyengar:
I'll note that the comment says this adds text on why stopping and then resuming sending older ack blocks is not recommended in response to #917, but it doesn't. Did you mean to add that text?

ianswett:
@janaiyengar After further discussion with Mike in person, I decided it'd be better to add text to the recovery doc and make this as straightforward and clear as possible.

PR #920, opened 2017-11-13
initial server packet number restricted to 31 bits

mcmanus:
this matches the reference in client initial packet number


martinthomson:
Hmm, the original text isn't that great. but I think that we'll just take on the 0/1 packet number issue and fix this properly.

mcmanus:
sgtm

Issue #587, opened 2017-06-06
Recovery: Handshake Packet is not using constant names, and uses wrong default


gloinul:
Recovery -03: Section 3.2.7.1:

If no previous RTT is available, the
   initial RTT defaults to 200ms.

This should reference the default konstant defined, and the value is currently set to 100 ms.

ianswett:
This is being removed in PR #921

gloinul:
So -10 resolves the mismatch about default. And I am fine with not using pseudo code constants in description.

PR #921, opened 2017-11-13
Editorial fixes and v1 of RTT calculation

### LINKED ISSUE #471: Retransmission of handshake frames
The draft currently says:

> Endpoints MUST retransmit handshake frames if not acknowledged within a time limit. 

I've been trying to be very careful about what it means to retransmit.  We don't retransmit packets, we create new packets; similarly, we don't retransmit frames, we create new frames that contain the data we want.

In the case of the first client packet, we can probably be a little more prescriptive.  The client can send another packet with the same contents, just with the packet number incremented.

ianswett:
Editorial fixes from my re-review of the recovery doc on the plane and add some text for the RTT calculation section, because it's important.

Fixes #471  and #587


draft-ietf-quic-recovery.md
```diff
@@ -194,8 +194,14 @@ round-trip time (RTT) is critical to these algorithms and is described first.
 
 ## Computing the RTT estimate
 
-(To be filled)
+RTT is calculated when an ACK frame arrives by computing the difference
+between the current time and the time the largest newly acked packet was sent.
+If no packets are newly acknowledged, RTT cannot be calculated. When RTT is
+calculated, the ack delay field from the ACK frame SHOULD be subtracted from
+the RTT as long as the result is positive.
```
marten-seemann: Is this the right thing to do?
Sending an ack delay that would make the RTT negative sounds like a protocol violation. If we want to be generous with a peer that might just have adjusted its system time, the right thing to do would be ignoring that RTT sample completely, wouldn't it?

ianswett: It's a good question.  GQUIC does what I wrote, but I'm not sure it's the best thing.  

The only reason I wouldn't want to throw it out entirely is that it may cause you to get no RTT samples in the worst case.

marten-seemann: Your text is consistent with the pseudocode. Do we need a new issue to track this?

ianswett: Possibly.  Alternately, I can add a sentence indicating that samples may be ignored if the ack delay is too large?  I'm not convinced we will or need to come to consensus on this, so it might be better to allow either until we have a strong reason to suggest one over the other.

draft-ietf-quic-recovery.md
```diff
@@ -166,14 +165,15 @@ not available.
...
-QUIC ACKs contain information that is equivalent to TCP SACK, but QUIC does not
+QUIC ACKs contain information that is similar to TCP SACK, but QUIC does not
 allow any acked packet to be reneged, greatly simplifying implementations on
 both sides and reducing memory pressure on the sender.
 
 ### More ACK Ranges
 
 QUIC supports up to 256 ACK ranges, opposed to TCP's 3 SACK ranges.  In high
-loss environments, this speeds recovery.
+loss environments, this speeds recovery, reduces spuriuos retransmits,
```
janaiyengar: typo: "spurious"

draft-ietf-quic-recovery.md
janaiyengar: Let's add this text in and open an issue. GQUIC currently subtracts this from RTT measurements, but if the peer's clock is off, this ack delay could be arbitrarily large, leading to a poor RTT estimate. Perhaps this should be protected by the measured minRTT as the min... but let's discuss that on an issue separately.

draft-ietf-quic-recovery.md
```diff
@@ -292,13 +298,12 @@ conditions:
 * If there are more than one unacknowledged packets, PTO SHOULD be scheduled for
   max(2*SRTT, 10ms).
 
-* If RTO is earlier, schedule a TLP alarm in its place. That is, PTO SHOULD be
-  scheduled for min(RTO, PTO).
+* If RTO (RTO, {{rto}}) is earlier, schedule a TLP alarm in its place. That is,
```
janaiyengar: Why the "RTO" inside the parenthesis? Did you mean to say "If RTO {{rto}} is earlier"?

draft-ietf-quic-recovery.md
```diff
@@ -308,10 +313,8 @@ A PTO value of at least 2*SRTT ensures that the ACK is overdue. Using a PTO of
 exactly 1*SRTT may generate spurious probes, and 2*SRTT is simply the next
 integral value of RTT.
 
-(TODO: These values of 2 and 1.5 are a bit arbitrary. Reconsider these.)
-
-If the Retransmission Timeout (RTO, {{rto}}) period is smaller than the computed
-PTO, then a PTO is scheduled for the smaller RTO period.
+The values of 2 and 1.5 are based on the TLP draft, but implementations MAY
```
janaiyengar: cite TLP draft

PR #922, opened 2017-11-13
Define kUsingTimeLossDetection

### LINKED ISSUE #534: UsingTimeLossDetection not defined
`UsingTimeLossDetection` is referred to in Section 3.2.3, but is not defined anywhere.

ianswett:
Fixes #534


marten-seemann:
I think you can use this constant in `DetectLostPackets` to replace the line `if (time_reordering_fraction != infinite)` (which isn't set to `infinite` anywhere, by the way).

ianswett:
Thanks Marten, updated.

PR #923, opened 2017-11-13
Use time_of_last_sent_packet

### LINKED ISSUE #540: time_of_last_sent_packet should be defined as a "variable of interest"
It sees like `time_of_last_sent_packet` should be defined somewhere as a connection-level variable?

ianswett:
time_of_last_sent_packet should be used to set the loss detection timer, instead of using now.

Fixes #540

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -687,7 +687,7 @@ Pseudocode for SetLossDetectionAlarm follows:
       alarm_duration = alarm_duration * (2 ^ handshake_count)
     else if (loss_time != 0):
       // Early retransmit timer or time loss detection.
-      alarm_duration = loss_time - now
+      alarm_duration = loss_time - time_of_last_sent_packet
     else if (tlp_count < kMaxTLPs):
       // Tail Loss Probe
       if (num_retransmittable_packets_outstanding == 1):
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -687,7 +687,7 @@ Pseudocode for SetLossDetectionAlarm follows:
       alarm_duration = alarm_duration * (2 ^ handshake_count)
     else if (loss_time != 0):
```
siyengar: this is also weird, it's possible that loss_time is greater than tlp timer or RTO, shouldn't we take the min(loss_time, rto_timer, tlp_timer) as the early retransmit timer, or at least min(loss_time, rto)

draft-ietf-quic-recovery.md
```diff
@@ -701,7 +701,8 @@ Pseudocode for SetLossDetectionAlarm follows:
       alarm_duration = max(alarm_duration, kMinRTOTimeout)
       alarm_duration = alarm_duration * (2 ^ rto_count)
 
-    loss_detection_alarm.set(now + alarm_duration)
+    loss_detection_alarm.set(time_of_last_sent_packet
```
siyengar: for handshake packets should this be the time since the last handshake packet was transmitted?

siyengar:
github won't let me comment on the particular line of code, however 

in detect loss packets the delay until lost is also wrong, it should be 

1/8 * max(srtt, lrtt) vs 9/8 -> or maybe it should be 1 /4 because we should at least wait for a 1/4 rtt
and it should be timereordering vs 1 + timereordering.

draft-ietf-quic-recovery.md
ianswett: That's a good point, given the other packets are not decryptable, and hence will not generate an acknowledgement until the handshake completes, #962 filed.

draft-ietf-quic-recovery.md
janaiyengar: I dont think this is an issue. This is at the beginning of the connection, and loss_time only applies to early retransmit. It's possible that loss_time is > TLP or RTO period, but that would be very very strange.

Issue #65, opened 2016-12-01
BLOCKED usage requirements


martinthomson:
Does this need to be MUST use?

igorlord:
Agreed. BLOCKED is described as a useful debugging tool. It should not be a MUST.

In fact, I'd go further and change BLOCKED frame type into DEBUG frame type and allow DEBUG frames to carry key/value pairs. Some keys could be "well-known" (BLOCKED). Others could be user-defined. This could quickly evolve into an even more useful debugging tool, possibly with the clients having an ability to turn up and down transport's "DEBUG levels".

martinthomson:
@igorlord, we have plenty of space for new frame types.  Adding another point of extension in the protocol, especially one that could be rarely used, is likely to cause problems.

janaiyengar:
We could change it to "MAY" but a debugging frame is useless if it isn't sent. What's the harm in requiring that it be sent?

igorlord:
I think the answer to MUST/SHOULD/MAY for the BLOCKED frame must be identical to the answer to whether this frame MUST/SHOULD/MAY be retransmitted in case of a loss.  See #452.

MikeBishop:
Closed by #924.

PR #924, opened 2017-11-14
BLOCKED, SHOULD, and retransmission

### LINKED ISSUE #452: BLOCKED semantics
It's possible that we don't care about the reliability of BLOCKED, but there is an inherent problem with the frame.

The quality of a signal like this is either in being able to capture the conditions that lead to it being generated, maybe in terms of looking at traces.  Or, it can be useful hollistically: by aggregating reports over a large set you can determine whether a change to flow control tuning was successful.

A sender of BLOCKED is currently obligated to retransmit lost BLOCKED frames.  That achieves the latter goal reasonably well, but it does little for the former because the state of the connection when BLOCKED was generated is unlikely to be the same when it is eventually retransmitted.  Even on the first transmission, the receiver of a BLOCKED frame still has to do quite a lot to work out what the state of the sender might have been.

BLOCKED would be more useful if it reported the limit that caused it to be generated.  That would also solve #451.

MikeBishop:
Fixes #452, #65.

Sending BLOCKED is a SHOULD (not a MUST), since it's a debug mechanism and not vital to protocol functionality.  Clarified retransmission rules to say that you SHOULD retransmit if the limit hasn't changed since the frame was sent, and not otherwise.  (If it changed and you're blocked again, you presumably sent a new frame already.)

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2118,9 +2118,9 @@ prevent the majority of middleboxes from losing state for UDP flows.
 
 ## BLOCKED Frame {#frame-blocked}
 
-A sender sends a BLOCKED frame (type=0x08) when it wishes to send data, but is
-unable to due to connection-level flow control (see {{blocking}}). BLOCKED
-frames can be used as input to tuning of flow control algorithms (see
+A sender SHOULD send a BLOCKED frame (type=0x08) when it wishes to send data,
+but is unable to due to connection-level flow control (see {{blocking}}).
+BLOCKED frames can be used as input to tuning of flow control algorithms (see
 {{fc-credit}}).
 
 The BLOCKED frame does not contain a payload.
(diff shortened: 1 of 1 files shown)


marten-seemann:
sgtm, but I'm still wondering if we should include the byte offset / max stream id in those debugging frames. It seems like we're making interpretation of these frame unnecessarily hard by having the receiver guess these values.

ianswett:
Interesting suggestion Marten.  Given we acknowledge MAX_STREAM_DATA frames and such, I think we get most of the information we need, but I could be wrong.

marten-seemann:
I'm not saying it's impossible, but you'd need a data structure keeping tracking of sent MAX_[DATA, STREAM_DATA, STREAM_ID] frames and which of them was acked. There would still be some uncertainty, since it's not guaranteed that the ACK for those frames is received before or at the same time as the debugging frame.

MikeBishop:
I don't think it would be unreasonable to package the offset or stream ID in question that triggered sending the frame.  That could also help the other side understand whether you're blocked because they misunderstimicated the window you would need, or if the packet containing their update was simply delayed and you should be fine now.

marten-seemann:
From the feedback here it sounds like this is worth writing a separate PR. I'll submit one soon. Don't want to block the merge of this one.

MikeBishop:
Split it to a second issue -- feel free to submit a PR against it.

PR #928, opened 2017-11-14
include debug infos in BLOCKED, STREAM_BLOCKED and STREAM_ID_BLOCKED

### LINKED ISSUE #927: Include blocking state in BLOCKED / STREAM_BLOCKED / STREAM_ID_BLOCKED
This has been suggested a couple times, most recently by @marten-seemann in #924:  If the flow control debug frames included the value of the block the sender is hitting, it simplifies a couple things:

  - The receiver can know whether they've unblocked the sender already
  - Differences between the two sides' view of the connection window can be identified
  - Retransmission logic might be easier, since it becomes easier to identify whether the state has changed

Splitting this off to unblock the other PR.

marten-seemann:
Fixes #927.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2123,7 +2123,22 @@ unable to due to connection-level flow control (see {{blocking}}). BLOCKED
 frames can be used as input to tuning of flow control algorithms (see
 {{fc-credit}}).
 
-The BLOCKED frame does not contain a payload.
+The BLOCKED frame is as follows:
+
+~~~
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                         Offset (i)                         ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+~~~
+
+The BLOCKED frame contains a single field.
+
+Offset:
+
+: A variable-length integer indicating the connection-level offset at which
+  the blocking occurred.
 
 
 ## STREAM_BLOCKED Frame {#frame-stream-blocked}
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2123,7 +2123,21 @@ unable to due to connection-level flow control (see {{blocking}}). BLOCKED
...
-The BLOCKED frame does not contain a payload.
+The BLOCKED frame is as follows:
+
+~~~
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                         Offset (i)                         ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+
```
MikeBishop: You need another `~~~` here to end the figure.

marten-seemann: Fixed.

draft-ietf-quic-transport.md
```diff
@@ -2140,14 +2155,21 @@ The STREAM_BLOCKED frame is as follows:
...
+The STREAM_BLOCKED frame contains two fields:
 
 Stream ID:
 
 : A variable-length integer indicating the stream which is flow control blocked.
 
+Offset:
+
+: A variable-length integer indicating the current offset of data sent on the
+stream.
```
ianswett: nit: I think you need to align 'stream' with 'A' for correct formatting?

draft-ietf-quic-transport.md
```diff
@@ -2157,8 +2179,22 @@ stream, but is unable to due to the maximum stream ID limit set by its peer (see
...
+|                        Stream ID (i)                        ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+~~~
+
+The STREAM_ID_BLOCKED frame contains a single field.
+
+Stream ID:
 
+: A variable-length integer indicating the stream ID of highest stream that
+could be openend.
```
ianswett: same nit on getting the right formatting

draft-ietf-quic-transport.md
```diff
@@ -2123,7 +2123,22 @@ unable to due to connection-level flow control (see {{blocking}}). BLOCKED
...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                         Offset (i)                         ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+~~~
+
+The BLOCKED frame contains a single field.
+
+Offset:
+
+: A variable-length integer indicating the current offset of data sent on the
```
martinthomson: current is ambiguous.  I presume that the intent is to signal the offset at which the blocking occurred.

draft-ietf-quic-transport.md
martinthomson: Same point here about "current"

draft-ietf-quic-transport.md
martinthomson: Typo on opened.

I think that we need to be really crisp about this.  Is this the value of MAX_STREAM_ID?  The alternative would be to signal the stream ID of the stream you wanted to create (which would be bad).

marten-seemann: No, that wasn't my intention. It should be the same value as sent in the MAX_STREAM_ID frame.

marten-seemann:
I pushed the fixes. Please have another look.

PR #939, opened 2017-11-16
Move version before packet number

### LINKED ISSUE #926: Future flexibility limited by QUIC version coming before packet number
If we ever want to change anything about the packet number in long headers, the version should be first, otherwise we're going to have to do something very awkward.

ianswett:
Closes #926

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -515,8 +515,8 @@ A Version Negotiation packet has long headers with a type value of 0x01 and is
 sent only by servers.  The Version Negotiation packet is a response to a client
 packet that contains a version that is not supported by the server.
 
-The packet number, connection ID and version fields echo corresponding values
-from the triggering client packet.  This allows clients some assurance that the
+The connection ID and version fields echo corresponding values from the
+triggering client packet.  This allows clients some assurance that the
 server received the packet and that the Version Negotiation packet was not
 carried in a packet with a spoofed source address.
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -940,8 +940,8 @@ successfully receives a response or it abandons the connection attempt.
 ### Handling Version Negotiation Packets {#handle-vn}
 
 When the client receives a Version Negotiation packet, it first checks that the
-packet number is 0 and the connection ID matches the connection ID the client
-sent.  If this check fails, the packet MUST be discarded.
+connection ID matches the connection ID the client sent.  If this check fails,
+the packet MUST be discarded.
 
 Once the Version Negotiation packet is determined to be valid, the client then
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -395,17 +395,17 @@ Connection ID:
...
   use of this field in more detail.
 
-Packet Number:
-
-: Octets 9 to 12 contain the packet number.  {{packet-numbers}} describes the
-  use of packet numbers.
-
 Version:
 
 : Octets 13 to 16 contain the selected protocol version.  This field indicates
```
MikeBishop: You'll need to renumber the octets, too.

ianswett: Thanks Mike, fixed

draft-ietf-quic-transport.md
```diff
@@ -515,8 +515,8 @@ A Version Negotiation packet has long headers with a type value of 0x01 and is
 sent only by servers.  The Version Negotiation packet is a response to a client
 packet that contains a version that is not supported by the server.
 
-The packet number, connection ID and version fields echo corresponding values
-from the triggering client packet.  This allows clients some assurance that the
+The connection ID and version fields echo corresponding values from the
+triggering client packet.  This allows clients some assurance that the
 server received the packet and that the Version Negotiation packet was not
 carried in a packet with a spoofed source address.
```
janaiyengar: I would add something like "The packet number field SHOULD be set to 0 and will be ignored by the receiver."

draft-ietf-quic-transport.md
```diff
@@ -940,9 +940,8 @@ successfully receives a response or it abandons the connection attempt.
 ### Handling Version Negotiation Packets {#handle-vn}
 
 When the client receives a Version Negotiation packet, it first checks that the
-packet number and connection ID match the values the client sent in a previous
-packet on the same connection.  If this check fails, the packet MUST be
-discarded.
+packet number is 0 and the connection ID matches the connection ID the client
```
janaiyengar: I don't think you need to check the packet number being 0. It's enough to check the connection ID and I would ignore the packet number at the receiver.

draft-ietf-quic-transport.md
ianswett: I think it's fine not to check the packet number(as updated), but if so, then advising what to set there seems unnecessary?

janaiyengar: The sender has to put something in the packet number, so I think it's better to be explicit about what it should do.

martinthomson: Note that if we decide to have a shorter packet header for Version Negotiation, a 0 is harmless. It advertises support for a version that cannot be supported, so it can be ignored.

I was going to suggest having explicit text about ignoring the packet number, but we might just remove it, so don't bother just yet.

PR #943, opened 2017-11-17
Explain that NST.max_stream_data is needed for 0-RTT.

ekr:
Note that we require 0xffffffff in order to keep things consistent.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1424,9 +1424,10 @@ compatible.  The client indicates which ALPN label it has chosen by placing that
 ALPN label first in the ALPN extension. In order to be usable for 0-RTT,
 the NewSessionTicket MUST contain the "max_early_data" extension with the
 value 0xffffffff; the amount of data which the client can send in 0-RTT
-is controlled by the "initial_max_data" and "initial_max_stream_data"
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -1421,7 +1421,12 @@ enable 0-RTT in TCP.
 All the restrictions on the use of 0-RTT apply, with the exception of the ALPN
 label, which MUST only change to a label that is explicitly designated as being
 compatible.  The client indicates which ALPN label it has chosen by placing that
-ALPN label first in the ALPN extension.
+ALPN label first in the ALPN extension. In order to be usable for 0-RTT,
+the NewSessionTicket MUST contain the "max_early_data" extension with the
+value 0xffffffff; the amount of data which the client can send in 0-RTT
+is controlled by the "initial_max_data" and "initial_max_stream_data"
+parameters supplied by the server. It is an error for the server to provide
+any other value for "max_early_data".
```
martinthomson: Which error?

ekr: PROTOCOL_VIOLATION perhaps?

martinthomson: Yeah, no point in defining anything more specific here.  If we did that for every parameter, we'd have a million useless error codes.

Issue #426, opened 2017-03-30
QUIC header format/demultiplexing


larseggert:
[I am raising this for @csperkins; text below is from him.]

If I understand correctly, the first octet of the new QUIC header is either:
```
 +-+-+-+-+-+-+-+-+
 |1|   Type (7)  |  Long header packet
 +-+-+-+-+-+-+-+-+
```
or
```
 +-+-+-+-+-+-+-+-+
 |0|C|K| Type (5)|  Short header packet
 +-+-+-+-+-+-+-+-+
```
Given the assigned types, this gives number values for the first octet in the ranges 0-3, 32-35, 64-67, 96-99, or 128-136.

RFC 7983 talks about how to demultiplex protocols running on a single UDP port, based on the values of the first octet. The important part is the new text table in Section 7, which demultiplexes based on the value of the first octet of the packet as follows:
```
                   +----------------+
                   |        [0..3] -+--> forward to STUN
                   |                |
                   |      [16..19] -+--> forward to ZRTP
                   |                |
       packet -->  |      [20..63] -+--> forward to DTLS
                   |                |
                   |      [64..79] -+--> forward to TURN Channel
                   |                |
                   |    [128..191] -+--> forward to RTP/RTCP
                   +----------------+
```
The QUIC headers don’t fit cleanly into this. The result is that it’s not possible to run demultiplex QUIC and WebRTC running on the same UDP port, since WebRTC uses STUN, DTLS, TURN and RTP/RTCP. Demultiplexing these is probably desirable, since it would let you switch from web traffic to interactive WebRTC or very low latency streaming without de-doing the NAT bindings on a port.

The simplest fix to allow the protocols to be demultiplexed is to set the top two bits of the first octet of a QUIC packet to 1 and reduce the size of the type fields:
```
 +-+-+-+-+-+-+-+-+
 |1|1|1|Type (5) |  Long header packet
 +-+-+-+-+-+-+-+-+

 +-+-+-+-+-+-+-+-+
 |1|1|0|C|K|Type3|  Short header packet
 +-+-+-+-+-+-+-+-+
```
There are probably other options that keep a larger type field, if we accept a less obvious bit packing, although even these likely lose one bit from the type field.

martinthomson:
I don't think that we need to multiplex QUIC with realtime media.  (If we multiplexed every conceivable protocol, we'd run out of bits.)

csperkins:
[...]

PR #956, opened 2017-11-22
Renumber packet types and invert Connection ID Flag

csperkins:
This commit makes the following changes:

- Renumber long header packet types to count from 0x7f downwards, rather than from 0x01 upwards
- Renumber short header packet types to count from 0x1f downward, rather than from 0x01 upward
- Invert the sense of the Connection ID Flag in the short header packets, to become an Omit Connection ID Flag

This is a purely syntactic change, with no semantic impact on QUIC.

The purpose of the changes is to make it possible to easily demultiplex QUIC and STUN packets running on a single UDP port, to enable peer to peer use of QUIC. It also enables demultiplexing of WebRTC packets and QUIC packets on a single UDP port.

Discussed in QUIC and AVTCORE working groups at IETF 100 (draft-aboba-avtcore-quic-multiplexing-01) and on QUIC mailing list.

Addresses issue #426

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -416,11 +416,11 @@ The following packet types are defined:
 
 | Type | Name                          | Section                     |
 |:-----|:------------------------------|:----------------------------|
-| 0x01 | Version Negotiation           | {{packet-version}}          |
-| 0x02 | Initial                       | {{packet-initial}}          |
-| 0x03 | Retry                         | {{packet-retry}}            |
-| 0x04 | Handshake                     | {{packet-handshake}}        |
-| 0x05 | 0-RTT Protected               | {{packet-protected}}        |
+| 0x7F | Version Negotiation           | {{packet-version}}          |
+| 0x7E | Initial                       | {{packet-initial}}          |
+| 0x7D | Retry                         | {{packet-retry}}            |
+| 0x7C | Handshake                     | {{packet-handshake}}        |
+| 0x7B | 0-RTT Protected               | {{packet-protected}}        |
 {: #long-packet-types title="Long Header Packet Types"}
 
 The header form, packet type, connection ID, packet number and version fields of
@@ -459,10 +459,10 @@ Header Form:
 
 : The most significant bit (0x80) of octet 0 is set to 0 for the short header.
 
-Connection ID Flag:
+Omit Connection ID Flag:
 
 : The second bit (0x40) of octet 0 indicates whether the Connection ID field is
-  present.  If set to 1, then the Connection ID field is present; if set to 0,
+  omitted.  If set to 0, then the Connection ID field is present; if set to 1,
   the Connection ID field is omitted.  The Connection ID field can
   only be omitted if the omit_connection_id transport parameter
   ({{transport-parameter-definitions}}) is specified by the intended recipient
@@ -481,8 +481,8 @@ Short Packet Type:
 
 Connection ID:
 
-: If the Connection ID Flag is set, a connection ID occupies octets 1 through 8
-  of the packet.  See {{connection-id}} for more details.
+: If the Omit Connection ID Flag is not set, a connection ID occupies octets
+  1 through 8 of the packet.  See {{connection-id}} for more details.
 
 Packet Number:
 
@@ -499,20 +499,20 @@ other fields.
 
 | Type | Packet Number Size |
 |:-----|:-------------------|
-| 0x01 | 1 octet            |
-| 0x02 | 2 octets           |
-| 0x03 | 4 octets           |
+| 0x1F | 1 octet            |
+| 0x1E | 2 octets           |
+| 0x1D | 4 octets           |
 {: #short-packet-types title="Short Header Packet Types"}
 
-The header form, connection ID flag and connection ID of a short header packet
-are version-independent.  The remaining fields are specific to the selected QUIC
-version.  See {{version-specific}} for details on how packets from different
-versions of QUIC are interpreted.
+The header form, omit connection ID flag, and connection ID of a short header
+packet are version-independent.  The remaining fields are specific to the
+selected QUIC version.  See {{version-specific}} for details on how packets
+from different versions of QUIC are interpreted.
 
 
 ## Version Negotiation Packet {#packet-version}
 
-A Version Negotiation packet has long headers with a type value of 0x01 and is
+A Version Negotiation packet has long headers with a type value of 0x7F and is
 sent only by servers.  The Version Negotiation packet is a response to a client
 packet that contains a version that is not supported by the server.
 
@@ -566,7 +566,7 @@ off-path attackers.
 
 ### Initial Packet {#packet-initial}
 
-The Initial packet uses long headers with a type value of 0x02.  It carries the
+The Initial packet uses long headers with a type value of 0x7E.  It carries the
 first cryptographic handshake message sent by the client.
 
 The client populates the connection ID field with randomly selected values,
@@ -594,7 +594,7 @@ Retry packet ({{packet-retry}}).
 
 ### Retry Packet {#packet-retry}
 
-A Retry packet uses long headers with a type value of 0x03.  It carries
+A Retry packet uses long headers with a type value of 0x7D.  It carries
 cryptographic handshake messages and acknowledgments.  It is used by a server
 that wishes to perform a stateless retry (see {{stateless-retry}}).
 
@@ -627,7 +627,7 @@ the server will also start at stream offset 0.
 
 ### Handshake Packet {#packet-handshake}
 
-A Handshake packet uses long headers with a type value of 0x04.  It is
+A Handshake packet uses long headers with a type value of 0x7C.  It is
 used to carry acknowledgments and cryptographic handshake messages from the
 server and client.
 
@@ -651,7 +651,7 @@ packets protected with 1-RTT keys are sent with short headers.  The different
 packet types explicitly indicate the encryption level and therefore the keys
 that are used to remove packet protection.
 
-Packets protected with 0-RTT keys use a type value of 0x05.  The connection ID
+Packets protected with 0-RTT keys use a type value of 0x7B.  The connection ID
 field for a 0-RTT packet is selected by the client.
 
 The client can send 0-RTT packets after receiving a Handshake packet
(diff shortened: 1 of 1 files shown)


marten-seemann:
I recently implemented the [header parsing logic for quic-go](https://github.com/lucas-clemente/quic-go/blob/master/internal/wire/header.go#L34-L72), such that it can handle both gQUIC and IETF QUIC. You might find my summary of the different formats for the type byte useful: https://docs.google.com/document/d/162BtfR84c6xTvAG6AaPK0dXSQXKahMrtJOXxNnpcwDw/edit?usp=sharing.

The current packet header format has the nice property that it is possible to decide if the packet is a gQUIC or an IETF QUIC packet by just looking at the type byte. This is possible due to the following properties:

### for packets sent by the server

For packets sent by the server the client already knows the QUIC version needed for parsing the packet, as soon as version negotiation is completed. Therefore, the only requirement here is that the type byte used for the Version Negotiation Packet in IETF QUIC is not used in gQUIC, and the type byte used for the gQUIC VNP is not used in IETF QUIC.
This property is guaranteed by the first bit of the type byte alone already (0x80 is unset in gQUIC, and set for an IETF QUIC VNP).

### for packets sent by the client

All gQUIC packets have bits 0x80 and 0x40 unset, and currently IETF QUIC packets either have 0x80 set (for long headers) or 0x40 set (for short headers, if connection ID omission is not allowed, which I assume is what server implementations will do).
This property would be changed by this PR (because the connection ID bit is inverted), so a server wouldn't be able to distinguish between gQUIC and short header packets.

csperkins:
I haven't looked at gQUIC, but does it matter that the "server wouldn't be able to distinguish between gQUIC and short header packets"? Can it tell from the initial long header packets if the flow is gQUIC or IETF QUIC?

marten-seemann:
In order to do that, you would have to associate the packet with a flow, which would mean that the server would have to extract the connection ID from the header first. This might or might not be possible, depending on how we shift around the location of the connection ID.
Furthermore, this would have the not so nice property that packet handling would not be truly stateless anymore (since you'd have to do lookups in the connection maps).

martinthomson:
First, assume that you require a connection ID at the server.  The general case for packets sent by clients in gQUIC is 0b00pp1000.  The change to short headers in this PR makes the typical IETF QUIC type byte 0b00k111pp, though obviously we don't make any promises about those 1 octets (and we will probably want to grease them).

One possible strategy we could pursue here is to disable greasing until we wean people off gQUIC.  Or - perhaps better - negotiate greasing.  Actually, I like that second idea, because it suits Colin's use case very nicely.

marten-seemann:
@martinthomson: Right, I missed that part. Relying on bit 0x4 might be the solution here.
It doesn't change your argument, but clients in gQUIC can set the Version Flag, so the general case will be 0b00pp100v.

In general, I don't think greasing should be negotiated (we should always grease as much as possible), but if that's the only way to make a transition possible, this wouldn't be a bad option.

martinthomson:
I wouldn't have suggested negotiation if not for this realtime case.  There, they probably want us to avoid using certain values (blame RFC 7983).  Negotiation would be a solution to that problem, and as long as the overwhelming majority of QUIC traffic greased, then it would all work out.  Of course, they negotiate sessions and could limit greasing to a narrower range of values.

csperkins:
Right – we likely want to grease by default, but have the ability to disable/limit greasing for certain flows where we know the greasing will conflict with other uses of the port.

dtikhonov:
I did a [write-up about this](https://www.ietf.org/mail-archive/web/quic/current/msg02009.html) some time ago. I think this change will make it easier to differentiate between gQUIC and QUIC.  Types 0x1F, 0x1E, and 0x1D would give us the following gQUIC interpretations, in order:

1. 2-byte packet number; connection ID, nonce, public reset, version bits set.
1. 2-byte packet number; connection ID, nonce, public reset bits set.
1. 2-byte packet number; connection ID, nonce, version bits set.

gQUIC public resets do not have two-byte packet numbers.  Likewise, gQUIC version negotiation packets are unlikely to have two-byte packet numbers.

ianswett:
I think this change doesn't change how difficult it is to support GQUIC and IETF QUIC at once.

The fact we're not moving the connection ID really helps, since we can always lookup by connection ID first, and then if nothing is found, try to parse it as a handshake packet that creates a new connection.

On the client side, the main issue would be version negotiation, since for every other packet, the client knows the version in use.  This VN codepoint is not used in GQUIC, so we should be all set.

But to pop the stack, I'm trying to walk through why the heuristics approach outlined in the draft isn't sufficient in practice.  The negatives like not being able to run DTLS and QUIC at the same time on the same 5-tuple seem like non-issues.  Can someone outline what the issues are that motivate this change?
https://tools.ietf.org/html/draft-aboba-avtcore-quic-multiplexing-01#section-2.3

The issue of greasing is real.  Unfortunately, I think we may need a way to communicate codepoints not to grease for WebRTC, GQUIC, and other future use cases.

If a server is running GQUIC and IETF QUIC at once, then it is actually using all the codepoints, just for two different protocols, so the space is effectively being greased.

martinthomson:
It's possible that we can still grease even for cases where multiplexing with RTP is desired, we just might multiplex less.  Would it help if I wrote up what I am considering specifying for greasing?

dtikhonov:
@ianswett writes:
> I think this change doesn't change how difficult it is to support GQUIC and IETF QUIC at once.

Could you please elaborate why you think that?

And (emphasis added):
> The fact we're not moving the connection ID really helps, since *we can always lookup by connection ID first*, and then if nothing is found, try to parse it as a handshake packet that creates a new connection.

This is quite expensive compared to simply examining the packet header!  One can also imagine a situation where a demultiplexer forwards packets to two separate (processes or machines) QUIC implementations: gQUIC and IETF QUIC.  The demux would not have the associated state.

The fallback to looking up a connection for each packet to determine which QUIC version it is should be used as last resort.  I hope we can avoid it.

marten-seemann:
I'm surprised this was merged already. The problem of stateless packet handling wasn't solved, since relying on bit 0x4 will not work as soon as we start greasing (or change anything else for the short headers).

ianswett:
Agreed on this merging sooner than I expected.  I realize we want to nail down the type byte, but I'm still not sure why we're doing this?

@dtikhonov You seemed to think it made supporting both at once easier than before?  Is that because all the new code points are either undefined or in practice unused in GQUIC?

@martinthomson Yes, I think it would help to have a specific greasing proposal written up that deals with both coexistence of GQUIC and IETF QUIC, as well as IETF QUIC and WebRTC.

csperkins:
> It's possible that we can still grease even for cases where multiplexing with RTP is desired, we just might multiplex less. 

When multiplexing, don’t grease the ranges being used by the protocols that you’re multiplexing with. You’re already exercising the ability to send packets using those bits.

> Would it help if I wrote up what I am considering specifying for greasing?

Likely, yes

dtikhonov:
@ianswett:  The reason I think this change makes it easier is that some of the [potential ambiguities](https://www.ietf.org/mail-archive/web/quic/current/msg02009.html) no longer exist.  The heuristics to tell one version from another using the new packet type numbering scheme based are described in [my original post above](https://github.com/quicwg/base-drafts/pull/956#issuecomment-346706234).  You state that you do not think this change make it easier to differentiate between packet types; I'd like to know why.

ianswett:
Our server always requires connection id to be present on packets it receives, so the ambiguity you mention in your original post is not a problem.

With the previous numbering scheme, on the server side, we can just look at the first two bits.  If either is 1, then it's IETF QUIC, and if both are 0, it's GQUIC.

In the new scheme, I can still check the first bit for the long header, and then I can check the nonce bit, and if it's that's 1, it must be IETF QUIC because the client never sends a diversification nonce.  

But I believe the nonce bit is the bit that would end up being the spin bit, which could start making things pretty tricky.  It's probably worth thinking through what happens if two bits are reserved for measurement.

dtikhonov:
> With the previous numbering scheme, on the server side, we can just look at the first two bits. If either is 1, then it's IETF QUIC, and if both are 0, it's GQUIC.

I see.  If one can assume that 0x40 (multipath) bit is never set in GQUIC, then this is indeed very easy.  I did not make this assumption in my analyses to err on the conservative side.

> But I believe the nonce bit is the bit that would end up being the spin bit, which could start making things pretty tricky. It's probably worth thinking through what happens if two bits are reserved for measurement.

Are you saying that the packet type in the short header would be narrower and occupy fewer than the currently-allocated low five bits?  (I apologize for not being on top of this development.)

marten-seemann:
> But I believe the nonce bit is the bit that would end up being the spin bit, which could start making things pretty tricky.

That would be really unpleasant. What is the reason for inverting the Connection ID flag?
We had a type byte format that worked perfectly in all corner cases, and by merging this PR, we just destroyed this feature and create a big problem for all people who have gQUIC deployed.

ianswett:
Yes, the 0x40 bit is always 0, because we never deployed multipath.

There's been quite a bit of discussion about the spin bit, including a new draft sent out today: 
https://datatracker.ietf.org/doc/html/draft-trammell-quic-spin-00

The current proposal makes the type shorter, but the nonce bit would still always be 1 in the above proposal, so as long as no other bits are used for anything, we should still be able to use the plan of looking at the first bit or the nonce bit to detect IETF QUIC?

But as Marten says, this PR definitely makes things more complex than I originally thought, particularly in light of other potential header changes.

janaiyengar:
I'm surprised to see this PR merged -- I don't think we had agreed on STUN types being necessary to multiplex with. In addition to concerns about GQUIC vs QUIC, there remains the high-level question of what do we need to multiplex QUIC with on the same port. This was a conversation that made good progress in Singapore, and I was expecting to continue it and finalize on list.

janaiyengar:
Let's have discussions on issues and not PRs. Let's please continue this on #426.

ronieven:
My understanding from the meeting was that there is no need to multiplex short header with stun so the change of the type values for the short header is a surprise and I did not see any discussion on the QUIC mailing list about this change. It is not mentioned also in the Singapore meeting notes.

csperkins:
We certainly need to demux QUIC short header packets with STUN, to maintain bindings. 

Changing the packet type values doesn't prevent the spin bit, although it might change which bit is used as the spin bit.

britram:
Note also that #989 would give us more degrees of freedom in the short header first octet.

ekr:
Colin is right here. You have to multiplex short headers with STUN b/c of consent checks, as well as candidates you learn about late

janaiyengar:
Can we please continue this discussion on #426?

Issue #950, opened 2017-11-21
"Push Stream Header" term and carriage


LPardue:
HTTP/QUIC section 4.4 says:

> A push stream is a server-initiated, unidirectional stream. A push stream includes a header (see Figure 1) that identifies the PUSH_PROMISE that it fulfills. This header consists of a Push ID, encoded as a variable-length integer. The Push ID identifies a server push (see Section 5.2.6).

I'm finding the term push stream header a little obtuse because it isn't a QUIC frame header, an HTTP/QUIC frame header, an HTTP header, nor an HPACK header-block, and all of those are in scope for a push. Since it could just be me, I don't have any alternatives to suggest at this stage.

I think what the text is saying is that a push stream is a sequence of QUIC STREAM frames, the first frame must contain the push stream header directly. In other words, there is no additional HTTP/QUIC framing for this (although I do wonder if an INITIATE_PUSH frame, that includes the Push ID might be way more explicit (at the cost of framing overhead)).

> After the push stream header, a push contains a response (Section 4.2), with response headers, a response body (if any) carried by DATA frames, then trailers (if any) carried by HEADERS frames.

For clarity, I wonder if it might be better to say:

_After the push stream header, a push contains a response (Section 4.2), with response header fields carried by HEADERS frames, a response body (if any) carried by DATA frames, then trailer fields (if any) carried by HEADERS frames._

MikeBishop:
Let's back up a second.  QUIC streams are bytestreams as far as the application is concerned.  It has no knowledge or dependency on how the bytestream is segmented for transport.  So the sequence of bytes on the stream is:

  - Stream header (only for push streams at the moment, might expand)
  - HTTP/QUIC Frames

In Martin's original unidirectional proposal, all streams had stream headers identifying the type of stream.  Currently only push streams have them, but a reasonable argument can be made for putting them on the control stream as well.  They're probably necessary if the header compression algorithm has its own control channels, too, so you can differentiate between types of unidirectional streams.

Once the dust settles on that, I'll probably reorganize to describe that the header exists on unidirectional streams, followed by a sequence of frames to the end of the stream.

LPardue:
Right, thanks for me reminding me of that unidirectional PR. I can see now the bits that were migrated over. The confusing bits in the current HTTP/QUIC doc (from my perspective) are the absence of generic definition of what a stream header is (hard when there is just one and it only contains one field), and the overloading of Push ID for different uses. 

[...]

PR #957, opened 2017-11-22
This clarifies where PUSH_PROMISE can be sent and gives a clue about …

### LINKED ISSUE #947: Can PUSH_PROMISE rules be relaxed to allow sending on a control stream?
Reading the latest HTTP/QUIC editor's copy. Section 4.4 on server push states that:

> The PUSH_PROMISE frame is sent on a response stream.

The recent changes in this area have moved away from Stream ID, PUSH_PROMISE frames now reference a Push ID. Push cancellation is achieved by the HTTP/QUIC CANCEL_PUSH frame, which is restricted to the control stream. 

So I got to wondering if there may be benefits allowing/restricting HTTP/QUIC PUSH_PROMISE frames on the control stream. It seems to me that CANCEL_PUSH forces implementations to be geared up for managing server push on the control stream anyway.

Maybe I'm missing the benefit of tying pushes to requests, RFC 7540 talks about the requirements but I can't infer the benefits. Relaxing the rule and incorporating something like cache digests might give an appropriately-configured server enough information to push just the correct things with a bit more performance (it is described as unsolicited right?). Another use case is to allow server push without having to hold streams open.

martinthomson:
…why.

Closes #947, #950.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -368,9 +368,9 @@ push to be associated with a request.  Ordering of a PUSH_PROMISE in relation to
 certain parts of the response is important (see Section 8.2.1 of {{!RFC7540}}).
 
 Unlike HTTP/2, the PUSH_PROMISE does not reference a stream; when a server
-fulfills a promise, the stream that carries the stream headers references the
-PUSH_PROMISE.  This allows a server to fulfill promises in the order that best
-suits its needs.
+fulfills a promise, the stream that carries the stream headers references a Push
+ID.  This allows a server to fulfill promises in the order that best suits its
+needs.
 
 The server push response is conveyed on a push stream.  A push stream is a
 server-initiated, unidirectional stream.  A push stream includes a header (see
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -362,10 +362,15 @@ a MAX_PUSH_ID frame.
...
-the stream that carries the stream headers references the PUSH_PROMISE.  This
-allows a server to fulfill promises in the order that best suits its needs.
+request. The PUSH_PROMISE frame is sent on the client-initiated, bidirectional
+stream that carried the request that generated the push.  This allows the server
+push to be associated with a request.  Ordering of a PUSH_PROMISE in relation to
+certain parts of the response is important (see Section 8.2.1 of {{!RFC7540}}).
+
+Unlike HTTP/2, the PUSH_PROMISE does not reference a stream; when a server
+fulfills a promise, the stream that carries the stream headers references the
+PUSH_PROMISE.  This allows a server to fulfill promises in the order that best
```
LPardue: Minor nit: would it be clearer to say "references the Push ID"? Multiple PUSH_PROMISE can reference same Push ID so a pushed stream is not satisfying any particular promise.

PR #958, opened 2017-11-24
Add separate transport parameter for unidirectional stream limit

### LINKED ISSUE #909: initial_max_stream_id and, uni and bidi streams
Now we have separate uni and bidirectional stream ID spaces, but initial_max_stream_id in transport parameter does not say it applies to 1) both, or 2) uni only, or 3) bidi only.

martinthomson:
Closes #909.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1058,12 +1058,13 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
    enum {
       initial_max_stream_data(0),
       initial_max_data(1),
-      initial_max_stream_id(2),
+      initial_max_stream_id_bidi(2),
       idle_timeout(3),
       omit_connection_id(4),
       max_packet_size(5),
       stateless_reset_token(6),
       ack_delay_exponent(7),
+      initial_max_stream_id_uni(8),
       (65535)
    } TransportParameterId;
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1128,12 +1129,16 @@ initial_max_data (0x0001):
...
-initial_max_stream_id (0x0002):
+initial_max_stream_id_bidi (0x0002):
 
 : The initial maximum stream ID parameter contains the initial maximum stream
-  number the peer may initiate, encoded as an unsigned 32-bit integer.  This is
-  equivalent to sending a MAX_STREAM_ID ({{frame-max-stream-id}}) immediately
-  after completing the handshake.
+  number the peer may initiate for bidirectional streams, encoded as an unsigned
+  32-bit integer.  This value MUST be a valid bidirectional stream ID for a
+  peer-initiated stream (that is, a server includes a value that have the least
```
LPardue: /s/have/has

LPardue:
The slight niggle that I have is the phrase "this is equivalent to sending a MAX_STREAM_ID". In spirit I think this is true, in practical terms, the closest MAX_STREAM_ID gets is a 30-bit varint. Transport params cannot express the full range streams IDs permitted by MAX_STREAM_ID.

draft-ietf-quic-transport.md
ianswett: I think the phrasing below is a bit cleaner, so I'd copy that.  "that is, the two least significant bits are set to 0 by a server and to 1 by a client"

tatsuhiro-t:
Sorry for late, but I think initial_max_stream_id_bidi and initial_max_stream_id_uni should allow 0.
The current text says 0 is not valid.

And also I'm wondering why initial_max_stream_id_uni must be sent by server.  And there is no indication for client although the text says about client case.
Should it put after ack_delay_exponent, I mean, its transmission is MAY.

marten-seemann:
I agree with tatsuhiro-t. Not every QUIC implementation might want to support both uni- and bidirectional streams, so having a way to effectively disable each of the types would be really helpful.
I suggest making 0 a valid value in both cases.

martinthomson:
I have what might be a better suggestion: define these values to default to zero and make them optional.

janaiyengar:
I prefer to have explicit intent with 0, but either works.

martinthomson:
I would also, but I think that this is better for two reasons:

* 0 is hard because it is only valid for client-initiated bidirectional streams, and you would have to create an exception in any validation checks.

* Now that we have multiple types, it is more onerous to explicitly signal for every type, especially if you never plan to use certain type (as we will for server-initiated bidirectional streams in hq).

PR #959, opened 2017-11-24
Move negotiated_version

### LINKED ISSUE #710: Versions in transport parameter force reformatting client hello on renegotiation
The transport parameter extension sent in the Client Hello includes the following:

>      select (Handshake.msg_type) {
>         case client_hello:
>            QuicVersion negotiated_version;
>            QuicVersion initial_version;
>            ...

This is annoying, since the "negotiated version" will not be known until negotiation is complete. As the extension is sent in the first flight with the Client Hello, this means that the value of this Client Hello will change between the first Client Initial packet that triggered a version negotiation, and the repeated Client Initial packet that is sent with the new version number. The text in section 7.1 says:

> Though the contents of a packet might not change in response to version negotiation, a client MUST increase the packet number it uses on every packet it sends. 

If we keep the "negotiated version" in the transport parameter extension, the content of the packet will always change, so this sentence is misleading.

I would prefer to only include the "initial version" in the transport parameter extension. The server can detect shenanigans if the initial version belong to its supported version list, yet is different from the version in the QUIC header. The client can detect shenanigans if the server supported list includes its own initial proposal.

There is an argument that the QUIC header might be forged. In that case, it seems better to include the negotiated version in the server parameters, before the list of supported versions, and trust the client to detect any disparity.

martinthomson:
This makes it possible for a client to use substantially-similar versions of
QUIC without having to regenerate a ClientHello.  That's not a huge benefit,
but given that it is no more vulnerable to downgrade attack than before, it
seems like a worthwhile thing to do.

Note that this will have the effect of violating the terms of the version
downgrade contract.  I do this knowing that this is still very much in
development and that no sane person would deploy this protocol in its current
form expecting it to be secure.  Conveniently, it appears that endpoints that
support the newer version will not be exposed to downgrade, only those that
support versions prior to the change (and even then, it's only if they can
cause the server to emit a transport parameter that looks like the current
version.  That is possible, but for QUIC version 4 only (assuming we retain the
same basic format for that long).

Closes #710.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1231,7 +1231,8 @@ detected.
 The client includes the initial_version field in its transport parameters.  The
 initial_version is the version that the client initially attempted to use.  If
 the server did not send a version negotiation packet {{packet-version}}, this
-will be identical to the negotiated_version.
+will be identical to the negotiated_version field in the server transport
+parameters.
 
 A server that processes all packets in a stateful fashion can remember how
 version negotiation was performed and validate the initial_version value.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1219,14 +1219,17 @@ New transport parameters can be registered according to the rules in
 
 ### Version Negotiation Validation {#version-validation}
 
-The transport parameters include three fields that encode version information.
-These retroactively authenticate the version negotiation (see
-{{version-negotiation}}) that is performed prior to the cryptographic handshake.
+Though the cryptographic handshake has integrity protection, two forms of QUIC
+version downgrade are possible.  In the first, an attacker replaces the QUIC
+version in the Initial packet.  In the second, a fake Version Negotiation packet
+is sent by an attacker.  To protect against these attacks, the transport
+parameters include three fields that encode version information.  These
+parameters are used to retroactively authenticate the choice of version (see
+{{version-negotiation}}).
 
 The cryptographic handshake provides integrity protection for the negotiated
 version as part of the transport parameters (see {{transport-parameters}}).  As
-a result, modification of version negotiation packets by an attacker can be
-detected.
+a result, attacks on version negotiation by an attacker can be detected.
 
 The client includes the initial_version field in its transport parameters.  The
 initial_version is the version that the client initially attempted to use.  If
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1270,8 +1270,8 @@ would have selected the negotiated version.  A client MUST terminate the
 connection with a VERSION_NEGOTIATION_ERROR error code if the current QUIC
 version is not listed in the supported_versions list.  A client MUST terminate
 with a VERSION_NEGOTIATION_ERROR error code if version negotiation occurred but
-it would have selected a different version if the versions in the
-supported_versions list appears in a Version Negotiation packet.
+it would have selected a different version based on the value of the
+supported_versions list.
 
 When an endpoint accepts multiple QUIC versions, it can potentially interpret
 transport parameters as they are defined by any of the QUIC versions it
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1075,10 +1075,10 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
    struct {
       select (Handshake.msg_type) {
          case client_hello:
-            QuicVersion negotiated_version;
             QuicVersion initial_version;
 
          case encrypted_extensions:
+            QuicVersion negotiated_version;
             QuicVersion supported_versions<4..2^8-4>;
 
          case new_session_ticket:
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1228,45 +1228,49 @@ version as part of the transport parameters (see {{transport-parameters}}).  As
...
-connection with a VERSION_NEGOTIATION_ERROR error code if the
-negotiated_version value is not included in the supported_versions list.  A
-client MUST terminate with a VERSION_NEGOTIATION_ERROR error code if
-version negotiation occurred but it would have selected a different version
-based on the value of the supported_versions list.
+connection with a VERSION_NEGOTIATION_ERROR error code if the current QUIC
+version is not listed in the supported_versions list.  A client MUST terminate
+with a VERSION_NEGOTIATION_ERROR error code if version negotiation occurred but
+it would have selected a different version if the versions in the
+supported_versions list appears in a Version Negotiation packet.
```
janaiyengar: This is a run-on. The old sentence reads fine -- leave it as it is.

draft-ietf-quic-transport.md
```diff
@@ -1228,45 +1228,49 @@ version as part of the transport parameters (see {{transport-parameters}}).  As
...
-  the version of QUIC that is in use MUST terminate the connection with a
-  VERSION_NEGOTIATION_ERROR error code.
-
-* The initial_version is the version that the client initially attempted to use.
-  If the server did not send a version negotiation packet {{packet-version}},
-  this will be identical to the negotiated_version.
+The client includes the initial_version field in its transport parameters.  The
+initial_version is the version that the client initially attempted to use.  If
+the server did not send a version negotiation packet {{packet-version}}, this
+will be identical to the negotiated_version.
```
marten-seemann: To the negotiated_version in the encrypted_extensions.

draft-ietf-quic-transport.md
```diff
@@ -1228,45 +1228,49 @@ version as part of the transport parameters (see {{transport-parameters}}).  As
...
+would have accepted the version included in the initial_version and the value
+differs from the QUIC version that is in use, the server MUST terminate the
+connection with a VERSION_NEGOTIATION_ERROR error.
+
+The server includes both the version of QUIC that is in use and a list of the
+QUIC versions that the server supports.
+
+The negotiated_version field is the version that is in use.  This MUST be set by
+the server to the value that is on the Initial packet that it accepts (not an
+Initial packet that triggers a Retry or Version Negotiation packet).  A client
```
marten-seemann: Why the Retry packet?

martinthomson: Both are (potentially) stateless from the server perspective.  You *could* save the packet number in the Retry cookie, but I don't think we need to require that servers do that.  Note that the size of this cookie is strictly limited.  It has to fit into the padding space in the Initial packet.

draft-ietf-quic-transport.md
martinthomson: I'm not allowed to use TLS message names, but I'll use "server transport parameters".

martinthomson:
@huitema, can you review this PR please?

draft-ietf-quic-transport.md
```diff
@@ -1075,10 +1075,10 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
...
       select (Handshake.msg_type) {
          case client_hello:
-            QuicVersion negotiated_version;
             QuicVersion initial_version;
 
          case encrypted_extensions:
+            QuicVersion negotiated_version;
             QuicVersion supported_versions<4..2^8-4>;
 
          case new_session_ticket:
```
huitema: LGTM.

draft-ietf-quic-transport.md
```diff
@@ -1228,45 +1228,50 @@ version as part of the transport parameters (see {{transport-parameters}}).  As
...
-connection with a VERSION_NEGOTIATION_ERROR error code if the
-negotiated_version value is not included in the supported_versions list.  A
-client MUST terminate with a VERSION_NEGOTIATION_ERROR error code if
-version negotiation occurred but it would have selected a different version
-based on the value of the supported_versions list.
+connection with a VERSION_NEGOTIATION_ERROR error code if the current QUIC
+version is not listed in the supported_versions list.  A client MUST terminate
+with a VERSION_NEGOTIATION_ERROR error code if version negotiation occurred but
+it would have selected a different version based on the value of the
+supported_versions list.
```
huitema: That's OK, but I had to read the text several times, so it probably should be clearer. How about something like that:

The whole point is to prevent version downgrade by a third party, by locking the version ID inside TLS extensions protected by the TLS protocol. The two types of downgrade attacks are:

1) MITM spoofing the client initial, replacing the original value by a downgrade value;
2) MITM dropping the client initial and sending a spoofed version negotiation packet that proposes the downgrade value, or alternately MITM intercepting a genuine negotiation packet and reducing the proposed list to a downgrade value.

Both types will be detected. In the first case, the server will see the initial version in the transport parameters sent by the client, and detect that it should have been accepted. In the second case, the client will see that the transport parameters of the server include a version that it would have preferred over the downgrade version.

PR #960, opened 2017-11-27
Make lint faster.

martinthomson:
This makes files to track what has been linted.  It runs `make -k lint`
in CI to avoid having partial results reported in CI.

BTW, I had no idea that this was so slow on windows until I ran it on my Windows box.  Wow.  I appreciate now that 25s is a little slow.  This restores it to 1s on that same machine, which isn't perfect, but certainly is better.  This could probably be improved with a python script for linting rather than the current kludge, but I don't have the cycles for that right now.

Closes #938.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,9 @@ jobs:
           keys:
             - refcache
       - run:
+          name: "Lint Drafts"
+          command: "make 'CLONE_ARGS=--reference ~/i-d-template' -k lint"
+      - run:
           name: "Build Drafts"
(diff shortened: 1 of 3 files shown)


MikeBishop:
I agree with you that this is simpler, but I'm not sure it gives the same advantages, either.  The modified time on a file seems to be the time the branch was checked out, if there was any difference in the file between the old and new branches.  That means this solution forces re-linting for the simple case of hopping back and forth between two branches while working on multiple PRs.  Mine does keep more state, but as a result is pretty good about not re-linting files if they've even been linted in that state.

MikeBishop:
Closing in favor of #1032.

PR #964, opened 2017-11-29
Randomize packet numbers more, not less

martinthomson:
Closes #864, #850.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -743,14 +743,25 @@ have special rules for populating the packet number field.
 
 ### Initial Packet Number {#initial-packet-number}
 
-The initial value for packet number MUST be selected from an uniform random
-distribution between 0 and 2^31-1.  That is, the lower 31 bits of the packet
-number are randomized.  {{?RFC4086}} provides guidance on the generation of
-random values.
+The initial value for packet number MUST be selected randomly from a range
+between 0 and 2^32 - 1024.  This value is selected so that Initial and Handshake
+packets exercise as many possible values for the Packet Number field as
+possible.
 
-The first set of packets sent by an endpoint MUST include the low 32-bits of the
-packet number.  Once any packet has been acknowledged, subsequent packets can
-use a shorter packet number encoding.
+Limiting the range allows both for loss of packets and for any stateless
+exchanges.  Packet numbers are incremented for subsequent packets, but packet
+loss and stateless handling can both mean that the first packet sent by an
+endpoint isn't necessarily the first packet received by its peer.  The first
+packet received by a peer cannot be 2^32 or greater or the recipient will
+incorrectly assume a packet number that is 2^32 values lower and discard the
+packet.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -743,14 +743,25 @@ have special rules for populating the packet number field.
...
+exchanges.  Packet numbers are incremented for subsequent packets, but packet
+loss and stateless handling can both mean that the first packet sent by an
+endpoint isn't necessarily the first packet received by its peer.  The first
+packet received by a peer cannot be 2^32 or greater or the recipient will
+incorrectly assume a packet number that is 2^32 values lower and discard the
+packet.
+
+Use of a secure random number generator {{?RFC4086}} is not necessary for
+generating the initial packet number, nor is it necessary that the value be
+uniformly distributed.
```
phluid61: [oblig.](https://xkcd.com/221/)

draft-ietf-quic-transport.md
```diff
@@ -743,14 +743,25 @@ have special rules for populating the packet number field.
...
+
+Packets with the long header include the low 32-bits of the packet number.
+Packets that use the short header can use a shorter packet number encoding.
```
janaiyengar: These last two sentences seem out of place. This section is about Initial packet number.

janaiyengar: Never mind. This seems fine.

draft-ietf-quic-transport.md
```diff
@@ -743,14 +743,25 @@ have special rules for populating the packet number field.
...
-The initial value for packet number MUST be selected from an uniform random
-distribution between 0 and 2^31-1.  That is, the lower 31 bits of the packet
-number are randomized.  {{?RFC4086}} provides guidance on the generation of
-random values.
-
-The first set of packets sent by an endpoint MUST include the low 32-bits of the
-packet number.  Once any packet has been acknowledged, subsequent packets can
-use a shorter packet number encoding.
+The initial value for packet number MUST be selected randomly from a range
+between 0 and 2^32 - 1024.  This value is selected so that Initial and Handshake
```
janaiyengar: Is this inclusive of 0 and 2 ^ 32 - 1024? I would say yes, and I don't actually care, but the spec should say it.

draft-ietf-quic-transport.md
martinthomson: Sure, I'll delete those.  I was trying to fix a paragraph that wasn't right in the first place.

draft-ietf-quic-transport.md
martinthomson: indeed

draft-ietf-quic-transport.md
martinthomson: I wanted to avoid invoking the spectre of #832, but I will clarify (the easiest is probably [0, 2^32-1024] because that makes a check using a bitmask easier).

PR #966, opened 2017-11-29
Remove remnants of the limit on ACK blocks

### LINKED ISSUE #906: QUIC ack frames now store more than 255 ack ranges
The existing text in the transport and recovery draft indicates QUIC ack frames can ack up to 255 ranges.  Now that we're using variable length integers, we can ack more than that.  

I'm not sure if that's a good thing or a bad thing, but at the least, we should update the text to be self-consistent.

martinthomson:
Closes #906

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -130,7 +130,7 @@ important to the loss detection and congestion control machinery below.
   machinery of QUIC underneath.
 
 * ACK frames contain acknowledgment information.  QUIC uses a SACK-based
-  scheme, where acks express up to 256 ranges.
+  scheme.
 
 ## Relevant Differences Between QUIC and TCP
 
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -130,7 +130,7 @@ important to the loss detection and congestion control machinery below.
   machinery of QUIC underneath.
 
 * ACK frames contain acknowledgment information.  QUIC uses a SACK-based
```
ianswett: Now that I read this, I don't think we want to use the word SACK here, since that has some baggage from TCP.  How about "ACK frames contain one or more ranges of acknowledged packets."?

draft-ietf-quic-transport.md
```diff
@@ -238,7 +238,7 @@ retransmission ambiguity problem.  QUIC acknowledgments also explicitly encode
 the delay between the receipt of a packet and its acknowledgment being sent, and
 together with the monotonically-increasing packet numbers, this allows for
 precise network roundtrip-time (RTT) calculation.  QUIC's ACK frames support up
```
ianswett: remove "up"

PR #968, opened 2017-11-29
Version 0 is Version Negotiation

### LINKED ISSUE #963: One packet type in invariants
The set of packet types is version-dependent, *except* that Version Negotiation has a fixed packet type across all versions.  We could make packet types entirely version-dependent by a simple change:  Version Negotiation packets always indicate a version of '0', and the type field is randomized on send / ignored on receipt.

MikeBishop:
Fixes #963.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -541,9 +541,9 @@ The layout of a Version Negotiation packet is:
 ~~~
 {: #version-negotiation-format title="Version Negotiation Packet"}
 
-The Type field is selected randomly by the server. The Connection ID field echos
-the corresponding value from the triggering client packet.  This allows clients
-some assurance that the server received the packet and that the Version
+The Type field is selected randomly by the server. The Connection ID field
+echoes the corresponding value from the triggering client packet.  This allows
+clients some assurance that the server received the packet and that the Version
 Negotiation packet is in fact from the server.  The Version field MUST be set to
 0x00000000.  The remainder of the Version Negotiation packet is a list of 32-bit
 versions which the server supports.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -551,8 +551,8 @@ Version Negotiation packet is in fact from the server.  The Version field MUST
 be set to 0x00000000.  The remainder of the Version Negotiation packet is a list
 of 32-bit versions which the server supports.
 
-A Version Negotiation packet is never explicitly acknowledged in an ACK frame by
-a client.  Receiving another Initial packet implicitly acknowledges a Version
+A Version Negotiation packet cannot be explicitly acknowledged in an ACK frame
+by a client.  Receiving another Initial packet implicitly acknowledges a Version
 Negotiation packet.
 
 See {{version-negotiation}} for a description of the version negotiation
@@ -777,8 +777,6 @@ constant:
 * the location and size of the Connection ID field in both header forms,
 
 * the location and size of the Version field in long headers,
-
-* the location and size of the Packet Number field in long headers, and
 
 * the format and semantics of the Version Negotiation packet.
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -302,7 +302,7 @@ deployed and used concurrently. Version negotiation is described in
 
 QUIC versions are identified using a 32-bit unsigned number.
 
-The version 0x00000000 is reserved to represent an invalid version.  This
+The version 0x00000000 is reserved to represent version negotiation.  This
 version of the specification is identified by the number 0x00000001.
 
 Version 0x00000001 of QUIC uses TLS as a cryptographic handshake protocol, as
@@ -416,11 +416,10 @@ The following packet types are defined:
 
 | Type | Name                          | Section                     |
 |:-----|:------------------------------|:----------------------------|
-| 0x7F | Version Negotiation           | {{packet-version}}          |
-| 0x7E | Initial                       | {{packet-initial}}          |
-| 0x7D | Retry                         | {{packet-retry}}            |
-| 0x7C | Handshake                     | {{packet-handshake}}        |
-| 0x7B | 0-RTT Protected               | {{packet-protected}}        |
+| 0x7F | Initial                       | {{packet-initial}}          |
+| 0x7E | Retry                         | {{packet-retry}}            |
+| 0x7D | Handshake                     | {{packet-handshake}}        |
+| 0x7C | 0-RTT Protected               | {{packet-protected}}        |
 {: #long-packet-types title="Long Header Packet Types"}
 
 The header form, packet type, connection ID, packet number and version fields of
@@ -512,14 +511,16 @@ from different versions of QUIC are interpreted.
 
 ## Version Negotiation Packet {#packet-version}
 
-A Version Negotiation packet has long headers with a type value of 0x7F and is
-sent only by servers.  The Version Negotiation packet is a response to a client
-packet that contains a version that is not supported by the server.
+A Version Negotiation packet is inherently not version-specific.  Such a packet
+has long headers and can be identified by the Version field being set to zero.
+Version Negotiation packets are sent only by servers.  The Version Negotiation
+packet is a response to a client packet that contains a version that is not
+supported by the server.
 
-The connection ID and version fields echo corresponding values from the
-triggering client packet.  This allows clients some assurance that the
-server received the packet and that the Version Negotiation packet was not
-carried in a packet with a spoofed source address.
+The Connection ID field echos the corresponding value from the triggering client
+packet.  This allows clients some assurance that the server received the packet
+and that the Version Negotiation packet is in fact from the server.  Values for
+the Packet Number and Type fields are selected at random.
 
 A Version Negotiation packet is never explicitly acknowledged in an ACK frame by
 a client.  Receiving another Initial packet implicitly acknowledges a Version
@@ -768,7 +769,7 @@ constant:
 
 * the location and size of the Packet Number field in long headers, and
 
-* the type, format and semantics of the Version Negotiation packet.
+* the format and semantics of the Version Negotiation packet.
 
 Implementations MUST assume that an unsupported version uses an unknown packet
 format. All other fields MUST be ignored when processing a packet that contains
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -512,14 +511,16 @@ from different versions of QUIC are interpreted.
...
+has long headers and can be identified by the Version field being set to zero.
+Version Negotiation packets are sent only by servers.  The Version Negotiation
+packet is a response to a client packet that contains a version that is not
+supported by the server.
 
-The connection ID and version fields echo corresponding values from the
-triggering client packet.  This allows clients some assurance that the
-server received the packet and that the Version Negotiation packet was not
-carried in a packet with a spoofed source address.
+The Connection ID field echos the corresponding value from the triggering client
```
martinthomson: echoes

draft-ietf-quic-transport.md
```diff
@@ -512,14 +511,16 @@ from different versions of QUIC are interpreted.
...
+packet.  This allows clients some assurance that the server received the packet
+and that the Version Negotiation packet is in fact from the server.  Values for
+the Packet Number and Type fields are selected at random.
```
janaiyengar: Nit: I would require this: "Values for ... MUST be randomly chosen."

draft-ietf-quic-transport.md
```diff
@@ -512,25 +511,24 @@ from different versions of QUIC are interpreted.
...
-
-The payload of the Version Negotiation packet is a list of 32-bit versions which
-the server supports, as shown below.
+The layout of a Version Negotiation packet is:
 
 ~~~
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+
+|1|   Type (7)  |
```
martinthomson: Rather than enshrine this as a "Type" field, I think that I'd prefer that this be marked "Unused" instead.

draft-ietf-quic-transport.md
```diff
@@ -543,6 +544,17 @@ the server supports, as shown below.
 ~~~
 {: #version-negotiation-format title="Version Negotiation Packet"}
 
+The value in the Unused field is selected randomly by the server. The Connection
```
marten-seemann: This will make it impossible to determine the packet type by looking at the type byte.

janaiyengar: That is true, but you can look at the type byte and version fields together to determine it. Also note that this is only sent on the server -> client direction, so you're not going to hit boxes that need to scale this sort of check.

marten-seemann: True, but why grease it at all? Note that with this proposal, you can't trust the long header type byte **at all**, since every reserved value could also just be the random value of the version negotiation packet.
This will make parsing long header packets quite painful, you will always have to look at the other fields to determine the true packet byte.
For middle boxes, the version negotiation packet will be clearly distinguishable anyway, so I don't see the need to play tricks with the type byte.

martinthomson: Removing version negotiation from consideration makes it possible to grease the type field completely.  Creating a carve-out for version negotiation makes it a little harder.

Long header packets are rare, so it seems like hitting the slow path is tolerable.

martinthomson:
@MikeBishop did you miss the start of my comment?

draft-ietf-quic-transport.md
marten-seemann: We could exclude the code points defined for Long Header types from the random values that can be used for the greased type byte of the version negotiation packet. That way, we get greasing and easy parsing at the same time.

MikeBishop:
@martinthomson, I don't think I missed it, though it's entirely possible I misinterpreted it.  You said:

> I think that we need to be clearer about the use of the long header here. I don't think that we should say that it has a long header, only that it sets the bit (so that recipients will be able to look for the version field). And we need to explain further about how this is expected to be identified by a recipient.

The PR now says:

> A Version Negotiation packet [...] does not use the packet headers defined above. Upon receipt by a client, it will appear to be a packet using the long header, but will be identified as a Version Negotiation packet based on the Version field.

Is there something specific about the use of the long header that needs to be clearer?

draft-ietf-quic-transport.md
MikeBishop: "the other fields" => version.  Since the type byte is version-specific, you should always confirm that the version is what you expect before attempting to interpret the type byte.  (Should we perhaps mention this in the details of handling incoming packets?)

marten-seemann: Ok, I didn't think about multiple version. Since the type byte version dependent, I'd argue that we shouldn't grease it at all for version negotiation packets.This would simplify the parsing logic, since you could identify the version negotiation packet by just looking at it's type byte (which is a QUIC invariant). For all other packets, you'd know the version anyway (because they're received after version negotiation).
If we want to grease, we could define a greasing mechanism for long header packets.

MikeBishop: But the greasing mechanism for long-header packets would likely be version-specific.  I think the point here is that "version 0" doesn't actually define any semantics for the Type field because it doesn't define any types; it defines a single type, and the value in that field is irrelevant once you've seen the version.

So the parsing logic would be:

```
if( in_handshake ) {
  switch(packet.version) {
    what_i_sent:  switch(packet.type) { ... }
    0:  version_negotiation(packet)
    default:  return;
  }
else {
  if(packet.version != connection.version) return;
  ...
}
```

mikkelfj:
Is bit 6 in first byte of short headers guaranteed to always indicate presence/absence of connection ID in short headers across all versions?
Is bit 7 in first byte guaranteed to alwats indicate long headers across all versions?

A version specific change could relate to co-existence with other protocols. This means very little should be decided about the first byte across versions. But enough to detect version negotiation.

Since detecting presence of version field puts restraints on the first byte it might be good to be very specific about which bits in the first byte remain unaffected across versions.

One the use of random values for greasing, I'm concerned that a dominating implementation in some environments might be sloppy about randomness and just put in the value 0, or 4 always, or by careless with PRNG exposure. I'd prefer some sort of hashing. It could be something simple like FNV-1a or Fletcher-16 seeded by connection ID.

draft-ietf-quic-transport.md
marten-seemann: The principle I'm advocating here (as well as in #426) is that a packet header should be identifiable by just looking at its type byte. So the parsing logic would be:

```
ParseHeader(our_version) {
  if(packet.type_byte == kTypeByteVersionNegotiation) {
    // parse version negotiation packet
  }
  else {
    // version dependent parsing of the packet header (version is our_version)
  }
}
```

Note that the **only** thing in the packet you have to look at to correctly parse the header is the type byte.

MikeBishop: We should move this to #963, since this is a discussion about whether the issue in #963 is a problem or a feature.

MikeBishop:
Martin has a draft with limited circulation so far pulling out the invariants into a separate document that hopefully addresses some of these concerns, @mikkelfj.  The current proposal in that draft is close to what's in 5.8, though:

- The long/short header bit
- The Connection ID and Version fields in long headers
- The Connection ID bit, and the Connection ID field if present, in short headers

draft-ietf-quic-transport.md
marten-seemann: Thanks. I must have missed that issue. I'll write up my argument there.

huitema: This looks like a gratuitous change. What exactly is the benefit of not having a version renegotiation packet?

janaiyengar: @huitema: can we discuss on #963 ?

martinthomson:
I looks like the only point of contention is the value we put in the type field.  Hopefully the draft I'm about to publish will help clarify.  That draft assumes that this PR is merged.  Switching based on the type byte is possible, and it won't appreciably change the logic.

There is no case where you don't check version before handling a packet with a long header.  So there aren't any more or fewer checks, just that they are in a different place.

janaiyengar:
@martinthomson : Agreed, but if you remove that question, i.e., assume that a type is assigned to a VN packet, then the only thing left in this PR is the removal of packet number from the VN packet. I am hearing pushback from the wg on the question of whether the VN type byte should be version-dependent, which is important to this particular PR, and we should discuss on #963 first. It's ok for your draft to assume this PR in the meanwhile.

draft-ietf-quic-transport.md
```diff
@@ -302,7 +302,7 @@ deployed and used concurrently. Version negotiation is described in
 
 QUIC versions are identified using a 32-bit unsigned number.
 
-The version 0x00000000 is reserved to represent an invalid version.  This
+The version 0x00000000 is reserved to represent version negotiation.  This
 version of the specification is identified by the number 0x00000001.
```
ianswett: OT Q: Why isn't version 0 represented by the number 0x00000000?

draft-ietf-quic-transport.md
```diff
@@ -768,7 +780,7 @@ constant:
 
 * the location and size of the Packet Number field in long headers, and
```
ianswett: I thought I removed this from the invariants?  If I missed it, shouldn't it be removed now?

draft-ietf-quic-transport.md
```diff
@@ -543,6 +544,17 @@ the server supports, as shown below.
...
 {: #version-negotiation-format title="Version Negotiation Packet"}
 
+The value in the Unused field is selected randomly by the server. The Connection
+ID field echoes the corresponding value from the triggering client packet.  This
+allows clients some assurance that the server received the packet and that the
+Version Negotiation packet is in fact from the server.  The Version field MUST
+be set to 0x00000000.  The remainder of the Version Negotiation packet is a list
+of 32-bit versions which the server supports.
+
+A Version Negotiation packet is never explicitly acknowledged in an ACK frame by
```
ianswett: It doesn't have a packet number anymore, so how could it be acknowledged?  Maybe change to "cannot be explicitly acknowledged.."

draft-ietf-quic-transport.md
MikeBishop: I'm a little confused by the question.  Are you asking why it's spelled out?

ianswett: I misread these two sentences as "The version 0x00000000 is reserved to represent an invalid version.  This(Ian's misinterpretation: the version negotiation version) version of the specification is identified by the number 0x00000001."

So I misinterpreted the use of the word "This", which is partly a product of the small snippet Github displays, but possibly it makes sense to clarify this now that the spec uses two versions?

PR #974, opened 2017-11-30
Followup to closing/draining changes

### LINKED ISSUE #973: Specify new error type, "closed by peer"?
The updated text in Section 7.9.3, "Immediate Close" states that: "_After receiving a closing frame, endpoints enter a draining period. An endpoint that receives a closing frame MAY send a single packet containing a CONNECTION_CLOSE frame before entering a draining period, but MUST NOT send further packets, which could result in a constant exchange of closing frames until the closing period on either peer ended_."

That's fine, but what error type should the peer use? No error? Should we specify a "closed by peer" error type?

martinthomson:
As Christian points out, we should recommend an error code.  Also, I used the
word "period" in a few places where "state" was more appropriate.  And, I
realized that if a shutdown has been negotiated, an endpoint might as well send
APPLICATION_CLOSE with whatever error is appropriate there rather than to use
CONNECTION_CLOSE, so it's back to closing frame rather than CONNECTION_CLOSE
only for the last gasp packet.

Closes #973.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1699,7 +1699,7 @@ to terminate the connection immediately.  Either closing frame causes all
 streams to immediately become closed; open streams can be assumed to be
 implicitly reset.
 
-After sending a closing frame, endpoints immediately enter a closing period.
+After sending a closing frame, endpoints immediately enter a closing state.
 During the closing period, an endpoint that sends a closing frame SHOULD respond
 to any packet that it receives with another packet containing a closing frame.
 To minimize the state that an endpoint maintains for a closing connection,
@@ -1716,11 +1716,12 @@ Note:
   control, which are not expected to be relevant for a closed connection.
   Retransmitting the final packet requires less state.
 
-After receiving a closing frame, endpoints enter a draining period.  An endpoint
+After receiving a closing frame, endpoints enter a draining state.  An endpoint
 that receives a closing frame MAY send a single packet containing a
-CONNECTION_CLOSE frame before entering a draining period, but MUST NOT send
-further packets, which could result in a constant exchange of closing frames
-until the closing period on either peer ended.
+closing frame before entering a draining state, using the NO_ERROR code if no
+other code is more appropriate.  An endpoint MUST NOT send further packets,
+which could result in a constant exchange of closing frames until the closing
+period on either peer ended.
 
 An immediate close can be used after an application protocol has arranged to
 close a connection.  This might be after the application protocols negotiates a
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1716,12 +1716,12 @@ Note:
   control, which are not expected to be relevant for a closed connection.
   Retransmitting the final packet requires less state.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1716,11 +1716,12 @@ Note:
   control, which are not expected to be relevant for a closed connection.
   Retransmitting the final packet requires less state.
 
-After receiving a closing frame, endpoints enter a draining period.  An endpoint
+After receiving a closing frame, endpoints enter a draining state.  An endpoint
 that receives a closing frame MAY send a single packet containing a
-CONNECTION_CLOSE frame before entering a draining period, but MUST NOT send
-further packets, which could result in a constant exchange of closing frames
-until the closing period on either peer ended.
+closing frame before entering a draining state, using the NO_ERROR code if no
```
huitema: You are replacing the "CONNECTION_CLOSE frame" reference by "closing frame". Actually, I think it is more logical to use the "Connection Close" frame there, since "Application Close" is suppose to only be done at the application's initiative.

martinthomson: oops, yeah, that's obvious now, thanks

draft-ietf-quic-transport.md
```diff
@@ -1699,7 +1699,7 @@ to terminate the connection immediately.  Either closing frame causes all
 streams to immediately become closed; open streams can be assumed to be
 implicitly reset.
 
-After sending a closing frame, endpoints immediately enter a closing period.
+After sending a closing frame, endpoints immediately enter a closing state.
```
janaiyengar: "a" closing state -> "the" closing state, probably.

draft-ietf-quic-transport.md
```diff
@@ -1716,11 +1716,12 @@ Note:
   control, which are not expected to be relevant for a closed connection.
   Retransmitting the final packet requires less state.
 
-After receiving a closing frame, endpoints enter a draining period.  An endpoint
+After receiving a closing frame, endpoints enter a draining state.  An endpoint
```
janaiyengar: ditto

draft-ietf-quic-transport.md
```diff
@@ -1716,11 +1716,12 @@ Note:
...
 that receives a closing frame MAY send a single packet containing a
-CONNECTION_CLOSE frame before entering a draining period, but MUST NOT send
-further packets, which could result in a constant exchange of closing frames
-until the closing period on either peer ended.
+closing frame before entering a draining state, using a CONNECTION_CLOSE frame
+and a NO_ERROR code if no other code is more appropriate.  An endpoint MUST NOT
```
janaiyengar: nit: since it is possible to send an APPLICATION_CLOSE frame as well, how about simply "using a CONNECTION_CLOSE frame and a NO_ERROR code if appropriate."

draft-ietf-quic-transport.md
huitema: I guess the correct text would be "send a CONNECTION CLOSE unless the application has already requested closing the connection." But that gets kind of long, and we can also let the developers decide. In any case, a peer in the closing state has to expect either due to the possibility of simultaneous close, so I guess it does not matter for interop.

PR #975, opened 2017-11-30
Define Min RTT and use it to check ack delay

### LINKED ISSUE #961: Ack delay in SRTT computation
The recovery draft (and GQUIC) use ack delay in SRTT computations, but with the caveat that "the ack delay field from the ACK frame SHOULD be subtracted from the RTT as long as the result is positive." 

The ack delay may be too large if the peer's clock was adjusted, but it can also happen if the peer uses a very granular clock, which can lead to arbitrarily large ack delays.

There are a few ways to deal with this, including:
1. current text, which simply drops the problematic ack delays,
2. ignore ack delays from a peer that screws up once (ack delay > measured RTT),
3. ignore RTT sample that contains problematic ack delay.

(1) is a bit aggressive. (2) is conservative: If ack delay is screwed up then you won't always be able to detect it, but once can be treated as a signal to not trust this peer's ack delays. (3) has been proposed but I'm not sure it addresses the right thing -- the RTT sample isn't bad, the ack delay is.

I'm leaning towards (2).

ianswett:
Closes #961

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -600,8 +600,8 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
      // min_rtt ignores ack delay.
      min_rtt = min(min_rtt, latest_rtt)
      // Adjust for ack delay if it's plausible.
-     if (latest_rtt - min_rtt > ack.ack_delay):
-       latest_rtt -= ack.delay
+     if (latest_rtt - min_rtt > ack_delay):
+       latest_rtt -= ack_delay
      // Based on {{?RFC6298}}.
      if (smoothed_rtt == 0):
        smoothed_rtt = latest_rtt
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -199,10 +199,14 @@ RTT is calculated when an ACK frame arrives by computing the difference
 between the current time and the time the largest newly acked packet was sent.
 If no packets are newly acknowledged, RTT cannot be calculated. When RTT is
 calculated, the ack delay field from the ACK frame SHOULD be subtracted from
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -199,11 +199,18 @@ RTT is calculated when an ACK frame arrives by computing the difference
...
 calculated, the ack delay field from the ACK frame SHOULD be subtracted from
-the RTT as long as the result is positive.
+the RTT as long as the result is larger than the Min RTT.  If the result is
+smaller than the min_rtt, the RTT should be used, but the ack delay field
+should be ignored.
 
 Like TCP, QUIC calculates both smoothed RTT and RTT variance as specified in
 {{?RFC6298}}.
 
+Min RTT is the minimum RTT measured over the connection, prior to adjusting
```
marten-seemann: A connection might be long lived, so it kind of feels weird to have this value calculated over the whole lifetime of the connection. Maybe this is ok, since we're only using Min RTT as a sanity check for the ACK delay sent by the peer.

draft-ietf-quic-recovery.md
```diff
@@ -588,6 +597,11 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
 
 
    UpdateRtt(latest_rtt):
```
marten-seemann: Add the second parameter here.

draft-ietf-quic-recovery.md
ianswett: BBR does use windowed min, but I think this is a fairly standard definition of min_rtt.

janaiyengar: I think it's ok... we can get fancy here (with windowed min), but I think this is good for a simple safeguard.

PR #976, opened 2017-11-30
Fix errors in handshake diagram

martinduke:
This diagram didn't completely catch up with the conversion to using AES-GCM on cleartext packets. I believe I've caught all the errors.


martinthomson:
Thanks Martin!

PR #977, opened 2017-12-01
Remove unnecessary pacing text and reword RTO text

### LINKED ISSUE #967: congestion control in bytes, but pacing advice in packets
"QUIC congestion control is specified in bytes due to finer control and the ease of appropriate byte counting" 

.. and then later ..

"In order to fairly compete with flows that are not pacing, it is recommended to not pace the first 10 sent packets when exiting quiescence."

Can we reasonably reword that advice into bytes? 10 MTU worth?

janaiyengar:
Closes #967

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -856,12 +856,13 @@ Acknowledgement or loss of tail loss probes are treated like any other packet.
 ## Retransmission Timeout
 
 When retransmissions are sent due to a retransmission timeout alarm, no change
-is made to the congestion window until the next acknowledgement arrives.  When
-an ack arrives, if packets prior to the first retransmission timeout are
-acknowledged, then the congestion window remains the same.  If no packets prior
-to the first retransmission timeout are acknowledged, the retransmission timeout
-has been validated and the congestion window must be reduced to the minimum
-congestion window and slow start is begun.
+is made to the congestion window until the next acknowledgement arrives.  The
+retransmission timeout is considered spurious when an ack is received that
+acknowledges packets sent prior to the first retransmission timeout.  The
+retransmission timeout is considered valid when an ack is received that
+acknowledges no packets sent prior to the first retransmission timeout.  In this
+case, the congestion window MUST be reduced to the minimum congestion window and
+slow start is re-entered.
 
 ## Pacing
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -865,8 +865,11 @@ congestion window and slow start is begun.
 
 ## Pacing
 
-It is RECOMMENDED that senders pace sending of all data, distributing the
-congestion window over the SRTT.
+It is RECOMMENDED that a sender pace sending of all data, distributing the
+congestion window over the SRTT.  This document does not specify a pacer.  As an
+example pacer, implementers are referred to the Fair Queue packet scheduler (fq
+qdisc) in Linux (3.11 onwards) as a well-known and publicly available
+implementation of a flow pacer.
 
 ## Pseudocode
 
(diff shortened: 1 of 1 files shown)


marten-seemann:
Was the text about the pacing rate just unnecessary or was it wrong? I recently implemented the pacing rate previously described here in quic-go, and the results so far look promising.

janaiyengar:
Not wrong -- it is what GQUIC does -- but those constants don't add much value, IMO. We do 2x during slow start because that's what the Linux fq qdisc does. I'm wary of specifying in an RFC something that's done by one implementation of pacing. You could choose to use 1.25x for the entire thing and I suspect you'll still be quite fine.

janaiyengar:
Happy to chat, and happy to open an issue for WG consideration. The spec currently does not specify a full pacer, which is appropriately out of scope. Linux's fq is not defined anywhere except in Linux code, and these constants apply to that flavor of pacer. (In contrast, the other constants we use apply to documented mechanisms, but we are borrowing constants only from Linux.) Simply stating these constants is incomplete, which is why I wanted to remove this entirely. I could change the text to have some sort of more generic language about GQUIC and Linux fq implementations; let me give that a shot.

draft-ietf-quic-recovery.md
```diff
@@ -850,30 +850,26 @@ the reduction to once per round trip.
...
-congestion avoidance.  In order to fairly compete with flows that are not
-pacing, it is recommended to not pace the first 10 sent packets when
-exiting quiescence.
+When retransmissions are sent due to a retransmission timeout alarm, no change
+is made to the congestion window until the next acknowledgement arrives.  When
+an ack arrives, if packets prior to the first retransmission timeout are
+acknowledged, then the congestion window remains the same.  If no packets prior
+to the first retransmission timeout are acknowledged, the retransmission timeout
+has been validated and the congestion window must be reduced to the minimum
+congestion window and slow start is begun.
```
ianswett: nit: begun is weird in retrospect.  Maybe re-entered?

janaiyengar: reworded paragraph

draft-ietf-quic-recovery.md
```diff
@@ -856,12 +856,13 @@ Acknowledgement or loss of tail loss probes are treated like any other packet.
...
 
 When retransmissions are sent due to a retransmission timeout alarm, no change
-is made to the congestion window until the next acknowledgement arrives.  When
-an ack arrives, if packets prior to the first retransmission timeout are
-acknowledged, then the congestion window remains the same.  If no packets prior
-to the first retransmission timeout are acknowledged, the retransmission timeout
-has been validated and the congestion window must be reduced to the minimum
-congestion window and slow start is begun.
+is made to the congestion window until the next acknowledgement arrives.  The
+retransmission timeout is considered spurious when an ack is received that
```
ianswett: nit: when the next ack acknowleges?

draft-ietf-quic-recovery.md
```diff
@@ -856,12 +856,13 @@ Acknowledgement or loss of tail loss probes are treated like any other packet.
...
+acknowledges packets sent prior to the first retransmission timeout.  The
+retransmission timeout is considered valid when an ack is received that
```
ianswett: same nit

draft-ietf-quic-recovery.md
janaiyengar: done

draft-ietf-quic-recovery.md
janaiyengar: done

janaiyengar:
Merging. Please add a comment if you want to change something post-merge, happy to do it.

ianswett:
Nope, LGTM, thanks for the improved text.

PR #985, opened 2017-12-05
Restructure packetization section

martinthomson:
This includes the changes from #849, which I have simply moved around.

I think that moving out the discussion on packet size to a subsection helps focus the attention on the most important part of this section, which is the discussion of how retransmission of frames works.

Closes #849.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2740,6 +2740,15 @@ that it sends.
 Strategies and implications of the frequency of generating acknowledgments are
 discussed in more detail in {{QUIC-RECOVERY}}.
 
+# Special Considerations for Packetization Layer PMTU Discovery
+
+The PADDING frame provides a useful option for PMTU probe packets that does not
+exist in other transports. QUIC Packets with no frames other than PADDING must be
+acknowledged, but need not be retransmitted if lost. PADDING probe packets, if
+lost, also do not delay delivery of application data. Their only impact on
+application data is the extent that they consume the congestion window and
+connection flow control credits.
+
 ## Special Considerations for PMTU Discovery
 
 Traditional ICMP-based path MTU discovery in IPv4 {{!RFC1191}} is potentially
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2776,12 +2776,13 @@ actually lost.
 
 ### Special Considerations for Packetization Layer PMTU Discovery
 
-The PADDING frame provides a useful option for PMTU probe packets that does
-not exist in other transports. PADDING frames generate acknowledgements, but
-their content need not be delivered reliably. PADDING frames may delay the
-delivery of application data, as they consume the congestion window. However,
-by definition their likely loss in a probe packet does not require delay-
-inducing retransmission of application data.
+
+The PADDING frame provides a useful option for PMTU probe packets that does not
+exist in other transports. PADDING frames generate acknowledgements, but their
+content need not be delivered reliably. PADDING frames may delay the delivery of
+application data, as they consume the congestion window. However, by definition
+their likely loss in a probe packet does not require delay- inducing
+retransmission of application data.
 
 When implementing the algorithm in Section 7.2 of {{!RFC4821}}, the initial
 value of search_low SHOULD be consistent with the IPv6 minimum packet size.
@@ -2789,9 +2790,8 @@ Paths that do not support this size cannot deliver Initial packets, and
 therefore are not QUIC-compliant.
 
 Section 7.3 of {{!RFC4821}} discusses tradeoffs between small and large
-increases in the size of probe packets. As QUIC probe packets need not
-contain application data, aggressive increases in probe size carry fewer
-consequences.
+increases in the size of probe packets. As QUIC probe packets need not contain
+application data, aggressive increases in probe size carry fewer consequences.
 
 
 # Streams: QUIC's Data Structuring Abstraction {#streams}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -154,8 +154,7 @@ Connection ID:
 
 QUIC packet:
 
-: A well-formed UDP payload that can be parsed by a QUIC receiver.  QUIC packet
-  size in this document refers to the UDP payload size.
+: A well-formed UDP payload that can be parsed by a QUIC receiver.
 
 
 ## Notational Conventions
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2740,7 +2697,57 @@ that it sends.
 Strategies and implications of the frequency of generating acknowledgments are
 discussed in more detail in {{QUIC-RECOVERY}}.
 
-## Special Considerations for PMTU Discovery
+
+## Packet Size
+
+Clients MUST ensure that the first packet in a connection, and any
+retransmissions of those octets, has a QUIC packet size of least 1200 octets.
+The packet size for a QUIC packet includes the QUIC header and integrity check,
```
martinduke: s/"packet size for QUIC packet"/"QUIC packet size"

Keep the terminology consistent, and shorter

draft-ietf-quic-transport.md
```diff
@@ -2740,7 +2697,57 @@ that it sends.
...
+but not the UDP or IP header.
+
+The initial client packet SHOULD be padded to exactly 1200 octets unless the
```
martinduke: s/initial client/QUIC initlal

draft-ietf-quic-transport.md
```diff
@@ -2740,7 +2697,57 @@ that it sends.
...
+client has a reasonable assurance that the Path Maximum Transmission Unit (PMTU)
+is larger.  Sending a packet of this size ensures that the network path supports
+an MTU of this size and helps reduce the amplitude of amplification attacks
+caused by server responses toward an unverified client address.
+
+Servers MUST ignore an initial plaintext packet from a client if its total size
```
martinduke: s/initial plaintext packet/QUIC initial packet

draft-ietf-quic-transport.md
```diff
@@ -2767,6 +2774,26 @@ provisional until QUIC's loss detection algorithm determines that the packet is
...
 
 
+### Special Considerations for Packetization Layer PMTU Discovery
+
+
+The PADDING frame provides a useful option for PMTU probe packets that does not
+exist in other transports. PADDING frames generate acknowledgements, but their
+content need not be delivered reliably. PADDING frames may delay the delivery of
+application data, as they consume the congestion window. However, by definition
+their likely loss in a probe packet does not require delay- inducing
```
martinduke: Eliminate space after dash.

draft-ietf-quic-transport.md
```diff
@@ -2740,7 +2697,57 @@ that it sends.
 Strategies and implications of the frequency of generating acknowledgments are
 discussed in more detail in {{QUIC-RECOVERY}}.
 
-## Special Considerations for PMTU Discovery
+
+## Packet Size
+
+Clients MUST ensure that the first packet in a connection, and any
+retransmissions of those octets, has a QUIC packet size of least 1200 octets.
```
janaiyengar: grammar nit: s/retransmissions/retransmission/ , since it's followed by "has a"

draft-ietf-quic-transport.md
janaiyengar: I don't think we need this; "QUIC packet size" is defined in Section 2.

draft-ietf-quic-transport.md
janaiyengar: Perhaps "QUIC Initial and Handshake packets" ?

draft-ietf-quic-transport.md
```diff
@@ -2740,7 +2697,57 @@ that it sends.
...
+
+## Path Maximum Transmission Unit
+
+The Path Maximum Transmission Unit (PMTU) is the maximum size of the entire IP
+header, UDP header, and UDP payload. The UDP payload includes the QUIC packet
+header, protected payload, and any authentication fields.
+
+All QUIC packets SHOULD be sized to fit within the estimated PMTU to avoid IP
+fragmentation or packet drops. To optimize bandwidth efficiency, endpoints
+SHOULD use Packetization Layer PMTU Discovery ({{!PLPMTUD=RFC4821}}) and MAY use
```
janaiyengar: readability nit: long sentence, which is fine, but I would split it at the "and": "... SHOULD use Packetization Layer PMTU Discovery ({{!PLPMTUD=RFC4821}}). Endpoints MAY use ..."

draft-ietf-quic-transport.md
```diff
@@ -2740,7 +2697,57 @@ that it sends.
...
+PMTU, setting the PMTU appropriately, and storing the result of previous PMTU
+determinations.
+
+In the absence of these mechanisms, QUIC endpoints SHOULD NOT send IP packets
+larger than 1280 octets. Assuming the minimum IP header size, this results in
+a QUIC packet size of 1232 octets for IPv6 and 1252 octets for IPv4.
+
+QUIC endpoints that implement any kind of PMTU discovery SHOULD maintain an
+estimate for each combination of local and remote IP addresses (as each pairing
+could have a different maximum MTU in the path).
```
janaiyengar: suggestion: remove the round brackets, and make the parenthetical phrase part of the main sentence.

draft-ietf-quic-transport.md
```diff
@@ -2740,7 +2697,57 @@ that it sends.
...
+
+QUIC depends on the network path supporting a MTU of at least 1280 octets. This
+is the IPv6 minimum and therefore also supported by most modern IPv4 networks.
```
janaiyengar: nit: s/the IPv6 minimum/the IPv6 minimum packet size/

draft-ietf-quic-transport.md
```diff
@@ -2740,7 +2697,57 @@ that it sends.
...
+An endpoint MUST NOT reduce their MTU below this number, even if it receives
```
janaiyengar: nit: s/their MTU/its MTU/

draft-ietf-quic-transport.md
```diff
@@ -2740,7 +2697,57 @@ that it sends.
...
+signals that indicate a smaller limit might exist.
+
+If a QUIC endpoint determines that the PMTU between any pair of local and remote
```
janaiyengar: There's obviously a good reason for this, which I can't seem to remember. Why do we disallow smaller packets? Can you add a note here about that as well, so it's obvious to any reader?

mikkelfj: My reading of this is that the maximum must not be reduced, but if you have less to send, go ahead. This will trigger packet loss and eventual connection failure when sending bulk data on a network that does not honor the 1200 minimum and therefore possibly trigger a connection migration if possible, or otherwise stop a hopeless quest.

mikkelfj: Another reason may be to prevent DoS attacks by injection of unverifiable network congestion control / status data.

draft-ietf-quic-transport.md
martinthomson: I never realized that "QUIC packet size" was defined there.  I think that this is the better place to add that definition.

draft-ietf-quic-transport.md
martinthomson: It's only Initial.  I'll reword this.

draft-ietf-quic-transport.md
martinthomson: Moar rewording coming up.

draft-ietf-quic-transport.md
martinthomson: or a new sentence

draft-ietf-quic-transport.md
martinthomson: MTU, not packet size (a 1280 MTU implies a smaller packet size)

draft-ietf-quic-transport.md
martinthomson: I'm going to request that you open an issue if you care about this.  This is old text and I'm not really equipped to deal with it right now.  I *think* that @mikkelfj is approximately right.  It might also be to avoid using paths that will drop random packets.

Frankly, I don't think that we need the requirement.  If you *know* that this has happened, you can deal.  The only real cost at that point is that you might be unable to create a new connection on that path.

PR #988, opened 2017-12-05
Minor mods to stream state machine text

janaiyengar:
Some rewording and added one normative.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3087,11 +3087,10 @@ state as a result of delayed delivery of packets.
 ### Bidirectional Stream States {#stream-bidi-states}
 
 A bidirectional stream is composed of a send stream and a receive stream.
-Implementations might represent the state of the bidirectional stream as a
-composite of the states of the constituent send and receive streams.  The
-simplest model presents the stream as "open" when either send or receive stream
-is in a non-terminal state and "closed" when both send and receive streams are
-in a terminal state.
+Implementations may represent states of the bidirectional stream as composites
+of send and receive stream states.  The simplest model presents the stream as
+"open" when either send or receive stream is in a non-terminal state and
+"closed" when both send and receive streams are in a terminal state.
 
 {{stream-bidi-mapping}} shows a more complex mapping of bidirectional stream
 states that loosely correspond to the stream states in HTTP/2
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3087,16 +3087,17 @@ state as a result of delayed delivery of packets.
 ### Bidirectional Stream States {#stream-bidi-states}
 
 A bidirectional stream is composed of a send stream and a receive stream.
-Implementations might present the state of the pair of streams as a composite of
-the states of the pair of streams.  The simplest model presents the stream as
-"open" when either send or receive stream is in a non-terminal state and
-"closed" when both send and receive streams are in a terminal state.
+Implementations might represent the state of the bidirectional stream as a
+composite of the states of the constituent send and receive streams.  The
```
ianswett: nits: 
I a prefer lower case 'may' to 'might' in the line above.
Should it states be plural?
I think constituent is unnecessary

draft-ietf-quic-transport.md
```diff
@@ -3140,15 +3141,15 @@ STOP_SENDING frame is received on a send stream that is already in the "Data
...
 
 An endpoint is expected to send another STOP_SENDING frame if a packet
-containing the frame is lost.  However, once either all stream data or a
-RST_STREAM frame has been received for the stream - that is, the stream is in
-any state other than "Recv" or "Size Known" - sending a STOP_SENDING frame is
-unnecessary.
+containing a previous STOP_SENDING is lost.  However, once either all stream
+data or a RST_STREAM frame has been received for the stream - that is, the
+stream is in any state other than "Recv" or "Size Known" - sending a
+STOP_SENDING frame is unnecessary.
```
ianswett: Above you changed it so it's SHOULD.  Should it be changed here as well?

draft-ietf-quic-transport.md
janaiyengar: done

draft-ietf-quic-transport.md
janaiyengar: I think of this as justification for why it's a SHOULD above, so I'd prefer to leave this text as is.

martinthomson: SHOULD doesn't make sense here.

Issue #912, opened 2017-11-12
No way to explicitly communicate an ack delay


ianswett:
Currently the recovery draft specifies a default ack delay of 25ms, which may be slightly too low in some environments and much too high in others(ie: datacenters).  In order to implement efficient loss recovery, the peer needs to know what ack delay is being used.  This should be communicated via transport params.

A similar proposal exists to add this to TCP, we should ensure it exists in QUIC as well.
https://www.ietf.org/proceedings/97/slides/slides-97-tcpm-tcp-options-for-low-latency-00.pdf

ianswett:
Discussion at IETF 100 included noting that this needs to be the max ack delay, and not an effort to communicate the exact delay in all circumstances.  Also, if it's in transport params, then it's not encrypted.  There was interest in communicating this to better inform TLP, however.

pravb:
Yes this is a deficiency in TCP and makes TLP less effective in cases where there is a single tail packet outstanding. I would request that we at least communicate the max ACK delay value in transport parameters. I would think communicating any changing values over time is optional and I would be fine with that not being included in V1. I also think it is likely that most implementations will simply pick a constant value for ACK delay rather than a dynamically tuned one.

pravb:
I don't want to conflate this with another related issue but communicating the maximum stretch ACking value back to sender can also be useful to implement ACK clocked congestion control like Reno. Today TCP implementations are using different thresholds for ABC and in many cases are sub-optimally growing the cwnd.

janaiyengar:
Closing this in favor of #981

PR #991, opened 2017-12-05
Specify MaxAckDelay for TLP and RTO

### LINKED ISSUE #981: Explicitly communicate and include ACK delay in RTO computations
The recovery spec says:

> RTT is calculated when an ACK frame arrives by computing the difference between the current time and the time the largest newly acked packet was sent. If no packets are newly acknowledged, RTT cannot be calculated. When RTT is calculated, the ack delay field from the ACK frame SHOULD be subtracted from the RTT as long as the result is larger than the Min RTT. If the result is smaller than the min_rtt, the RTT should be used, but the ack delay field should be ignored.

This is fine advice when computing the RTT, but there is a small issue arising later:

> When the final TLP packet is sent, the RTO period is set to max(SRTT + 4*RTTVAR, minRTO)

I believe that this RTO formula should also include some estimation of the ACK delay. When doing simulation with low latency links, I observe that SRTT converges to the low delay and RTTVAR to about zero. The main element in the overall delay is in fact the ACK delay. This leads to two possible solutions:

* Include the ACK Delay in the computation of the SRTT and RTTVAR,
* or, include an AckDelayVar component in the formulation of the RTO.

What do you think?

ianswett:
Closes #981

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -310,10 +310,13 @@ conditions:
   PTO SHOULD be scheduled for min(RTO, PTO).
 
 MaxAckDelay is the maximum ack delay supplied in an incoming ack frame.
-Ack delays that are aren't used for an RTT sample or reference ack-only packets
-are excluded.  When there is exactly one unacknowledged packet, the alarm
-duration includes time for a delayed acknowledgment to be received by including
-MaxAckDelay.
+MaxAckDelay excludes Ack delays that aren't included in an RTT sample because
+they're too large and those which reference an ack-only packet.  When there is
+exactly one unacknowledged packet, the alarm duration includes time for a
+delayed acknowledgment to be received by including MaxAckDelay.
+
+QUIC diverges from TCP by calculating MaxAckDelay dynamically, instead of
+assuming a constant value for all connections.
 
 A PTO value of at least 2*SRTT ensures that the ACK is overdue. Using a PTO of
 exactly 1*SRTT may generate spurious probes, and 2*SRTT is simply the next
@@ -370,9 +373,16 @@ one significantly increases resilience to packet drop in both directions, thus
 reducing the probability of consecutive RTO events.
 
 QUIC's RTO algorithm differs from TCP in that the firing of an RTO alarm is not
-considered a strong enough signal of packet loss. An RTO alarm fires only when
+considered a strong enough signal of packet loss, so does not result in an
+immediate change to CWND or recovery state. An RTO alarm fires only when
 there's a prolonged period of network silence, which could be caused by a change
 in the underlying network RTT.
+
+QUIC also diverges from TCP by including MaxAckDelay in the RTO period.  QUIC is
+able to explicitly model the ack delay via the ack delay field in the ack frame.
+Because QUIC is subtracting this delay from both SRTT and it is expected to
+reduce RTTVar, it is necessary to add it back, in order to compensate for
+the smaller SRTT and RTTVar.
 
 When an acknowledgment is received for a packet sent on an RTO event, any
 unacknowledged packets with lower packet numbers than those acknowledged MUST be
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -300,11 +300,7 @@ receiver.
 The alarm duration, or Probe Timeout (PTO), is set based on the following
 conditions:
 
-* If there is exactly one unacknowledged packet, PTO SHOULD be scheduled for
-  max(2*SRTT, 1.5*SRTT+MaxAckDelay)
-
-* If there are more than one unacknowledged packets, PTO SHOULD be scheduled for
-  max(2*SRTT, 10ms).
+* PTO SHOULD be scheduled for max(1.5*SRTT+MaxAckDelay, 10ms)
 
 * If RTO ({{rto}}) is earlier, schedule a TLP alarm in its place. That is,
   PTO SHOULD be scheduled for min(RTO, PTO).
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -301,21 +301,19 @@ The alarm duration, or Probe Timeout (PTO), is set based on the following
...
 
-kDelayedAckTimeout is the expected delayed ACK timer.  When there is exactly one
-unacknowledged packet, the alarm duration includes time for a delayed
-acknowledgment to be received by including kDelayedAckTimeout.
-
-The RECOMMENDED value for kDelayedAckTimeout is 25ms.
-
-(TODO: Add negotiability of delayed ack timeout.)
+MaxAckDelay is the maximum ack delay supplied in an incoming ack frame.
+Ack delays that are aren't used for an RTT sample or reference ack-only packets
```
janaiyengar: nit: It's not clear what it means to say Ack delays that reference ack-only packets... I would leave this at "aren't used for an RTT sample.", and remove "or reference ack-only packets"

draft-ietf-quic-recovery.md
```diff
@@ -517,9 +520,9 @@ transmit or exceeding the reordering window in time.
 sent_packets:
 : An association of packet numbers to information about them, including a number
   field indicating the packet number, a time field indicating the time a packet
-  was sent, and a bytes field indicating the packet's size.  sent_packets is
-  ordered by packet number, and packets remain in sent_packets until
-  acknowledged or lost.
+  was sent, a boolean indicating whether the packet is ack only, and a bytes
```
janaiyengar: s/ack only/ackable. I think this constraint needs to be more than just ack only packets.

draft-ietf-quic-recovery.md
```diff
@@ -569,6 +573,7 @@ Pseudocode for OnPacketSent follows:
    largest_sent_packet = packet_number
    sent_packets[packet_number].packet_number = packet_number
    sent_packets[packet_number].time = now
+   sent_packets[packet_number].ack_only = is_ack_only
```
janaiyengar: nit: is_ackable

draft-ietf-quic-recovery.md
```diff
@@ -602,6 +607,10 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
      // Adjust for ack delay if it's plausible.
      if (latest_rtt - min_rtt > ack_delay):
        latest_rtt -= ack_delay
+       // Only save into max ack delay if it's used
+       // for rtt calculation and is not ack only.
```
janaiyengar: ditto

draft-ietf-quic-recovery.md
```diff
@@ -300,30 +300,24 @@ receiver.
...
-(TODO: Add negotiability of delayed ack timeout.)
+MaxAckDelay is the maximum ack delay supplied in an incoming ack frame.
+MaxAckDelay excludes ack delays that aren't included in an RTT sample because
+they're too large and excludes those which reference an ack-only packet.
 
-A PTO value of at least 2*SRTT ensures that the ACK is overdue. Using a PTO of
-exactly 1*SRTT may generate spurious probes, and 2*SRTT is simply the next
-integral value of RTT.
+QUIC diverges from TCP by calculating MaxAckDelay dynamically, instead of
+assuming a constant value for all connections.  It includes this in all probe
```
janaiyengar: TCP doesn't assume a constant value for all connections. I would say something like "instead of implicitly including it in the SRTT and RTTVAR estimates".

draft-ietf-quic-recovery.md
```diff
@@ -300,30 +300,24 @@ receiver.
...
+timeouts, because it assume the ack delay may come into play, regardless of
+the number of packets outstanding.  TCP's TLP assumes if at least 2
+packets are outstanding, acks will not be delayed.
 
-The values of 2 and 1.5 are based on
-{{?LOSS-PROBE=I-D.dukkipati-tcpm-tcp-loss-probe}}, but implementations MAY
-experiment with other constants.
+A PTO value of at least 1.5*SRTT ensures that the ACK is overdue.  The of 1.5
```
janaiyengar: missing word: "The value of".

draft-ietf-quic-recovery.md
```diff
@@ -372,10 +366,17 @@ one significantly increases resilience to packet drop in both directions, thus
 reducing the probability of consecutive RTO events.
 
 QUIC's RTO algorithm differs from TCP in that the firing of an RTO alarm is not
-considered a strong enough signal of packet loss. An RTO alarm fires only when
+considered a strong enough signal of packet loss, so does not result in an
+immediate change to CWND or recovery state. An RTO alarm fires only when
```
janaiyengar: is CWND defined by here? or should it be "congestion window"? I'd prefer spelling it out in any case.

draft-ietf-quic-recovery.md
```diff
@@ -372,10 +366,17 @@ one significantly increases resilience to packet drop in both directions, thus
...
 there's a prolonged period of network silence, which could be caused by a change
 in the underlying network RTT.
 
+QUIC also diverges from TCP by including MaxAckDelay in the RTO period.  QUIC is
+able to explicitly model the ack delay via the ack delay field in the ack frame.
+Because QUIC is subtracting this delay from both SRTT and it is expected to
```
janaiyengar: s/is subtracting/subtracts/
It doesn't subtract this delay from SRTT... also this sentence is a bit of a run-on. Suggestion: "Since QUIC ignores this delay in its SRTT and RTTVAR computations, it is necessary to add this delay explicitly in the RTO computation.".

draft-ietf-quic-recovery.md
ianswett: I state that below, but I updated this text to be a bit more specific.

draft-ietf-quic-recovery.md

draft-ietf-quic-recovery.md
ianswett: Agreed, fixed.

draft-ietf-quic-recovery.md
ianswett: As discussed, in a future PR.

draft-ietf-quic-recovery.md
ianswett: ack

draft-ietf-quic-recovery.md
ianswett: ack

ianswett:
Also fixes #912

draft-ietf-quic-recovery.md
```diff
@@ -300,30 +300,24 @@ receiver.
...
+MaxAckDelay is the maximum ack delay supplied in an incoming ack frame.
+MaxAckDelay excludes ack delays that aren't included in an RTT sample because
+they're too large and excludes those which reference an ack-only packet.
 
-A PTO value of at least 2*SRTT ensures that the ACK is overdue. Using a PTO of
-exactly 1*SRTT may generate spurious probes, and 2*SRTT is simply the next
-integral value of RTT.
+QUIC diverges from TCP by calculating MaxAckDelay dynamically, instead of
+assuming a constant value for all connections.  It includes this in all probe
+timeouts, because it assume the ack delay may come into play, regardless of
```
janaiyengar: s/assume/assumes

draft-ietf-quic-recovery.md
```diff
@@ -372,9 +366,15 @@ one significantly increases resilience to packet drop in both directions, thus
...
-considered a strong enough signal of packet loss. An RTO alarm fires only when
-there's a prolonged period of network silence, which could be caused by a change
-in the underlying network RTT.
+considered a strong enough signal of packet loss, so does not result in an
+immediate change to congestion window or recovery state. An RTO alarm fires only
+when there's a prolonged period of network silence, which could be caused by a
+change in the underlying network RTT.
+
+QUIC also diverges from TCP by including MaxAckDelay in the RTO period.  QUIC is
+able to explicitly model the ack delay via the ack delay field in the ack frame.
```
janaiyengar: s/model the ack delay/model delay at the receiver/
s/ack frame/ACK frame/

PR #995, opened 2017-12-06
Move back to draft 07 style type byte

ianswett:
Un-does most of the changes in PR #956, to give time for more discussion and avoid causing implementers potentially unnecessary work.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -416,10 +416,10 @@ The following packet types are defined:
 
 | Type | Name                          | Section                     |
 |:-----|:------------------------------|:----------------------------|
-| 0x7F | Initial                       | {{packet-initial}}          |
-| 0x7E | Retry                         | {{packet-retry}}            |
-| 0x7D | Handshake                     | {{packet-handshake}}        |
-| 0x7C | 0-RTT Protected               | {{packet-protected}}        |
+| 0x01 | Initial                       | {{packet-initial}}          |
+| 0x02 | Retry                         | {{packet-retry}}            |
+| 0x03 | Handshake                     | {{packet-handshake}}        |
+| 0x04 | 0-RTT Protected               | {{packet-protected}}        |
 {: #long-packet-types title="Long Header Packet Types"}
 
 The header form, packet type, connection ID, packet number and version fields of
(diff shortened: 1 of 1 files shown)


ekr:
@martinthomson @ianswett are you planning to issue a -09 with this change? Should we use that for ID3?

ianswett:
@ekr If this lands, then I believe the answer is yes(along with complete change logs).  But that's subject to @martinthomson and @janaiyengar

janaiyengar:
@ekr: I'd like for @martinthomson to take a look at this PR, and if it lands, then we'll issue a draft-09.

larseggert:
So I did implement the current -08 scheme already... :-(

csperkins:
The discussion hasn't finished, but I thought there was consensus that some form of multiplexing was needed. This change looks like a step backwards.

marten-seemann:
I agree that it's better to first get this right, and then change all the implementations. There's no benefit from changing the same code over and over again.

janaiyengar:
@csperkins There was no formal consensus, but that's not the bar for getting changes into the draft. That said, we're still discussing the form and what we want to mux with, so the discussion is clearly not necessarily going to yield this outcome. In the meanwhile, implementations are required to use these new types for interop, and will likely have to change it again when we figure out the "right" answer. Given this is not the agreed upon answer yet, what does this PR get you?

huitema:
I have implemented the changes in Picoquic. Going back to the previous format would not help me one bit. Besides, the PR is *not* going back to the previous format -- it still does changes like merging server clear text and client clear text in a single code, removing the code point for renegotiation, and shifting all the other long header types by one. So it does not actually save much work for someone going from -07 to -08.

Useless, negative in fact.

marten-seemann:
@huitema: Client and Server Cleartext were merged into a single codepoint a long time before #956.

This PR is very helpful for people who have a deployment of gQUIC, since it's easy to demultiplex the old format with gQUIC, which is a big pain with #956. I don't know about the others maintaining a gQUIC deployments, but I don't see the point of going through the trouble to implement a new demultiplexing scheme for this change, just to change it again once we agree on yet another type byte format.

ianswett:
@huitema At this point, I think there are enough people who have implemented 08 that this should only be landed once we have a plan for greasing the type byte.

@marten-seemann Given we're not yet greasing, I think you can use the trick of looking at the nonce bit on the server side to differentiate GQUIC from IETF QUIC(No client to server packet ever has the nonce bit set to true in GQUIC).  We haven't done this yet in GQUIC, but I don't anticipate this will be a huge change?

ianswett:
Slight tweak to my last comment.  Apparently @janaiyengar asked implementers their preference on the slack channel.  I'm happy to let implementers decide.

janaiyengar:
Based on implementer input on the slack channel, the general sense seems to be that folks have implemented the changes this PR reverts, so we aren't going to merge this PR. Please note that there's active discussion on the question of packet types, which continues on #426. Please share further thoughts there.

MikeBishop:
Given that we've agreed not to do this, I'm closing it.

PR #1004, opened 2017-12-08
Gratuitous key schedule change

### LINKED ISSUE #1003: Update key derivation
TLS moved to shorter labels to reduce the number of times we call SHA-256.  We should do that.

Also, capitalization on "secret" is inconsistent in the example, credit to @mcmanus for finding that.  It's surprising that it took so long for anyone to complain.

martinthomson:
This changes the key schedule for QUIC so that it uses shorter labels.  Rather than a "tls13 " stem, it uses a "QUIC " stem.  The labels are all now shorter and more consistent in case.

Closes #1003.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -678,9 +678,9 @@ completion of the TLS handshake.  Data sent using 0-RTT keys might be replayed
 and so has some restrictions on its use, see {{using-early-data}}.  0-RTT keys
 are used after sending or receiving a ClientHello.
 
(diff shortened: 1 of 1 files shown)


marten-seemann:
Will this change be included in -09? Do we need to do this before or after the remote interim?

martinthomson:
It's definitely not in -08.  I definitely wouldn't want to make this part of anything we do this month.

janaiyengar:
@marten-seemann raises a good point -- let's not merge this until after we decide on #995.

ianswett:
PR #995 is no longer in play, so I think we can merge this.

PR #1005, opened 2017-12-08
Don't create multiple connections from the same ClientHello

martinthomson:
The TLS draft mentioned "Server Stateless Retry" a few times.  Now it doesn't.  That's a minor point.

I have also added a section that deals with the potential for Retry to be sent multiple times with different values.  A stateless version of NSS will generate a different cookie in response to every ticket it receives.  The difference is due to the need to generate a nonce for self-encryption.  That means that the client could end up with many HelloRetryRequest messages to work from.  

A client shouldn't proceed with more than one handshake here or we are well into the territory of violating some of the basic assumptions in the TLS security analysis (that ClientHello.random is a nonce).

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1130,15 +1130,22 @@ the transport will be lost at the server.  This includes the stream offset of
 stream 0, the packet number that the server selects, and any opportunity to
 measure round trip time.
 
-A server MUST send a TLS HelloRetryRequest in a Server Stateless Retry packet.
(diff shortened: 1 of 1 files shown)


mikkelfj:
Is there a way to know that extra HelloRetryRequests originate from the initial Client packet, and not as a result of the client trying to connect again, but still fails? In the first case additional retry packets should be dropped, but in the second case the connection should probably by aborted. Or can this simply be dealt with by the triving eventually timing out on connection attempts?

PR #1012, opened 2017-12-13
A more complete connection migration

### LINKED ISSUE #880: Probing and connection migration
#161 and #732 assume a make after break model, where only one network is available at any given time for an endpoint to migrate to. This is generally too simplistic, and it is common for mobile devices to have access to both WiFi and cellular. Connection migration needs to be more nuanced about probing and using the multiple available networks. Experience with GQUIC and MPTCP shows that having access to multiple networks is almost necessary for useful connection migration.

We should design migration carefully based on known best practices and experience. The current text is necessary but insufficient, and this issue is a placeholder for improving this set of mechanisms in the draft.

janaiyengar:
Closes #880.

This PR describes a more complete connection migration design. Among other small bits, this design includes the following:
- Allow a client to probe alternate paths without having to use them for data.
- Ensure that a server validates client ownership of a new address.


draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+#### Committing to a New Address
+
+When a server receives a packet containing a frame other than PATH_CHALLENGE,
+PATH_RESPONSE, or PADDING from a client address with a packet number that is the
+largest seen thus far on the connection, the server "commits" to this client
+address. The server MUST send all subsequent packets to this address, with the
+following exception.  PATH_CHALLENGE, PATH_RESPONSE, and PADDING frames are used
+for probing a network path, and receiving only these frames MUST NOT cause the
+server to commit to a new address.
```
martinthomson: I think that this can be simpler.  Any packet containing PATH_CHALLENGE is ignored.

You do allow STREAM frames with that packet in earlier text (and STREAM frames aren't a terrible waste of octets, so I can imagine people using the bandwidth for that rather than padding, even if the probability of loss is that much greater).  But this text would cause the recipient to latch.

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+When the server receives a PATH_RESPONSE from a client and verifies that the
+included data matches that sent by the server in a previous PATH_CHALLENGE to
+the client, the server considers the client address to which the corresponding
+PATH_CHALLENGE was sent as valid.
+
+A server MAY send additional PATH_CHALLENGE frames to handle packet loss,
+subject to the congestion controller's limits.
+
+A PATH_CHALLENGE frame used for validation MUST contain at least 12 randomly
+generated octets {{?RFC4086}}, making the payload sufficiently hard to guess.
```
martinthomson: This is a generic requirement, but you have it here in the server-specific section.

draft-ietf-quic-transport.md
```diff
@@ -2358,17 +2419,56 @@ Application Error Code:
...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+~~~
+
+Length:
+
+: This 8-bit value describes the length of the Data field.
+
+Data:
+
+: This variable-length field contains arbitrary data.
```
martinthomson: Can we make this a fixed length?

draft-ietf-quic-transport.md
```diff
@@ -2358,17 +2419,56 @@ Application Error Code:
...
+frame (that is, 0x0e).  If the content of a PATH_RESPONSE frame does not match
+the content of a PATH_CHALLENGE frame previously sent by the endpoint, the
+endpoint MAY generate a connection error of type UNSOLICITED_PATH_RESPONSE.
 
-An endpoint that receives an unsolicited PONG frame - that is, a PONG frame
-containing a payload that is empty MUST generate a connection error of type
-FRAME_ERROR, indicating the PONG frame (that is, 0x10d).  If the content of a
-PONG frame does not match the content of a PING frame previously sent by the
-endpoint, the endpoint MAY generate a connection error of type UNSOLICITED_PONG.
+A PATH_RESPONSE frame MUST NOT elicit acknowledgements.
```
martinthomson: I don't think that this is a necessary requirement.  Given that it's the packet that generates the acknowledgments, that PADDING causes ACKs to be sent, and that this will never fill an entire packet, I think that you can drop this line.

draft-ietf-quic-transport.md
```diff
@@ -3841,7 +3941,8 @@ Issue and pull request numbers are listed with a leading octothorp.
 
 ## Since draft-ietf-quic-transport-08
 
-No significant changes.
+- Added PATH_CHALLENGE and PATH_RESPONSE frames, removed PING with Data, removed
+  PONG frame and rewrote connection migration (#000)
```
erickinnear: This should be filled in with the correct number now

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+allow connections to survive changes to the client's address (client's IP
+address and/or port), such as those caused by a client migrating to a new
+network or a NAT rebinding changing the client's public address.
+
+
+### Path Validation and PMTU Verification
+
+PATH_CHALLENGE ({{frame-path-challenge}}) and PATH_RESPONSE
+({{frame-path-response}}) frames MAY be used for validating that an endpoint
+owns an address, or to verify the PMTU to/from an address.
```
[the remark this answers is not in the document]
janaiyengar: The client is verifying PMTU from its new address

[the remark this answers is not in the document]
martinthomson: I mean, this talks about verification toward the peer (outward bound), not from the peer (inbound).

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+A server validates a client's address by sending a PATH_CHALLENGE frame
+containing a payload that is hard to guess to that address.  The server
+considers the address to be valid when a PATH_RESPONSE frame containing the same
+payload is received from the client.
```
MikeBishop: Does it have to come back from a particular address?  Either way, perhaps say so.

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+Similarly, a client verifies the PMTU of a network path from a new client
+address to the server by sending a PMTU-sized packet carrying a PATH_CHALLENGE
+frame. The client considers the PMTU verified when a PATH_RESPONSE frame
+containing the same payload sent by the server to the new client address is
+received.
+
+
+### Client Behavior {#migration-client}
+
+A client triggers connection migration to a new address by sending a
```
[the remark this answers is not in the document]
MikeBishop: Perhaps "When a client intends to migrate to a new address, it prepares for the migration by...."

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+PATH_CHALLENGE as a probe or by switching to sending all packets to the new
```
MikeBishop: to => from.  If you take the proposed change above, perhaps expand this to talk about unintended/unexpected address changes, in which case the client will just keep sending from whatever address it has to work with.

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+address. The client SHOULD do PMTU verification before considering migration
+complete. The client also participates in an address validation that the server
```
MikeBishop: an => any

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+may initiate to confirm the client's ownership of its address (see
+{{migration-server}}).
+
+A client migrating to a new network might use a new connection ID for packets
+sent from the new address, see {{migration-linkability}} for further discussion.
+
+
+#### Verifying the PMTU
+
+When the client intends to use a new local address, it MAY establish server
```
MikeBishop: Or realizes that it is using a new path?  But how does it do that, based on receipt of a PATH_CHALLENGE from the server?

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+controller's limits {{QUIC-RECOVERY}}.
+
+The client SHOULD use fresh random data in every PATH_CHALLENGE frame so that it
+can perform more accurate measurements by associating the server's response with
+the causative PATH_CHALLENGE.
+
+When the client receives a PATH_RESPONSE from the server and verifies that the
+included data matches the data it sent in a previous PATH_CHALLENGE, the server
+is considered reachable from the new address and the PMTU of the new network
+path is considered verified.
```
MikeBishop: "the PMTU is considered verified" is a little too binary here.  You've mandated "exactly" 1200 bytes, which only verifies that 1200 is less than the PMTU.  The client would actually need to do a series of increasingly-large packets with PATH_CHALLENGEs before it actually knows the PMTU.  It might be worth integrating PATH_CHALLENGE into the existing PMTU section, referencing that, and simply saying that the client SHOULD re-probe PMTU whenever it becomes aware of a path change.

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+PATH_CHALLENGE frame in a PATH_RESPONSE frame.
+
+A client's PATH_RESPONSE MAY be bundled with other frames and does not require
+additional padding.
+
+A client MAY send more than one PATH_RESPONSE in response to a PATH_CHALLENGE,
+as allowed by the congestion controller, to ensure that packet loss does not
+unduly delay or cause failure of the server's validation process.
+
+A client may trigger the server's validation unwittingly when its public address
```
MikeBishop: Even with the non-caps "may" no longer being considered possibly normative, switching to "might" or "could" would probably still be clearer.

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+changes, such as when a NAT rebinds the client's private address to a new public
+address.  Though the client may be oblivious of such an address change, it MUST
```
MikeBishop: "oblivious to", or "unaware of"

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+respond to the consequent validation.  Receiving a PATH_CHALLENGE from the
```
MikeBishop: "Consequent" is a totally cromulent word for this context, but it might be a mite opaque to some readers.

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+was received.
+
+The server sends the PATH_RESPONSE in a packet that is padded to no more than
+the size of the client's packet carrying the PATH_CHALLENGE.  The server SHOULD
+pad the packet to the same size as the client's packet, unless the server has
+reasonable assurance of a smaller PMTU.
+
+The server need not retransmit its PATH_RESPONSE, since it is the client's
+responsibility to send more PATH_CHALLENGE frames when a response is not
+received for a prior one.
```
MikeBishop: Making them not retransmittable risks false measures of PMTU, doesn't it?  It becomes unclear to the other side whether a drop was random packet loss or the result of oversize datagrams.

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+Servers MUST ignore a PATH_CHALLENGE frame from a client that is carried in a
+packet smaller than 1200 octets in size.
+
+
+#### Validating an Unvalidated Client Address {#migration-validate}
```
MikeBishop: Is the "unvalidated" in the title actually useful?  Other than the "don't send other packets" restriction, is anything else about this flow different from revalidating a previously validated client address?

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+Before sending any other packets to a client address that is not yet considered
+valid, the server MUST send a PATH_CHALLENGE frame with its own validation data
+to this address to verify the client's ownership of it. The server's
+PATH_CHALLENGE MAY be bundled with other frames and does not require additional
+padding.
```
MikeBishop: This feels odd.  The server needs to re-probe PMTU as well; why no padding here?  Why can you bundle other frames if you're not supposed to be sending traffic to this unverified client?  You either need more symmetry here, or better explanation of why you're being asymmetric.

draft-ietf-quic-transport.md
MikeBishop: Seconded.  Shouldn't this simply be in the definition of the PATH_CHALLENGE frame?

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+A server MAY send additional PATH_CHALLENGE frames to handle packet loss,
+subject to the congestion controller's limits.
+
+A PATH_CHALLENGE frame used for validation MUST contain at least 12 randomly
+generated octets {{?RFC4086}}, making the payload sufficiently hard to guess.
+
+The server MUST use fresh random data in every PATH_CHALLENGE frame so that it
+can associate the client's response with the causative PATH_CHALLENGE.  Not
+using fresh random data risks exposing the server to an attacker defeating
+validation by sending a spoofed PATH_RESPONSE from a victim's address.
```
MikeBishop: Clarify.  The attacker (hopefully) can't spoof a PATH_RESPONSE, but could capture and replay one from a spoofed source address.

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+Until a client address is deemed valid, the server MUST limit the rate at
+which it sends data to this address. The server SHOULD NOT send more than an
+initial window's worth of bytes (as defined in {{QUIC-RECOVERY}}}) per estimated
+roundtrip period.  In the absence of this limit, the server risks being used for
+a denial of service attack against an unsuspecting victim.
+
+When validation of an address fails, the server SHOULD revert back to using the
+last validated client address, absent which the server MUST close the connection
```
MikeBishop: How can this be absent?  The client must have had a previously valid address, or we wouldn't have a connection.  Now, it might be that the client can no longer receive traffic at that address, in which case the connection will time out anyway.  Again, I'd refer to that section rather than trying to re-specify it here.

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+PATH_RESPONSE, or PADDING from a client address with a packet number that is the
+largest seen thus far on the connection, the server "commits" to this client
+address. The server MUST send all subsequent packets to this address, with the
+following exception.  PATH_CHALLENGE, PATH_RESPONSE, and PADDING frames are used
+for probing a network path, and receiving only these frames MUST NOT cause the
+server to commit to a new address.
+
+Receiving a packet from a new address but with a packet number that is not the
+largest seen thus far suggests that a reordered packet was received and MUST NOT
+cause the server to change its commitment.
```
MikeBishop: This is going to interact oddly with the NEW_CONNECTION_ID and packet number gap stuff.

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
-to receive packets from an older source address.
 
-An endpoint MUST validate that its peer can receive packets at the new address
-before sending any significant quantity of data to that address, or it risks
-being used for denial of service.  See {{migrate-validate}} for details.
+### Loss Detection and Congestion Control
+
+There may be apparent reordering at the receiver when an endpoint sends data and
+probes from/to multiple addresses during the migration period, since the two
+resulting paths may have different rountrip times.  A receiver of packets on
```
MikeBishop: round-trip

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+multiple paths will still send ACK frames covering all received packets.
+
+Both endpoints use one congestion control and one loss recovery context, as
```
MikeBishop: "Each endpoint uses a single congestion control and loss recovery context...."

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+described in {{QUIC-RECOVERY}}.  A sender MAY make an exception for probe
+packets so that their loss detection is independent and does not unduly cause
```
MikeBishop: Clarify "probe packets."  PMTU probing?  Address validation?  Both?

draft-ietf-quic-transport.md
```diff
@@ -1714,9 +1798,9 @@ state to draining, but it otherwise has no impact.
 An endpoint could receive packets from a new source address, indicating a
 connection migration ({{migration}}), while in the closing period. An endpoint
 in the closing state MUST strictly limit the number of packets it sends to this
-new address as though the address were not validated (see {{migrate-validate}}).
-A server in the closing state MAY instead choose to discard packets received
-from a new source address.
+new address as though the address were not validated (see
```
MikeBishop: I know this isn't new text, but why "as though the address were" rather than "since the address is not"?

draft-ietf-quic-transport.md
```diff
@@ -2358,17 +2419,56 @@ Application Error Code:
   {{app-error-codes}}).
 
 
+## PATH_CHALLENGE Frame {#frame-path-challenge}
+
+Endpoints can use PATH_CHALLENGE frames (type=0x0e) to check reachability to the
+peer, to verify a new path's PMTU, and to perform address validation during
+connection migration.
```
MikeBishop: Is PATH_CHALLENGE useful for PMTU validation once the address is known to be good?  Why couldn't you simply rely on ACKs to tell you whether the packet got through or not?

draft-ietf-quic-transport.md
MikeBishop: We already have a minimum.  Is there a scenario in which a server would want it to be longer?

draft-ietf-quic-transport.md
```diff
@@ -2358,17 +2419,56 @@ Application Error Code:
...
+
+: This 8-bit value describes the length of the Data field.
+
+Data:
+
+: This variable-length field contains arbitrary data.
+
+
+The sender of this frame MUST include at least one octet of data in the Data
+field.
```
MikeBishop: A larger minimum is provided above (eight octets, I think?).

draft-ietf-quic-transport.md
```diff
@@ -2358,17 +2419,56 @@ Application Error Code:
...
+
+The recipient of this frame MUST generate a PATH_RESPONSE frame
+({{frame-path-response}}) containing the same Data.  An endpoint that receives a
+PATH_CHALLENGE frame containing an empty payload MUST generate a connection
+error of type FRAME_ERROR, indicating the PATH_CHALLENGE frame (that is, 0x0e).
+
+A PATH_CHALLENGE frame MUST NOT elicit acknowledgements; the corresponding
```
MikeBishop: Frames don't get acknowledged; packets do.  Does this mean that a packet containing a PATH_CHALLENGE MUST NOT be acknowledged?  That's going to play merry hob with the loss recovery if you start putting actual data in there (padding with STREAM).

draft-ietf-quic-transport.md
mikkelfj: Is there sufficient text for guidance on how to handle a path that reverts to smaller PMTU during normal operation? I imagine a large stream frame in eternal futile retransmission because it isn't split into smaller frames in smaller packets.

huitema:
I have mixed feelings about this. First, I am not sure that this belongs to the frame layer. The path challenge is in many ways similar to a client initial packet. It requires the same kind of padding, for the same reason. And it carries similar content restrictions such as "no other frame allowed". A new packet type might be more appropriate than a frame type.

Second, this probably requires a three ways handshake, not just a question response. And it would be better served with a symmetric structure -- 
Maybe "your nonce, my nonce", so as to allow symmetry.

Then, there maybe a need for more attributes than nonces, especially if we want to handle NAT rebinding. Like, echoing the address from where this was sent.

huitema:
I guess my uneasiness stems from the PR's scope - supporting both involuntary transition like NAT remapping and voluntary transition like make before break. The latter addresses maybe 80% of the multi path issues, and I don't like addressing that without completing the multi path work. Sine full multipath is indeed out of scope for V1, how about simplifying drastically and *only addressing involuntary NAT rebinding*?

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+When a server receives a PATH_CHALLENGE frame from a client, it MUST generate a
+PATH_RESPONSE frame that contains the same data as the challenge. The server's
+PATH_RESPONSE MUST be sent to the address from which the client's PATH_CHALLENGE
+was received.
+
+The server sends the PATH_RESPONSE in a packet that is padded to no more than
+the size of the client's packet carrying the PATH_CHALLENGE.  The server SHOULD
+pad the packet to the same size as the client's packet, unless the server has
+reasonable assurance of a smaller PMTU.
```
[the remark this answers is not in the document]
huitema: Security consideration: if we allow PATH RESPONSE packets to be larger than the PATH CHALLENGE packet that triggered them, we enable a reflection DOD attack.

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+The server need not retransmit its PATH_RESPONSE, since it is the client's
+responsibility to send more PATH_CHALLENGE frames when a response is not
+received for a prior one.
+
+Servers MUST ignore a PATH_CHALLENGE frame from a client that is carried in a
+packet smaller than 1200 octets in size.
```
huitema: Yes. Security consideration as to why?

draft-ietf-quic-transport.md
huitema: Basically, starting a new path should reset the congestion control algorithm to ts starting condition. Just say so.

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+Both endpoints use one congestion control and one loss recovery context, as
+described in {{QUIC-RECOVERY}}.  A sender MAY make an exception for probe
+packets so that their loss detection is independent and does not unduly cause
+the congestion controller to reduce its sending rate.
+
+For instance, an endpoint may run a separate alarm when a PATH_CHALLENGE is
+sent, which is disarmed when the corresponding PATH_RESPONSE is received. If the
+alarm fires before the PATH_RESPONSE is received, the endpoint might send a new
+PATH_CHALLENGE, and restart the alarm, but for a longer period of time.
```
huitema: I see the discussion of the congestion control window, but I do not see an equivalent discussion of flow control and the MAX_DATA frame. Yet if the client migrates to a slow connection, MAX_DATA is going to be way too large. Where is that discussed?

draft-ietf-quic-transport.md
```diff
@@ -1541,112 +1731,6 @@ number. "packet_number_secret" is derived from the TLS key exchange,
 as described in Section 5.6 of {{QUIC-TLS}}.
```
huitema: The packet-number-gap trick only works if both the client and the server do it, and if both refrain from sending any packet on the old connection in parallel with the new connection. That seems very fragile. We should really consider encrypting the packet sequence number.

Note that all this discussion only applies to voluntary migration. It does not apply to NAT rebinding. If the client cannot anticipate the NAT rebinding, it will not be able to switch to a new connection ID or to insert a gap in the sequence numbers.

draft-ietf-quic-transport.md
huitema: I have a feeling that this text about the composition of challenge response is repeated multiple times. Can we simplify?

draft-ietf-quic-transport.md
```diff
@@ -2358,17 +2419,56 @@ Application Error Code:
...
+frame.  Its format is identical to the PATH_CHALLENGE frame
+({{frame-path-challenge}}).
 
-The PONG frame (type=0x0d) is sent in response to a PING frame that contains
-data.  Its format is identical to the PING frame ({{frame-ping}}).
+An endpoint that receives a PATH_RESPONSE frame containing an empty payload MUST
+generate a connection error of type FRAME_ERROR, indicating the PATH_RESPONSE
+frame (that is, 0x0e).  If the content of a PATH_RESPONSE frame does not match
+the content of a PATH_CHALLENGE frame previously sent by the endpoint, the
+endpoint MAY generate a connection error of type UNSOLICITED_PATH_RESPONSE.
```
huitema: Did we say before that path challenges do not need to be acked? Presumably, some servers will want to repeat them. Should the repeated challenge have exactly the same content as the original challenge?

draft-ietf-quic-transport.md
huitema: See issue #1034. The packet number gap is a complex mechanism that does not actually protect privacy, because client and server apply the same gap.

draft-ietf-quic-transport.md
```diff
@@ -1465,35 +1470,220 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+following exception.  PATH_CHALLENGE, PATH_RESPONSE, and PADDING frames are used
+for probing a network path, and receiving only these frames MUST NOT cause the
+server to commit to a new address.
+
+Receiving a packet from a new address but with a packet number that is not the
+largest seen thus far suggests that a reordered packet was received and MUST NOT
+cause the server to change its commitment.
+
+Upon committing to a new address, the server SHOULD reset its congestion
+controller and roundtrip time estimator to start new estimations.
```
ihlar: At this point it is still possible that a substantial amount of data is outstanding on the "previous" path, this data will be acked on the new path. I suppose that resetting the congestion controller and roundtrip time estimator means that such acknowledgements should be ignored when calculating the congestion window and the SRTT? 
This should probably be explicitly stated somewhere.

martinthomson: Excellent point.  "Only packets sent on the new path contribute to the state of the congestion control and RTT estimation on that path."

draft-ietf-quic-transport.md
janaiyengar: Good point, fixed.

draft-ietf-quic-transport.md
```diff
@@ -702,6 +702,10 @@ use it for all subsequent Handshake ({{packet-handshake}}) and 1-RTT
 Server's Version Negotiation ({{packet-version}}) and Retry ({{packet-retry}})
 packets MUST use connection ID selected by the client.
 
+Note that the connection ID may change during the connection via the use of
+NEW_CONNECTION_ID frames ({{frame-new-connection-id}}) and connection migration
+({{migration}}).
+
```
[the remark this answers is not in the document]
janaiyengar: This is discussed further below, in the linkability discussion. There's no normative language in this section, this is only a statement of fact.

janaiyengar:
Based on the comments on the PR and from the interim, I've done a rewrite of this patch. I am not changing packet number behavior yet, but I expect that's going to have to come out of the packet number encryption conversation. I am also limiting the migration protocol to client migration since making this symmetric adds sharp edges that need to be more carefully dealt with. I'm restricting the scope of this PR to the client/server use case that's most relevant for v1.

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,236 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+validation of the address of the spurious migration to be abandoned.
+
+
+### Loss Detection and Congestion Control
+
+The capacity available on the new path might not be the same as the old path.
+
+A client SHOULD reset its congestion controller and roundtrip time estimator
+prior to sending any non-probing packets from a new local address.
+
```
ihlar: As discussed earlier, I think it would make sense to add a text like:
 "Only packets sent on the new path contribute to the state of the congestion control and RTT estimation on that path."

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,236 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+Thus, the endpoint considers the path to be valid when a PATH_RESPONSE frame is
+received on the same path with the same payload as the PATH_CHALLENGE frame.
+
+The endpoint MAY send additional PATH_CHALLENGE frames to handle packet loss,
+but is subject to the following limit to avoid excessive network load: an
+endpoint SHOULD NOT send more than one PATH_CHALLENGE frame in 200ms.  This
+limit is approximately equivalent to the network load due to a new connection,
+and is based on the initial Handshake Timeout defined in {{QUIC-RECOVERY}}.
+
```
mikkelfj: 200ms is probably a sensible value, but for the life-time of the protocol it is hard to say what the network load of a new connection would be, not to mention that it could be totally different depening on the environment.

A different wording could be that is MUST be rate-limited to the estimated connection establishment time, and otherwise 200ms can be used as a default.

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,236 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+### Initiating Connection Migration {#initiating-migration}
+
+A client MAY initiate connection migration to a new local address in one of two
+ways.
+
+The client may immediately migrate the connection by sending all packets from
+the new local address.  Receiving acknowledgments for this data serves as proof
+of the server's reachability from the new address.
+
```
mikkelfj: Why is PATH_RESPONSE more strict than ACK's from just sending data wrt. local and remote addresses?

Should there be a rate limit on how often a migration can take place? It might stress the peer, and gaps in packet numbers can be a DoS on the peers implementation (I know gaps are up for discussion, but if/when there are gaps...)

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,236 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+address first using path validation {{migrate-validate}} and migrate the
+connection to the new address at a later time.  Failure of path validation
+simply means that the new local address is not usable for this connection and
+should not be fatal to the connection when alternative local addresses are
+available.
+
+A client migrating to a new local address SHOULD use a new connection ID for
+packets sent from that address, see {{migration-linkability}} for further
+discussion.
+
```
mikkelfj: Is SHOULD sufficient? One endpoint might not care, but it the other endpoint might.

draft-ietf-quic-transport.md
ianswett: I think I like Marcus suggestion here.  The text seems to presuppose an implementation approach(a single congestion controller that's reset), opposed to specifying the correct behavior.

janaiyengar: Modified this text.

draft-ietf-quic-transport.md
janaiyengar: I think this is adequate for now. The section on migration linkability has the real normative text anyways.

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,236 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+sends significantly more data than the client, connection migration might be
+used to amplify the volume of data that an attacker can generate toward a
+victim.
+
+Thus, on receiving and decrypting a packet from an unvalidated address, the
+server MUST immediately validate the client's new address to confirm the
+client's possession of the new address.
+
+Until a client's address is deemed valid, the server MUST limit the rate at
+which it sends data to this address.  The server MUST NOT send more than 4 * MSS
```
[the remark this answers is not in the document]
janaiyengar: It's a bit arbitrary, but I was shooting for something > min cwnd. Honestly, any number SGTM. Do you prefer min cwnd? 

I assumed the initial RTO period implied a default if path RTT was unknown. Happy to clarify if you think it's not clear.

draft-ietf-quic-transport.md
```diff
@@ -1487,8 +1487,8 @@ network.  This section describes the protocol for migrating a connection to a
 new client address.
 
 Migrating a connection to a new server address is left for future work. If a
-client receives packets from a new server address that are decryptable, the
-client MAY discard these packets.
+client receives packets in the connection from a new server address, the client
```
ianswett: nit: I don't think "in the connection" is necessary, since you begin the sentence with "Migrating a connection..."

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
 
 ## Connection Migration {#migration}
```
ianswett: If this is client specific, should it be called Client Connection Migration, at least for the time being?

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
-connection when they move networks.  This includes state that can be hard to
-recover such as outstanding requests, which might otherwise be lost with no easy
-way to retry them.
+QUIC allows connections to survive changes to endpoint addresses (that is, IP
+address and/or port), such as those caused by a client migrating to a new
+network.  This section describes the protocol for migrating a connection to a
+new client address.
+
+Migrating a connection to a new server address is left for future work. If a
+client receives packets in the connection from a new server address, the client
```
ianswett: nit: I don't think "in the connection" is necessary, since you begin the sentence with "Migrating a connection..."

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,236 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+sent, path validation is considered to have failed, even if the data matches
+that sent in the PATH_CHALLENGE.
+
+Thus, the endpoint considers the path to be valid when a PATH_RESPONSE frame is
+received on the same path with the same payload as the PATH_CHALLENGE frame.
+
+The endpoint MAY send additional PATH_CHALLENGE frames to handle packet loss,
+but is subject to the following limit to avoid excessive network load: an
+endpoint SHOULD NOT send more than one PATH_CHALLENGE frame in 200ms.  This
+limit is approximately equivalent to the network load due to a new connection,
```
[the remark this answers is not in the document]
ianswett: I agree that we should avoid being specific, which is one of the reasons I'm not a fan of saying 200ms here, and would prefer text such as: "an endpoint SHOULD NOT send a PATH_CHALLENGE more frequently than it would a client INITIAL, ensuring that connection migration is no more costly to a new path than establishing a new connection."

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
ianswett: I see why you chose 4*MSS.  In this document, 2*Min CWND seems preferable to a specific value to me.

I think it's worth clarifying that the RTO is likely based on default RTT/default RTO in this case, since if you haven't received a PATH_RESPONSE, I would expect you either have no estimate or a somewhat unverified estimate of RTT.

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+
+### Initiating Connection Migration {#initiating-migration}
+
+A client MAY initiate connection migration to a new local address in one of two
+ways.
+
+The client may immediately migrate the connection by sending all packets from
+the new local address.  Receiving acknowledgments for this data serves as proof
+of the server's reachability from the new address.  Note that since
```
martinthomson: No it doesn't.  A server could lie too.

If we want to do this right, we should try to always verify paths before using them. Obviously, when an endpoint migrates without its knowledge, that's not going to happen, so we would need to make allowances for that, but this says that it's OK to just move to a new address and start sending.

This is a big departure from existing systems.  ICE always verifies.  TCP always verifies.

I think that the important distinction is choice.  A client that is moved (because the network moves or rebinds) can keep sending because it's no different to route change in the network.  The client didn't move itself, it was moved.  But if the client decides to use a different network interface (wifi vs. cellular, for example), then we're into the MUST-validate territory.  I think that you need to move that distinction up into the introductory material.

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+acknowledgments may be received on any path, return reachability on the new path
+is not established. To establish return reachability on the new path, a client
+MAY concurrently initiate path validation {{migrate-validate}} on the new path.
+
+Alternatively, the client may probe for server reachability from the new local
```
martinthomson: s/may/could/

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+address first using path validation {{migrate-validate}} and migrate the
+connection to the new address at a later time.  Failure of path validation
```
martinthomson: s/at a later time/later/

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+simply means that the new local address is not usable for this connection and
+should not be fatal to the connection when alternative local addresses are
```
martinthomson: s/connection and should not be fatal [...]/connection.  Failure to validate a path does not cause the connection to end unless there are no valid alternative paths available./

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+available.
+
+A client migrating to a new local address SHOULD use a new connection ID for
+packets sent from that address, see {{migration-linkability}} for further
+discussion.
+
+A client MUST NOT initiate connection migration before the handshake is
+finished and the client has 1-RTT keys.
```
martinthomson: Another case where the distinction between voluntary and involuntary migration is important.

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+
+### Responding to Connection Migration
+
+A server may receive a packet from a new client address at any time during the
+connection after the handshake is complete. If the packet is decryptable, the
+client may be either probing from a new address or migrating immediately, as
```
martinthomson: s/decryptable/authenticated/
s/may/could/ or s/may/might/

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+In response to such a packet, the server MUST start sending subsequent packets
+to this client address and MUST initiate path validation to verify client
+ownership of the unvalidated address.
 
-Due to variations in path latency or packet reordering, packets from different
-source addresses might be reordered.  The packet with the highest packet number
-MUST be used to determine which path to use.  Endpoints also need to be prepared
-to receive packets from an older source address.
+A server MAY skip validation of a client address that it considers to be already
+valid.
```
martinthomson: Wow, this invites all sorts of speculation.  I think that you need more explanation here.

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
 
-An endpoint MUST validate that its peer can receive packets at the new address
-before sending any significant quantity of data to that address, or it risks
-being used for denial of service.  See {{migrate-validate}} for details.
+Note that the server MAY send data to an unvalidated client address, but it MUST
+protect against potential attacks as described in {{address-spoofing}} and
+{{on-path-spoofing}}.
+
+If a non-probing packet is received from a new client address but with a packet
+number that is not the largest seen thus far, the server must process the packet
```
martinthomson: s/must/can/

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+as usual, but it MUST ignore this client address under the assumption that it is
+not the client's most recent address.
```
martinthomson: Some explanation as to why here (the client might have migrated twice), would be good.

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+After verifying a new client address, the server SHOULD update any address
+validation tokens ({{address-validation}}) that it has issued to the client if
+those are no longer valid based on the changed address.
```
martinthomson: What does update mean?  Does that mean sending them again?

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
martinthomson: Everything that Ian said, except that I don't understand why 2xMinCWND.  Seems equally arbitrary.  MinCWND on the assumption of a default RT[TO] would be more defensible.

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+path is the same as the previous path.  For instance, a change in the client's
+port number is likely indicative of a rebinding in a middlebox and not a
+complete change in path.  This determination likely depends on heuristics, which
+could be imperfect; if the new path capacity is significantly reduced,
+ultimately this relies on the congestion controller responding to congestion
+signals and reducing send rates appropriately.
+
+There may be apparent reordering at the receiver when an endpoint sends data and
+probes from/to multiple addresses during the migration period, since the two
+resulting paths may have different rountrip times.  A receiver of packets on
```
martinthomson: roudtrip

draft-ietf-quic-transport.md
```diff
@@ -1735,9 +1829,9 @@ state to draining, but it otherwise has no impact.
 An endpoint could receive packets from a new source address, indicating a
 connection migration ({{migration}}), while in the closing period. An endpoint
 in the closing state MUST strictly limit the number of packets it sends to this
-new address as though the address were not validated (see {{migrate-validate}}).
-A server in the closing state MAY instead choose to discard packets received
-from a new source address.
+new address since the address is not validated (see {{migrate-validate}}). A
```
martinthomson: s/since/until
s/not //

draft-ietf-quic-transport.md
MikeBishop: Not really -- we've already said that the tuple can't change during the handshake, so involuntary can be treated the same way as a violation of this requirement by the server.

draft-ietf-quic-transport.md
MikeBishop: I understand the desire to leave implementations room to explore the envelope, but I agree this is a bit vague.  I'm assuming this is mostly intended to be read as "same IP, different port" or "return to an IP address the client had used/validated previously" (e.g. cellular, migrated to Wi-Fi, lost Wi-Fi).

draft-ietf-quic-transport.md
MikeBishop: Sending new ones that validate the new address.

janaiyengar: updated text

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
janaiyengar: added text and set the rate to min cwnd per rto.

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,236 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+address does not have the necessary cryptographic keys to read or respond to the
+PATH_CHALLENGE frame that is sent to it, even if it wanted to.
+
+To protect the connection from failing due to such a spurious migration, the
+server MUST revert to using the last validated client address when validation of
+a new client address fails. If the server has no state about the last validated
+client address, it MUST close the connection silently and discard any further
+packets received from the client for this connection.
+
+We note that receipt of packets with higher packet numbers from the legitimate
```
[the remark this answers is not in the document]
janaiyengar: done

draft-ietf-quic-transport.md
janaiyengar: the 'd' is redundant anyways, nobody pronounces it :-)
... though fixing english is not in charter, so fixing text

draft-ietf-quic-transport.md
janaiyengar: done

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,240 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+The new address is not considered valid until a PATH_RESPONSE frame containing
+the same payload is received, even if the packet containing the PATH_CHALLENGE
+frame is acknowledged.
+
+The PATH_RESPONSE frame MUST be received from the same remote address to which
+the corresponding PATH_CHALLENGE was sent. If a PATH_RESPONSE frame is received
+from a different remote address than the one to which the PATH_CHALLENGE was
+sent, path validation is considered to have failed, even if the data matches
+that sent in the PATH_CHALLENGE.
```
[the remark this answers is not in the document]
MikeBishop: Worth noting that one-way reachability *might* be a thing once we get to multipath.  If client->server has two valid paths, but server->client only one, so be it.  But I don't object to saying path validation has to test both directions.

draft-ietf-quic-transport.md
MikeBishop: ...except that it isn't.  The title of the section is still "Connection Migration."

draft-ietf-quic-transport.md
MikeBishop: I'm in favor of Ian's change, because if packets don't map to a connection they'll be discarded anyway.  The extra words just sound awkward.  However, it's not technically incorrect, so this is purely wordsmithing.

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,270 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+#### Initiation
+
+To initiate path validation, an endpoint sends a PATH_CHALLENGE frame containing
+a payload that is hard to guess on the path to be validated.
+
+An endpoint MAY send additional PATH_CHALLENGE frames to handle packet loss,
+but is subject to the following limit to avoid excessive network load: an
+endpoint SHOULD NOT send a PATH_CHALLENGE more frequently than it would a client
+INITIAL, ensuring that connection migration is no more load on a new path than
```
MikeBishop: Packet names are title-cased, not all-caps.  Also, I'd drop the colon lead-up.

`...but to avoid excessive network load, an endpoint SHOULD NOT... an Initial packet.  This ensures that....`

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,270 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+
+The endpoint MUST use fresh random data in every PATH_CHALLENGE frame so that it
+can associate the peer's response with the causative PATH_CHALLENGE.
+
+
+#### Response
+
+On receiving a PATH_CHALLENGE frame, an endpoint MUST respond immediately by
+echoing the data contained in the PATH_CHALLENGE frame in a PATH_RESPONSE frame,
+with the following stipulation.  Since a PATH_CHALLENGE may be sent from a
```
MikeBishop: "might be" - there's no permission involved here.

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,270 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+spoofed address, an endpoint MAY limit the rate at which it sends PATH_RESPONSE
+frames and MAY silently discard PATH_CHALLENGE frames that would cause it to
+respond at a higher rate.
+
+Path validation tests both directions of a path.  To ensure reachability in both
+directions, the PATH_RESPONSE MUST be sent on the same path as the triggering
+PATH_CHALLENGE: from the same local address on which the PATH_CHALLENGE was
+received, to the same remote address from which the PATH_CHALLENGE was received.
```
MikeBishop: and port, I presume?

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,270 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+establishing a new connection.
+
+The endpoint may receive packets containing other frames during this period, but
+a PATH_RESPONSE frame with appropriate data is required for the path validation
+to succeed.
+
+
+### Initiating Connection Migration {#initiating-migration}
+
+A client MAY initiate connection migration to a new local address in one of two
```
MikeBishop: This "MAY" seems superfluous, particularly since the first way also has a MAY.  I'd either lower-case it, "can", or "A client initiates...."  Also, I don't know that I'd frame these as two "ways" of migrating.  Ultimately, the client migrates by sending data from the new address/port.  A client MAY probe for reachability first, if it has the opportunity.

draft-ietf-quic-transport.md
```diff
@@ -1477,35 +1481,270 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
+the legitimate copy will be dropped as a duplicate. After a spurious migration,
+validation of the source address will fail because the entity at the source
+address does not have the necessary cryptographic keys to read or respond to the
+PATH_CHALLENGE frame that is sent to it, even if it wanted to.
+
+To protect the connection from failing due to such a spurious migration, the
+server MUST revert to using the last validated client address when validation of
+a new client address fails. If the server has no state about the last validated
+client address, it MUST close the connection silently and discard any further
+packets received from the client for this connection.
```
MikeBishop: How does this tie in to the closure/draining states described elsewhere?  Does the server enter draining?  Can it send Stateless Reset?

draft-ietf-quic-transport.md
janaiyengar: Ok, removed.

draft-ietf-quic-transport.md
janaiyengar: Fixed.

draft-ietf-quic-transport.md
janaiyengar: Ah, right. I'm always confused between may and might. This is a good way to think about it. Done.

draft-ietf-quic-transport.md
janaiyengar: an address is defined above as IP and/or port

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
janaiyengar: Yes, that would be quite a different deal. You'd want to be very careful when thinking about reachability in the multipath case, but as you say, that's for a different day.

draft-ietf-quic-transport.md
janaiyengar: Good suggestions. I've made restructured a bit so that it flows as you suggest instead of "two ways".

draft-ietf-quic-transport.md
janaiyengar: Good q. I've changed it so that the server basically discards all connection state, leaving subsequent packets to be treated generically. That is, the server could send Stateless Resets in response. I think this is reasonable.

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,271 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+
+To initiate path validation, an endpoint sends a PATH_CHALLENGE frame containing
+a payload that is hard to guess on the path to be validated.
+
+An endpoint MAY send additional PATH_CHALLENGE frames to handle packet loss.  An
+endpoint SHOULD NOT send a PATH_CHALLENGE more frequently than it would an
+Initial packet, ensuring that connection migration is no more load on a new path
+than establishing a new connection.
+
+The endpoint MUST use fresh random data in every PATH_CHALLENGE frame so that it
```
erickinnear: Here we indicate that this should be random data, but above we just said "hard to guess" -- I know we took out the note about what's considered hard to guess, but perhaps we should still include a note in the above paragraph that it's random?

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
erickinnear: I'd agree that requiring bi-directional reachability in this version is a good step, we can always remove that when defining true multipath.

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,271 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+
+#### Abandonment
+
+The endpoint SHOULD abandon path validation after sending some number of
+PATH_CHALLENGE frames or after some time has passed.  When setting this timer,
+implementations are cautioned that the new path could have a longer round-trip
+time than the original.  Again, to avoid excessive network load, an endpoint
+SHOULD NOT send more PATH_CHALLENGE frames than it would a client INITIAL,
+ensuring that connection migration is no more load on a new path than
+establishing a new connection.
```
erickinnear: Good to have this reminder so everything is in one place when implementing

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,271 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+
+The endpoint may receive packets containing other frames during this period, but
+a PATH_RESPONSE frame with appropriate data is required for the path validation
+to succeed.
```
erickinnear: Does this mean packets on the new path or on the old path? What are we trying to indicate here?

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,271 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+from the new local address.  Since the server's address is validated during
+connection establishment, receiving acknowledgments for this data serves as
+proof of the server's reachability from the new address.  Note that since
+acknowledgments may be received on any path, return reachability on the new path
+is not established. To establish return reachability on the new path, a client
+MAY concurrently initiate path validation {{migrate-validate}} on the new path.
+
+If it has the opportunity, a client MAY probe for server reachability from a
+new local address using path validation {{migrate-validate}} prior to migrating
+the connection to it.  Failure of path validation simply means that the new
```
erickinnear: nit: `to it` -> `to the new local address`

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,271 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+* the packet is a "non-probing" packet. That is, it contains a frame other than
+PATH_CHALLENGE, PATH_RESPONSE, or PADDING; and
+
+* the packet carries a packet number that is the largest seen thus far.
+
+Upon receiving such a packet, the server MUST abandon any path validation it is
+performing with other addresses on the expectation that those validations are
+likely to fail.  Abandoning path validation primarily means ceasing subsequent
+transmissions of the PATH_CHALLENGE frame.  An endpoint MUST ignore any
+subsequently received PATH_RESPONSE frames from abandoned addresses.
```
erickinnear: This could mean the client ends up spending some additional time if it wants to re-establish reachability on the other addresses / maintain that ahead of time (like is best for performance). Probably okay, but worth considering.

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,271 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+
+It is possible that the client is spoofing its source address to cause the
+server to send excessive amounts of data to an unwilling host.  If the server
+sends significantly more data than the client, connection migration might be
+used to amplify the volume of data that an attacker can generate toward a
+victim.
+
+Thus, on receiving and decrypting a packet from an unvalidated address, the
+server MUST immediately validate the client's new address to confirm the
+client's possession of the new address.
```
erickinnear: I worry that people may be confused by the perceived conflict between this and: 
```
A server MAY skip validation of a client address if it has been seen recently or
if the server has reasonable certainty that the new address is the result of a
NAT rebinding.
```

Can we resolve that somehow? 

Perhaps we could say above:
```
A server MAY consider validation to be complete for a client address that has been 
seen recently or if the server has reasonable certainty that the new address is the 
result of a NAT rebinding.
```
Or something to that effect (maybe `A server MAY deem a client address to be valid` to echo the next paragraph here).

draft-ietf-quic-transport.md
erickinnear: Ah, I see below that this allows us to avoid issues where someone tries to migrate us and we grab it back. Looks fine in that case.

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,271 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+Packets sent on the old path SHOULD NOT contribute to the state of the
+congestion control and RTT estimation on the new path. If it is using a single
+congestion controller and round-trip time estimator, the client SHOULD reset
+them prior to sending any non-probing packets from a new local address.
+
+After successful validation of the client's new address, the server SHOULD reset
+its congestion controller and round-trip time estimator prior to sending any
+further non-probing packets.
+
+An endpoint MUST NOT restore its send rate unless it is reasonably sure that the
```
erickinnear: Nit: `restore` -> `return to` ?

draft-ietf-quic-transport.md
erickinnear: I agree with Christian, this seems fragile, but that's for another issue and discussion

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,272 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+with the same payload as the PATH_CHALLENGE frame.
+
+
+#### Abandonment
+
+An endpoint SHOULD abandon path validation after sending some number of
+PATH_CHALLENGE frames or after some time has passed.  When setting this timer,
+implementations are cautioned that the new path could have a longer round-trip
+time than the original.  Again, to avoid excessive network load, an endpoint
+SHOULD NOT send more PATH_CHALLENGE frames than it would a client INITIAL,
```
MikeBishop: Same issue with "Initial" as at ~1519.  BTW, double check that you actually need near-identical text both places?

draft-ietf-quic-transport.md
```diff
@@ -3599,7 +3629,7 @@ entire round trip.
 For smooth operation of the congestion controller, it is generally considered
 best to not let the sender go into quiescence if avoidable.  To avoid blocking a
 sender, and to reasonably account for the possibiity of loss, a receiver should
-send a MAX_DATA or MAX_STREAM_DATA frame at least two roundtrips before it
+send a MAX_DATA or MAX_STREAM_DATA frame at least two round trips before it
```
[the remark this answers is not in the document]
janaiyengar: It's "round-trip time" or "round trip" elsewhere (I think, I'll make it consistent separately)

huitema:
I am looking at a stateless implementation of challenge/response. Not hard, but the 8 bytes challenge/response is a bit short.

draft-ietf-quic-transport.md
janaiyengar: Good point -- I don't think I need it here. The earlier text is adequate. Removed.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
martinthomson: I'd prefer:

> The connection ID could change over the lifetime of a connection, especially in response to connection migration (cite).  NEW_CONNECTION_ID frames (cite) are used to provide new values.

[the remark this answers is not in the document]
janaiyengar: done.

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,269 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+
+As described in {{migration-response}}, a server is required to validate the
+client's new address to confirm the client's possession of the new address.
+Until a client's address is deemed valid, the server MUST limit the rate at
+which it sends data to this address.  The server MUST NOT send more than a
+minimum congestion window's worth of data per estimated RTO period (as defined
+in {{QUIC-RECOVERY}}).  In the absence of this limit, the server risks being
+used for a denial of service attack against an unsuspecting victim.  Note that
+since the server will not have any round-trip time estimates to the new address,
+the RTO period is likely to be estimated based on the default initial RTT (see
```
martinthomson: "SHOULD be" not "likely to be"

I think that this text belongs in the section that deals with handling of migrations by the non-migrating peer.

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,269 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+#### Abandonment
+
+An endpoint SHOULD abandon path validation after sending some number of
+PATH_CHALLENGE frames or after some time has passed.  When setting this timer,
+implementations are cautioned that the new path could have a longer round-trip
+time than the original.
+
+Note that the endpoint might receive packets containing other frames on the new
+path, but a PATH_RESPONSE frame with appropriate data is required for path
+validation to succeed.
```
martinthomson: You might also hint here that a path challenge might be abandoned for other reasons.  Primarily, this happens if a path challenge is initiated as a result of a path migration and another path migration occurs before it completes.

You also need to talk about failure.  If this timer pops, then the path is invalid and cannot be used further.

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,269 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+address and/or port), such as those caused by a client migrating to a new
+network.  This section describes the protocol for migrating a connection to a
+new client address.
+
+Migrating a connection to a new server address is left for future work. If a
+client receives packets from a new server address, the client MAY discard these
+packets.
+
+
+### Path Validation {#migrate-validate}
```
martinthomson: I think that you should promote this section one level and put it before connection migration generally.

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,269 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+
+Path validation MAY be used at any time by either endpoint.  For instance, an
+endpoint might check that a peer is still in possession of its address after a
+period of quiescence.
+
+An endpoint MAY bundle PATH_CHALLENGE and PATH_RESPONSE frames with other
+frames, as appropriate.  For instance, an endpoint may pad a packet carrying a
+PATH_CHALLENGE for PMTU discovery, or an endpoint may bundle a PATH_RESPONSE
+with its own PATH_CHALLENGE.
+
```
martinthomson: Please add a paragraph disclaiming any suitability for UDP NAT traversal.  Though these mechanisms might be effective for the creation of NAT bindings that support NAT traversal, the expectation is that one or other peer (typically the server) is able to receive packets without first having sent a packet on that path.  Effective NAT traversal needs additional synchronization mechanisms that are not provided here.

janaiyengar: Added, PTAL.

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,269 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+#### Response
+
+On receiving a PATH_CHALLENGE frame, an endpoint MUST respond immediately by
+echoing the data contained in the PATH_CHALLENGE frame in a PATH_RESPONSE frame,
+with the following stipulation.  Since a PATH_CHALLENGE might be sent from a
+spoofed address, an endpoint MAY limit the rate at which it sends PATH_RESPONSE
+frames and MAY silently discard PATH_CHALLENGE frames that would cause it to
+respond at a higher rate.
+
+Path validation tests both directions of a path.  To ensure reachability in both
```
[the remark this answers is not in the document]
janaiyengar: done.

draft-ietf-quic-transport.md
janaiyengar: done.

draft-ietf-quic-transport.md
janaiyengar: removed.

draft-ietf-quic-transport.md
janaiyengar: done. I've simply moved this and the section below into the section above (Responding to Connection Migration).

draft-ietf-quic-transport.md
```diff
@@ -1475,37 +1479,269 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+
+### Initiating Connection Migration {#initiating-migration}
+
+A client can migrate a connection to a new local address by sending all packets
+from the new local address.  Since the server's address is validated during
+connection establishment, receiving acknowledgments for this data serves as
+proof of the server's reachability from the new address.  Note that since
+acknowledgments may be received on any path, return reachability on the new path
+is not established. To establish return reachability on the new path, a client
+MAY concurrently initiate path validation {{migrate-validate}} on the new path.
```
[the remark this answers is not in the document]
janaiyengar: Let's chat about this tomorrow.

PR #1013, opened 2017-12-13
Adds text about server padding handshake to PMTU

### LINKED ISSUE #1010: Server's initial plaintext packet should be PMTU sized
Sec 9 should include this, without which the PMTU discovery that happens with client + server handshake packets is not complete.

janaiyengar:
Closes #1010.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2714,12 +2714,18 @@ but not the UDP or IP header.
 
 The initial client packet SHOULD be padded to exactly 1200 octets unless the
 client has a reasonable assurance that the PMTU is larger.  Sending a packet of
-this size ensures that the network path supports an MTU of this size and helps
-reduce the amplitude of amplification attacks caused by server responses toward
-an unverified client address.
+this size ensures that the network path from the client to the server supports
+an MTU of this size and helps reduce the amplitude of amplification attacks
+caused by server responses toward an unverified client address.
 
 Servers MUST ignore an initial plaintext packet from a client if its total size
 is less than 1200 octets.
+
+Similarly, servers MUST ensure that the first handshake packet they send to
+clients, and any retransmissions of those octets, has a QUIC packet size that is
+the same as the received initial client packet, unless the server knows the PMTU
+to the client to be smaller. Sending a packet of this size ensures that the
+network path from the server to the client supports an MTU of this size.
 
 If a QUIC endpoint determines that the PMTU between any pair of local and remote
 IP addresses has fallen below 1280 octets, it MUST immediately cease sending


janaiyengar:
I understand that there's a PMTU requirement of > 1280 in the draft at the moment, but I have #1014 to track that. I'm happy to rewrite this PR depending on how that issue is resolved.

draft-ietf-quic-transport.md
```diff
@@ -2714,13 +2714,19 @@ but not the UDP or IP header.
...
-reduce the amplitude of amplification attacks caused by server responses toward
-an unverified client address.
+this size ensures that the network path from the client to the server supports
+an MTU of this size and helps reduce the amplitude of amplification attacks
+caused by server responses toward an unverified client address.
 
 Servers MUST ignore an initial plaintext packet from a client if its total size
 is less than 1200 octets.
 
+Similarly, servers MUST ensure that the first handshake packet they send to
```
martinthomson: Use capital H here and maybe be more explicit.  A Version Negotiation or Retry packet shouldn't be larger, and in fact should be kept small.  (We can't pad Version Negotiation anyway, except with greasing values, which isn't much use.)

draft-ietf-quic-transport.md
```diff
@@ -2714,13 +2714,19 @@ but not the UDP or IP header.
...
+clients, and any retransmissions of those octets, has a QUIC packet size that is
+the same as the received initial client packet, unless the server knows the PMTU
+to the client to be smaller. Sending a packet of this size ensures that the
```
martinthomson: This "smaller" clause here means that the server-to-client path could end up with a PMTU less than 1200.  I think that you need to set a minimum of 1200 here explicitly.

huitema: The client side "at least 1200 byte" is justified by a security concern, but that concern does not apply to server responses. I think we need some better text, and that the justification for the padding should come after the requirement of minimum 1280 MTU. Setting a minimum size for the first handshake packet ensures that the handshake will fail if the network does not support the recommended MTU, and is only justified if we want enforce the "minimum 1280" requirement.

But if we do that, maybe we should develop the idea in a full paragraph. For example, is it OK to meet the requirement in IPv4 by setting the DF bit to "fragmentation authorized"?

huitema: I am a bit skeptical about the requirement for "any retransmissions of those octets". It will be yet another special case in the handling of re-transmission. The net effect will be that servers pad all handshake message re-transmissions to 1200 bytes.

martinthomson: I agree that this would make it much more difficult to implement.  You need special handling for this case and I can see how you would end up with a fixed size for Handshake packets.  That would be unfortunate for several reasons.

martinthomson:
Next steps: suggestion was to *suggest* that a server might choose to pad its datagrams to the same size as the incoming datagrams prior to getting path validation.  A suggestion only, because the server might know something else about the MTU that suggests otherwise, and this only really provides minimum MTU validation, it's not really PMTUD.

janaiyengar:
Death by aging.

Issue #794, opened 2017-09-25
SNI requirement for HTTP/QUIC


igorlord:
Per @martinthomson in #495, SNI is currently not required to be used in QUIC, while it is required to be used in HTTP/2.  TLS 1.3 requires SNI to be implemented but apparently does not require it be to used.

This is a proposal to require SNI for QUIC.

The main reason for this is the scarcity of IPv4 addresses.  As long as something is not required, someone will not implement it.  As long as there are enough clients that do not use SNI, virtual service providers will continue to require a large number of IPv4 addresses.  Since IPv4 address conservation is currently a high priority for the Internet, QUIC should strongly encourage (or, better, require) technologies that further this conservation goal.

hardie:
On Mon, Sep 25, 2017 at 3:25 PM, Igor Lubashev <notifications@github.com>
wrote:

> Per @martinthomson <https://github.com/martinthomson> in #495
> <https://github.com/quicwg/base-drafts/issues/495>, SNI is currently not
> required to be used in QUIC, while it is required to be used in HTTP/2. TLS
> 1.3 requires SNI to be implemented but apparently does not require it be to
> used.
>
> This is a proposal to require SNI for QUIC.
>
> The main reason for this is the scarcity of IPv4 addresses. As long as
> something is not required, someone will not implement it. As long as there
> are enough clients that do not use SNI, virtual service providers will
> continue to require a large number of IPv4 addresses. Since IPv4 address
> conservation is currently a high priority for the Internet, QUIC should
> strongly encourage (or, better, require) technologies that further this
> conservation goal.
>
> Nit:  I believe you mean to require SNI, not SIN, though no doubt we'll
get the latter some way or another.

Substantive:  I believe that if we require SNI, we require some method of
preserving the privacy of that SNI.

I'm not sure that the TLS in TLS tunneling solution for encrypted SNI that
https://tools.ietf.org/html/draft-ietf-tls-sni-encryption-00 puts forward
is as neat in the QUIC case.  If the fronting server isn't a QUIC service
but is purely a TLS-in-TLS tunneling proxy, then it probably works as
described.  But if it is both a QUIC service and fronting for other
QUIC-in-TLS-in-TLS services,  then I think we need to carefully work
through how the second  aspects fits in with other QUIC assumptions about
where the transport/application boundary fits.

I think the dual service is something that we need to consider for some of
the same reasons set out in sni-encyption-00 regarding replay attacks: if
[...]

PR #1024, opened 2017-12-18
Expand connection management text 

### LINKED ISSUE #940: Coalescing in HTTP/QUIC
As @LPardue notes, the draft currently omits any language about connection reuse.  We'll need text for that.  As a first pass, note that there is no way to directly address an HTTP/QUIC endpoint by URI/URL.  So an HTTP/QUIC endpoint is somewhat analogous to an HTTP/2 endpoint for an origin which doesn’t appear in the DNS record.

With that as our starting point, here are some thoughts:

- The only reason you’re speaking HTTP/QUIC for example.com in the first place is because you have an Alt-Svc record for example.com.  If you have Alt-Svc records for other origins that happen to share the same certificate pointing to the same host:port, this should be coalesced.
    - It seems a pity that you would need to speak individually to each origin for this.  Coincidentally, Ben Schwartz and I were having a conversation about reviving Alt-Svc records in DNS….
- In the absence of Alt-Svc records, draft-ietf-httpbis-origin-frame permits coalescing in HTTP/2 based on the combination of the server’s possession of the correct certificate and assertion that it wants the traffic.  Like #371, there would need to be a very short draft that defines the ORIGIN frame for HTTP/QUIC as Exceptionally Similar™ to the one in draft-ietf-httpbis-origin-frame.

MikeBishop:
(Fixes #940, #794)

Adds the requirement for TLS as the handshake protocol (agreed in Chicago, but never added to draft), use of SNI, and discusses connection reuse aspects we've probably all been assuming but never wrote down.

I feel like we might still need a stronger text than "QUIC as the underlying transport," since HTTP/QUIC technically has a dependency on the ability for the client to initiate bidirectional streams and the server to initiate unidirectional streams.  There's no guarantee that arbitrary QUIC versions would include that, so we perhaps should state it.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -195,11 +195,13 @@ Once a connection exists to a server endpoint, this connection MAY be reused for
 requests with multiple different URI authority components.  The client MAY send
 any requests for which the client considers the server authoritative.
 
-This typically means that the client has received an Alt-Svc record from the
-request's origin in question which nominates the server endpoint as a valid HTTP
-Alternative Service for that origin.  Clients SHOULD NOT assume that an
-HTTP/QUIC endpoint is authoritative for other origins without an explicit
-signal.
+An authoritative HTTP/QUIC endpoint is typically discovered because the client
+has received an Alt-Svc record from the request's origin which nominates the
+endpoint as a valid HTTP Alternative Service for that origin.  As required by
+{{RFC7838}}, clients MUST validate that the nominated server can present a
+validated certificate for the origin before considering it authoritative.
+Clients SHOULD NOT assume that an HTTP/QUIC endpoint is authoritative for other
+origins without an explicit signal.
 
 A server that does not wish clients to reuse connections for a particular origin
 can indicate that it is not authoritative for a request by sending a 421
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -200,9 +200,9 @@ An authoritative HTTP/QUIC endpoint is typically discovered because the client
 has received an Alt-Svc record from the request's origin which nominates the
 endpoint as a valid HTTP Alternative Service for that origin.  As required by
 {{RFC7838}}, clients MUST check that the nominated server can present a valid
-certificate for the origin before considering it authoritative. Clients SHOULD
-NOT assume that an HTTP/QUIC endpoint is authoritative for other origins without
-an explicit signal.
+certificate for the origin before considering it authoritative. Clients MUST NOT
+assume that an HTTP/QUIC endpoint is authoritative for other origins without an
+explicit signal.
 
 A server that does not wish clients to reuse connections for a particular origin
 can indicate that it is not authoritative for a request by sending a 421
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -153,12 +153,13 @@ MAY omit supported versions for any reason.
 
 HTTP/QUIC relies on QUIC as the underlying transport.  The QUIC version being
 used MUST use TLS version 1.3 or greater as its handshake protocol.  The Server
-Name Indication extension {{!RFC6066}} MUST be included in the TLS handshake.
+Name Indication (SNI) extension {{!RFC6066}} MUST be included in the TLS
+handshake.
 
 QUIC connections are established as described in {{QUIC-TRANSPORT}}. During
 connection establishment, HTTP/QUIC support is indicated by selecting the ALPN
-token "hq" in the crypto handshake.  Support for other application-layer
-protocols MAY be offered in the same handshake.
+token "hq" in the TLS handshake.  Support for other application-layer protocols
+MAY be offered in the same handshake.
 
 While connection-level options pertaining to the core QUIC protocol are set in
 the initial crypto handshake, HTTP-specific settings are conveyed
@@ -198,10 +199,10 @@ any requests for which the client considers the server authoritative.
 An authoritative HTTP/QUIC endpoint is typically discovered because the client
 has received an Alt-Svc record from the request's origin which nominates the
 endpoint as a valid HTTP Alternative Service for that origin.  As required by
-{{RFC7838}}, clients MUST validate that the nominated server can present a
-validated certificate for the origin before considering it authoritative.
-Clients SHOULD NOT assume that an HTTP/QUIC endpoint is authoritative for other
-origins without an explicit signal.
+{{RFC7838}}, clients MUST check that the nominated server can present a valid
+certificate for the origin before considering it authoritative. Clients SHOULD
+NOT assume that an HTTP/QUIC endpoint is authoritative for other origins without
+an explicit signal.
 
 A server that does not wish clients to reuse connections for a particular origin
 can indicate that it is not authoritative for a request by sending a 421
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -95,7 +95,9 @@ frame name will be prefaced with "QUIC."  For example, "QUIC APPLICATION_CLOSE
 frames."  References without this preface refer to frames defined in {{frames}}.
 
 
-# QUIC Advertisement
+# Connection Setup and Management
+
+## Discovering an HTTP/QUIC Endpoint
 
 An HTTP origin advertises the availability of an equivalent HTTP/QUIC endpoint
 via the Alt-Svc HTTP response header or the HTTP/2 ALTSVC frame ({{!RFC7838}}),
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -182,6 +189,22 @@ itself as "hq-09-rickroll". Note that any label MUST conform to the "token"
...
 coordinate their experiments on the quic@ietf.org mailing list.
 
+## Connection Reuse
+
+Once a connection exists to a server endpoint, this connection MAY be reused for
+requests with multiple different URI authority components.  The client MAY send
+any requests for which the client considers the server authoritative.
+
+This typically means that the client has received an Alt-Svc record from the
+request's origin in question which nominates the server endpoint as a valid HTTP
```
martinthomson: This is potentially misleading.  Alt-Svc doesn't form the basis for authority.  I understand the intent here is to say that you don't get here without Alt-Svc, but the "This" in this sentence is immediately after the bit about authority.

MikeBishop: "This" does in fact refer to considering the server authoritative, so you're reading it correctly.  RFC 7838 does say that Alt-Svc provides authority.

> Alternative services are considered authoritative for an origin's resources....

martinthomson: I think that we're talking past each other.  The alternative is not authoritative because of Alt-Svc, that authority derives from the certificate.  This text implies that it derives from Alt-Svc, which is dangerously misleading.

MikeBishop: Perhaps.  I think this is an area that's in flux right now -- example.com:443 is authoritative for https://example.com/, but the certificate validates that you're actually talking to example.com:443.  Under any previous RFC, a server isn't authoritative simply because it has the certificate.

Alt-Svc allows an authoritative server to say that otherguy.org:8080 is co-authoritative with it.  You still need to validate the cert to verify that you've successfully reached the authoritative alternative, but it *is* the Alt-Svc delegation that makes it authoritative.  The certificate simply proves identity when connecting to that authoritative endpoint.

ORIGIN makes the stronger statement that a server can be authoritative *simply* because it has the certificate, if the client chooses to play by those new rules.  That's a new assertion in RFC [TBD], but not something contained in any existing RFC that I'm aware of.

I'll add some text saying that, as required by RFC 7838, the client MUST ensure that the certificate is valid for the origin being requested.

draft-ietf-quic-http.md
```diff
@@ -147,11 +149,16 @@ Origins SHOULD list only versions which are supported by the alternative, but
...
+
+HTTP/QUIC relies on QUIC as the underlying transport.  The QUIC version being
+used MUST use TLS version 1.3 or greater as its handshake protocol.  The Server
+Name Indication extension {{!RFC6066}} MUST be included in the TLS handshake.
 
-HTTP/QUIC connections are established as described in {{QUIC-TRANSPORT}}. During
+QUIC connections are established as described in {{QUIC-TRANSPORT}}. During
 connection establishment, HTTP/QUIC support is indicated by selecting the ALPN
-token "hq" in the crypto handshake.
+token "hq" in the crypto handshake.  Support for other application-layer
```
LPardue: nit: s/crypto/cryptographic to better align with Transport doc terminology. Seems like an opportunity to fix it up while in this area.

draft-ietf-quic-http.md
```diff
@@ -147,11 +149,16 @@ Origins SHOULD list only versions which are supported by the alternative, but
 MAY omit supported versions for any reason.
 
 
-# Connection Establishment {#connection-establishment}
+## Connection Establishment {#connection-establishment}
+
+HTTP/QUIC relies on QUIC as the underlying transport.  The QUIC version being
+used MUST use TLS version 1.3 or greater as its handshake protocol.  The Server
+Name Indication extension {{!RFC6066}} MUST be included in the TLS handshake.
```
LPardue: I missed the SNI reference because the acronym was not used, may help to add _(SNI)_.

RFC7540 also says clients MUST indicate target domain when negotiating TLS, is that requirement inherited here?

draft-ietf-quic-http.md
```diff
@@ -182,6 +189,24 @@ itself as "hq-09-rickroll". Note that any label MUST conform to the "token"
...
+
+Once a connection exists to a server endpoint, this connection MAY be reused for
+requests with multiple different URI authority components.  The client MAY send
+any requests for which the client considers the server authoritative.
+
+An authoritative HTTP/QUIC endpoint is typically discovered because the client
+has received an Alt-Svc record from the request's origin which nominates the
+endpoint as a valid HTTP Alternative Service for that origin.  As required by
+{{RFC7838}}, clients MUST validate that the nominated server can present a
+validated certificate for the origin before considering it authoritative.
```
LPardue: "validate a validated certificate" sounds odd.

draft-ietf-quic-http.md
MikeBishop: I tried a couple of different wordings with the requirement on what to include in SNI, but nothing felt right.  (Particularly in light of the Alt-Svc SNI draft, which suggests that the SNI might be something to which you don't actually intend to direct resources.)  I ultimately decided to leave it at MUST use the extension and MUST make sure you've seen a valid cert for an origin before using the connection for it.  If those two are related or not is a case-by-case choice by the implementation.

draft-ietf-quic-http.md
MikeBishop: How about "TLS," since I've already said we're bound to that anyway?  ;-)

MikeBishop:
Planning to merge soonish, unless @martinthomson is still unhappy with the authority text or someone else squawks.

draft-ietf-quic-http.md
```diff
@@ -182,6 +190,24 @@ itself as "hq-09-rickroll". Note that any label MUST conform to the "token"
...
 
+## Connection Reuse
+
+Once a connection exists to a server endpoint, this connection MAY be reused for
+requests with multiple different URI authority components.  The client MAY send
+any requests for which the client considers the server authoritative.
+
+An authoritative HTTP/QUIC endpoint is typically discovered because the client
+has received an Alt-Svc record from the request's origin which nominates the
+endpoint as a valid HTTP Alternative Service for that origin.  As required by
```
martinthomson: should we use "server" here rather than "endpoint"?

draft-ietf-quic-http.md
```diff
@@ -182,6 +190,24 @@ itself as "hq-09-rickroll". Note that any label MUST conform to the "token"
...
+{{RFC7838}}, clients MUST check that the nominated server can present a valid
+certificate for the origin before considering it authoritative. Clients SHOULD
+NOT assume that an HTTP/QUIC endpoint is authoritative for other origins without
```
martinthomson: "SHOULD NOT assume"?  "cannot" or "MUST NOT" perhaps

draft-ietf-quic-http.md
MikeBishop: What led me to that wording is that the same server might offer both TCP(HTTP/2) and UDP(QUIC) endpoints, so we're not necessarily discovering a new server.  It's ultimately an IP:port that we're connecting to, and I think "endpoint" is a more accurate term there.  Alternate terms for that concept welcome.

draft-ietf-quic-http.md
MikeBishop: Yeah, I don't remember what my logic was for that one.  Probably trying to dance around the ORIGIN-style trust model.  Either is fine.

mikkelfj:
@MikeBishop I'm not that familiar with SNI, but if it is required, will it work in scenarios where HTTP is not using name based addressing, such as in a RPC server cluster where only IP addresses are used for identification?

MikeBishop:
That's a good question -- the HTTP/2 language is 

> The TLS implementation MUST support the Server Name Indication (SNI) [TLS-EXT] extension to TLS.  HTTP/2 clients MUST indicate the target domain name when negotiating TLS.

I assume that's currently working for you...?

martinthomson:
That won't work, because SNI can't technically hold a IP literal.  However, it is common for clients to accept IP addresses in subjectAltName.  The only option there - if you care about compliance - is to make up an SNI (use the named from the reverse DNS tree perhaps) and to have the server effectively ignore the SNI and provide the IP-based certificate.  Practically speaking, if you don't have a name, then omitting SNI is probably fine.

Yet another reminder that this ecosystem is full of landmines.

LPardue:
By omission, do you mean an empty sni extension value, or no extension at all?

Http/Quic sounds looser than Http/2

"The Server Name Indication (SNI) extension [RFC6066] MUST be included in the TLS handshake."

martinthomson:
Yes, and that isn't accurate - it implies that the server is also bound by the requirement - which is technically allowed for servers, but the extension is rarely sent (it makes no sense for a server to send it).

PR #1027, opened 2017-12-18
Fix long header packet type mismatches

### LINKED ISSUE #1026: Packet type header value mismatches
It appears #968 updated the values in the table of packet types for the long header, but did not update the value in the description for each each packet type. 

For example the table currently says: 

| Type | Name                          | Section                     |
|:-----|:------------------------------|:----------------------------|
| 0x7F | Initial                       | {{packet-initial}}          |

While the description under Initial Packet says: 

> The Initial packet uses long headers with a type value of 0x7E

This mismatch appears to exist for all of the long header packet type values.

Lukazoid:
Fixes #1026.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -574,7 +574,7 @@ off-path attackers.
 
 ### Initial Packet {#packet-initial}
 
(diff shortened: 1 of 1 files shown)


mcmanus:
+1

PR #1031, opened 2017-12-24
Make figures comply with 65-character limit

MikeBishop:
In preparation for a hopefully-better linting approach, I found that the current linter is actually allowing slightly more characters than it says it is and we have a few longish figure lines.  No non-whitespace changes, but in recovery I did have to split two things across lines.  Ian, if you'd rather do intermediate calculations here, go for it.


draft-ietf-quic-recovery.md
```diff
@@ -590,37 +590,38 @@ When an ack is received, it may acknowledge 0 or more packets.
...
+      // for rtt calculation and is not ack only.
+      if (!sent_packets[ack.largest_acked].ack_only)
+        max_ack_delay = max(max_ack_delay, ack_delay)
+    // Based on {{?RFC6298}}.
+    if (smoothed_rtt == 0):
+      smoothed_rtt = latest_rtt
+      rttvar = latest_rtt / 2
+    else:
+      rttvar = 3/4 * rttvar + 1/4 *
+        abs(smoothed_rtt - latest_rtt)
```
janaiyengar: how about:
```
rttvar_sample = abs(smoothed_rtt - latest_rtt)
rttvar = 3/4 * rttvar + 1/4 * rttvar_sample
```

PR #1032, opened 2017-12-24
Python linting script

MikeBishop:
Per @martinthomson's suggestion, keep linting every time and just lint more quickly.  `make lint` now runs in about three seconds on WSL, which seems acceptable.  Depends on https://github.com/martinthomson/i-d-template/pull/127 to add Python 3.x support to the Docker image for Circle support and local support simultaneously.  Builds on #1031.


.lint.py
```diff
@@ -0,0 +1,64 @@
...
+
+            ## Skip tables
+            matchObj = re.match('^\s*\|',line)
+            if matchObj:
+                continue
+
+            ## Toggle figure state
+            matchObj = re.match('^[~`]{3,}',line)
+            if matchObj:
+                insideFigure = False if insideFigure else True
```
martinthomson:
```py
insideFigure = not insideFigure
```

.lint.py
```diff
@@ -0,0 +1,64 @@
...
+
+            ## Skip everything before abstract
+            if beforeAbstract:
+                matchObj = re.match('^--- abstract',line)
+                if matchObj:
+                    beforeAbstract = False
+                continue
+
+            ## Skip tables
+            matchObj = re.match('^\s*\|',line)
```
martinthomson: Use `'^\s*(?:\||{:)'` here so that you skip captions as well.

.lint.py
```diff
@@ -0,0 +1,64 @@
...
+            limit = maxFigureLineLength if insideFigure else maxLineLength
+            if length > limit:
+                foundError = True
+                print("{0}: Line is {1} characters; limit is {2}".format(linenumber,length,limit))
+                print(line)
+
+    sys.exit( 1 if foundError else 0)
+
+if __name__ == "__main__":
+   main(sys.argv[1:])
```
martinthomson: No EOL.

Makefile
```diff
@@ -16,13 +16,5 @@ latest:: lint
...
-	  if cat "$$f" | (l=0; while read -r a; do l=$$(($$l + 1)); echo -E "$$l:$$a"; done) | \
-	     sed -e '1,/--- abstract/d;/^[0-9]*: *|/d' | tr -d '\r' | grep '^[0-9]*:.\{81\}'; then \
-	    echo "$$f contains a line with >80 characters"; err=1; \
-	  fi; \
-	  if cat "$$f" | (l=0; while read -r a; do l=$$(($$l + 1)); echo -E "$$l:$$a"; done) | \
-	     sed -e '/^[0-9]*:~~~/,/^[0-9]*:~~~/p;/^[0-9]*:```/,/^[0-9]*:```/p;d' | \
-	     tr -d '\r' | grep '^[0-9]*:.\{66\}'; then \
-	    echo "$$f contains a figure with >65 characters"; err=1; \
-	  fi; \
+	  ./.lint.py -i $$f || err=1; \
```
martinthomson: Rather than iterate here, could you instead iterate in the script?  It would be more natural to drop the `-i` thing and iterate over the remaining arguments that getopt doesn't consume.

.lint.py
```diff
@@ -0,0 +1,64 @@
...
+def main(argv):
+    inputfile = ''
+    maxLineLength = 80
+    maxFigureLineLength = 65
+    foundError = False
+
+    try:
+        opts,args = getopt.getopt(argv,"i:lf")
+    except getopt.GetoptError:
+        print('.lint.py -i <input_file> [-l <line-length>] [-f <figure-line-length>]')
```
martinthomson: sys.stderr.write('...\n') for errors

.lint.py
```diff
@@ -0,0 +1,64 @@
...
+import sys,getopt,re
+
+def main(argv):
+    inputfile = ''
+    maxLineLength = 80
+    maxFigureLineLength = 65
+    foundError = False
+
+    try:
+        opts,args = getopt.getopt(argv,"i:lf")
```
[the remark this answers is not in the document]
MikeBishop: No -- ':' indicates that the argument is mandatory, and these have defaults.

.lint.py
MikeBishop: Somewhat more opaque, but sure.

.lint.py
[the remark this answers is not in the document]
martinthomson: From `man getopt`:

```
       -o, --options shortopts
              The short (one-character) options to be recognized.  If this option is not found, the  first  parameter
              of  getopt  that does not start with a '-' (and is not an option argument) is used as the short options
              string.  Each short option character in shortopts may be followed by one colon to  indicate  it  has  a
              required  argument,  and by two colons to indicate it has an optional argument.  The first character of
              shortopts may be '+' or '-' to influence the way options are parsed and output is generated  (see  sec‐
              tion SCANNING MODES for details).
```

Colon indicates whether the option is followed by an argument.  What you have here lets you do `./.lint.py -l`, but not `./.lint.py -l 20` (the "20" becomes an unattributed argument).

.lint.py
```diff
@@ -0,0 +1,56 @@
...
+for inputfile in args.files:
+    insideFigure = False
+    beforeAbstract = True
+    with open(inputfile,'U') as draft:
+        linecounter = 1
+        lines = draft.readlines()
+
+        abstract = re.compile('^--- abstract')
+        table = re.compile('^\s*(?:\||{:)')
+        figure = re.compile('^[~`]{3,}')
```
martinthomson: You can drop the `,` here.

martinthomson:
OK, this seems good.  A modest improvement on linux.  I'm going to autopep8 this and drop the extra comma.

```
$ git co pylint
$ time make lint

real    0m0.691s
user    0m0.395s
sys     0m0.355s
$ git co master
$ time make lint

real    0m1.008s
user    0m0.588s
sys     0m0.544s
```

Makefile
```diff
@@ -14,5 +14,13 @@ endif
 
 latest:: lint
 .PHONY: lint
+
+PYTHON := $(shell which python3)
+ifeq ($(PYTHON),)
+PYTHON := $(shell which python)
+endif
+
+ifneq ($(PYTHON),)
```
martinthomson: I don't think that we need this check.  We need python for xml2rfc.

MikeBishop:
Yeah, after this the bulk of the time spent on WSL is in make itself, not the lint phase.

```
mbishop@sea-wpwin:/mnt/c/Users/mbishop/Documents/GitHub/quic$ time make lint

real    0m8.921s
user    0m0.422s
sys     0m7.500s
mbishop@sea-wpwin:/mnt/c/Users/mbishop/Documents/GitHub/quic$ time ./.lint.py draft-ietf-quic-*.md

real    0m0.129s
user    0m0.031s
sys     0m0.063s
```

Makefile
```diff
@@ -14,5 +14,13 @@ endif
 
 latest:: lint
 .PHONY: lint
+
+PYTHON := $(shell which python3)
+ifeq ($(PYTHON),)
+PYTHON := $(shell which python)
+endif
```
martinthomson: Try:

```make
PYTHON := $(or $(shell which python3),$(shell which python))
```

martinthomson:
Merged manually.

PR #1036, opened 2018-01-04
Text proposal on user land warning about lower layer overhead to addr…

gloinul:
…ess issue #614

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2779,7 +2779,9 @@ previous PMTU determinations.
 
 In the absence of these mechanisms, QUIC endpoints SHOULD NOT send IP packets
 larger than 1280 octets. Assuming the minimum IP header size, this results in
-a QUIC packet size of 1232 octets for IPv6 and 1252 octets for IPv4.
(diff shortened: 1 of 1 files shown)


MikeBishop:
"operating system user land" is confusing and seems somewhat colloquial.  Also, it doesn't build because you exceed the line length.  You should rewrap to 80 characters per line.

martinthomson:
I don't think that this is specific to user space implementations.  There are two general classes of concern for user space implementations: timing and MTU.  Neither of these are special enough to warrant text (in my opinion).  In all cases, there are things that lowers layers of the stack might do to degrade service, either by adding delays or encapsulation.  Something like TURN is particularly annoying because the overheads can change over time, but it isn't all that different to a routing change causing packets to be sent on a path with a different MTU (on the contrary, it's actually more predictable).

If the intent is to say that you get better performance if you know more about lower layers of the stack, that is a truism.

martinthomson:
That might be a better approach.  I'd be OK with that.

gloinul:
@martinduke Okay, lets go with your suggestion. I have updated the pull request to only do that change.

martinthomson:
Thanks Magus and Martin!

PR #1037, opened 2018-01-05
Clarify some PMTU language

martinduke:
The introductory text to the packet size section has some syntax problems:
- It makes it sound like the Initial packet can be below 1200 Bytes
- The 1200 Byte restriction could, in principle, apply to an out of order 0RTT packet
- The "allow receipt" restriction is vague; the new language is IMO more prescriptive.


martinthomson:
This looks right, and I think that it is clearer.  Do you agree @janaiyengar?

Issue #714, opened 2017-08-08
Multiple connections on the same port


MikeBishop:
@mikkelfj has raised questions about whether we might be able to maintain many connections on the same port.  @ianswett has pointed out that other connection-ID-oriented protocols operate *all* connections on the same port to minimize the use of ephemeral ports.  However, if any of those connections are to the same back end, there will be duplicate 5-tuples, and QUIC's current model doesn't support simultaneous handshakes with the same 5-tuple.

If we want to enable this, we would need to have the server's first flight continue using the client-chosen Connection ID, but contain the new Connection ID.

janaiyengar:
Summarizing editors' discussion: We can use this new connection ID for only 1-RTT encrypted packets. We can further simplify the transport mechanism for server-chosen connection ID, by simply making it part of transport_params.

Note that #713 handles the case of server-chosen connection ID for Stateless Retries.

igorlord:
That said, we should make sure to discourage the use a single client port for all of its QUIC traffic.  Routers do EQMP based a on 5-tuple, and most of them will not be QUIC-aware for years.

igorlord:
@janaiyengar, could you clarify what you mean by: "We can further simplify the transport mechanism for server-chosen connection ID, by simply making it part of transport_params"?

We are starting our load balancer implementation for IETF QUIC, so details like this are very pertinent.

MikeBishop:
Basically, the reasoning in the meeting was this:  If you want multiple handshakes sharing the same 5-tuple in parallel, then you have to be able to identify which handshake a given packet belongs to.  That means the first packet from the server that gives you the server-chosen ID _also_ has to have the client-chosen ID so it can be correlated.  The client can start using the server-chosen ID after that, as currently defined.

If the server's choice of connection ID has to be carried in its first packet and that packet has to have the client-chosen connection ID, then it makes sense to just carry the new connection ID in the transport parameters, which are in that packet anyway for the normal case.  (Of course, we also need a way to carry it in a HelloRetryRequest, but that's #713.)

janaiyengar:
What Mike said. Thanks for speaking up about your implementation effort, Igor, that's useful to know. I think at this point, #713 and this issue are the only ones that we need to resolve... and actually for your load balancer implementation, you only really need this issue to be resolved. I'll spin up a PR soon.

igorlord:
[...]

Issue #713, opened 2017-08-08
Retry packet and Server-Chosen Connection ID


MikeBishop:
In Paris, there was discussion about allowing the server to change the Connection ID on a Server Stateless Retry.  However, the spec doesn't actually define how to do this.  Currently, it says to start a new connection, which could be read to imply a newly-random Connection ID.

For load-shedding, it would be nice to enable the server to point the client to a new server, which likely implies carrying it in the HelloRetryRequest.

janaiyengar:
From editors' discussion on this: 

Our joint recollection is that a new Connection ID can be sent along with the HRR, potentially in a TLS extension. The client uses it in the subsequent Client Initial. The server-side load balancer treating it as a client-chosen connection ID, which means that the server that sent the HRR should, if it wants redirection, generate this Connection ID accordingly.

For verification, the server can send a cookie along with HRR that contains the proposed connection ID, which the client echoes in new Client Initial.

ianswett:
Based on re-reading the draft, I don't believe this was ever written up.  Also, if we do this in an extension, which I like, can we stop having the server unilaterally change the connection ID and instead have the server inform the client to change the connection ID with the TLS extension?

That would solve #714 cleanly I believe.

PR #1041, opened 2018-01-09
Change connection ID with Transport Parameters

ianswett:
Change the connection ID by specifying a new server chosen connection ID in transport parameters.  Allows a new connection ID to be supplied in Retry packets(Fixing issue #713) and allows multiple handshakes to occur on the same 4-tuple, fixing #714.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -639,7 +639,7 @@ A Handshake packet uses long headers with a type value of 0x7D.  It is
 used to carry acknowledgments and cryptographic handshake messages from the
 server and client.
 
-The connection ID field in a Handshake packet contains a connection ID
+The connection ID field in a Handshake packet may contain a connection ID
 that is chosen by the server (see {{connection-id}}).
 
 The first Handshake packet sent by a server contains a randomized packet number.
@@ -690,14 +690,16 @@ The client MUST choose a random connection ID and use it in Initial packets
 ({{packet-initial}}) and 0-RTT packets ({{packet-protected}}).
 
 When the server receives a Initial packet and decides to proceed with the
-handshake, it chooses a new value for the connection ID and sends that in a
-Handshake packet ({{packet-handshake}}).  The server MAY choose to use the value
-that the client initially selects.
+handshake, it may chooses a new value for the connection ID and sends that in
+the server_connection_id transport parameter ({{transport-parameter-definitions}})
+of the Handshake packet ({{packet-handshake}}).
 
-Once the client receives the connection ID that the server has chosen, it MUST
+If the client receives a new connection ID that the server has chosen, it MUST
 use it for all subsequent Handshake ({{packet-handshake}}) and 1-RTT
 ({{packet-protected}}) packets but not for 0-RTT packets ({{packet-protected}}).
 
+The server MUST NOT switch to the new connection ID until has received a
+packet containing the connection ID from the client.  In particular, the
 Server's Version Negotiation ({{packet-version}}) and Retry ({{packet-retry}})
 packets MUST use connection ID selected by the client.
 
@@ -1078,6 +1080,7 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
       stateless_reset_token(6),
       ack_delay_exponent(7),
       initial_max_stream_id_uni(8),
+      server_connection_id(9),
       (65535)
    } TransportParameterId;
 
@@ -1214,6 +1217,11 @@ ack_delay_exponent (0x0007):
   default value of 3 is assumed (indicating a multiplier of 8).  The default
   value is also used for ACK frames that are sent in Initial, Handshake, and
   Retry packets.  Values above 20 are invalid.
+  
+server_connection_id (0x0009):
+: An 8 byte value to be used by the client on all future packets sent to the
+  server.  The server will switch to sending this connection ID once it
+  receives a packet using the connection ID from the client.
 
 
 ### Values of Transport Parameters for 0-RTT {#zerortt-parameters}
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -690,14 +690,16 @@ The client MUST choose a random connection ID and use it in Initial packets
 ({{packet-initial}}) and 0-RTT packets ({{packet-protected}}).
 
 When the server receives a Initial packet and decides to proceed with the
-handshake, it chooses a new value for the connection ID and sends that in a
-Handshake packet ({{packet-handshake}}).  The server MAY choose to use the value
-that the client initially selects.
+handshake, it may chooses a new value for the connection ID and sends that in
```
martinthomson: MAY choose I think

draft-ietf-quic-transport.md
```diff
@@ -690,14 +690,16 @@ The client MUST choose a random connection ID and use it in Initial packets
...
+the server_connection_id transport parameter ({{transport-parameter-definitions}})
+of the Handshake packet ({{packet-handshake}}).
```
martinthomson: A transport parameter can't be "of the Handshake packet".  You can drop from "of" onward.

draft-ietf-quic-transport.md
```diff
@@ -690,14 +690,16 @@ The client MUST choose a random connection ID and use it in Initial packets
...
 
-Once the client receives the connection ID that the server has chosen, it MUST
+If the client receives a new connection ID that the server has chosen, it MUST
```
martinthomson: Transport parameters are only really available once the handshake completes.  For one, they are in TLS EncryptedExtensions, which isn't necessarily right up front.  More importantly, though we could release them before the handshake completes, it makes it harder to reason about their authenticity.  Given that transport parameters are used for client address validation, that's problematic.

draft-ietf-quic-transport.md
```diff
@@ -690,14 +690,16 @@ The client MUST choose a random connection ID and use it in Initial packets
...
 use it for all subsequent Handshake ({{packet-handshake}}) and 1-RTT
 ({{packet-protected}}) packets but not for 0-RTT packets ({{packet-protected}}).
 
+The server MUST NOT switch to the new connection ID until has received a
+packet containing the connection ID from the client.  In particular, the
```
martinthomson: I think that it is easier to make the switch with the change to 1-RTT protected packets.  Creating another mini-protocol here complicates things considerably.  Outside of cases with packet loss (where you can't rely on the client being able to recover the transport parameters), the only real difference is that you are changing the connection ID sent with the client's second flight (it's set of Handshake packets).

Now, I agree that there is potentially some value in having the right markings on the client Handshake packets, but maybe the right answer here is to use NEW_CONNECTION_ID.

ianswett: NEW_CONNECTION_ID wouldn't change the encryption level issues.  And we'd have to allow it during the handshake, which we don't today.

draft-ietf-quic-transport.md
ianswett: Aren't they available at the server side when a CHLO is received?  And shouldn't they be available in the stateless retry?

ianswett:
Possibly I'm confused on the use of transport parameters.  It needs to be possible to negotiate in transport params, so they would have to be available to QUIC prior to the TLS handshake completing.

draft-ietf-quic-transport.md
nibanks: I agree with @ianswett, that the client should be the one that starts using the new connection ID first. Otherwise, you are back to one of the original problems of multiple clients on the same 4-tuple and it is impossible to determine which of them to deliver the newly received packet to.

ianswett:
One other concern about NEW_CONNECTION_ID is that the stateless retry is part of the TLS transcript, and so if the server chosen connection ID is in transport params, the handshake will fail at the end if the connection ID is changed.  I'm not sure why an attacker would want to insert a NEW_CONNECTION_ID frame in this context, but I think it would be impossible for the client to detect that it had occurred?

But if that's not an issue or we don't care, then NEW_CONNECTION_ID is workable.

My core goals are to solve the specified issues.  I do think it's better to have the client be the first to use the new connection ID.  The approach of the server suddenly changing the connection ID is very odd and as discussed, causes problems with multiple simultaneous handshakes.

MikeBishop:
"Negotiate" is perhaps overstating things in the transport parameters.

> During connection establishment, both endpoints make authenticated declarations of their transport parameters. These declarations are made unilaterally by each endpoint. Endpoints are required to comply with the restrictions implied by these parameters; the description of each parameter includes rules for its handling.

In what scenario do you think the transport parameters have to be available before the handshake completes?

ianswett:
The main one I had in mind was feature negotiation.  A client wants to negotiate non-standard feature A and the server replies that A is ok.  I guess a server could enumerate all non-standard features it supported and both could agree after the handshake?

There are probably other use cases I'm not thinking of, but maybe all of them can be worked around.

martinthomson:
It's possible to use the declarations in transport parameters to effect negotiation, sure.  If I declare feature A and you do too, then it's safe to assume that using feature A won't cause problems.  I'd say that we probably need that for - for example - changing the ACK frame format to support new information.  And calling that negotiation isn't a stretch.  Right now, we don't have anything that negotiates and the text reflects that.

As for NEW_CONNECTION_ID, sure, an attacker can insert a connection ID.  The likely effect is to cause packets to be routed incorrectly, likely disrupting the handshake.  We've agreed that if an attacker is able to disrupt the handshake to this degree, then we're not going to try to stop them.  An attacker with that capability could also choose to block the handshake or tamper with it in other ways that would cause failure.  If the concern is misrouting, they could easily rewrite the connection ID.

If the server is able to (maybe because it didn't use the connection ID) and decides to accept the incorrectly-marked packets, then it just became complicit and I'm not sure that we need to worry about that.

One thing that I was wrong about here was that transport parameters are available to a client because the TLS handshake is complete.  There's a nasty race condition there though.  The handshake is really only complete once the client has sent its last flight, and that means using handshake packet protection keys, but using transport parameters from the completed handshake.  It's rather awkward.

We would also need to be clear about what connection ID is used for Handshake packets that the client sends when there is packet loss and it has to acknowledge server Handshake packets.

ianswett:
Thanks Martin, did you want me to re-write this using the NEW_CONNECTION_ID frame or do you think the transport parameters approach is also workable?

martinthomson:
I don't know.  Maybe we can talk about this in Melbourne.

Issue #990, opened 2017-12-05
Neccessity and viability of hidden connection migration linkage


mikkelfj:
If it is desirable to break linkage between connections during path migration then the current approach of skipping random 2^32 packet numbers is likely insufficient.

An oberserver can easily detect decrease in traffic on one connection and a sudden new packet sequence that is higher at some offset. Depending on how many packets have been sent, the migration might just look like a new connection, or it might fall in a higher packet range indicating a migratin with some propability.

If packet numbers are reset upon migration with a new random number in the range 0..2^32 (ca.) such linkage is more limited. The argument goes both ways - resetting adds complexity, and it might not solve the problem:

The silence of one connection and the activity of a new connection could suggest linkage with some probability. When further considering GeoIP, even if on a different network, say wired vs mobile the chances of association increases. Add traffic pattern and timing analysis, and you can probably pin-point migrations with high accuracy.

So, what is the objective of preventing linkage and if it is worthwhile, shouldn't at least packet numbers be encrypted?

Network operators would likely be very upset about non-public packet numbers due to latency and loss monitoring.

Even with encrypted packet numbers linkage might not be too difficult given the amount of data available.

Would it make sense to encrypt the connection ID? Not likely due information in IP/UDP.

Some links on user tracking in Tor networks:
https://www.scmagazine.com/researchers-discover-dns-exploit-that-can-identify-tor-users/article/527591/
https://securelist.com/uncovering-tor-users-where-anonymity-ends-in-the-darknet/70673/

mikkelfj:
One thing that came to mind after writing the above is that the packet number is always transmitted modulo 2^32 or less. Thus a random gap of this range is no different from resetting the packet number assuming the oberserver does not know the real packet number of the connection. But that assumption is not safe depending on how reliably the network can be monitored by the observer.

mikkelfj:
PR https://github.com/quicwg/base-drafts/pull/1043 removes random skips and thus largely resolves this issue, thought it still doens't remove probabilistic linkage, which of couse, is hard.

martinthomson:
Yes, that is very difficult.  Traffic analysis and fingerprinting might further defeat our efforts.  But we do what we can.  I will close this with #1043 unless you think that I missed something (I don't believe that I did).

mikkelfj:
Sure

Issue #1030, opened 2017-12-21
Are random packet number skips still relevant for opportunistic ACK protection?


mikkelfj:
Do we still need to randomly skip packet numbers to protect against opportunistic ACK's know that handshakes are protected by connection ID encryption and considering there are no protection against early connection close injection for on-path attackers?

Having a predictable linear packet number space, aside from infrequent connection migrations, simplifies certain data strutures and large unexpected gaps can also be used by the receiver to detect abuse or massive failure.

Transport 5.4.1:
> The first Initial packet that is sent by a client contains a randomized packet number. All subsequent packets contain a packet number that is incremented by one, see (Section 5.7).

Transport 8.16:
> A sender MAY intentionally skip packet numbers to introduce entropy into the connection, to avoid opportunistic acknowledgement attacks. The sender SHOULD close the connection if an unsent packet number is acknowledged. The format of the ACK frame is efficient at expressing even long blocks of missing packets, allowing for large, unpredictable gaps.

marten-seemann:
The changes to the handshake don't change anything to the opportunistic ACK attack. This attack was always only possible by doing an unverified connection migration.

martinthomson:
I think that the main defense here is explicit path verification.  I would prefer that we remove gaps entirely.  That also means changing the linkability defense, which I want to do for other reasons.

marten-seemann:
I realized that my comment 2 weeks ago is wrong.

The optimistic ACK attack in QUIC comes in two variants. The first is a traffic redirection attack against a third party, which can be solved by explicit path verification. The second is a resource exhaustion attack against the server.

martinthomson:
How does a client mount that attack?  Do they send optimistic ACKs in order to get the server to open their congestion window and send more data?  Wouldn't that be a symmetric attack?

mikkelfj:
If the ACK is contained in a packet encrypted with the connection ID, the attacker must known the connection ID or guess 1/2^64. So effectively the attacker is able to listen to the transmission. How would path verification improve on this before the handshake is complete if the attacker can observe traffic and spoof the source?

marten-seemann:
@martinthomson: The client acks packets that is hasn't received. By doing so, the server will increase the congestion window for this connection and flood the network with packets. Packets will get lost, but since the client is just acking everything, none of those losses will be reported.

[...]

PR #1043, opened 2018-01-09
Mask packet numbers with a per-connection-ID key

### LINKED ISSUE #1034: The packet number gap specification needs work
I spent some time reading the spec of the "packet gap", the random offset applied to a packet number after migrating to a new connection ID. The gap is specified in 7.7.1.1 as:
~~~
Gap = HKDF-Expand-Label(packet_number_secret,
                        "QUIC packet sequence gap", sequence, 4)
~~~
The formula relies on a "packet number secret" derived from the master secret. At first, this seems a bit complicated -- why not just a packet number? The answer of course is that often only the lower 32 bits of the packet sequence number will be transmitted in the header, the higher 32 bits have to be guessed, and if it was just a 32 bit random offset the higher 32 bits would be guessed wrong half the time. And even if the spec mandated only a 31 bit random offset, the guessing would still be wrong half the time if the client migrated twice in a row. OK. So the complexity is justified. But then there are still three valid criticisms: spec is unclear; algorithm does not actually achieve privacy; and all that would not be necessary if the packet sequence number was encrypted.

The current text is hard to follow, due to the use of "sequence number" for both the packet sequence number and the "number of the chosen connection ID in the sequence of new connection ID". I had to read it multiple times to decipher what was actually requested. I would suggest using a more precise moniker than "sequence" to designate the sequence of the connection ID. Maybe "migration sequence".

Then, the algorithm does not achieve privacy. It is unclear from the text if the server also applies the gap, or not. But if it does apply the gap, it will be the same gap as the client, and adversaries can find that to link the old and the new context. And if it does not apply the gap, then adversaries can also link the old and new context. I would suggest that the spec specifies a different computation method for the server side gap, maybe something like:
~~~
Gap = HKDF-Expand-Label(packet_number_secret,
                        "QUIC packet sequence gap", sequence^0xFFFFFFFF, 4)
~~~

Finally, let's observe that all this shifting and offsetting of the packet sequence number would not be necessary if that number was encrypted. Encryption would allow use of a single sequence number on multiple paths without compromising privacy, and that would make the transition to multipath easier.

### LINKED ISSUE #1048: Packet number rework
This collates a number of issues with how we are managing packet number encoding, linkability, randomization and other things.  From the mail sent to the list:

There are number of problems with packet number gaps that I think we
need to address.  Most are minor.  For instance, the initial packet
number is not close to zero, which inflates the size of ACK frames
from the very beginning of a connection.

However, Christian just identified one that I think makes this more
important to address: when using a new connection ID, the gap is the
same size on both client and server.  This makes flows linkable.  The
gaps between the last packet on the old flow and the first packet on
the new flow can be calculated in both directions.  Flows can be
linked by finding flows where client-to-server packets have the same
(or very close) gaps as server-to-client packets.

With a few small tweaks to the current design, I think that we get a
much better solution.  Here is what I propose:

Firstly, packet numbers start at zero and always increase
monotonically, with no gaps.  A side benefit of this is that the first
ACK frame will have a small encoding for largest_acknowledged (1 octet
rather than 8 in most cases).

Packet numbers are never encoded directly onto the wire, they are
XORed with a masking value.  This value will change for different
packet protection keys, and for different connection IDs.  Client and
server will use different values so that gaps are hard to detect.

Thus, we will have a third key derivation:

   key = QHKDF-Expand(S, "key", key_length)
   iv  = QHKDF-Expand(S, "iv", iv_length)
   pnmask = QHKDF-Expand(S, "pnmask", 4)

To make this work seamlessly, I'm proposing a tweak to QHKDF-Expand
that includes the connection ID.  For that, the "Context" field that
we currently do not use seems like a good choice.

This will mean changes to the other uses of QHKDF-Expand:

* the handshake secret can be changed to be a fixed rather than a
derived value, so that we don't need to use the function (connection
ID will be added during expansion into the separate keys)

* during a key update, a zero-length or fixed connection ID will need
to be used so that there is no confusion about which connection ID was
in use at the time the update was made (editorially, I'm not certain
how to manage this one, I might just go back to using
HKDF-Expand-Label directly for that)

Finally, we need to be very clear that even if the connection ID is
omitted, those packets still use the implied connection ID in key
derivation.  We will probably want to recommend that endpoints include
a connection ID if they intend to support NEW_CONNECTION_ID or any of
the related functions, at least around the time when a connection ID
change is in place or the recipient of their packets could have
difficulty picking the right keys.  This is already a problem of
course, but this change would make it worse.

martinthomson:
This changes the key schedule so that connection ID is part of the key
derivation.  Secrets are generally unchanged, except that the handshake secret
is now fixed (we integrate the connection ID via key derivation).

Packet numbers now start from zero, with an additive masking value that is
derived in the same way that packet protection keys are.  Also, the type.

Closes #1048, #1034, #850, #990, #311.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -639,32 +639,68 @@ on the system used in TLS {{!TLS13}}.  The secrets that QUIC uses
 as the basis of its key schedule are obtained using TLS exporters (see Section
 7.5 of {{!TLS13}}).
 
-QUIC uses HKDF with the same hash function negotiated by TLS for
-key derivation.  For example, if TLS is using the TLS_AES_128_GCM_SHA256, the
-SHA-256 hash function is used.
+
+### QHKDF-Expand
+
+QUIC uses the Hash-based Key Derivation Function (HKDF) {{!HKDF=RFC5869}} with
+the same hash function negotiated by TLS for key derivation.  For example, if
+TLS is using the TLS_AES_128_GCM_SHA256, the SHA-256 hash function is used.
+
+Most key derivations in this document use the QHKDF-Expand function, which uses
+the HKDF expand function and is modelled on the HKDF-Expand-Label function from
+TLS 1.3 (see Section 7.1 of {{!TLS13}}). QHKDF-Expand differs from
+HKDF-Expand-Label in that it uses a different base label and omits the Context
+argument.
+
+~~~
+QHKDF-Expand(Secret, Label, Length) =
+   HKDF-Expand(Secret, QhkdfExpandInfo, Length)
+~~~
+
+The HKDF-Expand function used by QHKDF-Expand uses the PRF hash function
+negotiated by TLS, except for handshake secrets and keys derived from them (see
+{{handshake-secrets}}).
+
+Where the `info` parameter of HKDF-Expand is an encoded `QhkdfExpandInfo`
+structure.  This uses the current connection ID in the same place that the TLS
+1.3 HKDF-Expand-Label places a context string:
+
+~~~
+struct {
+   uint16 length = Length;
+   opaque label<6..255> = "QUIC " + Label;
+   uint8 connectionId<8> = ConnectionID;
+} QhkdfExpandInfo;
+~~~
+
+For example, the client packet protection key for a connection ID of
+0x1234567890abcdef uses an `info` parameter of:
+
+~~~
+info = (Hash.length / 256) || (Hash.length % 256) || 0x10 ||
+       "QUIC key" || 0x08 || 0x1234567890abcdef
+~~~
+
 
 ### Handshake Secrets {#handshake-secrets}
 
 Packets that carry the TLS handshake (Initial, Retry, and Handshake) are
-protected with secrets derived from the connection ID used in the
-client's Initial packet. Specifically:
+protected with a static secret.  Specifically:
 
 ~~~
-   quic_version_1_salt = afc824ec5fc77eca1e9d36f37fb2d46518c36639
+handshake_secret = 0x82f42f2e0c6d2b84b600a43af71c2db17353974a
 
-   handshake_secret = HKDF-Extract(quic_version_1_salt,
-                                   client_connection_id)
-
-   client_handshake_secret =
-      QHKDF-Expand(handshake_secret, "client hs", Hash.length)
-   server_handshake_secret =
-      QHKDF-Expand(handshake_secret, "server hs", Hash.length)
+client_handshake_secret =
+   QHKDF-Expand(handshake_secret, "client hs", Hash.length)
+server_handshake_secret =
+   QHKDF-Expand(handshake_secret, "server hs", Hash.length)
 ~~~
 
-The HKDF for the handshake secrets and keys derived from them uses the SHA-256
-hash function {{FIPS180}}.
+The hash function for HKDF when deriving handshake secrets and keys is SHA-256
+{{FIPS180}}.  The connection ID used with QHKDF-Expand is the connection ID
+chosen by the client.
 
-The salt value is a 20 octet sequence shown in the figure in hexadecimal
+The handshake secret is a 20 octet sequence shown in the figure in hexadecimal
 notation. Future versions of QUIC SHOULD generate a new salt value, thus
 ensuring that the keys are different for each version of QUIC. This prevents a
 middlebox that only recognizes one version of QUIC from seeing or modifying the
@@ -685,8 +721,8 @@ early_exporter_secret.  The QUIC 0-RTT secret is only used for protection of
 packets sent by the client.
 
 ~~~
-   client_0rtt_secret
-       = TLS-Exporter("EXPORTER-QUIC 0rtt", "", Hash.length)
+client_0rtt_secret =
+   TLS-Early-Exporter("EXPORTER-QUIC 0rtt", "", Hash.length)
 ~~~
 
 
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1498,46 +1536,17 @@ linkability between two points of network attachment.
...
-                        "QUIC packet sequence gap", sequence, 4)
-~~~
-
-The output of HKDF-Expand-Label is interpreted as a big-endian
-number. "packet_number_secret" is derived from the TLS key exchange,
-as described in Section 5.6 of {{QUIC-TLS}}.
+connection ID provided by the server.  Using a new connection ID will produce a
+new packet protection key and IV.  New values for obscuring header fields (see
+{{header-obscuring}}) ensure that packet numbers can't be used to link activity
+on paths that use different connection IDs.
```
mikkelfj: Is this optional. @ianswett suggested elsewhere that a packet might be dropped if it receives a packet with a tuple that does not match the connection ID. If so, a new connection ID must be used when transmitting on a new path. OTOH, but I am not sure about the implications with NAT rebinding.

draft-ietf-quic-transport.md
```diff
@@ -2377,8 +2386,11 @@ Unlike TCP SACKs, QUIC acknowledgements are irrevocable.  Once a packet has
 been acknowledged, even if it does not appear in a future ACK frame,
 it remains acknowledged.
 
-A client MUST NOT acknowledge Version Negotiation or Retry packets.  These
-packet types contain packet numbers selected by the client, not the server.
+A client MUST NOT acknowledge Version Negotiation or Retry packets.  Version
+Negotiation packets don't contain a packet number and Retry packets include the
+packet number from the Initial packet it responds to.  Rather than relying on
+ACK frames, these packets are implicitly acknowledged by the next Initial packet
+sent by the client.
```
mikkelfj: Not exclusive to this, but is there text to discuss DoS and amplification attacks related to triggering multiple VN and retry packets? One could imagine slowly triggering a lot of retry packets that are never responded to and though the server backs off, it might effectively send a flood to one target.

draft-ietf-quic-transport.md
martinthomson: It can't be optional if it is to be interoperable.

I'm not sure which point Ian was making.  If you lose connection ID and the address tuple changes, you are lost, but either of those can be enough to recover.  The address tuple is only sufficient if the server isn't routing based on connection ID though, and we expect that to be quite common.

draft-ietf-quic-transport.md
martinthomson: We've discussed this previously.  Retry and Version Negotiation are necessarily smaller than the Initial packet that triggers them.  It would therefore be more efficient for the attacker to send its packets directly toward the victim.

draft-ietf-quic-transport.md
mikkelfj: I think it was in relation to two separate connections during handshake in long form, so it might not actually apply here.

draft-ietf-quic-transport.md
```diff
@@ -707,8 +710,8 @@ packets MUST use connection ID selected by the client.
 The packet number is an integer in the range 0 to 2^62-1. The value is used in
 determining the cryptographic nonce for packet encryption.  Each endpoint
 maintains a separate packet number for sending and receiving.  The packet number
-for sending MUST increase by at least one after sending any packet, unless
-otherwise specified (see {{initial-packet-number}}).
+for sending starts at zero for the first packet set and MUST increase by one
+after sending a packet.
```
marten-seemann: As discussed in #1030, random packet number gaps still need to be permitted to defend against the optimistic ACK attack.
If we want to change the defense against this attack, I'd prefer to discuss this in a separate PR.

draft-ietf-quic-tls.md
```diff
@@ -704,86 +740,95 @@ an empty context.  The size of the secret MUST be the size of the hash output
...
 ~~~
-    struct {
-        uint16 length = Length;
-        opaque label<6..255> = "QUIC " + Label;
-        uint8 hashLength = 0;
-    } QuicHkdfLabel;
+struct {
+   uint16 length = Length;
+   opaque label<6..255> = "QUIC " + Label;
+   uint8 connectionId<0>;
```
MikeBishop: Is this included just to emphasize that it's zero-length, or does it actually include a byte representing the zero-length array?

draft-ietf-quic-tls.md
```diff
@@ -704,86 +740,95 @@ an empty context.  The size of the secret MUST be the size of the hash output
...
+: A connection ID might be omitted from the encoding of a packet, as is
+  permitted when the omit_connection_id transport parameter is used, but the
+  connection ID is still used by QHKDF-Expand to derive these values.
+
+The QUIC record protection initially starts with keying material derived from
+handshake keys.  When the TLS state machine reports that the ClientHello has
+been sent, 0-RTT keys can be generated and installed for writing, if 0-RTT is
+available.  Finally, the TLS state machine reports completion of the handshake,
+the 1-RTT keys can be generated and installed for writing.  Additionally, new
+keys are installed each time that the connection ID changes.
```
MikeBishop: Does this need to mention TLS key updates?

draft-ietf-quic-transport.md
```diff
@@ -606,9 +606,12 @@ A Retry packet uses long headers with a type value of 0x7E.  It carries
 cryptographic handshake messages and acknowledgments.  It is used by a server
 that wishes to perform a stateless retry (see {{stateless-retry}}).
 
-The packet number and connection ID fields echo the corresponding fields from
-the triggering client packet.  This allows a client to verify that the server
-received its packet.
+The connection ID field echoes the corresponding fields from the triggering
```
MikeBishop: One field

draft-ietf-quic-transport.md
```diff
@@ -606,9 +606,12 @@ A Retry packet uses long headers with a type value of 0x7E.  It carries
...
+client packet.  This allows a client to correlate a Retry with the Initial
+packet that it sends that the server received its packet.
```
MikeBishop: This sentence doesn't parse.

draft-ietf-quic-transport.md
```diff
@@ -642,11 +645,11 @@ server and client.
...
-The first Handshake packet sent by a server contains a randomized packet number.
-This value is increased for each subsequent packet sent by the server as
-described in {{packet-numbers}}.  The client increments the packet number from
-its previous packet by one for each Handshake packet that it sends (which might
-be an Initial, 0-RTT Protected, or Handshake packet).
+The first Handshake packet sent by a server contains a packet number of 0.  This
+value is increased for each subsequent packet sent by the server as described in
+{{packet-numbers}}.  The client increments the packet number from its previous
+packet by one for each packet that it sends (which might be an
+Initial, 0-RTT Protected, or Handshake packet).
```
MikeBishop: I know this isn't new text, but this statement about the client is redundant -- the client's behavior is already described at line 586.  (In fact, it might be worth consolidating them and simply describing that the first packet sent by each party has a packet number of zero and is incremented with each packet.)

draft-ietf-quic-transport.md
```diff
@@ -789,6 +773,60 @@ format. All other fields MUST be ignored when processing a packet that contains
...
+These changes are applied before packet protection, so the additional
+authenticated data (AAD) input includes masked values.
+
+These are not true cryptographic confidentiality protections, so entities other
+than endpoints are likely to be able to recover the underlying values by
+observing multiple packets.  For instance, in this version of QUIC, packet
+numbers still increase monotonically as long as the connection ID remains
+constant.  However, these values are different for client and server, and they
+change when a connection ID changes, together ensuring that flows with different
+connection IDs are not linkable based on the value of these fields.
```
MikeBishop: Strong statement; maybe we stick to making it more difficult to link?

ianswett:
I need some more time to think about this, but is there a discussion which led to this solution I missed?

Also, this PR is large.  Is it possible to make it more focused?

PR #1045, opened 2018-01-09
Use address validation tokens, not STK

### LINKED ISSUE #1035: Remove the STK in the Spoofed ACK Attack
Source Address Token has been removed from the quic draft upcoming versions.  We should replace the STK with the Address Validation Token or we can define the STK for the spoofed ACK Attack section .

martinthomson:
Closes #1035.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3650,7 +3650,7 @@ transport to cancel a stream in response to receipt of a STOP_SENDING frame.
 ## Spoofed ACK Attack
 
 An attacker might be able to receive an address validation token
-({{address-validation}}) from the server and then releases the IP address it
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3649,12 +3649,13 @@ transport to cancel a stream in response to receipt of a STOP_SENDING frame.
 
 ## Spoofed ACK Attack
 
-An attacker receives an STK from the server and then releases the IP address on
-which it received the STK.  The attacker may, in the future, spoof this same
+An attacker might be able to receive an address validation token
+({{address-validation}}) from the server and then releases the IP address it
```
MikeBishop: "be able to receive... and then release" or "receives... and then releases".

Issue #631, opened 2017-06-14
On-path calculation of RTT


britram:
As discussed at the interim, it might be desirable to allow devices on-path observing two sides of a QUIC connection (or the subset of a multipath QUIC connection) to accurately estimate the RTT of the flow, either for in-band operational purposes (AQM) or general monitoring and measurement purposes (cf. an informal survey of IPFIX information elements suggesting multiple vendors shipping devices that perform this measurement for TCP.)

This facility should require bilateral agreement of both endpoints before an RTT signal is available on path; i.e. it should be optional.

This was a large part of the requirement behind #269, #391, and #393, which were expressed in terms of mechanism instead of requirements. #609 contains a proposal addressing this issue which is separate from packet numbering, and is therefore compatible with proposals to make packet numbering inaccessible to on path devices (see #231)

This issue was filed in part to uplevel the discussion on #609 to requirements.

janaiyengar:
Thanks for filing this issue -- I'd like to move the discussion out of the PR and to this issue. 

On your issue description: I'm not sure that we need to have bilateral agreement. If the mechanism is simple, *and* if we think the information exposed is reasonable, then we could mandate it.

Two issues were raised on PR #609 :

- Maybe use a known/fixed packet type instead of spin bit. I went back and forth on this, and I'm leaning towards using a packet type. Irrespective of greasing, we have packet types, such as Client Initial, that middleboxes will want to identify; this type is no different. We need only specify a single short packet type (with a 4-byte packet number).

- Specify that this signal is per-path. Endpoints will continue receiving and sending packets as usual, so the signal will continuously be rallied between the endpoints despite any path changes. A middlebox that is on one of the two paths will stop or start seeing all packets, and there's nothing about the marked packet that requires it to be the "first" packet. The only place where it *might* matter is with multipath, but we're not there yet.

janaiyengar:
There's another use for this bit/packet type -- it gives a data receiver one RTT sample per round trip, without requiring a PING frame.

BTW, calling it Latency Spin Bit doesn't make sense -- there's positive or negative spin, no latency spin (unless you're thinking that the endpoints could fool the network about path latency with this bit :-))

mcmanus:
On Wed, Jun 14, 2017 at 3:49 PM, janaiyengar <notifications@github.com>
wrote:

> Thanks for filing this issue -- I'd like to move the discussion out of the
> PR and to this issue.
>
> On your issue description: I'm not sure that we need to have bilateral
[...]

PR #1046, opened 2018-01-09
Latency Spin Bit, 2018 edition

britram:
This PR adds a latency spin bit, taking text from [draft-trammell-quic-spin](https://tools.ietf.org/html/draft-trammell-quic-spin). It updates and replaces #609, and addresses #631. Information on usage of the Latency Spin Bit will appear in a separate PR on the manageability draft.

This version of the PR represents the minimum change. It does not grease the spin bit; spin bit greasing is discussed in [Section 6 of draft-trammell-quic-spin](https://tools.ietf.org/html/draft-trammell-quic-spin-01#section-6). It neither proposes a two-bit spin nor a spin-valid bit, the properties of which are explored in [Section 3.3 of draft-trammell-quic-spin](https://tools.ietf.org/html/draft-trammell-quic-spin-01#section-3.3).

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -469,7 +469,7 @@ Key Phase Bit:
   recipient of a packet to identify the packet protection keys that are used to
   protect the packet.  See {{QUIC-TLS}} for details.
 
-Latency Spin Bit: 
+Latency Spin Bit:
 
 : The fourth bit (0x10) of octet 0 is the latency spin bit; which is set as
 described in {{spin-bit}}.
@@ -806,7 +806,7 @@ each endpoint as follows:
   the server, if the packet has a short header and if it increases the
   highest packet number seen by the client from the server, it sets the spin
   value to the opposite of the spin bit in the received packet.
-  
+
   * The server initializes its spin value to 0. When it receives a packet from
   the client, if that packet has a short header and if it increases the
   highest packet number seen by the server from the client, it sets the spin
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -434,7 +434,7 @@ following sections.
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+
-|0|C|K| Type (5)|
+|0|C|K|S|Type(4)|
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -788,6 +793,29 @@ Implementations MUST assume that an unsupported version uses an unknown packet
...
+
+The latency spin bit enables latency monitoring from observation points on the
+network path. Each endpoint, client and server, maintains a spin value, 0 or
+1, for each QUIC connection, and sets the spin bit on packets it sends for
+that connection to the appropriate value. It also maintains the highest packet
+number seen from its peer on the connection. The value is then determined at
+each endpoint as follows:
+
+* The server initializes its spin value to 0. When it receives a packet from
+  the client, if that packet has a short header and if it increments the
```
ianswett: increments -> increases, since packets may be lost or reordered.

draft-ietf-quic-transport.md
```diff
@@ -788,6 +793,29 @@ Implementations MUST assume that an unsupported version uses an unknown packet
...
+## The Latency Spin Bit {#spin-bit}
+
+The latency spin bit enables latency monitoring from observation points on the
+network path. Each endpoint, client and server, maintains a spin value, 0 or
+1, for each QUIC connection, and sets the spin bit on packets it sends for
+that connection to the appropriate value. It also maintains the highest packet
+number seen from its peer on the connection. The value is then determined at
+each endpoint as follows:
+
+* The server initializes its spin value to 0. When it receives a packet from
```
ianswett: I would probably start with the client, since it initiates the process.  Also, it seems it doesn't matter what the server initializes to, since it always echoes the received client's value, so I would omit this sentence.

mikkelfj:
It might be better to have the server flip the spin bit because I could imagine a lot of client only libraries that simply ignores or fail to test the spin bit, whereas merely copying it has a greater chance of success.

draft-ietf-quic-transport.md
britram: Will change the order. 

However, since spinning only happens on the short header, I think it needs to be specified like this to handle the case that the server starts sending short header packets before the client does (e.g., because it's done sending all the non-ACK frames it needs to in the 0RTT part of the connection)

draft-ietf-quic-transport.md
britram: will fix, thanks!

britram:
re client vs server flip: it's worth noting that the endpoint responsible for flipping the bit has slightly more de facto ability to decide not to participate. For that reason, I'd leave the flip on the client side, as it is.

britram:
Superceded by #1283

PR #1054, opened 2018-01-16
Tweak send stream figure

### LINKED ISSUE #1008: Application open not define, Unidirectional Stream open missing
In figure fig-stream-send-states:

"Application Open" appears only in this figure, may be it should be defined somewhere.

Also "Open unidirectional stream" should change to the state "Open" for the sender stream states.

martinthomson:
The use of "Application Open" was confusing.  The text is clearer on this point, so simplify the diagram.

Closes #1008.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2960,7 +2960,7 @@ data to a peer.
 
 ~~~
        o
-       | Application Open
+       | Open Stream
        | Open Paired Stream (bidirectional)
        v
    +-------+
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2960,8 +2960,8 @@ data to a peer.
 
 ~~~
        o
-       | Open Stream
-       | Open Bidirectional Receive Stream
+       | Open Stream (Sending)
+       | Open Bidirectional Stream (Receiving)
        v
    +-------+
    | Open  | Send RST_STREAM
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2960,7 +2960,7 @@ data to a peer.
 
 ~~~
        o
-       | Application Open
+       | Open Stream
        | Open Paired Stream (bidirectional)
        v
```
mikkelfj: The term "Paired Stream" only appears in the figures. Shouldn't this be Open Unidirectional Stream, Open Bidirectional Stream?

martinthomson: Good point.  Does the latest change work for you?

mikkelfj: yeah - sort of - Receive in Send diagram is a bit confusing - but I suppose it means also get ready to receive data.

draft-ietf-quic-transport.md
```diff
@@ -2960,8 +2960,8 @@ data to a peer.
 
 ~~~
        o
-       | Application Open
-       | Open Paired Stream (bidirectional)
+       | Open Stream
+       | Open Bidirectional Receive Stream
```
janaiyengar: You have the directions flipped (should be Send here and Receive below). Admittedly pedantic, but how about "Open Bidirectional Stream (Sending part)"? Calling it a "Bidirectional Send Stream" sounds silly.

martinthomson: @janaiyengar do you have any suggestions for clarifying this?

ianswett: Agreed the current text is confusing.  Open Bidirectional Stream might be clearer.

martinthomson:
Is this better?

Issue #941, opened 2017-11-17
Resume HTTP/2 TLS from QUIC ?


krasic:
Section 2 and 3 of the HTTP mapping describe the progression from HTTP QUIC advertisement to QUIC connection establishment.

For the very first time accessing an HTTP origin, typical progression is:
   o make initial request via HTTP/2
       - this involves TLS 1.3 handshake
       - client see response contains QUIC advertisement
   o subsequent request to same origin 
        - client tries QUIC, does QUIC TLS 1.3 handshake
        - so two RTTs before first QUIC request goes out.

Would it be possible for QUIC to instead of treating it as an initial TLS 1.3 handshake, but rather  treating the first QUIC request as TLS zero RTT, based on resuming the HTTP/2 TLS session (over QUIC)?   Saving 1 RTT for that case?

siyengar:
Do you mean use the same session ticket from a TLS 1.3 connection over TCP for Quic 0-RTT? I don't think anything expressly prohibits x-protocol resumption right now, but we could do a security analysis on this.

britram:
see also https://tools.ietf.org/html/draft-kuehlewind-taps-crypto-sep for some of the implication of the general space of cross-protocol resumption. Seems like it might be useful to allow for this capability in the design, or at least not gratuitously prohibit it.

krasic:
siyengar@: yes.

ianswett:
+1 to making this possible.  We could add a MAY resume with the session ticket from the TCP/TLS 1.3 connection if available?

It would be nice to know if the resumption was likely to work or if the TCP and QUIC TLS 1.3 resumption tokens were completely disjoint.  So possibly we either need to encourage applications to make resumption back and forth between QUIC and TCP work OR include some explicit indicator?

davidben:
QUIC tickets need to include addition source-address information, right? That's not going to exist in TLS a priori. A TLS stack could put it in for you, but that's enough extra work that you probably shouldn't waste single-use tickets on that without an advertisement.

And then you've got the QUIC transport parameters that both sides need to recover for 0-RTT, but TLS over TCP is not going to negotiate that stuff. Though in https://github.com/quicwg/base-drafts/blob/master/draft-ietf-quic-transport.md#values-of-transport-parameters-for-0-rtt-zerortt-parameters, it looks like you all added an extension for the server to supply it independent of what the connection negotiated?

[...]

PR #1057, opened 2018-01-16
Remove cross-protocol resumption

martinthomson:
See https://github.com/quicwg/base-drafts/issues/941#issuecomment-352140849 for rationale.

There was text in the section that I removed.  That had to be restored to a more appropriate location.  I also found some related errors in another section.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1418,38 +1418,6 @@ EncryptedExtensions messages during the handshake.  The extension MAY be
 included in a NewSessionTicket message.
 
 
-## Priming 0-RTT
-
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -1372,13 +1386,15 @@ See {{useless}} for a discussion of these risks.
 To avoid receiving TLS packets that contain no useful data, a TLS implementation
 MUST reject empty TLS handshake records and any record that is not permitted by
 the TLS state machine.  Any TLS application data or alerts that is received
```
RyanAtGoogle: nit: "that is received" => "that are received"

PR #1058, opened 2018-01-16
Remove transport parameters from NewSessionTicket

### LINKED ISSUE #1015: Removing Transport Parameters from new_session_ticket
transport 7.4.2 says that if a new_session_ticket contains the transport parameters extension those parameters should be used as overrides when making future 0-rtt connections with the ticket. 

This is a little hairy to implement and seems to add a bunch of un-necessary complexity and confusion about which transport parameters to expect. The latter is especially concerning because a mixup there results in hard errors.

The motivation, afaict, is to allow smaller defaults for things that are flow controlled on 0rtt vs 1rtt. Presumably the 0-rtt connection would issue a bunch of flow control credits shortly after 1rtt keys are established to bring those two connection styles up to parity with each other.

I propose that transport parameters are removed from NST and the server simply always behaves in this way (i.e. use conservative values in transport parameters and then issue flow control credits when 1RT keys are established. irrespective of whether or not a 0-rtt phase was ever used).

Given that a server that is not going to authenticate its client can send flow control credits at 0.5RTT these updates arrive in the same flight as the handshake in the 1rtt case - i.e. its effectively not adding any latency.. certainly not enough latency to exhaust those smaller defaults.

The TLS APIs here can create quite a bit of complexity... e.g. great care needs to be taken to correctly correlate transport parameters with the right context.. it is a solvable problem, but the complexity seems un-necessary.

martinthomson:
Based on feedback from implementers, this is complex and ultimately not that valuable.

Closes #1015.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1094,9 +1094,6 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
          case encrypted_extensions:
             QuicVersion negotiated_version;
             QuicVersion supported_versions<4..2^8-4>;
-
-         case new_session_ticket:
-            struct {};
       };
       TransportParameter parameters<22..2^16-1>;
    } TransportParameters;
@@ -1218,12 +1215,12 @@ ack_delay_exponent (0x0007):
 
 ### Values of Transport Parameters for 0-RTT {#zerortt-parameters}
 
-Transport parameters from the server MUST be remembered by the client for use
-with 0-RTT data.  If the TLS NewSessionTicket message includes the
-quic_transport_parameters extension, then those values are used for the server
-values when establishing a new connection using that ticket.  Otherwise, the
-transport parameters that the server advertises during connection establishment
-are used.
+A client that attempts to send 0-RTT data MUST remember the transport parameters
+used by the server.  The transport parameters that the server advertises during
+connection establishment apply to all connections that are resumed using the
+keying material established during that handshake.  Remembered transport
+parameters apply to the new connection until the handshake completes and new
+transport parameters from the server can be provided.
 
 A server can remember the transport parameters that it advertised, or store an
 integrity-protected copy of the values in the ticket and recover the information
@@ -1306,8 +1303,7 @@ code.
 The server includes a list of versions that it would send in any version
 negotiation packet ({{packet-version}}) in the supported_versions field.  The
 server populates this field even if it did not send a version negotiation
-packet.  This field is absent if the parameters are included in a
-NewSessionTicket message.
+packet.
 
 The client validates that the negotiated_version is included in the
 supported_versions list and - if version negotiation was performed - that it
@@ -3867,7 +3863,7 @@ Issue and pull request numbers are listed with a leading octothorp.
 
 ## Since draft-ietf-quic-transport-08
 
-No significant changes.
+- Removed transport parameters from NewSessionTicket (#1015)
 
 ## Since draft-ietf-quic-transport-07
 
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1414,8 +1414,7 @@ quic_transport_parameters extension carries a TransportParameters when the
 version of QUIC defined in {{QUIC-TRANSPORT}} is used.
 
 The quic_transport_parameters extension is carried in the ClientHello and the
-EncryptedExtensions messages during the handshake.  The extension MAY be
-included in a NewSessionTicket message.
+EncryptedExtensions messages during the handshake.
 
 
 ## Priming 0-RTT


draft-ietf-quic-transport.md
```diff
@@ -1218,12 +1215,12 @@ ack_delay_exponent (0x0007):
...
-quic_transport_parameters extension, then those values are used for the server
-values when establishing a new connection using that ticket.  Otherwise, the
-transport parameters that the server advertises during connection establishment
-are used.
+A client that attempts to send 0-RTT data MUST remember the transport parameters
+used by the server.  The transport parameters that the server advertises during
+connection establishment apply to all connections that are resumed using the
+keying material established during that handshake.  Remembered transport
+parameters apply to the new connection until the handshake completes and new
+transport parameters from the server can be provided.
```
vasilvv: So, what is our current approach for ensuring that the transport parameter change does not put client in a sudden violation of them?  Do we have a text somewhere else in the text saying that flow control parameters may only increase, or does this still need to be addressed?

martinthomson: The existing text says that if the server accepts 0-RTT it can't change transport parameters to values that the client might violate with its 0-RTT: https://quicwg.github.io/base-drafts/draft-ietf-quic-transport.html#rfc.section.7.4.2

martinthomson:
Yes, that was - at least in part - the motivation for having separate transport parameters.  But the discussion on #1015 suggests that the complexity hit of that exceeds the impact of occasionally having to reject 0-RTT (or not offer it).

ianswett:
I can see the reason for removing the added complexity, but on a practical level, I'm unclear what should occur if a client remembers transport params that the server no longer honors.  If the intent is that the server should honor them, I think it makes sense to state that explicitly.

hardie:
On Wed, Jan 17, 2018 at 7:18 AM, ianswett <notifications@github.com> wrote:

> I can see the reason for removing the added complexity, but on a practical
> level, I'm unclear what should occur if a client remembers transport params
> that the server no longer honors. If the intent is that the server should
> honor them, I think it makes sense to state that explicitly.
>
> The current text is this:

A server can remember the transport parameters that it advertised, or store
an integrity-protected copy of the values in the ticket and recover the
information when accepting 0-RTT data. A server uses the transport
parameters in determining whether to accept 0-RTT data.

So its choices are "honor the transport parameters" and "reject 0-RTT
data", and I believe there are cases where it may need to do both.  In
particular, if it has to turn off a feature that is negotiated uses a
transport parameter, it likely has to move to "reject 0-RTT data" for any
ticket that uses the transport parameter to invoke that feature (see
section 7.3 for the related text on new transport parameters being used to
invoke new features).

ianswett:
Some transport parameters, like flow control and max stream ID are unilateral. If there are no transport params included with the session ticket, then I believe the only transport params the server receives in the client hello are the client transport params.  So when the server receives the client hello, it doesn't know what transport params the client is using.

Or possibly I'm confused.

janaiyengar:
Two questions:
- Does this mean that the server MUST remember the params along as long as the 0-RTT session key is valid?
- What's the expected action for a server that is unable to honor the 0-RTT params?
I think the text should be explicit about these.

hardie:
On Wed, Jan 17, 2018 at 10:56 AM, janaiyengar <notifications@github.com>
wrote:

> Two questions:
>
>    - Does this mean that the server MUST remember the params along as
>    long as the 0-RTT session key is valid?
>
> The text currently gives the server a choice:

"A server can remember the transport parameters that it advertised, or
store an integrity-protected copy of the values in the ticket and recover
the information when accepting 0-RTT data. "

So there are three choices: it can remember the parameters for the validity
period, store them in the ticket, or refuse 0-RTT.


>    - What's the expected action for a server that is unable to honor the
>    0-RTT params?
>    I think the text should be explicit about these.
>
>
If it doesn't know them (because it didn't store them in the ticket and
doesn't remember them),  then it has to refuse 0-RTT.

I think these are the right three choices, so if we need more text to cover
your questions and Ian's similar question, then we should draft a longer
exposition.  If you think these are the wrong choices, of course, that's a
different matter.

janaiyengar:
I think I'm agreeing with Ian that we probably ought to be more direct, specifically in stating that the server MUST do one of those two choices if it wants to do 0-RTT. So, I think using MUST and stating why would be helpful.

martinthomson:
That MUST already exists, here:

> A server MAY accept 0-RTT and subsequently provide different values for transport parameters for use in the new connection. If 0-RTT data is accepted by the server, the server MUST NOT reduce any limits or alter any values that might be violated by the client with its 0-RTT data. 

The text that Ted cited is (important) context for these stipulations.

martinthomson:
Making this part of -09, since that is going to make implementation easier, not harder.

PR #1065, opened 2018-01-23
Transport parameters are stream counts, not IDs

### LINKED ISSUE #1023: Is zero a valid stream-ID in Transport Parameters?
We found this issue in interop testing. One implementation would send transport parameters with max bidir ID and max unidir ID set to zero. The other implementation applied strictly was is in the spec, and rejected the connection. The spec says:

_initial_max_stream_id_bidi (0x0002):
    The initial maximum stream ID parameter contains the initial maximum stream number the peer may initiate for bidirectional streams, encoded as an unsigned 32-bit integer. This value MUST be a valid bidirectional stream ID for a peer-initiated stream (that is, the two least significant bits are set to 0 by a server and to 1 by a client). If an invalid value is provided, the recipient MUST generate a connection error of type TRANSPORT_PARAMETER_ERROR. Setting this parameter is equivalent to sending a MAX_STREAM_ID (Section 8.8) immediately after completing the handshake. The maximum bidirectional stream ID is set to 0 if this parameter is absent, preventing the creation of new bidirectional streams until a MAX_STREAM_ID frame is sent. Note that a default value of 0 does not prevent the cryptographic handshake stream (that is, stream 0) from being used._

If we interpret this strictly, 0 is illegal -- the way to encode 0 is simply to not provide the parameter. Should 0 actually be accepted?

MikeBishop:
Fixes #1023.

Adopts @huitema's suggestion from that issue.  Currently, we send transport parameters which encode max stream IDs for each type.  This forces the other side to then check that we picked an appropriate stream ID for the stream type associated to the parameter.

If you send an initial *count* rather than an initial ID, you not only save two bits, but it's impossible to encode an invalid value.

(Note that this does not in any way revert that we're using maximum stream IDs as our method of limiting open streams; we're not going back to max-concurrent.)

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1167,22 +1167,23 @@ initial_max_streams_bidi (0x0002):
   unsigned 16-bit integer.  If this parameter is absent or zero,
   application-owned bidirectional streams cannot be created until a
   MAX_STREAM_ID frame is sent.  Note that a value of 0 does not prevent the
-  cryptographic handshake stream (that is, stream 0) from being used. For
-  example, a value of 0x05 would be translated to a bidirectional stream ID of
-  20 by a client or 17 by a server.  Setting this parameter is equivalent to
-  sending a MAX_STREAM_ID ({{frame-max-stream-id}}) immediately after completing
-  the handshake containing the resulting Stream ID.
+  cryptographic handshake stream (that is, stream 0) from being used. Setting
+  this parameter is equivalent to sending a MAX_STREAM_ID
+  ({{frame-max-stream-id}}) immediately after completing the handshake
+  containing the resulting Stream ID. For example, a value of 0x05 would be
+  equivalent to receiving a MAX_STREAM_ID containing 20 when received by a
+  client or 17 when received by a server.
 
 initial_max_stream_id_uni (0x0008):
 
 : The initial maximum streams parameter contains the initial maximum number of
   application-owned unidirectional streams the peer may initiate, encoded as an
   unsigned 16-bit integer.  If this parameter is absent or zero, unidirectional
-  streams cannot be created until a MAX_STREAM_ID frame is sent.  For example, a
-  value of 0x05 would be translated to a value of 18 by a client or 19 by a
-  server.  Setting this parameter is equivalent to sending a MAX_STREAM_ID
-  ({{frame-max-stream-id}}) immediately after completing the handshake
-  containing the resulting Stream ID.
+  streams cannot be created until a MAX_STREAM_ID frame is sent.  Setting this
+  parameter is equivalent to sending a MAX_STREAM_ID ({{frame-max-stream-id}})
+  immediately after completing the handshake containing the resulting Stream ID.
+  For example, a value of 0x05 would be equivalent to receiving a MAX_STREAM_ID
+  containing 18 when received by a client or 19 when received by a server.
 
 omit_connection_id (0x0004):
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1162,10 +1162,10 @@ An endpoint MAY use the following transport parameters:
 
 initial_max_streams_bidi (0x0002):
 
-: The initial maximum streams parameter contains the initial maximum number of
-  application-owned bidirectional streams the peer may initiate, encoded as an
-  unsigned 16-bit integer.  If this parameter is absent or zero,
-  application-owned bidirectional streams cannot be created until a
+: The initial maximum bidirectional streams parameter contains the initial
+  maximum number of application-owned bidirectional streams the peer may
+  initiate, encoded as an unsigned 16-bit integer.  If this parameter is absent
+  or zero, application-owned bidirectional streams cannot be created until a
   MAX_STREAM_ID frame is sent.  Note that a value of 0 does not prevent the
   cryptographic handshake stream (that is, stream 0) from being used. Setting
   this parameter is equivalent to sending a MAX_STREAM_ID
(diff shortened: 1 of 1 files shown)


marten-seemann:
I like this change, since it's not possible to encode an invalid value.
I'd prefer to avoid talking about a count though, since this might lead to confusion which streams are counted (at the time that it receives the transport parameters, the client already initiated the crypto stream, and more streams in the 0-RTT case).

MikeBishop:
And those streams count.  Note that you can't encode a negative number, so it's valid to have initiated stream zero no matter what you do.  And if you're initiating streams in 0-RTT, it's because you have some cached value for these settings, and if the server isn't willing to give you at least that value it already MUST reject 0-RTT.

marten-seemann:
I'm aware it doesn't make a difference, I just think this would be clearer if we avoided talking about a count, and instead call this a wire encoding of the respective stream ID.

MikeBishop:
Something like "...contains the initial maximum bidirectional stream number the peer may initiate, encoded as an unsigned 32-bit integer containing the Stream ID minus the two least significant bits."?  I remain a little dubious that that's less confusing.

We encode Stream IDs elsewhere; if we're using a different encoding here, it feels clearer if that's because we're encoding a different thing.

huitema:
You *can* say 0. Just omit the parameter.

martinthomson:
That was what caused #1023 to be opened in the first place, so this is really only addressing the validation concern.

MikeBishop:
I went to add that to the PR, but it turns out that's what it already says:

> The maximum <type> stream ID is set to 0 if this parameter is absent, preventing the creation of new unidirectional streams until a MAX_STREAM_ID frame is sent.

However, there might be people who are sad about not being able to represent the default value.

MikeBishop:
I read the issue as being that if you actually send the default value, it's invalid.  This means that you can't encode something that's invalid, so it does feel like it addresses the root concern.

If people want a way to encode negative Stream IDs here, I think it gets a lot uglier.

huitema:
I would be fine with that suggestion:
~~~
Alternatively, you can subtract so that sending 0 actually means zero streams allowed. That would be more logical, but requires handling the underflow when subtracting.
~~~
It still has the property that all values are good. It just makes the processing a tiny bit more complicated.

draft-ietf-quic-transport.md
```diff
@@ -1160,33 +1160,32 @@ TRANSPORT_PARAMETER_ERROR.
...
+
+: The initial maximum streams parameter contains the initial maximum number of
+  bidirectional streams the peer may initiate, encoded as an unsigned 32-bit
+  integer.  The maximum bidirectional stream ID is set to 0 if this parameter is
+  absent or zero, preventing the creation of new bidirectional streams until a
+  MAX_STREAM_ID frame is sent.  Note that a value of 0 does not prevent the
+  cryptographic handshake stream (that is, stream 0) from being used. Non-zero
+  values are converted to a valid bidirectional stream ID for a peer-initiated
+  stream (that is, after shifting two bits to the left, clients subtract 0 and
+  servers subtract 3).  Setting this parameter is equivalent to sending a
```
janaiyengar: in addition to this algorithm, can you have an example? that ought to help.

draft-ietf-quic-transport.md
```diff
@@ -1160,33 +1160,29 @@ TRANSPORT_PARAMETER_ERROR.
...
-  generate a connection error of type TRANSPORT_PARAMETER_ERROR.  Setting this
-  parameter is equivalent to sending a MAX_STREAM_ID ({{frame-max-stream-id}})
-  immediately after completing the handshake.  The maximum unidirectional stream
-  ID is set to 0 if this parameter is absent, preventing the creation of new
-  unidirectional streams until a MAX_STREAM_ID frame is sent.
+: The initial maximum streams parameter contains the initial maximum number of
+  application-owned unidirectional streams the peer may initiate, encoded as an
+  unsigned 16-bit integer.  If this parameter is absent or zero, unidirectional
+  streams cannot be created until a MAX_STREAM_ID frame is sent.  For example, a
+  value of 0x05 would be translated to a value of 18 by a client or 19 by a
```
martinthomson: s/translated to a value/is equivalent to a MAX_STREAM_ID frame containing 18 when received by a client or 19.../

It's not clear what "value" is from context.  The same applies above.

draft-ietf-quic-transport.md
```diff
@@ -1162,27 +1162,29 @@ An endpoint MAY use the following transport parameters:
...
   MAX_STREAM_ID frame is sent.  Note that a value of 0 does not prevent the
-  cryptographic handshake stream (that is, stream 0) from being used. For
-  example, a value of 0x05 would be translated to a bidirectional stream ID of
-  20 by a client or 17 by a server.  Setting this parameter is equivalent to
-  sending a MAX_STREAM_ID ({{frame-max-stream-id}}) immediately after completing
-  the handshake containing the resulting Stream ID.
+  cryptographic handshake stream (that is, stream 0) from being used. Setting
+  this parameter is equivalent to sending a MAX_STREAM_ID
+  ({{frame-max-stream-id}}) immediately after completing the handshake
+  containing the resulting Stream ID. For example, a value of 0x05 would be
```
martinthomson: "resulting" doesn't make sense, because you don't describe a process that might produce a result.  Try "corresponding".

draft-ietf-quic-transport.md
```diff
@@ -1160,33 +1160,31 @@ TRANSPORT_PARAMETER_ERROR.
...
-  ({{frame-max-stream-id}}) immediately after completing the handshake.  The
-  maximum bidirectional stream ID is set to 0 if this parameter is absent,
-  preventing the creation of new bidirectional streams until a MAX_STREAM_ID
-  frame is sent.  Note that a default value of 0 does not prevent the
-  cryptographic handshake stream (that is, stream 0) from being used.
+initial_max_streams_bidi (0x0002):
+
+: The initial maximum bidirectional streams parameter contains the initial
+  maximum number of application-owned bidirectional streams the peer may
+  initiate, encoded as an unsigned 16-bit integer.  If this parameter is absent
```
marten-seemann: I'd prefer 32 bit here.

MikeBishop: Can you explain why?  Is there a situation in which you'd like to allow that many streams without receiving a MAX_STREAM_ID frame?  @martinthomson suggested reducing the size, and 16 bits still seems quite large for an initial value.

marten-seemann: You're right. I thought that there's no point in saving bytes here, because I overlooked that the transport parameters are sent by the server as well. So it actually makes sense to shave of 4 bytes (in total) here.

PR #257, opened 2017-02-02
do not normatively specify application delivery order

mcmanus:
for #252

PR #1070, opened 2018-01-25
Remove mandatory in-order delivery

### LINKED ISSUE #252: In-order delivery of stream data
transport 9.4 contains: "A receiver MUST ensure that received stream data is delivered to the application as an ordered byte-stream. Data received out of order MUST be buffered for later delivery, as long as it is not in violation of the receiver’s flow control limits."

I agree QUIC provides an ordered byte stream abstraction, but why do we have this requirement about in-order delivery to the application? Surely that makes a lot of sense, but it seems to me the only real requirement is that it be a byte stream because boundaries are not preserved..

martinthomson:
Pursuant to what seemed to be the agreement in Melbourne, this makes it
possible for an endpoint to deliver bytes out of order.  To @ianswett's
request, it also makes it mandatory for endpoints to provide the ability
to deliver in order.

Closes #252, #257.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2860,9 +2860,10 @@ between streams that are initiated by the client and server (see {{stream-id}}).
 Either type of stream can be created by either endpoint, can concurrently send
 data interleaved with other streams, and can be cancelled.
 
-Data that is received on a stream is delivered in order within that stream, but
-there is no particular delivery order across streams.  Transmit ordering among
-streams is left to the implementation.
+Stream offsets allow for the octets on a stream to be placed in order.  An
+endpoint MUST be capable of delivering data received on a stream in order.
+Implementations MAY choose to offer the ability to deliver data out of order.
+There is means of ensuring ordering between octets on different streams.
 
 The creation and destruction of streams are expected to have minimal bandwidth
 and computational cost.  A single STREAM frame may create, carry data for, and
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2860,9 +2860,10 @@ between streams that are initiated by the client and server (see {{stream-id}}).
 Either type of stream can be created by either endpoint, can concurrently send
 data interleaved with other streams, and can be cancelled.
 
-Data that is received on a stream is delivered in order within that stream, but
-there is no particular delivery order across streams.  Transmit ordering among
-streams is left to the implementation.
+Stream offsets allow for the octets on a stream to be placed in order.  An
+endpoint MUST be capable of delivering data received on a stream in order.
+Implementations MAY choose to offer the ability to deliver data out of order.
+There is means of ensuring ordering between octets on different streams.
```
ianswett: I'm not clear on this sentence.

draft-ietf-quic-transport.md
```diff
@@ -3099,10 +3100,10 @@ stream and is providing flow control credit.  A MAX_STREAM_DATA frame might
 arrive before a STREAM or STREAM_BLOCKED frame if packets are lost or reordered.
 
 In the "Recv" state, the endpoint receives STREAM and STREAM_BLOCKED frames.
-Incoming data is buffered and reassembled into the correct order for delivery to
-the application.  As data is consumed by the application and buffer space
-becomes available, the endpoint sends MAX_STREAM_DATA frames to allow the peer
-to send more data.
+Incoming data is buffered and can be reassembled into the correct order for
```
ianswett: I'd argue it's always reassembled, even if we deliver out of order, so I'd keep this as is, but I can see it both ways.

draft-ietf-quic-transport.md
MikeBishop: "There is means"?  Is this supposed to be "There is no means"?

draft-ietf-quic-transport.md
```diff
@@ -3266,7 +3267,7 @@ sender or during delivery to the application at the receiver.
 
 When new data is to be sent on a stream, a sender MUST set the encapsulating
 STREAM frame's offset field to the stream offset of the first byte of this new
-data.  The first byte of data that is sent on a stream has the stream offset 0.
+data.  The first byte of data that is sent on a stream has an offset of 0.
```
jpinner-lyft: I think this should read `The first byte of data on a stream has an offset of 0.`

What you send first depends on what your receive and if you're a proxy you don't want to have to necessarily buffer data until you receive offset 0 just to forward it.

PR #1071, opened 2018-01-25
consistent artwork typing

reschke:


martinthomson:
I don't know why you think we need so many tildes............

reschke:
...I don't know where the # of tildes came from. I just copied it from another instance in the same document.

PR #1076, opened 2018-01-29
Update change logs for -09

martinthomson:


draft-ietf-quic-transport.md
```diff
@@ -3865,6 +3865,17 @@ Issue and pull request numbers are listed with a leading octothorp.
 
 ## Since draft-ietf-quic-transport-08
 
+- Clarified requirements for BLOCKED usage (#65,  #924)
+- BLOCKED frame now includes reason for blocking (#452, #924, #927, #928)
+- Cleartext integrity as version independent (#568)
```
MikeBishop: This is very misleading, since the handshake packet protection is *not* an invariant.

PR #1078, opened 2018-01-29
Version Negotiation doesn't have a packet number

martinthomson:
And all that follows from that...

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -747,8 +747,9 @@ sending a packet with a number of 0x6b4264 requires a 16-bit or larger packet
 number encoding; whereas a 32-bit packet number is needed to send a packet with
 a number of 0x6bc107.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2375,8 +2376,11 @@ Unlike TCP SACKs, QUIC acknowledgements are irrevocable.  Once a packet has
 been acknowledged, even if it does not appear in a future ACK frame,
 it remains acknowledged.
 
-A client MUST NOT acknowledge Version Negotiation or Retry packets.  These
-packet types contain packet numbers selected by the client, not the server.
+A client MUST NOT acknowledge Version Negotiation or Retry packets.  Version
```
ianswett: nit: technically they CAN NOT acknowledge VN packets, since there's no packet number.

How about "A client MUST NOT acknowledge Retry packets.  Version Negotiation packets cannot be acknowledged because they do not contain a packet number."

Issue #725, opened 2017-08-11
Server might be unable to complete handshake due to MAX_STREAM_DATA


martinthomson:
MAX_DATA doesn't apply to stream 0, but MAX_STREAM_DATA does.  The server can increase the space available for the second flight of client handshake messages, but the client can't increase the space available to the server.  The initial value (in the client's transport parameters) determines how much space the server has to send.

Is this a problem?  should we exempt stream 0 (maybe just until the handshake completes)? or should we just advise against setting the value too low?

Note that the value is immutable, if a client wants a low value for general use, that might be a problem for the handshake.  I'm inclined to say that stream 0 can exceed the limit up until the point that the handshake completes, after which the limit strictly applies.  Then, if it has exceeded the limit, the server can't send anything until the client gives it more budget.

janaiyengar:
I don't have a strong opinion on this. This surely is not a problem in practice, but we need to specify it for completeness. I'm leaning towards being conservative here. How about the server sends a CONNECTION_CLOSE with a CERT_TOO_LARGE error?

martinthomson:
If it were only certificates, I'd be less worried.  The potential and imminent need for post quantum key exchange, which generally has large keys, is what makes me concerned that this isn't purely theoretical.  HANDSHAKE_TOO_LARGE maybe, but then we're left with a PQ problem to solve later.

mikkelfj:
Taking the long term view - quantum computing, lattice encryption, more bandwidth: what is sensible today might not be tomorrow. A max limit could be implied by version, or it could unenforced but with a possible implemetation dependent error message. What are the possible scenarios of abuse?

MikeBishop:
Obvious scenario for abuse:  Client discovers a server with a large handshake, spoofs a source address, and launches an amplification attack at a target.  Of course, such a malicious client would just give the server a nice, large flow control window to aid in the amplification attack, so that's slightly independent of the flow control limit.  (That is, the attack wouldn't be worsened by loosening the bounds on Stream 0.)

I could see doing something like "If it can't fit in N times the size of the client's packet, server MUST do source address validation before performing the large handshake."  However, that means we'll lose 0-RTT in post-quantum land, which makes me a little sad.  Validating the address by pickling it in the 0-RTT ticket could partially fix this, but requires mitigations against the attacker capturing a valid ticket from the victim and replaying it many times.  I think TLS 1.3 already suggests mitigations for this, doesn't it?

MikeBishop:
[...]

PR #1082, opened 2018-01-29
Clarify flow control issues on stream 0

### LINKED ISSUE #1074: Stream 0 exemption from flow control
@mikkelfj [asks](https://github.com/quicwg/base-drafts/commit/b52b40bb8ced2bbeda2e4f5d8ab00a2d21a3cc19#r27095625) whether the exemption of stream 0 also means that data counts towards MAX_DATA limits.

In other words, choose between:

* data sent on stream 0 counts as data sent, but it is permitted to exceed flow control limits
* data sent on stream 0 does not count as data sent for the purposes of flow control

MikeBishop:
Fixes #1074, #725 - these are similar problems around flow control on stream zero.

Stream 0 isn't bound to obey connection-level flow control, which necessarily means that it can't consume connection-level flow control either.  That was previously incorrect, since the definition of connection flow control repeatedly referred to "all streams."

On the other hand, Stream 0 can get stalled by stream-level flow control and the client isn't allowed to give it more credit.  This permits Stream 0 to "surge" until the end of the handshake, but then it has to wait for the client to give it more credit until it catches up.

~~~Finally, that surge could provoke an amplification attack if a server has a really large handshake message.  The proposed mitigation to that is to require address validation before sending more than N times the client's initial packet size; since the client has an initial minimum packet size, that can be simplified to N Handshake packets from the server.  I've arbitrarily chosen N=3, but happy to bikeshed that.  I'm not certain I've landed that restriction in the best place, so feel free to suggest moving it elsewhere entirely.~~~

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -650,6 +650,11 @@ described in {{packet-numbers}}.  The client increments the packet number from
 its previous packet by one for each Handshake packet that it sends (which might
 be an Initial, 0-RTT Protected, or Handshake packet).
 
+Servers MUST NOT send more than three Handshake packets in response to a
+client's Initial packet without validating the client's ownership of the
+address, either via a Retry packet or by receiving packets from the client
+in response to the server's Handshake packets.
+
 The payload of this packet contains STREAM frames and could contain PADDING and
 ACK frames.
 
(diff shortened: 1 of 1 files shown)


martinthomson:
Hah, I was working on exactly the same thing.  I prefer a different approach here.  I was just about to submit the PR.

MikeBishop:
I may have erred in conflating three separable items in this PR.

1. The handshake exemption is effectively the same as allowing the flow control window to grow without cryptographic verification, then snapping it shut again once the crypto is in place.  I like that better than "undoing" the credit granted in the clear, but they're effectively equivalent.
2. Either of these produces an amplification attack possibility from a server with a large certificate.  Such an attacker won't give you a small flow control window.  The fixed multiple is an attempt to mitigate that.
3. Fixing #1074 is connection-level flow control.

Let's take them separately -- (1) is an alternative approach to what you've done in #1083 , and I'd argue it's simpler.  Rather than allowing spoofable MAX_STREAM_DATA frames, it just extends the exemption you give to Initial and Reset to the entire handshake.

Either way, we need to limit the amplification possibility of large server handshakes as in (2).  I hear you that a fixed multiple "cannot work" -- better suggestions?

Would these work better as separate PRs?

martinthomson:
1 requires a fixed limit.  That's what I'm objecting to here.  I don't think that we can reasonably limit the size of a certificate chain that way.

2 is already addressed by congestion control and constraints on address validation.

I agree that 3 is separable, but I don't see much value in trying to do a two-step here, since it is all the same text.

MikeBishop:
Based on the conversation just now, I'll lift (2) out of this PR and create a new one once @janaiyengar's PATH_CHALLENGE frame lands.

PR #1083, opened 2018-01-29
Rationalize stream 0 flow control

### LINKED ISSUE #1074: Stream 0 exemption from flow control

martinthomson:
This leaves the exemption to the stream-level flow control for Initial and
Retry packets only (previously it was Initial only).  But, to address the
possibility that the flow control limit is too small for the entire handshake
it permits the sending of MAX_STREAM_DATA (and STREAM_BLOCKED) during the
handshake.

Closes #1074, #725, #1082.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1269,20 +1269,12 @@ both protected with 1-RTT keys and contains an `ACK` frame.
 
 ### Updates to Data and Stream Limits
 
-`MAX_DATA`, `MAX_STREAM_DATA`, `BLOCKED`, `STREAM_BLOCKED`, and `MAX_STREAM_ID`
-frames MUST NOT be sent unprotected.
+`MAX_DATA`, `BLOCKED`, `MAX_STREAM_ID`, and `STREAM_ID_BLOCKED` frames MUST NOT
+be sent unprotected.
 
-Though data is exchanged on stream 0, the initial flow control window on that
-stream is sufficiently large to allow the TLS handshake to complete.  This
-limits the maximum size of the TLS handshake and would prevent a server or
-client from using an abnormally large certificate chain.
-
-Stream 0 is exempt from the connection-level flow control window.
-
-Consequently, there is no need to signal being blocked on flow control.
-
-Similarly, there is no need to increase the number of allowed streams until the
-handshake completes.
+`MAX_STREAM_DATA` and `STREAM_BLOCKED` frames MUST NOT be sent unprotected
+unless they identify stream 0.  Stream 0 is exempt from the connection-level
+flow control window and stream limits.
 
 
 ### Handshake Failures
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3376,19 +3376,17 @@ connection-level data limits established by MAX_DATA.  That is, data sent on
 stream 0 is not counted against the limit expressed by MAX_DATA.  Data on stream
 0 is still subject to stream-level data limits and MAX_STREAM_DATA.
 
-Data sent in the Initial and Retry packets do not count toward stream 0 flow
-control limits.  These packet types do not permit the sending of data in
-multiple packets, so there is no opportunity for a peer to send MAX_STREAM_DATA
-frames.  Note that sending a Retry causes stream offsets and flow control to be
-reset for subsequent packets.  Thus, the only limit on the size of the
-cryptographic handshake messages these packets contain is determined by the MTU.
+Data sent in the Initial and Retry packets are allowed to temporarily exceed
+flow control limits on stream 0.  These packet types do not permit the sending
+of data in multiple packets, so there is no opportunity for a peer to send
+MAX_STREAM_DATA frames.  Note that sending a Retry causes stream offsets and
+flow control to be reset for subsequent packets.  Thus, the only limit on the
+size of the cryptographic handshake messages these packets contain is determined
+by the MTU.
 
-Endpoints MUST respect the stream flow control limit set by their peer when
-sending STREAM frames for stream 0 in packets other than Initial or Retry.  A
-client might exceed stream flow control limits when sending an Initial packet
-and so might need to await a MAX_STREAM_DATA frame from the server before it can
-sent additional STREAM frames.  Endpoints MAY send STREAM_BLOCKED frames for
-stream 0 if they are unable to send due to flow control limits.
+A client might reach or exceed stream flow control limits when sending an
+Initial packet and so might need to await a MAX_STREAM_DATA frame from the
+server before it can send additional STREAM frames.
 
 Flow control for stream 0 is reset after the handshake is completed.  The limit
 is set to the greater of the initial_max_stream_data transport parameter and
@@ -3396,6 +3394,9 @@ current stream offset (that is, the amount of data sent on the stream, excluding
 that sent in Retry and any Initial packet that caused a Retry packet to be
 sent).  This prevents an attacker from modifying an unauthenticated
 MAX_STREAM_DATA frame to inflate the flow control limit.
+
+Endpoints MAY send STREAM_BLOCKED frames for stream 0 in Handshake packets if
+they are unable to send due to flow control limits.
 
 
 ## Edge Cases and Other Considerations
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2825,7 +2825,6 @@ actually lost.
 
 ### Special Considerations for Packetization Layer PMTU Discovery
 
-
 The PADDING frame provides a useful option for PMTU probe packets that does not
 exist in other transports. PADDING frames generate acknowledgements, but their
 content need not be delivered reliably. PADDING frames may delay the delivery of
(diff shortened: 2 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3270,14 +3270,25 @@ delivery, as long as it is not in violation of the receiver's flow control
...
+multiple packets, so there is no opportunity for a peer to send MAX_STREAM_DATA
+frames.  Note also that the next packet sent after either of these packets might
+start at a stream offset of 0.  Thus, the only limit on the size of the
+cryptographic handshake messages these packets contain is determined by the MTU.
+
+Endpoints MUST respect the stream flow control limit set by their peer.  A
+client might exceed this limit when sending an Initial packet and so might need
+to await a MAX_STREAM_DATA frame from the server before it can sent additional
+STREAM frames.  A client MAY send ACK frames and STREAM_BLOCKED frames for
+stream 0.
```
MikeBishop: Given that ACK frames don't reference a stream, the idea of an ACK frame for stream 0 seems undefined.  If that phrase wasn't supposed to apply to the ACK frames as well, then I'm having trouble parsing it.

draft-ietf-quic-transport.md
```diff
@@ -3270,14 +3270,25 @@ delivery, as long as it is not in violation of the receiver's flow control
...
+
+Data sent in the Initial and Retry packets do not count toward stream 0 flow
+control limits.  These packet types do not permit the sending of data in
+multiple packets, so there is no opportunity for a peer to send MAX_STREAM_DATA
+frames.  Note also that the next packet sent after either of these packets might
+start at a stream offset of 0.  Thus, the only limit on the size of the
+cryptographic handshake messages these packets contain is determined by the MTU.
+
+Endpoints MUST respect the stream flow control limit set by their peer.  A
+client might exceed this limit when sending an Initial packet and so might need
```
MikeBishop: Actually, no, they can't -- ClientHello has to fit in the single Initial packet, which you say on 3280 doesn't count against this limit, so who cares?

A client might exceed this limit in a Handshake packet, if the handshake takes more than 1 RTT (or the Client Finished is huge for some reason, I suppose), though.  Is that what you're after?

draft-ietf-quic-transport.md
```diff
@@ -3270,14 +3270,25 @@ delivery, as long as it is not in violation of the receiver's flow control
...
+The cryptographic handshake stream, Stream 0, is exempt from the
+connection-level data limits established by MAX_DATA.  That is, data sent on
+stream 0 is not counted against the limit expressed by MAX_DATA.  Data on stream
+0 is still subject to stream-level data limits and MAX_STREAM_DATA.
+
+Data sent in the Initial and Retry packets do not count toward stream 0 flow
+control limits.  These packet types do not permit the sending of data in
+multiple packets, so there is no opportunity for a peer to send MAX_STREAM_DATA
+frames.  Note also that the next packet sent after either of these packets might
+start at a stream offset of 0.  Thus, the only limit on the size of the
```
MikeBishop: Hm.  Retry is fine, because you're resetting the transport in between, including all the stream states and contents.

But Initial?  You'll have different STREAM frames from different packets identifying the range [0,length) on stream 0 containing different data, and the logic to drop duplicate data needs to *not* drop it on stream 0?  Ick.

If this is the direction you want to go, I'd almost rather see a custom Handshake frame type and let 0 go back to being normal and application-owned.  This direction feels really weird.

draft-ietf-quic-tls.md
```diff
@@ -1269,20 +1269,12 @@ both protected with 1-RTT keys and contains an `ACK` frame.
...
-
-Stream 0 is exempt from the connection-level flow control window.
-
-Consequently, there is no need to signal being blocked on flow control.
-
-Similarly, there is no need to increase the number of allowed streams until the
-handshake completes.
+`MAX_STREAM_DATA` and `STREAM_BLOCKED` frames MUST NOT be sent unprotected
+unless they identify stream 0.  Stream 0 is exempt from the connection-level
+flow control window and stream limits.
```
MikeBishop: But isn't the whole problem here that tampering with these packets can't be detected?  An attacker could cause one side to believe that it had a huge flow control window on stream 0, which could then cause the connection to blow up at some future point if they attempted to send to that limit.  Subsequent, lower MAX_STREAM_DATA frames would be ignored, because they don't increase the largest-seen value.

martinthomson: Yeah, I'm prepared to acknowledge that as a limitation.  Perhaps we could say that the flow control limit is reset to the lower of the initial limit and the amount used once the handshake completes.

F**c I hate stream 0.

draft-ietf-quic-transport.md
martinthomson: The client should only ever send the same data in its Initial packet, so I'm not following your point.  I'm not saying that the Handshake that follows an Initial starts at offset 0 - just that another Initial might follow.

I can see the point for a Retry, in which case both peers will forget (or ignore as invalid) the first set of Initial packets.  There's a nasty gotcha there (#1085), but that probably isn't what you are talking about here.

draft-ietf-quic-transport.md
martinthomson: I should delete this sentence, I don't know where it came from.

Yes, I was looking for flow control for client certificates.

draft-ietf-quic-transport.md
martinthomson: These are two separate things:

* ACK frames
* STREAM_BLOCKED frames for stream 0.

Don't know how to make that clearer, since - as you say - your interpretation is nonsensical.

draft-ietf-quic-transport.md
MikeBishop: Ah, I see what you're getting at.  You're trying to warn that, if there's a Retry, the First Initial and Second Initial will both contain ClientHellos at offset zero, and Retry and the server's Handshake will both contain data starting at offset 0.  I got a totally different interpretation of what you wrote, clearly.

I'd either drop the note, or simply point to the section on Retry with a reminder that the transport state, including stream contents, gets reset following a Retry.

draft-ietf-quic-transport.md
MikeBishop: I think if you change Initial to Handshake, this is fine.

draft-ietf-quic-transport.md
MikeBishop: The statement that "A client MAY send ACK frames" is undoubtedly true, but feels like a bit of a non sequitur without some additional connection to the rest of the paragraph.  Perhaps, "A client in this state SHOULD send a STREAM_BLOCKED frame for stream 0 in a Handshake packet.  This packet MAY also include ACK frames."

draft-ietf-quic-transport.md
martinthomson: Yeah, I reworded.

draft-ietf-quic-transport.md
martinthomson: Yes, I noticed that when reworking this.

draft-ietf-quic-transport.md
martinthomson: Yeah, I removed ACK.

draft-ietf-quic-transport.md
```diff
@@ -3374,6 +3367,36 @@ received on all streams, which are used to check for flow control violations. A
...
 
+
+### Stream 0 Flow Control
+
+The cryptographic handshake stream, Stream 0, is exempt from the
+connection-level data limits established by MAX_DATA.  That is, data sent on
+stream 0 is not counted against the limit expressed by MAX_DATA.  Data on stream
+0 is still subject to stream-level data limits and MAX_STREAM_DATA.
+
+Data sent in the Initial and Retry packets do not count toward stream 0 flow
```
MikeBishop: I'm still a little bothered by "do not count," because that gives us the on-stream version of #1074.  Maybe say "is allowed to temporarily exceed stream 0's flow control limits"?  Otherwise, it looks like the flow control bound is separated from the stream offset, which isn't what you're going for.

draft-ietf-quic-transport.md
```diff
@@ -3374,6 +3367,36 @@ received on all streams, which are used to check for flow control violations. A
...
+control limits.  These packet types do not permit the sending of data in
+multiple packets, so there is no opportunity for a peer to send MAX_STREAM_DATA
+frames.  Note that sending a Retry causes stream offsets and flow control to be
+reset for subsequent packets.  Thus, the only limit on the size of the
+cryptographic handshake messages these packets contain is determined by the MTU.
+
+Endpoints MUST respect the stream flow control limit set by their peer when
+sending STREAM frames for stream 0 in packets other than Initial or Retry.  A
+client might exceed stream flow control limits when sending an Initial packet
```
MikeBishop: Better.  A server can still fall into the same state, though.  Maybe just say "An endpoint might have reached or exceeded stream flow control limits when sending an Initial or Handshake packet and so...."

draft-ietf-quic-tls.md
MikeBishop: Yeah, me too.  That flow control reset idea sounds unpleasant, but probably the best we can do if we have to tolerate window updates in unprotected packets.

martinthomson:
Merged #1083, so this is dead.

PR #1084, opened 2018-01-29
Remove misleading changelog entry

martinthomson:


MikeBishop:
Yeah, that's probably best.  It looks like the actual change (#724) happened in -06, and was the big PR clarifying the handling of incoming packets, when to do VN, etc.

PR #1086, opened 2018-01-31
Introduces PATH_CHALLENGE and PATH_RESPONSE frames

### LINKED ISSUE #1091: Don't overload PING frame semantics
The PING frame is currently overloaded for reachability and path validation semantics. This is icky to implement, especially since an ACK is adequate in one case, while a PONG frame is required for the other. It would be nice to use separate frames for path validation and leave PING to be used for only reachability.

janaiyengar:
Replaces PING with Data and PONG frames with PATH_CHALLENGE and PATH_RESPONSE frames.

Closes #1091.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2587,9 +2587,9 @@ Data:
 
 : This 8-byte field contains arbitrary data.
 
-A PATH_CHALLENGE frame containing at least 8 randomly generated {{?RFC4086}}
-octets is sufficient to ensure that it is easier to receive the packet than it
-is to guess the value correctly.
+A PATH_CHALLENGE frame containing 8 octets that are hard to guess is sufficient
+to ensure that it is easier to receive the packet than it is to guess the value
+correctly.
 
 The recipient of this frame MUST generate a PATH_RESPONSE frame
 ({{frame-path-response}}) containing the same Data.  An endpoint that receives a
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1566,16 +1566,19 @@ Once a connection is established, address validation is relatively simple (see
 endpoint validates a remote address by sending a PATH_CHALLENGE frame containing
 a payload that is hard to guess.  This frame MUST be sent in a packet that is
 sent to the new address.  Once a PATH_RESPONSE frame containing the same payload
-is received, the address is considered to be valid.  The PATH_RESPONSE frame can
-use any path on its return.
+is received, the address is considered to be valid. 
+
+The new address is not considered valid until a PATH_RESPONSE frame containing
+the same payload is received, even if the packet containing the PATH_CHALLENGE
+frame is acknowledged.
+
+The PATH_RESPONSE frame can use any path on its return.
 
 An endpoint MAY send multiple PATH_CHALLENGE frames to handle packet loss or to
 make additional measurements on a new network path.
 
 An endpoint MUST use fresh random data in every PATH_CHALLENGE frame so that it
-can associate the peer's response with the causative PATH_CHALLENGE.  Using
-fresh values allows the endpoint to measure the time taken to respond to a
-challenge.
+can associate the peer's response with the causative PATH_CHALLENGE.
 
 If the PATH_CHALLENGE frame is determined to be lost, a new PATH_CHALLENGE frame
 SHOULD be generated.  This PATH_CHALLENGE frame MUST include new data that is
@@ -1586,8 +1589,8 @@ recovering from possible loss of packets carrying PATH_CHALLENGE and
 PATH_RESPONSE frames, the endpoint MUST terminate the connection.  When setting
 this timer, implementations are cautioned that the new path could have a longer
 round trip time than the original.  The endpoint MUST NOT send a
-CONNECTION_CLOSE frame in this case; it has to assume that the remote peer does
-not want to receive any more packets.
+CONNECTION_CLOSE frame in this case; it has to assume that the remote peer
+cannot want to receive any more packets.
 
 If the remote address is validated successfully, the endpoint MAY increase the
 rate that it sends on the new path using the state from the previous path.  The
@@ -2592,9 +2595,7 @@ to ensure that it is easier to receive the packet than it is to guess the value
 correctly.
 
 The recipient of this frame MUST generate a PATH_RESPONSE frame
-({{frame-path-response}}) containing the same Data.  An endpoint that receives a
-PATH_CHALLENGE frame containing an empty payload MUST generate a connection
-error of type FRAME_ERROR, indicating the PATH_CHALLENGE frame (that is, 0x0e).
+({{frame-path-response}}) containing the same Data.
 
 
 ## PATH_RESPONSE Frame {#frame-path-response}
@@ -2603,11 +2604,9 @@ The PATH_RESPONSE frame (type=0x0f) is sent in response to a PATH_CHALLENGE
 frame.  Its format is identical to the PATH_CHALLENGE frame
 ({{frame-path-challenge}}).
 
-An endpoint that receives a PATH_RESPONSE frame containing an empty payload MUST
-generate a connection error of type FRAME_ERROR, indicating the PATH_RESPONSE
-frame (that is, 0x0f).  If the content of a PATH_RESPONSE frame does not match
-the content of a PATH_CHALLENGE frame previously sent by the endpoint, the
-endpoint MAY generate a connection error of type UNSOLICITED_PATH_RESPONSE.
+If the content of a PATH_RESPONSE frame does not match the content of a
+PATH_CHALLENGE frame previously sent by the endpoint, the endpoint MAY generate
+a connection error of type UNSOLICITED_PATH_RESPONSE.
 
 
 ## STREAM Frames {#frame-stream}
@@ -2711,12 +2710,12 @@ latency.
 
 Regular QUIC packets are "containers" of frames.  When an endpoint receives an
 ACK frame for one or more transmitted packets, all frames in the acknowledged
-packets are considered to have been delivered to the peer, with one exception.
-A PATH_CHALLENGE frame (see {{frame-path-challenge}}) is used to validate a
-peer's ownership of its address.  An ACK frame received for a PATH_CHALLENGE
-frame is not adequate to indicate that the PATH_CHALLENGE was in fact received.
-A PATH_CHALLENGE is considered acknowledged only when the corresponding
-PATH_RESPONSE (see {{frame-path-response}}) is received for it.
+packets are considered to have been received and processed by the peer, with one
+exception.  A PATH_CHALLENGE frame (see {{frame-path-challenge}}) is used to
+validate a peer's ownership of its address.  An ACK frame received for a
+PATH_CHALLENGE frame is not adequate to indicate that the PATH_CHALLENGE was in
+fact received.  A PATH_CHALLENGE is considered acknowledged only when the
+corresponding PATH_RESPONSE (see {{frame-path-response}}) is received for it.
 
 A packet is never retransmitted whole.  How an endpoint handles the loss of a
 frame depends on the type of the frame.  Some frames are simply retransmitted,
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1566,7 +1566,7 @@ Once a connection is established, address validation is relatively simple (see
 endpoint validates a remote address by sending a PATH_CHALLENGE frame containing
 a payload that is hard to guess.  This frame MUST be sent in a packet that is
 sent to the new address.  Once a PATH_RESPONSE frame containing the same payload
-is received, the address is considered to be valid. 
+is received, the address is considered to be valid.
 
 The new address is not considered valid until a PATH_RESPONSE frame containing
 the same payload is received, even if the packet containing the PATH_CHALLENGE
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2712,11 +2712,11 @@ latency.
 Regular QUIC packets are "containers" of frames.  When an endpoint receives an
 ACK frame for one or more transmitted packets, all frames in the acknowledged
 packets are considered to have been delivered to the peer, with one exception.
-A PATH_CHALLENGE frame (see {{path-challenge}}) is used to validate a peer's
-ownership of its address.  An ACK frame received for a PATH_CHALLENGE frame is
-not adequate to indicate that the PATH_CHALLENGE was in fact received.  A
-PATH_CHALLENGE is considered acknowledged only when the corresponding
-PATH_RESPONSE (see {{path-response}}) is received for it.
+A PATH_CHALLENGE frame (see {{frame-path-challenge}}) is used to validate a
+peer's ownership of its address.  An ACK frame received for a PATH_CHALLENGE
+frame is not adequate to indicate that the PATH_CHALLENGE was in fact received.
+A PATH_CHALLENGE is considered acknowledged only when the corresponding
+PATH_RESPONSE (see {{frame-path-response}}) is received for it.
 
 A packet is never retransmitted whole.  How an endpoint handles the loss of a
 frame depends on the type of the frame.  Some frames are simply retransmitted,
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2708,21 +2708,11 @@ implementation decision, and an implementation should be careful to delay
 conservatively, since any delay is likely to increase application-visible
 latency.
 
-Regular QUIC packets are "containers" of frames.  When an endpoint receives an
-ACK frame for one or more transmitted packets, all frames in the acknowledged
-packets are considered to have been received and processed by the peer, with one
-exception.  A PATH_CHALLENGE frame (see {{frame-path-challenge}}) is used to
-validate a peer's ownership of its address.  An ACK frame received for a
-PATH_CHALLENGE frame is not adequate to indicate that the PATH_CHALLENGE was in
-fact received.  A PATH_CHALLENGE is considered acknowledged only when the
-corresponding PATH_RESPONSE (see {{frame-path-response}}) is received for it.
-
-A packet is never retransmitted whole.  How an endpoint handles the loss of a
-frame depends on the type of the frame.  Some frames are simply retransmitted,
-some have their contents moved to new frames, and others are never
-retransmitted.
-
-When a packet is detected as lost, the sender re-sends any frames as necessary:
+Regular QUIC packets are "containers" of frames.  A packet is never
+retransmitted whole.  How an endpoint handles the loss of a frame depends on the
+type of the frame.  Some frames are simply retransmitted, some have their
+contents moved to new frames, and others are never retransmitted.  When a packet
+is detected as lost, the sender re-sends any frames as necessary:
 
 * All application data sent in STREAM frames MUST be retransmitted, unless the
   endpoint has sent a RST_STREAM for that stream.  When an endpoint sends a
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -849,8 +849,9 @@ explained in more detail as they are referenced later in the document.
 | 0x0a        | STREAM_ID_BLOCKED | {{frame-stream-id-blocked}} |
 | 0x0b        | NEW_CONNECTION_ID | {{frame-new-connection-id}} |
 | 0x0c        | STOP_SENDING      | {{frame-stop-sending}}      |
-| 0x0d        | PONG              | {{frame-pong}}              |
-| 0x0e        | ACK               | {{frame-ack}}               |
+| 0x0d        | ACK               | {{frame-ack}}               |
+| 0x0e        | PATH_CHALLENGE    | {{frame-path-challenge}}    |
+| 0x0f        | PATH_RESPONSE     | {{frame-path-response}}     |
 | 0x10 - 0x17 | STREAM            | {{frame-stream}}            |
 {: #frame-types title="Frame Types"}
 
@@ -1562,25 +1563,32 @@ consider the possibility that packets are sent without congestion feedback.
 
 Once a connection is established, address validation is relatively simple (see
 {{address-validation}} for the process that is used during the handshake).  An
-endpoint validates a remote address by sending a PING frame containing a payload
-that is hard to guess.  This frame MUST be sent in a packet that is sent to the
-new address.  Once a PONG frame containing the same payload is received, the
-address is considered to be valid.  The PONG frame can use any path on its
-return.  A PING frame containing 12 randomly generated {{?RFC4086}} octets is
-sufficient to ensure that it is easier to receive the packet than it is to guess
-the value correctly.
+endpoint validates a remote address by sending a PATH_CHALLENGE frame containing
+a payload that is hard to guess.  This frame MUST be sent in a packet that is
+sent to the new address.  Once a PATH_RESPONSE frame containing the same payload
+is received, the address is considered to be valid.  The PATH_RESPONSE frame can
+use any path on its return.  A PATH_CHALLENGE frame containing 12 randomly
+generated {{?RFC4086}} octets is sufficient to ensure that it is easier to
+receive the packet than it is to guess the value correctly.
 
-If the PING frame is determined to be lost, a new PING frame SHOULD be
-generated.  This PING frame MUST include a new Data field that is similarly
-difficult to guess.
+An endpoint MAY send multiple PATH_CHALLENGE frames to handle packet loss or to
+make additional measurements on a new network path.
+
+An endpoint MUST use fresh random data in every PATH_CHALLENGE frame so that it
+can associate the peer's response with the causative PATH_CHALLENGE,
+additionally helping the endpoint make more accurate path measurements.
+
+If the PATH_CHALLENGE frame is determined to be lost, a new PATH_CHALLENGE frame
+SHOULD be generated.  This PATH_CHALLENGE frame MUST include new data that is
+similarly difficult to guess.
 
 If validation of the new remote address fails, after allowing enough time for
-possible loss and recovery of packets carrying PING and PONG frames, the
-endpoint MUST terminate the connection.  When setting this timer,
-implementations are cautioned that the new path could have a longer round trip
-time than the original.  The endpoint MUST NOT send a CONNECTION_CLOSE frame in
-this case; it has to assume that the remote peer does not want to receive any
-more packets.
+recovering from possible loss of packets carrying PATH_CHALLENGE and
+PATH_RESPONSE frames, the endpoint MUST terminate the connection.  When setting
+this timer, implementations are cautioned that the new path could have a longer
+round trip time than the original.  The endpoint MUST NOT send a
+CONNECTION_CLOSE frame in this case; it has to assume that the remote peer does
+not want to receive any more packets.
 
 If the remote address is validated successfully, the endpoint MAY increase the
 rate that it sends on the new path using the state from the previous path.  The
@@ -1597,17 +1605,17 @@ After verifying an address, the endpoint SHOULD update any address validation
 tokens ({{address-validation}}) that it has issued to its peer if those are no
 longer valid based on the changed address.
 
-Address validation using the PING frame MAY be used at any time by either peer.
-For instance, an endpoint might check that a peer is still in possession of its
-address after a period of quiescence.
+Address validation using the PATH_CHALLENGE frame MAY be used at any time by
+either peer.  For instance, an endpoint might check that a peer is still in
+possession of its address after a period of quiescence.
 
 Upon seeing a connection migration, an endpoint that sees a new address MUST
 abandon any address validation it is performing with other addresses on the
 expectation that the validation is likely to fail.  Abandoning address
-validation primarily means not closing the connection when a PONG frame is not
-received, but it could also mean ceasing retransmissions of the PING frame.  An
-endpoint that doesn't retransmit a PING frame might receive a PONG frame, which
-it MUST ignore.
+validation primarily means not closing the connection when a PATH_RESPONSE frame
+is not received, but it could also mean ceasing subsequent transmissions of the
+PATH_CHALLENGE frame.  An endpoint MUST ignore any subsequently received
+PATH_RESPONSE frames from that address.
 
 
 ## Spurious Connection Migrations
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2352,17 +2337,56 @@ Application Error Code:
...
+
+The sender of this frame MUST include at least one octet of data in the Data
+field.
+
+The recipient of this frame MUST generate a PATH_RESPONSE frame
+({{frame-path-response}}) containing the same Data.  An endpoint that receives a
+PATH_CHALLENGE frame containing an empty payload MUST generate a connection
+error of type FRAME_ERROR, indicating the PATH_CHALLENGE frame (that is, 0x0e).
+
+A PATH_CHALLENGE frame MUST NOT elicit acknowledgements; the corresponding
```
ianswett: Are we clear on the meaning of elicit here?  We definitely intend for the packet containing the PATH_CHALLENGE frame to be acknowledged.  Given that, I'd be inclined to just consider it a frame we acknowledge and remove this paragraph.

draft-ietf-quic-transport.md
```diff
@@ -1562,25 +1563,32 @@ consider the possibility that packets are sent without congestion feedback.
...
-If the PING frame is determined to be lost, a new PING frame SHOULD be
-generated.  This PING frame MUST include a new Data field that is similarly
-difficult to guess.
+endpoint validates a remote address by sending a PATH_CHALLENGE frame containing
+a payload that is hard to guess.  This frame MUST be sent in a packet that is
+sent to the new address.  Once a PATH_RESPONSE frame containing the same payload
+is received, the address is considered to be valid.  The PATH_RESPONSE frame can
+use any path on its return.  A PATH_CHALLENGE frame containing 12 randomly
+generated {{?RFC4086}} octets is sufficient to ensure that it is easier to
+receive the packet than it is to guess the value correctly.
```
martinthomson: I think that we can remove this last sentence now that PATH_CHALLENGE is fixed size.  FWIW, I think that we could drop back to 8 happily.  12 raised some eyebrows, and the handshake uses fewer octets.  That said, 12 means that you could include 8 octets of randomness and a timestamp - or a timestamp and an 8 octet HMAC - so that you don't have to maintain state and you can still measure RTT.

draft-ietf-quic-transport.md
```diff
@@ -1562,25 +1563,32 @@ consider the possibility that packets are sent without congestion feedback.
...
+
+An endpoint MAY send multiple PATH_CHALLENGE frames to handle packet loss or to
+make additional measurements on a new network path.
+
+An endpoint MUST use fresh random data in every PATH_CHALLENGE frame so that it
+can associate the peer's response with the causative PATH_CHALLENGE,
+additionally helping the endpoint make more accurate path measurements.
```
martinthomson: Split the last clause here into a new sentence: Using fresh values allows the endpoint to measure the time taken to respond to a challenge.

draft-ietf-quic-transport.md
```diff
@@ -1562,25 +1563,32 @@ consider the possibility that packets are sent without congestion feedback.
...
+
+If the PATH_CHALLENGE frame is determined to be lost, a new PATH_CHALLENGE frame
+SHOULD be generated.  This PATH_CHALLENGE frame MUST include new data that is
```
martinthomson: Rather than SHOULD here, I think that you need to talk about sending multiple to deal with the potential of loss, but limiting the total time or number of challenges and eventually giving up.

draft-ietf-quic-transport.md
```diff
@@ -2352,17 +2337,56 @@ Application Error Code:
   {{app-error-codes}}).
 
 
+## PATH_CHALLENGE Frame {#frame-path-challenge}
 
-## PONG Frame {#frame-pong}
+Endpoints can use PATH_CHALLENGE frames (type=0x0e) to check reachability to the
+peer, to verify a new path's PMTU, and for address validation during connection
```
martinthomson: I think that you should drop the PMTU bit here.  It's the packet size that determines PMTU validity anyway.

draft-ietf-quic-transport.md
```diff
@@ -2352,17 +2337,56 @@ Application Error Code:
...
+establishment and connection migration.
 
-The PONG frame (type=0x0d) is sent in response to a PING frame that contains
-data.  Its format is identical to the PING frame ({{frame-ping}}).
+PATH_CHALLENGE frames contain a variable-length payload.
```
martinthomson: A fixed size is better.  No chance of messing it up that way.  As I said above, I think that 12 octets is pretty good (that's the size of the STUN transaction ID).

draft-ietf-quic-transport.md
```diff
@@ -2352,17 +2337,56 @@ Application Error Code:
...
+: This variable-length field contains arbitrary data.
 
-An endpoint that receives an unsolicited PONG frame - that is, a PONG frame
-containing a payload that is empty MUST generate a connection error of type
-FRAME_ERROR, indicating the PONG frame (that is, 0x10d).  If the content of a
-PONG frame does not match the content of a PING frame previously sent by the
-endpoint, the endpoint MAY generate a connection error of type UNSOLICITED_PONG.
+
+The sender of this frame MUST include at least one octet of data in the Data
+field.
```
martinthomson: To the above, one octet is of almost no value.  You would need many PATH_RESPONSE frames.

draft-ietf-quic-transport.md
```diff
@@ -2352,17 +2337,56 @@ Application Error Code:
...
+
+The recipient of this frame MUST generate a PATH_RESPONSE frame
+({{frame-path-response}}) containing the same Data.  An endpoint that receives a
+PATH_CHALLENGE frame containing an empty payload MUST generate a connection
+error of type FRAME_ERROR, indicating the PATH_CHALLENGE frame (that is, 0x0e).
```
martinthomson: When?  If the secondary value of PATH_CHALLENGE is to be realized, this should be done immediately.  Or are we expecting the ACK frame to include ACK Delay and we expect challengers to use that to correct their view of path delay?  With the potential for the ACK or response to follow a different path, it's hard to know how reliable this will be for measuring path RTT.

draft-ietf-quic-transport.md
martinthomson: I'm with Ian here.  I think that we can treat this frame as normal.  I can see how you might want to exclude PATH_RESPONSE from acknowledgment, but even there I would just let the ACKs flow.

draft-ietf-quic-transport.md
```diff
@@ -3868,6 +3896,11 @@ thanks to all.
 
 Issue and pull request numbers are listed with a leading octothorp.
 
+## Since draft-ietf-quic-transport-09
+
+- Added PATH_CHALLENGE and PATH_RESPONSE frames to replace PING with Data and
+  PONG frame. Changed ACK frame type from 0x0e to 0x0d. (#000)
```
martinthomson: Can you open an issue for this, so we can track it?

draft-ietf-quic-transport.md
```diff
@@ -2149,43 +2157,20 @@ than it has sent, unless this is a result of a change in the initial limits (see
...
-~~~
- 0                   1                   2                   3
- 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|   Length(8)   |                 Data (*)                    ...
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-~~~
-
-Length:
+The receiver of a PING frame simply need to acknowledge the packet containing
```
erickinnear: Nit: need -> needs

draft-ietf-quic-transport.md
```diff
@@ -2352,17 +2337,56 @@ Application Error Code:
   {{app-error-codes}}).
 
 
+## PATH_CHALLENGE Frame {#frame-path-challenge}
```
erickinnear: Should these fall after ACK frames in terms of ordering?

draft-ietf-quic-transport.md
erickinnear: I didn't think that we previously intended a packet containing only a PATH_CHALLENGE frame to be acknowledged, but I think that could make sense. 

Minor question though: 
What if we get a PATH_RESPONSE, but no ACK? Do we then retransmit the PATH_CHALLENGE? 

The previous comment about the packets potentially being sent on (or taking) different paths is interesting, because we could ACK along the current path, but send the PATH_RESPONSE along the same path the PATH_CHALLENGE came in on. This could provide valuable information to the endpoint about the viability of those two paths, although we still end up in the situation of a rapidly failing path causing the ACK to not be delivered, which is not the end of the world.

draft-ietf-quic-transport.md
erickinnear: If we require the PATH_RESPONSE to be sent on the same path as the PATH_CHALLENGE, do we place any requirements on the ACK? And does such a requirement make sense?

draft-ietf-quic-transport.md
```diff
@@ -2352,17 +2337,56 @@ Application Error Code:
...
+An endpoint that receives a PATH_RESPONSE frame containing an empty payload MUST
+generate a connection error of type FRAME_ERROR, indicating the PATH_RESPONSE
+frame (that is, 0x0e).  If the content of a PATH_RESPONSE frame does not match
+the content of a PATH_CHALLENGE frame previously sent by the endpoint, the
+endpoint MAY generate a connection error of type UNSOLICITED_PATH_RESPONSE.
+
+A PATH_RESPONSE frame MUST NOT elicit an acknowledgement.
 
 
 ## ACK Frame {#frame-ack}
```
erickinnear: Minor: This is now type `0x0d`

draft-ietf-quic-transport.md
erickinnear: Also, given that we stated that a new transmission of a PATH_CHALLENGE frame due to a lack of a PATH_RESPONSE should include different random data, how does that interact with loss recovery due to ACKs being present/not being present? 

(In other words, I can see an argument for either behavior, but we should make sure that we update below if we change this.)

draft-ietf-quic-transport.md
ianswett: I don't believe we have specified packet ordering if that's what you're referring to, so I'd prefer not to start now.

draft-ietf-quic-transport.md
ianswett: Fixed 8 bytes SGTM.

draft-ietf-quic-transport.md
```diff
@@ -1562,25 +1563,32 @@ consider the possibility that packets are sent without congestion feedback.
...
 
 If validation of the new remote address fails, after allowing enough time for
-possible loss and recovery of packets carrying PING and PONG frames, the
-endpoint MUST terminate the connection.  When setting this timer,
-implementations are cautioned that the new path could have a longer round trip
-time than the original.  The endpoint MUST NOT send a CONNECTION_CLOSE frame in
-this case; it has to assume that the remote peer does not want to receive any
-more packets.
+recovering from possible loss of packets carrying PATH_CHALLENGE and
+PATH_RESPONSE frames, the endpoint MUST terminate the connection.  When setting
```
nibanks: Why should the connection be terminated if the response wasn't received? What if the new path was just being interrogated for viability and it just wasn't viable (too much loss, blocked in one direction or got disconnected)? Should that really cause the good path to be killed as well?

draft-ietf-quic-transport.md
nibanks: +1 for fixed size.

draft-ietf-quic-transport.md
```diff
@@ -3620,10 +3648,10 @@ PROTOCOL_VIOLATION (0xA):
 : An endpoint detected an error with protocol compliance that was not covered by
   more specific error codes.
 
-UNSOLICITED_PONG (0xB):
+UNSOLICITED_PATH_RESPONSE (0xB):
 
-: An endpoint received a PONG frame that did not correspond to any PING frame
-  that it previously sent.
+: An endpoint received a PATH_RESPONSE frame that did not correspond to any
+  PATH_CHALLENGE frame that it previously sent.
```
nibanks: Do we really need to error out if a response if received for a challenge we don't have saved at the moment? What if the sender of the challenge ended up timing out the packet too aggressively and threw away its state for it? If we really need to reliability validate this, how long would the sender of the challenge need to keep this state around then?

draft-ietf-quic-transport.md
erickinnear: Ah no, sorry, this is *much* more minor: I'm referring to ordering within the document itself (although that might not be a concern at this point), which no longer matches the table above.

draft-ietf-quic-transport.md
erickinnear: This is specific to address validation, I think. 
So, if you're just probing new paths for viability then you're not doing this and it's totally fine to consider those packets lost and move on. 
If you're doing address validation and the remote side cannot prove that it actually owns that address, we need to not send it anything more. 

In practice, this is determined by what role you're playing (are you the one changing addresses or are you the one seeing the remote endpoint change).

nibanks: Let's imagine a client is on path A and probes path B with a challenge. The server receives this challenge and then responds with the response and a challenge of its own. Then, for some reason, that packet never makes it (too much loss, blocked in that direction, or client got disconnected). The both sides won't receive their challenge's response. The current spec text indicates both sides would immediately kill the connection on path A now. I am just questioning if that is the best course of action.

mikkelfj:
I must be missing something fundamental but ...

What is the real gain of a path challenge compared to just sending a packet on the desired path with an empty ping on a new path and see if an ACK gets through? Is to to enforce the ACK (aka path response) to follow the same path, and if so, I don't see any requirement that the path response is sent on a particular path?

Why must must the challenge be hard to guess if it happens on an encrypted channel? It would seem that the packet number with an auth tag would be sufficient, unless challenges can happen early on stream 0, which I don't think is the intention.

Is there a risk that the recipient is not cooperating, and hence the need for something unguessable? This then leans up against optimitistic ACK attacks.

Could this challenge thing be integrated with PMTU probing, considering that you need that knowledge to use the path, of course starting with the minimum size probe?

nibanks:
@mikkelfj As I understand it, the challenge response doesn't have to be on a particular path. The reason it needs to be hard to guess is because you are doing source address validation. You are trying to make sure the client isn't trying to dump traffic on some unsuspecting IP. If the client can guess the challenge and reply with the appropriate response on its own path, it could make the server think it does own that other IP, and start sending traffic to it.

mikkelfj:
OK, so the fundamental thing I was missing (which dawned after hitting return of course), is that it is not necessarily the client proping for a new path, but the server not trusting the peer to be on the path that it claims to be. So the sending a challenge is already following a return path.

@nibanks explained this while I was writing the above.

draft-ietf-quic-transport.md
erickinnear: This should probably be clarified in the text: I read this as only the server would be performing address validation for this example (since the client isn't seeing the server's address change) and therefore would be the only one required to kill the connection if the client cannot demonstrate that it owns the new address. 

In the scenario you mentioned, the client should be sending additional PATH_CHALLENGES when it doesn't get the response (and I think it's possible and probably a good idea for a server implementation to send a new one as well instead of waiting for the client to retransmit, if it wishes).

I agree with your general point, however -- I think we should clarify that the server must not send additional data to the new client's address rather than closing the connection as a whole. Otherwise I can trivially get the server to close your connection just by failing address validation on your behalf.

MikeBishop: I think the language should be around not considering that path viable.  If that was your only path, that means you'll have to kill the connection, yes.  If you think you still have the old path but you're wrong, you'll discover that via timeout soon enough.

draft-ietf-quic-transport.md
```diff
@@ -1562,25 +1563,32 @@ consider the possibility that packets are sent without congestion feedback.
...
-implementations are cautioned that the new path could have a longer round trip
-time than the original.  The endpoint MUST NOT send a CONNECTION_CLOSE frame in
-this case; it has to assume that the remote peer does not want to receive any
-more packets.
+recovering from possible loss of packets carrying PATH_CHALLENGE and
+PATH_RESPONSE frames, the endpoint MUST terminate the connection.  When setting
+this timer, implementations are cautioned that the new path could have a longer
+round trip time than the original.  The endpoint MUST NOT send a
+CONNECTION_CLOSE frame in this case; it has to assume that the remote peer does
+not want to receive any more packets.
```
MikeBishop: I'd question this piece, too.  Rather than "does not want," the implementation should be assuming that the other side *cannot* receive any more packets on the path being probed.  If you have another path, sending a CONNECTION_CLOSE or even maintaining the connection seems quite reasonable.

draft-ietf-quic-transport.md
```diff
@@ -1619,10 +1627,10 @@ the legitimate copy will be dropped as a duplicate.
...
 After a spurious migration, validation of the source address will fail because
 the entity at the source address does not have the necessary cryptographic keys
-to read or respond to the PING frame that is sent to it, even if it wanted to.
-Such a spurious connection migration could result in the connection being
-dropped when the source address validation fails.  This grants an attacker the
-ability to terminate the connection.
+to read or respond to the PATH_CHALLENGE frame that is sent to it, even if it
+wanted to.  Such a spurious connection migration could result in the connection
+being dropped when the source address validation fails.  This grants an attacker
+the ability to terminate the connection.
```
MikeBishop: How about a recommendation to probe both old and new addresses to guard against this?

draft-ietf-quic-transport.md
MikeBishop: We've already done one reorder of the document to match the table -- we'll probably need to do another as part of the grand editorial pass being discussed in a few months.

draft-ietf-quic-transport.md
MikeBishop: The point about PMTU is that you can send multiple PATH_CHALLENGEs on the new path in packets of different sizes and see which ones get through.  However, I'd recommend putting that in a section on probing for PMTU than making an offhand reference here.

draft-ietf-quic-transport.md
MikeBishop: The previous text says that the PATH_RESPONSE can be sent on any path, so currently we don't.

draft-ietf-quic-transport.md
MikeBishop: I think we should always acknowledge all packets.  I wouldn't be upset about including PATH_CHALLENGE/PATH_RESPONSE along with ACK in the context of "packets containing only ACKs don't trigger the sending of a ACK immediately, but are included the next time an ACK is generated."

draft-ietf-quic-transport.md
MikeBishop: Not minor -- you can create chaos doing this.  Ask me how I know.  😉

draft-ietf-quic-transport.md
MikeBishop: Defining the error code doesn't create a mandate to enforce and send it.  Implementations that know they have accounting for every PATH_CHALLENGE they've ever sent (e.g. because they've never sent any) can enforce this, while implementations which have discarded state can let an unknown PATH_RESPONSE pass without comment.

draft-ietf-quic-transport.md
janaiyengar: I'll change the text, but the point here was that an ACK frame is basically useless for acknowledging receipt of a PATH_CHALLENGE frame. The entire point of this frame is to require an echo of the random bytes from the peer... if an ACK were adequate, a PING frame is enough. I don't think it matters whether the peer acks the frame or not, since the endpoint shouldn't use the received ACK as an acknowledgment. This problem exists with PING+Data as well, it just isn't stated. 

I'll change this text.

draft-ietf-quic-transport.md
janaiyengar: This text is the same as previous, with PING+Data. I intend to change this in the PR that addresses other migration concerns, but not here.

draft-ietf-quic-transport.md
janaiyengar: The strategy that I'd like to suggest is to go back to the last best path. That said, this text is the same as it was previously with PING+Data and I'd like to not change migration behavior. I'd like to change this behavior but in a different PR that fixes migration. This PR simply replaces PING+Data / PONG with new frames.

draft-ietf-quic-transport.md
janaiyengar: Agreed, again not changing text here for what the draft currently states, since I'd like to simply get rid of PING+Data for now. This will change with fixing the rest of migration.

draft-ietf-quic-transport.md
janaiyengar: Done

draft-ietf-quic-transport.md
janaiyengar: Removed.

draft-ietf-quic-transport.md
janaiyengar: We should require PATH_RESPONSE to be sent to the same address as the PATH_CHALLENGE, though I'm open to arguments as to why not. That said, I'm not fixing migration bugs in this PR, just changing the PING+Data frame.

draft-ietf-quic-transport.md
janaiyengar: yup, this would not be happy. fixed.

draft-ietf-quic-transport.md
janaiyengar: Yes, good point. Filed #1091.

draft-ietf-quic-transport.md
martinthomson: You can still do stateless in 8 bytes, I guess.  It's just a weaker authenticator.  (I think that stateless is a good strategy actually.)

draft-ietf-quic-transport.md
martinthomson: Yes, let's just keep this constrained to replacement as much as possible.

draft-ietf-quic-transport.md
martinthomson: I think that it would be better to keep the special handling to a minimum.  It is easier to generate a PATH_RESPONSE and send on any path than require use of the same path.  That limits the special logic to the entity that is probing.  I'm happy to keep this simple - you ACK these frames as normal and you generate the response where ever and whenever it is easiest to do so.

draft-ietf-quic-transport.md
```diff
@@ -2593,6 +2564,52 @@ by a client in protected packets, because it is certain that the server is able
...
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+
+~~~
+
+Data:
+
+: This 8-byte field contains arbitrary data.
+
+A PATH_CHALLENGE frame containing at least 8 randomly generated {{?RFC4086}}
```
martinthomson: I think that we're making this too hard.  Can we just say hard to guess?  How they are generated can be left to the challenger.  (If TCP is our benchmark, then 32 bits of entropy is probably enough.)

janaiyengar: SGTM. I kept futzing around with saying the right thing, and it's probably adequate to say "hard to guess". Done.

draft-ietf-quic-transport.md
janaiyengar: Again, not for this PR, but the idea would be to demonstrate two way reachability using PATH_CHALLENGE and PATH_RESPONSE. Otherwise, you may end up in a situation where, for instance, UDP packets are blocked on one direction, but migration happens anyways since PATH_RESPONSE was sent from an address which didn't block UDP packets.

draft-ietf-quic-transport.md
```diff
@@ -1562,25 +1563,31 @@ consider the possibility that packets are sent without congestion feedback.
...
+is received, the address is considered to be valid.  The PATH_RESPONSE frame can
+use any path on its return.
+
+An endpoint MAY send multiple PATH_CHALLENGE frames to handle packet loss or to
+make additional measurements on a new network path.
+
+An endpoint MUST use fresh random data in every PATH_CHALLENGE frame so that it
+can associate the peer's response with the causative PATH_CHALLENGE.  Using
+fresh values allows the endpoint to measure the time taken to respond to a
+challenge.
```
ianswett: This gets a bit weird if the PATH_RESPONSE doesn't come back via the new path.

draft-ietf-quic-transport.md
ianswett: I would change this to "cannot" from "does not want to" as Mike suggested.

draft-ietf-quic-transport.md
ianswett: I think Jana's planning to improve this in a follow up, so I'm ok with it as is.

draft-ietf-quic-transport.md
```diff
@@ -2593,6 +2564,52 @@ by a client in protected packets, because it is certain that the server is able
...
+
+: This 8-byte field contains arbitrary data.
+
+A PATH_CHALLENGE frame containing 8 octets that are hard to guess is sufficient
+to ensure that it is easier to receive the packet than it is to guess the value
+correctly.
+
+The recipient of this frame MUST generate a PATH_RESPONSE frame
+({{frame-path-response}}) containing the same Data.  An endpoint that receives a
+PATH_CHALLENGE frame containing an empty payload MUST generate a connection
```
ianswett: There's no length, so I don't believe you can have an empty PATH_CHALLENGE frame

draft-ietf-quic-transport.md
```diff
@@ -2593,6 +2564,52 @@ by a client in protected packets, because it is certain that the server is able
...
+error of type FRAME_ERROR, indicating the PATH_CHALLENGE frame (that is, 0x0e).
+
+
+## PATH_RESPONSE Frame {#frame-path-response}
+
+The PATH_RESPONSE frame (type=0x0f) is sent in response to a PATH_CHALLENGE
+frame.  Its format is identical to the PATH_CHALLENGE frame
+({{frame-path-challenge}}).
+
+An endpoint that receives a PATH_RESPONSE frame containing an empty payload MUST
```
ianswett: Same comment: It can no longer be empty

draft-ietf-quic-transport.md
```diff
@@ -2692,10 +2709,19 @@ implementation decision, and an implementation should be careful to delay
 conservatively, since any delay is likely to increase application-visible
 latency.
 
-Regular QUIC packets are "containers" of frames; a packet is never retransmitted
-whole.  How an endpoint handles the loss of the frame depends on the type of the
-frame.  Some frames are simply retransmitted, some have their contents moved to
-new frames, and others are never retransmitted.
+Regular QUIC packets are "containers" of frames.  When an endpoint receives an
+ACK frame for one or more transmitted packets, all frames in the acknowledged
+packets are considered to have been delivered to the peer, with one exception.
```
ianswett: How about "received and processed by" instead of "delivered to", since processing is as important as receipt

draft-ietf-quic-transport.md
```diff
@@ -2692,10 +2709,19 @@ implementation decision, and an implementation should be careful to delay
...
+A PATH_CHALLENGE frame (see {{frame-path-challenge}}) is used to validate a
```
ianswett: I would prefer to leave this text out because it's more confusing than helpful.  It's ok to consider it delivered to the peer when it's acknowledged, the key is that one must receive a PATH_RESPONSE before the path has been verified, which you say above.

draft-ietf-quic-transport.md
```diff
@@ -1562,25 +1563,31 @@ consider the possibility that packets are sent without congestion feedback.
...
-return.  A PING frame containing 12 randomly generated {{?RFC4086}} octets is
-sufficient to ensure that it is easier to receive the packet than it is to guess
-the value correctly.
-
-If the PING frame is determined to be lost, a new PING frame SHOULD be
-generated.  This PING frame MUST include a new Data field that is similarly
-difficult to guess.
+endpoint validates a remote address by sending a PATH_CHALLENGE frame containing
+a payload that is hard to guess.  This frame MUST be sent in a packet that is
+sent to the new address.  Once a PATH_RESPONSE frame containing the same payload
```
ianswett: How about "The new address is not validated until a PATH_RESPONSE frame containing the same payload is received, even if the packet containing the PATH_RESPONSE frame is acknowledged."  (and then remove the text below I think is a bit confusing)

janaiyengar: Added. Which text is weird? The text about measuring time taken? It's caused me enough trouble so I'll remove it.

draft-ietf-quic-transport.md
janaiyengar: Yup, and validation gets weird when PATH_RESPONSE is sent via any path too. Again, I don't want to fix the connection migration mechanism here, so leaving it as is. But removed text about measuring time taken.

draft-ietf-quic-transport.md
janaiyengar: It depends on what the semantics of delivery are. What I say above is that an ACK implies that the frame has been received and processed by the peer. I don't think that's a safe assumption with PATH_CHALLENGE, since the entire point of it is to get something besides an ACK.

draft-ietf-quic-transport.md
janaiyengar: Done.

draft-ietf-quic-transport.md
janaiyengar: Yup. Happy to take this comment in the follow-up.

draft-ietf-quic-transport.md
janaiyengar: Good point. Removed the text, since it seems redundant now.

draft-ietf-quic-transport.md
janaiyengar: Done.

draft-ietf-quic-transport.md
janaiyengar: Done.

draft-ietf-quic-transport.md
ianswett: I can't find my comment, but I thought it was weird to take an RTT sample of one path in one direction and a different path in another, which is what happens if PATH_RESPONSE comes back on a different path from the one PATH_CHALLENGE is sent on.

draft-ietf-quic-transport.md
```diff
@@ -2692,10 +2708,19 @@ implementation decision, and an implementation should be careful to delay
...
-frame.  Some frames are simply retransmitted, some have their contents moved to
-new frames, and others are never retransmitted.
+Regular QUIC packets are "containers" of frames.  When an endpoint receives an
+ACK frame for one or more transmitted packets, all frames in the acknowledged
+packets are considered to have been received and processed by the peer, with one
+exception.  A PATH_CHALLENGE frame (see {{frame-path-challenge}}) is used to
+validate a peer's ownership of its address.  An ACK frame received for a
+PATH_CHALLENGE frame is not adequate to indicate that the PATH_CHALLENGE was in
+fact received.  A PATH_CHALLENGE is considered acknowledged only when the
+corresponding PATH_RESPONSE (see {{frame-path-response}}) is received for it.
```
ianswett: I still think this new text is more confusing than helpful.  You're going to have to write some new text that's more complete than this for connection migration, so I would keep this section as is for now.

ianswett: as is = as it was

janaiyengar: Removed offending text. Many things need fixing in migration.

janaiyengar:
I'm going to merge this PR, please feel free to comment on the issue if you want to see any further changes.

Issue #821, opened 2017-10-03
Require connection ID always


jpinner-lyft:
Intermediaries in a QUIC connection do not (and cannot) participate in transport parameter negotiation and endpoints in a connection may not be able to configure or even have knowledge of any intermediaries. Nonetheless, those intermediaries may require connection id to do proper address translation or connection management but have no way to signal to the endpoints that the `omit_connection_id` parameter must not be set.

In particular intermediaries (especially those acting on behalf of, or in concert with, the client and server -- e.g. L4 load-balancers, client gateways, configured proxies) that attempt to manage routing through address migration or NAT re-bindings would require connection ids in both directions.

Further, some aspects of client privacy are currently assumed to be achieved by clients not sending a connection id but the specification does not require this. And some aspects of connection id migration require the servers to send the client a new connection id preemptively which they are not currently required to do. Simply requiring connection ids without requiring the servers to provide clients with the ability to migrate them does not seem to provide sufficient privacy.

My suggestion here would be to modify the specification to always require connection ids be sent by both endpoints and further require endpoints to exchange the necessary information that allows connection id migration to be initiated at will by the client.

ianswett:
The original motivation for not including a connection id was saving bytes on the wire, not privacy.

Given both endpoints have to agree to omit it, and omitting it doesn't have a negative privacy impact, I don't see a negative of allowing it to be omitted.

jpinner-lyft:
This assumes that the endpoints have control and knowledge of the entire path to know that it's safe to omit.

janaiyengar:
Any intermediary has to be working with teh server or the client, since the keys aren't available to an intermediary otherwise. Isn't it adequate to have the end that teh intermediary is in cahoots with not send omit_connection_id? Meaning, if this is a reverse proxy, then the server *knows* to not send omit_connection_id, and itself always sends the connection ID. Why is that not enough?

jpinner-lyft:
@janaiyengar: Intermediaries do not need to be able to decrypt the packet to access the connection id and perform tasks like routing or NAT'ing. There are also many examples of routing intermediaries that are not under direct control of the client or the server (things like Amazon's ELB, WLAN access points, etc). In addition, with the ability to migrate the connection across different paths, the presence of such intermediaries may change throughout the course of the connection.

PR #1088, opened 2018-01-31
A 17 octet connection ID

### LINKED ISSUE #1052: Increase the size of connection ID
@vasilvv raised the question on the list.

64 bits seems like a lot until you start thinking about using cryptography, adding structured information to the identifier.  When you consider collision probabilities, those 64 bits seem a little tight.  Expanding connection ID to 128 bits, which might allow a little more space and enable simple encryption strategies.

On the other hand, a 64-bit overhead on every packet is already pretty costly.  We spent a lot of effort on shaving other parts of the protocol overhead down, sometimes at a fair expense in complexity.  Doubling the size would put the overhead equivalent to our AEAD.

martinthomson:
This comes out of the discussion about what it takes to construct a sensible
connection ID for NEW_CONNECTION_ID.  Trying to cram sufficient routing state
into a 64-bit identifier is tricky, especially since that information needs to
be encrypted and probably also authenticated (at least a little).  Using a
block cipher would be ideal, but all the 64-bit ciphers are pretty bad, whereas
all the 128-bit ones are both faster and more available.

The extra 8 bits is for accounting.  If you want to manage key rotation, a key
identifier is helpful.  Having extra space allows for that to be added without
forcing the block cipher to protect less than 128 octets, which can be hard.

The way that packet overheads are managed are different for long and short
headers.  In long headers, the overheads are much higher, but we don't send
many of those, so it's not a big deal.  For short headers, each endpoint sends
a truncate_connection_id transport parameter which sets the length of the
connection ID that will be used in short headers.  Each endpoint sets their own
limit and their peer has to follow this limit.  Anywhere from 0 (no connection
ID) to 17 (all of the connection ID) can be set.

This also closes the discussion about whether connection ID is a number or not.
With truncation, treating it as a number no longer makes sense.

Note that this change shows the corresponding changes to the invariants draft
on the assumption that it will need to be changed at the same time.  We can take
this before adoption, but it will need to be rebased if we do.

Closes #1052, #821, #834, #833, #881.


mcmanus:
thanks. I think this accurately reflects which way the wind was blowing in Melbourne. The most important property to me, which this reflects, is that the connID is actually a fixed size which might be encoded in shorthand in headers but is still clearly 17 bytes in other contexts.

A couple thoughts: It would be helpful to be explicit about which of the 17 are omitted when being truncated (i.e. the last or low bytes). truncate is probably technically sufficient but explicit is good.

In previous drafts the default was don't omit cid, and now it is do omit cid (i.e. truncate @0).. given that not all environments are robust without some cid, I think our default should be something other than omit completely.

mikkelfj:
>  It would be helpful to be explicit about which of the 17 are omitted

Yes I think so.

I was thinking perhaps it would be best to have the CID length exposed. Consider a HA-Proxy that does stateful routing. In a future version where it understood QUIC it could be configured to expect a certain CID length so it is not necessary to have it public. But it does complicate configuration and deployment.

Another concern is that a 17 byte CID needs to be hashed if used for state lookup at the client. If there were som assurances of collision resistance on the early bytes it would be simpler to manage.

draft-ietf-quic-transport.md
```diff
@@ -454,30 +456,22 @@ Header Form:
...
 
-: The third bit (0x20) of octet 0 indicates the key phase, which allows a
+: The second bit (0x40) of octet 0 indicates the key phase, which allows a
   recipient of a packet to identify the packet protection keys that are used to
   protect the packet.  See {{QUIC-TLS}} for details.
 
 Short Packet Type:
 
-: The remaining 5 bits of octet 0 include one of 32 packet types.
+: The remaining 6 bits of octet 0 include one of 32 packet types.
```
MikeBishop: If you have one more bit, you now have 64 packet types.

draft-ietf-quic-transport.md
```diff
@@ -494,15 +488,15 @@ other fields.
...
+| 0x3E | 2 octets           |
+| 0x3D | 4 octets           |
 {: #short-packet-types title="Short Header Packet Types"}
 
-The header form, omit connection ID flag, and connection ID of a short header
-packet are version-independent.  The remaining fields are specific to the
-selected QUIC version.  See {{version-specific}} for details on how packets
-from different versions of QUIC are interpreted.
+The header form and connection ID of a short header packet are
+version-independent.  The remaining fields are specific to the selected QUIC
```
MikeBishop: This makes the Connection ID in the short header a difficult invariant.  There might be one, but there's no way to identify how long it is.  Middleboxes that want to use the CID for routing, unless they're in active collaboration with the server, are kind of stuck.  Which might be intentional, but it seems awkward.

draft-ietf-quic-transport.md
```diff
@@ -682,11 +682,11 @@ sequence of frames, as described in {{frames}}.
 
 ## Connection ID {#connection-id}
 
-QUIC connections are identified by their 64-bit Connection ID.  All long headers
-contain a Connection ID.  Short headers indicate the presence of a Connection ID
-using the Omit Connection ID flag.  When present, the Connection ID is in the
-same location in all packet headers, making it straightforward for middleboxes,
-such as load balancers, to locate and use it.
+QUIC connections are identified by their 136-bit Connection ID.  All long
```
MikeBishop: You lower-cased all occurrences of "Connection ID" in this paragraph except this one.

draft-ietf-quic-transport.md
```diff
@@ -682,11 +682,11 @@ sequence of frames, as described in {{frames}}.
...
+headers contain a connection ID.  The connection ID is truncated in the short
```
MikeBishop: Since 17 bytes isn't truncated, perhaps "can be" truncated?

draft-ietf-quic-transport.md
```diff
@@ -682,11 +682,11 @@ sequence of frames, as described in {{frames}}.
...
+header, using between 0 and 17 octets.  When present, the connection ID is in
+the same location in all packet headers, making it straightforward for
+middleboxes, such as load balancers, to locate and use it.
```
MikeBishop: Except it's not straightforward; see above.

draft-ietf-quic-transport.md
```diff
@@ -703,6 +703,22 @@ use it for all subsequent Handshake ({{packet-handshake}}) and 1-RTT
...
+by setting truncate_connection_id to 0, or include any number of octets up to
+the full connection ID length of 17 octets.  Connection IDs are not truncated
+for any other use.
+
+Note:
+
+: The odd length of the connection ID is chosen to enable efficient construction
+  of the value using 128-bit ciphers with additional space for overhead.
+  Endpoints might negotiate the use of a truncated connection ID for use with
+  the short header.
```
MikeBishop: The second sentence of the note seems pointless, given that you've just talked about this and it's unrelated to the first sentence in the note.

draft-ietf-quic-transport.md
```diff
@@ -778,9 +794,9 @@ constant:
 
 * the location of the header form flag,
 
-* the location of the Omit Connection ID flag in short headers,
+* the location and size of the Connection ID field in long headers,
 
-* the location and size of the Connection ID field in both header forms,
+* the location of the Connection ID field in short headers,
```
MikeBishop: Perhaps worth re-emphasizing that the size is not invariant?

draft-ietf-quic-transport.md
```diff
@@ -1462,13 +1479,13 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
 ## Connection Migration {#migration}
 
-QUIC connections are identified by their 64-bit Connection ID.  QUIC's
-consistent connection ID allows connections to survive changes to the client's
-IP and/or port, such as those caused by client or server migrating to a new
-network.  Connection migration allows a client to retain any shared state with a
-connection when they move networks.  This includes state that can be hard to
-recover such as outstanding requests, which might otherwise be lost with no easy
-way to retry them.
+QUIC connections are identified by their 136-bit Connection ID or whatever part
```
MikeBishop: Along the lines of being explicit about which side we're truncating, maybe call this a "prefix" rather than a "part"?

draft-ietf-quic-transport.md
```diff
@@ -1890,6 +1897,46 @@ different static key or server identifier.
...
+requested be included in the short header.
+
+A server is expected to have a minimum number of octets that are needed for
+correct routing.  Thus, the server can always know that at least this many
+octets of the packet it receives contain a connection ID.
+
+If the server uses a longer connection ID than the client, there is the
+potential for the additional octets to be used to reveal that the stateless
+reset is not a regular packet.  This is not a serious issue, as it is already
+possible to use other fields to distinguish them from other packet.
```
MikeBishop: Really?  I thought we'd been able to stick to that goal so far.

draft-ietf-quic-transport.md
```diff
@@ -1890,6 +1897,46 @@ different static key or server identifier.
...
+
+If a client requires more octets of the connection ID for routing than the
+server requires, those octets will not appear in the packet that the server
+receives.  In this case, the stateless reset might not reach the client.
+
+A server can avoid this by ensuring that it always receives a connection ID that
+is equal to or longer than the connection ID requested by the client.  In
+choosing a connection ID, the server already knows the client's preferred
+transaction ID length, and so can request that the client provide a similarly
+long connection ID.
```
MikeBishop: This is probably the right choice, but again departs from our current semantics that servers don't have to look at the client's transport params before generating their own.

draft-ietf-quic-transport.md
```diff
@@ -1890,6 +1897,46 @@ different static key or server identifier.
...
+
+A server can construct the connection ID such that it includes a record the
```
MikeBishop: "a record of", or "an indicator of"

draft-ietf-quic-transport.md
martinthomson: I guess we have.  Assuming that we do packet number encryption.  Right now the packet number leaks quite a lot of information.

draft-ietf-quic-transport.md
martinthomson: Yeah, but this is still a degree shy of making it a negotiation.

draft-ietf-quic-transport.md
martinthomson: Removed that bit.  (I don't like making value judgments in specifications anyway.)

draft-ietf-quic-transport.md
martinthomson: I think that the purpose of the connection ID is for cooperating intermediaries, not random middleboxen.  I'm sure that others will disagree, as we're seeing with the proposal to encrypt packet numbers.

martinthomson:
> In previous drafts the default was don't omit cid, and now it is do omit cid (i.e. truncate @0).. given that not all environments are robust without some cid, I think our default should be something other than omit completely.

There are really only two values you can default to without making some sort of judgment about what sort of environment gets preferential treatment: 0 or 17.  I don't think that 17 is a great choice.  I would rather force the inclusion of the transport parameter in the common case than decide that 8 is the best number.

draft-ietf-quic-transport.md
```diff
@@ -682,11 +682,10 @@ sequence of frames, as described in {{frames}}.
 
 ## Connection ID {#connection-id}
 
-QUIC connections are identified by their 136-bit Connection ID.  All long
-headers contain a connection ID.  The connection ID is truncated in the short
+QUIC connections are identified by their 136-bit connection ID.  All long
+headers contain a connection ID.  The connection ID can truncated in the short
```
MikeBishop: can => can be

draft-ietf-quic-transport.md
```diff
@@ -890,18 +890,19 @@ Packets that can be associated with an existing connection are handled according
...
-ports.
-
-A packet that uses the short header could be associated with an existing
-connection with an incomplete cryptographic handshake.  Such a packet could be a
-valid packet that has been reordered with respect to the long-form packets that
-will complete the cryptographic handshake.  This might happen after the final
-set of cryptographic handshake messages from either peer.  These packets are
-expected to be correlated with a connection using the tuple of IP addresses and
-ports.  Packets that might be reordered in this fashion SHOULD be buffered in
+For connections with incomplete cryptographic handshakes packets with the long
```
MikeBishop: Probably want a comma after "handshakes"

mcmanus:
>  I would rather force the inclusion of the transport parameter in the common case 

seems reasonable

ekr:
I don't think this is the right answer. As I mentioned in Melbourne, I believe a better approach is to have the server provide a complete connection ID of variable length and then each side provides exactly that connection ID (or in the case of the server, potentially none at all). Aside from aesthetic considerations, one quite silly side effect of this design is that the client tells the server how to truncate the connection ID without knowing what the server's truncation is, so it's not at all impossible for the server to send you a connection ID where only 8 bytes are meaningful but the client doesn't know that and asks for 12 bytes.

In terms of process, I don't think that this has consensus and I read the minutes as pretty clearly saying that we were going to form a design team to come up with some proposals, and then the WG was going to decide, so this PR seems premature.

draft-ietf-quic-transport.md
martinthomson: That change is integral to the PR.  The bit only makes sense if you intend to send the connection ID on some packets but not others and if you think that an intermediary that is ignorant of the actual length being able to do something meaningful with both sets of packets.

draft-ietf-quic-transport.md
martinthomson: The construction of a stateless reset requires that the server provide a packet number.  It is probably using a connection ID that it used in the past, so packets can be correlated with previous packets.  The difference in packet numbers is observable.  A server has to pick a random packet number, which likely means a large gap.  That might be used to distinguish a stateless reset.

Packet number encryption fixes that problem.  I think that this will change if we decide to do that.

britram:
I share both of @ekr's concerns here.

draft-ietf-quic-transport.md
janaiyengar: Fair enough. I don't grok your point about using a connection ID from the past. 

This is another way that a stateless reset might be detectable: while sending a stateless reset the server does not know what the client had negotiated for connection ID truncation in the server -> client direction. If the client required the server to send a connection ID, the stateless reset would likely not carry one.

draft-ietf-quic-transport.md
janaiyengar: That's not the scenario I outlined in my comment. You run a server farm and purchase a middlebox from a vendor. If you have a connection ID bit, you don't have to configure the middlebox to align with server policy about connection ID, because packets are self-describing.

The bit makes sense only if the size is known to both the server and the middlebox however, so you have a good point. Both entities have to agree on a way to determine the size of the connection ID, and as we've discussed earlier, that can be achieved by making the connection ID a varlen. The spec doesn't have to make it a varlen. A middlebox can simply require that connection IDs are vended with a particular encoding by servers, and that's adequate.

I don't want to design this here; I'm noting that self-describing packets have value. I think this should be discussed separately from this PR. That said

mikkelfj:
> I share both of @ekr's concerns here.

Also note my proposal for asymmetric CID which would resolve some of those concerns. I think it would require a separate initial packet type when responding to a stateless retry, but then CID and CID length and in each direction can be place in transport params.
https://github.com/quicwg/base-drafts/issues/1089

draft-ietf-quic-transport.md
```diff
@@ -454,30 +456,22 @@ Header Form:
...
+: The remaining 6 bits of octet 0 include one of 64 packet types.
   {{short-packet-types}} lists the types that are defined for short packets.
 
 Connection ID:
 
-: If the Omit Connection ID Flag is not set, a connection ID occupies octets
-  1 through 8 of the packet.  See {{connection-id}} for more details.
+: A connection ID occupies between 0 and 17 octets inclusive, starting from
+  octet 1.  This value can be truncated from the full 17 octet connection ID.
+  See {{connection-id}} for more details.
```
huitema: So, if I read correctly, there is no indication in the short header of the presence and size of the connection ID field. In practice, this forces clients and servers to pick one same size for all connections. Yes, in theory it can be negotiated on a per connection basis, but you need to parse the CID to retrieve the connection context and find the result of that negotiation. Better not try to change the encoded length over the lifetime of the server...

draft-ietf-quic-transport.md
```diff
@@ -703,6 +702,23 @@ use it for all subsequent Handshake ({{packet-handshake}}) and 1-RTT
...
+truncate_connection_id to 0, or include any number of octets up to the full
+connection ID length of 17 octets.  Connection IDs are not truncated for any
+other use.
+
+Note:
+
+: The odd length of the connection ID is chosen to enable efficient construction
+  of the value using 128-bit ciphers with an additional octet for overheads such
+  as key identifiers.
+
```
huitema: OK. Now, in the direction from server to client, are we sending the server chosen CID, or the client chosen one? Because, the client has no way to know how many bytes of the server connection ID it might need before seeing the CID. Yet the client's transport parameters are sent in the client initial packet, before seeing the value chosen by the server. So the clients only practical choices are 0 and 17...

draft-ietf-quic-transport.md
martinthomson: A server has to pick a consistent minimum.  It could occasionally choose longer, but the load balancer would just ignore those extra bytes.

draft-ietf-quic-transport.md
martinthomson: Yes, there is general recognition that a client can't use stateless routing.  For a client, you are right.  This is why I would prefer some symmetric in connection ID selection, but that means making a hard call with respect to the stateless reset.

draft-ietf-quic-transport.md
```diff
@@ -1187,15 +1205,15 @@ initial_max_stream_id_uni (0x0008):
...
-  that source and destination IP address and port are sufficient for it to
-  identify a connection.  This parameter is zero length.  Absence of this
-  parameter means that the connection ID MUST be present in every packet sent to
-  this endpoint.
+: An 8-bit unsigned integer that indicates the length of the connection ID that
+  is to be put in packets that use the short header ({{short-header}}) that are
+  sent to this endpoint.  Values between 0 and 17 inclusive are valid; larger
+  values MUST be treated as a connection error of type
+  TRANSPORT_PARAMETER_ERROR.  This parameter is set to 0 if the transport
+  parameter is omitted, meaning that packets with short headers that are sent to
```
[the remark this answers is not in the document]
martinthomson: I just wanted to say that it defaults to 0.  I'll revise.

draft-ietf-quic-transport.md
martinthomson: My point is that if there were packets sent on this connection in the past, then the packet numbers there and in the stateless reset might reveal something.

The point about truncation is addressed in this PR (3 paragraphs down).  It's ugly though.

draft-ietf-quic-tls.md
```diff
@@ -552,7 +552,7 @@ older than 1.3 is negotiated.
 
 QUIC requires that the initial handshake packet from a client fit within the
 payload of a single packet.  The size limits on QUIC packets mean that a record
-containing a ClientHello needs to fit within 1171 octets.
+containing a ClientHello needs to fit within 1162 octets.
```
huitema: That's not the only change required to the TLS document. The connection ID is also used as in the creation of Handshake Secrets. Pointing out the change in length may not be strictly necessary, but it will be helpful for implementers.

draft-ietf-quic-transport.md
ianswett: In regards to the stateless reset issue for asymmetric connection IDs, one use case is picking a connection ID so the packet is received by the optimal receive queue and doesn't have to hop threads/CPUs.  For use cases like this, it's an optimization, but the machine should be able to route the packet correctly if it chooses to.

Also, I'll note that right now, the client doesn't get to pick any portion of the connection ID, so allowing the client to pick knowing that stateless reset might not route optimally seems acceptable, since the client can always pick the same value as the server, like today.

martinthomson:
Parking this.  We will need something to capture the design team conclusions, which included two connection IDs.

PR #1090, opened 2018-02-01
Mitigate amplification attacks during handshake

### LINKED ISSUE #38: First flight size restriction
... relative to the client's ClientHello.

We don't want to design a packet amplifier, but it's highly likely that the server's Certificate will cause the server's messages to be at least a little bigger than the initial messages from the client.  We do need servers to send HelloRetryRequest and do source address validation, but how often?

MikeBishop:
Lifted out of #1082 and building on #1086, this attempts to mitigate a potential amplification in which an attacker finds a server with a very large certificate, spoofs a source address, and then uses the server's handshake to generate traffic.  It limits the server from sending more than three Handshake packets to an unvalidated source address.  Suggested ways of validating are Retry packets and PATH_CHALLENGE.

Fixes #38.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -651,14 +651,17 @@ its previous packet by one for each Handshake packet that it sends (which might
 be an Initial, 0-RTT Protected, or Handshake packet).
 
 Servers MUST NOT send more than three Handshake packets without receiving a
-packet from a verified source address.  If the server expects to generate more
-than three Handshake packets in response to an Initial packet, it SHOULD include
-a PATH_CHALLENGE frame in each Handshake packet that it sends.  After receiving
-at least one valid PATH_RESPONSE frame in a Handshake packet, the server can
-send its remaining Handshake packets.  (Servers MAY instead perform address
-validation using a Retry packet; this requires less state on the server, but
-could involve additional computational effort depending on implementation
-choices.)
+packet from a verified source address.  Source addresses can be verified
+through an address validation token, receipt of the final cryptographic message
+from the client, or by receiving a valid PATH_RESPONSE frame from the client.
+
+If the server expects to generate more than three Handshake packets in response
+to an Initial packet, it SHOULD include a PATH_CHALLENGE frame in each Handshake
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -650,8 +650,18 @@ described in {{packet-numbers}}.  The client increments the packet number from
 its previous packet by one for each Handshake packet that it sends (which might
 be an Initial, 0-RTT Protected, or Handshake packet).
 
-The payload of this packet contains STREAM frames and could contain PADDING and
-ACK frames.
+Servers MUST NOT send more than three Handshake packets without receiving a
+packet from a verified source address.  If the server expects to generate more
+than three Handshake packets in response to an Initial packet, it SHOULD include
+a PATH_CHALLENGE frame in each Handshake packet that it sends.  After receiving
+at least one valid PATH_RESPONSE frame in a Handshake packet, the server can
```
martinthomson: You don't need to say where the frame is in this sentence.  You do need to change the TLS doc to permit the use of these frames in Handshake packets.

draft-ietf-quic-transport.md
```diff
@@ -650,8 +650,18 @@ described in {{packet-numbers}}.  The client increments the packet number from
...
+send its remaining Handshake packets.  (Servers MAY instead perform address
```
martinthomson: s/MAY/can and remove the parentheses.

draft-ietf-quic-transport.md
```diff
@@ -650,8 +650,18 @@ described in {{packet-numbers}}.  The client increments the packet number from
 its previous packet by one for each Handshake packet that it sends (which might
 be an Initial, 0-RTT Protected, or Handshake packet).
 
-The payload of this packet contains STREAM frames and could contain PADDING and
-ACK frames.
+Servers MUST NOT send more than three Handshake packets without receiving a
+packet from a verified source address.  If the server expects to generate more
```
martinthomson: You should probably explain that the server is only able to verify a client source address after receiving an address validation token or the final cryptographic handshake message from the client (in TLS, we can tell earlier, but that's not a transport-level guarantee).

PR #1097, opened 2018-02-05
Alt-Svc quic= a list

### LINKED ISSUE #1093: hq Alt-Svc quic parameter and structured headers
I think the current syntax of the "quic" parameter for hq Alt-Svc is, after recent changes, incompatible with structured headers. For reference, https://github.com/httpwg/http-extensions/issues/281 provides history of the issue. This ticket highlights that HTTP/QUIC is now incompatible. 

Please close this out if a non-issue.

MikeBishop:
Fixes #1093.

Replaces the multiple occurrences of the quic= parameter with a quoted string containing a comma-delimited list.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -127,7 +127,10 @@ port.
 
 This document defines the "quic" parameter for Alt-Svc, which MAY be used to
 provide version-negotiation hints to HTTP/QUIC clients. QUIC versions are
-four-octet sequences with no additional constraints on format. Syntax:
+four-octet sequences with no additional constraints on format.  Leading zeros
+SHOULD be omitted for brevity.
+
+Syntax:
 
 ~~~ abnf
 quic = DQUOTE version-number [ "," version-number ] * DQUOTE
@@ -138,14 +141,16 @@ For example, if a server supported both version 0x00000001 and the version
 rendered in ASCII as "Q034", it could specify the following header:
 
 ~~~ example
-Alt-Svc: hq=":49288";quic="1,51303334"
+Alt-Svc: hq=":49288";quic="1,dadababa,51303334"
 ~~~
 
 Where multiple versions are listed, the order of the values reflects the
 server's preference (with the first value being the most preferred version).
-Leading zeros SHOULD be omitted for brevity. Origins SHOULD list only versions
-which are supported by the alternative, but MAY omit supported versions for any
-reason.
+Reserved versions MAY be listed, but unreserved versions which are not supported
+by the alternative SHOULD NOT be present in the list. Origins MAY omit supported
+versions for any reason.
+
+Clients MUST ignore any included versions which they do not support.
 
 
 ## Connection Establishment {#connection-establishment}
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -130,23 +130,22 @@ provide version-negotiation hints to HTTP/QUIC clients. QUIC versions are
...
-Leading zeros SHOULD be omitted for brevity.  When multiple versions are
-supported, the "quic" parameter MAY be repeated multiple times in a single
-Alt-Svc entry.  For example, if a server supported both version 0x00000001 and
-the version rendered in ASCII as "Q034", it could specify the following header:
+For example, if a server supported both version 0x00000001 and the version
+rendered in ASCII as "Q034", it could specify the following header:
 
 ~~~ example
-Alt-Svc: hq=":49288";quic=1;quic=51303334
+Alt-Svc: hq=":49288";quic="1,51303334"
```
martinthomson: Unrelated to this, but can you add one of the greasing versions to this list?  We don't have any use of those yet, and it would be good to include something in the a-f range (which that does).

draft-ietf-quic-http.md
```diff
@@ -138,14 +141,16 @@ For example, if a server supported both version 0x00000001 and the version
 rendered in ASCII as "Q034", it could specify the following header:
 
 ~~~ example
-Alt-Svc: hq=":49288";quic="1,51303334"
+Alt-Svc: hq=":49288";quic="1,dadababa,51303334"
```
LPardue: 0xdadababa values is not explained in preceding paragraph.

LPardue:
Just sanity check, the reserved value 0x00000000 would result in something like quic="" or quic="1,", is that a good thing? Perhaps better to prohibit advertising this value?

martinthomson:
@LPardue, that sounds like a feature to me.

MikeBishop:
Should I include a double comma in the list, just to demonstrate?  (Facetious -- I actually think having empty-string as a possible value is unfortunate.)

MikeBishop:
Actually, never mind, this is fine.  `version-number` is at least one digit long, so version zero would be rendered as `0`, and who cares?

LPardue:
Good point. In that case ignore me on 0.

LPardue:
Now that we've gone back to this list format, is it invalid to repeat the the quic parameter in a single Alt-Svc?

draft-ietf-quic-http.md
```diff
@@ -127,27 +127,36 @@ port.
...
-Alt-Svc: hq=":49288";quic=1;quic=51303334
+Alt-Svc: hq=":49288";quic="1,1abadaba,51303334,0"
 ~~~
 
-Where multiple versions are listed, the order of the values reflects the
-server's preference (with the first value being the most preferred version).
-Origins SHOULD list only versions which are supported by the alternative, but
-MAY omit supported versions for any reason.
-
+A client acting on this header would drop the reserved versions (because it does
```
janaiyengar: For clarity, can you add a note first about 1abadaba and 0 being reserved versions, with a ref to Section 4 of the transport draft?

PR #1100, opened 2018-02-06
Extract request cancellation from GOAWAY

### LINKED ISSUE #756: Cancellation of requests with (partial) responses
We should be a lot clearer about what it means to receive HTTP_REQUEST_CANCELLED for a stream where part of the response was received.  What if you receive (in the same packet even) a STREAM frame containing the headers block and a 200 status code, followed by a cancellation of that stream?

This also applies to GOAWAY, which implicitly sends HTTP_REQUEST_CANCELLED for all requests with stream ID > N.

MikeBishop:
Fixes #756.

Currently, all the discussion about CANCELLED requests is bound up in GOAWAY.  While GOAWAY is one way that requests can be CANCELLED, there are others.  This lifts out the text about request cancellation to its own section, then shortens the GOAWAY section to reference it.

I don't think there are new normative requirements here, but there are definitely some requirements that are easier to find now.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -383,12 +383,12 @@ RST_STREAM or STOP_SENDING frames, as appropriate) with an error type of
 HTTP_REQUEST_CANCELLED ({{http-error-codes}}).  When the client cancels a
 request or response, it indicates that the response is no longer of interest.
 
-When the server cancels (HTTP_REQUEST_CANCELLED) either direction of the request
-stream, it indicates that no application processing was performed.  The client
-can treat requests cancelled by the server as though they had never been sent at
-all, thereby allowing them to be retried later on a new connection.  Servers
-MUST NOT use the HTTP_REQUEST_CANCELLED status for requests which were partially
-or fully processed.
+When the server cancels either direction of the request stream using
+HTTP_REQUEST_CANCELLED, it indicates that no application processing was
+performed.  The client can treat requests cancelled by the server as though they
+had never been sent at all, thereby allowing them to be retried later on a new
+connection.  Servers MUST NOT use the HTTP_REQUEST_CANCELLED status for requests
+which were partially or fully processed.
 
   Note:
   : In this context, "processed" means that some data from the stream was
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -383,12 +383,12 @@ RST_STREAM or STOP_SENDING frames, as appropriate) with an error type of
 HTTP_REQUEST_CANCELLED ({{http-error-codes}}).  When the client cancels a
 request or response, it indicates that the response is no longer of interest.
 
-When the server cancels either direction of the request stream, it indicates
-that no application processing was performed.  The client can treat requests
-cancelled by the server as though they had never been sent at all, thereby
-allowing them to be retried later on a new connection.  Servers MUST NOT use the
-HTTP_REQUEST_CANCELLED status for requests which were partially or fully
-processed.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -376,6 +376,31 @@ as a stream error of type HTTP_CONNECT_ERROR ({{http-error-codes}}).
...
 
+### Request Cancellation
+
+Either client or server can cancel requests by closing the stream (QUIC
+RST_STREAM or STOP_SENDING frames, as appropriate) with an error type of
+HTTP_REQUEST_CANCELLED ({{http-error-codes}}).  When the client cancels a
+request or response, it indicates that the response is no longer of interest.
+
+When the server cancels either direction of the request stream, it indicates
+that no application processing was performed.  The client can treat requests
```
martinthomson: I don't think that this can be true.  It might indicate that no processing was performed, but we can't prohibit a server from cancelling streams after it does something, and this implies that requirement.

(This WAS true, because a variation on this text was used in the context of GOAWAY.)

draft-ietf-quic-http.md
```diff
@@ -832,33 +855,19 @@ identified by a QUIC MAX_STREAM_ID frame, and MAY be zero if no requests were
...
-If a stream is cancelled after receiving a complete response, the client MAY
-ignore the cancellation and use the response.  However, if a stream is cancelled
-after receiving a partial response, the response SHOULD NOT be used.
-Automatically retrying such requests is not possible, unless this is otherwise
-permitted (e.g., idempotent actions like GET, PUT, or DELETE).  Requests on
-Stream IDs less than or equal to the Stream ID in the GOAWAY frame might have
-been processed; their status cannot be known until they are completed
+indicated in the GOAWAY frame, those requests are considered cancelled
+({{request-cancellation}}).  Endpoints SHOULD reset any streams above this ID
+with the error code HTTP_REQUEST_CANCELLED.  Servers MAY also cancel requests on
```
martinthomson: s/Endpoints/Both server and client/ just to make that clear.

draft-ietf-quic-http.md
```diff
@@ -832,33 +855,19 @@ identified by a QUIC MAX_STREAM_ID frame, and MAY be zero if no requests were
 processed.  Servers SHOULD NOT increase the MAX_STREAM_ID limit after sending a
 GOAWAY frame.
 
-  Note:
-  : In this context, "processed" means that some data from the stream was
-    passed to some higher layer of software that might have taken some action as
-    a result.
-
-Once sent, the server will refuse requests sent on streams with an identifier
+Once sent, the server MUST cancel requests sent on streams with an identifier
```
martinthomson: This MUST is now inconsistent with the SHOULD below: "Endpoints SHOULD reset [...]".  Pick one.

draft-ietf-quic-http.md
MikeBishop: I think the clarification needed here is that I've defined "cancel" := "RST_STREAM with CANCELLED".  The server can, of course, RST_STREAM with some other error code at any time.  We can, I think, prohibit the server from using a particular error code after it does something.

MikeBishop: (Of course, we still have to spec the client behavior if the server fails in its obligation, which is what the subsequent text is about.)

MikeBishop:
I think the resolution to both the MUST/SHOULD and the clarity of "endpoints" is to say that the server MUST cancel, but the client SHOULD go ahead and cancel as well.

martinthomson:
Yeah, that seems right.

PR #1107, opened 2018-02-12
Frame types that are Protocol Violations

### LINKED ISSUE #1047: Handling of Frames Inappropriate for Unidirectional Streams
Something I can't seem to find in the transport doc: if I have a unidirectional stream, various frame types don't make any sense.

Send-only streams shouldn't see STREAM, STREAM_BLOCKED, or RST_STREAM frames.
Receive-only streams shouldn't see MAX_STREAM_DATA, STOP_SENDING

There are several courses of action here:
- blow up the connection with error code XXXX
- blow up the stream with error code XXXX
- drop the whole containing packet as somehow compromised, and continue
- ignore the frame and continue.

If we blow it up,  I think there's a decent case for STREAM_DIRECTION_ERROR or something similar.

Early responders on the mailing list thought it should be a fatal connection error. I don't have a strong opinion but we should document the correct behavior.

martinduke:
Specify that certain frames on unidirectional streams tear down the connection. STREAM, STREAM_BLOCKED, and RST_STREAM can't be on a send-only, and MAX_STREAM_DATA and STOP_SENDING can't be received on a receive-only.

Fixes #1047

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2656,8 +2656,8 @@ are present in the frame.
   final offset of the stream.  Setting this bit indicates that the frame
   marks the end of the stream.
 
-An endpoint that receives a STREAM frame for a send-only stream MUST terminate the
-connection with error PROTOCOL_VIOLATION.
+An endpoint that receives a STREAM frame for a send-only stream MUST terminate
+the connection with error PROTOCOL_VIOLATION.
 
 A STREAM frame is shown below.
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2103,6 +2103,12 @@ of the maximum amount of data that can be sent on a stream.
 An endpoint that receives a MAX_STREAM_DATA frame for a receive-only stream
 MUST terminate the connection with error PROTOCOL_VIOLATION.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2097,6 +2100,9 @@ change in the initial limits (see {{zerortt-parameters}}).
 The MAX_STREAM_DATA frame (type=0x05) is used in flow control to inform a peer
 of the maximum amount of data that can be sent on a stream.
 
+An endpoint that receives a MAX_STREAM_DATA frame for a receive-only stream
+MUST terminate the connection with error PROTOCOL_VIOLATION.
+
```
mikkelfj: What about when receiving MAX_STREAM_DATA on a send or bidi stream but that has not yet been opened?

martinduke: MAX_STREAM_DATA is defined in the state machine as one way to open a stream.

mikkelfj: In section 10.2.3 it says "The receiver only sends MAX_STREAM_DATA in the “Recv” state. " So I assume it would be a protocol error to send it in other states.

martinduke: Ah, I see now that this is for the receive end of bidirectional streams only. So, yes, you're right.

PR #1112, opened 2018-02-15
qcram: section reference format

reschke:


draft-ietf-quic-qcram.md
```diff
@@ -362,7 +362,7 @@ the number of concurrently blocked streams.
 
 ### Fixed overhead.
 
-HPACK defines overhead as 32 bytes ({{!RFC7541}} Section 4.1).  As described
+HPACK defines overhead as 32 bytes ({{!RFC7541}}, Section 4.1).  As described
```
afrind: Is there a consistent format we should use for references to RFC7541?  Or it's ok that one is 'see Section X.Y of' and the other is ', Section X.Y'?

draft-ietf-quic-qcram.md
```diff
@@ -103,7 +103,7 @@ PUSH_PROMISE frames (see {{hq-frames}}).
 If a header block contains no vulnerable header fields, BLOCKING MUST be 0.
 This implies that the header fields are represented either as references
 to dynamic table entries which are known to have been received, or as
-Literal header fields (see {{RFC7541}} Section 6.2).
+Literal header fields (see Section 6.2 of {{RFC7541}} ).
```
afrind: Is there an extraneous space before ) ?

draft-ietf-quic-qcram.md
martinthomson: I have been using "Section X.Y of {{CITATION}}" always.  The other acceptable form is "{{CITATION}}, Section X.Y", but there is value in consistency.

draft-ietf-quic-qcram.md
reschke: yes

draft-ietf-quic-qcram.md
reschke: ...depends on context. The first form works a bit better in prose, the second one in parentheses. IMHO.

PR #1114, opened 2018-02-15
Simplify QCRAM abstract

LPardue:
A terse strawman proposal.

State what not why, make it more similar to HPACK abstract.

### DIFF
--- a/draft-ietf-quic-qcram.md
+++ b/draft-ietf-quic-qcram.md
@@ -32,8 +32,8 @@ author:
 
 --- abstract
 
-This specification defines QCRAM, a compression format for efficiently 
-representing HTTP header fields, to be used in HTTP over QUIC. This is a 
+This specification defines QCRAM, a compression format for efficiently
+representing HTTP header fields, to be used in HTTP over QUIC. This is a
 variation of HPACK header compression that avoids head-of-line blocking.
 
 --- middle
--- a/draft-ietf-quic-qcram.md
+++ b/draft-ietf-quic-qcram.md
@@ -32,15 +32,9 @@ author:
 
 --- abstract
 
-The design of the core QUIC transport subsumes many HTTP/2 features, prominent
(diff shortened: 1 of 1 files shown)


LPardue:
build failed due to trailing whitespace >.<

Please judge on content.

martinthomson:
I like the content and have fixed the whitespace.

draft-ietf-quic-qcram.md
```diff
@@ -32,15 +32,9 @@ author:
...
-multiplexing free of head-of-line (HoL) blocking between streams. In HTTP/2,
-multiplexed streams can suffer HoL blocking due to TCP.
-
-If HTTP/2's HPACK is used for header compression, HTTP/QUIC is still vulnerable
-to HoL blocking, because of HPACK's assumption of in-order delivery.  This draft
-defines QCRAM, a variation of HPACK and mechanisms in the HTTP/QUIC mapping that
-allow the flexibility to avoid header-compression-induced HoL blocking.
+This specification defines QCRAM, a compression format for efficiently
+representing HTTP header fields, to be used in HTTP over QUIC. This is a
+variation of HPACK header compression that avoids head-of-line blocking.
```
dtikhonov: "Avoids" is an incorrect term to use: we differentiate avoidance and minimization.  In the default mode, QCRAM minimizes head-of-line blocking.

afrind: Agreed.  @dtikhonov, would you prefer 'minimizes' here, or do you think we should elaborate? 

'...that minimizes or avoids...'
'...that seeks to reduce...'

dtikhonov: As far as abstracts go, the current QCRAM abstract is short.  To keep it that way, "seeks to reduce" is probably the best way to go.  Discussion about minimization vs. avoidance does not have to be surfaced to the level of the abstract.

LPardue:
SGTM, I'll incorporate that text into PR when at the computer next.

PR #1120, opened 2018-02-21
use the 0x8 short header bit for demultiplexing gQUIC

marten-seemann:
This PR changes the code points for the short header packet types, such that the 0x8 bit is always unset. This will allow implementations of gQUIC to demultiplex gQUIC packets with short header packets sent by the client (gQUIC clients will always set the 0x8 bit).

For implementers, the only thing that needs to be changed is the short header packet type:
- 0x1F => 0x17
- 0X1E => 0x16
- 0x1D => 0x15

Apparently Github doesn't allow me to request reviews, but probably @ianswett, @martinthomson, @janaiyengar and @csperkins should have a look at this.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -486,8 +486,8 @@ Key Phase Bit:
   recipient of a packet to identify the packet protection keys that are used to
   protect the packet.  See {{QUIC-TLS}} for details.
 
-\[\[RFC editor: please remove the Google QUIC Demultipexing bit before
-publication.]]
+\[\[Editor's Note: this section should be removed and the bit definitions
+changed before this draft goes to the IESG.]]
 
 Google QUIC Demultipexing Bit:
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -489,6 +489,13 @@ Key Phase Bit:
 \[\[Editor's Note: this section should be removed and the bit definitions
 changed before this draft goes to the IESG.]]
 
+Fourth bit:
+
+: The fourth bit (0x10) of octet 0 is set to 1.
+
+\[\[Editor's Note: this section should be removed and the bit definitions
+changed before this draft goes to the IESG.]]
+
 Google QUIC Demultipexing Bit:
 
 : The fifth bit (0x8) of octet 0 is set to 0. This allows implementations of
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -524,9 +524,9 @@ other fields.
 
 | Type | Packet Number Size |
 |:-----|:-------------------|
-| 0x7  | 1 octet            |
-| 0x6  | 2 octets           |
-| 0x5  | 4 octets           |
+| 0x0  | 1 octet            |
+| 0x1  | 2 octets           |
+| 0x2  | 4 octets           |
 {: #short-packet-types title="Short Header Packet Types"}
 
 The header form, omit connection ID flag, and connection ID of a short header
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -493,7 +493,8 @@ Google QUIC Demultipexing Bit:
 
 : The fifth bit (0x8) of octet 0 is set to 0. This allows implementations of
   Google QUIC to distinguish Google QUIC packets from short header packets sent
-  by a client.
+  by a client because Google QUIC servers expect the connection ID to always be
+  present.
   The special interpretation of this bit SHOULD be removed from this
   specification when Google QUIC has finished transitioning to the new header
   format.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -451,7 +451,7 @@ following sections.
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+
-|0|C|K| Type (5)|
+|0|C|K|1|0|T T T|
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                                                               |
 +                     [Connection ID (64)]                      +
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -486,9 +486,18 @@ Key Phase Bit:
   recipient of a packet to identify the packet protection keys that are used to
   protect the packet.  See {{QUIC-TLS}} for details.
 
+Google QUIC Demultipexing Bit:
+
+: The fifth bit (0x8) of octet 0 is set to 0. This allows implementations of
+  Google QUIC to distinguish Google QUIC packets from short header packets sent
+  by a client.
+  The special interpretation of this bit SHOULD be removed from this
```
martinthomson: This is fine, but you might be better prefixing this with a note to the RFC editor.  You can see one of those in the section on versioning.  In that note, request removal of the bit.  Hopefully we will get to that before this goes to publication.

martinthomson:
You need write access to request review apparently.  And reviewers need to be in the organization.  The mention should suffice.

draft-ietf-quic-transport.md
marten-seemann: I added the note. Just removing this bit won't be enough though, we'll need to add this bit to whatever greasing mechanism we decide for.

csperkins:
Looks okay to me.

draft-ietf-quic-transport.md
```diff
@@ -486,9 +486,21 @@ Key Phase Bit:
...
   protect the packet.  See {{QUIC-TLS}} for details.
 
+\[\[RFC editor: please remove the Google QUIC Demultipexing bit before
+publication.]]
+
+Google QUIC Demultipexing Bit:
+
+: The fifth bit (0x8) of octet 0 is set to 0. This allows implementations of
+  Google QUIC to distinguish Google QUIC packets from short header packets sent
+  by a client.
```
ianswett: How about "from short header packets sent by a client because Google QUIC servers expect the connection ID to always be present." to clarify why this approach works.

marten-seemann: Updated.

draft-ietf-quic-transport.md
ianswett: In an ideal world, we'd be able to move onto this and then start greasing later after GQUIC has been deprecated.  In reality, I suspect we'll either need a mechanism not to grease certain codepoints in the short header or not to grease at all.  I tend to favor the former, since I think it's generally useful(ie: for WebRTC).

draft-ietf-quic-transport.md
```diff
@@ -486,9 +486,22 @@ Key Phase Bit:
   recipient of a packet to identify the packet protection keys that are used to
   protect the packet.  See {{QUIC-TLS}} for details.
 
+\[\[RFC editor: please remove the Google QUIC Demultipexing bit before
+publication.]]
```
MikeBishop: For clarity, we don't want them to actually eliminate the bit, just the text describing why it's there.  I'd just steal the Note from the changelog asking that "this section" be removed, and put it immediately after (rather than before) the header.

draft-ietf-quic-transport.md
```diff
@@ -511,9 +524,9 @@ other fields.
 
 | Type | Packet Number Size |
 |:-----|:-------------------|
-| 0x1F | 1 octet            |
-| 0x1E | 2 octets           |
-| 0x1D | 4 octets           |
+| 0x7  | 1 octet            |
+| 0x6  | 2 octets           |
+| 0x5  | 4 octets           |
```
MikeBishop: Technically, the reason we were counting back instead of forward was to assist in demuxing.  If we're changing these again anyway, do we still need this unintuitive property?

marten-seemann: I’m not sure if we still need it, i don’t think so. @csperkins?

draft-ietf-quic-transport.md
```diff
@@ -486,9 +486,22 @@ Key Phase Bit:
   recipient of a packet to identify the packet protection keys that are used to
   protect the packet.  See {{QUIC-TLS}} for details.
 
+\[\[RFC editor: please remove the Google QUIC Demultipexing bit before
```
martinthomson: @mnot suggests that we make this a simple editors note and that it say:

```
\[\[Editor's Note: this section should be removed and the bit definitions changed before this draft goes to the IESG.]]
```

draft-ietf-quic-transport.md
```diff
@@ -486,9 +486,22 @@ Key Phase Bit:
...
+publication.]]
+
+Google QUIC Demultipexing Bit:
+
+: The fifth bit (0x8) of octet 0 is set to 0. This allows implementations of
```
martinthomson: You need to mention the fourth bit also.

draft-ietf-quic-transport.md
martinthomson: Let's start counting from 0.  As it is, this arrangement collides with DTLS, but that's hard to avoid.

draft-ietf-quic-transport.md
```diff
@@ -486,9 +486,22 @@ Key Phase Bit:
...
+  Google QUIC to distinguish Google QUIC packets from short header packets sent
+  by a client because Google QUIC servers expect the connection ID to always be
+  present.
+  The special interpretation of this bit SHOULD be removed from this
+  specification when Google QUIC has finished transitioning to the new header
+  format.
+
```
janaiyengar: Can you add a paragraph for reserving the 4th bit as well? Maybe call it the STUN Demultiplexing Bit?

draft-ietf-quic-transport.md
janaiyengar: If the plan is to go back to the current types after Google has moved to iQUIC, it really doesn't matter. So I don't care, but leaving this unintuitive property in may be a good thing if that's the long term outcome.

csperkins: If we count from zero, and later set Google QUIC Demultipexing bit back to zero, then we collide with STUN for short header packets with a Connection ID present and key phase zero. Counting down avoids that collision.

If we're going to support P2P QUIC, then having a long-term stable way of easily demuxing with STUN seems important, so I'd prefer to avoid that collision.

martinthomson: @csperkins, the current pattern is 0xx10xxx.  That never collides with STUN.  (The omission of bit 0x10 from the PR is a little confusing.)

csperkins: sure - I was responding to the suggestion that the gQUIC demux bits be removed before IETF QUIC goes final, in which case we would have a collision.

PR #1134, opened 2018-02-21
qcram: remove authors from acknowledgements

### LINKED ISSUE #1132: qcram acks
It's a bit confusing that the acks actually include the document authors.

afrind:
Fixes #1132

And trim an extra whitespace


martinthomson:
To Mike's point, if you make branches in this repo, they will be built for you and you can see and share the results.  For example: https://quicwg.github.io/base-drafts/httpq/

PR #1135, opened 2018-02-21
Tls error

huitema:
I would like to add an error code, to use during the handshake when the TLS parameters proposed by the server cannot be accepted by the client. Example would be a bad certificate.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3724,10 +3724,6 @@ PROTOCOL_VIOLATION (0xA):
 : An endpoint detected an error with protocol compliance that was not covered by
   more specific error codes.
 
-TLS_NEGOTIATION_FAILURE (0xB):
-
-: An endpoint detected an error during the TLS negotiation.
-
 UNSOLICITED_PATH_RESPONSE (0xB):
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3724,6 +3724,10 @@ PROTOCOL_VIOLATION (0xA):
 : An endpoint detected an error with protocol compliance that was not covered by
   more specific error codes.
 
+TLS_NEGOTIATION_FAILURE (0xB):
```
nibanks: You need to fix `UNSOLICITED_PATH_RESPONSE (0xB)` below so there isn't collision on the error codes. That being said, what's wrong with `TLS_HANDSHAKE_FAILED` defined in the QUIC TLS spec?

huitema: Nothing wrong with TLS_HANDSHAKE_FAILED. Apart from not being documented in the transport draft... I updated the PR to just a pointer to the TLS draft.

huitema:
https://www.facebook.com/hatepseudoscience/photos/pb.163735987107605.-2207520000.1519251574./1023329751148220/

martinthomson:
Love the subtle trolling in the image there :)

PR #1137, opened 2018-02-23
add an error code for rejecting new connections

marten-seemann:


draft-ietf-quic-transport.md
```diff
@@ -926,8 +926,11 @@ server continues below.
 
 If the packet is an Initial packet fully conforming with the
 specification, the server proceeds with the handshake ({{handshake}}).
-This commits the server to the version that the client
-selected.
+This commits the server to the version that the client selected.
+
+If the server currently doesn't accept any new connections, it SHOULD send a
+SERVER_BUSY error code. A client that receives this error code SHOULD NOT
+immediately try to reconnect to this server.
```
MikeBishop: "send an error code" isn't the same terminology we've used elsewhere.  I'd use something like "If a server isn't currently accepting any new connections, it SHOULD send a Handshake packet containing a CONNECTION_CLOSE frame with error code SERVER_BUSY."

draft-ietf-quic-transport.md
```diff
@@ -3676,6 +3679,9 @@ INTERNAL_ERROR (0x1):
 : The endpoint encountered an internal error and cannot continue with the
   connection.
 
+SERVER_BUSY (0x2):
+: The server is currently busy and does not accept any new connections.
```
MikeBishop: Others have a line between entry and definition.  I don't particularly care for that style, but let's be consistent anyway.

draft-ietf-quic-transport.md
marten-seemann: Thanks for the suggestion. That sounds better.

PR #1143, opened 2018-02-28
QPACK Blocking Setting

### LINKED ISSUE #1140: Non-blocking decoder
The QCRAM draft currently enables an encoder to operate in blocking or non-blocking mode, because the encoder always knows which header blocks have been processed by the decoder.  However, there's currently no mechanism for a decoder to require an encoder to operate in non-blocking mode.  During design team discussions, this seemed to be a desirable property -- do we want to add a setting to support this?

There was also some discussion that a decoder which permitted blocking might still want to constrain how many blocked decoder instances it had, and therefore put a bound on the number of blocking QCRAM frames which should be permitted.

MikeBishop:
Based on #1141, fixes #1140.  Creates a setting that indicates blocking references are permitted.  We can, if desired, extend this further and make the setting an integer.  0-RTT issues are dealt with by adding this to the existing 0-RTT discussion under the same terms.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -781,10 +781,10 @@ The following settings are defined in HTTP/QUIC:
   : An integer with a maximum value of 2^30 - 1.
 
   SETTINGS_MAX_HEADER_LIST_SIZE (0x6):
-  : An integer with a maximum value of 2^30 - 1
+  : An integer with a maximum value of 2^30 - 1.
 
   SETTINGS_BLOCKING_HEADER_REFERENCES (0x7):
-  : A Boolean
+  : An integer with a maximum value of 2^16 - 1.
 
 #### Initial SETTINGS Values
 
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -778,13 +778,15 @@ Settings which are integers use the QUIC variable-length integer encoding.
 The following settings are defined in HTTP/QUIC:
 
   SETTINGS_HEADER_TABLE_SIZE (0x1):
-  : An integer with a maximum value of 2^30 - 1.
+  : An integer with a maximum value of 2^30 - 1.  The default value is 4,096
+    bytes.
 
   SETTINGS_MAX_HEADER_LIST_SIZE (0x6):
-  : An integer with a maximum value of 2^30 - 1.
+  : An integer with a maximum value of 2^30 - 1.  The default value is
+    unlimited.
 
   SETTINGS_BLOCKING_HEADER_REFERENCES (0x7):
-  : An integer with a maximum value of 2^16 - 1.
+  : An integer with a maximum value of 2^16 - 1.  The default value is 100.
 
 #### Initial SETTINGS Values
 
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -785,7 +785,7 @@ The following settings are defined in HTTP/QUIC:
   : An integer with a maximum value of 2^30 - 1.  The default value is
     unlimited.
 
-  SETTINGS_BLOCKING_HEADER_REFERENCES (0x7):
+  SETTINGS_QPACK_BLOCKED_STREAMS (0x7):
   : An integer with a maximum value of 2^16 - 1.  The default value is 100.
 
 #### Initial SETTINGS Values
@@ -1454,6 +1454,7 @@ The entries in the following table are registered by this document.
 | Reserved                   | 0x4  | N/A                     |
 | Reserved                   | 0x5  | N/A                     |
 | MAX_HEADER_LIST_SIZE       | 0x6  | {{settings-parameters}} |
+| QPACK_BLOCKED_STREAMS      | 0x7  | {{settings-parameters}} |
 |----------------------------|------|-------------------------|
 
 ## Error Codes {#iana-error-codes}
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -783,6 +783,9 @@ The following settings are defined in HTTP/QUIC:
   SETTINGS_MAX_HEADER_LIST_SIZE (0x6):
   : An integer with a maximum value of 2^30 - 1
 
+  SETTINGS_BLOCKING_HEADER_REFERENCES (0x7):
+  : A Boolean
+
 #### Initial SETTINGS Values
 
 When a 0-RTT QUIC connection is being used, the client's initial requests will
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -100,9 +100,10 @@ which might not yet have been processed by the recipient.  Such references
 are called "vulnerable," because the loss of a different packet can keep
 the reference from being usable.
 
-The encoder can choose on a per-header-block basis whether to favor higher
-compression ratio (by permitting vulnerable references) or HoL resilience (by
-avoiding them).
+The decoder can signal that it is willing to process vulnerable references by
+setting SETTINGS_BLOCKING_HEADER_REFERENCES to true.  In this case, the encoder
+can choose on a per-header-block basis whether to favor higher compression ratio
+(by permitting vulnerable references) or HoL resilience (by avoiding them).
 
 The header block contains a Base Index (see {{absolute-index}}) which is used to
 correctly index entries, regardless of reordering in the transport (see
@@ -112,6 +113,7 @@ least the value of the Base Index.  While blocked, header field data MUST remain
 in the blocked stream's flow control window.  When the Base Index is zero, the
 frame contains no references to the dynamic table and can always be processed
 immediately.
+
 
 # Wire Format
 
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -102,9 +102,9 @@ received.
 Each header block contains a Largest Reference (see {{absolute-index}}) which
 identifies the table state necessary for decoding. If the greatest absolute
 index in the dynamic table is less than the value of the Largest Reference, the
-stream is considered "blocked."  While blocked, header field data MUST remain in
-the blocked stream's flow control window.  When the Largest Reference is zero,
-the frame contains no references to the dynamic table and can always be
+stream is considered "blocked."  While blocked, header field data should remain
+in the blocked stream's flow control window.  When the Largest Reference is
+zero, the frame contains no references to the dynamic table and can always be
 processed immediately.
 
 A decoder can permit the possibility of blocked streams by setting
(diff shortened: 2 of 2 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -100,10 +100,13 @@ which might not yet have been processed by the recipient.  Such references
 are called "vulnerable," because the loss of a different packet can keep
 the reference from being usable.
 
-The encoder can choose on a per-header-block basis whether to favor higher
+The decoder can signal that it is willing to process vulnerable references by
```
martinthomson: I still don't like the term "vulnerable references" here.  The point is that the decoder is willing to block on incomplete or in-progress header table updates or not.  "Vulnerable" as a concept isn't helpful, especially here.

draft-ietf-quic-http.md
```diff
@@ -786,7 +786,10 @@ The following settings are defined in HTTP/QUIC:
   : An integer with a maximum value of 2^30 - 1.
 
   SETTINGS_MAX_HEADER_LIST_SIZE (0x6):
-  : An integer with a maximum value of 2^30 - 1
+  : An integer with a maximum value of 2^30 - 1.
+
+  SETTINGS_BLOCKING_HEADER_REFERENCES (0x7):
+  : An integer with a maximum value of 2^16 - 1.
```
martinthomson: These need defaults.

draft-ietf-quic-qpack.md
```diff
@@ -115,6 +118,9 @@ state which the peer might not have received yet, the BLOCKING flag MUST be
 set.  If the peer does not yet have the appropriate state, such blocks
 might not be processed on arrival.
 
+An encoder MUST NOT have more header blocks outstanding with the BLOCKING flag
+set than the value of SETTINGS_BLOCKING_HEADER_REFERENCES at any time.
```
martinthomson: So what does a decoder do if it has to block more times than it said it was willing to?  You should also note that an encoder might violate this agreement but not be detected in doing so.

MikeBishop:
Perhaps the settings should be in QPACK instead.  I was keeping it that way because that's what HTTP/2 did, but we're breaking with precedent other places we feel like we've learned better now, so... perhaps.  However, that looks to have enough ripple effects on the HTTP doc that I'd prefer to do that in a separate PR.  This is now #1323.

afrind:
I agree with @dtikhonov that blocking is worse.  It's neither blocking nor vulnerable, but vulnerable-to-HoL-blocking.  I say change it back for now and defer coming up with a better term for another issue.

MikeBishop:
Here's another stab at it -- references are "vulnerable" if the encoder sends them without knowing that the decoder has the entries.  They're "blocking" if they arrive at the decoder and it actually doesn't.  Because each vulnerable reference *could be* blocking, the encoder needs to keep its number of vulnerable references below the number of blocking references the decoder will tolerate.

draft-ietf-quic-qpack.md
```diff
@@ -95,15 +95,26 @@ reintroduced by HPACK when the loss includes a HEADERS frame.
...
-which might not yet have been processed by the recipient.  Such references
-are called "vulnerable," because the loss of a different packet can keep
-the reference from being usable.
-
-The encoder can choose on a per-header-block basis whether to favor higher
+In the example above, the header block on the second stream contained a
+reference to data which might not yet have been processed by the recipient. Such
+references are called "vulnerable," because the loss of a different packet can
+keep the reference from being usable.  If the reference cannot be immediately
+processed on receipt, the reference is considered "blocking."
```
afrind: It's more blocked than blocking (the missing one is blocking)?

draft-ietf-quic-qpack.md
```diff
@@ -494,7 +505,9 @@ Indexed-Duplicate representation instead (see {{indexed-duplicate}}).
 
 For header blocks encoded in non-blocking mode, the encoder needs to forego
 indexed representations that refer to table updates which have not yet been
-acknowledged with {{feedback}}.
+acknowledged with {{feedback}}.  Since all table updates are processed in
+sequence on the control stream, an index into the dynamic table is sufficient to
+track which entries have been processed.
```
afrind: change processed to acknowledged to match the previous sentence?

afrind:
The setting is still called BLOCKING_HEADER_REFERENCES, but the text now talks about blocked streams.  Should we update this as well?  SETTINGS_MAX_STREAMS_BLOCKED_ON_HEADERS is kind of a mouthful.

MikeBishop:
How about SETTINGS_QPACK_BLOCKED_STREAMS?

draft-ietf-quic-qpack.md
```diff
@@ -95,17 +95,29 @@ reintroduced by HPACK when the loss includes a HEADERS frame.
...
 ## Avoiding Head-of-Line Blocking in HTTP/QUIC {#overview-hol-avoidance}
 
-In the example above, the second stream contained a reference to data
-which might not yet have been processed by the recipient.  Such references
-are called "vulnerable," because the loss of a different packet can keep
-the reference from being usable.
-
-The encoder can choose on a per-header-block basis whether to favor higher
+Because QUIC does not guarantee order between data on different streams, a
+reference might be received on a request stream to an entry which is not yet in
```
martinthomson: request streams aren't the only affected streams

"a header block might reference an entry in the dynamic table that has not yet been received."

draft-ietf-quic-qpack.md
```diff
@@ -95,17 +95,29 @@ reintroduced by HPACK when the loss includes a HEADERS frame.
...
+the dynamic table. References are called "vulnerable" when they refer to entries
+whose receipt has not been confirmed by the peer, because loss or reordering can
+keep the reference from being immediately usable.  If the reference cannot be
+immediately processed on receipt, the stream is considered "blocked."
```
martinthomson: I *still* don't like "vulnerable" and I think that I understand better now.  The problem is that you are putting the cart before the horse.  The problem is that the decoder might block.  The focus on the encoder side is wrong.

Hence, I suggest, without the word "vulnerable" at all:

> [The QUIC is not ordered preface]. A decoder that receives a reference to a dynamic table entry that has not been received is "blocked".
>
> An encoder can decide whether to risk having a header block becoming blocked at a decoder.  Compression efficiency can be improved by referencing dynamic table entries that are still in transit, but if there is packet loss or reordering a decoder can become blocked.  An encoder avoids the risk of blocking by only referencing dynamic table entries that have been acknowledged, though this means using literals, which make the header block larger.  Larger header blocks can result in an encoder become blocked on congestion or flow control limits.
>
> A decoder can permit the possibility of blocking references using the SETTINGS_QPACK_BLOCKING_REFERENCES setting.  This setting sets an upper bound on the number of header blocks that can reference unacknowledged entries in the dynamic table.  An encoder respects this setting when deciding whether to reference unacknowledged dynamic table entries.
>
> Every header block contains a Largest Reference (see ...).  This is used to determine the minimum number of entries need to be present in the dynamic table before the header block can be successfully decoded.  A decoder becomes blocked if the Largest Reference is greater than the number of dynamic table entries that it has received.
>
> An encoder MUST limit the number of unacknowledged header blocks that reference unacknowledged dynamic table entries to the value of the SETTINGS_...  A decoder that is blocked on a number of header blocks equal to the limit it advertised SHOULD treat receipt of an additional blocked header block [awkward] as a stream error of type HTTP_...

Or something like that.

draft-ietf-quic-qpack.md
```diff
@@ -547,9 +559,9 @@ have outstanding (unacknowledged) references.
 An encoder MUST ensure that a header block which references a dynamic table
 entry is not received by the decoder after the referenced entry has already been
 evicted, and MUST ensure that the decoder will not suffer head-of-line blocking
-if the decoder has not opted to receive blocking references. Even if the decoder
-is willing to process blocking references, the encoder might choose to avoid
-them in certain cases.
+if the decoder has not opted to receive vulnerable references. Even if the
```
martinthomson: End the sentence after "evicted".  Then "An encoder also respects the limit set by the decoder on the number of header blocks that are allow to be blocked."

draft-ietf-quic-qpack.md
afrind: I think this is a workable direction.  We should be consistent about blocked vs blocking (I prefer blocked), and what is blocked - is it the stream, header block, or reference?

A single header block may contain many references to unacknowledged entries, so it's not blocked references the decoder is counting, but header blocks.  It seems logical that the entire stream should be blocked until the blocked header block is decoded, but perhaps the text needs to be more explicit about it.

All of this to say that I like Mike's suggestion: SETTINGS_QPACK_BLOCKED_STREAMS.

martinthomson: Yes, I think that the unit is header blocks.  A decoder is blocked on the decoding of a header block.  

We can generalize that to streams, since the correspondence is perfect to streams: a blocked header block is a blocked stream.  In fact, we should...

Thinking more about that, if an endpoint is silly and they read a blocked header block, then concentrating on header blocks might be wrong.  A message with trailers has two header blocks and if both are blocked, you might expect the encoder to only account for one lot of blocking.  But a silly decoder might find itself blocked on both header and trailer at the same time.  Specifying this as the number of streams that have a decoder blocked on the decoding of a header block with unresolved references to the dynamic table is ideal.

MikeBishop: I agree that the unit should be streams; the decoder SHOULD NOT be reading even the body of the header block, which means it won't see any further header blocks on that stream.  A cautious encoder might choose to count blocks for simplicity, but the decoder's count will almost certainly be one per stream.  A silly decoder might read further in an attempt to over-optimize, and specifying the blocking count to be streams helps keep them from blowing up.

draft-ietf-quic-qpack.md
```diff
@@ -96,36 +96,37 @@ reintroduced by HPACK when the loss includes a HEADERS frame.
...
-will be less than or equal to the encoder's count of references which are
-vulnerable (potentially blocking).  If the decoder encounters more blocked
+header block might reference an entry in the dynamic table that has not yet been
+received.
+
+Each header block contains a Largest Reference (see {{absolute-index}}) which
+identifies the table state necessary for decoding. If the greatest absolute
+index in the dynamic table is less than the value of the Largest Reference, the
+stream is considered "blocked."  While blocked, header field data MUST remain in
+the blocked stream's flow control window.  When the Largest Reference is zero,
```
martinthomson: I don't think that there is an interoperability requirement here.  Holding data outstanding is advice, not a requirement.

draft-ietf-quic-qpack.md
```diff
@@ -96,36 +96,37 @@ reintroduced by HPACK when the loss includes a HEADERS frame.
...
+
+A decoder can permit the possibility of blocked streams by setting
+SETTINGS_QPACK_BLOCKED_STREAMS to a non-zero value.  This setting specifies an
+upper bound on the number of streams which can be blocked.
+
+An encoder can decide whether to risk having a stream become blocked. If
+permitted by the value of SETTINGS_QPACK_BLOCKED_STREAMS, compression efficiency
+can be improved by referencing dynamic table entries that are still in transit,
+but if there is loss or reordering the stream can become blocked at the decoder.
+An encoder avoids the rist of blocking by only referencing dynamic table entries
```
martinthomson: typo: risk

draft-ietf-quic-qpack.md
```diff
@@ -96,36 +96,37 @@ reintroduced by HPACK when the loss includes a HEADERS frame.
...
+which have been acknowledged, but this means using literals. Since literals make
+the header block larger, this can result in the encoder becoming blocked on
+congestion or flow control limits.
+
+An encoder MUST limit the number of streams which could become blocked to the
+value of SETTINGS_QPACK_BLOCKED_STREAMS at all times. Note that the decoder's
+count of streams which are actually blocked will be at most the encoder's count
+of streams which could become blocked.  If the decoder encounters more blocked
```
martinthomson: How about simply: "A decoder might not be forced to block on streams that risk being blocked."

draft-ietf-quic-qpack.md
```diff
@@ -95,24 +95,36 @@ reintroduced by HPACK when the loss includes a HEADERS frame.
...
+header block might reference an entry in the dynamic table that has not yet been
+received.
+
+Each header block contains a Largest Reference (see {{absolute-index}}) which
+identifies the table state necessary for decoding. If the greatest absolute
+index in the dynamic table is less than the value of the Largest Reference, the
+stream is considered "blocked."  While blocked, header field data should remain
+in the blocked stream's flow control window.  When the Largest Reference is
+zero, the frame contains no references to the dynamic table and can always be
+processed immediately.
```
afrind: We probably need to explain when a stream becomes unblocked.

Issue #1128, opened 2018-02-21
Duplication can use different indexing


martinthomson:
A duplication instruction never references the static table.  Currently however, this is possible, and it's not a great idea.

Aside from it being pointless to duplicate entries in the static table (it costs space and time to do this, and references to these new entries will have longer encodings than static table references), in the current design the encoding of the duplication instruction is such that virtually every duplication instruction has to use a two octet encoding because exactly one entry in the dynamic table can be duplicated in one octet.  Duplication of this entry is completely useless because duplicating it does nothing to improve its position in the table.

mikkelfj:
Can please refer to QCRAM or HTTP on this and similar issues - it’s otherwise hard to decipher context.

afrind:
One proposed solution comes from the QPACK draft, which is to use a bit in the instruction to indicate if a referenced index is static or dynamic.  I found that this actually simplified parts of the implementation compared to HPACK/QCRAM's unified index space.  If we separate the instruction space, then we can define duplication to always operate on the dynamic table index space.  It also may compress slightly better, since more dynamic indexes can be encoded by a single byte.

afrind:
@mikkelfj : is the -qcram tag insufficient to differentiate? Would you prefer something in the issue title?

LPardue:
When I read these issues that are pushed in email there is no visibility of
the tag.

As a non-editor I cannot set tags for an issue, so it makes it hard to
imply scope.

This situation is not unliveable.

However, some projects define a namespace to be used in the issue title in
order to help such situations. "qcram" and "hq" might be enough to help to
disambiguate from transport, security etc.

mikkelfj:
@afrind yes QCRAM in the title is fine, or qcram:, just something. The issue is with email as @LPardue says.

martinthomson:
@afrind, the extra bit is less efficient overall (see also why arithmetic coding is superior to Huffman).  Especially given that the static table is a fixed size.  I don't find the indexing that onerous, so I don't think that the extra bit would help.  That is, unless we get a proposal for a new, bigger static table.  Having references to the dynamic table start at (for example) 200, would make the bit a good investment.

afrind:
If we leave the on-the-wire index space unified, is the proposal to use a different indexing scheme for this instruction only (eg: skip the + 62)?  I'm not crazy about having two different indexing schemes in the design, but otherwise it would solve the issues you raise above.

martinthomson:
[...]

PR #1144, opened 2018-02-28
QCRAM opcodes with non-byte-aligned string literals

MikeBishop:
Based on #1141 and sparked by a comment on that PR that I'd like to discuss separately.  There are effectively four instructions in a table update:

1. Insert with static table name reference
2. Insert with dynamic table name reference
3. Insert with no reference
4. Duplicate (a.k.a. insert with name+value reference)

#1141 follows HPACK's lead in treating insert as a single instruction with internal flags describing different possible references.  That leads to effectively using two-bit prefixes for (1)/(2-3), a sentinel value in the index to differentiate (2)/(3), and (4) gets a single-bit instruction code even though it's the least common.  This PR treats them as four instructions and uses two-bit prefixes for each.

The impact of this is break-even for literals which reference a name from the header table.  The benefit is that you save a byte on almost every header which is inserted without a name reference.

The cost is:

- "Roughly 15 minutes of hacking (to add the extra argument)" in @martinthomson's estimation
- An extra byte required to duplicate entries which are at indexes 64-127, modulo #1128.
- It's no longer as simple to look at one bit and say that bit has a consistent meaning

### DIFF
--- a/draft-ietf-quic-qcram.md
+++ b/draft-ietf-quic-qcram.md
@@ -198,11 +198,11 @@ string literal (see Section 5.2 of [RFC7541]).
 {: title="Insert Header Field -- Indexed Name"}
 
 
-### Insert Without Reference
+### Insert Without Name Reference
 
 An addition to the header table where both the header field name and the header
 field value are represented as string literals (see {{primitives}}) starts with
-the '00' two-bit pattern.
+the '01' two-bit pattern.
 
 The name is represented as a 6-bit prefix string literal, while the value is
 represented as an 8-bit prefix string literal.
@@ -210,7 +210,7 @@ represented as an 8-bit prefix string literal.
 ~~~~~~~~~~ drawing
      0   1   2   3   4   5   6   7
    +---+---+---+---+---+---+---+---+
-   | 0 | 0 | H | Name Length (5+)  |
+   | 0 | 1 | H | Name Length (5+)  |
    +---+---+---+-------------------+
    |  Name String (Length octets)  |
    +---+---------------------------+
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qcram.md
```diff
@@ -175,17 +197,21 @@ always non-zero; a zero index is reserved for literal names.
 ~~~~~~~~~~
 {: title="Insert Header Field -- Indexed Name"}
 
-Otherwise, the header field name is represented as a string literal (see Section
-5.2 of [RFC7541]). A value 0 is used in place of the table reference, followed
-by the header field name.
+
+### Insert Without Reference
```
martinthomson: Without Name Reference

draft-ietf-quic-qcram.md
```diff
@@ -175,17 +197,21 @@ always non-zero; a zero index is reserved for literal names.
...
+represented as an 8-bit prefix string literal.
 
 ~~~~~~~~~~ drawing
      0   1   2   3   4   5   6   7
    +---+---+---+---+---+---+---+---+
-   | 1 | 0 |           0           |
-   +---+---+-----------------------+
-   | H |     Name Length (7+)      |
-   +---+---------------------------+
+   | 0 | 0 | H | Name Length (5+)  |
```
martinthomson: Almost a bike shed: invert the second bit here.  Then you can reasonably say that you are describing 11 (insert with static name reference), 10 (insert with dynamic name reference), 01 (insert with new name), 00 (duplication).

MikeBishop: I'm not wholly sure that makes a difference, but I suppose there's an argument to be made that '00' is the "unique" one.

MikeBishop:
Having separately discussed bringing back table size change, there are now five instructions, so "00" represents "uncommon things" with a third bit to differentiate duplicate versus size change.  Purely by coincidence, that means size change is now identical to the HPACK instruction.

Issue #1146, opened 2018-02-28
Version Negotiation Routing Problem


mikkelfj:
The invariants draft mentions the version negotiation packet.

If QUIC moves to asymmetric connection ID's it may become a problem because the connection ID may be used to route through middleware at both endpoints.

It is also not clear that the version negotation packet is indeed a version negotiation packet is a version negotiation packet because the only reliable information is that it is a long header format. Therefore it is hard for middleware to make exceptions for that packet, assuming it could then route the packet correctly, which is not a given.

Of course there are also details about the length of the connection ID, which have not been settled, but I trust that will be updated once decided.

mikkelfj:
https://github.com/quicwg/base-drafts/pull/1151 will resolve this issue.

martinthomson:
I don't see how #1151 fails to fix this.  The client puts its preferred value in the packet it sends and the server has to use that value.  These values are available to all middleboxes on the path.  No exception is needed for Version Negotiation.

mikkelfj:
@martinthomson I really like the changes in #1151 and it does solve most concerns. You argue my remaining issue is a matter of expections and that it is not specific to vneg:

The concern is that the server and the TLS handshake places some restrictions on the initial client source CID because it is required to be random. If the client exists in a complex p2p server network it might want detailed control of routing bits in CID. A middelbox could do NAT style routing setup and remember the CID for reverse routing, but that is an assumption on infrastructure. In particular the reverse route might take an entirely different path so NAT won't work at all. Encrypting the CID is an option even if less than 16 bytes, but not all infrastructure might want to do that.

At the same time there are (or ought to be) requirements on CIDs to be unique per tuple and for linkability concerns etc. so there is a broader discussion on what a CID must fulfill. Since a CID is allowed to be zero length, it cannot be very hard requirements in general.

So, removing randomness restrictions on CIDs and in place provide recommendations or requirements for specific usecases (linkability) would remove the concerns I still have.

Note that randomness is less important now that the server can always choose it's own CID.

martinthomson:
Only the destination connection ID needs to be random in that proposal.  The client sets the source connection ID in the Initial packet to whatever it likes.  The idea there being that the server will use this value in its destination connection ID on the return path.

mikkelfj:
Oh, maybe I'm not thinking clearly here. In vneg dest is random, fine. On normal handshake destination is replaced by servers chosen source. And then both CID's are free. Yes.

martinthomson:
[...]

PR #1151, opened 2018-03-02
Symmetric connection IDs

### LINKED ISSUE #1089: Asymmetric connection ID
This is probably a v2 thing

The multiple connections per tuple has been shot down for v1, but let's consider the general case:

Two hosts in one or more data centers establish peer to peer communication. Each side has its own routing infrastructure and need routing guidance in the CID.

There is no particular reason why a client and a server should use the same CID on the same path, after all, it really is two paths, one for each direction. This also makes it possible to require CID in one direction, but not the other, or to have different routing information encoded. It would also allow each peer full control of the construction of the CID being received. This means that the peer can cryptographically reject any packet that does not have a valid CID long before any packet AEAD is applied, thus defending against a flooding attack. There is no need to negotiate any encryption for this to work, just as is the case for the server decided CID in current 17 byte CID proposal.

As to multiple connections on same tuple and regarding initial packet CID:

The initial CID should appear random, but can be any pseudo random function. The server generated CID is then NOT returned in the packet header, but in a separate field, possibly in transport params. For version negotiation the initial CID is used in reply. For stateless retry the initial CID is used, but a second server CID is added to the package.

### LINKED ISSUE #1101: Asymmetric Connection IDs
Sorry to totally randomize things, but several of us had an offline discussion about asymmetric connection IDs, in which each side specifies the connection ID that the other
side should send with. Here's a sketch of what that would look like.

PROTOCOL OVERVIEW
```
Client                                      Server

Initial [CID=XXX] {recv-CID=YYY} ---------------->
<-------------- Cleartext [CID=YYY] {recv-CID=ZZZ}
Cleartext [CID=ZZZ] ----------------------------->
<-------------------------- Short header [CID=YYY]
Short header [CID=ZZZ] -------------------------->
```
The initial packet contains:
- In the header CID field, XXX, which is either
  (a) a randomly chosen dummy CID
  (b) a CID which it received from the server in a stateless reset
- The transport parameters contain the client's receive CID (YYY in
  the diagram above).
  
All the server's packets are sent with the client's receive CID and
the transport parameters contain the server's receive CID (ZZZ).

Subsequent packets from the client are sent with the server's receive
CID (ZZZ) 

Finally, you can send NEW_CONNECTION_ID in either direction to provide
a new connection ID for the other side to use.

OTHER DETAILS
This design doesn't require variable length CIDs but it's a natural
fit. In particular, it doesn't require the C bit in the long header
because each side gets to indicate exactly how long a CID it wants,
and so sending an empty CID means "don't send me a CID". I haven't
heard a good reason why we should let the packet sender decide whether
to send a CID or not, and so this is largely for the benefit of
middleboxes which doesn't seem like a good reason.

Orthogonally, here's a sketch of a long header with a variable
length CID:

```
   +-+-+-+-+-+-+-+-+---------------+
   |0|C|K| Type (5)|  CID length   |
   |                                                               |
   +                     [Connection ID (*)]                       +
   |                                                               |
   |                      Packet Number (8/16/32)                ...
   |                     Protected Payload (*)                   ...
```

Given that the CID is now variable length, we might or might not
want to move it to after the PN.

OPEN ISSUES/PROBLEMS
1. If the server's handshake gets reordered, the client might need to
send some ACKs with the initial CID. However, we've agreed that the
client's IP address has to be stable, so this isn't a
problem. Alternately, you could change C->S CIDs in the short header
if that was easier. This might be an argument for the idea of putting
both CIDs in the long header that I proposed in Melbourne.

2. Any asymmetric ID scheme inherently makes stateless reset more
difficult, because the server doesn't have the client's CID. It could
use a random CID or the server's CID, though. One might not think this
is so bad because it's really only an issue for a client which needs
stateless routing of packets (the client itself is of course
stateful).  Previously, this topology was just impossible, so one
could argue that this was strictly better, as the client will just
fall back to timeout if the routing fails. Of course, you might
think that it was really bad too.

martinthomson:
Lots of changes here.

Obviously people will feel inclined to nitpick at the length encoding.  Feel
free, but I think that I have good reasons for my choices there.

The NEW_CONNECTION_ID thing is tricky.  As was the analysis around the
handshake packet keys.  Also, note the trick in Version Negotiation packets
about swapping the connection IDs.  That makes client packet handling much
simpler.

The big loss is in stateless reset, which I think that I've covered well
enough.

Closes #1089, #1052, #821.


draft-ietf-quic-invariants.md
```diff
@@ -257,9 +271,9 @@ Version Negotiation packets do not use integrity or confidentiality protection.
 A specific QUIC version might authenticate the packet as part of its connection
 establishment process.
 
-The Connection ID field in a Version Negotiation packet contains the Connection
-ID from the packet that was received.  This provides some protection against
-injection of Version Negotiation packets by off-path attackers.
+The Connection ID fields in a Version Negotiation packet contain the Connection
+ID values from the packet that was received.  This can provide some protection
+against injection of Version Negotiation packets by off-path attackers.
```
mikkelfj: ID values in reversed order.

draft-ietf-quic-tls.md
```diff
@@ -703,6 +704,22 @@ ensuring that the keys are different for each version of QUIC. This prevents a
...
+  that included the Retry packet for that property.
+
+Note(2):
+
+: HKDF-Extract uses HMAC {{?RFC2104}}, which produces the same output for values
+  ending in zero octets and values that omit those octets.  Thus,
+  0x12345678900000 and 0x1234567890 produce the same result.  A client is
+  required to randomize the value of Destination Connection ID, which avoids
+  this problem.
+
```
mikkelfj: Can we avoid requiring it to be random, or allow it to be only partially random? If a client has intimate knowledge of the server infrastructure, it might want to route the packet, or add some guidance in the initial destination ID.

draft-ietf-quic-transport.md
```diff
@@ -557,12 +562,20 @@ The layout of a Version Negotiation packet is:
...
+The value in the Unused field is selected randomly by the server.
+
+The server MUST include the value from the Source Connection ID field of the
+packet it receives in the Destination Connection ID field; similarly, the value
+of the Source Connection ID field is set to the Destination Connection ID from
+the received packet.  The value for Source Connection ID is randomly selected by
+a client and so gives clients some assurance that the server received the packet
+and that the Version Negotiation packet was not generated by an off-path
+attacker.
+
```
mikkelfj: The the client sends a random source ID, it cannot be used to route the vneg packet back to the client. It could be partially a random. Even if encrypted, it wouldn't really be random. It seems that the need for nonces somewhat conflicts with the idea of using the ID for routing. It could also make the connection ID longer than necessary just to ensure initial entropy that isn't needed subsequently.

vneg routing issue: https://github.com/quicwg/base-drafts/issues/1146

draft-ietf-quic-transport.md
```diff
@@ -594,10 +607,14 @@ off-path attackers.
...
-packet from the server, the connection ID field uses the value provided by the
-server.
+If the client has not previously received a Retry packet from the server, it
+populates the Destination Connection ID field with a randomly selected value.
+This MUST NOT be zero length, but can be any other length.  If the client
+received a Retry packet and is sending a second Initial packet, then it uses the
+value from the Source Connection ID in the Retry packet.
+
+The client populates the Source Connection ID field with a value of its choosing
+and sets the low bits of the ConnID Len field to match.
```
mikkelfj: Again same issue with randomness as for vneg.

mikkelfj:
Although it adds even more bulk, it might make sense to include a separate 32 bit nonce value in long headers. It would allow ID's to be routable and ensure entropy even when ID's are not in use.

draft-ietf-quic-invariants.md
```diff
@@ -178,16 +191,17 @@ version-specific semantics are marked with an X.
...
+or Version fields.
+
+The remainder of the packet has version-specific semantics.
 
 
 ## Connection ID
 
-A connection ID is an opaque 64-bit field.
+A connection ID is an opaque field.  A connection ID can be 0 octets in length,
+or between 8 and 22 octets (inclusive).
```
nibanks: Why limit the allowable range to 8 and 22 (inclusively)? Practically, the size just needs to be large enough for the number of connections on a given tuple, right? So if a client only needs to share with a couple of connections, 1 byte might be sufficient.

draft-ietf-quic-transport.md
```diff
@@ -435,10 +450,11 @@ The following packet types are defined:
 | 0x7C | 0-RTT Protected               | {{packet-protected}}        |
 {: #long-packet-types title="Long Header Packet Types"}
 
-The header form, packet type, connection ID, packet number and version fields of
-a long header packet are version-independent. The types of packets defined in
-{{long-packet-types}} are version-specific.  See {{QUIC-INVARIANTS}} for
-details on how packets from different versions of QUIC are interpreted.
+the header form, type, connection ID lengths octet, destination and source
```
nibanks: Should `the` be capitalized?

draft-ietf-quic-transport.md
```diff
@@ -881,33 +909,30 @@ Incoming packets are classified on receipt.  Packets can either be associated
...
 
-If there is no connection ID, but the packet matches the address/port tuple
-of a connection where the host did not require connection IDs, QUIC processes
-the packet as part of that connection. Endpoints MUST drop packets that omit
-connection IDs if they do not meet both of these criteria.
+If the Destination Connection ID is zero length, but the packet matches the
+address/port tuple of a connection where the host did not require connection
+IDs, QUIC processes the packet as part of that connection. Endpoints MUST drop
+packets with zero-length Destination Connection ID fields if they do not meet
+both of these criteria.
```
nibanks: This adds a bit of complexity that I'm not sure is worth it. As I understand it, it would require a look up table of all remote address/port tuples in use, that would have to be checked for every new connection attempt. If we don't do this, and leave it up to the peer to make the smart choice, one of two things happen:

1) The do things correctly and everything works.
2) They have two connections on the same tuple, 1 with a CID and one without. The one without breaks down because it cannot distinguish its short header packets from the other one.

If the peer does 2) above, then it's their own fault and they suffer the consequences.

draft-ietf-quic-tls.md
MikeBishop: Tempting to simply require that the final octet not be zero.  If the client has knowledge of how to generate a Connection ID, it can respin in the 1/16 chance it generates one that ends in zeros.

Or alternatively, append a fixed non-zero octet to the Connection ID to side-step this problem.

draft-ietf-quic-transport.md
```diff
@@ -288,16 +289,12 @@ processing.
...
-address. QUIC provides automatic cryptographic verification of a rebound client,
-since the client continues to use the same session key for encrypting and
-decrypting packets. The consistent connection ID can be used to allow migration
-of the connection to a new server IP address as well, since the Connection ID
-remains consistent across changes in the client's and the server's network
-addresses.
+QUIC connections carry a connection ID, an opaque value assigned by the intended
+recipient of the packet.  Using a consistent connection ID allows connections to
+survive changes to the sender's IP and port, such as those caused by NAT
+rebindings or by changes in network connectivity.  QUIC provides mechanisms that
```
MikeBishop: "mechanisms to," I'd say.  The verification doesn't happen by itself.

draft-ietf-quic-transport.md
```diff
@@ -288,16 +289,12 @@ processing.
...
+verify addresses of endpoints that change network connectivity, while
+maintaining session continuity.
```
MikeBishop: This phrase doesn't feel right with a comma, but without the comma it's ambiguous which idea it's referring to.  Maybe reword?

draft-ietf-quic-transport.md
```diff
@@ -370,12 +367,12 @@ keys are established.
 ~~~~~
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
-+-+-+-+-+-+-+-+-+
-|1|   Type (7)  |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|1|   Type (7)  | ConnID Len(8) |
```
MikeBishop: Same here.

draft-ietf-quic-transport.md
MikeBishop: Yes, it should.

draft-ietf-quic-transport.md
MikeBishop: This also conflicts with a paragraph or so lower, which says that the Source CID is "of [the client's] choosing" and later says "this might be a zero length value."

draft-ietf-quic-transport.md
```diff
@@ -623,8 +640,13 @@ A Retry packet uses long headers with a type value of 0x7E.  It carries
 cryptographic handshake messages and acknowledgments.  It is used by a server
 that wishes to perform a stateless retry (see {{stateless-retry}}).
 
-The server includes a connection ID of its choice in the connection ID field.
-The client MUST use this connection ID for any subsequent packets that it sends.
+The server populates the Destination Connection ID with the connection ID that
+the client included in the Source Connection ID of the Initial packet.  This
+might be a zero length value.
```
MikeBishop: Adjective preceding a noun, so zero-length.

draft-ietf-quic-transport.md
```diff
@@ -712,23 +737,26 @@ sequence of frames, as described in {{frames}}.
...
+header contains two connection IDs: the Destination Connection ID is chosen by
+the recipient of the packet and is used to provide consistent routing; the
+Source Connection ID is used to set the Destination Connection ID used by a
+peer.
 
-The client MUST choose a random connection ID and use it in Initial packets
-({{packet-initial}}) and 0-RTT packets ({{packet-protected}}).
+During the handshake, packets with the long header are used to establish the
+Destination Connection ID that each endpoint uses.  Each endpoint sends the
+value that they wish their peer to use in the Source Connection ID field.  Upon
```
MikeBishop: It should be clear from context, but the sentence by itself is slightly ambiguous.  Perhaps "Each endpoint uses the Source Connection ID field to communicate the CID which should be used in the Destination CID field of packets being sent to them."?

draft-ietf-quic-transport.md
```diff
@@ -712,23 +737,26 @@ sequence of frames, as described in {{frames}}.
...
+receiving a packet, each endpoint sets the Destination Connection ID to match
+the value of the Source Connection ID that they receive.
```
MikeBishop: Discussion of receiving multiple packets with differing Source CID fields might be warranted.  Do I set this value every time I receive a packet?  First time I see a change?  What if the change isn't until later?

draft-ietf-quic-transport.md
```diff
@@ -712,23 +737,26 @@ sequence of frames, as described in {{frames}}.
...
-use it for all subsequent Handshake ({{packet-handshake}}) and 1-RTT
-({{packet-protected}}) packets but not for 0-RTT packets ({{packet-protected}}).
+The very first packet sent by a client client includes a random value for
+Destination Connection ID.  The same value MUST be used for all 0-RTT packets
+sent on that connection ({{packet-protected}}).  This randomized value is used
+to determine the handshake packet protection keys (see Section 5.2.2 of
+{{QUIC-TLS}}).
 
 Server's Version Negotiation ({{packet-version}}) and Retry ({{packet-retry}})
 packets MUST use connection ID selected by the client.
```
MikeBishop: I know you didn't touch this, but it's missing a "the."  Also, should this say "both Source and Destination CIDs"?

draft-ietf-quic-transport.md
MikeBishop: With the `omit` setting removed, I think there is no more "did not require Connection IDs," only connections where the Connection ID is zero-length.  At that point, this is a duplicate of CID collision.

In that case, the language can be something like "If the Destination Connection ID matches multiple connections (such as because they all have zero-length connection IDs), the address/port tuple is used to identify the correct connection.  Endpoints MUST drop packets with zero-length connection IDs if they cannot be matched to a connection."

draft-ietf-quic-transport.md
```diff
@@ -1498,31 +1513,42 @@ being used for denial of service.  See {{migrate-validate}} for details.
...
+recovers the packet number by adding the cumulative packet number gap to its
+expected packet number.  An endpoint MUST discard packets that contain a smaller
+gap than it advertised.
+
+An endpoint that receives a packet from a new address with either a previously
+unused connection ID or an empty connection ID MUST use the next available
+connection ID for any packets it sends to that address.  Failing to do this
+would allow for use of that connection ID to link activity on new paths.  An
+endpoint SHOULD use a new connection ID if it observes a change in address
+without a change in connection ID.
```
MikeBishop: To avoid linkability, we also should say that if you observe a change in the peer's connection ID, you should also change CIDs yourself.  Otherwise, the connection IDs in the reverse direction provide correlation.

draft-ietf-quic-invariants.md
```diff
@@ -150,10 +150,25 @@ version-specific semantics are marked with an X.
 {: #fig-long title="QUIC Long Header"}
 
 A QUIC packet with a long header has the high bit of the first octet set to 1.
+All other bits in that octet are version specific.
+
+The next octet contains the length in octets of the two Connection IDs (see
+{{connection-id}}) that follow.  Each length is encoded as a 4 bit unsigned
+integer.  The length of the Destination Connection ID occupies the high bits of
+the octet and the length of the Source Connection ID occupying the low bits of
+the octet.  An encoded length of 0 indicates that the connection ID is also 0
```
ianswett: I think there are valid use cases for shorter connection IDs(ie: 4 bytes in particular), so I'd be biased towards using 5 bits for each length.  I'm not sure if I'd want to overlap this with the first byte to reduce the number of bits we'd want to grease or just deal with randomizing 3 bits in each byte.

draft-ietf-quic-transport.md
```diff
@@ -2320,15 +2352,28 @@ Sequence:
   the value selected during the handshake comes immediately before the first
   value that a server can send.
 
+Length:
+
+: An 8-bit unsigned integer containing the length of the connection ID.  Values
+  less than 8 and greater than 22 are invalid and MUST be treated as a
+  connection error of type PROTOCOL_VIOLATION.
```
MikeBishop: You have the room here to spell it out, but there's something to be said for using the same encoding other places the length is written.  Maybe just use four bits and zero/ignore the upper four?

draft-ietf-quic-invariants.md
```diff
@@ -150,10 +150,25 @@ version-specific semantics are marked with an X.
...
-A QUIC packet with a long header has two fixed fields immediately following the
-first octet: a 64-bit Connection ID (see {{connection-id}}) and a 32-bit Version
-(see {{version}}).
+The connection ID lengths are followed by connection IDs.  The connection ID
+selected by the recipient of the packet (the Destination Connection ID) is
+followed by the connection ID selected by the sender of the packet (the Source
+Connection ID).
+
+After both Connection IDs, a 32-bit Version (see {{version}}) is followed by a
+version-specific payload.
```
ianswett: Not in this PR, but I think moving to varlen connection IDs motivates moving the version in front of the connection IDs, because then it's in a fixed spot, which is nice.

Previously the connection ID was always in a fixed spot, but now it's offset changes by 1 byte based on long/short header anyway, so we might as well have it move by 5 bytes instead of 1.  Conveniently, an 8 byte connection ID fits into a typical 16 byte cacheline in either location.

This isn't to say that the connection ID format in long headers should be removed from the invariants, because I think that turns out to be impractical.

draft-ietf-quic-transport.md
```diff
@@ -594,10 +607,14 @@ off-path attackers.
 The Initial packet uses long headers with a type value of 0x7F.  It carries the
 first cryptographic handshake message sent by the client.
 
-The client populates the connection ID field with randomly selected values,
-unless it has received a packet from the server.  If the client has received a
-packet from the server, the connection ID field uses the value provided by the
-server.
+If the client has not previously received a Retry packet from the server, it
+populates the Destination Connection ID field with a randomly selected value.
+This MUST NOT be zero length, but can be any other length.  If the client
```
ianswett: nit: I think it reads more cleanly to say "This MUST be at least 8 bytes."

draft-ietf-quic-transport.md
MikeBishop: For short headers, an endpoint can simplify its job by always using the same length CID, but if it doesn't then it probably needs to keep the connections in a tree by CID.  Check the first byte, see if there are any CIDs that begin with that byte, and so on, until it either finds a complete match or a lookup failure.

Then, if there's a lookup failure, it might have been zero-length and you were accidentally reading the packet number and encrypted payload as a CID.  So now you look into your separate bag of connections where the CID is zero-length and try to match by tuple.

But wait!  You also might have had a collision between the packet number and a CID, so packets which can't be validated by the selected connection *also* need to be checked against the zero-length set.  And we pray that doesn't happen very often, because this seems like the worst-case path through the code.

Ick.

draft-ietf-quic-invariants.md
```diff
@@ -150,10 +150,25 @@ version-specific semantics are marked with an X.
...
+the octet.  An encoded length of 0 indicates that the connection ID is also 0
+octets in length.  Non-zero encoded lengths are increased by 7 to get the full
+length of the connection ID.  For example, an octet with the value 0xa0
+describes a 17 octet Destination Connection ID and a zero octet Source
+Connection ID.
 
-A QUIC packet with a long header has two fixed fields immediately following the
-first octet: a 64-bit Connection ID (see {{connection-id}}) and a 32-bit Version
-(see {{version}}).
+The connection ID lengths are followed by connection IDs.  The connection ID
```
ekr: perhaps "by the connection IDs"

draft-ietf-quic-invariants.md
ekr: +1 to Ian here.

draft-ietf-quic-invariants.md
```diff
@@ -150,10 +150,25 @@ version-specific semantics are marked with an X.
...
+selected by the recipient of the packet (the Destination Connection ID) is
```
ekr: This isn't quite correct for the initial Initial packet. Maybe "associated with the recipient"

draft-ietf-quic-tls.md
```diff
@@ -552,7 +552,8 @@ older than 1.3 is negotiated.
 
 QUIC requires that the initial handshake packet from a client fit within the
 payload of a single packet.  The size limits on QUIC packets mean that a record
-containing a ClientHello needs to fit within 1171 octets.
+containing a ClientHello needs to fit within 1131 octets, though endpoints can
+reduce the size of their connection ID to increase by up to 22 octets.
```
ekr: You'll of course need to change this if you ahnge the length encoding.

draft-ietf-quic-tls.md
```diff
@@ -679,13 +680,13 @@ client packet protection key would use HKDF-Expand with an `info` parameter of
 ### Handshake Secrets {#handshake-secrets}
 
 Packets that carry the TLS handshake (Initial, Retry, and Handshake) are
-protected with a secret derived from the connection ID used in the client's
-Initial packet.  Specifically:
+protected with a secret derived from the Destination Connection ID field from
+the client's Initial packet.  Specifically:
```
ekr: Good catch.

draft-ietf-quic-tls.md
```diff
@@ -703,6 +704,22 @@ ensuring that the keys are different for each version of QUIC. This prevents a
...
+  that the server received its packet; the client has to rely on the exchange
+  that included the Retry packet for that property.
+
+Note(2):
+
+: HKDF-Extract uses HMAC {{?RFC2104}}, which produces the same output for values
+  ending in zero octets and values that omit those octets.  Thus,
+  0x12345678900000 and 0x1234567890 produce the same result.  A client is
+  required to randomize the value of Destination Connection ID, which avoids
+  this problem.
```
ekr: I believe this is incorrect.

HKDF-Extract is:

```
PRK = HMAC-Hash(salt, IKM)
```

In this case, salt is the handshake_salt (used as the HMAC key) and the CID is the IKM (used as the message. HMAC does right-pad the key with 0s (before XORing in the pad) but it does not handle the message that way.

draft-ietf-quic-transport.md
```diff
@@ -288,16 +289,12 @@ processing.
...
-connections to survive changes to the client's IP and port, such as those caused
-by NAT rebindings or by the client changing network connectivity to a new
-address. QUIC provides automatic cryptographic verification of a rebound client,
-since the client continues to use the same session key for encrypting and
-decrypting packets. The consistent connection ID can be used to allow migration
-of the connection to a new server IP address as well, since the Connection ID
-remains consistent across changes in the client's and the server's network
-addresses.
+QUIC connections carry a connection ID, an opaque value assigned by the intended
+recipient of the packet.  Using a consistent connection ID allows connections to
```
ekr: Technically, QUIC packets carry a connection ID.

draft-ietf-quic-transport.md
```diff
@@ -3987,6 +4031,9 @@ Issue and pull request numbers are listed with a leading octothorp.
 - Added PATH_CHALLENGE and PATH_RESPONSE frames to replace PING with Data and
   PONG frame. Changed ACK frame type from 0x0e to 0x0d. (#1091, #1086)
 
+- Made connection IDs symmetrical and variable length.  Removed
```
ekr: The fact that I called this "asymmetric" and you called it "symmetrical" suggests that both names are bad. Maybe "unidirectional" is the term we want?

draft-ietf-quic-invariants.md
```diff
@@ -135,12 +135,12 @@ version-specific semantics are marked with an X.
 ~~~
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
-+-+-+-+-+-+-+-+-+
-|1|X X X X X X X|
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|1|X X X X X X X| ConnID Len(8) |
```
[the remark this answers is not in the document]
martinthomson: I guess so.  This seemed a little easier to understand without having to read the text that follows.

draft-ietf-quic-invariants.md
martinthomson: I actually think that shorter connection IDs is harmful.  If we are to encourage the use of NEW_CONNECTION_ID, the space needs to be large enough to make the use of a new identifier meaningful.

draft-ietf-quic-invariants.md
martinthomson: Yep, I think that's a good idea.  We should make that move sooner rather than later if we do.

draft-ietf-quic-invariants.md
martinthomson: I thought carefully about this.

Shorter values make it much harder to come up with a scheme for preserving unlinkability if you use NEW_CONNECTION_ID.  Thus, I want to prohibit 1 octet connection IDs.  2 seems similarly challenged.

draft-ietf-quic-tls.md
martinthomson: Yeah, I'm close to killing this sentence, but it's somewhat useful to know where the limits are.

draft-ietf-quic-tls.md
martinthomson: Ahh, yes, the key is a problem, not the IKM.  I've never understood the ordering choice in the design of HKDF.

draft-ietf-quic-transport.md
martinthomson: Difficult.  A complete redo maybe.

draft-ietf-quic-transport.md
martinthomson: An understandable misinterpretation.

Client sets destination to random, source to whatever it needs.
In VN, server sets destination to client's source and source to client's destination.

Thus, it's very hard to explain.  And clearly I failed.

draft-ietf-quic-transport.md
martinthomson: Good catch.  I've put in a rule for this.  I think that it's a sensible rule, but you can judge that for yourself.

draft-ietf-quic-transport.md
martinthomson: "don't do that"

draft-ietf-quic-transport.md
martinthomson: There's a good reason to make the headers we expose to the cesspit use a crappy encoding, but less so here.

draft-ietf-quic-transport.md
martinthomson: I did that intentionally :)  But the changelog can be better.  I might avoid all those words and say what they are instead.

draft-ietf-quic-transport.md
```diff
@@ -1498,31 +1523,42 @@ being used for denial of service.  See {{migrate-validate}} for details.
...
+gap associated with the connection ID that it does use.
+
+An endpoint that receives a packet that is marked with a new connection ID
+recovers the packet number by adding the cumulative packet number gap to its
+expected packet number.  An endpoint MUST discard packets that contain a smaller
+gap than it advertised.
+
+An endpoint that receives a packet either with a previously unused connection ID
+or with an empty connection ID and from a new remote address MUST use the next
+available connection ID for any packets it sends to that address.  Failing to do
```
mirjak: What happens if there is no new connection ID available (anymore)?

draft-ietf-quic-invariants.md
```diff
@@ -178,16 +191,17 @@ version-specific semantics are marked with an X.
 
 A QUIC packet with a short header has the high bit of the first octet set to 0.
 
-A QUIC packet with a short header includes an optional connection ID and no
-version field.  The second bit of that octet (that is, 0x40) determines whether
-the connection ID is present.  If the second bit is cleared, a 64-bit connection
-ID immediately follows the first octet.  If the second bit is set, the remainder
-of the packet has version-specific semantics.
+A QUIC packet with a short header includes a Destination Connection ID.  The
```
mirjak: Just double checking: The short header does actually not include a connection ID if the length was zero in the long header, right?

draft-ietf-quic-transport.md
```diff
@@ -2320,15 +2362,28 @@ Sequence:
   the value selected during the handshake comes immediately before the first
   value that a server can send.
 
+Length:
+
+: An 8-bit unsigned integer containing the length of the connection ID.  Values
+  less than 8 and greater than 22 are invalid and MUST be treated as a
+  connection error of type PROTOCOL_VIOLATION.
```
mirjak: What is actually the use case for changing the connection ID length during a connection? Is that really needed?

draft-ietf-quic-invariants.md
```diff
@@ -154,18 +154,18 @@ All other bits in that octet are version specific.
...
+ID is also 0 octets in length.  Non-zero encoded lengths are increased by 7 to
+get the full length of the connection ID.  For example, an octet with the value
+0xa0 describes a 17 octet Destination Connection ID and a zero octet Source
 Connection ID.
 
-The connection ID lengths are followed by connection IDs.  The connection ID
-selected by the recipient of the packet (the Destination Connection ID) is
-followed by the connection ID selected by the sender of the packet (the Source
-Connection ID).
+The connection ID lengths are followed by a two connection IDs.  The connection
```
MikeBishop: by ~a~ two

draft-ietf-quic-transport.md
MikeBishop: "When an endpoint changes network connectivity, QUIC provides mechanisms to verify new addresses while maintaining session continuity.", maybe?

draft-ietf-quic-transport.md
```diff
@@ -712,26 +737,39 @@ sequence of frames, as described in {{frames}}.
...
-packets MUST use connection ID selected by the client.
+A connection ID is used to ensure consistent routing of packets.  The long
+header contains two connection IDs: the Destination Connection ID is chosen by
+the recipient of the packet and is used to provide consistent routing; the
+Source Connection ID is used to set the Destination Connection ID used by a
+peer.
+
+During the handshake, packets with the long header are used to establish the
+Destination Connection ID that each endpoint uses.  Each endpoint uses the
+Source Connection ID field to communicate the connection that is used in the
```
MikeBishop: "communicate the connection that is used" doesn't read properly to me.  Should this be "connection ID"?

draft-ietf-quic-transport.md
```diff
@@ -1498,31 +1523,42 @@ being used for denial of service.  See {{migrate-validate}} for details.
...
-packet number.  A server SHOULD discard packets that contain a smaller gap than
-it advertised.
+to correlate activity between those paths.  An endpoint that moves between
+networks might not wish to have their activity correlated by any entity other
+than a server. The NEW_CONNECTION_ID message can be sent by both endpoints to
+provide an unlinkable connection ID for use in case a peer wishes to explicitly
+break linkability between two points of network attachment.
+
+An endpoint might need to send packets on multiple networks without receiving
+any response from the server.  To ensure that the endpoint is not linkable
```
MikeBishop: You're mixing "endpoint/peer" and "client/server" language here.  While it's (currently) correct that the endpoint on the previous line must be a client, you might want to pick one or the other for clarity.

draft-ietf-quic-transport.md
```diff
@@ -1498,31 +1523,42 @@ being used for denial of service.  See {{migrate-validate}} for details.
...
+number gaps for connection IDs that it skips in addition to the packet number
+gap associated with the connection ID that it does use.
+
+An endpoint that receives a packet that is marked with a new connection ID
+recovers the packet number by adding the cumulative packet number gap to its
+expected packet number.  An endpoint MUST discard packets that contain a smaller
+gap than it advertised.
+
+An endpoint that receives a packet either with a previously unused connection ID
+or with an empty connection ID and from a new remote address MUST use the next
```
MikeBishop: If it's an empty connection ID and a new remote address, how would the packet even be associated with the connection?  I think this is a normative requirement for a situation which can't occur.

draft-ietf-quic-transport.md
MikeBishop: A few reasonable choices would be to continue with the current CID (it's the peer's fault for letting you run out, so they've lost their linkability), or to prohibit sending until you get a NEW_CONNECTION_ID frame.  The latter has the nasty corner case that, if you're prohibited from sending ACKs and the packet containing your new CID gets dropped, you can't make forward progress until you RTO.

draft-ietf-quic-transport.md
```diff
@@ -3987,6 +4041,10 @@ Issue and pull request numbers are listed with a leading octothorp.
 - Added PATH_CHALLENGE and PATH_RESPONSE frames to replace PING with Data and
   PONG frame. Changed ACK frame type from 0x0e to 0x0d. (#1091, #1086)
 
+- Endpoint now set the connection ID that their peer uses.  Connection IDs are
```
MikeBishop: Endpoint**s**.

draft-ietf-quic-tls.md
ekr: AIUI, the idea is that HMAC is a PRF, with the IKM being the message,  so this whitens the IKM.

draft-ietf-quic-tls.md
ekr: @MikeBishop: as noted above, I don't think this is necessary.

draft-ietf-quic-invariants.md
martinthomson: Yep, either that or it includes a zero-length connection ID :)

draft-ietf-quic-transport.md
martinthomson: Hmm, only if it does trial decryption for all connections.  Probably unrealistic.  Trimming "with an empty connection ID and" out.

draft-ietf-quic-transport.md
martinthomson: Opened #1163

draft-ietf-quic-transport.md
martinthomson: Well, if you are migrating to new load balancing infrastructure, it might help.  Note that you probably need an indicator of length in the connection ID itself to make this work, or a period where the common prefix is distinct.

Basically, I couldn't see any reason not to allow this, given that it's the endpoint that sends NEW_CONNECTION_ID that absorbs the complexity cost.

draft-ietf-quic-transport.md
igorlord: Please, drop: "or from a new remote address".  No "either" there.  Otherwise, you are setting yourself up for an attack, where the attacker just replays the same packet from many different addresses toward a peer, draining the list of connection IDs that peer has.  Then, when there is a genuine connection migration, the peer has no connection IDs left.

draft-ietf-quic-transport.md
```diff
@@ -1498,31 +1523,42 @@ being used for denial of service.  See {{migrate-validate}} for details.
...
+An endpoint that receives a packet that is marked with a new connection ID
+recovers the packet number by adding the cumulative packet number gap to its
+expected packet number.  An endpoint MUST discard packets that contain a smaller
+gap than it advertised.
+
+An endpoint that receives a packet either with a previously unused connection ID
+or with an empty connection ID and from a new remote address MUST use the next
+available connection ID for any packets it sends to that address.  Failing to do
+this would allow for use of that connection ID to link activity on new paths.
+An endpoint SHOULD use a new connection ID if it observes a change in address
```
igorlord: See above why SHOULD should be SHOULD NOT.  I.e. you do not want an attacker resend the same packet from many src IPs.  In any case, the linkability cause is probably lost, if even one packet has made it from one peer to another with an old connection ID over the new path (even in one direction).

draft-ietf-quic-transport.md
igorlord: Thanks for updating. The level of control required is not much, really, especially if the connection is over ipv6 (ie. the number of unique ip addresses under attacker's control is enormous).

The ability to collect and correlate data from multiple independent networks to follow nodes across migrations assumes a rather awesome level of control and resources already.

draft-ietf-quic-transport.md
```diff
@@ -1528,31 +1560,44 @@ being used for denial of service.  See {{migrate-validate}} for details.
...
+gap than it advertised.
+
+An endpoint that receives a successfully authenticated packet with a previously
+unused connection ID MUST use the next available connection ID for any packets
+it sends to that address.  To avoid changing connection IDs multiple times when
+packets arrive out of order, endpoints MUST change only in response to a packet
+that increases the largest received packet number.  Failing to do this could
+allow for use of that connection ID to link activity on new paths.  There is no
+need to move to a new connection ID if the address of a peer changes without
+also changing the connection ID.
```
mikkelfj: I'm not fully up to speed on connection migration, so bear with me, but...

What if both endpoints migrate? For example B has sent two NEW_ID_B(1,2) to A, and A sends two NEW_ID_A(1,2) to B. A migrates and consumes the first NEW_ID_B(1). Then B migrates and consumes NEW_ID_A(1). But because either end has migrated, they are now on new infrastructure and the load balancers no longer understand NEW_ID_A(2) or NEW_ID_B(2).

It seems to me that you cannot have a sequential NEW_ID - assuming an endpoint only migrates to one location at a time (no multipath) there must a be a latest NEW_ID that will work, but the older might not. It is just trial an error until something works?

mikkelfj:
Maybe I'm just not reading careful enough, but could we perhaps change the language slightly - despite that I have been thinking about this problem for a long time, I still got confused. There are 3 different connection ID's. The initial connection ID chosen randomly by client, and the client chosen ID, and the server chosen ID. Each of these are sometimes a Destination ID or a Source ID depending on direction.

In some place there is a wording along "Even if the server changes the ID, the same ID must be used". It would be helpful to clarify which kind of ID and exactly for how long.

draft-ietf-quic-transport.md
martinthomson: Migration happens with short headers, so there is usually one connection ID in each direction.

For instance, the client might be sending SCID3 and the server might send CCID2.  If the client moves, then it sends SCID4.  The server sees the new connection ID from the client and moves to sending CCID3.

Note that connection IDs are currently strictly sequential, but that is not a property we need to preserve if we encrypt packet numbers.  Any connection ID should work (with some caveats that I can go into more).

martinthomson:
Not following you there.  There are actually 4 discrete values for connection ID that might be in use during the handshake (the server can update its choice if it sends Retry).

I added this recently, does it help?

> During the handshake, an endpoint might receive multiple packets with the long
header, and thus be given multiple opportunities to update the Destination
Connection ID it sends.  An client MUST only change the value it sends in the
Destination Connection ID field in response to the first packet of each type
(Retry, or Handshake) that it receives; a server MUST only change its value
based on an Initial packet.  This avoids problems that might arise from
stateless processing of multiple Initial packets producing different connection
IDs.

mikkelfj:
Yes this clarifies the last part. I still wonder if might help to provide a name to the initial Destination ID, such as Client Probe ID, or whatever. Note typo "An client MUST".

draft-ietf-quic-transport.md
mikkelfj: Hmm - in the above example: If the server later moves, it uses CCID4 and the client sees this and sends SCID5. But SCID5 was send in a NEW_ID message before the server moved to use CCID4 and the server is now behind infrastructure that does not understand SCID5.

If the migration where the server sends CCID4 also marks the smallest valid SCID, there should be no issue - perhaps this is already the case.

martinthomson: It is the responsibility of the endpoint to ensure that any connection ID it provides to its peer is capable of reaching it, no matter where it moves to.  The only stipulation being that packets arrive at an address that the endpoint control.  Either the endpoint or intermediaries over which it has influence, like a load balancer.

draft-ietf-quic-transport.md
```diff
@@ -404,26 +401,44 @@ Long Packet Type:
...
-  use of this field in more detail.
+Destination Connection ID:
+
+: The Destination Connection ID field starts at octet 2 and is either 0 octets
+  in length or between 4 and 18 octets. {{connection-id}} describes the use of
+  this field in more detail.
+
+Source Connection ID:
+
+: The Source Connection ID field starts at octet 2 and is either 0 octets in
```
ianswett: They can't both start at octet 2, unless the destination Connection ID is 0 length.

draft-ietf-quic-transport.md
```diff
@@ -732,26 +764,39 @@ sequence of frames, as described in {{frames}}.
...
+Source Connection ID field to specify the connection ID that is used in the
+Destination Connection ID field of packets being sent to them.  Upon receiving a
+packet, each endpoint sets the Destination Connection ID it sends to match the
+value of the Source Connection ID that they receive.
+
+During the handshake, an endpoint might receive multiple packets with the long
+header, and thus be given multiple opportunities to update the Destination
+Connection ID it sends.  A client MUST only change the value it sends in the
+Destination Connection ID field in response to the first packet of each type
+(Retry, or Handshake) that it receives; a server MUST only change its value
```
ianswett: I think an algorithm based on packet number might be more robust than specifying the first packet of each type?  Largest packet number wins?

Also, I would say the server MUST set it's value based on an Initial packet.  Given there was no connection before, I wouldn't call it a change of connection ID.

draft-ietf-quic-transport.md
```diff
@@ -1502,13 +1534,13 @@ connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
...
-QUIC connections are identified by their 64-bit Connection ID.  QUIC's
-consistent connection ID allows connections to survive changes to the client's
-IP and/or port, such as those caused by client or server migrating to a new
-network.  Connection migration allows a client to retain any shared state with a
-connection when they move networks.  This includes state that can be hard to
-recover such as outstanding requests, which might otherwise be lost with no easy
-way to retry them.
+A QUIC connection can be identified using the Destination Connection ID included
+in packets.  QUIC's consistent connection ID allows connections to survive
+changes to the client's IP and/or port, such as those caused by client or server
```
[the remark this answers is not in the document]
MikeBishop: It might in the future, but server connection migration isn't defined.  Certainly they pave the way, though.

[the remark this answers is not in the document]
ianswett: SGTM, ignore this comment for now.

draft-ietf-quic-transport.md
```diff
@@ -163,7 +163,8 @@ Connection:
 
 Connection ID:
```
janaiyengar: s/Connection ID/Destination Connection ID/

draft-ietf-quic-transport.md
```diff
@@ -163,7 +163,8 @@ Connection:
...
 
-: The 64-bit unsigned number used as an identifier for a QUIC connection.
+: An opaque identifier that is used to identify a QUIC connection.  Each
```
janaiyengar: This sentence sounds incomplete now. How about "to identify a QUIC connection at an endpoint"?

draft-ietf-quic-transport.md
```diff
@@ -404,26 +401,44 @@ Long Packet Type:
...
+DCIL and SCIL:
+
+: Octet 1 contains the lengths of the two connection ID fields that follow it.
+  These lengths are encoded as two 4-bit unsigned integers. The Destination
+  Connection ID Length (DCIL) field occupies the 4 high bits of the octet and
+  the Source Connection ID Length (SCIL) field occupies the 4 low bits of the
+  octet.  An encoded length of 0 indicates that the connection ID is also 0
+  octets in length.  Non-zero encoded lengths are increased by 3 to get the full
+  length of the connection ID, producing a length between 4 and 18 octets
+  inclusive.  For example, an octet with the value 0x50 describes an 8 octet
```
janaiyengar: 8-octet

draft-ietf-quic-transport.md
```diff
@@ -404,26 +401,44 @@ Long Packet Type:
...
+  Destination Connection ID and a zero octet Source Connection ID.
```
janaiyengar: zero-octet

draft-ietf-quic-transport.md
janaiyengar: +1. I would remove the starting position.

draft-ietf-quic-transport.md
```diff
@@ -577,12 +589,20 @@ The layout of a Version Negotiation packet is:
...
-of 32-bit versions which the server supports.
+The value in the Unused field is selected randomly by the server.
+
+The server MUST include the value from the Source Connection ID field of the
+packet it receives in the Destination Connection ID field.  The value for Source
+Connection ID MUST be copied from the Destination Connection ID of the received
+packet, which is initially randomly selected by a client.  Echoing both
+connection IDs gives clients some assurance that the server received the packet
+and that the Version Negotiation packet was not generated by an off-path
+attacker.
```
janaiyengar: This text is great. Can you use this text above in the VN section of the invariants draft?

draft-ietf-quic-invariants.md
```diff
@@ -257,9 +272,12 @@ Version Negotiation packets do not use integrity or confidentiality protection.
...
 
-The Connection ID field in a Version Negotiation packet contains the Connection
-ID from the packet that was received.  This provides some protection against
-injection of Version Negotiation packets by off-path attackers.
+The Connection ID fields in a Version Negotiation packet contain the Connection
+ID values from the packet that was received by the server, inverted to ensure
+correct routing of the packet toward the client based on the Destination
+Connection ID.  Echoing the value of the client's Destination Connection ID in
+the Source Connection ID field can provide some protection against injection of
+Version Negotiation packets by off-path attackers.
```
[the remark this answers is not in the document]
janaiyengar: The text in the VN Packet section of the transport draft is great. I would simply use that same text here.

draft-ietf-quic-transport.md
```diff
@@ -614,10 +634,14 @@ off-path attackers.
 The Initial packet uses long headers with a type value of 0x7F.  It carries the
 first cryptographic handshake message sent by the client.
 
-The client populates the connection ID field with randomly selected values,
-unless it has received a packet from the server.  If the client has received a
-packet from the server, the connection ID field uses the value provided by the
-server.
+If the client has not previously received a Retry packet from the server, it
+populates the Destination Connection ID field with a randomly selected value.
```
janaiyengar: Is this a single random value for all packets until one is received from teh server? Or is it ok for the client to send a new random in each packet?

draft-ietf-quic-transport.md
```diff
@@ -678,8 +707,10 @@ A Handshake packet uses long headers with a type value of 0x7D.  It is
 used to carry acknowledgments and cryptographic handshake messages from the
 server and client.
 
-The connection ID field in a Handshake packet contains a connection ID
-that is chosen by the server (see {{connection-id}}).
+The Destination Connection ID field in a Handshake packet contains a connection
+ID that is chosen by the recipient of the packet; the Source Connection ID
```
janaiyengar: The server's Initial packet also carries a dest connection ID chosen by the client, right?

draft-ietf-quic-transport.md
```diff
@@ -712,12 +743,13 @@ packet types explicitly indicate the encryption level and therefore the keys
...
 Packets protected with 0-RTT keys use a type value of 0x7C.  The connection ID
-field for a 0-RTT packet is selected by the client.
+fields for a 0-RTT packet MUST match the values used in the Initial packet
+({{packet-initial}}).
 
 The client can send 0-RTT packets after receiving a Handshake packet
 ({{packet-handshake}}), if that packet does not complete the handshake.  Even if
 the client receives a different connection ID in the Handshake packet, it MUST
-continue to use the connection ID selected by the client for 0-RTT packets, see
+continue to use the same Destination Connection ID for 0-RTT packets, see
```
janaiyengar: What should a receiver do if a subsequent packet is received with a different source connection ID?

draft-ietf-quic-transport.md
```diff
@@ -732,26 +764,39 @@ sequence of frames, as described in {{frames}}.
...
+A connection ID is used to ensure consistent routing of packets.  The long
+header contains two connection IDs: the Destination Connection ID is chosen by
+the recipient of the packet and is used to provide consistent routing; the
+Source Connection ID is used to set the Destination Connection ID used by a
+peer.
+
+During the handshake, packets with the long header are used to establish the
+Destination Connection ID that each endpoint uses.  Each endpoint uses the
+Source Connection ID field to specify the connection ID that is used in the
+Destination Connection ID field of packets being sent to them.  Upon receiving a
```
janaiyengar: remove "field"

draft-ietf-quic-transport.md
```diff
@@ -732,26 +764,39 @@ sequence of frames, as described in {{frames}}.
...
-packets MUST use connection ID selected by the client.
+A connection ID is used to ensure consistent routing of packets.  The long
+header contains two connection IDs: the Destination Connection ID is chosen by
+the recipient of the packet and is used to provide consistent routing; the
+Source Connection ID is used to set the Destination Connection ID used by a
+peer.
+
+During the handshake, packets with the long header are used to establish the
+Destination Connection ID that each endpoint uses.  Each endpoint uses the
+Source Connection ID field to specify the connection ID that is used in the
```
janaiyengar: remove "field"

draft-ietf-quic-transport.md
```diff
@@ -732,26 +764,39 @@ sequence of frames, as described in {{frames}}.
...
+Destination Connection ID field in response to the first packet of each type
+(Retry, or Handshake) that it receives; a server MUST only change its value
+based on an Initial packet.  This avoids problems that might arise from
+stateless processing of multiple Initial packets producing different connection
+IDs.
+
+Short headers only include the Destination Connection ID and omit the explicit
+length.
+
+The very first packet sent by a client client includes a random value for
```
janaiyengar: too many clients

draft-ietf-quic-transport.md
janaiyengar: +1. I can't reason about packet type changes easily, but largest packet number seems robust.

draft-ietf-quic-transport.md
```diff
@@ -1856,9 +1899,23 @@ following layout:
...
 
-A server copies the connection ID field from the packet that triggers the
-stateless reset.  A server omits the connection ID if explicitly configured to
-do so, or if the client packet did not include a connection ID.
+This design ensures that a stateless reset packet is - to the extent possible -
+indistinguishable from a regular packet.
+
+A server generates a random 18-octet Destination Connection ID field.  For a
+client that requires that the server include a connection ID, this will mean
+that this value differs from previous packets with two consequences:
```
janaiyengar: too many "that"s... and this is a bit of a run-on sentence.

draft-ietf-quic-transport.md
```diff
@@ -1856,9 +1899,23 @@ following layout:
...
+
+* The packet might not reach the client.  If the Destination Connection ID is
+  critical for routing toward the client, then this packet could be incorrectly
+  routed and dropped.  This causes the stateless reset to be ineffective in
```
janaiyengar: remove "and dropped".

draft-ietf-quic-transport.md
```diff
@@ -1856,9 +1899,23 @@ following layout:
...
+  causing errors to be quickly detected and recovered.  In this case, clients
+  will need to rely on other methods such as timers to detect that the
```
janaiyengar: "other methods - such as timers - to detect"

draft-ietf-quic-transport.md
```diff
@@ -1856,9 +1899,23 @@ following layout:
...
+  connection has failed.
+
+* The connection ID can be used by entities other than the client to identify
```
janaiyengar: "The randomly generated connection ID"

draft-ietf-quic-transport.md
```diff
@@ -1856,9 +1899,23 @@ following layout:
...
+  this as a potential stateless reset.  A server that occasionally uses
```
janaiyengar: Capitalize Stateless Reset

draft-ietf-quic-transport.md
```diff
@@ -2350,15 +2401,28 @@ Sequence:
   the value selected during the handshake comes immediately before the first
   value that a server can send.
 
+Length:
+
+: An 8-bit unsigned integer containing the length of the connection ID.  Values
+  less than 4 and greater than 18 are invalid and MUST be treated as a
+  connection error of type PROTOCOL_VIOLATION.
```
janaiyengar: It's a bit unfortunate that this length is a different encoding than DSIL and SSIL. Nothing to do about it I suppose.

draft-ietf-quic-transport.md
```diff
@@ -2350,15 +2401,28 @@ Sequence:
...
 
 Stateless Reset Token:
 
 : A 128-bit value that will be used to for a stateless reset when the associated
   connection ID is used (see {{stateless-reset}}).
 
+An endpoint MUST NOT send this frame if it currently requires that its peer send
+packets with a zero-length Destination Connection ID.  Changing the length of a
+connection ID to or from zero-length makes it very difficult to identify when
+the new connection ID was used.  An endpoint that is sending packets with a a
```
janaiyengar: I don't understand this sentence. Which is the "new connection ID"? I'm inclined to remove this sentence entirely.

janaiyengar: too many articles ("with a a")

draft-ietf-quic-transport.md
mikkelfj: There is a discussion on list about version negotiation attacks by man-on-the-side. One proposed solution was to queue some packets and wait for the best outcome. I'm not so sure what is best, but it overlaps with the considerations in the above text and comments.

draft-ietf-quic-transport.md
```diff
@@ -732,26 +764,39 @@ sequence of frames, as described in {{frames}}.
...
+length.
+
+The very first packet sent by a client client includes a random value for
+Destination Connection ID.  The same value MUST be used for all 0-RTT packets
+sent on that connection ({{packet-protected}}).  This randomized value is used
+to determine the handshake packet protection keys (see Section 5.2.2 of
+{{QUIC-TLS}}).
+
+A Version Negotiation ({{packet-version}}) packet MUST use both connection IDs
+selected by the client, inverted to ensure correct routing toward the client.
```
mikkelfj: inverted -> swapped, mirrored, reflected?

draft-ietf-quic-invariants.md
```diff
@@ -178,16 +192,17 @@ version-specific semantics are marked with an X.
...
 A QUIC packet with a short header has the high bit of the first octet set to 0.
 
-A QUIC packet with a short header includes an optional connection ID and no
-version field.  The second bit of that octet (that is, 0x40) determines whether
-the connection ID is present.  If the second bit is cleared, a 64-bit connection
-ID immediately follows the first octet.  If the second bit is set, the remainder
-of the packet has version-specific semantics.
+A QUIC packet with a short header includes a Destination Connection ID.  The
+short header does not include the Connection ID Lengths, Source Connection ID,
+or Version fields.
```
siyengar: Two sentences like 

"A load balancer may uses variable length connection IDs for different clients in order to grease the length. Even though the short header does not have an explicit length field for the connection ID, a load balancer could encode the length of the connection ID in the connection ID with it's own encoding scheme"

might be useful to clarify the intention of not having length fields in the short headers

draft-ietf-quic-transport.md
```diff
@@ -451,11 +467,9 @@ following sections.
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+
-|0|C|K|1|0|T T T|
+|0|K|1|1|0|T T T|
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                                                               |
-+                     [Connection ID (64)]                      +
-|                                                               |
+|                Destination Connection ID (0..148)           ...
```
kazuho: Maybe this needs be corrected to `Destination Connection ID (0/32..144)`?

I see the same issue in the definition of the Stateless Reset (it says 148) and also in short header definition in the invariants draft.

draft-ietf-quic-invariants.md
ianswett: This seems potentially helpful to me.

draft-ietf-quic-transport.md
```diff
@@ -1115,7 +1166,6 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
       initial_max_data(1),
       initial_max_stream_id_bidi(2),
       idle_timeout(3),
-      omit_connection_id(4),
       max_packet_size(5),
       stateless_reset_token(6),
       ack_delay_exponent(7),
```
mikkelfj: Renumber?

draft-ietf-quic-transport.md
martinthomson: Largest packet number fails for the simple reason that it might allow for multiple changes.  Lowest packet number works better.  And that's all I'm looking for.  Latching is what we are looking for, not a continuous sequence of changes.

Also, I added a note to ignore packets with different Source Connection ID values.  That deals neatly with the problem where two server instances both attempt to start a connection with you as a result of you sending multiple Initial packets.

draft-ietf-quic-transport.md
martinthomson: Done.

draft-ietf-quic-transport.md
martinthomson: Oh, I had a whole thing about multiple random values being great for measuring RTT then I remembered that you don't really want to be stuck with trial decryption for the server's first packet.

draft-ietf-quic-transport.md
martinthomson: Servers don't send Initial packets.

draft-ietf-quic-transport.md
martinthomson: I added text that says that you discard those.

draft-ietf-quic-transport.md
martinthomson: Actually I want to add field at least once.  When "Destination Connection ID" is used, it always refers to the field.  "Connection ID" refers to the concept.  Is that OK?

mikkelfj:
Just a thought:

Middleware or man-in-the-middle could replace the Initial Random Connection ID. This could be done to direct traffic further into the infrastructure or it could be a DoS attack. This may not be a sufficiently strong case to do anything about it considering all else that is possible to do at this point, and it could even be beneficial on occasion. But there is the option of having the client verifying it via TP or similar - e.g. as part of crypto key derivation - I don't think this is currently the case, though I'm not sure.

draft-ietf-quic-transport.md
martinthomson: Maybe later.

draft-ietf-quic-transport.md
martinthomson: It isn't Germanic anywhere else.

draft-ietf-quic-transport.md
martinthomson: This requirement seemed to need justification, that's all.

martinthomson:
A server has to treat Initial with suspicion anyway.  Nothing new in that.

draft-ietf-quic-invariants.md
mirjak: or in the manageability statement...?

mirjak:
It also probably something to discuss in the manageability statement, no?

draft-ietf-quic-invariants.md
martinthomson: I think that text on how to construct a connection ID that has the sorts of properties we care about (unlinkability, etc...) would be good things to add to that sort of draft.

draft-ietf-quic-transport.md
```diff
@@ -779,32 +777,37 @@ Source Connection ID is used to set the Destination Connection ID used by a
...
 
 During the handshake, an endpoint might receive multiple packets with the long
 header, and thus be given multiple opportunities to update the Destination
 Connection ID it sends.  A client MUST only change the value it sends in the
-Destination Connection ID field in response to the first packet of each type
-(Retry, or Handshake) that it receives; a server MUST only change its value
-based on an Initial packet.  This avoids problems that might arise from
-stateless processing of multiple Initial packets producing different connection
-IDs.
+Destination Connection ID in response to the first packet of each type (Retry,
```
ianswett: I'm still concerned this 'first packet of each type' rule ends up being more complex to reason about than the connection ID on the largest processed packet number.

draft-ietf-quic-transport.md
```diff
@@ -2326,9 +2390,7 @@ The NEW_CONNECTION_ID is as follows:
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                          Sequence (i)                       ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                                                               |
-+                        Connection ID (64)                     +
-|                                                               |
+|   Length (8)  |          Connection ID (32..144)            ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
nibanks: Does it really make sense to support multiple different length Connection IDs for the same connection? Why not just require all Connection IDs be the same length as the original, and then remove this length field altogether? I'm just thinking of the additional complexity of multiple different length Connection IDs, per connection, would be for hardware offload.

draft-ietf-quic-transport.md
```diff
@@ -732,26 +770,44 @@ sequence of frames, as described in {{frames}}.
...
-use it for all subsequent Handshake ({{packet-handshake}}) and 1-RTT
-({{packet-protected}}) packets but not for 0-RTT packets ({{packet-protected}}).
-
-Server's Version Negotiation ({{packet-version}}) and Retry ({{packet-retry}})
-packets MUST use connection ID selected by the client.
+A connection ID is used to ensure consistent routing of packets.  The long
+header contains two connection IDs: the Destination Connection ID is chosen by
+the recipient of the packet and is used to provide consistent routing; the
+Source Connection ID is used to set the Destination Connection ID used by a
+peer.
```
nibanks: Personally, I think changing `a peer` to `the peer` sounds better.

draft-ietf-quic-transport.md
mikkelfj: I tend to agree. If it were the case that a connection ID was path specific it would be a different matter, but it is required to work for all paths. It could be argued that removing the ability to change length would complicate multi-path design later on - but that is hard to predict.

Another point if view is that a connection ID might be related to infrastructure provided and it might change suddenly for a large number of live connections. But still, it seems a bit contrived.

draft-ietf-quic-invariants.md
```diff
@@ -165,11 +181,9 @@ version-specific semantics are marked with an X.
...
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+
-|0|C|X X X X X X|
+|0|X X X X X X X|
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                                                               |
-+                     [Connection ID (64)]                      +
-|                                                               |
+|                Destination Connection ID (0..144)           ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
mirjak: Given the dest conn ID is variable length now, can we maybe have the packet number before the connection id in the header?

nibanks: Packet number isn't an invariant. So it cannot precede any invariant field (Connection ID).

MikeBishop: Technically, it could, but there would have to be an invariant way of determining the offset of the Connection ID field.

ianswett: As Nick implied, I think the best order is version, connection ID(s), then packet number, regardless of what we decide on encrypting packet numbers.

draft-ietf-quic-transport.md
```diff
@@ -577,12 +589,20 @@ The layout of a Version Negotiation packet is:
...
-allows clients some assurance that the server received the packet and that the
-Version Negotiation packet is in fact from the server.  The Version field MUST
-be set to 0x00000000.  The remainder of the Version Negotiation packet is a list
-of 32-bit versions which the server supports.
+The value in the Unused field is selected randomly by the server.
+
+The server MUST include the value from the Source Connection ID field of the
+packet it receives in the Destination Connection ID field.  The value for Source
+Connection ID MUST be copied from the Destination Connection ID of the received
+packet, which is initially randomly selected by a client.  Echoing both
```
janaiyengar: These first two sentences are redundant. I think you can remove the first sentence.

draft-ietf-quic-transport.md
```diff
@@ -732,26 +771,44 @@ sequence of frames, as described in {{frames}}.
...
+Connection ID that they receive.
+
+During the handshake, an endpoint might receive multiple packets with the long
+header, and thus be given multiple opportunities to update the Destination
+Connection ID it sends.  A client MUST only change the value it sends in the
+Destination Connection ID in response to the first packet of each type it
+receives from the server (Retry or Handshake); a server MUST set its value based
+on the Initial packet.  Any additional changes are not permitted; subsequent
+packets of those types include a different Source Connection ID, they MUST be
+discarded.  This avoids problems that might arise from stateless processing of
```
janaiyengar: This sentence needs grammar. I think you might be missing an "if" in "if subsequent packets of those types ..."

draft-ietf-quic-transport.md
```diff
@@ -732,26 +771,44 @@ sequence of frames, as described in {{frames}}.
...
+multiple Initial packets producing different connection IDs.
+
+Short headers only include the Destination Connection ID and omit the explicit
+length.  The length of the Destination Connecton ID field is expected to be
```
janaiyengar: s/Connecton/Connection/

draft-ietf-quic-transport.md
```diff
@@ -732,26 +771,44 @@ sequence of frames, as described in {{frames}}.
...
+known to endpoints.  Endpoints with a load balancer that uses a connection ID
+can agree to a fixed or minimum length for connection IDs with necessary
+information in the fixed portion.  If that fixed portion instead encodes an
+explicit length, the entire connection ID can vary in length and still be used
+by the load balancer.
```
janaiyengar: This part about load balancers is confusing. I would at least make it a separate paragraph. I do wonder if this is necessary ... it's basically a suggestion for how load balancers are implemented, and I would be more explicit about that in this para. Here's a suggested rephrasing:

"Endpoints using a connection-ID based load balancer could agree with the load balancer on a fixed or minimum length and on an encoding for connection IDs. This fixed portion could encode an explicit length, which allows the entire connection ID to vary in length and still be used by the load balancer."

draft-ietf-quic-transport.md
```diff
@@ -901,33 +958,30 @@ Incoming packets are classified on receipt.  Packets can either be associated
 with an existing connection, or - for servers - potentially create a new
 connection.
 
-Hosts try to associate the packet with an existing connection. If the
-packet has a connection ID corresponding to an existing connection, QUIC
+Hosts try to associate the packet with an existing connection. If the packet has
```
janaiyengar: s/the packet/a packet/

draft-ietf-quic-transport.md
```diff
@@ -901,33 +958,30 @@ Incoming packets are classified on receipt.  Packets can either be associated
...
-Initial packet with a server-generated connection ID and will be processed
-accordingly. Clients SHOULD discard any packets with new connection IDs that
-do not meet these criteria.
-
-Note that a successfully associated packet may be a Version Negotiation
-packet, which is handled in accordance with {{handle-vn}}.
+Valid packets sent to clients always include a Destination Connection ID that
+matches the value the client selects.  Clients that choose to receive
+zero-length connection IDs can use the address/port tuple to identify a
+connection.  Packets that don't match an existing connection can be discarded.
```
janaiyengar: s/can be discarded/MAY be discarded/

PR #1139, opened 2018-02-23
Generating Acknowledgements

ianswett:
Fixes the remaining portion of #230

PR #1153, opened 2018-03-04
A few proposed edits for #1139

martinthomson:


draft-ietf-quic-recovery.md
```diff
@@ -439,12 +439,11 @@ receiving a second full-sized packet.
 Out-of-order packets SHOULD be acknowledged more quickly, in order
 to accelerate loss recovery.  The receiver SHOULD send an immediate ACK
 when it receives a new packet which is not one greater than the
-largest received packet number.  If a receiver processes multiple
-packets before sending any ACK frames in response, they MAY not
-send an immediate ack if the received packets form a contiguous
-sequence starting one larger than the largest acked.
+largest received packet number.  A receiver MAY delay acknowledgment by up
+to its maximum ack delay as long as it receives a contiguous sequence
+starting at the largest acknowledged.
```
ianswett: So, what I was hoping to convey is that if you process 5 packets, and they don't arrive in order, but there end up being no new gaps after processing, then it's ok to not send an immediate ack.

In practice, this is a pretty good thing to do, particularly on higher bandwidth or lower powered devices.  Maybe this deserves a full paragraph with a clearer explanation?

martinthomson: Yes, that is what I understood.  If you want to add, "even if those packets don't arrive in sequence", that might make that clearer.

MikeBishop:
Since Ian partially incorporated this change and merged the parent PR, what's the correct parent/state of this PR?

ianswett:
I hope it's no longer necessary, but I'm open to changes to the current text.

martinthomson:
I trust Ian's judgment here.

Issue #1133, opened 2018-02-21
qcram document title should include the term "QCRAM"


reschke:

afrind:
It may be too late to ask this question, but I thought I heard someone say we should have reserved the name 'QPACK' for our final design choice.  Have we burned that bridge, and if not, should we  rename QCRAM to QPACK, especially given how similar QCRAM is to HPACK?

ianswett:
I was thinking the same thing last night in regards to renaming QCRAM to QPACK.  I think it may reduce the confusion about what QCRAM is.

MikeBishop:
HAHAHAhahahHAHahaha.......  🤣🤣🤣

MikeBishop:
More seriously, I have no objections.  The name is irrelevant to the technical merits, and if it improves the understanding of folks who weren't part of the design team, call it whatever you want.

krasic:
+1

afrind:
If no one objects, I'll make the change from QCRAM to QPACK at the end of the week.

MikeBishop:
Worth changing the doc name as well, even though it means another -00?

janaiyengar:
+1 -- QPACK SGTM

afrind:
Renamed to QPACK and included in the title

PR #1164, opened 2018-03-06
Rename QCRAM to QPACK

afrind:
* As discussed in #1133, QPACK is a more natural name given its similarity to HPACK.

I will note that based on the current set of open issues and PRs, QCRAM may diverge from HPACK more over time.  The core idea of a shared rotating dynamic table doesn't seem to be going anywhere, however.


MikeBishop:
This should include a search-and-replace in the HTTP draft as well.

Issue #1166, opened 2018-03-06
Moving version before connection ID in the long header


ianswett:
As I mentioned on PR #1151, if we change to asymmetric connection IDs with variable length, it would be ideal to have the version come before the connection IDs in the long header, because then it would be at a fixed offset and the destination connection ID would also still be at a fixed(but 4 byte later) offset.

Due to the importance of Connection ID, there is no intent to remove Connection ID from the invariants, though this change may enable some extra flexibility far down the road.

helpingothers4u:
I cannot remember my password and I am on my phone now that is my kindel
I'm trying get on in

Angela Risner

PR #1167, opened 2018-03-06
Move version

martinthomson:


ianswett:
Closes Issue #1166

mikkelfj:
If VNEG is an invariant and easily observable, why not fix the unused bits to 0 for this specifical instance where version is 0. This would make it trivial to test for VNEG. It's not like VNEG could later use the unused bits for anything useful later on.

MikeBishop:
See #963 for the previous round of discussion on that topic, but basically.  We don't want a packet type fixed across versions -- particularly since even the fact that the field *is* a packet type isn't actually an invariant.  The low bits of the first byte are simply version-dependent.

martinthomson:
Mikkel's point here, which I think is valid, is that for *version 0*, we could define the semantics of the type/unused bits.  Requiring the value to be 0 as part of that would be reasonable and might allow for other values to carry some other sort of signal (though what that might be is hard to know right now).

Issue #962, opened 2017-11-28
Handshake packets should be retransmitted based on the last sent handshake packet


ianswett:
Instead of using time_of_last_sent_packet, handshake packets should use time_of_last_sent_handshake_packet(to be defined), because if the handshake hasn't completed, protected packets will not be acknowledged, because they cannot be decrypted and processed.  The only other valid frames to send pre-handshake are acks and connection close.  Acks do not cause an acknowledgement and connection closes are terminal.

siyengar:
+1, this is a follow up to my comment in your PR?

PR #1171, opened 2018-03-06
Define time_of_last_sent_handshake_packet

ianswett:
Fixes Issue #962

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -537,6 +537,9 @@ largest_sent_before_rto:
 time_of_last_sent_packet:
 : The time the most recent packet was sent.
 
+time_of_last_sent_handshake_packet:
+: The time the most reccent handshake packet was sent.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -619,18 +623,24 @@ are as follows:
   ACK frame.  If true, it is still expected an ack will be received for
   this packet, but it is not congestion controlled.
 
+* is_handshake_packet: A boolean that indicates whether a packet is part of
```
marten-seemann: How does this apply to ack-only packets sent during the handshake?

It we're regarding them as handshake packets here, we probably need to change `OnPacketSend`, such that `time_of_last_sent_handshake_packet` is only updated for retransmittable handshake packets.

ianswett: Ah, I was thinking a handshake packet was a packet with handshake data in it, not based on packet type, but I'll restructure to make it clear.

PR #1180, opened 2018-03-12
allow sending of TLP and RTO probe packets when congestion limited

marten-seemann:
We already have text allowing to send TLP and RTO packets when congestion limited in the respective TLP and RTO sections.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -285,7 +285,7 @@ alarm for TCP as well, and this document incorporates this advancement.
 Timer-based loss detection implements the spirit of TCP's Tail Loss Probe
 and Retransmission Timeout mechanisms.
 
-### Tail Loss Probe
+### Tail Loss Probe {#tlp}
 
 The algorithm described in this section is an adaptation of the Tail Loss Probe
 algorithm proposed for TCP {{TLP}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -885,7 +885,9 @@ specified in bytes due to finer control and the ease of appropriate byte
 counting {{?RFC3465}}.
 
 QUIC hosts MUST NOT send packets if they would increase bytes_in_flight
-(defined in {{vars-of-interest}}) beyond the available congestion window.
+(defined in {{vars-of-interest}}) beyond the available congestion window, unless
+the packet is a probe packet sent after the TLP or RTO alarm fires, as described
+in {{tlp}} and {{rto}}.
```
martinthomson: Now it looks like this is a definitive list, but we have an exemption for packets that only carry ACK frames as well.

ianswett: ACK only packets don't increase bytes_in_flight, but that does remind me that handshake packets also may increase bytes_in_flight above the congestion window.

marten-seemann: Handshake packets are not congestion limited? Where does it say so?

ianswett: Sorry, I meant that similarly to TLP or RTO packets, handshake packets count towards bytes in flight, but can be sent even if bytes in flight exceeds congestion window.

marten-seemann: Can you point me to the section where this is specified? I can't find it.

ianswett: I had this in mind: "When the alarm fires, the sender MUST retransmit all unacknowledged handshake data. On each consecutive firing of the handshake alarm, the sender SHOULD double the handshake timeout and set an alarm for this period."

PR #1185, opened 2018-03-13
Move opportunistic ACK defense to security considerations

### LINKED ISSUE #305: Packet number skipping and entropy
> skipping packet numbers between 1 and 255 effectively provides up to 8 bits of efficient entropy on  demand, which should be adequate protection against most opportunistic acknowledgement attacks.

Aside from the odd phrasing, I think that this is incorrect.  Aside from the obvious point that 1-255 is only  7.994 bits of entropy, you also have to take into account the probability that a sender chooses to skip (a probability of 2^-8 would give the answer here).  Add to that the fact that any entropy accumulates with every packet sent.  Choosing to send or withhold each packet number could add up to a bit of entropy over time, so the difficulty in guessing could be higher over time.

No matter what strategy it's always safe for an attacker to guess high, so the amount of effective entropy against an attacker is probably less than claimed.  Of course, guessing high has an effect on congestion control, because that is marked by the sender as loss.  That could be as effective in mitigating the optimistic acknowledgment attack as the gap itself.

Then there is real packet loss.  If you randomly decide to skip 255 numbers, then real packet loss on the packet after the gap makes it difficult to acknowledge (you have to send two ACK frames if there is a gap >255).  Burst loss makes that even more likely.

In any case, I'd recommend removing any claim about entropy.  I also recommend removing any mention of specific numbers.  Skipping more packets is fine, but it means that you run the packet number space down faster and you get less efficiency in ACK packets.

martinthomson:
Closes #305, #1030.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2486,12 +2486,6 @@ cannot be acknowledged because they do not contain a packet number.  Rather than
 relying on ACK frames, these packets are implicitly acknowledged by the next
 Initial packet sent by the client.
 
-A sender MAY intentionally skip packet numbers to introduce entropy into the
-connection, to avoid opportunistic acknowledgement attacks.  The sender SHOULD
-close the connection if an unsent packet number is acknowledged.  The format of
-the ACK frame is efficient at expressing even long blocks of missing packets,
-allowing for large, unpredictable gaps.
-
 An ACK frame is shown below.
 
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3896,6 +3890,15 @@ secure key, the attacker will not be able to generate forward-secure protected
 packets with ACK frames.
 
 
+## Opportunistic ACK Attack
+
+An endpoint that acknowledges packets it has not received might cause a
+congestion controller to permit sending at rates beyond what the network
+supports.  An endpoint MAY skip packet numbers when sending packets to detect
+this.  An endpoint can then drop connection state and stop sending packets
+toward the misbehaving peer.
```
marten-seemann: That just means closing the connection, doesn't it? Why don't we say it explicitly?

MikeBishop: Because "closing" generally means generating an error code, etc. to inform a peer that it made a mistake.  This is malicious behavior.  Also, by not making it normative, we leave room for any implementation-specific responses.

draft-ietf-quic-transport.md
```diff
@@ -3896,6 +3890,15 @@ secure key, the attacker will not be able to generate forward-secure protected
 packets with ACK frames.
 
 
+## Opportunistic ACK Attack
+
+An endpoint that acknowledges packets it has not received might cause a
+congestion controller to permit sending at rates beyond what the network
+supports.  An endpoint MAY skip packet numbers when sending packets to detect
+this.  An endpoint can then drop connection state and stop sending packets
```
janaiyengar: this -> this behavior

PR #1192, opened 2018-03-14
Clients ignore packets of unknown versions

### LINKED ISSUE #827: client handling of unknown versions
Sec. 7.1 details the server handling for packets with unknown versions but has no mention on client behavior when it receives packets with unknown versions.

The section does detail MUST and MUST NOT behavior for handling version negotiation packets.

An additional requirement should be added that clients must ignore packets with unknown versions.

martinthomson:
Closes #827.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -888,6 +888,9 @@ that are encrypted with a key it has not yet computed. Clients MAY drop these
 packets, or MAY buffer them in anticipation of later packets that allow it to
 compute the key.
 
+If a client receives a packet that has an unsupported version, it MUST discard
+that packet.
+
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -888,6 +888,9 @@ that are encrypted with a key it has not yet computed. Clients MAY drop these
 packets, or MAY buffer them in anticipation of later packets that allow it to
 compute the key.
 
+If a client receives a packet that has an unsupported version, it MUST discard
+that packet.
+
```
mikkelfj: Isn't that an invariant? Otherwise the next version could require the client to do a best effort on unknown versions, and a client would not know what guides the world.

PR #1193, opened 2018-03-14
Remove implicit stream opening

### LINKED ISSUE #896: Implicitly opened streams add complexity
PR#894 reminded me that implicitly opened streams add complexity to the stream text and I'm not aware of any real value.

Efficient implementations would have to store a small amount of data to indicate a stream ID less than the largest had not been used, but I think that's true either way.

I realize there are cases when one COULD use this as a feature(ie: in protocols where both sides send info right after the TCP handshake), but you could also just wait for the stream frame to arrive.

I realize this has been brought up before, but now that we've resolved the uni/bidi direction, I think it makes sense to reconsider whether this is a feature or a bug.

martinthomson:
Closes #896.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3030,11 +3030,11 @@ Unidirectional streams use the applicable state machine directly.  Bidirectional
 streams use both state machines.  For the most part, the use of these state
 machines is the same whether the stream is unidirectional or bidirectional.  The
 conditions for opening a stream are slightly more complex for a bidirectional
-stream because the opening of either send or receive streams causes the stream
+stream because the opening of either send or receive sides causes the stream
 to open in both directions.
 
-An endpoint can open sending streams up to its maximum stream limit in any
-order, however endpoints SHOULD open streams for each type in order.
+An endpoint can open streams up to its maximum stream limit in any order,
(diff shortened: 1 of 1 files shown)


marten-seemann:
What is the reason for this change? Rereading #896, I don't think there was a consensus for making this change, and good arguments for keeping implicit stream opening.

ianswett:
The editors discussed this, and implementations may implicitly open streams, but nothing in the QUIC transport spec or HTTP mappings relies on implicitly opened streams.  We decided it was better to simplify the text and not require streams to be implicitly opened, and leave it up to implementations.

We're not aware of any problems this causes, but please tell us if there are any.

ianswett:
It was noted that for applications where either side my speak first, such as the HTTP mapping's control stream, two unidirectional streams works well.  Though in that case, implicit opened didn't fully solve the need anyway.

marten-seemann:
@ianswett: I don't think this is up to implementations now. If you implicitly open streams (and return them to the application via an `AcceptStream()` API, which returns streams opened by the peer in ascending order), the application could decide to trigger sending of a STOP_SENDING frame, which according to 8d450dbc1a10f8df095492940b7fc88592cc4c29 is now forbidden.

PR #1196, opened 2018-03-14
removing some dupe references

seanturner:
I-D nits complained about dupe references to RFC 1191 and RFC 8201.  I also just went with convention to use RFC#### as opposed to PLPMTUD.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2873,8 +2873,8 @@ header, protected payload, and any authentication fields.
 
 All QUIC packets SHOULD be sized to fit within the estimated PMTU to avoid IP
 fragmentation or packet drops. To optimize bandwidth efficiency, endpoints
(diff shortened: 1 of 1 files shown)


martinthomson:
Can we use the nice labels everywhere?  Is this a result of not using those labels in other places?

seanturner:
We can use the nice labels, but we can't mix.  And, I went all matchy-matchy because we don't have cute names for all the others.  Note that the I-D nit looked like this:

~~~~
-- Duplicate reference: RFC1191, mentioned in 'RFC1191', was also mentioned in 'PMTUDv4'.

-- Duplicate reference: RFC4821, mentioned in 'RFC4821', was also mentioned in 'PLPMTUD'.
~~~~

martinthomson:
Hey @seanturner, does this work OK for you?

martinthomson:
OBE, Mike fixed this already.

Issue #829, opened 2017-10-03
Can a server use a Handshake packet for ACK of the last client Handshake packet


huitema:
Interesting question. Picoquic and Quickly have an interop issue because Quickly acks the client finished message (client clear text) in a server clear text packet carrying an ACK. Picoquic ignores the server clear text packet because it has already sent the client finished and has moved to encrypted operation. Who is right?

janaiyengar:
Why did the server (Quicly) not move to 1-RTT and send the server ack in a 1-RTT packet? That's what I would have expected.

ekr:
What's the text that leads you to believe that it should do that.

kazuho:
@janaiyengar 
> Why did the server (Quicly) not move to 1-RTT and send the server ack in a 1-RTT packet? That's what I would have expected.

ACK for the cleartext packet that carried ClientFinished was sent in a cleartext packet that carried ServerFinished.

ekr:
FWIW, this is what I do as well.

huitema:
I am concerned that this creates a big special case. If I have to accept a clear text ACK for the Finished message, I need to keep track of the packet numbers that can be ACKed with that message -- ACKing 1RTT data would be wrong. That's a very big wart in the code.

kazuho:
@huitema 
> I am concerned that this creates a big special case. If I have to accept a clear text ACK for the Finished message, I need to keep track of the packet numbers that can be ACKed with that message -- ACKing 1RTT data would be wrong. That's a very big wart in the code.

We cannot use the timing at which we finish TLS handshake as a moment to switch from ACK over cleartext to ACK over protected packets, considering the fact that 0-RTT packets need to be ACKed using protected packets, and that 0-RTT packets and cleartext packets can be sent in mixed order.

huitema:
@kazuho You wrote:

> ACK for the cleartext packet that carried ClientFinished was sent in a cleartext packet that carried 
> ServerFinished.

That's not what I see in the logs. I see:

  -> Client initial (carries client hello)
  <- Server Clear Text (several message carry server flight, including server finished. Message embed 
       ACK of Client Initial
  -> Client clear text: carries client finished, and ACK of server clear text messages
  <- Server Clear Text: only carries an ACK Frame for the client clear text.

[...]

PR #1201, opened 2018-03-14
Addressing issue#1147

### LINKED ISSUE #1147: Sentence fragment in Key Update section
Line 1041:

> As shown in {{quic-tls-handshake}} and {{ex-key-update}}, there is never a
> situation where there are more than two different sets of keying material that
> might be received by a peer.  Once both sending and receiving keys have been
> updated,

The paragraph ends mid-sentence.

seanturner:
Closes #1147, but this might be wrapped up in #829.


seanturner:
Turns out it's not :)

PR #1202, opened 2018-03-14
Specify pacing in a bit more detail.

### LINKED ISSUE #1056: "CanSendPacket" function or equivalent function
The congestion control section doesn't provide a function that describes how a sender might decide whether to send a packet (or, similarly, but more complicated, a function that describes what can be sent in a packet at the current moment).

Ideally this would include pacing and congestion limits.  And it would include the exclusion for ACK-only packets (#1055).  

Anyone addressing this might consider #837 at the same time.

janaiyengar:
Closes #1056.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -951,11 +951,20 @@ slow start is re-entered.
 
 ## Pacing
 
-It is RECOMMENDED that a sender pace sending of all data, distributing the
-congestion window over the SRTT.  This document does not specify a pacer.  As an
-example pacer, implementers are referred to the Fair Queue packet scheduler (fq
-qdisc) in Linux (3.11 onwards) as a well-known and publicly available
-implementation of a flow pacer.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -951,11 +951,20 @@ slow start is re-entered.
...
 
-It is RECOMMENDED that a sender pace sending of all data, distributing the
-congestion window over the SRTT.  This document does not specify a pacer.  As an
-example pacer, implementers are referred to the Fair Queue packet scheduler (fq
-qdisc) in Linux (3.11 onwards) as a well-known and publicly available
-implementation of a flow pacer.
+This document does not specify a pacer, but it is RECOMMENDED that a sender pace
+sending of all data based on input from the congestion controller.  For example,
+a pacer might distribute the congestion window over the SRTT when used with a
+window-based controller, and a pacer might simply use the rate estimate of a
```
ianswett: or a pacer might use the rate estimate...

PR #1210, opened 2018-03-15
Improve Push text

### LINKED ISSUE #1175: HTTP/QUIC supports server push as described in [RFC7540], except where it doesn't
In the process of discussing some design for HTTP/QUIC PUSH_PROMISE, I was reminded that promised requests must fit within certain criteria. I thought this was defined in the HTTP mapping but it is actually referenced back to RFC7540 section 8.2. There is nothing inherently wrong with that. (for the curious, the text I wanted says "Promised requests MUST be cacheable (see [RFC7231], Section 4.2.3),  MUST be safe (see [RFC7231], Section 4.2.1), and MUST NOT include a request body.")

However, having read HTTP/QUIC section 3.4 again it comes across as a bit awkward (hence the annoying issue title). 

I think this is just an editorial issue to reword and/or restructure the section a little.

MikeBishop:
Attempts to resolve #1175.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -439,10 +439,10 @@ push by sending a MAX_PUSH_ID frame (see {{frame-max-push-id}}). A server cannot
 use server push until it receives a MAX_PUSH_ID frame.
 
 As with server push for HTTP/2, the server initiates a server push by sending a
-PUSH_PROMISE frame that includes request header fields attributed to the
-request.  Promised requests MUST be cacheable (see {{!RFC7231}}, Section 4.2.3),
-MUST be safe (see {{RFC7231}}, Section 4.2.1), and MUST NOT include a request
-body.
+PUSH_PROMISE frame (see {{frame-push-promise}}) that includes request headers
+for the promised request.  Promised requests MUST be cacheable (see
+{{!RFC7231}}, Section 4.2.3), MUST be safe (see {{RFC7231}}, Section 4.2.1), and
+MUST NOT include a request body.
 
 The PUSH_PROMISE frame is sent on the client-initiated, bidirectional stream
 that carried the request that generated the push.  This allows the server push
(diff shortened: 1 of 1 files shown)


MikeBishop:
On (4), it's because Martin reused the RST_STREAM codepoint for CANCEL_PUSH, and the frames are presented in codepoint order.  Whether that's a "good" reason is debatable, but that's the reason anyway.  I'll take a look at the others.

draft-ietf-quic-http.md
```diff
@@ -433,28 +433,43 @@ frame.
...
 ## Server Push
 
-HTTP/QUIC supports server push as described in {{!RFC7540}}. During connection
-establishment, the client enables server push by sending a MAX_PUSH_ID frame
-(see {{frame-max-push-id}}).  A server cannot use server push until it receives
-a MAX_PUSH_ID frame.
+HTTP/QUIC supports server push in a similar manner to {{!RFC7540}}, but uses
+different mechanisms. During connection establishment, the client enables server
+push by sending a MAX_PUSH_ID frame (see {{frame-max-push-id}}). A server cannot
+use server push until it receives a MAX_PUSH_ID frame.
```
martinthomson: Which makes me wonder (probably for the second time) whether adding a setting is valuable or not.  I guess not because MAX_PUSH_ID can be send immediately after SETTINGS, and having two ways to do the same thing isn't great.

draft-ietf-quic-http.md
```diff
@@ -433,28 +433,43 @@ frame.
...
 
 As with server push for HTTP/2, the server initiates a server push by sending a
-PUSH_PROMISE frame that includes request header fields attributed to the
-request. The PUSH_PROMISE frame is sent on the client-initiated, bidirectional
-stream that carried the request that generated the push.  This allows the server
-push to be associated with a request.  Ordering of a PUSH_PROMISE in relation to
+PUSH_PROMISE frame (see {{frame-push-promise}}) that includes request headers
+for the promised request.  Promised requests MUST be cacheable (see
+{{!RFC7231}}, Section 4.2.3), MUST be safe (see {{RFC7231}}, Section 4.2.1), and
+MUST NOT include a request body.
```
martinthomson: I think that we want to reference the requirements in h2, not repeat them.

LPardue: I'd be happy with that

MikeBishop: Ah, I though part of the point of the original issue was that these requirements weren't in the spec, but just pointed off to RFC7540.  I'll replace this with a pointer to the specific section.

draft-ietf-quic-http.md
MikeBishop: By that logic, we shouldn't have transport parameters for max stream ID either -- just send MAX_STREAM_ID frames in your first packet.  But that would prevent servers from using 0.5-RTT data, since they haven't received authenticated packets from the client yet.

HTTP has no such restriction, since all data is inside the crypto.  We can just use the frames.

Issue #591, opened 2017-06-06
is_retransmittable semantics


gloinul:
In draft-ietf-quic-recovery-03 there appear to me to be an overloading of the use of the "is_retransmittable" flag. It appears to be used both for the purpose to indicate if there is a frame that can and need be retransmitted and if that frame should be counted as lost and impact congestion control. I think these two purposes should be separated and clarified as being two different ones. The reason is that if we create extensions that will not require retransmission of frames but needs to be included in the congestion control response a significant rewrite would be needed. I also think it would clarify things for some of the control frames also if they should be included or not, even if not regularly retransmitted.

ianswett:
Agreed that these concepts need to be separated.

I replaced is_retransmittable with is_ack_only in PR #780, because padding is not retransmittable, but does count towards bytes in flight.  My intent is to make everything that is not an ack only packet count towards bytes_in_flight.

But I think there's still some more editorial work to clarify the separation you describe.

gloinul:
This is still not clear in -10, will keep it open.

ianswett:
The plan is to rename this to needs_acknowledgement as well, because QUIC rarely retransmits frames and never retransmits packets.

PR #1211, opened 2018-03-15
No more Ackable

ianswett:
Fixes Issue #591

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -232,7 +232,7 @@ congestion-controlled packet triggers the Early Retransmit process, as follows.
 If there are unacknowledged congestion-controlled packets still pending, they should
 be marked as lost. To compensate for the reduced reordering resilience, the sender
 SHOULD set an alarm for a small period of time. If the unacknowledged
-congstion-controlled packets are not acknowledged during this time, then these
+congestion-controlled packets are not acknowledged during this time, then these
 packets MUST be marked as lost.
 
 An endpoint SHOULD set the alarm such that a packet is marked as lost no earlier
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -100,9 +100,12 @@ interpretation of TCP loss detection mechanisms.
 Every packet may contain several frames.  We outline the frames that are
 important to the loss detection and congestion control machinery below.
 
-* Congestion controlled frames are frames that count towards bytes in
+* Congestion-controlled frames are those that count towards bytes in
   flight and need acknowledgement.  The most common are STREAM frames,
   which typically contain application data.
+
+* Congestion-controlled packets are those that contain at least one
+  congestion-controlled frame.
 
 * Crypto handshake data is sent on stream 0, and uses the reliability
   machinery of QUIC underneath.
@@ -540,8 +543,9 @@ max_ack_delay:
 
 reordering_threshold:
 : The largest packet number gap between the largest acked
-  congestion controlled packet and smaller unacknowledged congestion
-  controlled packet before it is declared lost.
+  congestion-controlled packet and an unacknowledged
+  congestion-controlled packet with a lower packet number
+  before it is declared lost.
 
 time_reordering_fraction:
 : The reordering window as a fraction of max(smoothed_rtt, latest_rtt).
@@ -909,7 +913,7 @@ the reduction to once per round trip.
 ## Tail Loss Probe
 
 A TLP packet MUST NOT be blocked by the sender's congestion controller. The
-sender MUST however count these bytes as additional bytes in flight, since a TLP
+sender MUST however count these bytes as additional bytes-in-flight, since a TLP
 adds network load without establishing packet loss.
 
 Acknowledgement or loss of tail loss probes are treated like any other packet.
@@ -971,13 +975,13 @@ are described in this section.
 
 bytes_in_flight:
 : The sum of the size in bytes of all sent packets that contain at least
-  one congestion controlled frame, and have not been acked or declared
+  one congestion-controlled frame, and have not been acked or declared
   lost. The size does not include IP or UDP overhead.
   Packets only containing ACK frames do not count towards bytes_in_flight
   to ensure congestion control does not impede congestion feedback.
 
 congestion_window:
-: Maximum number of bytes in flight that may be sent.
+: Maximum number of bytes-in-flight that may be sent.
 
 end_of_recovery:
 : The largest packet number sent when QUIC detects a loss.  When a larger
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -223,8 +223,8 @@ and implementers are encouraged to explore this space.
 
 ### Early Retransmit
 
-Unacknowledged packets close to the tail may have fewer than
-kReorderingThreshold number of congestion-controlled packets sent after them.
+Unacknowledged packets close to the tail may have fewer than kReorderingThreshold
+congestion-controlled packets sent after them.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -100,14 +100,15 @@ interpretation of TCP loss detection mechanisms.
...
+* Congestion controlled frames are frames that count towards bytes in
+  flight and need acknowledgement.  The most common are STREAM frames,
+  which typically contain application data.
 
 * Crypto handshake data is sent on stream 0, and uses the reliability
   machinery of QUIC underneath.
 
 * ACK frames contain acknowledgment information.  ACK frames contain one or more
-  ranges of acknowledged packets.
+  ranges of acknowledged packets and are not congestion controlled.
```
marten-seemann: This now sounds like ACKs never count towards the bytes_in_flight. As far as I understand the rest of this PR, this only applies for ACK-only packets though, right?

ianswett:
Thanks Marten, reverted to what it was before.

draft-ietf-quic-recovery.md
```diff
@@ -538,9 +539,9 @@ max_ack_delay:
   RTT sample less than min_rtt.
 
 reordering_threshold:
-: The largest delta between the largest acked
-  retransmittable packet and a packet containing retransmittable frames before
-  it's declared lost.
+: The largest packet number gap between the largest acked
+  congestion controlled packet and smaller unacknowledged congestion
```
nibanks: This is the first reference you have to `congestion controlled packet`. You haven't explicitly defined what this means. I assume it is just a packet that contains at least 1 congestion controlled frame, but I think it would be worth explicitly stating that above.

draft-ietf-quic-recovery.md
```diff
@@ -538,9 +539,9 @@ max_ack_delay:
...
+  controlled packet before it is declared lost.
```
nibanks: Why does `reordering_threshold` care about congestion-controlled or not? If you have the following outstanding packets:

1. ACK only
2. ACK only
3. ACK only
4. MAX_DATA only

And then you get an ACK for packet 4, should you not consider any of the previous packets as lost? I understand that marking a non-congestion controlled packet as lost might have no effect, but it is lost nonetheless, right?

draft-ietf-quic-recovery.md
```diff
@@ -100,8 +100,9 @@ interpretation of TCP loss detection mechanisms.
 Every packet may contain several frames.  We outline the frames that are
 important to the loss detection and congestion control machinery below.
 
-* Retransmittable frames are frames requiring reliable delivery.  The most
-  common are STREAM frames, which typically contain application data.
+* Congestion controlled frames are frames that count towards bytes in
+  flight and need acknowledgement.  The most common are STREAM frames,
```
janaiyengar: hyphenate bytes-in-flight

draft-ietf-quic-recovery.md
```diff
@@ -100,8 +100,9 @@ interpretation of TCP loss detection mechanisms.
 Every packet may contain several frames.  We outline the frames that are
 important to the loss detection and congestion control machinery below.
 
-* Retransmittable frames are frames requiring reliable delivery.  The most
-  common are STREAM frames, which typically contain application data.
+* Congestion controlled frames are frames that count towards bytes in
```
janaiyengar: "frames are those"

draft-ietf-quic-recovery.md
marten-seemann: @nibanks You don't even save packets 1 to 3 (in the sent_packets), so when receiving the ACK for 4, you wouldn't even remember that you ever sent them.

draft-ietf-quic-recovery.md
janaiyengar: "congestion-controlled"

draft-ietf-quic-recovery.md
janaiyengar: "unacknowledged congestion-controlled packet with a lower packet number"

draft-ietf-quic-recovery.md
nibanks: Yes, I agree you don't *need* to save them, but if you want to, for purely a statistical measurement of loss, couldn't you remember them?

marten-seemann: You *could* do that, but you'll probably have to be very careful about loss at the tails. I'm not sure if this would be an improvement over a measurement that just uses retransmittable packets though.

mikkelfj: How would you deal with optimistic ACK attack if you forget? It is valid to ACK an unackable packet if it is together with other packets, right?

MikeBishop: All packets except Reset are ackable, and so they are included in the ACK when it's sent.  ACK-only packets don't provoke an ACK in response, however.

draft-ietf-quic-recovery.md
```diff
@@ -970,9 +971,9 @@ are described in this section.
 
 bytes_in_flight:
 : The sum of the size in bytes of all sent packets that contain at least
-  one retransmittable or PADDING frame, and have not been acked or
-  declared lost. The size does not include IP or UDP overhead.
-  Packets only containing ACK frames do not count towards byte_in_flight
+  one congestion controlled frame, and have not been acked or declared
```
MikeBishop: hyphenate

draft-ietf-quic-recovery.md
```diff
@@ -733,8 +734,8 @@ Pseudocode for SetLossDetectionAlarm follows:
 ~~~
  SetLossDetectionAlarm():
     // Don't arm the alarm if there are no packets with
-    // retransmittable data in flight.
-    if (num_retransmittable_packets_outstanding == 0):
+    // congestion controlled data in flight.
```
MikeBishop: hyphenate

draft-ietf-quic-recovery.md
```diff
@@ -711,10 +712,10 @@ response to 0RTT packets.
 
 #### Tail Loss Probe and Retransmission Alarm
 
-Tail loss probes {{?TLP}} and
-retransmission timeouts {{?RFC6298}} are an alarm based mechanism to recover
-from cases when there are outstanding retransmittable packets, but an
-acknowledgement has not been received in a timely manner.
+Tail loss probes {{?TLP}} and retransmission timeouts {{?RFC6298}}
+are an alarm based mechanism to recover from cases when there are
+outstanding congestion controlled packets, but an acknowledgement has
```
MikeBishop: hyphenate

draft-ietf-quic-recovery.md
ianswett: Thanks, done above.

draft-ietf-quic-recovery.md
nibanks: @ianswett What are your thoughts here? Shouldn't all packets, congestion controlled or not, be capable of being lost and therefore be able to be detected as lost?

mikkelfj: This is one of the great unsolved mysteries of the universe - is information capable of getting lost.

ianswett:
I seem to be having problems getting any consensus on a replacement for retransmittable, so I'm going to keep it as retransmittable and we can continue to discuss a replacement name.
Leading suggestions:
1) Congestion-controlled (as in this PR now)
2) In-flight (because it's included in bytes_in_flight)

ianswett:
If people would like to continue discussing naming, please open a separate issue.  I think this PR clarifies the original Issue enough to consider it fixed.

martinthomson:
@ianswett, remember that you are the editor, and questions of choice of words do not generally require consensus (not that that helps much, of course).

PR #1213, opened 2018-03-15
Retransmit data, not frames

ianswett:


draft-ietf-quic-transport.md
```diff
@@ -3419,7 +3419,7 @@ cryptographic handshake.  This includes the retransmission of the second flight
 of client handshake messages, that is, the TLS Finished and any client
 authentication messages.
 
-STREAM frames that are determined to be lost SHOULD be retransmitted before
+STREAM data in frames determined to be lost SHOULD be retransmitted before
```
MikeBishop: Data in STREAM frames.  We wouldn't all-caps it if it's talking about "stream data".

mikkelfj: Could it be stated more explicitly that retransmitted data may have different framing and overlap, just to avoid the odd interop problem?

PR #1214, opened 2018-03-15
Change CID after inactivity

### LINKED ISSUE #1087: Recommend simulating migration after quiescence
A connection that is unused for a while might trigger NAT rebinding when the NAT times out.  We can recommend that a client use the next available connection ID at that point.  Nothing much is lost by doing so, and it increases the use of NEW_CONNECTION_ID.

MikeBishop:
Fixes #1087, builds on #1012 or successors.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1762,6 +1762,12 @@ recovers the packet number by adding the cumulative packet number gap to its
 expected packet number.  An endpoint MUST discard packets that contain a smaller
 gap than it advertised.
 
+Clients MAY change connection ID at any time based on implementation-specific
+concerns.  For example, after a period of network inactivity NAT rebinding might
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1876,6 +1876,12 @@ connection IDs, but it MUST ensure that it applies the associated packet number
 gaps for connection IDs that it skips in addition to the packet number gap
 associated with the connection ID that it does use.
 
+Clients MAY change connection ID at any time based on implementation-specific
+concerns.  For example, after a period of network inactivity NAT rebinding might
+occur when the client begins sending data again. A client might wish to break
+linkability by employing a new connection ID when resuming traffic after
```
martinthomson: when sending packets after a period of inactivity.

"resuming traffic" is less precise

PR #1215, opened 2018-03-15
Remove confusing text about ports

### LINKED ISSUE #929: Why MUST a server use the same port for HTTP/QUIC?
Section 3 on QUIC advertisement finishes with the following paragraph

> Servers MAY serve HTTP/QUIC on any UDP port. Servers MUST use the same port across all IP addresses that serve a single domain, and SHOULD NOT change this port.

I have a few questions:

1. Where does the requirement to use the same port come from?
  a) Does this lock server owners into a single port for eternity? (per-user port advertisement seems unlikely to me)
  b) What happens if a server does listen on different ports? Is it a connection error?
  c) I can see some hypothetical problems, e.g. where Alt-Svc is used for load shedding to a different origin that resolves to an anycast IP, leading to violation. Similar to what has been discussed on #560 and #253.
2. Should clients pro-actively ignore non-compliant servers?
3. Why are port changes recommended against?
4. What should clients do if they see an Alt-Svc within the a HTTP/QUIC connection that changes the port.?

I appreciate the connection of this paragraph to QUIC advertisement. However, it actually seems quite fundamental guidance for HTTP/QUIC server deployments on how they should approach domain, IP and port usage. This makes me wonder if it should be broken out into a separate (sub)section or that it would pay to pull this information out into the manageability document.

MikeBishop:
This is just one of the practicalities of an Alt-Svc deployment; it's not necessary to call it out here.  (Though perhaps it's worth a mention in the manageability document, since QUIC is the most likely candidate to hit it.)

Fixes #929.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -132,7 +132,7 @@ following header in any response:
 Alt-Svc: hq=":50781"
 ~~~
 
-On receipt of an Alt-Svc header indicating HTTP/QUIC support, a client MAY
+On receipt of an Alt-Svc record indicating HTTP/QUIC support, a client MAY
 attempt to establish a QUIC connection to the indicated host and port and, if
 successful, send HTTP requests using the mapping described in this document.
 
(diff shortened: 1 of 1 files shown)


martinthomson:
Interesting consequences for the work we did to regain control over UDP/443, but this is indeed correct.

PR #1219, opened 2018-03-16
loss recovery timers are based on time of last retransmittable packet

marten-seemann:
If I understand the loss recovery correctly, we should base the loss recovery timers (TLP and RTO) on the send time of the last retransmittable packet, not the last packet sent.

@janaiyengar, @ianswett: Is that correct?

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -506,8 +506,8 @@ largest_sent_before_rto:
 : The last packet number sent prior to the first retransmission
   timeout.
 
-time_of_last_sent_packet:
-: The time the most recent packet was sent.
+time_of_last_sent_retransmittable_packet:
+: The time the most recent retransmittable packet was sent.
 
 time_of_last_sent_handshake_packet:
 : The time the most recent packet containing handshake data was sent.
@@ -578,7 +578,7 @@ follows:
    min_rtt = infinite
    max_ack_delay = 0
    largest_sent_before_rto = 0
-   time_of_last_sent_packet = 0
+   time_of_last_sent_retransmittable_packet = 0
    time_of_last_sent_handshake_packet = 0
    largest_sent_packet = 0
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -506,8 +506,8 @@ largest_sent_before_rto:
 : The last packet number sent prior to the first retransmission
   timeout.
 
-time_of_last_sent_packet:
-: The time the most recent packet was sent.
+time_of_last_sent_retransmittable_packet:
```
ianswett: I'm trying to remove the use of retransmittable from this document in a different PR, so I'd prefer not to add it here.  Can you change the text below to be "The time the most recent congestion-controlled packet was sent."?  For the variable rename, is there something shorter we could come up with that doesn't use the word retransmittable?

marten-seemann: I used retransmitted here since your PR isn't merged yet.

Maybe we can call these packets ackable instead of congestion controlled? I'm not sure if I really like this term, since ACK-only packets can also be acked (later, as soon as a retransmittable packet is received), but I can't think of anything better right now.

mikkelfj: time_of_last_ack_eliciting_packet?

mcmanus:
"reliable information" ?

marten-seemann:
I think the discussion about the right term belongs in #1211. Ian chose "congestion controlled" there, and I'm fine with this term, although it makes variable names pretty long.

If we merge this PR here first, #1211 should be rebased and change the variable name. If we merge #1211 first, I'll rebase this PR to reflect whatever term we came up with.

mikkelfj:
We could define two terms more explicitly elsewhere - for example Signal packet and Reliable packet where signal packets must elicit an ACK and Carrier packets SHOULD elicit an ack.
It still sort of suggests that the packet is retransmitted, but then, the logic as at packet level.

The name would then be, as @mcmanus suggests:

time_of_last_sent_reliable_packet

ianswett:
I'd rather not introduce the word reliable here, since this term is really
about whether it counts towards bytes_in_flight and therefore whether it
invokes an ack.

Even in a use case where nothing is ever retransmitted, congestion control
and acknowledgement behavior are still relevant.

siyengar:
looks good to me. I'm also fine with the use of the word ``retransmittable``

PR #1222, opened 2018-03-17
Only pace retransmittable packets

ianswett:


mikkelfj:
You might occasionally if you think things could  be clogging up and that small ack packets has a greater chance of survival.

ianswett:
ACK-only packets shouldn't be limited by the pacer, otherwise the max ack delay may be unintentionally exceeded without measuring it.

draft-ietf-quic-recovery.md
```diff
@@ -927,15 +927,18 @@ slow start is re-entered.
...
+rate estimate of a rate-based controller.
 
 An implementation should take care to architect its congestion controller to
 work well with a pacer.  For instance, a pacer might wrap the congestion
 controller and control the availability of the congestion window, or a pacer
-might pace out packets handed to it by the congestion controller.
+might pace out packets handed to it by the congestion controller.  In either
+arrangement, packets only containing ACK frames should not be paced, to prevent
+spurious timeout based retransmissions and to ensure the intended max ack delay
+is honored.
```
janaiyengar: This sentence is a bit unwieldy. How about:
"Timely delivery of ACK frames is important for efficient loss recovery.  Packets containing only ACK frames should therefore not be paced, to avoid delaying their delivery to the peer."

MikeBishop:
Didn't we talk yesterday about pacing ACK-only packets as well, to ensure that the recipient isn't flooding the network with an "every other incoming packet" ACK strategy?

ianswett:
Mike, I think we should add some text about rate limiting based on Monday's discussion, but actually pacing acks risks exceeding the intended delayed ack timer, which could have unintended consequences I'd rather avoid.

PR #1229, opened 2018-03-18
Rename RetransmitAllHandshakePackets

### LINKED ISSUE #625: RetransmitAllHandshakePackets() and SetLossDetectionAlarm()
`RetransmitAllHandshakePackets()` is not further defined in the loss-recovery ID. But it is implied that when executing this, one would also execute `OnPacketSent()`, which produces the effect hat `SetLossDetectionAlarm()` is called twice for the same transmission event. That is clearly not intended.

ianswett:
Fixes #625


ianswett:
bytes_in_flight increases, but I think you're getting at a point, which is that at some point, all unencrypted packets have to be removed from bytes_in_flight without losing them.

PR #1231, opened 2018-03-18
Restrict ACK-clocking to ACK-only frames.

ekr:
As written, this would forbid the case where A sends a single stream
frame and then B responds with three stream frames and piggybacks
ACKs along with each. That seems silly and a lot of work. Rather,
this should just be about ACK-only frames.


draft-ietf-quic-transport.md
```diff
@@ -2538,10 +2538,10 @@ ACK Block (repeated):
 Implementations MUST NOT generate packets that only contain ACK frames in
 response to packets which only contain ACK frames. However, they MUST
 acknowledge packets containing only ACK frames when sending ACK frames in
-response to other packets.  Implementations MUST NOT send more than one ACK
-frame per received packet that contains frames other than ACK frames.  Packets
-containing non-ACK frames MUST be acknowledged immediately or when a delayed
-ack timer expires.
+response to other packets.  Implementations MUST NOT send more than one packet
+containing only ACK frames per received packet that contains frames other than
```
ianswett: I have no objection, but is the motivation for this the coming CRYPTO_ACK/ACK split?

ekr: No. It was purely that the text didn't match my expectations

draft-ietf-quic-transport.md
```diff
@@ -2538,20 +2538,24 @@ ACK Block (repeated):
...
 an ACK frame has been acknowledged, the packets it acknowledges SHOULD NOT be
 acknowledged again.
 
-A receiver that is only sending ACK frames will not receive acknowledgments for
-its packets.  Sending an occasional MAX_DATA or MAX_STREAM_DATA frame as data is
-received will ensure that acknowledgements are generated by a peer.  Otherwise,
-an endpoint MAY send a PING frame once per RTT to solicit an acknowledgment.
+Because ACK-only packets are not acknowledged, a receiver that is only
+sending ACK frames will only receive acknowledgements for its packets
+if the sender piggybacks them on other packets. In order to enable
```
marten-seemann: Why does it play a role how the sender bundles frames into packets?

draft-ietf-quic-transport.md
```diff
@@ -2538,20 +2538,24 @@ ACK Block (repeated):
...
+this, senders SHOULD prioritize sending ACKs along with their data.
```
marten-seemann: I don't think you should prioritize ACKs, if the ACKs don't ack any new retransmittable packets.

draft-ietf-quic-transport.md
```diff
@@ -2538,20 +2538,24 @@ ACK Block (repeated):
...
+In addition, if a receiver regularly issues additional flow control
+credit rather than just issuing infinite flow control credit at the
```
marten-seemann: Issuing infinite flow control credit is a really bad idea. I don't think we should mention this as an option here.

draft-ietf-quic-transport.md
ianswett: I think prioritize is a loaded word, maybe "senders SHOULD bundle ACKs with non-ACK frames when possible."

draft-ietf-quic-transport.md
```diff
@@ -2538,20 +2538,24 @@ ACK Block (repeated):
...
 To limit ACK blocks to those that have not yet been received by the sender, the
 receiver SHOULD track which ACK frames have been acknowledged by its peer.  Once
 an ACK frame has been acknowledged, the packets it acknowledges SHOULD NOT be
 acknowledged again.
 
-A receiver that is only sending ACK frames will not receive acknowledgments for
-its packets.  Sending an occasional MAX_DATA or MAX_STREAM_DATA frame as data is
-received will ensure that acknowledgements are generated by a peer.  Otherwise,
-an endpoint MAY send a PING frame once per RTT to solicit an acknowledgment.
+Because ACK-only packets are not acknowledged, a receiver that is only
```
ianswett: ACK-only packets are acknowledged, but they do not cause an ACK frame to be sent.

Maybe: ""Because ACK frames are not sent in response to ACK-only packets, ..."

draft-ietf-quic-transport.md
ianswett: I would remove this sentence.

draft-ietf-quic-transport.md
ekr: Consider the case where the receiver issued 1MB of flow credit and the sender sends 950K of data. In this case the receiver might just send ACK-only packets which the sender is forbidden to send its own ACK-only packets. Thus, if the sender never bundles ACKs, the receiver will never get any.

marten-seemann: Where is the problem? You don't expect to get any ACK for ACK-only packets, they don't count towards bytes-in-flight, they are never retransmitted.

ekr: Well, the text I am updating assumes it is a problem, I'm just fixing it up.

With that said, while the packets don't count towards bytes-in-flight, you generally end up storing the data about the to-be-ACKed packets (as implied by the following text):

```To limit ACK blocks to those that have not yet been received by the sender, the
receiver SHOULD track which ACK frames have been acknowledged by its peer.  Once
an ACK frame has been acknowledged, the packets it acknowledges SHOULD NOT be
acknowledged again.```

Thus, you can't garbage collect this state until those ACKs are ACKed.

mikkelfj: I think that is an unfortunate requirement (SHOULD NOT be ack'ed again):
You want to ack packets again because you collapse gaps into larger blocks. Only the oldest ACK'ed block should not be acked again.

ekr: @mikkelfj: I suggest you file a separate issue.

mikkelfj: https://github.com/quicwg/base-drafts/issues/1234

marten-seemann: The text as it stands makes sense: **If** a peer wants an ACK for an ACK (e.g. to prune the lowest acked packet), it can include a retransmittable frame. This depends on which ack-pruning algorithm it uses, and it has nothing to do with how frames are bundled.

ekr: That part of the text is correct. What's not correct is the claim that you will not receive ACKs if you do not do so.

PR #1235, opened 2018-03-19
Change Index-Duplicate instruction code point

### LINKED ISSUE #1142: Header table size change
QCRAM doesn't afford the encoder the ability to specify the table size it is using, which is a takeback from HPACK.  In QCRAM, the encoder is using the maximum table size permitted by the decoder regardless.  Are we okay with that?

HPACK broke the "maximum you can use" and "size I am using" pieces apart because, although the encoder can choose to mostly-forget entries that are above the size it uses, the decoder is left tracking useless items.  (Also, it allows the encoder to play tricks like setting the table size to zero and then returning it to the previous size to flush the table on an idle connection.)

afrind:
Changed the instruction for duplicate from 001/5+ to 1/7+.  This is because Table Size Update is otherwise a valid QPACK instruction

Fixes #1142

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -253,17 +253,17 @@ Indexed-Duplicate representation instead (see {{indexed-duplicate}}).
 ~~~~~~~~~~  drawing
     0 1 2 3 4 5 6 7
    +-+-+-+-+-+-+-+-+
-   |1|Index(7+)    |
+   |1|  Index (7+) |
    +-+-+-+---------+
 ~~~~~~~~~~
 {:#fig-index-with-duplication title="Indexed Header Field with Duplication"}
 
-*Indexed-Duplicates* insert a new entry into the dynamic table which duplicates
+*Indexed-Duplicate* inserts a new entry into the dynamic table which duplicates
 an existing entry. {{RFC7541}} allows duplicate HPACK table entries, that is
 entries that have the same name and value.
 
 This repurposes the HPACK instruction for Indexed Header Field (see Section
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -253,7 +253,7 @@ Indexed-Duplicate representation instead (see {{indexed-duplicate}}).
 ~~~~~~~~~~  drawing
     0 1 2 3 4 5 6 7
    +-+-+-+-+-+-+-+-+
-   |0|0|1|Index(5+)|
+   |1|Index(7+)    |
```
MikeBishop: Center, and probably a space before the parenthesis since there's room.

draft-ietf-quic-qpack.md
```diff
@@ -262,8 +262,8 @@ Indexed-Duplicate representation instead (see {{indexed-duplicate}}).
 an existing entry. {{RFC7541}} allows duplicate HPACK table entries, that is
 entries that have the same name and value.
 
-This replaces the HPACK instruction for Dynamic Table Size Update (see Section
-6.3 of {{RFC7541}}, which is not supported by HTTP over QUIC.
+This repurposes the HPACK instruction for Indexed Header Field (see Section
+6.1 of {{RFC7541}}.
```
MikeBishop: This wording might be confusing to some readers.  It might be more accurate to say that you're defining the behavior of this instruction on the control stream, while it retains its usual meaning on the request streams.

MikeBishop:
I like that this makes the instructions sets have fewer invalid values.  I dislike that it gives a short prefix code to one of the least-common instructions.

PR #1237, opened 2018-03-19
Clarify QPACK eviction races

### LINKED ISSUE #1129: Referencing strategy
(This is a little more nebulous than my other issues, but I think that there are concrete things that can be written down that will improve things.)

Referencing entries near the end of the table in a new header block is hazardous. If you just compress as you do in HPACK, you can end up pinning entries in the header table, making it impossible to add new entries. Thus, you have to stop referencing entries as they get nearer to the table capacity limit.

The draft isn't particularly good about explaining this exposure or outlining strategies that might be used to manage this (or the impact it might have on the selection of header table size, for example).

One framework for thinking about this is probably:

An encoder attempts new header fields to the header table at a certain rate.  Under ideal conditions, the insertion of header fields is never blocked by outstanding references to entries that might be evicted as a result of the insertion.  Under these ideal conditions, header fields enter the table at some rate, H.  Note that a new connection might see a higher rate of insertions, but the expectation is that the empty table will be able to absorb those insertions.  That is, H averages out in the long term to |H|.

New removals from the table are possible after one round trip if the header field is not referenced in that round trip.  Thus, an encoder preserves its ability to insert at a rate of |H| if it refrains from using the last |H|\*RTT of the table.  To allow for both delays in delivering acknowledgments and variance in RTT and H, an encoder might choose to use a higher value that takes variance into consideration.  In extremis: (Mh\*|H| + Nh\*Hvar)\*(Mr\*RTT + Nr\*RTTvar) where Mx > 1 and Nx > 0.  An encoder could measure these values and adapt the amount of buffer it reserves, or choose typical values.

A decoder should allow for this safety margin and - assuming no other constraints on space - set a table size that is this amount larger than is ideal for compression.  A decoder has to choose values based on typical (or some percentile, e.g., 90th) values, because it can't resize the table for each connection.

afrind:
Fixes #1129

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -905,27 +905,36 @@ the decoder.
 
 ### Blocked Eviction
 
-The encoder MUST NOT permit an entry to be evicted while a reference to that
-entry remains unacknowledged.  If a new header to be inserted into the dynamic
-table would cause the eviction of such an entry, the encoder MUST NOT emit the
-insert instruction until the reference has been processed by the decoder and
-acknowledged. This is called a blocked eviction.
-
-When the encoder encounters a blocked eviction, the encoder SHOULD emit a
-literal representation for the new header in order to avoid encoding delays, and
-MAY insert the header into the table later if desired.
+The encoder MUST NOT emit an instruction that evicts an entry while a reference
+to that entry remains unacknowledged. When the encoder encounters a blocked
+eviction in the course of encoding a header block, it SHOULD emit a literal
+representation for the new header, and MAY insert the header into the table
+later if desired.
 
 To ensure that blocked evictions are rare, the encoder SHOULD avoid referencing
-any eviction-prone entries, which are dynamic table entries that may be evicted
-soon.  Rather than reference an eviction-prone entry, the encoder SHOULD emit a
-Duplicate representation (see {{duplicate}}), and reference the duplicate
+any eviction-prone entries, which are dynamic table entries that might be
+evicted soon.  Rather than reference an eviction-prone entry, the encoder SHOULD
+emit a Duplicate representation (see {{duplicate}}), and reference the duplicate
 instead.
 
-To identify eviction-prone entries, the encoder may maintain a draining index,
-which is the smallest index in the dynamic table that it will emit a
+To identify eviction-prone entries, the encoder can maintain a draining index,
+which is the smallest absolute index in the dynamic table that it will emit a
 reference for.  As new entries are inserted, the encoder increments the
 draining index such that the amount of free and draining space in the dyanmic
 table is larger than its target threshold.
+
+~~~~~~~~~~  drawing
+   +----------------+-----------------------+------------+
+   | Eviction-prone |     Referenceable     | Free Space |
+   |    Entries     |        Entries        |            |
+   +----------------+-----------------------+------------+
+   ^                ^                       ^
+   |                |                       |
+ Drop Point    Draining Index           Base Index /
+                                      Insertion Point
+~~~~~~~~~~
+{:#fig-eviction-prone title="Encoder's View of Dynamic Table with
+draining index"}
 
 
 ### Blocked Decoding
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -318,7 +318,7 @@ decoder that permit the encoder to track the decoder's state.  These events are:
 
 Regardless of whether a header block contained blocking references, the
 knowledge that it has been processed permits the encoder to evict
-entries to which no unacknowledged references remain; see {{blocked-eviction}}.
+entries to which no unacknowledged references remain; see {{blocked-insertion}}.
 When a stream is reset or abandoned, the indication that these header blocks
 will never be processed serves a similar function; see {{stream-cancellation}}.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -225,28 +225,28 @@ with the `HTTP_HPACK_DECOMPRESSION_FAILED` error code.
...
+referencing any eviction-prone entries, which are dynamic table entries that may
+be evicted soon.  Rather than reference an eviction-prone entry, the encoder
+SHOULD emit an Indexed-Duplicate representation (see {{indexed-duplicate}}),
+and reference the duplicate instead.
+
+To identify eviction-prone entries, the encoder may maintain a draining index,
+which is the smallest index in the dynamic table that it will emit a
+reference for.  As new entries are inserted, the encoder increments the
+draining index such that the amount of free and draining space in the dyanmic
+table is larger than its target threshold.
```
martinthomson: I'm fairly stupid during meetings that I pay attention to, but I didn't follow this last paragraph very well.  Smallest implies small numbers, but I think that you mean smallest "base" (which is the inverse of indices).

I probably don't really understand what you are trying to say, so consider tweaking for clarity. That might mean more text.

afrind: Sorry, I meant the smallest absolute index.  I wouldn't call it a base, since I think of a base as an absolute index that you encode against.  I'm guessing that alone isn't enough to clarify?

I think of draining space as space in the table occupied by valid entries that the encoder chooses not to use.

Here's the little snippet of code that I use to implement, where `minUsable_` represents my draining index:

```
  while (capacity_ - bytes_ + drainedBytes_ < table_->minFree() &&
         minUsable_ <= baseIndex_) {
    VLOG(5) << "Draining absolute index " << minUsable_;
    drainedBytes_ += (*table_)[absoluteToInternal(minUsable_++)].bytes();

  }
```

draft-ietf-quic-qpack.md
```diff
@@ -557,20 +557,36 @@ the decoder.
...
-avoid encoding delays, and MAY insert the header into the table later if
-desired.
-
-To ensure that the blocked eviction case is rare, references to the oldest
-entries in the dynamic table SHOULD be avoided.  When one of the oldest entries
-in the table is still actively used for references, the encoder SHOULD emit an
-Indexed-Duplicate representation instead (see {{indexed-duplicate}}).
+The encoder MUST NOT emit an instruction that evicts an entry while a reference
+to that entry remains unacknowledged. When the encoder encounters a blocked
+eviction in the course of encoding a header block, it SHOULD emit a literal
```
MikeBishop: SHOULD implies that the encoder has a different option which is appropriate in certain circumstances.  What else could the encoder do, and when would it choose that?  (I don't think it does, in which case I'd just skip the normative language here and say that's what the encoder does.)

draft-ietf-quic-qpack.md
```diff
@@ -557,20 +557,36 @@ the decoder.
...
+which is the smallest absolute index in the dynamic table that it will emit a
+reference for.  As new entries are inserted, the encoder increments the
+draining index such that the amount of free and draining space in the dyanmic
+table is larger than its target threshold.
+
+~~~~~~~~~~  drawing
+   +----------------+-----------------------+------------+
+   | Eviction-prone |     Referenceable     | Free Space |
+   |    Entries     |        Entries        |            |
+   +----------------+-----------------------+------------+
```
MikeBishop: I'm not sure if "Free Space" is necessarily the correct thing here, at least in a steady state (which is the only time eviction is an issue).  I suppose it's sometimes possible to insert something small without evicting anything, but I'm not sure if that's relevant.

draft-ietf-quic-qpack.md
afrind: ok

draft-ietf-quic-qpack.md
afrind: There's usually some free space because it's not possible to evict a partial header.   Maybe I could make it look smaller?  The text talks about the sum of free space and draining space so I wanted to include it in the picture.

martinthomson: The point here is not that there might be some free space (there is almost always free space because header fields are not necessarily a neat fit for the available space), but that the encoder needs to keep its hands off those entries that are close to eviction.  Eviction-prone is misleading, because it implies that there is a probability of eviction.

I think that this map is misleading though.  There are three important classifications here:

* Entries that are near eviction, such that new insertions might be blocked by outstanding references to those entries.
* Entries that are recently added (i.e., those that are "vulnerable"), such that references to them might cause the receiver to block.
* Other entries that are acknowledged, but not near eviction.

The line between those that are "vulnerable" and acknowledged is crisp.  The point here is that the line between near-eviction and merely acknowledged is fuzzy and subject to heuristics.

I think that you also want to say that keeping the near-eviction index in the range of acknowledged entries will prevent perverse behaviour.  If the two ever cross, you can get duplication on every header block, which causes unnecessary churn.  (It's also possible to get that without the two crossing with a naive implementation, but I don't think that we need to get into that too much.)

martinthomson:
Has this been superseded by #1357?

martinthomson:
@MikeBishop, ping.

afrind:
Yes, I'll come back to this.  It was low pri relative to the other design changes I wanted to land for -01

draft-ietf-quic-qpack.md
afrind: I'm not sure I agree that the draining index always needs to be in the acknowledged range.  Consider the case where I insert a full table's worth of headers and none have yet been acknowledged.  An encoder will still want to stop referencing the oldest ones so they can be evicted soon, even if it causes duplication.  For the purposes of this section, I'm not sure I want to add the 'acknowledged index' to the picture.

MikeBishop: I'm with @afrind.  The pathological case I think you're describing is an encoder that won't / isn't permitted to block deciding that it should duplicate an old entry, but still can't use either one.  If it doesn't duplicate it more than once, that should be okay.

draft-ietf-quic-qpack.md
```diff
@@ -871,21 +871,38 @@ the decoder.
...
-avoid encoding delays, and MAY insert the header into the table later if
-desired.
-
-To ensure that the blocked eviction case is rare, references to the oldest
-entries in the dynamic table SHOULD be avoided.  When one of the oldest entries
-in the table is still actively used for references, the encoder SHOULD emit an
-Duplicate representation instead (see {{duplicate}}).
+The encoder MUST NOT emit an instruction that evicts an entry while a reference
+to that entry remains unacknowledged.  When the encoder encounters a blocked
+eviction, it emits a literal representation for the new header, and MAY insert
```
martinthomson: "When the encoder encounters a blocked eviction, ..." involves a logical leap, or an assumption about the structure of encoders that hasn't been explained.  We all assume that encoders emit new table modifications as the result of processing header blocks, but that isn't necessary.

Perhaps: "An encoder MUST NOT insert an entry into the dynamic table (or duplicate an existing entry) if that would cause an entry with unacknowledged references to be evicted.  For header blocks that might rely on the newly added entry, the encoder can use a literal representation and maybe insert the entry later."

draft-ietf-quic-qpack.md
```diff
@@ -871,21 +871,38 @@ the decoder.
 
 ### Blocked Eviction
```
martinthomson: The title here might be better as "Blocked Dynamic Table Insertions".  Evictions don't get blocked, they are the immovable object here.  It's more the case that insertions are what get blocked.

draft-ietf-quic-qpack.md
```diff
@@ -871,21 +871,38 @@ the decoder.
...
-To ensure that the blocked eviction case is rare, references to the oldest
-entries in the dynamic table SHOULD be avoided.  When one of the oldest entries
-in the table is still actively used for references, the encoder SHOULD emit an
-Duplicate representation instead (see {{duplicate}}).
+The encoder MUST NOT emit an instruction that evicts an entry while a reference
+to that entry remains unacknowledged.  When the encoder encounters a blocked
+eviction, it emits a literal representation for the new header, and MAY insert
+the header into the table later if desired.
+
+To ensure that blocked evictions are rare, the encoder SHOULD avoid referencing
```
martinthomson: "To ensure that the encoder is not prevented from adding new entries, the encoder can avoid referencing entries that are near the point at which they would be evicted.  Rather than ..."

draft-ietf-quic-qpack.md
```diff
@@ -871,21 +871,38 @@ the decoder.
...
+   | Draining |          Referenceable          | Unused |
+   | Entries  |             Entries             | Space  |
+   +----------+---------------------------------+--------+
+   ^          ^                                 ^
+   |          |                                 |
+ Drop Point  Draining Index               Base Index /
+                                          Insertion Point
+~~~~~~~~~~
+{:#fig-eviction-prone title="Encoder's View of Dynamic Table with
+draining index"}
```
martinthomson: This has to be a single line.  Also, check whether You Want Title Case or Sentence case.  I think that we use Title Case elsewhere.

draft-ietf-quic-qpack.md
```diff
@@ -871,21 +871,38 @@ the decoder.
...
+
+Determining which entries are too close to eviction to reference is an encoder
+preference.  One heuristic is to target a fixed amount of available space in the
+dynamic table: either unused space or space that can be reclaimed by evicting
+unreferenced entries.  To achieve this, the encoder can maintain a draining
+index, which is the smallest absolute index in the dynamic table that it will
+emit a reference for.  As new entries are inserted, the encoder increments the
+draining index such that the amount of unused space and space occupied by
+draining entries table is larger than its target threshold.  Draining entries
+will eventually become evictable as any references to them are acknowledged.
```
martinthomson: "As new entries are inserted, the encoder increases the draining index to maintain the section of the table that it will not reference.  Draining entries - entries with an absolute index lower than this the draining index - will not accumulate new references.  The number of unacknowledged references to draining entries will eventually become zero, making the entry available for eviction."

Issue #1121, opened 2018-02-21
Separate QCRAM streams


martinthomson:
QCRAM doesn't need to use the control stream for its activities.  I propose two streams:

1. A stream for header updates.

2. A stream for header block acknowledgments.

These could probably be arranged as a bidirectional stream that is initiated by the encoding side with header updates on the send direction and acknowledgments in the other direction.  However, the timing of creation of these streams suggests that unidirectional streams are better.  It is possible that header blocks would need to be acknowledged before the encoder commits any changes to the header table, and having the acknowledgments blocked on the creation of a remote stream would add unnecessary delays.

This separation allows for several advantages and I can't see any significant downsides:

* Framing overheads can be removed entirely (or significantly reduced).  Neither stream requires any framing in practice (header table updates can be a contiguous stream of changes), and acknowledgments can be a continuous sequence of integers.

* An encoder can be given exclusive use of a write stream for sending header updates and a read stream for receiving acknowledgments.  A decoder can be given exclusive access to the reciprocal streams.

* More streams means less unnecessary blocking.  Though there is generally a cost to more streams, these streams have no interaction with other control stream functions (SETTINGS, PRIORITY, etc...).  Increased independence means fewer chances for head-of-line blocking.

LPardue:
Reading between lines, is this to say the QCRAM streams would be a special
type akin to Push streams?

Would you have the first bytes sent on a unidirectional stream be some
magic?

afrind:
We could reserve certain stream numbers for these streams in the HTTP mapping, or we could have some kind of magic as you suggest to identify the purpose of the stream.

I think you would need some framing on the header update stream to define the boundaries where the decoder would send acknowledgements.  It's also somewhat simpler to not start parsing instructions when the data might end in the middle of an instruction.

martinthomson:
You don't send acknowledgments to header updates.  No need.

I prefer the magic number at the start of the stream.  Then I don't need to build a stack that knows about stream identifiers (see also #910, which I support in full).

MikeBishop:
@afrind, this ties to @martinthomson's observation that BLOCKING is unnecessary.  Rather than making the "Base" value be the greatest in the table that was filled when the frame was encoded, you make it the greatest actually used by the frame.  If that's not the newest entry, then it's not blocking, but the decoder doesn't care about that distinction, it just cares whether it has all referenced entries.

This does break your idea to use the HEADERS frames on the control stream as virtual checkpoints, though.

[...]

Issue #1122, opened 2018-02-21
Framing QCRAM payloads


martinthomson:
It's not clear from reading QCRAM what the expected pattern of framing is for the three types of message that it exchanges (header blocks, header updates, and header block acknowledgments).  This could be clearer.  

I think that the document assumes is that these are all frames in the hq sense, but #1121 might make that less necessary.

afrind:
In #1121, there was some discussion about the value in framing header updates within HEADERS frames for two reasons:  1) the length field of headers means that decoders do not need to handle partial instructions and 2) the decoder uses the group of headers included in the HEADERS block as a signal for when to generate a HEADERS_ACK, which updates the encoder's view of decoder state.

There was also discussion about the cost of this framing for table updates.  On my admittedly small sample size of 1 HAR file, adding 6 bytes of framing per control block didn't change the compression ratio.  251 requests across 7 origins generated 29 control stream updates, adding 174 bytes of overhead.  This doesn't include the QUIC framing since you pay that cost even if you don't frame the header updates.

If we move updates off the control stream, per #1121, we can reduce the framing overhead further, at some cost to code reusability.

Issue #910, opened 2017-11-11
Applications and Stream IDs


MikeBishop:
@nibanks previously proposed a QUIC API that didn't expose specific Stream IDs to applications.  That doesn't actually work for HTTP/QUIC as currently defined, but there are only two places where the HTTP/QUIC implementation needs to reference a Stream ID.

- PRIORITY frames reference requests by Stream ID.  The solution to this would be to expand #701 to use a Request ID for client-initiated requests as well as Push ID.  (At first glance, these don't need to be throttled since QUIC MAX_STREAM_ID effectively controls this.)
- The long-lived control stream is a specific stream ID.  This could be eliminated by putting a stream header (again, &agrave; la #701) on unidirectional streams that identifies them as control streams.  We still need the control stream to be unique to support order-dependent frames like PRIORITY.

If HTTP goes this direction, there's an associated question for the abstraction design team as to whether we want to explicitly recommend that the Stream ID is always an internal property of the transport.  (This would make it easier to make changes to Stream IDs and how they work in future versions, incidentally.)

nibanks:
I'd like to qualify my proposal to say that the `stream = connection.OpenNewStream(type)` API would just internally create a new stream with the next available ID for the given type (unidirectional or bidirectional) and wouldn't indicate the Stream ID to the application by default. But instead there would just be a helper function on the Stream to query the ID, `stream.GetID()` for instance.

Would this be enough? Essentially that API would only support creating stream in order, as far as the ID space goes.

MikeBishop:
I think that would be enough to implement HTTP/QUIC now, provided that the HTTP layer ensures the first created unidirectional streams are for the control streams.  So the question becomes whether there's a desire for more isolation than that in HTTP.

martinthomson:
@MikeBishop, I think that making a change here would be helpful in resolving #1121.  Do you think that it is time to start talking about this again?

PR #1238, opened 2018-03-19
Move QPACK control instructions to dedicated streams

afrind:
1. Give QPACK encoders and decoders their own unidirectional streams (hardcoded to 6,7,10,11 for now)
2. Define simple framing for encoder stream
3. Move HEADER_ACK from HTTP over QUIC frame to QPACK instruction
4. Introduce Table State Synchronize instruction

Fixes #1120 and #1121

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -127,19 +127,75 @@ considered blocked by the decoder and can not be processed until all entries in
 the range `[1, Depends Index]` have been added.  While blocked, header
 field data MUST remain in the blocked stream's flow control window.
 
-# HPACK extensions
+# Wire Format
 
-## Allowed Instructions
+QPACK instructions occur in three locations:
 
-HEADERS frames on the Control Stream SHOULD contain only Literal with
-Incremental Indexing and Indexed with Duplication (see {{indexed-duplicate}})
-representations.  Frames on this stream modify the dynamic table state
-without generating output to any particular request.
+ - Table updates are carried by a unidirectional stream from encoder to
+   decoder.
+   Instructions on this stream modify the dynamic table state
+   without generating output to any particular request.
+ - Acknowledgements of table modifications and header processing are carried by
+   a unidirectional stream from decoder to encoder.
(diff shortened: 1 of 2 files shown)


martinthomson:
Rather than hard-code, I would *greatly* prefer that this used a stream header.

afrind:
Can you explain the basis of your preference?  A client has no other unidirectional streams defined in HTTP over QUIC, and a server must process/acknowledge at least one header block before opening any push streams.  It is further very unlikely the server would open a push stream without first encoding a header block that inserts at least one header to the dynamic table.  So in practice it's going to work out to these streams anyways.

How do you see the encode/decode streams as materially different from the control stream, which is hardcoded?

Aside: it occurs to me that a server could compress a PUSH_PROMISE a lot better if it compressed against the client's table, which of course doesn't work ;)

MikeBishop:
Looks reasonable.  The source of the stream header is, I think, the discussion in #910 about eliminating references to particular stream IDs in the application layer.  You can make a QUIC interface that depends only on receiving a stream and reading the bytes to figure out what to do with it, rather than needing to open / listen on *particular* streams by ID.

I think that model is important for things that could crop up mid-connection, but I don't feel as strongly about the header for streams that are unique and persistent.

MikeBishop:
BTW, I think you might have meant #1122 rather than #1120.  #1120 is another PR that has already been merged.

draft-ietf-quic-qpack.md
```diff
@@ -164,14 +164,27 @@ unmodified.
...
+## QPACK Encoder Stream
 
 Table updates can add a table entry, possibly using existing entries to avoid
 transmitting redundant information.  The name can be transmitted as a reference
 to an existing entry in the static or the dynamic table or as a string literal.
 For entries which already exist in the dynamic table, the full entry can also be
 used by reference, creating a duplicate entry.
 
+Each set of encoder insructions is prefaced by its length, encoded as a variable
+length integer with an 8-bit prefix.
```
MikeBishop: We should probably specify that these need to be complete instructions (i.e. no truncation), and require an error if one of the length-prefixed blocks ends mid-instruction.

PR #1245, opened 2018-03-20
Reference RFC6928

### LINKED ISSUE #592: Recovery: Size of MSS and relation to MTU
In draft-ietf-quic-recovery-03, section 4.3: 

   kDefaultMss (default 1460 bytes):  The default max packet size used
      for calculating default and minimum congestion windows.

   kInitialWindow (default 10 * kDefaultMss):  Default limit on the
      amount of outstanding data in bytes.

   kMinimumWindow (default 2 * kDefaultMss):  Default minimum congestion
      window.

The document doesn't discuss MSS more, however it clearly have significant impact on the protocol behavior. Thus, I really like to get some clarifications on how MSS are handled in QUIC. Some questions I have are:

1. What is the relation between MSS and IP layer MTU?
2. Is the constants above intended to be related to actual packets possible to send, where MSS = MTU is important, or only to indicate a certain number of bytes? 
3. What manipulation is allowed of MSS during a connections life time. In QUIC change of MTU has to be expected for several reasons. Thus, should MSS be updated due to MTU changes, and how are they performed?

ianswett:
Partially fixes #592

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -959,8 +959,9 @@ kDefaultMss (default 1460 bytes):
 : The default max packet size used for calculating default and minimum
   congestion windows.
 
-kInitialWindow (default 10 * kDefaultMss):
-: Default limit on the amount of outstanding data in bytes.
+kInitialWindow (default 14600 bytes):
+: Default limit on the initial amount of outstanding data in bytes.
+  Taken from {{?RFC6928}}.
 
 kMinimumWindow (default 2 * kDefaultMss):
 : Default minimum congestion window.


draft-ietf-quic-recovery.md
```diff
@@ -955,15 +955,9 @@ Constants used in congestion control are based on a combination of RFCs,
...
-: The default max packet size used for calculating default and minimum
-  congestion windows.
-
-kInitialWindow (default 10 * kDefaultMss):
-: Default limit on the amount of outstanding data in bytes.
-
-kMinimumWindow (default 2 * kDefaultMss):
-: Default minimum congestion window.
+kInitialWindow (default 14600 bytes):
+: Default limit on the initial amount of outstanding data in bytes.
```
gloinul: I would note that RFC 6928 states the following as the TCP initial window:

   More precisely, the upper bound for the initial window will be

         min (10*MSS, max (2*MSS, 14600))                            (1)

I would recommend that one do calculate this value for the Initial value, as MSS may be based on a 9K MTU, thus allowing two full packets in those cases.

gloinul: Also if mss is smaller than 1460 then more than 10 packets can be sent. Is that intentional? I am slightly worried about packet bursts. Are there a pacing requirement?

martinthomson:
@ianswett, I leave this with you.

PR #1249, opened 2018-03-26
A more complete connection migration

### LINKED ISSUE #880: Probing and connection migration

janaiyengar:
Closes #880, closes #1012.

Moved #1012 here because unicorns are not nice.

This PR describes a more complete connection migration design, including:
- Allowing an endpoint to probe alternate paths without having to use them for data.
- Ensure that an endpoint validates peer ownership of a new address.


draft-ietf-quic-transport.md
```diff
@@ -1451,33 +1454,279 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
-shared state with a connection when they move networks.  This includes state
-that can be hard to recover such as outstanding requests, which might otherwise
-be lost with no easy way to retry them.
+QUIC allows connections to survive changes to endpoint addresses (that is, IP
+address and/or port), such as those caused by a endpoint migrating to a new
+network.  This section describes the process by which an endpoint migrates to a
+new address.
+
+An endpoint MUST NOT initiate connection migration before the handshake is
+finished and the endpoint has 1-RTT keys.
```
martinthomson: Let's say that the handshake is complete instead.  Otherwise the server can migrate too soon.

draft-ietf-quic-transport.md
```diff
@@ -1451,33 +1454,279 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+
+An endpoint uses a new connection ID for probes sent from a new local address,
+see {{migration-linkability}} for further discussion.
+
+Receiving a PATH_CHALLENGE frame from a peer indicates that the peer is probing
+for reachability on a path. An endpoint sends a PATH_RESPONSE in response as per
+{{migrate-validate}}.
+
+PATH_CHALLENGE, PATH_RESPONSE, and PADDING frames are "probing" frames. If an
+endpoint has received packets containing only these frames from a given peer
```
martinthomson: "Until an endpoint receives packets containing other frames from a given peer address, it MUST NOT [...]"

draft-ietf-quic-transport.md
```diff
@@ -1451,33 +1454,279 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
-to receive packets from an older source address.
+Each endpoint validates its peer's address during connection establishment.
+Therefore, a migrating endpoint can send to its peer knowing that the peer is
+willing to receive at the peer's current address. Thus an endpoint can migrate
+to a new local address without first validating the peer's address.
 
-An endpoint MUST validate that its peer can receive packets at the new address
-before sending any significant quantity of data to that address, or it risks
-being used for denial of service.  See {{migrate-validate}} for details.
+However, the new path might not support the endpoint's current sending
```
martinthomson: s/However/When migrating/

draft-ietf-quic-transport.md
```diff
@@ -1451,33 +1454,279 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+rate. Therefore, the endpoint resets its congestion controller, as described in
+{{migration-cc}}.
+
+Receiving acknowledgments for this data serves as proof of the peer's
```
martinthomson: s/this data/data sent on the new path/

draft-ietf-quic-transport.md
```diff
@@ -1451,33 +1454,279 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+to send excessive amounts of data to an unwilling host.  If the endpoint sends
+significantly more data than the spoofing peer, connection migration might be
+used to amplify the volume of data that an attacker can generate toward a
+victim.
+
+As described in {{migration-response}}, an endpoint is required to validate a
+peer's new address to confirm the peer's possession of the new address.  Until a
+peer's address is deemed valid, an endpoint MUST limit the rate at which it
+sends data to this address.  The endpoint MUST NOT send more than a minimum
+congestion window's worth of data per estimated round-trip time (as defined in
```
martinthomson: mention kMinimumWindow

draft-ietf-quic-transport.md
```diff
@@ -1451,33 +1454,279 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+{{QUIC-RECOVERY}}).  In the absence of this limit, an endpoint risks being used
+for a denial of service attack against an unsuspecting victim.  Note that since
+the endpoint will not have any round-trip time measurements to this address, the
+estimate SHOULD be the default initial value (see {{QUIC-RECOVERY}}).
+
+An endpoint SHOULD NOT apply this rate limit when the endpoint skips validation
```
martinthomson: Remove the "SHOULD NOT" (no interop requirement here).  Maybe:

If an endpoint skips validation of a peer address as described in {{migration-responses}}, it does not need to limit its sending rate.

draft-ietf-quic-transport.md
```diff
@@ -1451,33 +1454,279 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+of the peer's address, as described in {{migration-response}}.
+
+
+#### Handling Address Spoofing by an On-path Attacker {#on-path-spoofing}
+
+An on-path attacker could cause a spurious connection migration by capturing and
+forwarding a packet such that it arrives before the legitimate copy of that
+packet.  Such a packet will appear to be a legitimate connection migration and
+the legitimate copy will be dropped as a duplicate. After a spurious migration,
```
martinthomson: Two uses of "legitimate"

s/the legitimate copy will be dropped as a duplicate/the original packet appears to be a duplicate and will be dropped/

draft-ietf-quic-transport.md
```diff
@@ -1451,33 +1454,279 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+validation of the source address will fail because the entity at the source
+address does not have the necessary cryptographic keys to read or respond to the
+PATH_CHALLENGE frame that is sent to it, even if it wanted to.
```
martinthomson: drop from the comma

draft-ietf-quic-transport.md
```diff
@@ -1451,33 +1454,279 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+does not unduly cause the congestion controller to reduce its sending rate.  For
+instance, an endpoint may run a separate alarm when a PATH_CHALLENGE is sent,
+which is disarmed when the corresponding PATH_RESPONSE is received.  If the
+alarm fires before the PATH_RESPONSE is received, the endpoint might send a new
+PATH_CHALLENGE, and restart the alarm for a longer period of time.
 
 
 ### Privacy Implications of Connection Migration {#migration-linkability}
 
+
```
martinthomson: Oops

draft-ietf-quic-transport.md
janaiyengar: I don't follow. It does say "the handshake is finished". What change are you suggesting?

martinthomson: 1-RTT keys != handshake finished, at least for a server.

martinthomson: nevermind, misreading on my part

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,276 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+An endpoint MAY bundle PATH_CHALLENGE and PATH_RESPONSE frames that are used for
+path validation with other frames.  For instance, an endpoint may pad a packet
+carrying a PATH_CHALLENGE for PMTU discovery, or an endpoint may bundle a
+PATH_RESPONSE with its own PATH_CHALLENGE.
+
+
+### Initiation
+
+To initiate path validation, an endpoint sends a PATH_CHALLENGE frame containing
+a random payload that is hard to guess on the path to be validated.
```
ianswett: nit: "that is hard to guess" seems redundant to random, so I would remove it.

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,276 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
-be lost with no easy way to retry them.
+QUIC allows connections to survive changes to endpoint addresses (that is, IP
+address and/or port), such as those caused by a endpoint migrating to a new
+network.  This section describes the process by which an endpoint migrates to a
+new address.
+
+An endpoint MUST NOT initiate connection migration before the handshake is
+finished and the endpoint has 1-RTT keys.
+
+This document limits migration of connections to new client addresses.
```
ianswett: It seems like this is the only paragraph that's client specific now.  Is there a reason to keep this paragraph now?

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,276 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+Receiving acknowledgments for data sent on the new path serves as proof of the
+peer's reachability from the new address.  Note that since acknowledgments may
+be received on any path, return reachability on the new path is not
+established. To establish return reachability on the new path, an endpoint MAY
+concurrently initiate path validation {{migrate-validate}} on the new path.
+
+
+### Responding to Connection Migration {#migration-response}
+
+Receiving a packet from a new peer address containing a non-probing frame
```
ianswett: If a packet contains a PATH_CHALLENGE and a STREAM frame, then it contains a non-probing frame, but I don't think the intent is to immediately migrate.

How about "Receiving a packet from a new peer address that does not contain a PATH_CHALLENGE frame indicates the peer has migrated to that address."

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,276 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+to send excessive amounts of data to an unwilling host.  If the endpoint sends
+significantly more data than the spoofing peer, connection migration might be
+used to amplify the volume of data that an attacker can generate toward a
+victim.
+
+As described in {{migration-response}}, an endpoint is required to validate a
+peer's new address to confirm the peer's possession of the new address.  Until a
+peer's address is deemed valid, an endpoint MUST limit the rate at which it
+sends data to this address.  The endpoint MUST NOT send more than a minimum
+congestion window's worth of data per estimated round-trip time (kMinimumWindow,
```
ianswett: Instead of the minimum window, we could use the limit in response to client initial of 3x, since I think we want migration to be governed by most of the rules of a new connection.

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,276 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+
+### Loss Detection and Congestion Control {#migration-cc}
+
+The capacity available on the new path might not be the same as the old path.
+Packets sent on the old path SHOULD NOT contribute to the state of the
+congestion control and RTT estimation on the new path. If it is using a single
+congestion controller and round-trip time estimator, the client SHOULD reset
+them prior to sending any non-probing packets from a new local address.
+
+After successful validation of the client's new address, the server SHOULD reset
```
ianswett: This is almost identical to the end of the paragraph above it.

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,276 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+is significantly reduced, ultimately this relies on the congestion controller
+responding to congestion signals and reducing send rates appropriately.
+
+There may be apparent reordering at the receiver when an endpoint sends data and
+probes from/to multiple addresses during the migration period, since the two
+resulting paths may have different round-trip times.  A receiver of packets on
+multiple paths will still send ACK frames covering all received packets.
+
+An endpoint MAY use a single congestion control context and a single loss
+recovery context, as described in {{QUIC-RECOVERY}}.  A sender MAY make
```
ianswett: MAY -> SHOULD

draft-ietf-quic-transport.md
MikeBishop: There's more that's needed for a server migration, even though the mechanisms we have so far can be used in both.  I have a separate branch with the immediate-server-migration case, and we can follow on with "full" server migration if time permits.

draft-ietf-quic-transport.md
MikeBishop: I think that's a substantive change to the design.  If you don't intend to migrate, you're not using that path for anything except probing.  If you're including a STREAM frame, then you are migrating, though you might also want to explicitly probe to be sure the migration is working.

ianswett: Above the text says it's ok to bundle other frames with PATH_CHALLENGE, so I assumed STREAM frames were ok.  Is that not true?  Whatever we decide, this should be clear.

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,276 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+from the same local address on which the PATH_CHALLENGE was received, to the
+same remote address from which the PATH_CHALLENGE was received.
+
+
+### Completion
+
+A new address is considered valid when a PATH_RESPONSE frame is received
+containing data that was sent in a previous PATH_CHALLENGE. Receipt of an
+acknowledgment for a packet containing a PATH_CHALLENGE frame is not adequate
+validation, since the acknowledgment can be spoofed by a malicious client.
```
MikeBishop: "client" => "peer"

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,276 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
 connection if the integrity check fails with a PROTOCOL_VIOLATION error code.
 
 
+## Path Validation {#migrate-validate}
+
+Path validation is used by an endpoint to verify reachability of a peer over a
```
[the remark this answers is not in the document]
janaiyengar: The last sentence in this para is supposed to capture that. It's not bidirectional in the sense that the peer has to do its own validation... so saying "both directions" can be a bit misleading

draft-ietf-quic-transport.md
janaiyengar: What Mike said.

draft-ietf-quic-transport.md
janaiyengar: It's ok to bundle other frames, as the server might, when it's verifying client ownership of the new address. But if a client sends anything but probing frames to the server, the server considers this as a migration, not a probe.

draft-ietf-quic-transport.md
janaiyengar: I'd rather leave this at min cwnd for now, and we can fix this later if necessary.

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,275 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+An endpoint MAY bundle PATH_CHALLENGE and PATH_RESPONSE frames that are used for
+path validation with other frames.  For instance, an endpoint may pad a packet
+carrying a PATH_CHALLENGE for PMTU discovery, or an endpoint may bundle a
+PATH_RESPONSE with its own PATH_CHALLENGE.
+
+
+### Initiation
+
+To initiate path validation, an endpoint sends a PATH_CHALLENGE frame containing
+a random payload on the path to be validated.
```
ekr: The requirement is not that the payload be *random* but rather that it be *unpredictable*. For instance, it would be fine to have the payload be generated as:
```
retransmission # || HMAC(key, remote IP || retransmission #)
```

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,275 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+
+
+### Response
+
+On receiving a PATH_CHALLENGE frame, an endpoint MUST respond immediately by
+echoing the data contained in the PATH_CHALLENGE frame in a PATH_RESPONSE frame,
+with the following stipulation.  Since a PATH_CHALLENGE might be sent from a
+spoofed address, an endpoint MAY limit the rate at which it sends PATH_RESPONSE
+frames and MAY silently discard PATH_CHALLENGE frames that would cause it to
+respond at a higher rate.
```
ekr: I actually don't think that this is necessary. Here is my reasoning. PATH_CHALLENGES are cryptographically protected, and therefore if you are doing duplicate suppression on incoming messages, any spoofed PATH_CHALLENGE will be responded to only once, so as long as the other side is reasonable about how many PATH_CHALLENGEs it generates, then an attacker cannot cause you to send to quickly.

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,275 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+A new address is considered valid when a PATH_RESPONSE frame is received
+containing data that was sent in a previous PATH_CHALLENGE. Receipt of an
+acknowledgment for a packet containing a PATH_CHALLENGE frame is not adequate
+validation, since the acknowledgment can be spoofed by a malicious peer.
+
+For path validation to be successful, a PATH_RESPONSE frame MUST be received
+from the same remote address to which the corresponding PATH_CHALLENGE was
+sent. If a PATH_RESPONSE frame is received from a different remote address than
+the one to which the PATH_CHALLENGE was sent, path validation is considered to
+have failed, even if the data matches that sent in the PATH_CHALLENGE.
```
ekr: What does "have failed" mean? That you abort the entire challenge operation? That gives an attacker a trivial way to stop migration. Or does it mean something else.

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,275 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+
+Additionally, the PATH_RESPONSE frame MUST be received on the same local address
+from which the corresponding PATH_CHALLENGE was sent.  If a PATH_RESPONSE frame
+is received on a different local address than the one from which the
+PATH_CHALLENGE was sent, path validation is considered to have failed, even if
+the data matches that sent in the PATH_CHALLENGE.  Thus, the endpoint considers
+the path to be valid when a PATH_RESPONSE frame is received on the same path
+with the same payload as the PATH_CHALLENGE frame.
```
ekr: Same question here as above.

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,275 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+### Probing a New Path
+
+An endpoint MAY probe for peer reachability from a new local address using path
+validation {{migrate-validate}} prior to migrating the connection to the new
+local address.  Failure of path validation simply means that the new path is not
+usable for this connection.  Failure to validate a path does not cause the
+connection to end unless there are no valid alternative paths available.
+
+An endpoint uses a new connection ID for probes sent from a new local address,
+see {{migration-linkability}} for further discussion.
```
ekr: Does this mean that if the other side hasn't provided enough connection IDs you are just stalled?

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,275 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+risks being used for a denial of service attack against an unsuspecting victim.
+Note that since the endpoint will not have any round-trip time measurements to
+this address, the estimate SHOULD be the default initial value (see
+{{QUIC-RECOVERY}}).
+
+If an endpoint skips validation of a peer address as described in
+{{migration-response}}, it does not need to limit its sending rate.
+
+
+#### Handling Address Spoofing by an On-path Attacker {#on-path-spoofing}
```
ekr: The procedure here doesn't seem like it works very well, as the attacker can force you onto a new path for quite some time by spoofing a single packet. you recover when one of two things happen:

1. You receive a higher-numbered packet.
2. Validation fails.

But both of these could take a really long time if one side is kind of quiet.

I have a nice attack that works if you allow server migration as well, but even this seems like it's not great.

draft-ietf-quic-transport.md
```diff
@@ -1451,29 +1454,275 @@ failure.  If integrity protection is performed by QUIC, QUIC MUST abort the
...
+network.  This section describes the process by which an endpoint migrates to a
+new address.
+
+An endpoint MUST NOT initiate connection migration before the handshake is
+finished and the endpoint has 1-RTT keys.
+
+This document limits migration of connections to new client addresses.
+Migrating a connection to a new server address is left for future work. If a
+client receives packets from a new server address, the client MAY discard these
+packets.
```
ekr: Isn't the answer that it MUST NOT follow the other procedures.

ekr:
To add to my previous comment about attacks, in cases where one side has totally migrated (i.e., only one path is valid), then an attacker can also mess things up pretty severely by just spoofing a *response* to a PATH_CHALLENGE. I.e.,

- I am receiving packets from A
- A migrates and so sends me a packet from A'
- I send a PATH_CHALLENGE to A'
- If the attacker forwards the next packet from A' (likely a PATH_RESPONSE) but rewrites the source address to B, then I will consider the validation failed.

PR #1251, opened 2018-03-27
Server's preferred address

### LINKED ISSUE #560: IP Anycast with server IP+Port renegotiation
Many large DNS server deployments rely on IP Anycast for directing UDP requests to the "closest" server. As we consider QUIC for DNS, it would be nice to support the same kind of deployment. However, DNS over QUIC requires server-side state. The instantaneous nature of IP routing could cause anycast packets to be delivered to the "wrong" server - a server instance that does not hold the state for the connection.

This could be solved if after the initial exchange the server indicated an alternative  non-anycast IP address, and maybe an alternative port number. The client could then switch to using that address, assuring then that the QUIC connection would remain stable even if IP routing did change.

There is are obvious security issues, but they can be mitigated. The server indication would need be protected against spoofing, which probably means that it would have to come as a 1-RTT packet. A malicious server could redirect client traffic towards an unsuspecting party, in some kind of DOS attack. This can be mitigated if the client sends a verification packet to the new address, and only starts using the new address for "regular" traffic if the verification packet is acknowledged.

MikeBishop:
Fixes #560; this is a specific post-handshake handoff by the server to a new IP address.  A more general ability to hand off to a different server address is left for the future, though it would probably take the form of a frame which is substantially similar to the new transport option.

One new requirement which isn't really specific to this but which this makes more obvious:  If you're trying to do connection migration, but you get a Stateless Reset on the new path, that doesn't mean the connection gets dropped -- it just means that IP from your new address took you to the wrong server, and your new address can't be used to continue the connection.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -4283,6 +4283,10 @@ thanks to all.
 
 Issue and pull request numbers are listed with a leading octothorp.
 
+## Since draft-ietf-quic-transport-11
+
+- Enable server to prompt migration to a preferred address (#560,#1251).
+
 ## Since draft-ietf-quic-transport-10
 
 - Swap payload length and packed number fields in long header (#1294)
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -762,9 +762,8 @@ A Version Negotiation ({{packet-version}}) packet MUST use both connection IDs
 selected by the client, swapped to ensure correct routing toward the client.
 
 The connection ID can change over the lifetime of a connection, especially in
-response to connection migration ({{migration-client}}}). NEW_CONNECTION_ID
-frames ({{frame-new-connection-id}}) are used to provide new connection ID
-values.
+response to connection migration ({{migration}}}). NEW_CONNECTION_ID frames
+({{frame-new-connection-id}}) are used to provide new connection ID values.
 
 ## Packet Numbers {#packet-numbers}
 
@@ -904,8 +903,8 @@ connection establishment intertwines version negotiation with the cryptographic
 and transport handshakes to reduce connection establishment latency, as
 described in {{handshake}}.  Once established, a connection may migrate to a
 different IP or port at either endpoint, due to NAT rebinding or mobility, as
-described in {{migration-client}}}.  Finally a connection may be terminated by
-either endpoint, as described in {{termination}}.
+described in {{migration}}}.  Finally a connection may be terminated by either
+endpoint, as described in {{termination}}.
 
 ## Matching Packets to Connections {#packet-handling}
 
@@ -1262,7 +1261,7 @@ stateless_reset_token (0x0006):
 preferred_address (0x0004):
 
 : The server's Preferred Address is used to effect a server address migration at
-  the end of the handshake, as described in {{migration-server}}.
+  the end of the handshake, as described in {{preferred-address}}.
 
 A client MUST NOT include a stateless reset token or a preferred address.  A
 server MUST treat receipt of either transport parameter as a connection error of
@@ -1491,7 +1490,7 @@ therefore unlikely to be successful.
 This token can be provided to the cryptographic handshake immediately after
 establishing a connection.  QUIC might also generate an updated token if
 significant time passes or the client address changes for any reason (see
-{{migration-client}}}).  The cryptographic handshake is responsible for
+{{migration}}}).  The cryptographic handshake is responsible for
 providing the client with the token.  In TLS the token is included in the ticket
 that is used for resumption and 0-RTT, which is carried in a NewSessionTicket
 message.
@@ -1527,8 +1526,8 @@ and a specific peer address, where an address is the two-tuple of IP address and
 port.  Path validation tests that packets can be both sent to and received from
 a peer.
 
-Path validation is used during connection migration (see {{migration-client}}
-and {{migration-server}}) by the migrating endpoint to verify reachability of a
+Path validation is used during connection migration (see {{migration}} and
+{{preferred-address}}) by the migrating endpoint to verify reachability of a
 peer from a new local address. Path validation is also used by the peer to
 verify that the migrating endpoint is able to receive packets sent to the its
 new address.  That is, that the packets received from the migrating endpoint do
@@ -1623,7 +1622,7 @@ failure. Primarily, this happens if a connection migration to a new path is
 initiated while a path validation on the old path is in progress.
 
 
-## Client Connection Migration {#migration-client}
+## Connection Migration {#migration}
 
 QUIC allows connections to survive changes to endpoint addresses (that is, IP
 address and/or port), such as those caused by a endpoint migrating to a new
@@ -1872,31 +1871,32 @@ number. "packet_number_secret" is derived from the TLS key exchange,
 as described in Section 5.6 of {{QUIC-TLS}}.
 
 
-## Server Connection Migration {#migration-server}
+## Server's Preferred Address {#preferred-address}
 
 QUIC allows servers to accept connections on one IP address and attempt to
 transfer these connections to a more preferred address shortly after the
 handshake.  This section describes the protocol for migrating a connection to a
-new server address.
+preferred server address.
 
 Migrating a connection to a new server address mid-connection is left for future
 work. If a client receives packets from a new server address not indicated by
 the preferred_address transport parameter, the client SHOULD discard these
 packets.
 
-### Initiating Connection Migration
+### Communicating A Preferred Address
 
-A server initiates connection migration by including the preferred_address
+A server conveys a preferred address by including the preferred_address
 transport parameter in the TLS handshake.
 
 Once the handshake is finished, the client MUST initiate path validation (see
-{{migrate-validate}}) of the server's preferred IP address using the connection
-ID provided in the preferred_address transport parameter.
+{{migrate-validate}}) of the server's preferred address using the connection ID
+provided in the preferred_address transport parameter.
 
 If path validation succeeds, the client SHOULD immediately begin sending all
 non-probe packets to the new server address using the new connection ID and
 discontinue use of the old server address.  If path validation fails, the client
 MUST continue sending all future packets to the server's original IP address.
+
 
 ### Responding to Connection Migration
 
@@ -1911,12 +1911,12 @@ is the largest packet number seen so far, the server begins sending to the
 client exclusively from its preferred IP address.
 
 
-## Interaction of Client and Server Migration
+### Interaction of Client Migration and Preferred Address
 
-A client might need to perform a connection migration before the server's
-connection migration has completed.  In this case, the client SHOULD perform
-path validation to both the original and preferred server address from the
-client's new address concurrently.
+A client might need to perform a connection migration before it has migrated to
+the server's preferred address.  In this case, the client SHOULD perform path
+validation to both the original and preferred server address from the client's
+new address concurrently.
 
 If path validation of the server's preferred address succeeds, the client MUST
 abandon validation of the original address and migrate to using the server's
@@ -1929,6 +1929,7 @@ address, the server MUST protect against potential attacks as described in
 {{address-spoofing}} and {{on-path-spoofing}}.  In addition to intentional
 simultaneous migration, this might also occur because the client's access
 network used a different NAT binding for the server's preferred address.
+
 
 ## Connection Termination {#termination}
 
@@ -1994,11 +1995,11 @@ draining.  A key update might prevent the endpoint from moving from the closing
 state to draining, but it otherwise has no impact.
 
 An endpoint could receive packets from a new source address, indicating a client
-connection migration ({{migration-client}}}), while in the closing period. An
-endpoint in the closing state MUST strictly limit the number of packets it sends
-to this new address until the address is validated (see {{migrate-validate}}). A
-server in the closing state MAY instead choose to discard packets received from
-a new source address.
+connection migration ({{migration}}}), while in the closing period. An endpoint
+in the closing state MUST strictly limit the number of packets it sends to this
+new address until the address is validated (see {{migrate-validate}}). A server
+in the closing state MAY instead choose to discard packets received from a new
+source address.
 
 
 ### Idle Timeout
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1153,10 +1153,10 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
    } TransportParameters;
 
    struct {
-     enum { IPv4(4), IPv6(6), (15)} ip_version;
-     opaque ip_address<4..2^8-1>;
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1792,6 +1812,64 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+
+### Responding to Connection Migration
+
+A server might receive a packet addressed to its preferred IP address at any
+time during the connection after the handshake is completed.  If this packet
+contains a PATH_CHALLENGE frame, the server sends a PATH_RESPONSE frame as per
+{{migrate-validate}}, but the server MUST continue sending all other packets
+from its original IP address.
+
+Once the server has received a non-probing packet on its preferred address which
```
ianswett: I'm still confused on what probing and non-probing packets are.  Did I just miss the definition?

MikeBishop: Line 1612 defines the "probing frames" and talks about "packets containing other frames," but you're correct that there's not as clear a definition as I'd thought.  Let's add that in the top-level migration PR.  @janaiyengar?

ianswett: @janaiyengar Already merged that, so I'd suggest we clarify it in this PR?

ianswett:
And thanks for writing this up Mike.  This is a great traffic management feature, and it's great to have this in an early draft, so different groups can ensure it works as intended.

draft-ietf-quic-transport.md
```diff
@@ -1791,6 +1811,64 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+work. If a client receives packets from a new server address not indicated by
+the preferred_address transport parameter, the client SHOULD discard these
+packets.
+
+### Initiating Connection Migration
+
+A server initiates connection migration by including the preferred_address
+transport parameter in the TLS handshake.
+
+Once the handshake is finished, the client MUST initiate path validation (see
```
janaiyengar: Requiring the client to use the server's preferred address seems unnecessary.

draft-ietf-quic-transport.md
```diff
@@ -1791,6 +1811,64 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+{{migrate-validate}}) of the server's preferred IP address using the connection
+ID provided in the preferred_address transport parameter.
+
+If path validation succeeds, the client SHOULD immediately begin sending all
+non-probe packets to the new server address using the new connection ID and
```
janaiyengar: s/non-probe/non-probing/

draft-ietf-quic-transport.md
janaiyengar: Fixed in #1253

draft-ietf-quic-transport.md
```diff
@@ -1791,6 +1811,64 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+
+A server might receive a packet addressed to its preferred IP address at any
+time during the connection after the handshake is completed.  If this packet
+contains a PATH_CHALLENGE frame, the server sends a PATH_RESPONSE frame as per
+{{migrate-validate}}, but the server MUST continue sending all other packets
+from its original IP address.
+
+Once the server has received a non-probing packet on its preferred address which
+is the largest packet number seen so far, the server begins sending to the
+client exclusively from its preferred IP address.
```
janaiyengar: What should the server do if a subsequent packet from the client appears on the old address?

draft-ietf-quic-transport.md
```diff
@@ -1791,6 +1811,64 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+
+
+## Interaction of Client and Server Migration
+
+A client might need to perform a connection migration before the server's
+connection migration has completed.  In this case, the client SHOULD perform
+path validation to both the original and preferred server address from the
+client's new address concurrently.
```
janaiyengar: This is tricky. Where should the client send data if its old address is unusable?
(I want to spend more cycles thinking about this, but I'll be out. Do make sure that all corner cases are covered; I worry that there are more here.)

draft-ietf-quic-transport.md
```diff
@@ -1791,6 +1811,64 @@ number. "packet_number_secret" is derived from the TLS key exchange,
 as described in Section 5.6 of {{QUIC-TLS}}.
 
 
+## Server Connection Migration {#migration-server}
```
janaiyengar: I don't think this should be called Server Connection Migration. There's a symmetry with Client Connection Migration that this name suggests, but that's not true, since the assumptions are quite different. How about "Using a Preferred Server Address"?

Along the same lines, I think you should state any assumptions here clearly, since these have implications. I can think of at least one: The server is not actually changing network attachment points. I believe this assumption is implicit in this design.

draft-ietf-quic-transport.md
```diff
@@ -1791,6 +1811,64 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+
```
janaiyengar: I would add a short description of where this is expected to be used.

draft-ietf-quic-transport.md
martinthomson: I like (and prefer) the suggested name change.

When you say that not changing network attachment is a necessary assumption, why is that important?  The new address will be validated and congestion control rebooted in the same way that any other migration would be.  So even if this points to a completely different server instance (with keys being shipped across as needed), it still works.

draft-ietf-quic-transport.md
martinthomson: This feature is only effective if it can be relied upon.  That said, it's a big deal, so we should talk about it.

mikkelfj:
This proposal is sort of acting like ones own DNS server. I'm not against doing this, but I think are corner cases to consider.
- The same server can use multiple load balancers and it is normal (and slow) to have clients fail over when one load balancer goes belly up. A server that wants to move the client may want to list multiple IP addresses on the new destination. It could be resolved by randomly picking one per connection, only the server might not know that one is down.
- Is it always the case that a server wends on the same address as it receives (via translation in path if necessary)?
- Certs are tied an IP and changing IP is an attack vector on the servers configuration system - though being able to present a cert on the original addres gives some level comfort.

draft-ietf-quic-transport.md
huitema: Let's keep it simple. This PR is about triggering a migration at the end of the handshake. Given the prevalence of NAT, the path challenge sent by the client to a new server address will very often arrive at the server with a different source address than the source address of the client initial packet, and there is not much the client can do about that. The default source address of the client might also change, but from a server point of view that's just the same as any client migration. So in most cases we will see "challenge from client, response and challenge from server, response from client." Both server and clients will have to consider whether sending data before the verification completes, but that's what they always do.

draft-ietf-quic-transport.md
```diff
@@ -1485,6 +1505,11 @@ path validation with other frames.  For instance, an endpoint may pad a packet
 carrying a PATH_CHALLENGE for PMTU discovery, or an endpoint may bundle a
 PATH_RESPONSE with its own PATH_CHALLENGE.
 
+Differences in routing on the Internet might cause the same destination address
+and connection ID to reach a different server instance which does not posses the
+necessary connection state. Receiving a Stateless Reset in response to a probing
+packet SHOULD NOT terminate the connection, but MUST cause the endpoint to
+consider path validation to have failed.
```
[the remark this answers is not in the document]
huitema: Really? The use case is that the server sends both a unicast address and a connection ID. In theory, the unicast address guarantees that the packet will reach the specific server farm that sent the redirect, and the connection ID guarantees that the load balancer will send the packet to the right server in the farm. Receiving a stateless reset is an indication that something is seriously wrong.

Or maybe you are not addressing this specific PR, but rather making a generic statement about probes. In which case I would suggest moving this to a different PR.

draft-ietf-quic-transport.md
MikeBishop: I think the necessary assumption is that the server maintains the ability to send/receive from both addresses through the whole process.  So if the client's attempt to hand off fails, it can still continue with the original address.  In the more general P2P case, the server might not actually continue to have access to the original address, because it is actually changed network connections.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
MikeBishop: Well, there are two pieces.  NAT rebinding means that, even if the client doesn't change its local address, the server will see a simultaneous migration.  So forbidding it seems problematic, since the server can't enforce it.  If it can't be totally forbidden, we could say that you SHOULD NOT attempt to perform a local migration until the handoff has completed.

I'm not convinced that abandoning one probe (because that local address is gone) and starting two in parallel is a combinatorial explosion -- it's two paths to probe, with an embedded preference between them if they both work.

draft-ietf-quic-transport.md
MikeBishop: The server can't actually know if the client's probe fails to reach it for some reason, so it's going to have to tolerate clients failing to switch anyway.  This could probably become a SHOULD without changing logic on anyone's side.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
MikeBishop: Yes, it's something I realized when discussing the simultaneous migration case.  If the client probes both the original (anycast) and preferred (unicast) address from the new interface, the result is a race between a Stateless Reset from the anycast peer and a PATH_RESPONSE from the unicast peer.  Since the anycast endpoint is likely to be closer, the Stateless Reset will likely win.

However, it's not specific to this PR, so it's now #1259.

MikeBishop:
For simplicity, I'm inclined to say it's the server's job to send a single working IP address to the client.  When we have a more complete server migration story, the server can just try multiple times / send multiple addresses.

Yes, just like in path validation, the server responds on the same address to the same address.

I don't see this as an attack, primarily because it's under the server's control.  The server could use a dedicated IP per cert to do handshakes, then hand off to a shared IP for existing connections.  No certs involved there.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
martinthomson: It increases the complexity of implementations.  Right now, you probe one path, and if a new path presents itself, you abandon any previous path.  Having to move from 1 to N (even if N == 2), is more complex than we need.  The odds of needing to move just after the handshake are low enough that I would prefer to just say that the client can hold off on migration until after the server completes.  I guess that your response is that this only affects discretionary migrations.  But we can at least write that consideration down.

MikeBishop:
Removed the dependency on #1259.

draft-ietf-quic-transport.md
```diff
@@ -1249,9 +1258,14 @@ stateless_reset_token (0x0006):
 : The Stateless Reset Token is used in verifying a stateless reset, see
   {{stateless-reset}}.  This parameter is a sequence of 16 octets.
 
-A client MUST NOT include a stateless reset token.  A server MUST treat receipt
-of a stateless_reset_token transport parameter as a connection error of type
-TRANSPORT_PARAMETER_ERROR.
+preferred_address (0x0004):
+
+: The server's Preferred Address is used to effect a server address migration at
```
janaiyengar: s/a server address migration/change of a server address/

draft-ietf-quic-transport.md
```diff
@@ -1852,6 +1871,73 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+time after the handshake is completed.  If this packet contains a PATH_CHALLENGE
+frame, the server sends a PATH_RESPONSE frame as per {{migrate-validate}}, but
+the server MUST continue sending all other packets from its original IP address.
+
+The server SHOULD also initiate path validation of the client using its
+preferred address and the address from which it received the client probe.  This
+helps to guard against spurious migration initiated by an attacker.
+
+Once the server has completed its path validation and has received a non-probing
+packet on its preferred address which is the largest packet number seen so far,
```
janaiyengar: How about: s/has received a non-probing packet on its preferred address which is the largest packet number seen so far/has received a non-probing packet with a new largest packet number  on its preferred address/

draft-ietf-quic-transport.md
```diff
@@ -1791,6 +1811,64 @@ number. "packet_number_secret" is derived from the TLS key exchange,
...
+abandon validation of the original address and migrate to using the server's
+preferred address.  If path validation of the server's preferred address fails,
+but validation of the server's original address succeeds, the client MAY migrate
+to using the original address from the client's new address.
+
+If the connection to the server's preferred address is not from the same client
+address, the server MUST protect against potential attacks as described in
+{{address-spoofing}} and {{on-path-spoofing}}.  In addition to intentional
+simultaneous migration, this might also occur because the client's access
+network used a different NAT binding for the server's preferred address.
```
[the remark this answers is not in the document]
janaiyengar: Agree with Martin -- I think this needs more text.

draft-ietf-quic-transport.md
```diff
@@ -1150,6 +1151,14 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
       };
       TransportParameter parameters<22..2^16-1>;
    } TransportParameters;
+
+   struct {
+     enum { IPv4(4), IPv6(6), (15)} ipVersion;
+     opaque ipAddress<4..2^8-1>;
```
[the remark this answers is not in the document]
MikeBishop: The length prefix has to be at least one byte to accommodate any IP addresses; since it's a byte, it can represent a length up to 255 bytes.  Obviously, it wouldn't actually be 255 for IPv4 or IPv6.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
MikeBishop: I've added some text, but it feels duplicative of {{address-spoofing}}.  Concrete suggestions of "more text" would be welcome.

rpaulo:
This looks good to me!

draft-ietf-quic-transport.md
```diff
@@ -1150,6 +1151,14 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
       };
       TransportParameter parameters<22..2^16-1>;
    } TransportParameters;
+
+   struct {
+     enum { IPv4(4), IPv6(6), (15)} ipVersion;
```
martinthomson: unbalanced space within curlies

GuShuhengSimon:
@MikeBishop, seems this support is removed ? now it's covered by {Connection Migration} part?
Then server's anycast IP => unicast IP migration would still not well supported by QUIC currently? Since the {Connection Migration} requires the "endpoint"(Server in this case) sustain a stable address, and need a full handshake being performed, but in this case, the next CHLO from client during 1-RTT would possibly routed to another "anycast" IP servers which makes anyCast IP => uniCast IP failure.

DavidSchinazi:
Hi @GuShuhengSimon this feature is supported by QUIC. It uses the `preferred_address` transport parameter.

See [Section 9.6 of draft-19](https://tools.ietf.org/html/draft-ietf-quic-transport-19#section-9.6)

PR #1253, opened 2018-03-27
Probing packet

janaiyengar:
Define probing/non-probing packet and frames.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1570,8 +1570,8 @@ finished and the endpoint has 1-RTT keys.
 
 This document limits migration of connections to new client addresses.
 Migrating a connection to a new server address is left for future work. If a
-client receives packets from a new server address, the client MAY discard these
-packets.
+client receives packets from an unknown server address, the client MAY discard
+these packets.
 
 
 ### Probing a New Path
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1589,9 +1589,13 @@ Receiving a PATH_CHALLENGE frame from a peer indicates that the peer is probing
...
 
-PATH_CHALLENGE, PATH_RESPONSE, and PADDING frames are "probing" frames. Until an
-endpoint receives packets containing other frames from a given peer address, it
-MUST NOT send other types of frames to that address.
+PATH_CHALLENGE, PATH_RESPONSE, and PADDING frames are "probing frames", and all
+other frames are "non-probing frames".  A packet containing only probing frames
+is a "probing packet", and a packet containing any other frame is a "non-probing
+packet".
+
+An endpoint MUST NOT send non-probing frames to a peer address until it receives
```
ianswett: Is this statement true?  What about NAT rebinds?

MikeBishop: If this statement is strictly followed, no one ever migrates because they're waiting for the other side to do so.  (At which the other side has violated a MUST, and you really ought to kill the connection to that naughty endpoint!)

For both client and server migration, this actually points in the same direction, so let's accept the non-generic here:  "A server MUST NOT send non-probing packets to a client address until it receives a non-probing packet from that address."

PR #1254, opened 2018-03-28
Remove unnecessary text from #1253

martinthomson:
This paragraph:

> A server MUST NOT send non-probing frames to a client's address until the server
> receives a non-probing packet from that address.

Is true only with the restriction of migration to client addresses.  It
complicates the server migration story if we ever want to do that (as we
already have with the proposal for preferred addresses).

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1594,9 +1594,6 @@ other frames are "non-probing frames".  A packet containing only probing frames
 is a "probing packet", and a packet containing any other frame is a "non-probing
 packet".
 
-A server MUST NOT send non-probing frames to a client's address until the server
-receives a non-probing packet from that address.
-
 
 ### Initiating Connection Migration {#initiating-migration}
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1569,12 +1569,11 @@ An endpoint MUST NOT initiate connection migration before the handshake is
 finished and the endpoint has 1-RTT keys.
 
 This document limits migration of connections to new client addresses.
-Additionally, clients are responsible for initiating all migrations.  Servers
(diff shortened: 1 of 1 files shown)


MikeBishop:
No, actually, we discussed this in the editors' meeting.  It's true for server migration as well -- the server prompts the client to probe the new address, but there too, the server does not switch until the client has probed successfully and switched over.  In other words, there's an implicit assumption that the server has access to both addresses and never has a "make-after-break" transition.

What it's not necessarily true for is P2P; at that point, you really need ICE or similar.

martinthomson:
I still think that the text has to go.  What value does it provide?  I mean, other than making it hard to potentially enable server migration, that is.

MikeBishop:
In the client case, it provides the value that it's the client that decides when to switch.  The server can't just decide to start sending you data on an address that you probed from but didn't intend to switch to.  In the server case, it means that the client still drives the point of handover, because only the client knows when all probes have completed successfully.

In a future generic server migration case, a server might decide it's going to switch, but it still needs the client to initiate traffic to that address for NAT purposes.  However, that's probing, so this text need not apply.

draft-ietf-quic-transport.md
```diff
@@ -1569,12 +1569,15 @@ An endpoint MUST NOT initiate connection migration before the handshake is
 finished and the endpoint has 1-RTT keys.
 
 This document limits migration of connections to new client addresses.
-Migrating a connection to a new server address is left for future work. If a
-client receives packets from an unknown server address, the client MAY discard
-these packets.
+Additionally, clients are responsible for initiating all migrations.  Servers
+do not send non-probing packets (see {{probing}}) toward a client address until
+it sees a non-probing packet from that address.  Additionally, if a client
```
MikeBishop: Two "additionally"s in three lines seems a bit redundant, as well as repetitive.  Perhaps drop the second one?

PR #1263, opened 2018-04-04
Avoid failure from spoofed/delayed PATH_RESPONSE

martinduke:
Bogus PATH_RESPONSE in a Handshake packet seems like an obvious DoS attack.

Also, I added some language about storing PATH_CHALLENGE data for a really long time if we're going to send an error. I would actually prefer that we get rid of this error code entirely, as it seems like a great way to accidentally tear down connections when the new path's RTT is way higher than expected.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2784,8 +2784,11 @@ frame.  Its format is identical to the PATH_CHALLENGE frame
 
 If the content of a PATH_RESPONSE frame does not match the content of a
 PATH_CHALLENGE frame previously sent by the endpoint, the endpoint MAY generate
-a connection error of type UNSOLICITED_PATH_RESPONSE.
-
+a connection error of type UNSOLICITED_PATH_RESPONSE. Endpoints MUST NOT
+generate this error in response to a PATH_RESPONSE frame in a Handshake Packet.
+Endpoints SHOULD NOT generate this error if they are not storing the
+PATH_CHALLENGE data well in excess of any plausible packet round-trip time on
+a new path, as a late response would result in connection teardown.
 
 ## STREAM Frames {#frame-stream}
 


draft-ietf-quic-transport.md
```diff
@@ -2784,8 +2784,11 @@ frame.  Its format is identical to the PATH_CHALLENGE frame
 
 If the content of a PATH_RESPONSE frame does not match the content of a
 PATH_CHALLENGE frame previously sent by the endpoint, the endpoint MAY generate
-a connection error of type UNSOLICITED_PATH_RESPONSE.
-
+a connection error of type UNSOLICITED_PATH_RESPONSE. Endpoints MUST NOT
+generate this error in response to a PATH_RESPONSE frame in a Handshake Packet.
```
martinthomson: I don't think that we need to worry about this.  We have agreed not to specifically defend against tampering with the handshake.  Note that you need to be on-path for this to work, because of the handshake packet protection design.

draft-ietf-quic-transport.md
```diff
@@ -2784,8 +2784,11 @@ frame.  Its format is identical to the PATH_CHALLENGE frame
...
+Endpoints SHOULD NOT generate this error if they are not storing the
+PATH_CHALLENGE data well in excess of any plausible packet round-trip time on
+a new path, as a late response would result in connection teardown.
```
martinthomson: I don't think that we need to say this either, because if the endpoint isn't sure that it sent the challenge, then permission to end the connection isn't implied.

MikeBishop:
#1265 would subsume this, if it goes to the extent Patrick wants.  And perhaps this PR should address that issue.  But assuming we're not actually removing the error entirely....

Just because we've agreed we're not going to go out of our way to make the handshake tamper-proof, that doesn't mean we have to ignore low-hanging fruit.  This is goodness, and the cost is effectively zero.

martinduke:
@martinthomson makes some good points about the handshake. Has anyone done an audit of possible on-path and off-path attacks on the handshake? If a bogus PATH_RESPONSE is neither cheaper nor more effective than other handshake attacks, then I agree there's no point in worrying about it.

As for the discarded challenge issue, by definition a "should" isn't something we need to say. I'll leave it to the editors as to whether it's worthwhile or not.

ianswett:
@bemasc Has spent quite a bit of time thinking about on and off-path attacks on the handshake.

bemasc:
> Bogus PATH_RESPONSE in a Handshake packet seems like an obvious DoS attack.

@martinduke Alas, the QUIC handshake is full of such attacks.  The specification generally requires or recommends that clients terminate the handshake if the remote party appears to be misbehaving, and an on-path attacker can easily trigger that termination behavior in many different ways.

I'm very interested in resolving this problem, but I think it's too early be trying to fix this in the QUIC specification, which is still very much in flux.  Once the QUIC protocol is finalized, perhaps we can have a discussion about what it would take to harden the handshake.

martinduke:
Thanks @bemasc. Is there an open issue for this assessment? If so, I am happy to close this PR.

martinthomson:
I will close this for now.  We currently have consensus not to harden the handshake against disruption by an on-path attacker.  I think that there is plenty of interest in looking at hardening the handshake against tampering and disruption, but those efforts aren't currently in scope.  Any attempt to do that will likely need to be marked `quicv2`, as Ben notes.

PR #1269, opened 2018-04-06
Change the UDP port sometimes too

martinthomson:
As discussed on-list, changing the UDP port at the same time as changing the connection ID makes good sense, if only to ensure that migration is exercised.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1770,7 +1770,9 @@ source UDP port when sending traffic after a period of inactivity.  Changing the
 UDP port from which it sends packets at the same time might cause the packet to
 appear as a connection migration. This ensures that the mechanisms that support
 migration are exercised even for clients that don't experience NAT rebindings or
-genuine migrations.
+genuine migrations.  Changing port number can cause a peer to reset its
+congestion state (see {{migration-cc}}), so this action SHOULD only be changed
+infrequently.
 
 An endpoint that receives a successfully authenticated packet with a previously
 unused connection ID MUST use the next available connection ID for any packets
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1763,9 +1763,14 @@ gap than it advertised.
 
 Clients MAY change connection ID at any time based on implementation-specific
 concerns.  For example, after a period of network inactivity NAT rebinding might
-occur when the client begins sending data again. A client might wish to reduce
-linkability by employing a new connection ID when sending traffic after a period
-of inactivity.
+occur when the client begins sending data again.
+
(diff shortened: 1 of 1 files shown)


britram:
+1 to CID rotation being pointless if we don't do this. However, this seems dangerous if not done carefully, and it links CID rotation rate to the maximum usable UDP port consumption rate.

Note that QUIC presently doesn't have an advisory on-path state clearance signal (we discussed this as a signal bound to public reset in #20, but it morphed into Stateless Reset, which has completely different semantics), which means that there is neither a present nor yet a potential future method to reduce the NAT state load this will generate.

mikkelfj:
UDP could also have a recommended client port when used with client CID, then the port does not leak information.

mirjak:
If you are behind an NAT and change your source port, you might also get a new outer IP address which would lead to path validation and congestion control reset at the server. Nor sure if you want to risk that every 10ms.

draft-ietf-quic-transport.md
```diff
@@ -1763,9 +1763,14 @@ gap than it advertised.
 
 Clients MAY change connection ID at any time based on implementation-specific
 concerns.  For example, after a period of network inactivity NAT rebinding might
-occur when the client begins sending data again. A client might wish to reduce
-linkability by employing a new connection ID when sending traffic after a period
-of inactivity.
+occur when the client begins sending data again.
+
+A client might wish to reduce linkability by employing a new connection ID and
+source UDP port when sending traffic after a period of inactivity.  Changing the
```
ianswett: This may have a performance impact if the server interprets this is a full migration and not just a NAT rebind.  I think it's worth noting that as well as that this may cause the server to perform path validation before sending any significant data?

chris-wood:
@huitema @britram Out of curiosity, do either of you have data on middleboxes that may not handle port rotation well?

martinthomson:
I've added text about the consequences of port number change, and advised that it only be done infrequently. @chris-wood, see the list for discussion about port rotation: infrequent changes should be fine based on that.

draft-ietf-quic-transport.md
```diff
@@ -1763,9 +1763,16 @@ gap than it advertised.
...
+UDP port from which it sends packets at the same time might cause the packet to
+appear as a connection migration. This ensures that the mechanisms that support
+migration are exercised even for clients that don't experience NAT rebindings or
+genuine migrations.  Changing port number can cause a peer to reset its
+congestion state (see {{migration-cc}}), so this action SHOULD only be changed
```
ianswett: "so this action SHOULD only be changed infrequently." is not quite right.  Did you intend "so this action SHOULD only be taken infrequently."?  Or "the port SHOULD only be changed infrequently."?

PR #1273, opened 2018-04-08
tls: "Packet Number Gap" in transport has moved

reschke:


martinthomson:
Ideally this would have been **re**moved, but consensus is hard.

PR #1274, opened 2018-04-09
Permit CONNECTION_CLOSE in Handshake

### LINKED ISSUE #1270: CONNECTION_CLOSE missing from handshake packet
Section “5.4.3. Handshake Packet” states
>The payload of this packet contains STREAM frames and could contain PADDING, ACK, PATH_CHALLENGE, or PATH_RESPONSE frames.

CONNECTION_CLOSE is forgotten here (e.g., when TLS error occurs, when server is busy)

martinthomson:
Closes #1270.

@janaiyengar, anyone, can you remember why we allow PATH_* frames in Handshake messages?  They don't seem to serve any particular purpose.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -644,7 +644,8 @@ but could involve additional computational effort depending on implementation
 choices.
 
 The payload of this packet contains STREAM frames and could contain PADDING,
-ACK, PATH_CHALLENGE, or PATH_RESPONSE frames.
+ACK, PATH_CHALLENGE, or PATH_RESPONSE frames.  Handshake packets MAY contain
+CONNECTION_CLOSE frames if the handshake is unsuccessful.
 
 
 ## Protected Packets {#packet-protected}


nibanks:
If the client can acknowledge any of the server's Handshake packets, doesn't that prove ownership of the address, assuming the server's packet numbers are not predictable?

ianswett:
An acknowledgement only proves the peer saw the right packet number, which may be possible to guess, depending on what we decide with packet number randomization/etc.

nibanks:
I agree it can only work if the packet number isn't predictable, and that depends on the final scheme we decide to use for packet numbers. But if it isn't predictable, receiving a packet from an IP address that you sent the packet to, containing information that could only be retrieved from that original packet seems to me, to prove they can receive any packets sent to that IP address.

MikeBishop:
It's for the server's three-packet limit.  If the server's first flight is more than three packets, it includes a PATH_CHALLENGE in each of those three.  The client can send back a Handshake frame that contains only a PATH_RESPONSE, permitting the server to finish sending its first flight as soon as it gets any of them.

PR #1282, opened 2018-04-10
QUIC is QUIC, and does not stand for anything.

ariel-anieli:
https://www.ietf.org/mail-archive/web/quic/current/msg03855.html

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -137,8 +137,7 @@ protocol, including the conceptual design, wire format, and mechanisms of the
 QUIC protocol for connection establishment, stream multiplexing, stream and
 connection-level flow control, connection migration, and data reliability.
 
-QUIC used to be known as 'Quick UDP  Internet Connections', but the present
-document advises it be used no  more: QUIC does not stand for anything.
+QUIC is a name, not an acronym.
 
 Accompanying documents describe QUIC's loss detection and congestion control
 {{QUIC-RECOVERY}}, and the use of TLS 1.3 for key negotiation {{QUIC-TLS}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -137,6 +137,9 @@ protocol, including the conceptual design, wire format, and mechanisms of the
 QUIC protocol for connection establishment, stream multiplexing, stream and
 connection-level flow control, connection migration, and data reliability.
 
+QUIC used to be known as 'Quick UDP  Internet Connections', but the present
+document advises it be used no  more: QUIC does not stand for anything.
+
 Accompanying documents describe QUIC's loss detection and congestion control
```
mikkelfj: I belive the feedback on the mailing list was that the RFC is not the place to make historical references.

>QUIC used to be known as 'Quick UDP  Internet Connections', but the present
 +document advises it be used no  more: QUIC does not stand for anything.

could perhaps be reformulated as

> QUIC is a name and not an acronym.

ariel-anieli:
@mikkelfj, here is the new pull.

draft-ietf-quic-transport.md
ariel-anieli: Fixed in the new pull, @mikkelfj. Have you seen it?

mikkelfj: now I have :)

MikeBishop:
I'm a little dubious of throwing this sentence as a stand-alone paragraph in the Introduction, as it seems unrelated to most of the other text there.  Maybe it would fit better at the end of Section 2, when we're defining other terms and notations?

ariel-anieli:
Right, @MikeBishop. A new pull will change it.

martinthomson:
Thanks Mike, I had the same objection, but didn't think of the suggestion.

ariel-anieli:
Fixed, @MikeBishop.

martinthomson:
So few words, so much effort.  Thanks @ariel-anieli.

PR #1283, opened 2018-04-10
Spin Bit Reservation

britram:
As discussed in London and confirmed on the list, this PR reserves bit 0x04 of the first octet of the short header for the spin bit, referencing a [draft](https://britram.github.io/draft-trammell-quic-spin/draft-ietf-quic-spin-exp.html) to be submitted shortly for the details of the spin bit.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -89,6 +89,16 @@ informative:
     target:
      "https://web.archive.org/web/20150315054838/http://ha.ckers.org/slowloris/"
 
+  SPIN-BIT-EXP:
+    title: "The QUIC Latency Spin Bit"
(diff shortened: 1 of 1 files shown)


britram:
This PR supercedes #1046.

mnot:
Brian,

The result of the discussion in London was that the bits would be reserved for experimentation -- not specifically for the Spin Bit.

martinthomson:
That's OK, I fixed it.

ronieven:
My understanding of the meeting and the meeting notes is that the bit is reserved for spin bit experimentation since this is what the discussion was about and that there will be a reference to the spin-exp draft. There is no mention of general experiment usage in the meeting notes.

PR #1284, opened 2018-04-10
Grammar fix: contents "determine," not "determines."

dtikhonov:


dtikhonov:
I thought about "content determines," but then I decided that messages have more than one content.

martinthomson:
This is fine.

PR #1289, opened 2018-04-12
Removed contractions from text

ariel-anieli:
Fixed Point 3 of https://www.ietf.org/mail-archive/web/quic/current/msg03844.html


ariel-anieli:
Yes, @LPardue; I fixed it.

martinthomson:
I don't think that we need to do this.  These are common contractions that are well understood by readers of English.

PR #1291, opened 2018-04-13
Changelog for -11

martinthomson:
Also includes a script to help with tagging.  You shouldn't need to use that unless you are tagging docs that you aren't an author on.


LPardue:
I know this has been merged but.. 

Http doc missed out that frame diagrams were added

LPardue:
Ok!

PR #1299, opened 2018-04-17
Makefile: add target to generate tags for I-D sections

dtikhonov:


MikeBishop:
This seems like it might be a more generally useful addition to https://github.com/martinthomson/i-d-template, rather than being specific to QUIC.

dtikhonov:
Superceded by PR #1302 -- closing.

MikeBishop:
Exactly: my point is that you're making the PR into QUIC's local Makefile instead of the template's Makefile.  (Also, you can update PRs by pushing more commits to the same branch, even replacing commits with `--force` if needed.  Updates don't require a new PR.)

dtikhonov:
Well, if you guys don't want it, you can just close PR #1302 and write that you don't want it. I know *I* will still use tags. :)

PR #1300, opened 2018-04-17
QUIC-TRANSPORT: Add missing anchor to Packet Number Gap section

dtikhonov:


MikeBishop:
This may also be a version thing -- there's an implicit anchor on the name of the section (as here).

dtikhonov:
This is true -- there is no need for this change.  Thanks.

This also means I should improve the tag-generating code.

PR #1302, opened 2018-04-18
Generate tags file for explicit and implicit anchors in the drafts

dtikhonov:
This replaces PR #1299.


dtikhonov:
Closing -- will open a PR against "the template," as @MikeBishop recommends.

PR #1303, opened 2018-04-18
Remove term "application-owned streams"

### LINKED ISSUE #1297: Meaning of "application-owned streams" is unclear
This wording -- "application-owned stream" -- is only used to describe the QUIC transport parameters.  I tracked down the introduction of this concept to f6157a0b51f14ee1583df3cc7408d8ae86174180.

Based on the diff, I gather that this simply means "not the crypto stream," at least at the time of this writing. Nevertheless, I think that the concept of an application-owned stream should be defined before it is used.

LPardue:
Fixes #1297. 

Just a proposal, shoot it down...

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1219,9 +1219,9 @@ initial_max_stream_id_uni (0x0008):
   an unsigned 16-bit integer.  If this parameter is absent or zero,
   unidirectional streams cannot be created until a MAX_STREAM_ID frame is
   sent. Setting this parameter is equivalent to sending a MAX_STREAM_ID
-  ({{frame-max-stream-id}}) immediately after completing the handshake 
-  containing the corresponding Stream ID. For example, a value of 0x05 would be 
-  equivalent to receiving a MAX_STREAM_ID containing 18 when received by a 
+  ({{frame-max-stream-id}}) immediately after completing the handshake
+  containing the corresponding Stream ID. For example, a value of 0x05 would be
+  equivalent to receiving a MAX_STREAM_ID containing 18 when received by a
   client, or 19 when received by a server.
 
 max_packet_size (0x0005):
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1202,28 +1202,27 @@ An endpoint MAY use the following transport parameters:
 initial_max_streams_bidi (0x0002):
 
 : The initial maximum bidirectional streams parameter contains the initial
-  maximum number of application-owned bidirectional streams the peer may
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1202,28 +1202,27 @@ An endpoint MAY use the following transport parameters:
...
-  MAX_STREAM_ID frame is sent.  Note that a value of 0 does not prevent the
-  cryptographic handshake stream (that is, stream 0) from being used. Setting
-  this parameter is equivalent to sending a MAX_STREAM_ID
-  ({{frame-max-stream-id}}) immediately after completing the handshake
-  containing the corresponding Stream ID. For example, a value of 0x05 would be
-  equivalent to receiving a MAX_STREAM_ID containing 20 when received by a
-  client or 17 when received by a server.
+  maximum number of bidirectional streams the peer may initiate, encoded as
+  an unsigned 16-bit integer.  Note that the cryptographic handshake stream
+  (that is, stream 0) is exempt, a value of zero does not prevent it from being
```
MikeBishop: The issue is that stream 0 is still (pending design team) a bidirectional stream initiated by the peer.  So a valid interpretation of this new text would be:

- 0:  Special case, can still use crypto stream
- 1:  One bidirectional stream permitted, which is the crypto stream
- 2:  Two bidirectional streams permitted, one of which is the crypto stream
...and so on.

The point of "application-owned" is not just that zero can't be blocked by this setting, but to clarify that the number available to the application isn't one less than this value due to 0.

MikeBishop:
Oh, and your build failures are from trailing whitespace on lines 1222-1224.

LPardue:
The intent of the change is that stream 0 is (right now) always exempt from calculation. I.e it is reserved and should not be included in the conversion from number of streams to stream ID. The conversion is non-intuitive to the unfamiliar reader. So, I think we agree in principle but disagree on phrasing.

More broadly, if the term application-owned stream is not used anywhere else, I see little value in coining it here. It makes assumption that all non 0 streams are application owned. Why may be the a different topic of discussion, hence the tussle here.

MikeBishop:
I agree, we're trying to say the same thing, that zero doesn't count for purposes of this setting.  But still, saying nothing about it except the example doesn't fix the problem, because instead of being confused a reader might reach the wrong conclusion and keep going.

Perhaps "maximum number of bidirectional streams the peer may initiate **in addition to stream 0**..."?

martinthomson:
This is going to be exploded by the stream 0 design team effort, so we'll close it now.

MikeBishop:
Obsoleted by the design team's proposal (all streams are application-owned) -- we'll leave the issue and adjust the wording when that work lands.

LPardue:
Sgtm and saves some work!

PR #1304, opened 2018-04-19
Move Retry to QUIC Layer

nibanks:
Move the stateless retry method of client source address validation out of TLS and into the QUIC layer.


nibanks:
Oops. Didn't mean to create this. Sorry for the noise.

Issue #1105, opened 2018-02-10
Segmented Packet Numbers


mikkelfj:
There is a long ongoing discussion on the mailing list about packet number encryption.
Here is proposal extracted from that discussion which does not involve packet number encryption
(as proposed here https://github.com/quicwg/base-drafts/pull/1079).

NOTE: this is not a consensus and I am not personally convinced this is the best option, but it is one option.

The packet number is segmented in a lower 32-bit counter (nonce) and an upper 30+bit path identifier (PID). The packet number is globally unique across all paths and is used directly in ACK frames, but it is never transmitted in clear text on the wire. The lower 32-bit nonce is transmitted in clear text in place of the packet number as a varint. The nonce starts at zero and increments for every packet sent on a given path and a path cease to be valid no later than when the counter reaches 2^32-1. The PID starts at 0 and is incremented for every new path in use. It is not strictly required to increment by one, but each path MUST have a unique PID that is known to both endpoints. There must be a mapping from connection ID to a PID pair.

This solution provides dense ACK ranges only having gaps where a new path is in use. The packet number directly identifies that path the packet was transmitted on which could help congestion control.

It may be relevant to allow more than 30 bits for the PID range because the packet number space is exhausted relatively fast, but then the packet number cannot be encoded in a 64 bit varint.

It is also proposed to use asymmetric CID's so there is a 1-1 mapping from PID to CID, but it is not strictly necessary (https://github.com/quicwg/base-drafts/issues/1089).

Additionally, if each path is given two PID's differing only by only the lower bit, this can be used instead of skipping packet numbers in gapless ACK attack mitigation (https://github.com/quicwg/base-drafts/issues/1092). If the CID is absent, a bit in the header can select which PID is in use. 
(EDIT: I think the above won't work because it is easy to guess the next nonce anyway, but if you randomly switch and stick to either PID for a while, it could work).

The major upside to the approach:

1. Avoids PN encryption which could add an estimated 18% overhead for low-latencey applications.
(https://www.ietf.org/mail-archive/web/quic/current/msg03292.html)
2. AEAD IV (nonce/packetnumber) not in clear text affects hardware offload and pipelined operation.
3. Reduces ACK fragmentation.
4. Simplifies per path network monitoring
5. Easily handles multi-path.
6. Congenstion control might be easier to handle across paths because the PN identifies its path directly and loss estimation and ACK history truncation can be done per path.
7. Nonce prevents linkage across paths.

The major downside to the approach:

[...]

PR #1307, opened 2018-04-19
Change Packet Number Gap to Packet Number Offset

nibanks:
This is a proposed solution to the privacy/linkability problem by slightly modifying the existing packet number gap solution to make it a per connection ID packet number offset. The offset is used to transform the wire encoding of the packet number, but the connection still uses the same monotonically increasing packet number space (no gaps).

This solution requires very little per packet CPU overhead, compared to PNE.

Unlike the PNE proposal (#1079), this doesn't fix the ossification/greasing problem as well. Instead, I'd like to treat that as a separate problem (and a separate PR) and attempt to solve that with a non-cryptographic solution (shuffle for instance).

Closes:
- #1174
- #1034
- #990

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -832,6 +832,27 @@ Use of a secure random number generator {{?RFC4086}} is not necessary for
 generating the initial packet number, nor is it necessary that the value be
 uniformly distributed.
 
+### New Connection ID Packet Numbers
+
+Packets with with new connection IDs have special rules for encoding the packet
+number in the headers.  They use an offset (see {{packet-number-offset}}) to
+transform the packet number via the following:
+
+~~~
+transformed_packet_number = (packet_number + offset) % 2^62
+~~~
+
+Once transformed, the normal rules of packet number encoding are followed, only
+encoding the least number of bits required.
+
+When receiving a packet with a new connection ID, the full transformed packet
+number must be first computed. It is decoded similarly to how a normal packet
+number would be decoded, but instead of comparing the number with the largest
+acknowledged packet number, it compares it to the largest packet number
+transformed with the same equation above. Once the full transformed packet
+number is decoded, the packet number offset can be subtracted to get the true
+packet number.
+
 
 # Frames and Frame Types {#frames}
 
@@ -1784,7 +1805,7 @@ break linkability between two points of network attachment.
 
 An endpoint might need to send packets on multiple networks without receiving
 any response from its peer.  To ensure that the endpoint is not linkable
-across each of these changes, a new connection ID and packet number gap are
+across each of these changes, a new connection ID and packet number offset are
 needed for each network.  To support this, each endpoint sends multiple
 NEW_CONNECTION_ID messages.  Each NEW_CONNECTION_ID is marked with a sequence
 number.  Connection IDs MUST be used in the order in which they are numbered.
@@ -1796,15 +1817,11 @@ new connection IDs using the NEW_CONNECTION_ID frame.
 An endpoint which wishes to break linkability upon changing networks MUST use
 the connection ID provided by its peer as well as incrementing the packet
 sequence number by an externally unpredictable value computed as described in
-{{packet-number-gap}}. Packet number gaps are cumulative.  An endpoint might
-skip connection IDs, but it MUST ensure that it applies the associated packet
-number gaps for connection IDs that it skips in addition to the packet number
-gap associated with the connection ID that it does use.
+{{packet-number-offset}}. A packet number offset is computed per connection ID,
+and is independent of any other packet number offset.
 
 An endpoint that receives a packet that is marked with a new connection ID
-recovers the packet number by adding the cumulative packet number gap to its
-expected packet number.  An endpoint MUST discard packets that contain a smaller
-gap than it advertised.
+recovers the packet number by subtracting the packet number offset.
 
 Clients MAY change connection ID at any time based on implementation-specific
 concerns.  For example, after a period of network inactivity NAT rebinding might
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -832,6 +832,27 @@ Use of a secure random number generator {{?RFC4086}} is not necessary for
 generating the initial packet number, nor is it necessary that the value be
 uniformly distributed.
 
+### New Connection ID Packet Numbers
+
+Packets with with new connection IDs have special rules for encoding the packet
+number in the headers.  They use an offset (see {{packet-number-offset}}) to
+transform the packet number via the following:
```
marten-seemann: Why does this only apply to new connection IDs? Wouldn't it be easier to apply this transformation to every packet number?

huitema:
I think that's a terrible idea. It perpetuates the fallacy that offsetting packet numbers provides privacy. We have established that it doesn't in an important case, when there is overlap between two connections. The matching pattern of holes and sequence number increment is more that sufficient to establish correlation in these cases.

We would be much better off just getting rid of the gaps altogether. They force complex synchronization between sender and receiver, effectively the same management cost as managing encryption keys, but they do not provide nearly as much benefit as encryption.

mikkelfj:
@nibanks this looks a bit like my segmented packet numbers proposal but not that this proposal has a fixed offset per connection ID and that a connection ID is replaced before that offset is exhausted.

This avoids any linkage other than NAT rebinding.

https://github.com/quicwg/base-drafts/issues/1105

You can further mutilate the clean packet number sequence with a reversible permutation

http://stackoverflow.com/a/12996028
```
   uint32_t x;

    x = ((uint32_t)(size_t)key) ^ (uint32_t)(MY_SEED);

    x = ((x >> 16) ^ x) * 0x45d9f3bUL;
    x = ((x >> 16) ^ x) * 0x45d9f3bUL;
    x = ((x >> 16) ^ x);
    return x;
```

And you can further encrypt that permuation using simple XOR magic - apparently IPSec style.

mikkelfj:
Just using a random seed might be sufficient to deter the average middle box and with CID update there is no privacy leak.
The resulting PN can be used as AEAD IV since it is a permutation of a unique non-repeating sequence. After CID update the AEAD keys must change, but then it is about time anyway.

mikkelfj:
I suppose a more elaborate encryption would go like

```
uint32_t key[4];
hkdf(key, context);

uint32_t x;

    x = ((uint32_t)(size_t)PN) ^ key[0];
    x = ((x >> 16) ^ x) * UINT32_C(0x45d9f3b) ^ key[1];
    x = ((x >> 16) ^ x) * UINT32_C(0x45d9f3b) ^ key[2] ;
    x = ((x >> 16) ^ x) * UINT32_C(0x45d9f3b) ^ key[3];
    x = ((x >> 16) ^ x);
    return x;
```
Due to brute force the encryption is not strong, but combined with per path CID and update before 2^32 PN's it should be robust since the PN encryption is not privacy.

To speed things up, one or two multiplication rounds can be taken out.

See also
 http://stackoverflow.com/a/12996028

mikkelfj:
The above cannot be used directly as IV because the permuation is no longer a 1 to 1 mapping but a key guided mapping, so the encrypted PN will not work as an IV. The same issue should not be the case with the initial seeded value.

draft-ietf-quic-transport.md
MikeBishop: The gap in the actual packet number was introduced at the point you sent the first packet with a new CID.  This proposal is a transformation of the actual packet number for wire presentation, so it needs to occur every time.

draft-ietf-quic-transport.md
```diff
@@ -832,6 +832,27 @@ Use of a secure random number generator {{?RFC4086}} is not necessary for
...
+
+~~~
+transformed_packet_number = (packet_number + offset) % 2^62
```
MikeBishop: This might actually be okay, but the possibility that the transformed packet number might wrap around to zero before being truncated feels a little odd.

draft-ietf-quic-transport.md
```diff
@@ -832,6 +832,27 @@ Use of a secure random number generator {{?RFC4086}} is not necessary for
...
+~~~
+
+Once transformed, the normal rules of packet number encoding are followed, only
+encoding the least number of bits required.
```
MikeBishop: "only" and "the least" feels duplicative.  Maybe just replace the part with the comma with a reference to the section where the truncation is defined?

Better yet, since this is effectively a transform on every packet number, put this in the section about truncation (it becomes a description of how the 64-bit number is encoded on the wire) and make this section purely about determining what the offset will be.

mikkelfj:
> if you switched keys because you used a new CID to probe, what key do you use back on the original path while you wait for the probe to complete?

I forgot about my own proposal: No, rekeying is not needed for each new CID. The segmented packet number idea means that each CID is given an internal incremental high bit value. So the IV is the CID counter + the visible PN, permutated or not. Hence the AEAD can last much longer than CID's

However, for the sake of privacy, you need to do a a high bit increment for each CID, but that is also very cheap.

huitema:
This design falls in the category of "obfuscation", analyzed in https://github.com/quicwg/base-drafts/wiki/Summary-of-the-PN-encryption-issues-and-alternatives#alternative-pn-encryption. It does not provide privacy, and also does not protect against ossification. You would be better off leaving the PN in clear text.

mikkelfj:
> You would be better off leaving the PN in clear text.
That was also my original intend, but it does offer privacy because it is unique per CID and there is no visible gap to be compared across CID's (modulo NAT rebinding). As to obfuscation, if seeded with an unguessable value, and filtered for the zero case, it would not be unbreakable, but neither would it be very trivial to make use of in a busy switch. So it comes down to how much you wan't to harden against ossification - if the answer is "very" then full PNE is the only way.

ad-l:
@MikeBishop is right - the proposal only makes sense if it applies to all packets, so the text about this being only for new connection IDs must be removed

draft-ietf-quic-transport.md
ad-l: I think that's a clear mistake, the transformation does apply to all packet numbers on the network

nibanks:
No longer necessary as PNE was merged.

PR #1312, opened 2018-04-20
Deemphasize TCP SACKs reference in ACK frame description

dtikhonov:
It is odd to describe something by saying what is *not* like.  In
addition, not all readers of the document will be familiar with
TCP SACKs.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2581,9 +2581,9 @@ received and processed.  A sent packet that has never been acknowledged is
 missing. The ACK frame contains any number of ACK blocks.  ACK blocks are
 ranges of acknowledged packets.
 
-QUIC acknowledgements are irrevocable:  Once a packet has been acknowledged,
-even if it does not appear in a future ACK frame, it remains acknowledged.
-(This is unlike TCP SACKs.)
+QUIC acknowledgements are irrevocable.  Once acknowledged, a packet remains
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2581,9 +2581,9 @@ received and processed.  A sent packet that has never been acknowledged is
 missing. The ACK frame contains any number of ACK blocks.  ACK blocks are
 ranges of acknowledged packets.
 
-Unlike TCP SACKs, QUIC acknowledgements are irrevocable.  Once a packet has
-been acknowledged, even if it does not appear in a future ACK frame,
-it remains acknowledged.
+QUIC acknowledgements are irrevocable:  Once a packet has been acknowledged,
```
martinthomson: Period, not colon.

draft-ietf-quic-transport.md
```diff
@@ -2581,9 +2581,9 @@ received and processed.  A sent packet that has never been acknowledged is
...
+even if it does not appear in a future ACK frame, it remains acknowledged.
+(This is unlike TCP SACKs.)
```
martinthomson: Don't include an entire sentence in a parenthetical.  It would be fine to remove the paretheses here.

Of course, now that you touched this, I notice that we don't have a TCP SACK reference here, and you can't just use an unexpanded acronym like this without both expansion and a reference.

janaiyengar:
I think this addresses Martin's comments, so I'll merge.

PR #1313, opened 2018-04-20
Remove awkward sentence

### LINKED ISSUE #1311: Awkward phrase in ACK Frame section of the Transport Draft
In the Transport Draft 11, the [ACK Frame section (7.15)](https://tools.ietf.org/html/draft-ietf-quic-transport-11#section-7.15) begins as follows:

```
   Receivers send ACK frames (type=0x0d) to inform senders which packets
   they have received and processed.  A sent packet that has never been
   acknowledged is missing.  The ACK frame contains any number of ACK
   blocks.  ACK blocks are ranges of acknowledged packets.
```

What is the purpose of the second sentence, both in the context of the ACK Frame section and outside thereof?

I think it should be removed.  Or at least rephrased.

ianswett:
Fixes #1311

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2577,9 +2577,8 @@ Application Error Code:
 ## ACK Frame {#frame-ack}
 
 Receivers send ACK frames (type=0x0d) to inform senders which packets they have
-received and processed.  A sent packet that has never been acknowledged is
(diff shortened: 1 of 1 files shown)


martinthomson:
This is fine, but you should probably let @janaiyengar or myself review or merge this sort of change in future.

PR #1314, opened 2018-04-20
Better Indexing in QPACK

### LINKED ISSUE #1125: Remove BLOCKING flag and Depends
I can find no good use for the "Depends" field in header blocks.  Base is sufficient.

### LINKED ISSUE #1130: Base and the Static Table
It is not explained at all that the base is applied to the dynamic table only.

That is, index 1 is a straight index into the static table and that index 62 will use the current number of inserts and the base to determine the actual index into the dynamic table (this is the first table entry only when the number of table inserts and base are the same, if the base is smaller, then it will be further along the table).

To be clear, my logic for finding entries is (absent some other bounds-checking, that real code cannot safely omit):

```python
    if i <= 0: error
    if i <= len(staticTable): return staticTable[i-1]
    delta = tableBase - base
    if delta < 0: error
    return dynamicTable[i - len(staticTable) - 1 + delta]
```

It is probably equally valid to assume that the indices for the static table also increment with each iteration of base, but that leads to approximately the same code anyway.

### LINKED ISSUE #1136: Base/depends of 0 and dynamic table entries
We should explicitly note that a base/depends of 0 is the best choice for header blocks that don't reference the dynamic table.  Setting this means that it is impossible to reference any dynamic entries.

### LINKED ISSUE #1145: 0-based indexing
QPACK (?) uses the same 1-based indexing that HPACK does.  But if we contemplate #1144, then all that 1-based indexing does is create an unused value.  That's annoying.  0-based indexes are more common and are natural.  We tolerated 1-based indexing in HPACK because it make name indexes somewhat easier to manage.

### LINKED ISSUE #1176: When are missing Q-thingey references fatal?
@afrind raises a good hypothetical case:

- Encoder sends a header block on a request stream, which is delayed in the network
- Encoder side resets the stream, which is lost in the network
- Encoder sends another header block on the control stream that evicts an entry referenced in the first block, and this is delivered immediately.
- The first delayed header block arrives

From the encoder's side, it's valid to evict something referenced by the first block, because the stream has been reset.  From the decoder's side, there's a reference to something already evicted, because it doesn't know about the stream reset.

The solution to this seems to be that missing references on request/push streams are stream errors, not connection errors.  Missing references on the control stream still need to be connection errors.

MikeBishop:
Based on #1141.

Adds a section which more clearly (I hope!) explaining how indices work in different contexts.  Cleans up a whole mess of index-related issues in the process.

Fixes #1176
Fixes #1145
Fixes #1136
Fixes #1130
Fixes #1125

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -185,21 +185,21 @@ d = count of entries dropped
 {: title="Dynamic Table Indexing - Control Stream"}
 
 Because frames from request streams can be delivered out of order with
-instructions on the control stream, relative indices are based on the Base Index
-(see {{absolute-index}}) included in the HEADERS or PUSH_PROMISE frame.  The
-Base Index is the absolute index of the entry which has the relative index of
-zero when interpreting the frame.  The relative indices of entries do not change
+instructions on the control stream, relative indices are relative to the Base
+Index at the beginning of the header block (see {{absolute-index}}). The Base
+Index is the absolute index of the entry which has the relative index of zero
+when interpreting the frame.  The relative indices of entries do not change
 while interpreting headers on a request or push stream.
 
 ~~~~~ drawing
+             Base Index
+                 |
+                 V
     +---+-----+-----+-----+-------+
     | n | n-1 | n-2 | ... |  d+1  |  Absolute Index
     +---+-----+  -  +-----+   -   +
               |  0  | ... | n-d-3 |  Relative Index
               +-----+-----+-------+
-                 ^
-                 |
-             Base Index
 
 n = count of entries inserted
 d = count of entries dropped
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -204,7 +204,7 @@ while interpreting headers on a request or push stream.
 n = count of entries inserted
 d = count of entries dropped
 ~~~~~
-{: title="Dynamic Table Indexing - Request Stream"}
+{: title="Example Dynamic Table Indexing - Request Stream"}
 
 Entries with an absolute index greater than a frame's Base Index cannot be
 referenced by that frame.  If the decoder encounters a reference to an entry
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -209,8 +209,8 @@ d = count of entries dropped
 Entries with an absolute index greater than a frame's Base Index cannot be
 referenced by that frame.  If the decoder encounters a reference to an entry
 which has already been dropped from the table, this MUST be treated as a stream
-error of type `HTTP_QPACK_DECOMPRESSION_FAILED` error code.  If this reference
-occurs on the control stream, this MUST be treated as a session error.
+error of type `HTTP_QPACK_DECOMPRESSION_FAILED`.  If this reference occurs on
+the control stream, this MUST be treated as a session error.
 
 
 ## HEADERS Frames on the Control Stream
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -102,30 +102,16 @@ the reference from being usable.
 
 The encoder can choose on a per-header-block basis whether to favor higher
 compression ratio (by permitting vulnerable references) or HoL resilience (by
-avoiding them). This is signaled by the BLOCKING flag in HEADERS and
-PUSH_PROMISE frames (see {{QUIC-HTTP}}).
+avoiding them).
 
-If a header block contains no vulnerable header fields, BLOCKING MUST be 0.
-This implies that the header fields are represented either as references
-to dynamic table entries which are known to have been received, or as
-Literal header fields (see Section 6.2 of {{RFC7541}}).
-
-If a header block contains any header field which references dynamic table
-state which the peer might not have received yet, the BLOCKING flag MUST be
-set.  If the peer does not yet have the appropriate state, such blocks
-might not be processed on arrival.
-
-The header block contains a prefix ({{absolute-index}}). This prefix contains
-table offset information that establishes total ordering among all headers,
-regardless of reordering in the transport (see {{overview-absolute}}).
-
-In blocking mode, the prefix additionally identifies the minimum state required
-to process any vulnerable references in the header block (see `Depends Index` in
-{{overview-absolute}}).  The decoder keeps track of which entries have been
-added to its dynamic table.  The stream for a header with BLOCKING flag set is
-considered blocked by the decoder and can not be processed until all entries in
-the range `[1, Depends Index]` have been added.  While blocked, header
-field data MUST remain in the blocked stream's flow control window.
+The header block contains a Base Index (see {{absolute-index}}) which is used to
+correctly index entries, regardless of reordering in the transport (see
+{{indexing}}).  The stream for a header is considered blocked by the decoder and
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -164,6 +150,69 @@ unmodified.
...
+
+n = count of entries inserted
+d = count of entries dropped
+~~~~~
+{: title="Dynamic Table Indexing - Request Stream"}
+
+Entries with an absolute index greater than a frame's Base Index cannot be
+referenced by that frame.  If the decoder encounters a reference to an entry
+which has already been dropped from the table, this MUST be treated as a stream
+error of type `HTTP_QPACK_DECOMPRESSION_FAILED` error code.  If this reference
```
LPardue: > this MUST be treated as a stream error of type `HTTP_QPACK_DECOMPRESSION_FAILED` error code

seems a bit clunky

MikeBishop: Yeah, that's the discussion in #1176.  It's clunky, but what's better?  Otherwise, a delayed packet can cause your entire compression context to blow up.  The other alternatives discussed require major changes to how the HTTP layer uses QUIC that I'm not comfortable with.

I did have one other idea, which is to add a second acknowledgment type to QPACK, and require it to be sent if a stream was aborted without reading to the end.  Then the encoder has to consider even reset streams to still be outstanding until it receives something from the other side.

LPardue: Sorry, simpler issue, I just meant the language was clunky, "stream error of type X error code". Reducing to using type or error code would work I think

LPardue:
Lgtm

draft-ietf-quic-qpack.md
```diff
@@ -164,6 +150,69 @@ unmodified.
...
+a HEADERS frame as new entries are inserted.
+
+~~~~~ drawing
+    +---+---------------+-------+
+    | n |      ...      | d + 1 |  Absolute Index
+    + - +---------------+   -   +
+    | 0 |      ...      | n-d-1 |  Relative Index
+    +---+---------------+-------+
+      ^                     |
+      |                     V
```
dtikhonov: While RFC7541 has Insertion Point on the left and Dropping Point on the right, I think it would be more intuitive if the Absolute Index grew left-to-right.

draft-ietf-quic-qpack.md
```diff
@@ -164,6 +150,69 @@ unmodified.
...
+Because frames from request streams can be delivered out of order with
+instructions on the control stream, relative indices are based on the Base Index
+(see {{absolute-index}}) included in the HEADERS or PUSH_PROMISE frame.  The
+Base Index is the absolute index of the entry which has the relative index of
+zero when interpreting the frame.  The relative indices of entries do not change
+while interpreting headers on a request or push stream.
+
+~~~~~ drawing
+    +---+-----+-----+-----+-------+
+    | n | n-2 | n-3 | ... |  d+1  |  Absolute Index
```
dtikhonov: I think you missed `n-1`?

draft-ietf-quic-qpack.md
MikeBishop: Whatever we do, either absolute or relative will be starting at the greatest number and counting down.  Given that relative is what's sent in instructions, I found that to be more intuitive, personally.  I'm open to swapping them if most people prefer, though.

draft-ietf-quic-qpack.md
MikeBishop: Good catch!

draft-ietf-quic-qpack.md
```diff
@@ -164,6 +150,69 @@ unmodified.
...
+
+~~~~~ drawing
+    +---+-----+-----+-----+-------+
+    | n | n-1 | n-2 | ... |  d+1  |  Absolute Index
+    +---+-----+  -  +-----+   -   +
+              |  0  | ... | n-d-3 |  Relative Index
+              +-----+-----+-------+
+                 ^
+                 |
+             Base Index
```
afrind: Since Base Index is an absolute index, would it be more intuitive to have this on the top of the drawing pointing down?

draft-ietf-quic-qpack.md
```diff
@@ -164,6 +150,69 @@ unmodified.
...
+      |                     V
+Insertion Point         Dropping Point
+
+n = count of entries inserted
+d = count of entries dropped
+~~~~~
+{: title="Dynamic Table Indexing - Control Stream"}
+
+Because frames from request streams can be delivered out of order with
+instructions on the control stream, relative indices are based on the Base Index
```
afrind: "...relative indices are relative to the Base Index..." perhaps?

draft-ietf-quic-qpack.md
```diff
@@ -164,6 +150,69 @@ unmodified.
...
+              |  0  | ... | n-d-3 |  Relative Index
+              +-----+-----+-------+
+                 ^
+                 |
+             Base Index
+
+n = count of entries inserted
+d = count of entries dropped
+~~~~~
+{: title="Dynamic Table Indexing - Request Stream"}
```
afrind: Include the word 'example' in the caption?

draft-ietf-quic-qpack.md
```diff
@@ -225,9 +273,8 @@ represented as an 8-bit prefix string literal.
 ### Duplicate {#indexed-duplicate}
 
 Duplication of an existing entry in the dynamic table starts with the '000'
-three-bit pattern.  The index of the existing entry is represented as an integer
-with a 5-bit prefix. Table indices are always non-zero; a table index of zero
-MUST be treated as a decoding error.
+three-bit pattern.  The absolute index of the existing entry is represented as
```
afrind: Why absolute?  Because it's ordered on the control stream, this can be relative to N=number of inserts.

draft-ietf-quic-qpack.md
```diff
@@ -164,6 +150,69 @@ unmodified.
...
+Insertion Point         Dropping Point
+
+n = count of entries inserted
+d = count of entries dropped
+~~~~~
+{: title="Dynamic Table Indexing - Control Stream"}
+
+Because frames from request streams can be delivered out of order with
+instructions on the control stream, relative indices are based on the Base Index
+(see {{absolute-index}}) included in the HEADERS or PUSH_PROMISE frame.  The
```
afrind: You mention Base Index here like it's a field in the HEADERS or PUSH_PROMISE frame, but it looks more like the beginning of the header block below.

draft-ietf-quic-qpack.md
afrind: I have a slight preference for absolute index growing left-to-right, but I can live with it as is too.

draft-ietf-quic-qpack.md
MikeBishop: #1128.  You're almost certainly going to duplicate entries which have the largest possible relative index in your table.  Unless your table has wrapped multiple times, that means they have a relatively *small* absolute index.  On the flip side, the size of this will grow continuously over the lifetime of a connection while the relative index is at least bounded.

For ideal compression, this would be relative to the total number of drops (so the last item in the table is 0, etc.), but I was really hesitant to introduce a third (fourth, given #1239) way of indexing entries in the same table.

MikeBishop: Okay, you've convinced me.  😀

PR #1316, opened 2018-04-23
remove incorrect statement about the connection ID in server packet

marten-seemann:
The randomized Destination Connection ID is contained in Retry as well as in Handshake packets sent by the server. Or am I missing something obvious here?


MikeBishop:
Those packets from the server contain the CID that the server wants the client to start using.  See the next couple paragraphs of that section and 4.7.  The server can always choose the client's randomly selected one if it just needs to disambiguate, but the server never blindly echos the client's randomly-selected ID.

ianswett:
Good point Mike.  That being said, I find the way it's currently stated a bit confusing.

MikeBishop:
Perhaps the next paragraph needs a lead-in explaining why the SCID in the server's packet might be different than the DCID the client used.  It currently describes the mechanics -- when the client should pick up a new value to use in future packets.  It doesn't link that to a server's desire to select the CID to provide the context.

PR #1324, opened 2018-04-24
fix wording that said only long header packets could be coalesced

mcmanus:


draft-ietf-quic-transport.md
```diff
@@ -372,10 +372,10 @@ The interpretation of the fields and the payload are specific to a version and
 packet type.  Type-specific semantics for this version are described in the
 following sections.
 
-End of the Payload field (which is also the end of the long header packet) is
-determined by the value of the Payload Length field.  Senders can coalesce
-multiple long header packets into one UDP datagram.  See {{packet-coalesce}} for
-more details.
+End of the Payload field (which is also the end of the long header
```
MikeBishop: Not your text, but you touched it:  "**The** end"

PR #1325, opened 2018-04-24
Fix initial_max_streams transport param

martinduke:
This was renamed in one place in draft-11 (for bidi), but uni wasn't renamed. The change also missed numerous references in the document.


martinthomson:
OBE

PR #1327, opened 2018-04-25
Retransmit data, not packets

ianswett:


draft-ietf-quic-recovery.md
```diff
@@ -496,7 +496,7 @@ loss_detection_alarm:
 : Multi-modal alarm used for loss detection.
 
 handshake_count:
-: The number of times the handshake packets have been
+: The number of times all unacknowledged handshake data has been
```
siyengar: "The number of times all unacknowledged handshake data has been transmitted"

without receiving an ack seems redundant :)

ianswett: It's consistent with tlp_count and rto_count because all get reset to 0 when an ack is received.

MikeBishop: I think the point is that "unacknowledged ... without receiving an ack" is either redundant (if it's unacknowledged, of course you didn't receive an ack) or ambiguous (without receiving an ack of *that* data, or without receiving *any* acks for anything?) depending how you read it.

ianswett: It's without receiving an ack for any new packets.  Does that deserve clarification in all 3 spots?

PR #1345, opened 2018-05-10
Narrow lint target to just the drafts

martinthomson:
Note that this doesn't filter out non-markdown drafts, so this will fail in the general case, but I don't think that we'll have that problem here.

Fixes martinthomson/i-d-template#143 for @MikeBishop.  I hope.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ endif
 
 ifneq ($(PYTHON),)
 lint::
-	@$(PYTHON) ./.lint.py draft-*.md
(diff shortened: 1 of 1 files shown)


martinthomson:
I see.  I think that this is an easy fix, but we'll see.  We would still need this change, but also some more changes on the template.  In progress....

martinthomson:
@MikeBishop retry please.

MikeBishop:
Yep, works now.  Thank you!

PR #1346, opened 2018-05-10
Stateless reset fixes

### LINKED ISSUE #466: Stateless Reset from clients
A client may operate on equal footing with a server, i.e. it is not always a HTTP browser client, but currently a client cannot send a PUBLIC_RESET:

> A Public Reset packet is only sent by servers and is used to abruptly terminate communications. Public Reset is provided as an option of last resort for a server that does not have access to the state of a connection.

The problem is also that the purpose of PUBLIC_RESET is vague. The problem here is primarily that an endpoint might loose state during a crash and cannot tell peer to terminate the connection. The related use case for informing middle boxes is a separate topic.

Use case: Datacenters might have private networking, but not across datacenters. QUIC could be used to form a peer-to-peer server network. There might be edge routers in each datacenter (e.g. into a Kubernetes network), or not, but either way end-points may have high volume bidirectional traffic that needs to stop. The situation can also happen within datacenters with private networks with co-location where QUIC protects against other customers on the network.

Edge routers may have to also tear down the connection across datacenters (border gateway routing) which is then the separate middlebox issue.

Lost-state teardown need not involve any public activity and could avoid an obvious attack vector by using a non-public reset mechanism, but middlebox signalling need this. For the endpoint tear down, it could be replacing PUBLIC_RESET with a hint on lost connection which triggers surviving endpoint to perform a low volume liveness check, e.g. PING/ack at most every x ms.

Further consideration is that post-crash a server might have to load state from persistent storage in order to perform a reset proof that is not predictable. If this can be avoided via hint/live-ness verification, persistent state reset might be avoidable.

also discussed here:
https://github.com/quicwg/base-drafts/issues/462#issuecomment-297383356

martinthomson:
Inspired by @martinduke's #1328, I set to cleaning a few things up.  You can see separate commits here that:

* Take Martin's simplification suggestions
* Reword a few things based on review
* Make stateless reset symmetric
* Add a paragraph limiting the size of stateless reset (for amplification attack reasons)
* Add a point that @kazuho and I discussed somewhere about the scope of the stateless reset key and connection IDs (which opens a different DoS vector if they don't line up correctly).  That's partly addressed by the inclusion of an instance identifier in the calculation of the token, but the point is that you probably want instance X to be able to reset connections for instance Y if Y is out of commission, so you end up with the potential for DoS if you aren't careful to ensure that X and Y use different connection IDs.

Closes #466, #1328.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2082,23 +2082,24 @@ of the packet protection AEAD.  More random octets might be necessary if the
 endpoint could have negotiated a packet protection scheme with a larger minimum
 AEAD expansion.
 
-An endpoint cannot determine the source connection ID from a packet with a short
-header, therefore it cannot set the destination connection ID in the stateless
+An endpoint cannot determine the Source Connection ID from a packet with a short
+header, therefore it cannot set the Destination Connection ID in the stateless
 reset packet.  The destination connection ID will therefore differ from the
-value used in previous packets.  A random destination connection ID makes the
+value used in previous packets.  A random Destination Connection ID makes the
 connection ID appear to be the result of moving to new connection ID that was
 provided using the NEW_CONNECTION_ID frame ({{frame-new-connection-id}}).
 
 Using a randomized connection ID results in two problems:
 
 * The packet might not reach the peer.  If the Destination Connection ID is
-  critical for routing toward the client, then this packet could be incorrectly
+  critical for routing toward the peer, then this packet could be incorrectly
   routed.  This causes the stateless reset to be ineffective in causing errors
-  to be quickly detected and recovered.  In this case, clients will need to rely
-  on other methods - such as timers - to detect that the connection has failed.
+  to be quickly detected and recovered.  In this case, endpoints will need to
+  rely on other methods - such as timers - to detect that the connection has
+  failed.
 
 * The randomly generated connection ID can be used by entities other than the
-  client to identify this as a potential stateless reset.  A server that
+  peer to identify this as a potential stateless reset.  An endpoint that
   occasionally uses different connection IDs might introduce some uncertainty
   about this.
 
@@ -2109,9 +2110,9 @@ A stateless reset is not appropriate for signaling error conditions.  An
 endpoint that wishes to communicate a fatal connection error MUST use a
 CONNECTION_CLOSE or APPLICATION_CLOSE frame if it has sufficient state to do so.
 
-This stateless reset design is specific to QUIC version 1.  A server that
+This stateless reset design is specific to QUIC version 1.  An endpoint that
 supports multiple versions of QUIC needs to generate a stateless reset that will
-be accepted by clients that support any version that the server might support
+be accepted by peers that support any version that the endpoint might support
 (or might have supported prior to losing state).  Designers of new versions of
 QUIC need to be aware of this and either reuse this design, or use a portion of
 the packet other than the last 16 octets for carrying data.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2070,91 +2069,114 @@ following layout:
...
+for that connection.
+
+An endpoint that loses state can use the same method to generate a valid
+Stateless Reset Secret.  The connection ID comes from the packet that the
+endpoint receives.  An instance that receives a packet for another instance
+might recover the instance identifier from the connection ID.
+
+This design relies on the peer always sending a connection ID in its packets so
+that the endpoint can use the connection ID from a packet to reset the
+connection.  An endpoint that uses this design cannot allow its peers to use a
```
martinduke: "... cannot allow its peers to send packets with a zero-length destination connection ID"

martinthomson:
Thanks for the quick review.

Nothing specifically prevents an endpoints from sending stateless reset in response to a long header.  I've opened #1348 in case I missed something, because I have to move on to something else now :(.

draft-ietf-quic-transport.md
```diff
@@ -2160,8 +2160,8 @@ might recover the instance identifier from the connection ID.
 
 This design relies on the peer always sending a connection ID in its packets so
 that the endpoint can use the connection ID from a packet to reset the
-connection.  An endpoint that uses this design cannot allow its peers to use a
-zero-length connection ID.
+connection.  An endpoint that uses this design cannot allow its peers to send
+packets with a zero-length connection ID.
```
martinduke: "...zero-length destination connection ID"

draft-ietf-quic-transport.md
```diff
@@ -2070,91 +2069,114 @@ following layout:
...
+connection that is reset by revealing the Stateless Reset Token cannot be reused
+for new connections at the same instance without first changing to use a
+different static key or instance identifier.
+
+Any set of instances that share a static key for the purposes of stateless reset
+MUST allocate connection IDs from the same space without reusing connection IDs
+as long as the key is valid.  If instances that share a stateless reset key
+allow connections with the same connection ID to be created, then the stateless
+reset token for one connection could be used to terminate any connection that
+has the same connection ID.
```
MikeBishop: This looks like you're trying to close #1258 and #1259, too.  However, I don't think this language is strong enough.  It's not just that it MUST allocate from the same space without reusing; it MUST NOT generate a stateless reset for a connection ID that it would not have allocated, even if it has the proper key to do so.

draft-ietf-quic-transport.md
```diff
@@ -2070,91 +2069,114 @@ following layout:
...
 
 Revealing the Stateless Reset Token allows any entity to terminate the
 connection, so a value can only be used once.  This method for choosing the
-Stateless Reset Token means that the combination of server instance, connection
-ID, and static key cannot occur for another connection.  A connection ID from a
-connection that is reset by revealing the Stateless Reset Token cannot be
-reused for new connections at the same server without first changing to use a
-different static key or server identifier.
+Stateless Reset Token means that the combination of instance, connection ID, and
+static key cannot occur for another connection.  A connection ID from a
```
MikeBishop: By the same logic, I think the combination of connection ID and static key cannot be allowed to recur, even if it's on a different instance.

draft-ietf-quic-transport.md
MikeBishop: Actually, I think the "instance identifier" discussed above would mitigate this.  It will generate an SR, but it won't be valid from a different instance.

martinthomson:
Notes from editor's meeting: the last paragraph can be removed, then we will add another PR to address the routing issue underlying all this..

draft-ietf-quic-transport.md
```diff
@@ -2070,91 +2069,107 @@ following layout:
 This design ensures that a stateless reset packet is - to the extent possible -
 indistinguishable from a regular packet with a short header.
 
-A server generates a random 18-octet Destination Connection ID field.  For a
-client that depends on the server including a connection ID, this will mean that
-this value differs from previous packets.  Ths results in two problems:
+The message consists of a header octet, followed by random octets of arbitrary
+length, followed by a Stateless Reset Token.
+
+The endpoint SHOULD send a packet with a short header.
```
janaiyengar: This seems like a strange SHOULD. The Stateless Reset packet is defined here with a short header. How can this be a long header?

draft-ietf-quic-transport.md
```diff
@@ -2070,91 +2069,107 @@ following layout:
...
+
+Assuming a short header, the Random Octets field needs to include at least 20
```
janaiyengar: If the above comment holds, remove "Assuming a short header"

draft-ietf-quic-transport.md
```diff
@@ -2070,91 +2069,107 @@ following layout:
...
+An endpoint SHOULD NOT send a stateless reset that is significantly larger than
+the packet it receives.  Endpoints MUST discard packets that are too small to be
+valid QUIC packets.  With the set of AEAD functions defined in {{QUIC-TLS}},
+packets less than 19 octets long are never valid.
+
+An endpoint cannot determine the Source Connection ID from a packet with a short
+header, therefore it cannot set the Destination Connection ID in the stateless
+reset packet.  The destination connection ID will therefore differ from the
+value used in previous packets.  A random Destination Connection ID makes the
+connection ID appear to be the result of moving to new connection ID that was
```
janaiyengar: "moving to new connection ID" -> "moving to a new connection ID"

draft-ietf-quic-transport.md
```diff
@@ -2070,91 +2069,107 @@ following layout:
...
+provided using the NEW_CONNECTION_ID frame ({{frame-new-connection-id}}).
```
janaiyengar: s/the/a/

draft-ietf-quic-transport.md
```diff
@@ -2070,91 +2069,107 @@ following layout:
...
-either cannot be decrypted or is marked as a duplicate packet.  The client then
-compares the last 16 octets of the packet with the Stateless Reset Token
-provided by the server in its transport parameters.  If these values are
-identical, the client MUST enter the draining period and not send any further
-packets on this connection.  If the comparison fails, the packet can be
-discarded.
+An endpoint detects a potential stateless reset when a packet with a short
+header either cannot be decrypted or is marked as a duplicate packet.  The
+endpoint then compares the last 16 octets of the packet with the Stateless Reset
+Token provided by its peer, either from the NEW_CONNECTION_ID frame or the
```
janaiyengar: s/from/in/
s/the NEW_CONNECTION_ID frame/a NEW_CONNECTION_ID frame/
s/server/server's/

draft-ietf-quic-transport.md
```diff
@@ -2070,91 +2069,107 @@ following layout:
...
+the case where state is lost, so this approach is suboptimal.
 
 A single static key can be used across all connections to the same endpoint by
 generating the proof using a second iteration of a preimage-resistant function
-that takes three inputs: the static key, the server's connection ID (see
-{{connection-id}}), and an identifier for the server instance.  A server could
-use HMAC {{?RFC2104}} (for example, HMAC(static_key, server_id ||
+that takes three inputs: the static key, the connection ID chosen by the
+endpoint (see {{connection-id}}), and an instance identifier.  An endpoint could
+use HMAC {{?RFC2104}} (for example, HMAC(static_key, instance_id ||
```
janaiyengar: IIUC, incorporating an instance_id doesn't work for connections that are diverted to a new instance, which is a common case, possibly even the most common case. Server restarts are unlikely to be a common use case, since usually servers are drained before restarting, and server crashes are not frequent. Short-term routing flaps however are common.  I think the text should say that for Stateless Reset to work across instances, the HMAC could exclude the instance_id. (Correct me if my understanding is wrong.)

martinthomson: I've addressed that below.  But I've expanded on it a little in my latest changes.  The trick here is that you can recover the instance ID using the connection ID.  I will follow up with a change that includes the fixes we discussed.  Part of that change involves removing the instance ID from this.

MikeBishop: And of course, the issue is that if you enable servers to SR after a routing flap, you also enable those servers to be used as oracles.  The only defense is to make it impossible for the routing to be affected by the attacker.

janaiyengar: @martinthomson : Ah, right, I see it's addressed below. Given this is an example construction, I would suggest making that explicitly clear up at the beginning of this paragraph.
@MikeBishop : Yes, that is true, but defending the routing infrastructure against attacks seems out of scope for us, since similar attacks can be launched against TCP as well, where you'd receive an RST or ICMP unreachable.

MikeBishop: Defending it is out of scope.  Noting that you create an attack vector if you fail to defend it yourself, however, bears mentioning.

martinthomson:
See #1386 for the follow-up.

Pascalh2001:
Hi, 
About Stateless Reset, on one side you have : 
> A single static key can be used across all connections to the same endpoint by
 generating the proof using a second iteration of a preimage-resistant function
that takes three inputs: the static key, the connection ID chosen by the
endpoint (see {{connection-id}}), and an instance identifier.

But on the other side (section about Connection IDs) : 
> Short headers only include the Destination Connection ID and omit the explicit
length.  The length of the Destination Connection ID field is expected to be
known to endpoints.

So if the server loose state, how do you retrieve the Destination Connection Id from a short header ?
Is it mandatory for an endpoint to use the same length for all CIDs it uses and to be able to remind this length after a hard reboot ? 

('ve been reading your work and discussions since recently, great project!)

mikkelfj:
@Pascalh2001 you could imagine one server cluster settling on 16-bit identifiers throughout, and another cluster that uses a 4-bit type prefix from which the length can be derived, and a third cluster that register connection ID's in an in-memory database.

You save bandwidth by leaving it up to the endpoint rather than forcing the length to be transported at all times.

MikeBishop:
A cluster that wants the length present simply includes the length of the CID as part of the CID it tells the client to send it.  The encoding inside is totally up to the server, so that might be the explicit length, or a four-bit encoding of it, or even a single bit that toggles between the two lengths the cluster typically uses.

PR #1347, opened 2018-05-10
Write down spoofed packet guarantees

### LINKED ISSUE #645: Handshake guarantees/compromises
We need to decide what guarantees we will assume prior to establishing common keys.  This touches on things like

* whether we want to abort the connection in response to spoofed packets, or whether we can ignore them
* how the client does source address validation for the server
* whether the connection ID needs to be stable during the handshake or whether the 5-tuple is enough

This covers #624, #594, #554, #608, and likely other issues.

martinthomson:
We agreed to this a while back, but never got around to putting the conclusions in the document.  We've done some of the work by writing down the rules for use of CONNECTION_CLOSE and TLS alerts, and the related changes, but the basic principles never got articulated.

This attempts to outline what we do to authenticate packets that we receive.

* Once the handshake is done, it's mostly simple: if the AEAD fails, drop it.
* We have a small carve-out for some types of ICMP, with some strong limitations on what that information can be used for.  @igorlord might expand the PMTU usage to cover DestUnreach, but these uses permit a very narrow range of actions.
* We also have stateless reset, which is another snowflake.
* The bulk of the text here concerns the handshake.  I've covered the two mechanisms we tie into the use of connection IDs:
** Echoing the value (the only protection we have for Version Negotiation, for instance).
** The AEAD that uses the connection ID as input.

And then it explains that our only defense against an on-path attacker is to fail to establish a connection.  I think that covers all the necessary points.

Closes #645.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -4021,7 +4021,7 @@ transport to cancel a stream in response to receipt of a STOP_SENDING frame.
 
 As an encrypted and authenticated transport QUIC provides a range of protections
 against denial of service.  Once the cryptographic handshake is complete, QUIC
-endpoints discard packets that are not authenticated, greatly limiting the
+endpoints discard most packets that are not authenticated, greatly limiting the
 ability of an attacker to interfere with existing connections.
 
 Once a connection is established QUIC endpoints might accept some
@@ -4035,35 +4035,35 @@ attack from off the network path.  All QUIC packets contain proof that the
 recipient saw a preceding packet from its peer.
 
 The first mechanism used is the source and destination connection IDs, which are
-required to match those set by a peer.  Except for an Initial packet and
-stateless reset, an endpoint only accepts packets that include a destination
+required to match those set by a peer.  Except for an Initial and stateless
+reset packets, an endpoint only accepts packets that include a destination
 connection that matches a connection ID the endpoint previously chose.  This is
 the only protection offered for Version Negotiation packets.
 
 The destination connection ID in an Initial packet is selected by a client to be
 unpredictable, which serves an additional purpose.  The packets that carry the
-cryptographic handshake are protected with a key that is derived from the
+cryptographic handshake are protected with a key that is derived from this
 connection ID and salt specific to the QUIC version.  This allows endpoints to
 use the same process for authenticating packets that they receive as they use
 after the cryptographic handshake completes.  Packets that cannot be
 authenticated are discarded.  Protecting packets in this fashion provides a
-strong assurance that the sender of the packet saw previous packets and
-understood them.
+strong assurance that the sender of the packet saw the Initial packets and
+understood it.
 
 These protections are not intended to be effective against an attacker that is
-able to receive QUIC packets.  Such an attacker can potentially send packets
-that will be accepted by QUIC clients or servers.  This version of QUIC only
-attempts to detect this sort of attack.  Interference with the handshake is
-detected when the cryptographic handshake fails.  For the most part, this is a
-property that the cryptographic handshake protocol {{QUIC-TLS}} is required to
-provide, though additional validation is required for version negotiation (see
+able to receive QUIC packets prior to the connection being established.  Such an
+attacker can potentially send packets that will be accepted by QUIC clients or
+servers.  This version of QUIC attempts to detect this sort of attack, but it
+expects that endpoints will fail to establish a connection rather than
+recovering.  For the most part, the cryptographic handshake protocol
+{{QUIC-TLS}} is responsible for detecting tampering during the handshake, though
+additional validation is required for version negotiation (see
 {{version-validation}}).
 
 Endpoints are permitted to use other methods to detect and attempt to recover
-from interference with the handshake.  For instance, if invalid packets can be
-identified then they can be discarded.  This can be complex or difficult and so
-no specific method is mandated in this version of QUIC.
-
+from interference with the handshake.  Invalid packets may be identified and
+discarded using other methods, but no specific method is mandated in this
+document.
 
 
 ## Spoofed ACK Attack
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -4047,18 +4047,17 @@ connection ID and salt specific to the QUIC version.  This allows endpoints to
 use the same process for authenticating packets that they receive as they use
 after the cryptographic handshake completes.  Packets that cannot be
 authenticated are discarded.  Protecting packets in this fashion provides a
-strong assurance that the sender of the packet saw the Initial packets and
+strong assurance that the sender of the packet saw the Initial packet and
 understood it.
 
 These protections are not intended to be effective against an attacker that is
 able to receive QUIC packets prior to the connection being established.  Such an
-attacker can potentially send packets that will be accepted by QUIC clients or
-servers.  This version of QUIC attempts to detect this sort of attack, but it
-expects that endpoints will fail to establish a connection rather than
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -4017,6 +4017,55 @@ transport to cancel a stream in response to receipt of a STOP_SENDING frame.
 
 # Security Considerations
 
+## Handshake Denial of Service
+
+As an encrypted and authenticated transport QUIC provides a range of protections
+against denial of service.  Once the cryptographic handshake is complete, QUIC
+endpoints discard packets that are not authenticated, greatly limiting the
```
janaiyengar: how about "most packets", since there clearly are unauthenticated packets that are accepted.

draft-ietf-quic-transport.md
```diff
@@ -4017,6 +4017,55 @@ transport to cancel a stream in response to receipt of a STOP_SENDING frame.
...
+is a stateless reset ({{stateless-reset}}) which relies on the token being kept
+secret until it is used.
+
+During the creation of a connection, QUIC only provides protection against
+attack from off the network path.  All QUIC packets contain proof that the
+recipient saw a preceding packet from its peer.
+
+The first mechanism used is the source and destination connection IDs, which are
+required to match those set by a peer.  Except for an Initial packet and
+stateless reset, an endpoint only accepts packets that include a destination
```
janaiyengar: "Except for Initial and Stateless Reset packets"

draft-ietf-quic-transport.md
```diff
@@ -4017,6 +4017,55 @@ transport to cancel a stream in response to receipt of a STOP_SENDING frame.
...
+connection that matches a connection ID the endpoint previously chose.  This is
+the only protection offered for Version Negotiation packets.
+
+The destination connection ID in an Initial packet is selected by a client to be
+unpredictable, which serves an additional purpose.  The packets that carry the
+cryptographic handshake are protected with a key that is derived from the
+connection ID and salt specific to the QUIC version.  This allows endpoints to
```
janaiyengar: s/the connection ID/this connection ID/

draft-ietf-quic-transport.md
```diff
@@ -4017,6 +4017,55 @@ transport to cancel a stream in response to receipt of a STOP_SENDING frame.
...
+use the same process for authenticating packets that they receive as they use
+after the cryptographic handshake completes.  Packets that cannot be
+authenticated are discarded.  Protecting packets in this fashion provides a
+strong assurance that the sender of the packet saw previous packets and
```
janaiyengar: s/saw previous packets and understood them/saw the Initial packet and understood it/

draft-ietf-quic-transport.md
```diff
@@ -4017,6 +4017,55 @@ transport to cancel a stream in response to receipt of a STOP_SENDING frame.
...
+understood them.
+
+These protections are not intended to be effective against an attacker that is
+able to receive QUIC packets.  Such an attacker can potentially send packets
+that will be accepted by QUIC clients or servers.  This version of QUIC only
+attempts to detect this sort of attack.  Interference with the handshake is
```
janaiyengar: s/only attempts to detect/attempts to only detect/

draft-ietf-quic-transport.md
```diff
@@ -4017,6 +4017,55 @@ transport to cancel a stream in response to receipt of a STOP_SENDING frame.
...
+detected when the cryptographic handshake fails.  For the most part, this is a
+property that the cryptographic handshake protocol {{QUIC-TLS}} is required to
+provide, though additional validation is required for version negotiation (see
+{{version-validation}}).
+
+Endpoints are permitted to use other methods to detect and attempt to recover
+from interference with the handshake.  For instance, if invalid packets can be
+identified then they can be discarded.  This can be complex or difficult and so
```
janaiyengar: Suggested replacement for the last two sentences: "Invalid packets may be identified and discarded using other methods, but no specific method is mandated in this document."

draft-ietf-quic-transport.md
martinthomson: Sure.  I was hoping that we could ultimately authenticate even those packets that don't have packet protection in some way.  For instance, by ensuring that the ICMP unreachable includes bytes from a packet we sent.  But maybe that's stretching the definition too far.

draft-ietf-quic-transport.md
martinthomson: I've reworded this to be a little clearer (I hope).  I think that it is important to say that we don't expect endpoints to do anything other than fail when an attacker tampers with the handshake.  I realize that the lead-in to this paragraph didn't scope that properly though.  Have another look.

draft-ietf-quic-transport.md
```diff
@@ -4017,6 +4017,55 @@ transport to cancel a stream in response to receipt of a STOP_SENDING frame.
...
+the only protection offered for Version Negotiation packets.
+
+The destination connection ID in an Initial packet is selected by a client to be
+unpredictable, which serves an additional purpose.  The packets that carry the
+cryptographic handshake are protected with a key that is derived from this
+connection ID and salt specific to the QUIC version.  This allows endpoints to
+use the same process for authenticating packets that they receive as they use
+after the cryptographic handshake completes.  Packets that cannot be
+authenticated are discarded.  Protecting packets in this fashion provides a
+strong assurance that the sender of the packet saw the Initial packets and
```
MikeBishop: packets plural?

draft-ietf-quic-transport.md
```diff
@@ -4035,35 +4035,35 @@ attack from off the network path.  All QUIC packets contain proof that the
 recipient saw a preceding packet from its peer.
 
 The first mechanism used is the source and destination connection IDs, which are
-required to match those set by a peer.  Except for an Initial packet and
-stateless reset, an endpoint only accepts packets that include a destination
+required to match those set by a peer.  Except for an Initial and stateless
+reset packets, an endpoint only accepts packets that include a destination
```
janaiyengar: capitalize stateless reset

draft-ietf-quic-transport.md
janaiyengar: yeah, this should be "packet"

draft-ietf-quic-transport.md
```diff
@@ -4035,35 +4035,35 @@ attack from off the network path.  All QUIC packets contain proof that the
...
 
 These protections are not intended to be effective against an attacker that is
-able to receive QUIC packets.  Such an attacker can potentially send packets
-that will be accepted by QUIC clients or servers.  This version of QUIC only
-attempts to detect this sort of attack.  Interference with the handshake is
-detected when the cryptographic handshake fails.  For the most part, this is a
-property that the cryptographic handshake protocol {{QUIC-TLS}} is required to
-provide, though additional validation is required for version negotiation (see
+able to receive QUIC packets prior to the connection being established.  Such an
+attacker can potentially send packets that will be accepted by QUIC clients or
```
janaiyengar: s/QUIC clients or servers/the QUIC endpoints/ ?

draft-ietf-quic-transport.md
```diff
@@ -4035,35 +4035,35 @@ attack from off the network path.  All QUIC packets contain proof that the
...
+servers.  This version of QUIC attempts to detect this sort of attack, but it
+expects that endpoints will fail to establish a connection rather than
+recovering.  For the most part, the cryptographic handshake protocol
```
janaiyengar: I think this is fine.

draft-ietf-quic-transport.md
martinthomson: That would be a first.  It's lowercase everywhere else.

PR #1353, opened 2018-05-14
Octets that are sent twice

martinthomson:
This is obvious, but never stated.  We discussed checking that they stay constant and agreed that requiring a check would be crazy.  However, we never wrote anything down.  I think that a MAY is fine.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3616,9 +3616,9 @@ MUST be less than 2^62.
 
 QUIC makes no specific allowances for partial reliability or delivery of stream
 data out of order.  Endpoints are expected to deliver stream data to an
-application as an ordered byte-stream.  Delivering an order byte-stream requires
-that an endpoint buffer any data that is received out of order, up to the
-advertised flow control limit.
+application as an ordered byte-stream.  Delivering an ordered byte-stream
+requires that an endpoint buffer any data that is received out of order, up to
+the advertised flow control limit.
 
 An endpoint could receive the same octets multiple times; octets that have
 already been received can be discarded.  The value for a given octet MUST NOT
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3614,13 +3614,16 @@ data.  The first octet of data on a stream has an offset of 0.  An endpoint is
 expected to send every stream octet.  The largest offset delivered on a stream
 MUST be less than 2^62.
 
-A receiver MUST ensure that received stream data is delivered to the application
-as an ordered byte-stream.  Data received out of order MUST be buffered for
-later delivery, as long as it is not in violation of the receiver's flow control
-limits.  An endpoint could receive the same octets multiple times; octets that
-have already been received can be discarded.  The value for a given octet MUST
-NOT change if it is sent multiple times; an endpoint MAY treat receipt of a
-changed octet as a connection error of type PROTOCOL_VIOLATION.
+QUIC makes no specific allowances for partial reliability or delivery of stream
+data out of order.  Endpoints are expected to deliver stream data to an
+application as an ordered byte-stream.  Delivering an order byte-stream requires
+that an endpoint buffer any data that is received out of order, up to the
+advertised flow control limit.
+
+An endpoint could receive the same octets multiple times; octets that have
+already been received can be discarded.  The value for a given octet MUST NOT
+change if it is sent multiple times; an endpoint MAY treat receipt of a changed
+octet as a connection error of type PROTOCOL_VIOLATION.
 
 An endpoint MUST NOT send data on any stream without ensuring that it is within
 the data limits set by its peer.  The cryptographic handshake stream, Stream 0,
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3612,10 +3612,15 @@ When new data is to be sent on a stream, a sender MUST set the encapsulating
 STREAM frame's offset field to the stream offset of the first byte of this new
 data.  The first octet of data on a stream has an offset of 0.  An endpoint is
 expected to send every stream octet.  The largest offset delivered on a stream
-MUST be less than 2^62. A receiver MUST ensure that received stream data is
-delivered to the application as an ordered byte-stream.  Data received out of
-order MUST be buffered for later delivery, as long as it is not in violation of
-the receiver's flow control limits.
+MUST be less than 2^62.
+
+A receiver MUST ensure that received stream data is delivered to the application
```
ianswett: I think Patrick wanted to allow out of order delivery, and I thought the agreement was along the lines of "A receiver MUST provide the received stream data to the application as an ordered byte-stream".  The text you wrote implies it's in-order is the only possible option.

martinthomson: That text isn't changed.  It's what we litigated (at some length), but moved into a new paragraph, because the old one was getting unwieldy.

You are right, though.  It seems to imply that it has to deliver in order, when the intent was that it be capable of doing so.

dtikhonov:
From the proposed change:
> The value for a given octet MUST NOT change if it is sent multiple times...

Why require that it not change?  While inappropriate for HQ, mutable content (perhaps combined with out-of-order delivery) may make for some interesting protocol options.

MikeBishop:
Because it's then unpredictable what version (or combination of versions) will be received by the application.  We'd need more versioning mechanisms if we wanted to employ that properly.

While I agree that the text around ordering reads as stricter than the agreed-upon intent, it's not actually modified by this PR.  The change introduced by this PR is fine; I don't feel strongly about fixing the in-order text while we're here versus creating a separate PR.

dtikhonov:
> Because it's then unpredictable what version (or combination of versions) will be received by the application. We'd need more versioning mechanisms if we wanted to employ that properly.

An application might be OK with it.  What if the content is versioned purely as a function of time?

How about this scenario in HQ: serving a static file.  An implementation may simply `read(2)` the file contents into STREAM frames as packets are sent out, including when lost frames are resent.  The requirement that octets not be modified forces either keeping unacknowledged STREAM frames in memory or locking the file somehow.

MikeBishop:
In that scenario, consider that bytes 0-800 and 1000-1200 are delivered on the first try, then the file changes, and bytes 800-1000 need to be resent.  If the file has changed in the meantime and the server is just reading from disk, you get a window into the new file while still having the outer bytes taken from the old file.

And yes, to avoid that, the server needs to hold read access to the file until it's done reading for good, or retain the content that it transferred, or at the least listen for filesystem change notifications.

I can see value in a solution (probably at the application layer) to abort a response and indicate that the resource has changed.  But randomly intermixing content from the old and new versions without indication of which source it came from, which would be the output of the current structure, seems like an invitation to invalid state.

mikkelfj:
The wonderful mmap api does not guarantee MAP_PRIVATE view remains private on Linux if someone else writes to the file.

I think the requirement of unchanged data is the only sane approach at protocol level, but it can be viewed as a system requirement not the sole responsibility of the QUIC engine because that could range from very expensive to near impossible. Hence, if some other process with adequate permissions overwrite a mmap'ed file being transmitted, that process just violated the QUIC spec.

martinthomson:
@ianswett, @janaiyengar, @mcmanus, can you check  that I'm accurate in my updated statement here.  I think that avoiding MUST is sensible here.

martinthomson:
@dtikhonov, if you want to change the value of an octet, I think that requires some sort of negotiation.  In other words, that's a new feature.

draft-ietf-quic-transport.md
```diff
@@ -3612,10 +3612,18 @@ When new data is to be sent on a stream, a sender MUST set the encapsulating
...
 expected to send every stream octet.  The largest offset delivered on a stream
-MUST be less than 2^62. A receiver MUST ensure that received stream data is
-delivered to the application as an ordered byte-stream.  Data received out of
-order MUST be buffered for later delivery, as long as it is not in violation of
-the receiver's flow control limits.
+MUST be less than 2^62.
+
+QUIC makes no specific allowances for partial reliability or delivery of stream
+data out of order.  Endpoints are expected to deliver stream data to an
+application as an ordered byte-stream.  Delivering an order byte-stream requires
```
ianswett: in-order byte stream?

ianswett: Or ordered?

dtikhonov:
@MikeBishop, @martinthomson -- I am not advocating for this, just want to get to the rationale behind this requirement.

martinthomson:
It's been an unstated assumption.  We're just documenting it, that's all.

PR #1355, opened 2018-05-15
QPACK:  Updated Static Table

### LINKED ISSUE #904: HTTP header compression: static table fit for QUIC
The HTTP header compression mechanism is still TBC but as I understand it, the proposals keep the same HPACK static table. This was formed from the "most frequent header fields used by popular web sites", and I wonder if that is still the case for today's web. For example, perhaps we'll see more usage of `alt-svc` in HTTP/QUIC, in order to keep the session active.

One benefit to keeping the static table would be to allow implementations to reuse of their object representation of the table, however I don't know how realistic that would be.

MikeBishop:
Fixes #904.

Takes a snapshot of the current draft on the wiki and turns it into a PR.

One open issue:  The txt output squeezes the "index" column enough that three-digit indices wrap the third digit.  If anyone knows how to coerce the table layout in kramdown-rfc2629, that would be great.

If it's easier for review purposes, you can also view this [here](https://quicwg.github.io/base-drafts/qpack_static/draft-ietf-quic-qpack.html#rfc.appendix.A).

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -1117,17 +1117,16 @@ Description:
 | 108   | access-control-allow-credentials | FALSE                                                                                                  |
 | 109   | expect-ct                        |                                                                                                        |
 | 110   | content-security-policy          | script-src 'none'; object-src 'none'; base-uri 'none'                                                  |
-| 111   | transfer-encoding                | chunked                                                                                                |
-| 112   | upgrade-insecure-requests        | 1                                                                                                      |
-| 113   | x-served-by                      |                                                                                                        |
-| 114   | x-requested-with                 | xmlhttprequest                                                                                         |
-| 115   | access-control-request-headers   | content-type                                                                                           |
-| 116   | access-control-request-headers   | x-requested-with                                                                                       |
-| 117   | access-control-request-method    | post                                                                                                   |
-| 118   | access-control-request-method    | get                                                                                                    |
-| 119   | purpose                          | prefetch                                                                                               |
-| 120   | x-csrf-token                     | null                                                                                                   |
-| 121   | x-csrf-token                     | undefined                                                                                              |
+| 111   | upgrade-insecure-requests        | 1                                                                                                      |
+| 112   | x-served-by                      |                                                                                                        |
+| 113   | x-requested-with                 | xmlhttprequest                                                                                         |
+| 114   | access-control-request-headers   | content-type                                                                                           |
+| 115   | access-control-request-headers   | x-requested-with                                                                                       |
+| 116   | access-control-request-method    | post                                                                                                   |
+| 117   | access-control-request-method    | get                                                                                                    |
+| 118   | purpose                          | prefetch                                                                                               |
+| 119   | x-csrf-token                     | null                                                                                                   |
+| 120   | x-csrf-token                     | undefined                                                                                              |
 
 # Change Log
 
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -117,9 +117,7 @@ addressed.
 ## Static Table {#table-static}
 
 The static table consists of a predefined static list of header fields, each of
-which has a fixed index over time.  Its entries are defined in Appendix A of
-{{!RFC7541}}. Note that because HPACK did not use zero-based references, there
-is no value at index zero of the static table.
+which has a fixed index over time.  Its entries are defined in {{static-table}}.
 
 ## Dynamic Table {#table-dynamic}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -715,6 +714,133 @@ None.
...
+| 105   | timing-allow-origin              | \*                                                                                                     |
+| 106   | access-control-expose-headers    | content-length                                                                                         |
+| 107   | access-control-allow-credentials | TRUE                                                                                                   |
+| 108   | access-control-allow-credentials | FALSE                                                                                                  |
+| 109   | expect-ct                        |                                                                                                        |
+| 110   | content-security-policy          | script-src 'none'; object-src 'none'; base-uri 'none'                                                  |
+| 111   | transfer-encoding                | chunked                                                                                                |
```
dtikhonov: "chunked" is prohibited by the protocol.  From [draft-ietf-quic-http-11]:

```
   The "chunked" transfer encoding defined in Section 4.1 of [RFC7230]
   MUST NOT be used.
```

LPardue:
Is there a strong case for the CONNECT method? I'd be intrigued to know how many deployed HTTP proxies support connect over H2 let alone HTTP/QUIC. 

I guess one argument for is that it can reduce the handshake overhead, dynamic table is of less use here because many times the CONNECT call is only used once?

MikeBishop:
CONNECT is defined in the draft, and we know that HTTP/2 and gQUIC proxies exist (though they are rare).  However, we could certainly follow 7541's example and keep only the most common methods.

LPardue:
Aside from CONNECTION discussion the PR LGTM, be great to close out #904 after all this time. Thanks for the work!

draft-ietf-quic-qpack.md
```diff
@@ -215,8 +215,7 @@ and follows the definitions in [RFC7541] without modification.
 Entries in the QPACK static and dynamic tables are addressed separately.
 
 Entries in the static table have the same indices at all times.  The static
-table is defined in Appendix A of {{!RFC7541}}. Note that because HPACK did not
-use zero-based references, there is no value at index zero of the static table.
+table is defined in {{static-table}}.
```
LPardue: I think {{static-table}} might be missing...

mikkelfj:
I still don't see any x-forwarded-for and friends - when used they are used a lot, even if it might not be visible in the studied dataset.

draft-ietf-quic-qpack.md
MikeBishop: Implicit anchor on the title of the section; turns into "Appendix A" in the rendered document.

draft-ietf-quic-qpack.md
```diff
@@ -715,6 +714,132 @@ None.
...
+| 106   | access-control-expose-headers    | content-length                                                                                         |
+| 107   | access-control-allow-credentials | TRUE                                                                                                   |
+| 108   | access-control-allow-credentials | FALSE                                                                                                  |
+| 109   | expect-ct                        |                                                                                                        |
+| 110   | content-security-policy          | script-src 'none'; object-src 'none'; base-uri 'none'                                                  |
+| 111   | upgrade-insecure-requests        | 1                                                                                                      |
+| 112   | x-served-by                      |                                                                                                        |
```
drwilco: While a very common header, it doesn't serve any purpose other than debugging. Maybe scrap it from here to encourage people to stop using it?

draft-ietf-quic-qpack.md
```diff
@@ -715,6 +714,132 @@ None.
...
+| 0     | :path                            | /                                                                                                      |
+| 1     | referer                          |                                                                                                        |
+| 2     | content-length                   | 0                                                                                                      |
+| 3     | date                             |                                                                                                        |
+| 4     | last-modified                    |                                                                                                        |
+| 5     | :authority                       |                                                                                                        |
+| 6     | expires                          |                                                                                                        |
```
drwilco: Considering `Expires` is HTTP/1.0, maybe don't include it to encourage people to start using `Cache-Control: max-age`?

martinthomson:
Good enough.  Editors discussed this and for some reason I don't understand would prefer not to have this change in the -12 drafts.  Something about the potential for churn here being hard to deal with.  I don't buy that argument, we're still making major structural changes, so a little churn is expected.  @MikeBishop and @afrind merge at your discretion.

MikeBishop:
My comment was that I'm not certain we'll have sufficient review for consensus before we drop -12, which isn't quite the same statement.  If we do, fantastic.

draft-ietf-quic-qpack.md
```diff
@@ -715,6 +714,132 @@ None.
...
+| 10    | if-modified-since                |                                                                                                        |
+| 11    | if-none-match                    |                                                                                                        |
+| 12    | if-range                         |                                                                                                        |
+| 13    | set-cookie                       |                                                                                                        |
+| 14    | age                              | 0                                                                                                      |
+| 15    | link                             |                                                                                                        |
+| 16    | content-disposition              | attachment; filename="f.txt"                                                                           |
```
mnot: drop the value

draft-ietf-quic-qpack.md
```diff
@@ -715,6 +714,132 @@ None.
...
+| 17    | range                            | bytes=0-                                                                                               |
+| 18    | authorization                    |                                                                                                        |
+| 19    | :scheme                          | https                                                                                                  |
+| 20    | :scheme                          | http                                                                                                   |
+| 21    | :status                          | 200                                                                                                    |
+| 22    | :status                          | 304                                                                                                    |
+| 23    | :path                            | /index.html                                                                                            |
```
mnot: We already have /, this seems vaguely implementation-specific

draft-ietf-quic-qpack.md
mnot: +1

draft-ietf-quic-qpack.md
```diff
@@ -715,6 +714,132 @@ None.
...
+| 80    | accept-ranges                    | bytes                                                                                                  |
+| 81    | alt-svc                          | clear                                                                                                  |
+| 82    | x-xss-protection                 | 1; mode=block                                                                                          |
+| 83    | host                             |                                                                                                        |
+| 84    | pragma                           | no-cache                                                                                               |
+| 85    | pragma                           | public                                                                                                 |
+| 86    | x-powered-by                     |                                                                                                        |
```
mnot: Same as @drwilco 's comment below

draft-ietf-quic-qpack.md
```diff
@@ -715,6 +714,132 @@ None.
...
+| 79    | content-encoding                 | br                                                                                                     |
+| 80    | accept-ranges                    | bytes                                                                                                  |
+| 81    | alt-svc                          | clear                                                                                                  |
+| 82    | x-xss-protection                 | 1; mode=block                                                                                          |
+| 83    | host                             |                                                                                                        |
+| 84    | pragma                           | no-cache                                                                                               |
+| 85    | pragma                           | public                                                                                                 |
```
mnot: No.

draft-ietf-quic-qpack.md
```diff
@@ -715,6 +714,132 @@ None.
...
+| 53    | cache-control                    | no-cache, no-store, must-revalidate                                                                    |
+| 54    | cache-control                    | max-age=604800                                                                                         |
+| 55    | accept-encoding                  | gzip, deflate, br                                                                                      |
+| 56    | accept-encoding                  | gzip, deflate                                                                                          |
+| 57    | vary                             | accept-encoding                                                                                        |
+| 58    | vary                             | origin                                                                                                 |
+| 59    | vary                             | accept-encoding,user-agent                                                                             |
```
mnot: There's no space after the comma. I know that we're basing this on traffic currently seen, but once this is the static table, implementations are going to adjust what they emit to match, making this a pseudo-standard. So I'd suggest either adding a space or dropping this one.

draft-ietf-quic-qpack.md
MikeBishop: To `pragma` entirely, or this particular value?  It's outside spec, but it's commonly used.

mnot: This particular value definitely, but probably 84 too. We shouldn't be encouraging HTTP/1.0 semantics and this isn't going to result in any real-world perceivable performance benefit.

draft-ietf-quic-qpack.md
```diff
@@ -1004,6 +1002,120 @@ Description:
...
+| 28    | accept-encoding                  | gzip, deflate, br                                     |
+| 29    | accept-language                  |                                                       |
+| 30    | authorization                    |                                                       |
+| 31    | cache-control                    | max-age=0                                             |
+| 32    | cache-control                    | max-age=604800                                        |
+| 33    | cache-control                    | max-age=2592000                                       |
+| 34    | cache-control                    | no-cache                                              |
+| 35    | cache-control                    | no-cache, no-store, must-revalidate                   |
+| 36    | cache-control                    | public, max-age=31536000                              |
+| 37    | content-disposition              |                                                       |
```
dtikhonov: You have two identical values -- at 3 and 37.

MikeBishop: Whoops!

draft-ietf-quic-qpack.md
```diff
@@ -1004,6 +1002,119 @@ Description:
...
+| 46    | content-type                     | image/png                                             |
+| 47    | content-type                     | text/css                                              |
+| 48    | content-type                     | text/html; charset=utf-8                              |
+| 49    | content-type                     | text/javascript                                       |
+| 50    | content-type                     | text/javascript; charset=utf-8                        |
+| 51    | content-type                     | text/plain                                            |
+| 52    | content-type                     | text/plain;charset=utf-8                              |
+| 53    | range                            | bytes=0-                                              |
+| 54    | server                           |                                                       |
+| 55    | strict-transport-security        | max-age=10886400; includesubdomains; preload          |
```
martinthomson: I have a request.  Take this out and move :status: 103 up into the one-byte range.

draft-ietf-quic-qpack.md
```diff
@@ -1004,6 +1002,119 @@ Description:
...
+| 32    | cache-control                    | max-age=604800                                        |
+| 33    | cache-control                    | max-age=2592000                                       |
+| 34    | cache-control                    | no-cache                                              |
+| 35    | cache-control                    | no-cache, no-store, must-revalidate                   |
+| 36    | cache-control                    | public, max-age=31536000                              |
+| 37    | content-encoding                 | br                                                    |
+| 38    | content-type                     | application/dns-message                               |
+| 39    | content-type                     | application/javascript                                |
+| 40    | content-type                     | application/json                                      |
+| 41    | content-type                     | application/json; charset=utf-8                       |
```
martinthomson: Maybe we shouldn't include charset parameters either.

MikeBishop: Per discussion, we'll drop charsets from everything except text/html and perhaps text/plain.

draft-ietf-quic-qpack.md
```diff
@@ -1004,6 +1002,119 @@ Description:
...
+| 42    | content-type                     | application/x-www-form-urlencoded                     |
+| 43    | content-type                     | application/x-www-form-urlencoded; charset=utf-8      |
+| 44    | content-type                     | image/gif                                             |
+| 45    | content-type                     | image/jpeg                                            |
+| 46    | content-type                     | image/png                                             |
+| 47    | content-type                     | text/css                                              |
+| 48    | content-type                     | text/html; charset=utf-8                              |
+| 49    | content-type                     | text/javascript                                       |
+| 50    | content-type                     | text/javascript; charset=utf-8                        |
```
martinthomson: https://datatracker.ietf.org/doc/draft-ietf-dispatch-javascript-mjs/ deprecates this, so remove both text/javascript options.

draft-ietf-quic-qpack.md
```diff
@@ -1004,6 +1002,119 @@ Description:
 
 --- back
 
+# Static Table
+
+| Index | Name                             | Value                                                 |
+| ----- | -------------------------------- | ----------------------------------------------------- |
+| 0     | :authority                       |                                                       |
+| 1     | :path                            | /                                                     |
+| 2     | age                              | 0                                                     |
```
mnot: Seems pretty low-value; will never appear in a request, and it's just three letters.

draft-ietf-quic-qpack.md
```diff
@@ -1004,6 +1002,119 @@ Description:
...
+| 18    | :method                          | HEAD                                                  |
+| 19    | :method                          | OPTIONS                                               |
+| 20    | :method                          | POST                                                  |
+| 21    | :method                          | PUT                                                   |
+| 22    | :scheme                          | http                                                  |
+| 23    | :scheme                          | https                                                 |
+| 24    | :status                          | 200                                                   |
+| 25    | :status                          | 304                                                   |
+| 26    | accept                           | */*                                                   |
+| 27    | accept                           | application/dns-message                               |
```
mnot: Seriously?

draft-ietf-quic-qpack.md
```diff
@@ -1004,6 +1002,119 @@ Description:
...
+| 28    | accept-encoding                  | gzip, deflate, br                                     |
+| 29    | accept-language                  |                                                       |
+| 30    | authorization                    |                                                       |
+| 31    | cache-control                    | max-age=0                                             |
+| 32    | cache-control                    | max-age=604800                                        |
+| 33    | cache-control                    | max-age=2592000                                       |
+| 34    | cache-control                    | no-cache                                              |
+| 35    | cache-control                    | no-cache, no-store, must-revalidate                   |
```
mnot: this has equivalent semantics to 'no-store'. I know some variation on this is common practice, but it's bad practice, and I doubt that there's enough convergence on this particular spelling to make the codification of that bad practice worthwhile. Just do `no-store`.

draft-ietf-quic-qpack.md
```diff
@@ -1004,6 +1002,119 @@ Description:
...
+| 52    | content-type                     | text/plain;charset=utf-8                              |
+| 53    | range                            | bytes=0-                                              |
+| 54    | server                           |                                                       |
+| 55    | strict-transport-security        | max-age=10886400; includesubdomains; preload          |
+| 56    | strict-transport-security        | max-age=31536000                                      |
+| 57    | strict-transport-security        | max-age=31536000; includesubdomains                   |
+| 58    | strict-transport-security        | max-age=31536000; includesubdomains; preload          |
+| 59    | user-agent                       |                                                       |
+| 60    | vary                             | accept-encoding                                       |
+| 61    | vary                             | accept-encoding, user-agent                           |
```
mnot: This is actively bad practice; it should not be encouraged.

draft-ietf-quic-qpack.md
```diff
@@ -1004,6 +1002,119 @@ Description:
...
+| 65    | :status                          | 204                                                   |
+| 66    | :status                          | 206                                                   |
+| 67    | :status                          | 302                                                   |
+| 68    | :status                          | 400                                                   |
+| 69    | :status                          | 403                                                   |
+| 70    | :status                          | 404                                                   |
+| 71    | :status                          | 421                                                   |
+| 72    | :status                          | 425                                                   |
+| 73    | :status                          | 500                                                   |
+| 74    | :status                          | 503                                                   |
```
mnot: Are putting three-character status codes in the static table really necessary?

draft-ietf-quic-qpack.md
```diff
@@ -1004,6 +1002,119 @@ Description:
...
+| 76    | access-control-allow-credentials | FALSE                                                 |
+| 77    | access-control-allow-credentials | TRUE                                                  |
+| 78    | access-control-allow-headers     | *                                                     |
+| 79    | access-control-allow-headers     | cache-control                                         |
+| 80    | access-control-allow-headers     | content-type                                          |
+| 81    | access-control-allow-headers     | x-requested-with                                      |
+| 82    | access-control-allow-methods     | get                                                   |
+| 83    | access-control-allow-methods     | get, post, options                                    |
+| 84    | access-control-allow-methods     | options                                               |
+| 85    | access-control-allow-origin      | *                                                     |
```
mnot: All of the above are response headers; is it worth having so many variations?

draft-ietf-quic-qpack.md
```diff
@@ -1004,6 +1002,119 @@ Description:
...
+| 90    | access-control-request-method    | post                                                  |
+| 91    | alt-svc                          | clear                                                 |
+| 92    | content-encoding                 | gzip                                                  |
+| 93    | content-security-policy          | script-src 'none'; object-src 'none'; base-uri 'none' |
+| 94    | early-data                       | 1                                                     |
+| 95    | expect-ct                        |                                                       |
+| 96    | forwarded                        |                                                       |
+| 97    | if-range                         |                                                       |
+| 98    | origin                           |                                                       |
+| 99    | purpose                          | prefetch                                              |
```
mnot: This is going away AIUI.

draft-ietf-quic-qpack.md
```diff
@@ -1004,6 +1002,119 @@ Description:
...
+| 79    | access-control-allow-headers     | cache-control                                         |
+| 80    | access-control-allow-headers     | content-type                                          |
+| 81    | access-control-allow-headers     | x-requested-with                                      |
+| 82    | access-control-allow-methods     | get                                                   |
+| 83    | access-control-allow-methods     | get, post, options                                    |
+| 84    | access-control-allow-methods     | options                                               |
+| 85    | access-control-allow-origin      | *                                                     |
+| 86    | access-control-expose-headers    | content-length                                        |
+| 87    | access-control-request-headers   | content-type                                          |
+| 88    | access-control-request-headers   | x-requested-with                                      |
```
mnot: People using this header usually don't understand what they're getting; e.g., they use it for access control.

mikkelfj: I don't remember what it means 5 minutes after I read about it due to lack of hint of intuition, but CORS is used a lot, especially in CDN, so it makes a lot of sense to have, if it it means what I think it means.

draft-ietf-quic-qpack.md
```diff
@@ -1030,90 +1030,84 @@ Description:
...
+| 92    | origin                           |                                                       |
+| 93    | purpose                          | prefetch                                              |
+| 94    | server                           |                                                       |
+| 95    | timing-allow-origin              | *                                                     |
+| 96    | upgrade-insecure-requests        | 1                                                     |
+| 97    | user-agent                       |                                                       |
+| 98    | x-forwarded-for                  |                                                       |
+| 99    | x-frame-options                  | deny                                                  |
+| 100   | x-frame-options                  | sameorigin                                            |
+| 101   | x-requested-with                 | xmlhttprequest                                        |
```
mnot: x-requested-with might be common, but it's being sent by old JS frameworks; it serves no real purpose, so we shouldn't be optimising for it.

PR #1357, opened 2018-05-18
Core concepts for QPACK

MikeBishop:
Borrows a lot of RFC7540 text into the introduction to make it less critical to fully understand HPACK prior to reading the QPACK spec.  (Note that it's not a goal of this PR to make this spec fully independent of HPACK, but rather to have the core concepts live in QPACK and reference specific sections of RFC7541 for concrete items like the Huffman table, integer encodings, etc.)

Also moved the Indexing discussion into the Introduction to live with this text, since it's thematically related.  Along the way, removed the text that QPACK is a "relatively non-intrusive extension to HPACK" since, well, it's not.  ;-)

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -52,46 +52,162 @@ code and issues list for this draft can be found at
 # Introduction
 
 The QUIC transport protocol was designed from the outset to support HTTP
-semantics, and its design subsumes many of the features of HTTP/2.  QUIC's
-stream multiplexing comes into some conflict with header compression.  A key
-goal of the design of QUIC is to improve stream multiplexing relative to HTTP/2
-by eliminating HoL (head of line) blocking, which can occur in HTTP/2.  HoL
-blocking can happen because all HTTP/2 streams are multiplexed onto a single TCP
-connection with its in-order semantics.  QUIC can maintain independence between
-streams because it implements core transport functionality in a fully
-stream-aware manner.  However, the HTTP/QUIC mapping is still subject to HoL
-blocking if HPACK is used directly.  HPACK exploits multiplexing for greater
-compression, shrinking the representation of headers that have appeared earlier
-on the same connection.  In the context of QUIC, this imposes a vulnerability to
-HoL blocking (see {{hol-example}}).
+semantics, and its design subsumes many of the features of HTTP/2.  HTTP/2 used
+HPACK ({{!RFC7541}}) for header compression, but QUIC's stream multiplexing
+comes into some conflict with HPACK.  A key goal of the design of QUIC is to
+improve stream multiplexing relative to HTTP/2 by reducing head-of-line
+blocking.  If HPACK were used for HTTP/QUIC, it would induce head-of-line
+blocking due to built-in assumptions of a total ordering across frames on all
+streams.
 
 QUIC is described in {{?QUIC-TRANSPORT=I-D.ietf-quic-transport}}.  The HTTP/QUIC
 mapping is described in {{!QUIC-HTTP=I-D.ietf-quic-http}}. For a full
 description of HTTP/2, see {{?RFC7540}}. The description of HPACK is
 {{!RFC7541}}, with important terminology in Section 1.3.
 
-QPACK modifies HPACK to allow correctness in the presence of out-of-order
-delivery, with flexibility for implementations to balance between resilience
-against HoL blocking and optimal compression ratio.  The design goals are to
-closely approach the compression ratio of HPACK with substantially less
-head-of-line blocking under the same loss conditions.
+QPACK reuses core concepts from HPACK, but is redesigned to allow correctness in
+the presence of out-of-order delivery, with flexibility for implementations to
+balance between resilience against head-of-line blocking and optimal compression
+ratio.  The design goals are to closely approach the compression ratio of HPACK
+with substantially less head-of-line blocking under the same loss conditions.
 
-QPACK is intended to be a relatively non-intrusive extension to HPACK; an
-implementation should be easily shared within stacks supporting both HTTP/2 over
-(TLS+)TCP and HTTP/QUIC.
+## Header Tables
 
-## Head-of-Line Blocking in HPACK {#hol-example}
+Like HPACK, QPACK uses two tables for associating header fields to indexes.  The
+static table (see {{table-static}}) is predefined and contains common header
+fields (some of them with an empty value).  The dynamic table (see
+{{table-dynamic}}) built up over the course of the connection and can be used by
+the encoder to index header fields repeated in the encoded header lists.
 
-HPACK enables several types of header representations, one of which also adds
-the header to a dynamic table of header values.  These values are then available
-for reuse in subsequent header blocks simply by referencing the entry number in
-the table.
+These two tables are indexed in a different manner than in HPACK; see
+{{indexing}} for more details.
 
-If the packet containing a header is lost, that stream cannot complete header
-processing until the packet is retransmitted.  This is unavoidable. However,
-other streams which rely on the state created by that packet *also* cannot make
-progress. This is the problem which QUIC solves in general, but which is
-reintroduced by HPACK when the loss includes a HEADERS frame.
+### Static Table {#table-static}
+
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -52,46 +52,162 @@ code and issues list for this draft can be found at
...
+
+This mechanism can be used to completely clear entries from the dynamic table by
+setting a maximum size of 0, which can subsequently be restored.
+
+### Indexing
+
+Entries in the QPACK static and dynamic tables are addressed separately.
+
+While entries in the static table have the same indices at all times, entries
+are inserted into the dynamic table over time.  Each entry possesses both an
```
martinthomson: change

draft-ietf-quic-qpack.md
```diff
@@ -52,46 +52,162 @@ code and issues list for this draft can be found at
...
+absolute index which is fixed for the lifetime of that entry and a relative
+index which changes over time based on the context of the reference. The first
+entry inserted has an absolute index of "1"; indices increase sequentially with
+each insertion.
```
martinthomson: You need to explain that relative indexes increase in the opposite direction to the absolute index.  This is the most confusing aspect of this entire thing.

draft-ietf-quic-qpack.md
```diff
@@ -52,46 +52,162 @@ code and issues list for this draft can be found at
...
+n = count of entries inserted
+d = count of entries dropped
+~~~~~
+{: title="Example Dynamic Table Indexing - Control Stream"}
+
+Because frames from request streams can be delivered out of order with
+instructions on the control stream, relative indices are relative to the Base
+Index at the beginning of the header block (see {{absolute-index}}). The Base
+Index is the absolute index of the entry which has the relative index of zero
+when interpreting the frame.  The relative indices of entries do not change
```
martinthomson: You can drop the "which has" onwards.  I find it confusing.  How about: "The Base Index is an absolute index.  All subsequent references use relative index, with the entry identified in the Base Index having a relative index of 0."

draft-ietf-quic-qpack.md
```diff
@@ -52,46 +52,162 @@ code and issues list for this draft can be found at
...
+    +---+-----+  -  +-----+   -   +
+              |  0  | ... | n-d-3 |  Relative Index
+              +-----+-----+-------+
+
+n = count of entries inserted
+d = count of entries dropped
+~~~~~
+{: title="Example Dynamic Table Indexing - Request Stream"}
+
+Entries with an absolute index greater than a frame's Base Index can be
```
martinthomson: Explain why: "A header block on the request stream can reference entries added after the entry identified by the Base Index.  This allows an encoder to process a header block in a single pass and include references to entries added while processing this (or other) header blocks.  Newly added entries are referenced using Post-Base instructions.  Indices for Post-Base instructions increase in the same direction as absolute indices, but the zero value is one higher than the Base Index."

martinthomson: Also, is the Largest Reference a Post-Base index?  It should be.

draft-ietf-quic-qpack.md
```diff
@@ -52,46 +52,162 @@ code and issues list for this draft can be found at
...
+QPACK reuses core concepts from HPACK, but is redesigned to allow correctness in
+the presence of out-of-order delivery, with flexibility for implementations to
+balance between resilience against head-of-line blocking and optimal compression
+ratio.  The design goals are to closely approach the compression ratio of HPACK
+with substantially less head-of-line blocking under the same loss conditions.
 
-QPACK is intended to be a relatively non-intrusive extension to HPACK; an
-implementation should be easily shared within stacks supporting both HTTP/2 over
-(TLS+)TCP and HTTP/QUIC.
+## Header Tables
```
martinthomson: Make this h1.

draft-ietf-quic-qpack.md
```diff
@@ -52,46 +52,162 @@ code and issues list for this draft can be found at
...
-progress. This is the problem which QUIC solves in general, but which is
-reintroduced by HPACK when the loss includes a HEADERS frame.
+### Static Table {#table-static}
+
+The static table consists of a predefined static list of header fields.  Its
+entries are defined in Appendix A of {{!RFC7541}}. Note that because HPACK did
+not use zero-based references, there is no value at index zero of the static
+table.
+
+### Dynamic Table {#table-dynamic}
```
martinthomson: This would benefit from subsections so that people can find Absolute and Relative indexing, "Base Index", Post-Base, etc...

draft-ietf-quic-qpack.md
MikeBishop: Clarity

draft-ietf-quic-qpack.md
MikeBishop: Not quite, though LR and Base are given one as absolute and one relative to the other. LR might be less than Base, or more, or the same.

draft-ietf-quic-qpack.md
martinthomson: table [change] over time ?

draft-ietf-quic-qpack.md
martinthomson: I don't think that there is any value in having Largest Reference with a lower absolute value than the Base Index.  I mean, sure, it can happen, but why would you encode that?

MikeBishop: Because a likely algorithm would be that Base is the largest entry you've inserted in the dynamic table, before you've examined the incoming header set.  If you don't wind up using entries that are in flight (or any entries, in the extreme case), then Largest Reference will indicate that the most recent entries aren't needed.

martinthomson: Right.  Of course.

draft-ietf-quic-qpack.md
```diff
@@ -52,46 +52,168 @@ code and issues list for this draft can be found at
...
+increase sequentially with each insertion.
+
+The relative index begins at zero and increases in the opposite direction from
+the absolute index.  Determining which entry has a relative index of "0" depends
+on the context of the reference.
+
+On the control stream, a relative index of "0" always refers to the most
+recently inserted value in the dynamic table.  Note that this means the
+entry referenced by a given relative index can change while interpreting
+a HEADERS frame as new entries are inserted.
```
afrind: HEADERS frames do not insert entries.

draft-ietf-quic-qpack.md
```diff
@@ -52,46 +52,168 @@ code and issues list for this draft can be found at
...
+
+~~~~~ drawing
+    +---+---------------+-------+
+    | n |      ...      | d + 1 |  Absolute Index
+    + - +---------------+   -   +
+    | 0 |      ...      | n-d-1 |  Relative Index
```
afrind: Nit: d + 1 has spaces around the operator and n-d-1 does not.  My preference would be to expand the column so both could have spaces.

draft-ietf-quic-qpack.md
```diff
@@ -52,46 +52,168 @@ code and issues list for this draft can be found at
...
+
+The dynamic table consists of a list of header fields maintained in first-in,
+first-out order.  The dynamic table is initially empty.  Entries are added by
+instructions on the Encoder Stream (see {{encoder-stream}}).
+
+Before a new entry is added to the dynamic table, entries are evicted from the
+end of the dynamic table until the size of the dynamic table is less than or
+equal to (maximum size - new entry size) or until the table is empty.
+
+If the size of the new entry is less than or equal to the maximum size, that
```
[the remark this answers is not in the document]
MikeBishop: If you're doing this intentionally, that's a better approach and might be worth mentioning.  However, I think we need to define what happens in this situation regardless, and the HPACK precedent is *a* logical choice.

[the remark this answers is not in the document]
martinthomson: Given that there is zero utility from blowing the table out like this, I think that we might be better prohibiting that outright.  In HPACK, this caused the header to be emitted as well, but here all you are doing is wasting bytes.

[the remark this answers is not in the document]
MikeBishop: So inserting a header that's larger than the static table is a hard error?  Works for me.

PR #1358, opened 2018-05-18
Move setting registrations into QPACK

### LINKED ISSUE #1323: QPACK settings in HTTP
On #1143, @martinthomson said:
> if -http depends on -qcram, I think that we need to define configuration parameters in -qcram and explain how the settings are used to set these parameters in -http. Right now the drafts are tightly coupled.

He's probably correct, but there appear to be enough ripple effects that I'd prefer to handle this change separately.

MikeBishop:
Fixes #1323.  Trying to describe how the max table size was used made me realize we never actually say, which is why I also wrote #1357; there's a reference that needs to be updated in whichever PR lands second.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -777,18 +777,13 @@ Settings which are integers use the QUIC variable-length integer encoding.
 
 #### Defined SETTINGS Parameters {#settings-parameters}
 
-The following settings are defined in HTTP/QUIC:
-
-  SETTINGS_HEADER_TABLE_SIZE (0x1):
-  : An integer with a maximum value of 2^30 - 1.  The default value is 4,096
-    bytes.
+The following setting is defined in HTTP/QUIC:
 
   SETTINGS_MAX_HEADER_LIST_SIZE (0x6):
   : An integer with a maximum value of 2^30 - 1.  The default value is
     unlimited.
 
-  SETTINGS_QPACK_BLOCKED_STREAMS (0x7):
-  : An integer with a maximum value of 2^16 - 1.  The default value is 100.
+Additional settings MAY be defined by extensions to HTTP/QUIC.
 
 #### Initial SETTINGS Values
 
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-http.md
```diff
@@ -777,18 +777,13 @@ Settings which are integers use the QUIC variable-length integer encoding.
...
 #### Defined SETTINGS Parameters {#settings-parameters}
 
-The following settings are defined in HTTP/QUIC:
-
-  SETTINGS_HEADER_TABLE_SIZE (0x1):
-  : An integer with a maximum value of 2^30 - 1.  The default value is 4,096
-    bytes.
+The following setting is defined in HTTP/QUIC:
 
   SETTINGS_MAX_HEADER_LIST_SIZE (0x6):
```
afrind: Does anything in this document explain what SETTINGS_MAX_HEADER_LIST_SIZE does?

draft-ietf-quic-http.md
```diff
@@ -777,18 +777,13 @@ Settings which are integers use the QUIC variable-length integer encoding.
...
   : An integer with a maximum value of 2^30 - 1.  The default value is
     unlimited.
 
-  SETTINGS_QPACK_BLOCKED_STREAMS (0x7):
-  : An integer with a maximum value of 2^16 - 1.  The default value is 100.
+Additional settings MAY be defined by extensions to HTTP/QUIC.
```
afrind: Are we calling QPACK an extension to HTTP/QUIC?  Or it is for purposes of SETTINGS?

draft-ietf-quic-http.md
```diff
@@ -1418,13 +1413,11 @@ The entries in the following table are registered by this document.
 |----------------------------|------|-------------------------|
 | Setting Name               | Code | Specification           |
 |----------------------------|:----:|-------------------------|
-| HEADER_TABLE_SIZE          | 0x1  | {{settings-parameters}} |
```
afrind: It strikes me as bizarre that 0x1 is omitted from this table / or that QPACK - as an extension or subordinate document - gets to define the lowest numbered setting value.

draft-ietf-quic-http.md
MikeBishop: Good point.  RFC7540 doesn't have much, but it has more than HQ does currently.

draft-ietf-quic-http.md
MikeBishop: Kind of.  It's required to parse HEADERS and PUSH_PROMISE, but it is its own thing.  Kind of a mandatory extension, I suppose?

PR #1359, opened 2018-05-19
Unidirectional Stream Headers in HTTP

### LINKED ISSUE #910: Applications and Stream IDs

MikeBishop:
Fixes #910, implements the idea discussed in #1323.

Unidirectional streams have a header which indicates what type they are.  Two types are defined by HTTP (control stream, push stream); two by QPACK (encoder stream, decoder stream).  Others can be defined in the future, somewhat analogously to new frame types.  (But if you send one the peer doesn't understand, you get a STOP_SENDING.)

Also revived HTTP_CLOSED_CRITICAL_STREAM, which got dropped somewhere along the way, but was still referenced.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -219,9 +219,9 @@ While connection-level options pertaining to the core QUIC protocol are set in
 the initial crypto handshake, HTTP/QUIC-specific settings are conveyed in the
 SETTINGS frame. After the QUIC connection is established, a SETTINGS frame
 ({{frame-settings}}) MUST be sent by each endpoint as the initial frame of their
-respective HTTP control stream (Stream ID 2 or 3, see {{stream-mapping}}). The
-server MUST NOT send data on any other stream until the client's SETTINGS frame
-has been received.
+respective HTTP control stream (see {{control-streams}}). The server MUST NOT
+send data on any other stream until the client's SETTINGS frame has been
+received.
 
 ## Connection Reuse
 
@@ -1277,8 +1277,8 @@ Because the Flags field is not present in generic HTTP/QUIC frames, those frames
 which depend on the presence of flags need to allocate space for flags as part
 of their frame payload.
 
-Other than these issues, frame type HTTP/2 extensions are typically portable to
-QUIC simply by replacing Stream 0 in HTTP/2 with Stream 2 or 3 in HTTP/QUIC.
+Other than this issue, frame type HTTP/2 extensions are typically portable to
+QUIC simply by replacing Stream 0 in HTTP/2 with a control stream in HTTP/QUIC.
 HTTP/QUIC extensions will not assume ordering, but would not be harmed by
 ordering, and would be portable to HTTP/2 in the same manner.
 
@@ -1551,15 +1551,10 @@ Specification:
 
 The entries in the following table are registered by this document.
 
-<<<<<<< HEAD
-<<<<<<< HEAD
-=======
->>>>>>> f61a811... Review feedback
 | ---------------------------- | ------ | ------------------------- |
 | Setting Name                 | Code   | Specification             |
 | ---------------------------- | :----: | ------------------------- |
 | Reserved                     | 0x2    | N/A                       |
-<<<<<<< HEAD
 | NUM_PLACEHOLDERS             | 0x3    | {{settings-parameters}}   |
 | Reserved                     | 0x4    | N/A                       |
 | Reserved                     | 0x5    | N/A                       |
@@ -1575,26 +1570,6 @@ Name:
 
 Specification:
 : {{settings-parameters}}
-=======
-|----------------------------|------|-------------------------|
-| Setting Name               | Code | Specification           |
-|----------------------------|:----:|-------------------------|
-| Reserved                   | 0x2  | N/A                     |
-| Reserved                   | 0x3  | N/A                     |
-| Reserved                   | 0x4  | N/A                     |
-| Reserved                   | 0x5  | N/A                     |
-| MAX_HEADER_LIST_SIZE       | 0x6  | {{settings-parameters}} |
-| QPACK_BLOCKED_STREAMS      | 0x7  | {{settings-parameters}} |
-|----------------------------|------|-------------------------|
->>>>>>> 493a2bb... WIP: Unidirectional stream headers
-=======
-| Reserved                     | 0x3    | N/A                       |
-| Reserved                     | 0x4    | N/A                       |
-| Reserved                     | 0x5    | N/A                       |
-| MAX_HEADER_LIST_SIZE         | 0x6    | {{settings-parameters}}   |
-| QPACK_BLOCKED_STREAMS        | 0x7    | {{settings-parameters}}   |
-| ---------------------------- | ------ | ------------------------- |
->>>>>>> f61a811... Review feedback
 
 ## Error Codes {#iana-error-codes}
 
@@ -1647,8 +1622,6 @@ The entries in the following table are registered by this document.
 | HTTP_WRONG_STREAM_DIRECTION         | 0x0010     | Unidirectional stream in wrong direction | {{http-error-codes}}   |
 | HTTP_MALFORMED_FRAME                | 0x01XX     | Error in frame formatting or use         | {{http-error-codes}}   |
 | ----------------------------------- | ---------- | ---------------------------------------- | ---------------------- |
-<<<<<<< HEAD
-=======
 
 ## Stream Types {#iana-stream-types}
 
@@ -1677,22 +1650,12 @@ Sender:
 
 The entries in the following table are registered by this document.
 
-<<<<<<< HEAD
-| ---------------- | ------ | -------------------------- |
-| Stream Type      | Code   | Specification              |
-| ---------------- | :----: | -------------------------- |
-| Control Stream   | 0x43   | {{control-streams}}        |
-| Push Stream      | 0x50   | {{server-push}}            |
-| ---------------- | ------ | -------------------------- |
->>>>>>> 493a2bb... WIP: Unidirectional stream headers
-=======
 | ---------------- | ------ | -------------------------- | ------ |
 | Stream Type      | Code   | Specification              | Sender |
 | ---------------- | :----: | -------------------------- | ------ |
 | Control Stream   | 0x43   | {{control-streams}}        | Both   |
 | Push Stream      | 0x50   | {{server-push}}            | Server |
 | ---------------- | ------ | -------------------------- | ------ |
->>>>>>> f61a811... Review feedback
 
 
 --- back
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -530,7 +530,7 @@ use server push until it receives a MAX_PUSH_ID frame.  Only servers can push;
 if a server receives a client-initiated push stream, this MUST be treated as a
 stream error of type HTTP_PUSH_LIMIT_EXCEEDED.
 
-A push stream is indicated by a stream type of `0x80`, followed by the Push ID
+A push stream is indicated by a stream type of `0x50`, followed by the Push ID
 of the promise that it fulfills, encoded as a variable-length integer.
 
 ~~~~~~~~~~ drawing
@@ -1625,6 +1625,8 @@ The entries in the following table are registered by this document.
 | HTTP_PUSH_LIMIT_EXCEEDED            | 0x000B     | Maximum Push ID exceeded                 | {{http-error-codes}}   |
 | HTTP_DUPLICATE_PUSH                 | 0x000C     | Push ID was fulfilled multiple times     | {{http-error-codes}}   |
 | HTTP_MALFORMED_FRAME                | 0x01XX     | Error in frame formatting or use         | {{http-error-codes}}   |
+| HTTP_UNKNOWN_STREAM_TYPE            | 0x0D       | Unknown unidirectional stream type       | {{http-error-codes}}   |
+| HTTP_TOO_CONTROLLING                | 0x0E       | Multiple control streams                 | {{http-error-codes}}   |
 | ----------------------------------- | ---------- | ---------------------------------------- | ---------------------- |
 <<<<<<< HEAD
 =======
@@ -1655,7 +1657,7 @@ The entries in the following table are registered by this document.
 | Stream Type      | Code   | Specification              |
 | ---------------- | :----: | -------------------------- |
 | Control Stream   | 0x43   | {{control-streams}}        |
-| Push Stream      | 0x80   | {{server-push}}            |
+| Push Stream      | 0x50   | {{server-push}}            |
 | ---------------- | ------ | -------------------------- |
 >>>>>>> 493a2bb... WIP: Unidirectional stream headers
 
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -506,8 +506,8 @@ the peer's discretion.
 
 ###  Control Streams
 
-The control stream is indicated by a stream type of `0x43`.  Data on this stream
-consists of HTTP frames, as defined in {{frames}}.
+The control stream is indicated by a stream type of `0x43` (ASCII 'C').  Data on
+this stream consists of HTTP frames, as defined in {{frames}}.
 
 Each side MUST initiate a single control stream at the beginning of the
 connection and send its SETTINGS frame as the first frame on this stream.  Only
@@ -530,8 +530,9 @@ use server push until it receives a MAX_PUSH_ID frame.  Only servers can push;
 if a server receives a client-initiated push stream, this MUST be treated as a
 stream error of type HTTP_WRONG_STREAM_DIRECTION.
 
-A push stream is indicated by a stream type of `0x50`, followed by the Push ID
-of the promise that it fulfills, encoded as a variable-length integer.
+A push stream is indicated by a stream type of `0x50` (ASCII 'P'), followed by
+the Push ID of the promise that it fulfills, encoded as a variable-length
+integer.
 
 ~~~~~~~~~~ drawing
  0                   1                   2                   3
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -512,8 +512,8 @@ Each side MUST initiate a single control stream at the beginning of the
 connection and send its SETTINGS frame as the first frame on this stream.  Only
 one control stream per peer is permitted; receipt of a second stream which
 claims to be a control stream MUST be treated as a connection error of type
-HTTP_TOO_CONTROLLING.  If the control stream is closed at any point, this MUST
-be treated as a connection error of type HTTP_CLOSED_CRITICAL_STREAM.
+HTTP_WRONG_STREAM_COUNT.  If the control stream is closed at any point, this
+MUST be treated as a connection error of type HTTP_CLOSED_CRITICAL_STREAM.
 
 A pair of unidirectional streams is used rather than a single bidirectional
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-http.md
```diff
@@ -432,17 +409,76 @@ either by identifying the stream that carries a request or by using a Push ID
...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+~~~~~~~~~~
+{: #fig-push-stream-header title="Push Stream Header"}
 
-As with server push for HTTP/2, the server initiates a server push by sending a
-PUSH_PROMISE frame (see {{frame-push-promise}}) that includes request headers
-for the promised request.  Promised requests MUST conform to the requirements in
+The remaining data on this stream consists of HTTP frames, as defined in
+{{frames}}, and carries the response side of an HTTP message exchange as
+described in {{request-response}}.  The request headers of the exchange is
```
LPardue: s/is/are

draft-ietf-quic-http.md
```diff
@@ -432,17 +409,76 @@ either by identifying the stream that carries a request or by using a Push ID
...
+carried by a PUSH_PROMISE frame (see {{frame-push-promise}}) on the request
```
LPardue: on the request stream?

draft-ietf-quic-http.md
```diff
@@ -452,12 +488,8 @@ certain parts of the response is important (see Section 8.2.1 of {{!RFC7540}}).
...
 Unlike HTTP/2, the PUSH_PROMISE does not reference a stream; it contains a Push
 ID. The Push ID uniquely identifies a server push. This allows a server to
-fulfill promises in the order that best suits its needs.
-
-When a server later fulfills a promise, the server push response is conveyed on
-a push stream.  A push stream is a server-initiated, unidirectional stream.  A
-push stream identifies the Push ID of the promise that it fulfills, encoded as a
-variable-length integer.
+fulfill promises in the order that best suits its needs.  When a server later
+fulfills a promise, the server push response is conveyed on a push stream.
```
LPardue: Perhaps this paragraph now fits better above the push stream header drawing?

draft-ietf-quic-http.md
```diff
@@ -1457,23 +1488,55 @@ Specification:
...
+  including the layout semantics of its payload.
+
+The entries in the following table are registered by this document.
+
+| ---------------- | ------ | -------------------------- |
+| Stream Type      | Code   | Specification              |
+| ---------------- | :----: | -------------------------- |
+| Control Stream   | 0x43   | {{control-streams}}        |
+| Push Stream      | 0x50   | {{server-push}}            |
+| ---------------- | ------ | -------------------------- |
```
LPardue: @I think it would be useful to state explicitly which direction that unidirectional stream types apply to i.e. client-initiated, server-initiated or both. Then we can just have a generic rule that says if a stream header is received on a directional that is not permitted by the type specification, then generate a generic error with HTTP_WRONG_STREAM_TYPE (or something else).

draft-ietf-quic-http.md
```diff
@@ -432,17 +409,76 @@ either by identifying the stream that carries a request or by using a Push ID
 Only a client can send PRIORITY frames.  A server MUST NOT send a PRIORITY
 frame.
 
+## Unidirectional Streams
+
+Unidirectional streams in both directions possess a stream header that indicates
+the stream's purpose.  Some purposes are defined in this document; other
+purposes MAY be defined by extensions to HTTP/QUIC.
+
```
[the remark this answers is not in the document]
martinthomson: Or shorter: "Unidirectional streams, in either direction, are used for a range of purposes. The purpose is indicated by a stream type, which is sent as a single octet header at the start of the stream.  The format and structure of data that follows this header is determined by the stream type.

draft-ietf-quic-http.md
```diff
@@ -432,17 +409,76 @@ either by identifying the stream that carries a request or by using a Push ID
...
+ 0 1 2 3 4 5 6 7
++-+-+-+-+-+-+-+-+
+|Stream Type (8)|
++-+-+-+-+-+-+-+-+
+~~~~~~~~~~
+{: #fig-stream-header title="Unidirectional Stream Header"}
+
+The stream header consists of a single byte, indicating the type of data to
+follow.  The format and structure of that data are defined by the specification
+of the stream type.
```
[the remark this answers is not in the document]
martinthomson: s/MAY/can/  (not an interoperability requirement, but a statement of fact)

draft-ietf-quic-http.md
```diff
@@ -432,17 +409,76 @@ either by identifying the stream that carries a request or by using a Push ID
...
+
+If the header indicates a stream type which is not supported by the recipient,
+the stream MUST be aborted (using a QUIC STOP_SENDING frame) with an error of
+type HTTP_UNKNOWN_STREAM_TYPE.  The remainder of the stream SHOULD NOT be read
+by the recipient.
```
martinthomson: If your strategy here is to permit the use of these extensions without negotiation, the possibility that stream errors might be promoted to connection errors remain.  You should be clearer on this point.

I also don't think that you can require the stream to be stopped.  I won't send STOP_SENDING if the stream closes before I get around to it.

And for "SHOULD NOT be read", that's not necessary.  The semantics are undefined, so whether or not the data is read is immaterial.

draft-ietf-quic-http.md
```diff
@@ -432,17 +409,76 @@ either by identifying the stream that carries a request or by using a Push ID
...
-## Server Push
+The control stream is indicated by a stream type of `0x43`.  Data on this stream
+consists of HTTP frames, as defined in {{frames}}.
 
-HTTP/QUIC supports server push in a similar manner to {{!RFC7540}}, but uses
+Each side MUST initiate a single control stream at the beginning of the
+connection and send its SETTINGS frame as the first frame on this stream.  Only
+one control stream is permitted; receipt of a second stream which claims to be a
+control stream MUST be treated as a connection error of type
+HTTP_TOO_CONTROLLING.  If the control stream is closed at any point, this MUST
```
martinthomson: I'm not a big fan of these very-specific error codes.  A "you screwed up" code is sufficient for these cases.

draft-ietf-quic-http.md
```diff
@@ -432,17 +409,76 @@ either by identifying the stream that carries a request or by using a Push ID
...
+be treated as a connection error of type HTTP_CLOSED_CRITICAL_STREAM.
```
martinthomson: Have you considered creating a special stream for "settings", that is closed when all settings-related information is complete?  That would allow for sending MAX_PUSH_ID there with some certainty about status.  You could then reserve the "control" stream type for PRIORITY, MAX_PUSH_ID, CANCEL_PUSH, and GOAWAY frames.

draft-ietf-quic-http.md
```diff
@@ -432,17 +409,76 @@ either by identifying the stream that carries a request or by using a Push ID
...
+
+Since most connection-level concerns will be managed by QUIC, the primary use of
+the control streams will be for the SETTINGS frame when the connection opens and
+for PRIORITY frames subsequently.
```
martinthomson: and push-related control frames, and GOAWAY

draft-ietf-quic-http.md
```diff
@@ -432,17 +409,76 @@ either by identifying the stream that carries a request or by using a Push ID
...
+
+HTTP/QUIC server push is similar to what is described in {{!RFC7540}}, but uses
 different mechanisms. During connection establishment, the client enables server
 push by sending a MAX_PUSH_ID frame (see {{frame-max-push-id}}). A server cannot
-use server push until it receives a MAX_PUSH_ID frame.
+use server push until it receives a MAX_PUSH_ID frame.  Only servers can push;
+if a server receives a client-initiated push stream, this MUST be treated as a
+stream error of type HTTP_PUSH_LIMIT_EXCEEDED.
+
+A push stream is indicated by a stream type of `0x50`, followed by the Push ID
```
martinthomson: Care to share your rationale for picking these values?

draft-ietf-quic-http.md
```diff
@@ -1457,23 +1488,55 @@ Specification:
...
+| HTTP_CLOSED_CRITICAL_STREAM         | 0x0F       | Critical stream was closed               | {{http-error-codes}}   |
+| ----------------------------------- | ---------- | ---------------------------------------- | ---------------------- |
+
+## Stream Types {#iana-stream-types}
+
+This document establishes a registry for HTTP/QUIC unidirectional stream types.
+The "HTTP/QUIC Stream Type" registry manages an 8-bit space.  The "HTTP/QUIC
+Stream Type" registry operates under either of the "IETF Review" or "IESG
+Approval" policies {{?RFC8126}} for values between 0x00 and 0xef, with values
+between 0xf0 and 0xff being reserved for Experimental Use.
```
martinthomson: inclusive or exclusive?  (You might need to fix you copied too.)

draft-ietf-quic-qpack.md
```diff
@@ -179,17 +179,23 @@ x ...
 QPACK instructions occur in three locations, each of which uses a separate
 instruction space:
 
- - Table updates are carried by a unidirectional stream from encoder to decoder.
-   Instructions on this stream modify the dynamic table state without generating
-   output to any particular request.
+ - Table updates are carried by a unidirectional stream of type `0x48` from
+   encoder to decoder. Instructions on this stream modify the dynamic table
+   state without generating output to any particular request.
```
martinthomson: You name this "encoder stream" in two places later, but don't establish the name here where it is defined.  "An encoder stream is a unidirectional stream that carries table updates..." and "An acknowledgment stream is a unidirectional stream that carries acknowledgments..."

Is it better as "table update stream"?

draft-ietf-quic-http.md
```diff
@@ -249,27 +249,16 @@ the HTTP framing layer. A QUIC receiver buffers and orders received STREAM
...
-on QUIC stream 4, with subsequent requests on stream 8, 12, and so on.
-
-Server push uses server-initiated, unidirectional streams.  Thus, the server's
-first push consumes stream 7 and subsequent pushes use stream 11, 15, and so on.
+QUIC reserves the first client-initiated, bidirectional stream (Stream 0) for
+cryptographic operations. All remaining client-initiated bidirectional streams
+are used for requests and responses.  A bidirectional stream ensures that the
+response can be readily correlated with the request. This means that the
+client's first request occurs on QUIC stream 4, with subsequent requests on
+stream 8, 12, and so on. HTTP/QUIC does not use server-initiated bidirectional
```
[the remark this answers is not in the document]
MikeBishop: I don't really like the comma in that construction, since it's describing which flavor of Ydirectional stream rather than two independent descriptors of the stream.  So I fixed it by removing the other commas.  😉

"HTTP/QUIC" is used 60 times in this document, "HTTP over QUIC" is used 5.  I think of them equivalently to "HTTP/2" and "HTTP version 2" in RFC7540.  I'll do a pass separately to see if it's appropriate to trim any other "over"s.

draft-ietf-quic-http.md
MikeBishop: Reasonable suggestion; let's maybe do that as a second stage, though?

draft-ietf-quic-http.md
MikeBishop: True, the control stream has expanded again since this text was first written elsewhere.  Let's just drop that paragraph.

draft-ietf-quic-http.md
MikeBishop: If we think a table and 0x00, 0x01, etc. is better, I can do that easily enough.  However, 0x00 in particular seems like a first byte you could get reasonably by accident.  I think non-sequential usage will make it somewhat less likely that bugs/experiments will collide, so I selected these via ASCII values.

Control streams are ASCII "C", push is ASCII "P", and header compression uses "H" and "h".  It has a nice side-effect of viewing the stream data in a UI that renders the bytes into ASCII/UTF-8.

draft-ietf-quic-http.md
martinthomson: Of course :)

draft-ietf-quic-http.md
martinthomson: Just wondering.  I don't do ASCII conversion in my head, so the numbers looked a little odd.  There is no real reason to pick numbers sequentially.

draft-ietf-quic-qpack.md
MikeBishop: Not specific to this PR, but happy to address separately.

draft-ietf-quic-http.md
```diff
@@ -432,54 +409,99 @@ either by identifying the stream that carries a request or by using a Push ID
...
+###  Control Streams
+
+The control stream is indicated by a stream type of `0x43`.  Data on this stream
+consists of HTTP frames, as defined in {{frames}}.
+
+Each side MUST initiate a single control stream at the beginning of the
+connection and send its SETTINGS frame as the first frame on this stream.  Only
+one control stream per peer is permitted; receipt of a second stream which
+claims to be a control stream MUST be treated as a connection error of type
+HTTP_TOO_CONTROLLING.  If the control stream is closed at any point, this MUST
```
afrind: lol.  HTTP_CONTROL_FREAK

draft-ietf-quic-http.md
```diff
@@ -432,54 +409,99 @@ either by identifying the stream that carries a request or by using a Push ID
...
+recipient, the stream SHOULD be treated as a stream error of type
+HTTP_UNKNOWN_STREAM_TYPE.  The semantics of the remainder of the stream are
+unknown. Implementations SHOULD NOT send stream types the peer is not already
+known to support, since a stream error MAY be promoted to a connection error at
+the peer's discretion.
 
-HTTP/QUIC supports server push in a similar manner to {{!RFC7540}}, but uses
+###  Control Streams
+
+The control stream is indicated by a stream type of `0x43`.  Data on this stream
```
afrind: Maybe add "(ASCII 'C')", etc - otherwise everyone is going to ask.

draft-ietf-quic-qpack.md
```diff
@@ -179,18 +179,25 @@ x ...
 QPACK instructions occur in three locations, each of which uses a separate
 instruction space:
 
- - Table updates are carried by a unidirectional stream from encoder to decoder.
-   Instructions on this stream modify the dynamic table state without generating
-   output to any particular request.
- - Acknowledgements of table modifications and header processing are carried by
-   a unidirectional stream from decoder to encoder.
+ - The encoder stream is a unidirectional stream of type `0x48` which carries
```
afrind: H and D, not E and D?

MikeBishop: H and h, actually, for header compression.

draft-ietf-quic-http.md
MikeBishop: It now is, GitHub just doesn't see the diff that way.

LPardue:
Martin makes a good point, I kind of agree but thinking this through for something else I ask the question "what makes this an extension to HTTP"? To elaborate, if all we are doing is reserving some of QUICs streams to sit independently alongside HTTP request/response, then what of HTTP does this extend?

I don't think an answer is hard, nor required but this is yet another variable to consider (and there is not much guidance) on how to select an extension format.

LPardue:
To be less vague, consider how WebSockets could be implemented in QUIC with this unidirectional stream approach.

MikeBishop:
My thought had been that things which the sender has in hand all at once should be frames, while things which develop over long stretches should be a stream.  Of course, that's all in the eye of the beholder, but it would argue for an ALTSVC frame but a QPACK stream.

I think the thing this extends of HTTP is that the HTTP implementation owns the full set of QUIC streams.  If an HTTP plugin chooses to pipe a stream off to some secondary function, that's its prerogative.

martinthomson:
I like Mike's view here: if it's discrete and finite always, use a frame.  They are easier to use.

If it might run on continuously, use a stream.

Other factors might change that though: ordering relative to other activity often forces things into being frames.  And head of line blocking concerns might push towards a stream.  Even a small, discrete message might need to sit in a receive buffer until some condition is fulfilled, which you don't want to have blocking an important stream.

MikeBishop:
I'm inclined to follow this PR with a separate editorial one expounding on when you might want to use each.  However, this one is parked for the coming week or two, so I might create that one before this one lands.

LPardue:
Thinking through this further, I think unidirectional streams *need* to be typed, otherwise you severely limit the use of non-reserved stream IDs. 

Without this change, the early streams are reserved for control and QPACK. All other server-initated unidirectional streams effectively end up belonging to Server Push. The first byte will attempt to be parsed as a Push ID, and probably pass! This prohibits my unbound server push extension, which would like to send PUSH_PROMISE frames on server-initiated streams.

martinthomson:
That argument might be used to require typing of bidirectional streams.  What do you think of that @LPardue ?

LPardue:
For hq is think bidirectional type is not required. The prime interaction pattern is request/response and bidi suits that well. 

H2 sets a precedent for reserving a bidi stream for other uses via the CONNECT method, the WebSocket extension is an example in practice.

So, a hq extension that needs 2-way comms could either use a pair of unidi streams, or emulate/enhance CONNECT.

MikeBishop:
I agree with the general premise that if we want to retain the flexibility to do something other than push with these streams in this version, there has to be a type indicator.

MikeBishop:
Critical flaw with this proposal:  If extra unidirectional stream types are no longer a breaking protocol change, the "rickroll" variant described in "Draft Version Guidance" is no longer a valid example of an incompatible experiment.  ;-)

martinthomson:
Did we resolve this on the list?  It seemed like the objections were addressed, but I'm biased on this one.

MikeBishop:
@LPardue, I had said I would come back to this and add greasing values here too.  However, given that an unknown stream type is defined to be an error (and implementations are cautioned that peers can promote these errors to connection errors), sending grease values and mandating that receivers not treat them any differently will be a bit fragile.

I think to do that, we'd need to either prohibit unknown unidirectional streams being promoted, which seems an unfortunate exception case, or special-case the grease types, which rather defeats the purpose.

As a result, I'm merging this without grease and will open an issue to consider greasing later.

LPardue:
@MikeBishop I agree with the dilemma. Now that this change has landed, let me update my unbound server push extension specification and implementation so we can have something tangible for any new ticket. 

Many thanks!

janaiyengar:
I'm not thrilled about this change... perhaps this is worth discussing in Montreal once more, if only to make sure that this doesn't go in silently.

LPardue:
Well since we are necromancing...

I have implemented this in an experimental branch of https://github.com/bbc/nghq, along with the extension and it's specification.

PR #1366, opened 2018-05-21
Adds subheadings for post-base instructions

martinthomson:
These are different enough that they deserve a heading.  And some fixups on the diagrams.


afrind:
Or subhead one more level since the text is really subordinate to the preceding instruction?

martinthomson:
Note that h4 (four '#') is already below the waterline for our ToC.  h3 should be the goal.

PR #1369, opened 2018-05-22
Reword linkability on migration text

### LINKED ISSUE #1364: Connection migration spec is hard to understand
In section 6.8.5.  Privacy Implications of Connection Migration, the base transport spec says:
~~~
An endpoint that to break linkability upon changing networks MUST use a previously unused connection ID provided by its peer. Protection of packet numbers ensures that packet numbers cannot be used to correlate connections. Other properties of packets, such as timing and size, might be used to correlate activity, but no explicit correlation can be used to link activity across paths.
~~~
"That to break" is a bit awkward. Something must have been left out in the editing process...

martinthomson:
Closes #1364.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1809,9 +1809,9 @@ in the order in which they are numbered.
 Upon changing networks an endpoint MUST use a previously unused connection ID
 provided by its peer.  This eliminates the use of the connection ID for linking
 activity from the same connection on different networks.  Protection of packet
-numbers ensures that packet numbers cannot be used to correlate activity.  Other
-properties of packets, such as timing and size, might be used to correlate
-activity, but these changes ensure that the packet header cannot be used.
+numbers ensures that packet numbers cannot be used to correlate activity.
+This does not prevent other properties of packets, such as timing and size, 
+from being used to correlate activity.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1806,11 +1806,12 @@ support this, multiple NEW_CONNECTION_ID messages are needed.  Each
...
-previously unused connection ID provided by its peer.  Protection of packet
-numbers ensures that packet numbers cannot be used to correlate connections.
-Other properties of packets, such as timing and size, might be used to correlate
-activity, but no explicit correlation can be used to link activity across paths.
+Upon changing networks an endpoint MUST use a previously unused connection ID
+provided by its peer.  This eliminates the use of the connection ID for linking
+activity from the same connection on different networks.  Protection of packet
+numbers ensures that packet numbers cannot be used to correlate activity.  Other
+properties of packets, such as timing and size, might be used to correlate
+activity, but these changes ensure that the packet header cannot be used.
```
mikkelfj: "but these changes" is a bit vague - how about "but using a path unique connection ID enures ..."

huitema: Looks good to me, but I agree with @Mikkelfj, "but these changes" is loose text and wont stand well after another change or three.

PR #1372, opened 2018-05-22
Proposal for adding ECN support to QUIC.

gloinul:
This pull request contains an initial adaptation of what was in the design teams Wiki, but with more explanatory glue, and modifications to places where ACK or ACK_ECN frames can be used for example. 

Recovery draft changes are only pseudo code level so far and need a bit more work.


IngJohEricsson:
As regards to the pseudo code. 
Is there any formalized syntax for identification of frame types 
Now we have 

if (ack is of type ACK_ECN):

guess that is perhaps not correct syntax or ?
Other that I believe that input from other is appreciated.
/Ingemar

martinthomson:
@gloinul, thanks for putting this together.  I have opened a pull request on your repo that you should check and merge.  I've fixed typos and the build-related problems.  Once we're done with that, we can talk about the other editorial comments I have for you.

draft-ietf-quic-transport.md
```diff
@@ -2849,6 +2997,79 @@ by a client in protected packets, because it is certain that the server is able
...
+
+ECT_1:
+
+: Initial value = 0, incremented when a packet marked ECT(1) is received
+
+CE:
+
+: Initial value = 0, incremented when a packet marked CE is received
+
+Reception of duplicate packets SHOULD NOT increment the counters.
```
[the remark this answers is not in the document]
larseggert: Agree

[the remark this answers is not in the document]
gloinul: Disagree, see my answer on issue #1405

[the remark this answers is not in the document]
larseggert: ah, yes

draft-ietf-quic-transport.md
```diff
@@ -1410,6 +1416,71 @@ a single packet.
...
+larger that the amount of ECT marked packets that have been transmitted.
+
+### Continous Verification of ECN {#ecn-continous-verification}
+
+If the ECN capabiity check was successful and the endpoint continus to send ECT
+marked packets then continous verification is applied. This is to detect any
+cases when ECN field is bleached, that is, zeroed out by a network node, likely
+as the result of a routing changes since the ECN capability check.
+
+For each ACK_ECN frame that is received, the total number of ACKed packets are
```
ianswett: Suggestion: "For each received ACK_ECN frame, the total number of newly acknowledged packets can be compared to the total increase in ECN counters.  If the increase in ECN counters is larger, then an ECN failure has occurred and ECN should be disabled."

draft-ietf-quic-transport.md
```diff
@@ -2816,6 +2960,10 @@ ACK Block (repeated):
 
 ### Sending ACK Frames
 
+ACK Frames MAY be sent when all packets to be acknowledge had an IP header with
```
ianswett: acknowledged

draft-ietf-quic-transport.md
```diff
@@ -2816,6 +2960,10 @@ ACK Block (repeated):
...
+the ECN field marked as Not-ECT. If any packets where marked as ECT or ECN-CE
```
ianswett: where -> were

draft-ietf-quic-transport.md
ianswett: Also, I think leading with this paragraph here is a bit confusing.  I would prefer not adding anything here about ACK_ECN and placing text in the ACK_ECN section that must be used when new marks are received.

draft-ietf-quic-transport.md
```diff
@@ -4196,6 +4416,23 @@ limit mitigates the effect of the stream commitment attack.  However, setting
 the limit too low could affect performance when applications expect to open
 large number of streams.
 
+## Explicit Congestion Notification Attacks
+
+The ECN bits {{!RFC3168}} are an unauthenticated signal from the network. An
+on-path attacker may manipulate the value of the field. Thus, affecting the
```
ianswett: I think this reads more clearly as one sentence.   ie: "An on-path attacker may manipulate the value of the field, affecting the congestion avoidance behavior of the sender."

draft-ietf-quic-transport.md
```diff
@@ -4196,6 +4416,23 @@ limit mitigates the effect of the stream commitment attack.  However, setting
...
+congestion avoidance behavior of the sender. By clearing any CE marks the
+connection can help drive a bottle neck queue into a loss regime. By setting
+the ECN field to CE marking it can drive down the senders congestion window
+thus resulting in reduced throughput. The later could equally be accomplished
+by dropping packets for the connection. Section 18 and 19 of {{!RFC3168}}
+discusses the effects of undesired manipulation of the ECN field in more
+details.
+
+If a receiver would not have packet duplication detection and not discard any
```
ianswett: How about "If a receiver does not discard duplicate packets, an off-path attacker can retransmit packets with ECN bits set and manipulate the senders congestion avoidance state."

draft-ietf-quic-transport.md
```diff
@@ -4196,6 +4416,23 @@ limit mitigates the effect of the stream commitment attack.  However, setting
...
+duplicates an off-path attacker that can receive copies of the connection's
+packets can manipulate the senders congestion avoidance state. If packet
```
ianswett: packet duplicates -> duplicate packets

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
kazuho: #1439 proposes a different scheme that removes the requirement on the receiver-side to detect duplicates.

draft-ietf-quic-transport.md
```diff
@@ -1061,20 +1063,21 @@ negotiation, which is the same no matter which reserved version was sent.
 A server MAY therefore send different reserved version numbers in the Version
 Negotiation Packet and in its transport parameters.
 
-A client MAY send a packet using a reserved version number.  This can be used to
-solicit a list of supported versions from a server.
+A client MAY send a packet using a reserved version number. This can be
```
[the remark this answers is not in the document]
gloinul: I have reverted those that aren't necessary to make lint happy.

draft-ietf-quic-transport.md
gloinul: Yes, except that larger needs to be less in this case.

draft-ietf-quic-transport.md
gloinul: Okay, I removed this paragraph, and moved the MAY statement to the ACK_ECN frame, the MUST statement was already present.

draft-ietf-quic-recovery.md
```diff
@@ -417,9 +417,9 @@ to accelerate loss recovery.  The receiver SHOULD send an immediate ACK
 when it receives a new packet which is not one greater than the
 largest received packet number.
 
-Also, reception of an packet marked as ECN Congestion Experience
+Similarly, reception of an packet marked as ECN Congestion Experience
 (ECN-CE) SHOULD be acknowledged immediately to quicker react to
-congesiton events. Additional ECN-CE marks received during the same
+congestion events. Additional ECN-CE marks received during the same
```
ianswett: I don't think you need to say "during the same recovery period" since we're not proscribing any receiver side behavior corresponding to whether the marks are received within the same recovery period.

draft-ietf-quic-recovery.md
```diff
@@ -904,9 +904,9 @@ experimentation in using other response functions as discussed in
 
 The ACK_ECN frame defined in {{QUIC-TRANSPORT}} does not provide
 information on which of the newly acknowledged packets that
```
ianswett: nit: remove "that"

draft-ietf-quic-recovery.md
```diff
@@ -1085,19 +1084,24 @@ acked_packet from sent_packets.
...
-ack_ce_cntr.
+previous_ecn_ce_ctr.
 
 ~~~
-   OnPacketsMarked(ce_counter):
-     if (ce_counter > ack_ce_cntr):
-       // update ack_ce_cntr
-       ack_ce_cntr = ce_counter
-     if (!InRecovery(largest_acked_packet)):
+   CongestionEvent(packet_number):
```
ianswett: I'd prefer OnCongestionEvent to be consistent with the other names.

draft-ietf-quic-transport.md
```diff
@@ -1442,16 +1441,15 @@ endpoint, an ACK_ECN frame is transmitted back. This ACK_ECN frame indicates how
...
-larger than the number of ACKed packets then the path is ECN capable. This
-capability check thus has verified that one direction of the path between the
-peers is free from issues with ECN bleaching and that the application does not
-experience problems with access to the ECN field in the IP header. In this case
-the marking of the generated packets with ECT continues.
+When ACK_ECN is received, the counter values can be used to confirm the expected
+number of marks were received. ECT marked packets can become remarked as CE
+along the path between the peers. The ACK_ECN counters are added together and
+compared to total number of ACKed packets, if the sum of the counters are equal
+too or larger than the number of ACKed packets then the path is ECN capable.
```
ianswett: too -> to, or just remove the too entirely

draft-ietf-quic-transport.md
```diff
@@ -2882,6 +3020,83 @@ by a client in protected packets, because it is certain that the server is able
...
+|              Number of ECN-CE marked packets (i)            ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+~~~
+{: #ECN-BLOCK-FIG title="ECN Block"}
+
+
+### ECN counters
+
+The receiver side should implement three 64-bit counters that are copied to the
+ECN block when an ACK_ECN frame is generated:
```
martinthomson: Rather than phrase this as "implement these counters" just define the fields in the frame.

gloinul:
Note, I have just added a paragraph for an issue related to persistent ACK loss for some packets. Such events causes an insensitiveness in the continuous verification equal to the number of packets that the receiver seen with ECN markings and never acknowledged in the sender. Please review this late addition.

gloinul:
Based on the discussion in #1405 and the text that is in master branch towards next version requiring frame types to be idempotent I will add text so that the ACK_ECN frame will have this property even if packet duplication occurs. I will do so in what I consider a straight forward and simple way. I will require that only the first received ECN field for an packet number is added to the counters. This should be straightforward as the receiver's ACK data structure should allow one to determine if a packet is older than anything retained, new or duplicate.

mirjak:
Actually if possible, CE should never be ignored because a network node setting CE does not know if a packet is a duplicate or not...

ianswett:
@mirjak  In general, QUIC doesn't process duplicate packets and QUIC doesn't ACK anything it can't process, so ignoring duplicate marked packets matches with this model.

I suspect a lot more marked packets will be dropped because they were undecryptable(ie: 0RTT) than because of duplication, since duplication is very rare in my experience, and when it does occur, it'd commonly be better to ignore the second mark anyway, since the marking would occur before the duplication.

draft-ietf-quic-transport.md
```diff
@@ -2887,6 +3033,95 @@ by a client in protected packets, because it is certain that the server is able
...
 
+## ACK_ECN Frame {#frame-ack-ecn}
+
+A QUIC connection MUST keep counters for each ECN codepoint, recording
+the number of packets that were received with the corresponding ECN
+codepoint in the IP header. If the header is not readable from the
+application, the codepoint 00 (Not-ECT) MUST be assumed. If any packet
+are duplicated by the network then only the value of the ECN field of
+the packet copy first received SHALL be included in the counters, the ECN
+field value for a duplicate SHALL be ignored. This to prevent the
```
kazuho: Clarification question: considering the use of "SHALL," does this imply that an endpoint MUST detect all duplicates?

I understand that it is possible to implement such a scheme; an endpoint will keep track of the cut-off PN (the biggest PN that it would ignore) as well as all the PNs that it have seen above the cut-off PN. But I wonder if the intent here is to require all implementations to have such defense.

gloinul: Yes, that is the intention, to avoid getting ECN field values for duplicates added to the counters for packets that are included in any ACK. So if a receiver has included it in an ACK, it need to remember that this PN already have an ECN field valued added to the counters. But it doesn't need to detect duplicates for packet that are discarded directly after decryption, i.e. older than the cut-off PN.   

This is straightforward and fulfills the requirements that appears to currently exist. If #1405 concludes it would be easier to determine if there are other approaches including other ECN field value encodings as discussed in #1439.

kazuho: Thank you for the clarification. My two cents go to that this is an unnecessary burden, because IMO while idempotency is required for frame payloads, it is not required for ECN signal that is injected by an on-path device (even though we need some defense against injection attacks).

mirjak: This is incorrect. The requirements that currently exists in tsv is that ALL CEs must be counted because the network does not know if a packet is a duplicate or not!

ianswett: I believe the reason TCP counts all CEs is because TCP retransmits packets, not because of an effort to count packets the network inadvertently duplicated.  QUIC does not duplicate packets and instead uses new packet numbes for retransmissions, so I believe this embodies the intent of the requirement.

mirjak: No, that is not true. The network marks a packet as congested because it is congested and want to signal this to the endpoint. The network does not distinguish between retransmission or duplicates or anything because all these packets need the same amount of network resources and therefore cause congestion. RFC8085 even says 
"A receiver that enables the use of ECN for a UDP port MUST check
      the ECN field at the receiver for each UDP datagram that it
      receives on this port." 
and 
"This MUST report the
      presence of datagrams received with a CE-mark by providing a
      mechanism to feed this congestion information back to the sending
      application."
You cannot just ignore an CE because it's on a retransmission or a duplicated packet!

The problem with duplicates it that a packet could have been duplicated before or after it was CE marked. If it gets duplicated after it was marked, this is not a real congestion indication and if counted overstates the congestion. However, if it gets marked after it was duplicated, it is an indication of congestion. As you don't know when it got marked, you need to be conservative and account for it.

mirjak: Also QUIC does process duplicate packets to the point where it can detect that it is duplicated which probably already gives you some confidence that it is a valid quic packet and therefore you can also check the ECN IP mark at this point of time.

ianswett: Fair enough, then I think that text wasn't written with QUIC in mind.

As I mentioned previously, I think this is an edge case that is not a practical problem in real networks.  Unnecessary packet duplication is the real problem, not it's implications on ECN.

mirjak: I agree that this is an edge case and should hopefully not be a problem but saying in the spec that duplicated shall be ignored is not the right thing to do.

janaiyengar: Let's back up just a bit.

First, QUIC receivers are far less likely to receive duplicates because as @ianswett pointed out, retransmissions are not duplicates. As @mirjak notes, this doesn't mean that the network won't do stupid things, but that really is the exception, not the rule. Duplicates in QUIC are necessarily network duplication. This is a significant difference in terms of *how much* this problem matters. I think suppression is fine, since the tradeoff is between enabling an on-the-side performance attack vs. an infinitesimally more aggressive response to a CE marked packet.

Second, I agree with @kazuho that requiring duplicate detection should not be required, if only because I can cook up conditions under which an endpoint won't be able to do this without infinite state.  Consider an endpoint that nacks a packet (sends an ack with this packet in a gap), receives an ack for this nack, and moves its "left edge" past that PN. Now, it receives a packet with that PN. It cannot verify that this packet is a duplicate, since it may never have received the original.

I'd like to point out that this is ultimately a performance attack, and as with any other such attacks that use unauthenticated bits (ICMP?), we can provide some guidance to the receiver of a duplicate. I think the text here can changed a bit, with little to no real difference in impact.

I'll propose the following text: "The ECN counters that an endpoint reports include values of the ECN field only from packets that are subsequently newly acknowledged.  Note that an endpoint may newly acknowledge a duplicate packet if it no longer has state about the first receipt of the same packet. An endpoint may decrypt and drop a duplicate packet if it has state about the first receipt of the same packet, and in this case, the endpoint will not increase the ECN counters based on the duplicate packet."

martinthomson: The problem with that text is that it assumes something quite dangerous: that a receiver might accept a packet when it isn't sure that it has received it before.  And content might be different.

What @ianswett suggested before is better: if you move the trailing edge forward, then packets that you believe to be behind that edge are dropped unconditionally.  Or, more concisely stated: you only accept packets that you know you haven't accepted before.  That errs on the side of dropping rather than accepting, which I believe to be safe.

kazuho: > The problem with that text is that it assumes something quite dangerous: that a receiver might accept a packet when it isn't sure that it has received it before. And content might be different.

I am not sure if the content could be different. I'd assume that the only case in which it could be different is when the sender reuses a packet number. I do not think that we need to detect that (and if we do, it should be a PROTOCOL_VIOLATION rather than a drop).

draft-ietf-quic-transport.md
```diff
@@ -4200,6 +4434,20 @@ limit mitigates the effect of the stream commitment attack.  However, setting
 the limit too low could affect performance when applications expect to open
 large number of streams.
 
+## Explicit Congestion Notification Attacks {#security-ecn}
+
+An on-path attacker may manipulate the value of the field, affecting the
+congestion avoidance behavior of the sender. Removing any ECN-CE marking causes
+senders to maintain or increase their sending rate beyond that the path can
+sustain, which will eventually result in loss. Adding an ECN-CE marking causes
+senders to reduce their sending rate. The later could equally be accomplished by
```
martinthomson: latter

draft-ietf-quic-transport.md
```diff
@@ -4200,6 +4434,20 @@ limit mitigates the effect of the stream commitment attack.  However, setting
...
+dropping packets for the connection. Section 18 and 19 of {{!RFC3168}} discusses
+the effects of undesired manipulation of the ECN field in more details.
```
martinthomson: detail

draft-ietf-quic-transport.md
```diff
@@ -4200,6 +4434,20 @@ limit mitigates the effect of the stream commitment attack.  However, setting
...
+
+If a receiver does not discard duplicate packets, an off-path attacker can
+retransmit packets with ECN bits set and manipulate the senders congestion
+avoidance state. If duplicate packets are dropped, the off-path attacker will
```
martinthomson: s/dropped/discarded

draft-ietf-quic-transport.md
gloinul: @martinthomson it was not my intention to require infinite state. I will see if I can reformulate to say that for any packet within the window of PNs that you will accept to process at all you need to maintain duplication detection before adding the ECN field values to the counters. 

@mirjak TSVWG has asked if there was anything from the QUIC ECN work that was of general nature and should be discussed in TSVWG and I think this is one of those things. However, my opinion is that due to ECN's unauthenticated nature, it is reasonable to suppress network duplicates. Otherwise use of ECN makes one more open to attacks to drive down the congestion window for a connection than when not using it. Yes, the connection will be slightly less responsive to the congestion signal, how much is depending on the amount of network duplication that occur.

draft-ietf-quic-transport.md
```diff
@@ -1427,6 +1432,77 @@ a single packet.
...
+non-destructive congestion notification by a network node.  That is, packets are
+marked instead of being discarded by routers and other devices along a network
+path. QUIC endpoints perform capability verification on connection establishment
+as well as connection migration. ECN can be used unidirectionally,
+bidirectionally, or disabled depending on endpoint and path capabilities.
+
+The capability check makes use of the ACK_ECN frame in {{frame-ack-ecn}}. Each
+endpoint individually performs an ECN capability check for its send path. The
+check is performed by attempting to set the ECN bits in the IP header to either
+ECT(0) or ECT(1) for packets generated by the endpoint, clients starting with
```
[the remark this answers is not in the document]
gloinul: I think this proposal is very good, and agree about the IP packet thing and Stream 0 change. The only thing I 100% happy is the explicit mention of ECT(0) being the default.

draft-ietf-quic-transport.md
```diff
@@ -1427,6 +1432,77 @@ a single packet.
...
+packets that was marked as ECT in the sender, or if the comparision of the
+counters with the total of acknolwedged packets indicates that not all packets
+arrive as ECT(0), ECT(1) or ECN-CE, then the path is determined as not ECN
+Capable.  If the endpoint determine the path as currently not ECN capable it
+SHALL stop marking the packets as ECT, and instead mark them as Not-ECT. If the
+connection is migrated then the ECN capability check is rerun as specified in
+{{ecn-connection-migration}}.
+
+
+### Continuous Verification of ECN {#ecn-continuous-verification}
```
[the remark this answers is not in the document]
gloinul: If one is interested in seeing how this two counter proposal looks, please look at #1460.

draft-ietf-quic-transport.md
```diff
@@ -2887,6 +2964,94 @@ by a client in protected packets, because it is certain that the server is able
...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                     Largest Acknowledged (i)                ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                          ACK Delay (i)                      ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                       ACK Block Count (i)                   ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                          ACK Blocks (*)                     ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                           ECN Block                         ...
```
ianswett: As discussed on the list, I think it makes sense to but the ACK Block Count and ACK blocks at the end of this frame, to make it simpler to truncate the ack blocks if necessary.

gloinul: Agree, if no ones protests, I will implement that change tomorrow.

gloinul: Now implemented

draft-ietf-quic-transport.md
```diff
@@ -2999,10 +2999,10 @@ appended at the end.
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                       ACK Block Count (i)                   ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
ianswett: Can you move the ACK Block Count down as well.  It's odd to have that up here.

ianswett: Also, how about putting the ECN info in this diagram, rather than having it in a separate diagram?

gloinul: Yes, I merged the ECN counters and text into a single section for the ACK_ECN frame. I think that is very reasonable. So please check this latest change.

draft-ietf-quic-transport.md
```diff
@@ -2887,6 +2964,73 @@ by a client in protected packets, because it is certain that the server is able
 to decipher the packet.
 
 
+## ACK_ECN Frame {#frame-ack-ecn}
+
+A QUIC connection MUST keep counters for each ECN codepoint, recording the
+number of packets that were received with the corresponding ECN codepoint in the
+IP header. If the header is not readable from the application, the codepoint 00
+(Not-ECT) MUST be assumed. If any packet are duplicated by the network then only
```
larseggert: s/packet/packets/

larseggert: s/by the network/by the network,/

draft-ietf-quic-transport.md
```diff
@@ -2887,6 +2964,73 @@ by a client in protected packets, because it is certain that the server is able
...
+the value of the ECN field of the packet copy first received SHALL be included
+in the counters. This to prevent the on-the-side attack ({{security-ecn}}) and
+ensure that ACK_ECN frames becomes idempotent in the event of packet
```
larseggert: s/becomes/are/

draft-ietf-quic-transport.md
```diff
@@ -2887,6 +2964,73 @@ by a client in protected packets, because it is certain that the server is able
...
+duplication. Note, a receiver is not required to maintain indefinite state for
+which packet numbers have been received far into the history. Packets discarded
+for this reason, their ECN values are also not counted.
```
larseggert: These sentences read awkwardly. Maybe "when information about old packets is discarded at a receiver, any ECN information associated with them is also discarded" or something like that?

draft-ietf-quic-transport.md
```diff
@@ -2887,6 +2964,73 @@ by a client in protected packets, because it is certain that the server is able
...
+
+ACK_ECN Frame MUST be used when when an endpoint is acknowledging a packet were
```
larseggert: s/ACK_ECN Frame/An ACK_ECN frame/

There are other instances where "Frame" is capitalized; they should also be lowercased IMO for consistency with the rest of the docs.

draft-ietf-quic-transport.md
```diff
@@ -1427,6 +1431,63 @@ a single packet.
 
 In TLS, the Retry packet type is used to carry the HelloRetryRequest message.
 
+## ECN Capability Check {#ecn-capability-check}
+
+Explicit Congestion Notification (ECN) {{!RFC3168}} is feature that allows for
```
larseggert: s/allows for/enables/

draft-ietf-quic-transport.md
```diff
@@ -1427,6 +1431,63 @@ a single packet.
...
+endpoint records this path as not being ECN capable and it ceases marking of
+packets.
+
+IP packets sent on a new path SHOULD be marked with ECT(0) or ECT(1) to verify
+that the new path supports ECN, see {{ecn-connection-migration}}.
+
+### Continuous Verification of ECN {#ecn-continuous-verification}
+
+If the ECN capability check was successful and the endpoint continues to send
+ECT marked packets then continuous verification is applied. This is to detect
```
larseggert: s/packets then/packets, then/

draft-ietf-quic-transport.md
```diff
@@ -1427,6 +1431,63 @@ a single packet.
...
+any cases when ECN field is bleached, that is, zeroed out by a network node,
```
larseggert: Above, the text used the term "remove markings". We should pick one term - zeroed, bleached, removed - consistently.

draft-ietf-quic-transport.md
```diff
@@ -1427,6 +1431,63 @@ a single packet.
...
+likely as the result of a routing changes since the ECN capability check.
+
+For each received ACK_ECN frame, the total number of newly acknowledged packets
+can be compared to the total increase in ECN counters. If the increase in ECN
+counters is less, then an ECN failure has occurred and ECN should be disabled.
```
larseggert: less than what?

draft-ietf-quic-transport.md
```diff
@@ -1427,6 +1431,63 @@ a single packet.
...
+ECN is also disabled in case an ACK frame is received acknowledging any ECT sent
+packet.
+
+If the acknowledgements from the receiver are lost such that one or more packet
```
larseggert: s/packet/packets/

draft-ietf-quic-transport.md
```diff
@@ -1427,6 +1431,63 @@ a single packet.
...
+are received by the receiver, but never acknowledged to the sender an
+insensitivity to bleaching will be created. In this situation the ECN counters
+reported will have increase, but the sender side total for acknowledged packets
```
larseggert: s/increase/increased/

draft-ietf-quic-transport.md
```diff
@@ -1427,6 +1431,63 @@ a single packet.
...
+will not have increased. Thus, a number of bleached packets equal to the number
+of packets that failed to be acknowledged can be received before triggering the
+continuous verification. To address this issue the sender detect the case when
+the ECN counters grows more than number of acknowledged packets when a ACK_ECN
+frame is received. In such cases a new comparison point is created by storing
+the current number of totally acknowledged packets and latest ECN counters. Then
+comparison are done by subtracting these stored values from the respective
+counters prior to the comparison. Note that any out-of-order ACK_ECN frames
+can't be used for determining any loss of acknowledgements.
```
larseggert: This whole paragraph is really difficult to understand. "Comparison points" are not mentioned anywhere else, for starters.

draft-ietf-quic-transport.md
```diff
@@ -1783,6 +1845,21 @@ Note that receipt of packets with higher packet numbers from the legitimate peer
...
+### ECN Capability Check for Migrated Connection {#ecn-connection-migration}
+
+Each new path is probed to determine whether it supports ECN. Packets sent on
+the new path are sent in IP packets with an ECT marking as described in
+{{ecn-capability-check}}.
+
+Markings, or absence of markings, on packets sent on multiple paths can make it
+difficult to correctly attribute counters with markings on specific packets.
+Recording the packet number when connection migration occurred might help in
+correlating increases in counters with packets sent on the new path.
```
larseggert: Might help how?

draft-ietf-quic-transport.md
```diff
@@ -2887,6 +2964,73 @@ by a client in protected packets, because it is certain that the server is able
...
+|                        ECN-CE Count (i)                     ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                       ACK Block Count (i)                   ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                          ACK Blocks (*)                     ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+~~~
+{: #ACN_ECN_FRAME_FORMAT title="ACK_ECN Frame Format"}
+
+The receiver side report three ECN counters in the ACK_ECN frame. These counters
```
larseggert: s/side //

draft-ietf-quic-transport.md
```diff
@@ -2887,6 +2964,73 @@ by a client in protected packets, because it is certain that the server is able
...
+counts the number of packets marked with this codepoint since the start of the
```
larseggert: s/counts/count/

janaiyengar:
@martinthomson : I assume you're planning to merge my PR on top of this?

martinthomson:
Your PR went into master at the same time.  This closed when I merged your changes in.

PR #1379, opened 2018-05-23
Use tilde instead of backticks, again

martinthomson:
Unless you read the HTML, this doesn't appear to be a problem.


MikeBishop:
I prefer the HTML output the backticks generate, frankly.  But in the text version, it's wrapped in quote marks, which isn't really what we want.  So I agree we need this change, but I wouldn't mind if the template styled the tilde blocks (`<samp>`) more like it does the backtick blocks (`<pre>`).

martinthomson:
Yeah, the problem is that kramdown doesn't see ``` as special.  It reads it as `<code></code><code>`, which is an inline elements, which causes the mess you see.

PR #1399, opened 2018-05-30
Implicit acknowledgment via header block acknowledgment

### LINKED ISSUE #1370: Implicitly acknowledging table updates by acknowleding header blocks
When you acknowledge a header block, you also implicitly acknowledge the largest reference used in that header block.  You can't acknowledge it otherwise.

Should we require that an encoder track this information so that a decoder can reduce the number of acknowledgments that it sends?  It's more complicated to do that, but it can save some bytes and the associated churn.  I haven't implemented delayed acknowledgments for QPACK, but this would possibly make that more effective.

martinthomson:
I'm sure that this could be better, but here you go.

Closes #1370.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -508,7 +508,7 @@ already reached the decoder's limit for blocked streams.
 | 1 |     Insert Count (7+)     |
 +---+---------------------------+
 ~~~~~~~~~~
-{:#fig-size-sync title="Table Size Synchronize"}
+{:#fig-size-sync title="Table State Synchronize"}
 
 ### Header Acknowledgement
 
@@ -534,7 +534,7 @@ blocks within a stream have been fully processed.
 
 An encoder MUST treat receipt of a Header Acknowledgment as also acknowledging
 any dynamic table entries that the header block referenced.  That is, this
-instruction is also processed as a Table Size Synchronize instruction with a
+instruction is also processed as a Table State Synchronize instruction with a
 value matching the Largest Reference of the corresponding header block.
 
 A decoder MUST track increases to the largest acknowledged dynamic table entry
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -491,7 +491,14 @@ instruction begins with the '1' one-bit pattern. The instruction specifies the
 total number of dynamic table inserts and duplications since the last Table
 State Synchronize, encoded as a 7-bit prefix integer.  The encoder uses this
 value to determine which table entries are vulnerable to head-of-line blocking.
-A decoder MAY coalesce multiple synchronization updates into a single update.
+
+A decoder MAY coalesce multiple synchronization updates into a single update.  A
+decoder MAY rely on Header Acknowledgement instructions to indirectly
+acknowledge changes to the dynamic table.  Relying on Header Acknowledgment
+instructions exclusively leads to blocking if the encoder waits for an entry to
+be acknowledged before using it.  This happens if the encoder wants to avoid the
+risk of blocking at the decoder, or the decoder sets the
+SETTINGS_QPACK_BLOCKED_STREAMS lower than the number of active streams.
 
 ~~~~~~~~~~ drawing
   0   1   2   3   4   5   6   7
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -523,6 +530,12 @@ blocks within a stream have been fully processed.
 ~~~~~~~~~~
 {:#fig-header-ack title="Header Acknowledgement"}
 
+An encoder MUST treat receipt of a Header Acknowledgment as also acknowledging
+any dynamic table entries that the header block referenced.  That is, this
+instruction is also processed as a Table Size Synchronize instruction with a
+value matching the Largest Reference of the corresponding header block.
```
MikeBishop: Not quite, because TSS is a count of the inserts received since the last TSS, instead of an absolute index.

martinthomson: Damn, really?  That's more efficient, I agree, but I completely missed that.  (My fault, not the spec.)  Since these are on the same stream, I think that we can make TSS incremental and avoid having two counters.

Now I have to fix my code...

draft-ietf-quic-qpack.md
```diff
@@ -491,7 +491,14 @@ instruction begins with the '1' one-bit pattern. The instruction specifies the
 total number of dynamic table inserts and duplications since the last Table
 State Synchronize, encoded as a 7-bit prefix integer.  The encoder uses this
 value to determine which table entries are vulnerable to head-of-line blocking.
-A decoder MAY coalesce multiple synchronization updates into a single update.
+
+A decoder MAY coalesce multiple synchronization updates into a single update.  A
+decoder MAY rely on Header Acknowledgement instructions to indirectly
+acknowledge changes to the dynamic table.  Relying on Header Acknowledgment
+instructions exclusively leads to blocking if the encoder waits for an entry to
```
afrind: I don't think it would lead to blocking, but it would lead to never using the dynamic table?

draft-ietf-quic-qpack.md
afrind: You mean TSS will be incremental from the largest acknowledged largest reference or most recent TSS?

draft-ietf-quic-qpack.md
```diff
@@ -491,7 +491,14 @@ instruction begins with the '1' one-bit pattern. The instruction specifies the
...
+be acknowledged before using it.  This happens if the encoder wants to avoid the
+risk of blocking at the decoder, or the decoder sets the
+SETTINGS_QPACK_BLOCKED_STREAMS lower than the number of active streams.
```
afrind: I had to read this many times to figure out what you are saying - you are explaining why the encoder might be waiting for acknowledgement before using the headers in the dynamic table.  I think it might be more clear if you change the second bit to something like

"... or the encoder has already reached the decoder's limit for blocked streams".

draft-ietf-quic-qpack.md
martinthomson: Yeah.  That way you maintain a single value for largest acknowledged entry and that is updated by TSS and header acknowledgment.  It's simpler and ensures correctness more directly.

draft-ietf-quic-qpack.md
```diff
@@ -491,7 +491,14 @@ instruction begins with the '1' one-bit pattern. The instruction specifies the
 total number of dynamic table inserts and duplications since the last Table
 State Synchronize, encoded as a 7-bit prefix integer.  The encoder uses this
 value to determine which table entries are vulnerable to head-of-line blocking.
-A decoder MAY coalesce multiple synchronization updates into a single update.
+
+A decoder MAY coalesce multiple synchronization updates into a single update.  A
+decoder MAY rely on Header Acknowledgement instructions to indirectly
```
MikeBishop: If the meaning of TSS is changing, then it's no longer a MAY for the decoder to let Header Acknowledgement instructions acknowledge changes.  The decoder MUST track the encoder's perspective of what has been acknowledged, and SHOULD send TSS in addition if entries have been added that aren't referenced.

MikeBishop:
I think this discussion no longer belongs inside the TSS instruction.  If the tracking of decoder state involves the encoder combining information from HA and TSS, and the decoder has to track the encoder's inferences from HA to generate TSS, this is a complex enough system to deserve its own subsection.

Then TSS simply says that it advances the greatest acknowledged index by however many entries, and references that subsection.

martinthomson:
OK, that's not going to happen before I leave.  Let's continue next week.

draft-ietf-quic-qpack.md
```diff
@@ -489,9 +489,18 @@ After processing a set of instructions on the encoder stream, the decoder will
 emit a Table State Synchronize instruction on the decoder stream.  The
 instruction begins with the '1' one-bit pattern. The instruction specifies the
 total number of dynamic table inserts and duplications since the last Table
-State Synchronize, encoded as a 7-bit prefix integer.  The encoder uses this
-value to determine which table entries are vulnerable to head-of-line blocking.
-A decoder MAY coalesce multiple synchronization updates into a single update.
+State Synchronize or Header Acknowledgement that increased the largest
```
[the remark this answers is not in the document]
martinthomson: Yes, but I want to do that later.

PR #1400, opened 2018-05-30
Acknowledge stream resets too

### LINKED ISSUE #1371: Tracking of header blocks for reset streams
There isn't much advice in hq or qpack about tracking of header blocks, and it's one of the hardest parts to implement.  The nastiest bit is dealing with reset streams.  If a stream is reset, then all the header blocks are no longer blocked, and none of the table entries they reference need to be tracked.

However..., tracking can only stop once the peer has received the reset, otherwise the peer might read the pre-reset data and block, increasing the number of blocked streams.  So we need to know when the peer stops caring, either because they sent STOP_SENDING or received RST_STREAM.  But we don't have a signal for this.  The header block acknowledgment only applies to a single header block, not the entire stream.

It's also a timely reminder that APIs need to surface when an inbound stream is stopped (STOP_SENDING) or reset (RST_STREAM).  In either case, an HTTP receiver would need to send whatever signal we're missing here.

martinthomson:
This builds on #1399, because I realized that it would be good to point out that there is no implicit acknowledgment of header table updates here.

Closes #1371.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -508,7 +508,7 @@ already reached the decoder's limit for blocked streams.
 | 1 |     Insert Count (7+)     |
 +---+---------------------------+
 ~~~~~~~~~~
-{:#fig-size-sync title="Table Size Synchronize"}
+{:#fig-size-sync title="Table State Synchronize"}
 
 ### Header Acknowledgement
 
@@ -534,7 +534,7 @@ blocks within a stream have been fully processed.
 
 An encoder MUST treat receipt of a Header Acknowledgment as also acknowledging
 any dynamic table entries that the header block referenced.  That is, this
-instruction is also processed as a Table Size Synchronize instruction with a
+instruction is also processed as a Table State Synchronize instruction with a
 value matching the Largest Reference of the corresponding header block.
 
 A decoder MUST track increases to the largest acknowledged dynamic table entry
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -491,7 +491,14 @@ instruction begins with the '1' one-bit pattern. The instruction specifies the
 total number of dynamic table inserts and duplications since the last Table
 State Synchronize, encoded as a 7-bit prefix integer.  The encoder uses this
 value to determine which table entries are vulnerable to head-of-line blocking.
-A decoder MAY coalesce multiple synchronization updates into a single update.
+
+A decoder MAY coalesce multiple synchronization updates into a single update.  A
+decoder MAY rely on Header Acknowledgement instructions to indirectly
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -536,26 +536,29 @@ instruction is also processed as a Table Size Synchronize instruction with a
...
-### Stream Reset Acknowledgement
+### Stream Cancellation
 
 A stream that is reset might have multiple outstanding header blocks.  A decoder
-that receives a stream reset before the end of a stream generates a Stream Reset
-Acknowledgment instruction on the decoder stream.  This signals to the encoder
-that any references to the dynamic table are no longer outstanding.
+that receives a stream reset before the end of a stream generates a Stream
+Cancellation instruction on the decoder stream.  Similarly, a decoder that
+abandons reading of a stream need to signal this using the Stream Cancellation
```
MikeBishop: need => needs

draft-ietf-quic-qpack.md
```diff
@@ -536,6 +536,31 @@ instruction is also processed as a Table Size Synchronize instruction with a
...
+
+A stream that is reset might have multiple outstanding header blocks.  A decoder
+that receives a stream reset before the end of a stream generates a Stream
+Cancellation instruction on the decoder stream.  Similarly, a decoder that
+abandons reading of a stream needs to signal this using the Stream Cancellation
+instruction.  This signals to the encoder that all references to the dynamic
+table on that stream are no longer outstanding.
+
+An encoder cannot infer from this acknowledgement that any dynamic table entries
+referenced have been received.
```
afrind: I think you're trying to say that the encoder can't make any assumptions about the state of the dynamic table, as can be done with header acknowledgement.

"...cannot infer from this acknowledgement any updates to the dynamic table have been received" ?

draft-ietf-quic-qpack.md
```diff
@@ -523,6 +532,41 @@ blocks within a stream have been fully processed.
...
+### Stream Cancellation
+
+A stream that is reset might have multiple outstanding header blocks.  A decoder
+that receives a stream reset before the end of a stream generates a Stream
+Cancellation instruction on the decoder stream.  Similarly, a decoder that
+abandons reading of a stream needs to signal this using the Stream Cancellation
+instruction.  This signals to the encoder that all references to the dynamic
+table on that stream are no longer outstanding.
+
+An encoder cannot infer from this acknowledgement that any updates to the
```
afrind: s/acknowledgement/instruction/ ?

martinthomson: This builds on top of #1399 (You can view the changes here: https://github.com/quicwg/base-drafts/compare/implicit-ack-header-block...ack-reset-stream  I assumed - incorrectly - that you had already seen these changes, so pointed this at master rather than the other PR.

PR #1406, opened 2018-06-01
Clarify 'packet protection' -> 'packet number protection'

ghedo:


mikkelfj:
I'm not convinced this is a proper change. The packer number protection key in terms of label used, and length, is based on a the packet protection algorithm, not a separate isolated packet number protection algorithm.

ghedo:
@mikkelfj I'm not sure I follow, do you mean "The size of the packet number protection key is determined by the packet number protection algorithm" should instead be "The size of the packet number protection key is determined by the packet protection algorithm"? That is, remove the second "number" I added in that sentence?

draft-ietf-quic-tls.md
```diff
@@ -826,8 +826,8 @@ in Section 5.3 of {{!TLS13}}, the IV length is the larger of 8 or N_MIN (see
 Section 4 of {{!AEAD=RFC5116}}; all ciphersuites defined in {{!TLS13}} have
 N_MIN set to 12).
 
-The size of the packet protection key is determined by the packet protection
-algorithm, see {{pn-encrypt}}.
+The size of the packet number protection key is determined by the packet number
+protection algorithm, see {{pn-encrypt}}.
```
MikeBishop: The existing statement is true as well.  Perhaps say that the size of both keys depends on the negotiated AEAD?

ghedo: So, the way I read it, is that key *and IV* length (i.e. key_length and iv_length) is actually defined in the paragraph before, but nowhere it says how long is the packet number encryption key (i.e. pn_key_length) which is what confused me (also considering the "see {{pn-encrypt}}" which didn't make a whole lot of sense to me since packet number encryption wasn't mentioned at all in that sentence).

ghedo: Though I'm fine adding both, it just seems superfluous.

mikkelfj:
> @mikkelfj I'm not sure I follow, do you mean "The size of the packet number protection key is determined by the packet number protection algorithm" should instead be "The size of the packet number protection key is determined by the packet protection algorithm"? That is, remove the second "number" I added in that sentence?

Yes

ghedo:
> Yes

Ok, so I thought about that too, but "packet number protection algorithm" is explicitly defined in the pn-encrypt section which says "The packet number protection algorithm depends on the negotiated AEAD", so it kind of seemed like a a repetition that the pn protection key length is the same as protection algorithm key length *and* that pn protection algo is also the same as protection algo (that is, negotiated AEAD).

Again, don't have a strong opinion and would be happy to change, but seems superfluous.

martinthomson:
Thanks for clarifying this.  We can consider separately improving the description of the packet protection stuff, but given that it's likely to disappear very soon, let's not make extra work for ourselves.

PR #1410, opened 2018-06-03
Describe QPACK Feedback mechanisms

MikeBishop:
This feels like a huge wall of text, and I could use some help making it more readable.  However, it attempts to pull out the text from #1399 into its own section and describe the process of making sure encoder and decoder are on the same page.

Closes #1399 and closes #1400 because I have those commits in here.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -508,7 +508,7 @@ already reached the decoder's limit for blocked streams.
 | 1 |     Insert Count (7+)     |
 +---+---------------------------+
 ~~~~~~~~~~
-{:#fig-size-sync title="Table Size Synchronize"}
+{:#fig-size-sync title="Table State Synchronize"}
 
 ### Header Acknowledgement
 
@@ -534,7 +534,7 @@ blocks within a stream have been fully processed.
 
 An encoder MUST treat receipt of a Header Acknowledgment as also acknowledging
 any dynamic table entries that the header block referenced.  That is, this
-instruction is also processed as a Table Size Synchronize instruction with a
+instruction is also processed as a Table State Synchronize instruction with a
 value matching the Largest Reference of the corresponding header block.
 
 A decoder MUST track increases to the largest acknowledged dynamic table entry
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -529,14 +529,14 @@ entries might cause a stream to become blocked, as described in
 ~~~~~~~~~~
 {:#fig-size-sync title="Table State Synchronize"}
 
-A decoder SHOULD emit a Table State Synchronize after receiving new dynamic
-table entries if the most recently inserted entry was not the Largest Reference
-of a blocked stream.  Decoders which support blocking MAY delay this instruction
-briefly to see if new header blocks arrive which reference the new entries, but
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -265,6 +265,52 @@ might not actually become blocked on every stream which risks becoming blocked.
...
+
+Regardless of whether a header block contained blocking references, the
+knowledge that it was processed successfully permits the encoder to avoid
+evicting entries while references remain outstanding; see {{blocked-eviction}}.
+When a stream is reset or abandoned, the indication that these header blocks
+will never be processed serves a similar function.
+
+For the encoder to identify which dynamic table entries can be safely used
+without a stream becoming blocked, the encoder tracks the Largest Known Received
+absolute index in the decoder's state.
```
afrind: "in the decoder's state" sounds kind of weird. Maybe "the encoder tracks the absolute index of the decoder's Largest Known Received entry.

draft-ietf-quic-qpack.md
```diff
@@ -265,6 +265,52 @@ might not actually become blocked on every stream which risks becoming blocked.
...
+
+When blocking references are permitted, the encoder uses acknowledgement of
+header blocks for this purpose.  If a header block was potentially blocking, the
```
afrind: "for this purpose" reads ambiguously to me.

draft-ietf-quic-qpack.md
```diff
@@ -265,6 +265,52 @@ might not actually become blocked on every stream which risks becoming blocked.
...
+successful processing of the block by the decoder implies that all dynamic table
```
afrind: receipt of a header acknowledgement implies this

MikeBishop:
Redone on top of #1400; look at the most recent commit if you want to see just this change.  The suggestion to land the text in the appropriate instructions and just forward-reference from the intro section does substantially reduce the "wall of text," and I think it reads much better.

draft-ietf-quic-qpack.md
```diff
@@ -485,21 +513,30 @@ server's header blocks and table updates.
...
-| 1 |     Insert Count (7+)     |
-+---+---------------------------+
+| 1 | 0 |   Insert Count (6+)   |
++---+---+-----------------------+
 ~~~~~~~~~~
-{:#fig-size-sync title="Table Size Synchronize"}
+{:#fig-size-sync title="Table State Synchronize"}
+
+A decoder SHOULD emit a Table State Synchronize after receiving new dynamic
+table entries if the most recently inserted entry was not the Largest Reference
```
afrind: Since we are going to remove length prefix, 'entries' here is vague.  How many entries?  I think in the  common case the inserts will get processed before the references, but a blocking decoder would not want to send TSS without delaying to check for a referencing header block, so perhaps the SHOULD here is misleading?

Maybe it would help to describe the tradeoff -

A decoder chooses when to emit TSS instructions.  Emitting a TSS after adding each new dynamic table entry will provide the most timely feedback to the encoder, but could be redundant with other decoder feedback.  By delaying a TSS, a decoder might be able to coalesce multiple TSS instructions, or replace them entirely with Header Acknowledgements.  However, delaying too long may lead to compression inefficiencies if the encoder waits for an entry to be acknowledged before using it.

PR #1411, opened 2018-06-04
What if invariants change?

### LINKED ISSUE #1260: Invariants extensbility
If future versions of QUIC require us to add more fields to the invariants, how would we go about it? Would we add the fields after the PN or before it? We lack a field that specifies the offset to encrypted payload start and there will be no more reserved bits left if the spin bit experiment is successful.

MikeBishop:
Fixes #1260.


pravb:
Looks good to me.

Just a nit on text that this commit isn't changing.
> The primary goal of this document is to ensure that it is possible to deploy new versions of QUIC. By documenting the _things_ that can't change
It would perhaps be cleaner to say documenting the _properties_ that can't change instead.

Issue #1243, opened 2018-03-19
ICMP and ICMPv6 PMTUD with asymmetric connection-ids


enygren:
Asymmetric connection-ids introduce problems for handling ICMP and ICMPv6 PMTUD messages (such as ICMPv6 packet-too-big) in conjunction with stateless load balancers.  For example, if a server sends out a packet containing a client connection-id that is too big, a router or tunnel end-point may send back an ICMPv6 PTB message containing the packet header.  Because the PTB message with the packet header only contains the client connecton-id, it isn't possible to route to the proper server.

While ICMP/ICMPv6 PTB messages are best treated as optimizations, being able to get them can be very useful for identifying the source off loss quickly as being a too-large packet.

Note that other ICMP(v6) messages such as destination-unreachable also have a similar problem with asymmetric connection-ids.

Note that related topics are under active discussion in other working groups, such as:
* draft-bonica-intarea-frag fragile-01
* draft-ietf-tsvwg-datagram-plpmtud

enygren:
One proposal based on discussions would be to allow Long headers (with both client and server connection-ids) in certain situations following the handshake.  In particular, when performing a PMTUD probe (or otherwise sending a packet with a larger size than the current PMTU), using a Long form header would mean that a stateless load balancer would still be able to route it based on the payload of the ICMP(v6) message, plus it might make it safer and easier for the server to validate that the packet was send by a device on-path.  Both endpoints would still of course need to have reasonable lower bounds on PMTU as well.

Similarly, such a scheme could also make it more reasonable to support ICMP(v6) destination-unreachable messages in a way that wasn't vulnerable to parallel-to-path resets but could allow devices such as NATs to eventually squelch connections where state is no longer present.  For example, if the Nth and subsequent (3rd?) retransmits of the same data was sent with a full Long header (with both connection-ids) and an ICMP(v6) response is returned that contains both connection-ids (plus enough data from what was sent for validation) then this is a good sign that the node sending is both on-path and that this is a degenerate situation that should only have been reached after multiple losses and thus likely from a legitimate active middle-box.

(Note that client/server above are symmetric as both end-points have the same challenge with roles reversed.)

igorlord:
We used to allow long-header packets after handshake, but got rid of them for the lack of a use case.  This is a valid use case, so it would be good to bring back these long-header packets.

martinthomson:
[...]

PR #1412, opened 2018-06-05
Updated ICMP PMTU section

igorlord:
Per request of the editors, here is a draft PR for ICMP PMTU section.
I am deliberately not including text for #1243 here.  (That can be a subsequent PR.)

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3684,63 +3684,66 @@ which requires ICMPv6 error messages to contain "as much of invoking packet as
 possible without the ICMPv6 packet exceeding the minimum IPv6 MTU", and
 {{!RFC1812}}, which states that ICMPv4 error messages "SHOULD contain as much of
 the original datagram as possible without the length of the ICMP datagram
-exceeding 576 bytes".  ICMP messages without an on-path prooff are sent in
+exceeding 576 bytes".  ICMPv4 messages without an on-path proof are sent in
 accordance with the minimum requirments of {{!ICMP=RFC0792}} and contain fewer
 octets past the IP header (typically only 8 octets).
 
 The minimum required validation of ICMP messages with an on-path proof invoves
-verifying that the message was sent by this endpoint with at least 1-2^32
-probability and it is still outstanding (not acknowledged and not deemed lost).
-If a QUIC endpoint does not perform this minimum validation, it SHOULD treat the
-packet as an ICMP message without an on-path proof.
+verifying that the message was sent by a network element that has observed a
+QUIC packet that is still outstanding (not acknowledged and not deemed lost).
+To perform the on-path validation, the sender would need to remember at least 8
+octets from the encrypted portion of outstanding QUIC packets.  Alternatively,
+the sender would need to remember a 64-bit hash of the first 64 octets of the
+outstanding QUIC packets.  If a QUIC endpoint does not perform this minimum
+validation, it SHOULD treat the packet as an ICMP message without an on-path
+proof.
 
 As noted in {{?RFC5927}}, using ICMP messages without an on-path proof exposes
 the protocol implementation to off-path attacks and requires mitigations.
 
-Even ICMP messages without an on-path proof SHOULD undergo some validation, such
-as:
+ICMP messages without an on-path proof SHOULD undergo as much validation as
+possible.  For example, such validation could include storing IP ID fields of
+sent datagrams to validate that received ICMP messages are refering to
+outstanding packets.
 
-* Set the IPv4 Don't Fragment (DF) bit on a small proportion of packets, so that
-  most invalid ICMP messages arrive when there are no DF packets outstanding,
-  and can therefore be identified as spurious.
+Any ICMP message that fails validation MUST be discarded.
 
-* Store IP ID field of the sent datagrams to validate that ICMP message is
-  refering to an outstanding packet.
+Any reduction in Path MTU due to an ICMP Packet Too Big (PTB) message is
+provisional until QUIC's loss detection algorithm determines that the packet is
+actually lost.  ICMP PTB messages are ICMPv4 type 3 code 4 ({{!ICMP}}) packets
+and ICMPv6 type 2 ({{!ICMPv6}}) packets.
 
-Any ICMP messages that fail validation MUST be discarded.
+#### ICMP PTB During Handshake  {#icmp-ptb-during-handshake}
 
 It is important that any problems are detected quickly during the connection
 handshake, because the client may be able to mitigate them by switching to
 alternative IP addresses or protocols.  Hence, an endpoint SHOULD reduce Path
-MTU in response to an ICMP Packet Too Big (PTB) message during a handshake,
-unless then would cause a reduction to a Path MTU value smaller than 1280
-octets.
+MTU in response to an ICMP PTB message during a handshake, unless then would
+cause a reduction to a Path MTU value smaller than 1280 octets.
 
-If during a handshake a client receives an ICMP TPB message that requests it to
-reduce Path MTU to a value smaller than 1280 octets, then:
+If a client receives an ICMP PTB message that requests it to reduce Path MTU to
+a value smaller than 1280 octets during a handshake, then:
 
-* If the client has another IP address for the server to try, the client should
+* If the client has another IP address for the server to try, the client SHOULD
   restart the connection to another IP.
 
 * Otherwise, if the client can fail over to another protocol, and the ICMP
-  packet has on-path validation, the client should retry connecting with another
+  packet has on-path validation, the client SHOULD retry connecting with another
   protocol.
+
+#### ICMP PTB After Handshake   {#icmp-ptb-after-handshake}
 
 If an ICMP PTB message is received after handshake, and the claimed Path MTU is
 at least 1280 octets for messages with on-path validation or 1392 for messages
 without on-path validations, the Path MTU SHOULD be set accordingly.  Otherwise,
 Path MTU probing of {{!PLPMTUD}} SHOULD be attempted.
 
-Any reduction in Path MTU due to a report contained in an ICMP message is
-provisional until QUIC's loss detection algorithm determines that the packet is
-actually lost.
-
 
 ### Special Considerations for Packetization Layer PMTU Discovery
 
 The PING frame provides a useful option for PMTU probe packets. PING
 frames generate acknowledgements, but they need not be delivered reliably. As a
-result, the loss of PING frames in probe packets does not require
+result, the loss of PING frames in probe packets does not require a
 delay-inducing retransmission. However, PING frames do consume congestion
 window, which may delay the transmission of subsequent application data.
 
@@ -3754,16 +3757,23 @@ increases in the size of probe packets. As QUIC probe packets need not contain
 application data, aggressive increases in probe size carry fewer consequences.
 
 
-## Responding to ICMP "Unreachable" messages {#icmp-unreach}
+## Responding to ICMP Unreachable Messages {#icmp-unreach}
 
-When a QUIC endpoint receives an ICMP "Unreachable" message during a handshake,
-the response SHOULD be identical to receiving an ICMP TPB message that announces
-a Path MTU smaller than 1280 octets (see {{icmp-pmtu}}).
+When a QUIC endpoint receives an ICMP Unreachable message (other than PTB
+message) during a handshake, the response SHOULD be identical to receiving an
+ICMP PTB message that announces a Path MTU smaller than 1280 octets during a
+handshake (see {{icmp-ptb-during-handshake}}).
 
-When an ICMP "Unreachable" message is received after the handshake, the QUIC
-endpoint should send a PATH_CHALLENGE frame ({{frame-path-challenge}}).  Sending
-PATH_CHALLENGE frames on the same path due to ICMP "Unreachable" messages should
-be rate limited.
+When an ICMP Unreachable message (other than PTB message) is received after the
+handshake, the server SHOULD ignore this messages, since this message could be
+due to a NAT losing its binding or due to an attack.
+
+The client, however, SHOULD attempt to verify this ICMP message by sending
+additional packets containing PING frames ({{frame-ping}}) as well as possibly
+other frames.  If the client receives no QUIC packets from the server for the
+duration of multiple round-trip times after receiving an ICMP Unreachable, the
+client SHOULD deem the connection dead and attempt to reconnect (preferably to a
+different server IP, if one is available).
 
 
 # Streams: QUIC's Data Structuring Abstraction {#streams}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3649,20 +3649,19 @@ header, protected payload, and any authentication fields.
 
 All QUIC packets SHOULD be sized to fit within the estimated PMTU to avoid IP
 fragmentation or packet drops. To optimize bandwidth efficiency, endpoints
-SHOULD use Packetization Layer PMTU Discovery ({{!PLPMTUD=RFC4821}}).  Endpoints
-MAY use classical PMTU Discovery ({{!PMTUDv4=RFC1191}}, {{!PMTUDv6=RFC8201}})
-for detecting the PMTU, setting the PMTU appropriately, and storing the result
-of previous PMTU determinations (see {{icmp-pmtu}}).
+SHOULD use Packetization Layer PMTU Discovery ({{!PLPMTUD=RFC4821}}), which is
+an extension to the classical ICMP-based PMTU Discovery ({{!PMTUDv4=RFC1191}},
+{{!PMTUDv6=RFC8201}}).
 
-In the absence of these mechanisms, QUIC endpoints SHOULD NOT send IP packets
-larger than 1280 octets. Assuming the minimum IP header size, this results in
-a QUIC packet size of 1232 octets for IPv6 and 1252 octets for IPv4. Some
-QUIC implementations MAY wish to be more conservative in computing allowed
-QUIC packet size given unknown tunneling overheads or IP header options.
+In the absence of PMTU discovery mechanisms, QUIC endpoints SHOULD NOT send IP
+packets larger than 1280 octets. Assuming the minimum IP header size, this
+results in a QUIC packet size of 1232 octets for IPv6 and 1252 octets for
+IPv4. Some QUIC implementations MAY wish to be more conservative in computing
+allowed QUIC packet size given unknown tunneling overheads or IP header options.
 
-QUIC endpoints that implement any kind of PMTU discovery SHOULD maintain an
-estimate for each combination of local and remote IP addresses.  Each pairing of
-local and remote addresses could have a different maximum MTU in the path.
+QUIC endpoints that implement PMTU discovery SHOULD maintain an estimate for
+each combination of local and remote IP addresses.  Each pairing of local and
+remote addresses could have a different maximum MTU in the path.
 
 QUIC depends on the network path supporting a MTU of at least 1280 octets. This
 is the IPv6 minimum MTU and therefore also supported by most modern IPv4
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3182,40 +3182,73 @@ QUIC packets on the affected path.  This could result in termination of the
...
+If a QUIC endpoint does not perform this minimum validation, it SHOULD treat the
+packet as an ICMP message without an on-path proof.
+
+As noted in {{?RFC5927}}, using ICMP messages without an on-path proof exposes
+the protocol implementation to off-path attacks and requires mitigations.
+
+Even ICMP messages without an on-path proof SHOULD undergo some validation, such
+as:
 
 * Set the IPv4 Don't Fragment (DF) bit on a small proportion of packets, so that
```
ianswett: I thought we were recommending setting DF on all packets?

mikkelfj: I also wondered ...

igorlord: That was one of the original proposals.  But I agree -- this is not the best idea.  I'll revise.

draft-ietf-quic-transport.md
```diff
@@ -3228,6 +3261,18 @@ increases in the size of probe packets. As QUIC probe packets need not contain
...
+## Responding to ICMP "Unreachable" messages {#icmp-unreach}
+
+When a QUIC endpoint receives an ICMP "Unreachable" message during a handshake,
+the response SHOULD be identical to receiving an ICMP TPB message that announces
+a Path MTU smaller than 1280 octets (see {{icmp-pmtu}}).
+
+When an ICMP "Unreachable" message is received after the handshake, the QUIC
+endpoint should send a PATH_CHALLENGE frame ({{frame-path-challenge}}).  Sending
+PATH_CHALLENGE frames on the same path due to ICMP "Unreachable" messages should
+be rate limited.
```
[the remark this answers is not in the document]
igorlord: I'll rewrite this completely.

draft-ietf-quic-transport.md
```diff
@@ -3157,9 +3157,9 @@ header, protected payload, and any authentication fields.
 All QUIC packets SHOULD be sized to fit within the estimated PMTU to avoid IP
 fragmentation or packet drops. To optimize bandwidth efficiency, endpoints
 SHOULD use Packetization Layer PMTU Discovery ({{!PLPMTUD=RFC4821}}).  Endpoints
-MAY use PMTU Discovery ({{!PMTUDv4=RFC1191}}, {{!PMTUDv6=RFC8201}}) for
-detecting the PMTU, setting the PMTU appropriately, and storing the result of
-previous PMTU determinations.
+MAY use classical PMTU Discovery ({{!PMTUDv4=RFC1191}}, {{!PMTUDv6=RFC8201}})
```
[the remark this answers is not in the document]
igorlord: How strongly do you feel against word "classical"?  It is used in other RFCs for the purpose I am using it here.  See https://tools.ietf.org/html/rfc4821 and https://tools.ietf.org/html/rfc5320.

draft-ietf-quic-transport.md
```diff
@@ -3182,40 +3182,73 @@ QUIC packets on the affected path.  This could result in termination of the
...
+the original datagram as possible without the length of the ICMP datagram
+exceeding 576 bytes".  ICMP messages without an on-path prooff are sent in
+accordance with the minimum requirments of {{!ICMP=RFC0792}} and contain fewer
+octets past the IP header (typically only 8 octets).
 
-As a result, endpoints that implement PMTUD in IPv4 SHOULD take steps to
-mitigate this risk. For instance, an application could:
+The minimum required validation of ICMP messages with an on-path proof invoves
+verifying that the message was sent by this endpoint with at least 1-2^32
+probability and it is still outstanding (not acknowledged and not deemed lost).
```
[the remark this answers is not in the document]
igorlord: You are right.  This needs revising.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
martinduke: How about "ICMP-based"? This eliminates all ambiguity and saves people the trouble of clicking through references to know what we're talking about.

draft-ietf-quic-transport.md
```diff
@@ -3182,40 +3187,71 @@ QUIC packets on the affected path.  This could result in termination of the
...
-Traditional ICMP-based path MTU discovery in IPv4 {{!PMTUDv4}} is potentially
-vulnerable to off-path attacks that successfully guess the IP/port 4-tuple and
-reduce the MTU to a bandwidth-inefficient value. TCP connections mitigate this
-risk by using the (at minimum) 8 bytes of transport header echoed in the ICMP
-message to validate the TCP sequence number as valid for the current
-connection. However, as QUIC operates over UDP, in IPv4 the echoed information
-could consist only of the IP and UDP headers, which usually has insufficient
-entropy to mitigate off-path attacks.
+ICMP error messages used in classical Path MTU discovery may be classified into
+messages with an on-path proof and without an on-path proof.  ICMP messages with
```
martinduke: I disagree with the statement that some ICMP messages have no proof.  The 8-byte trailer is sufficient as a proof for TCP, and even if the QUIC case there is enough entropy if we store the right things.

draft-ietf-quic-transport.md
```diff
@@ -3182,40 +3182,73 @@ QUIC packets on the affected path.  This could result in termination of the
...
+
+It is important that any problems are detected quickly during the connection
+handshake, because the client may be able to mitigate them by switching to
+alternative IP addresses or protocols.  Hence, an endpoint SHOULD reduce Path
+MTU in response to an ICMP Packet Too Big (PTB) message during a handshake,
+unless then would cause a reduction to a Path MTU value smaller than 1280
+octets.
+
+If during a handshake a client receives an ICMP TPB message that requests it to
+reduce Path MTU to a value smaller than 1280 octets, then:
```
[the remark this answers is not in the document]
igorlord: This is specifically done during the handshake.  The language here is trying to stay in agreement with the decision that PMTU < 1280 does not support QUIC.  Which ones of the statements is the concern:
1. to switch IPs (but stick w/ QUIC), if multiple IPs are available for the server?
2. to switch protocols, if we have an on-path validation of the signal?

Would you rather see all PMTU<1280 signals ignored (and not switch to TCP, if available, for example)?

draft-ietf-quic-transport.md
```diff
@@ -3182,40 +3187,71 @@ QUIC packets on the affected path.  This could result in termination of the
...
-example, the IP ID or UDP checksum) to further authenticate incoming Datagram
-Too Big messages.
+ICMP messages without an on-path proof SHOULD undergo some validation, such as
+storing IP ID fields of sent datagrams to validate that received ICMP messages
+are refering to outstanding packets.
 
-* Any reduction in PMTU due to a report contained in an ICMP packet is
-provisional until QUIC's loss detection algorithm determines that the packet is
-actually lost.
+Any ICMP messages that fail validation MUST be discarded.
```
martinduke: The problem with MUSTs in this section is that some user space implementations may not have any control over ICMP handling logic.

draft-ietf-quic-transport.md
```diff
@@ -3182,40 +3187,71 @@ QUIC packets on the affected path.  This could result in termination of the
...
 
+Any reduction in Path MTU due to a report contained in an ICMP Packet Too Big
+message (PTB) provisional until QUIC's loss detection algorithm determines that
```
martinduke: ... SHOULD be provisional...

draft-ietf-quic-transport.md
```diff
@@ -3182,40 +3187,71 @@ QUIC packets on the affected path.  This could result in termination of the
...
+It is important that any problems are detected quickly during the connection
+handshake, because the client may be able to mitigate them by switching to
+alternative IP addresses or protocols.  Hence, an endpoint SHOULD reduce Path
+MTU in response to an ICMP PTB message during a handshake, unless then would
+cause a reduction to a Path MTU value smaller than 1280 octets.
+
+If a client receives an ICMP PTB message that requests it to reduce Path MTU to
+a value smaller than 1280 octets during a handshake, then:
+
+* If the client has another IP address for the server to try, the client should
```
martinduke: SHOULD

draft-ietf-quic-transport.md
```diff
@@ -3182,40 +3187,71 @@ QUIC packets on the affected path.  This could result in termination of the
...
+  restart the connection to another IP.
 
+* Otherwise, if the client can fail over to another protocol, and the ICMP
+  packet has on-path validation, the client should retry connecting with another
```
martinduke: SHOULD

draft-ietf-quic-transport.md
```diff
@@ -3182,40 +3187,71 @@ QUIC packets on the affected path.  This could result in termination of the
...
-actually lost.
+Any ICMP messages that fail validation MUST be discarded.
 
+Any reduction in Path MTU due to a report contained in an ICMP Packet Too Big
+message (PTB) provisional until QUIC's loss detection algorithm determines that
+the packet is actually lost.  ICMP PTB messages are ICMPv4 type 3 code 4
+({{!ICMP}}) and ICMPv6 type 2 ({{!ICMPv6}}).
 
-### Special Considerations for Packetization Layer PMTU Discovery
+#### ICMP PTB During Handshake  {#icmp-ptb-during-handshake}
```
martinduke: I like how you separate the handshake issues here, but these also apply to PLPMTUD, yes?

draft-ietf-quic-transport.md
```diff
@@ -3182,40 +3187,71 @@ QUIC packets on the affected path.  This could result in termination of the
...
+
+It is important that any problems are detected quickly during the connection
+handshake, because the client may be able to mitigate them by switching to
+alternative IP addresses or protocols.  Hence, an endpoint SHOULD reduce Path
```
martinduke: This is a significant change. The spec suggests ("SHOULD") that PLPMTUD be used. ICMP-based PMTUD is strictly a MAY.

draft-ietf-quic-transport.md
```diff
@@ -3182,40 +3187,71 @@ QUIC packets on the affected path.  This could result in termination of the
...
+* Otherwise, if the client can fail over to another protocol, and the ICMP
+  packet has on-path validation, the client should retry connecting with another
+  protocol.
 
-The PADDING frame provides a useful option for PMTU probe packets. PADDING
+#### ICMP PTB After Handshake   {#icmp-ptb-after-handshake}
+
+If an ICMP PTB message is received after handshake, and the claimed Path MTU is
+at least 1280 octets for messages with on-path validation or 1392 for messages
+without on-path validations, the Path MTU SHOULD be set accordingly.  Otherwise,
```
martinduke: Same comment: we're elevating ICMP response from a MAY to a SHOULD.

draft-ietf-quic-transport.md
martinduke: Zooming out a bit, I think the original text of this section is a bit clearer about what the real problem is. I do think it could use a paragraph like "endpoints SHOULD use fields from the QUIC header, if provided in the ICMP message, to validate that a QUIC packet in flight might have generated the ICMP message." And leave the rest unchanged.

draft-ietf-quic-transport.md
```diff
@@ -3182,40 +3187,71 @@ QUIC packets on the affected path.  This could result in termination of the
...
-mitigate this risk. For instance, an application could:
+The minimum required validation of ICMP messages with an on-path proof invoves
+verifying that the message was sent by this endpoint with at least 1-2^32
+probability and it is still outstanding (not acknowledged and not deemed lost).
+If a QUIC endpoint does not perform this minimum validation, it SHOULD treat the
+packet as an ICMP message without an on-path proof.
 
-* Set the IPv4 Don't Fragment (DF) bit on a small proportion of packets, so that
-most invalid ICMP messages arrive when there are no DF packets outstanding, and
-can therefore be identified as spurious.
```
martinduke: I guess you deleted this because it didn't meet your 1/2^32 standard?

draft-ietf-quic-transport.md
```diff
@@ -3182,40 +3182,73 @@ QUIC packets on the affected path.  This could result in termination of the
...
-* Store additional information from the IP or UDP headers from DF packets (for
-example, the IP ID or UDP checksum) to further authenticate incoming Datagram
-Too Big messages.
+* Otherwise, if the client can fail over to another protocol, and the ICMP
+  packet has on-path validation, the client should retry connecting with another
+  protocol.
 
-* Any reduction in PMTU due to a report contained in an ICMP packet is
+If an ICMP PTB message is received after handshake, and the claimed Path MTU is
+at least 1280 octets for messages with on-path validation or 1392 for messages
```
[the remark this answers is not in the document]
igorlord: Somewhat arbitrary, examining various tunneling protocols out there to see what seems like a conservative-enough number that is not too small to be significantly concerned with degraded performance.  This number would certainly need a wg consensus.

The goal was to make sure that in the common case (no attack), we can switch to a workable PMTU immediately.

Of course, if the wg believes that 1280 is "large enough" to use it immediately, this would simplify things.

draft-ietf-quic-transport.md
```diff
@@ -3228,6 +3261,18 @@ increases in the size of probe packets. As QUIC probe packets need not contain
...
 
 
+## Responding to ICMP "Unreachable" messages {#icmp-unreach}
+
+When a QUIC endpoint receives an ICMP "Unreachable" message during a handshake,
+the response SHOULD be identical to receiving an ICMP TPB message that announces
+a Path MTU smaller than 1280 octets (see {{icmp-pmtu}}).
+
+When an ICMP "Unreachable" message is received after the handshake, the QUIC
+endpoint should send a PATH_CHALLENGE frame ({{frame-path-challenge}}).  Sending
```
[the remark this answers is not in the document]
igorlord: Fixing and rewriting this.

draft-ietf-quic-transport.md
igorlord: I'd say that not just the QUIC header but, more importantly, the QUIC encrypted section, since it has more entropy.

But, I agree, instead of coming up with metrics, we should just leave the specifics undefined.  I like your wording.  Will adapt and incorporate.

igorlord: (As for an 8-byte proof, it is sufficient for TCP but not QUIC.  Hence this entire discussion.)

draft-ietf-quic-transport.md
igorlord: The language in this section says that validation is a SHOULD.  But if that validation fails, discarding is a MUST.  (And if the validation is not done, since it is not mandatory, then the signal SHOULD be treated as a signal without an on-path proof.)

Do you still think this is a problem?

draft-ietf-quic-transport.md
igorlord: PLPMTUD take time to do its work (it is loss/timeout-based).  The purpose of this discussion is to identify conditions, when you can trust ICMP immediately and not have to resort to PLPMTUD (especially during handshake).

draft-ietf-quic-transport.md
igorlord: Yes, this would need to be reconfirmed.  I do believe that we should encourage implementations to use the best available signal, including ICMP, especially when we have reasons to trust it.  Performance is important for QUIC, and it is double-important for handshake, and ICMP works much faster than PLPMTUD.

draft-ietf-quic-transport.md
igorlord: LOL.  Mostly because I received comments that we recommend DF to be set on all packets, and this is going against that recommendation.

(But being pedantic, that standard was for on-path proof, while this discussion is for validation packets w/o on-path proof.)

More importantly, the entire standard thing is gone now.  I will further revise it to use your latest feedback as to the wording (ie. "use the data from the packet" instead of prescribing standards or which data to use).

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
janaiyengar: I think the context does make it clear, but @martinduke's suggestion seems reasonable.

draft-ietf-quic-transport.md
```diff
@@ -3187,40 +3187,76 @@ QUIC packets on the affected path.  This could result in termination of the
 connection if an alternative path cannot be found.
 
 
-### IPv4 PMTU Discovery {#v4-pmtud}
+### Classical ICMP-based path MTU discovery {#icmp-pmtu}
```
janaiyengar: ditto. Also, Capitalize This Section Title.

draft-ietf-quic-transport.md
```diff
@@ -3187,40 +3187,76 @@ QUIC packets on the affected path.  This could result in termination of the
...
+
+ICMP error messages used in classical Path MTU discovery may be classified into
+messages with an on-path proof and without an on-path proof.  ICMP messages with
+an on-path proof are the messages sent in accordance with {{!ICMPv6=RFC4443}},
```
janaiyengar: the messages -> messages

draft-ietf-quic-transport.md
```diff
@@ -3187,40 +3187,76 @@ QUIC packets on the affected path.  This could result in termination of the
...
+possible without the ICMPv6 packet exceeding the minimum IPv6 MTU", and
+{{!RFC1812}}, which states that ICMPv4 error messages "SHOULD contain as much of
+the original datagram as possible without the length of the ICMP datagram
+exceeding 576 bytes".  ICMPv4 messages without an on-path proof are sent in
+accordance with the minimum requirments of {{!ICMP=RFC0792}} and contain fewer
+octets past the IP header (typically only 8 octets).
+
+The minimum required validation of ICMP messages with an on-path proof invoves
+verifying that the message was sent by a network element that has observed a
+QUIC packet that is still outstanding (not acknowledged and not deemed lost).
```
janaiyengar: This sentence is a bit hard to follow. Also, it's possible that the ICMP message was received after the sender deemed it lost, since the sender may have been too earnest in marking it lost. Perhaps remove the requirement that it not be marked as lost yet? How about the following: "To validate an ICMP message containing on-path proof, the proof MUST be from a QUIC packet that is not yet acknowledged."

draft-ietf-quic-transport.md
```diff
@@ -3187,40 +3187,76 @@ QUIC packets on the affected path.  This could result in termination of the
...
+To perform the on-path validation, the sender would need to remember at least 8
```
janaiyengar: To perform the on-path validation -> To validate a received ICMP message

draft-ietf-quic-transport.md
```diff
@@ -3187,40 +3187,76 @@ QUIC packets on the affected path.  This could result in termination of the
...
+octets from the encrypted portion of outstanding QUIC packets.  Alternatively,
+the sender would need to remember a 64-bit hash of the first 64 octets of the
+outstanding QUIC packets.  If a QUIC endpoint does not perform this minimum
```
janaiyengar: the outstanding => outstanding

draft-ietf-quic-transport.md
```diff
@@ -3187,40 +3187,76 @@ QUIC packets on the affected path.  This could result in termination of the
...
+validation, it SHOULD treat the packet as an ICMP message without an on-path
```
janaiyengar: SHOULD -> MUST

draft-ietf-quic-transport.md
```diff
@@ -3187,40 +3187,76 @@ QUIC packets on the affected path.  This could result in termination of the
...
+proof.
+
+As noted in {{?RFC5927}}, using ICMP messages without an on-path proof exposes
```
janaiyengar: an on-path proof -> on-path proof

draft-ietf-quic-transport.md
```diff
@@ -3187,40 +3187,76 @@ QUIC packets on the affected path.  This could result in termination of the
...
+the protocol implementation to off-path attacks and requires mitigations.
```
janaiyengar: the protocol implementation -> an endpoint

draft-ietf-quic-transport.md
```diff
@@ -3187,40 +3187,76 @@ QUIC packets on the affected path.  This could result in termination of the
...
+
+ICMP messages without an on-path proof SHOULD undergo as much validation as
```
janaiyengar: an on-path proof -> on-path proof

draft-ietf-quic-transport.md
```diff
@@ -3187,40 +3187,76 @@ QUIC packets on the affected path.  This could result in termination of the
...
+possible.  For example, such validation could include storing IP ID fields of
+sent datagrams to validate that received ICMP messages are refering to
+outstanding packets.
```
janaiyengar: This paragraph seems quite redundant. I would replace it with one sentence with recommendation for what state a sender can store to validate proof in ICMP messages.

draft-ietf-quic-transport.md
```diff
@@ -3187,40 +3187,76 @@ QUIC packets on the affected path.  This could result in termination of the
...
-could consist only of the IP and UDP headers, which usually has insufficient
-entropy to mitigate off-path attacks.
+#### ICMP PTB During Handshake  {#icmp-ptb-during-handshake}
 
-As a result, endpoints that implement PMTUD in IPv4 SHOULD take steps to
-mitigate this risk. For instance, an application could:
+It is important that any problems are detected quickly during the connection
+handshake, because the client may be able to mitigate them by switching to
+alternative IP addresses or protocols.  Hence, an endpoint SHOULD reduce Path
+MTU in response to an ICMP PTB message during a handshake, unless then would
```
janaiyengar: then -> they

draft-ietf-quic-transport.md
igorlord: Re-reading PLPMTUD, the draft says explicitly that it is an extension to the classical (ICMP-based) PMTU discovery process and its prescriptions are to be used in cases where ICMP-based PMTU discovery cannot be done.  So say that PLPMTUD is a SHOULD is effectively saying that ICMP is a SHOULD (as much as practicable).

draft-ietf-quic-transport.md
igorlord: @janaiyengar is suggesting that not doing (or not being able to do) a validation means it is a ~~SHOULD~~ MUST to treat the signal as not having an on-path proof.  It makes sense to me.

ianswett:
@igorlord  This hasn't been updated in a while and I think PR #2036 may have covered some of the items you were working on.  Can you either update this PR or close it as being now obsolete?

igorlord:
@ianswett, yes, much has changed since.  Let's close this PR.  Instead, I've posted a simpler #2108 and #2109 that may need a short discussion.

PR #1419, opened 2018-06-05
Remove sequence number from NEW_CONNECTION_ID

### LINKED ISSUE #1280: HoLB from NCID frames
Since NEW_CONNECTION_ID frames are defined to be sequential (because the packet number gaps are cumulative), receipt of NCID frames out of order requires reassembly; the later CIDs aren't usable without the previous ones.  This is particularly pernicious in the situations described in #1276, where one side is effectively out of CIDs, while the other side believes it has many future IDs issued.  In an extreme case, the exhausted side might be unable to send packets at all, causing the other side not to receive ACKs which would permit it to repair the loss.

Possible solutions include:
- Eliminating the packet number gap entirely via #1079 
- Switching from a gap to an offset, so that CIDs need not be used sequentially

In either case, the CIDs associated with a connection would become an unordered set, rather than a list in fixed order.

martinthomson:
Closes #1280.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1817,9 +1817,7 @@ new connection IDs using the NEW_CONNECTION_ID frame.
 An endpoint might need to send packets on multiple networks without receiving
 any response from its peer.  To ensure that the endpoint is not linkable across
 each of these changes, a new connection ID is needed for each network.  To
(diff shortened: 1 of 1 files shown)


martinthomson:
Just to clarify intent here.  I tried to find text to add back, but I just can't see there being any value in it.  I could say explicitly that any connection ID can be used, but the text remaining effectively implies that.

PR #1422, opened 2018-06-06
Placeholders for PRIORITY

### LINKED ISSUE #441: fix priority grouping node ambiguity

### LINKED ISSUE #1421: Stream 0 DT proposal conflicts with PRIORITY
After the design team output, all client-initiated bidirectional streams are used for requests, including Stream 0.  Since Stream 0 will be a valid, actual client request, this creates a conflict with the use of Stream ID 0 in PRIORITY to indicate the root of the priority tree.

MikeBishop:
**DO NOT MERGE**:  Dependent on #1398.  Fixes #441, fixes #1421.

Primarily for illustrative purposes, this is what I envision the addition of placeholders to the PRIORITY scheme would look like.  Key points:

- Flags for Stream ID vs. Push ID are now a two-bit type field
- Filled in the fourth value as "root of the tree," fixing #1421, since Stream 0 is now a valid request stream
- Server has a setting to declare how many placeholders you get
- Client can use placeholders 0..N-1
- Server can prune closed streams aggressively (after 2 RTT to ensure client knows they're closed)


MikeBishop:
I'd prefer not to do that because HTTP/QUIC doesn't currently have a hard dependency on QUICv1, and the properties of the Stream ID space could easily change across versions.

martinthomson:
@MikeBishop, if you want to get this in for -13, you have a little bit of time left.

PR #1431, opened 2018-06-08
Move stream type into frame type

### LINKED ISSUE #1430: Interleaved Stream IDs are awkward
In case we hadn't noticed before, at the Kista interim it became very clear how we all trip over the N/N+4 progression of stream IDs of the same type.  It's quite awkward.  As someone said, we'll all eventually get used to it, and so will the world... unless we don't have to.

It's possible to envision the lowest two bits of the Stream ID as flags which indicate the type, leaving the "real" ID as the upper bits.  However, by encoding them in the low bits of an integer, that becomes annoying to represent and work with easily.

If they were the high-order bits, that would be more convenient, but would of course blow out the varint encoding.  (Of course, two high-order bits of stream type and a 62-bit Stream ID fit nicely into a 64-bit variable assuming that implementations want to store them together once the varint is expanded.)  If varint could start mid-byte, that would also be an option, but that makes the varint parser more annoying.

The obvious solution, if we want to bite the bullet and go there, is to move the stream type being referenced into the frame type.  This was awkward when client-initiated bidirectional stream 0 was special, but now that crypto is moving to a special stream outside the Stream ID space, I think this transition makes more sense.  (Crypto is essentially a fifth type of stream with special properties: there can be only one, and most stream management frames don't apply to it.)

Besides making Stream IDs easier to keep track of, this resolves the inconsistency between `initial_max_(uni|bidi)_streams` and MAX_STREAM_ID -- the Stream ID and the count are the same thing, if the type is specified separately.  By reclaiming two bits per stream reference, this also saves a small number of bytes on the wire -- every STREAM, STOP_SENDING, RST_STREAM, etc. becomes a byte shorter for streams 16-63, 4096-16384, etc. of each stream type.  The same benefits will accrue to HTTP PRIORITY frames, where the stream type is already known from the fact that it's a stream carrying an HTTP request.

The biggest cost comes in combination with our decision to make frame types varint:  By spending two bits of frame type to indicate length and moving stream types into the type byte as well, there are only two frame types values left before we spill into a second byte, and we'll those for CRYPTO and CRYPTO_CLOSE.  (If we add something else and something has to spill over, I'd start with things that are sent only once per connection, like the *_CLOSE frames.)

MikeBishop:
During the discussion that became #1430 in Kista, @janaiyengar asked me to write this up, just so we could see what it looks like.

As the issue mentions, this becomes more viable now that the crypto stream won't be part of any of the stream type spaces, but off on its own.  To keep this sensible, I've deleted the stream 0 carve-outs where they were in the way of this change, but it won't actually be fully consistent until #1377 lands as well.

Fixes #1430.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -887,22 +887,22 @@ explained in more detail as they are referenced later in the document.
 | Type Value  | Frame Type Name   | Definition                  |
 |:------------|:------------------|:----------------------------|
 | 0x00        | PADDING           | {{frame-padding}}           |
-| 0x01        | RST_STREAM        | {{frame-rst-stream}}        |
-| 0x02        | CONNECTION_CLOSE  | {{frame-connection-close}}  |
-| 0x03        | APPLICATION_CLOSE | {{frame-application-close}} |
-| 0x04        | MAX_DATA          | {{frame-max-data}}          |
-| 0x05        | MAX_STREAM_DATA   | {{frame-max-stream-data}}   |
-| 0x06        | MAX_STREAM_ID     | {{frame-max-stream-id}}     |
-| 0x07        | PING              | {{frame-ping}}              |
-| 0x08        | BLOCKED           | {{frame-blocked}}           |
-| 0x09        | STREAM_BLOCKED    | {{frame-stream-blocked}}    |
-| 0x0a        | STREAM_ID_BLOCKED | {{frame-stream-id-blocked}} |
-| 0x0b        | NEW_CONNECTION_ID | {{frame-new-connection-id}} |
-| 0x0c        | STOP_SENDING      | {{frame-stop-sending}}      |
-| 0x0d        | ACK               | {{frame-ack}}               |
-| 0x0e        | PATH_CHALLENGE    | {{frame-path-challenge}}    |
-| 0x0f        | PATH_RESPONSE     | {{frame-path-response}}     |
-| 0x10 - 0x17 | STREAM            | {{frame-stream}}            |
+| 0x01        | CONNECTION_CLOSE  | {{frame-connection-close}}  |
+| 0x02        | APPLICATION_CLOSE | {{frame-application-close}} |
+| 0x03        | MAX_DATA          | {{frame-max-data}}          |
+| 0x04        | PING              | {{frame-ping}}              |
+| 0x05        | BLOCKED           | {{frame-blocked}}           |
+| 0x06        | NEW_CONNECTION_ID | {{frame-new-connection-id}} |
+| 0x07        | ACK               | {{frame-ack}}               |
+| 0x08        | PATH_CHALLENGE    | {{frame-path-challenge}}    |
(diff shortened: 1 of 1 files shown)


LPardue:
Clarification: does this mean the stream intent at the QUIC layer as opposed to application layer mapping type?

mikkelfj:
> Clarification: does this mean the stream intent at the QUIC layer as opposed to application layer mapping type?

As I read it, this is clearly at the QUIC layer, and QUIC (transport) doesn't really deal much with the application layer.

It might be good with some additional formal naming conventions such as peer initiated bidi stream, locally initiated uni stream, etc. such that HTTP QUIC and friends has a meaningful language and possibly also assign a number to the stream type as a value in 0 .. 3 but keeping the actual stream id specific to the stream type at both layers.

MikeBishop:
The debugging aspect is mitigated if you stick the type in the top two bits of a 64-bit integer locally.  So you'd see 0x00000002, 0x40000002, 0x80000002, or 0xc0000002.  Or you'd output with a prefix, e.g. CU2, SB2, etc.  But you can't encode that way on the wire, or only one of the types can be less than 4 bytes.

Yeah, I thought about trying to condense it to three types per frame, where one of them means "the only unidirectional stream I could be talking about" and the interpretation varies based on who's sending it.  It's only a step further than the two-type frames that imply directionality based on the sender and only provide bi/uni in a bit.  But I feel like the cost in consistent representation outweighed saving the frame types.  We already have the language that says you need to reject a frame that talks about a stream the sender isn't allowed to talk about, so that's no worse, even if no better.

How does it ruin PRIORITY again?

martinthomson:
Having the extra space turns out to be handy: https://github.com/quicwg/base-drafts/pull/1422#pullrequestreview-127164641

ianswett:
It's interesting to see this, but I favor the current approach unless this fixes some important issue I'm unaware of.  I don't think something being awkward reaches the bar of "important issue" at this point in the process.

martinthomson:
I agree with Ian here.  What we have isn't ideal, but you can get used to it.  It would be easier if people didn't feel that it was necessary to correct someone every time they say N+1 but mean N+4, when intent is otherwise clear, but that is a small tax to pay.

FWIW, I experimented with an encoding for stream identifiers that used a 2 bit prefix and a varint that was 4/12/28/60.  The performance hit was considerable.

MikeBishop:
@martinthomson, I'm curious about the source of the performance hit.  Was it simply from using a non-byte-aligned varint?  If so, that seems separate from this.

martinthomson:
So I don't know for certain, I just had the numbers in the end.  I suspect that it was just the extra code needed to pull two lots of two bits from the first octet.  Obviously having an extra field is more work, but it was more costly than I anticipated.  I probably have the code on a branch still...

martinthomson:
@MikeBishop are we resolved on this?  (I don't have the code I wrote, sadly).

janaiyengar:
I'm fine with closing this.

PR #1437, opened 2018-06-11
Don't cite octet indices any more

### LINKED ISSUE #1436: octet 1 of Long Header
In https://github.com/quicwg/base-drafts/blob/master/draft-ietf-quic-transport.md#long-header-long-header

It says that octet 1 of the long header contains the DCIL/SCIL values, but the diagram above has the beginning of the Version in octet 1.

martinthomson:
Also reflow.

Closes #1436.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -315,15 +315,16 @@ Version:
 
 DCIL and SCIL:
 
-: Octet 1 contains the lengths of the two connection ID fields that follow it.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -443,7 +444,7 @@ Google QUIC Demultipexing Bit:
 Reserved:
 
 : The sixth, seventh, and eighth bits (0x7) of octet 0 are reserved for
```
MikeBishop: You've removed "octet 1" above, but still have "octet 0" here.

PR #1441, opened 2018-06-12
Statement of idempotency

### LINKED ISSUE #1424: Idempotency of Frame Types
In the repeated discussions of duplicated packets and what happens if an implementation accidentally processes them, it has become clear that an important property of QUIC in this respect is that every frame type is idempotent.  It doesn't matter if you receive them once or a dozen times, nor how they are reordered (only provided they're not received before they're sent, of course).  All that matters is that each reliable frame, or the information it carries, is received.

This is perhaps an important property to document, both for future iterations of the WG who are expanding QUIC, and for extension frames as they will need to have this property as well.

MikeBishop:
Fixes #1424.

Marked as editorial, since this is just a statement of what is already true about all frames anyway.  However, it's implicit that this should also be true of frames defined in the future or in extensions, so it's a bit broader than the usual editorial change.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -906,6 +906,11 @@ explained in more detail as they are referenced later in the document.
 | 0x10 - 0x17 | STREAM            | {{frame-stream}}            |
 {: #frame-types title="Frame Types"}
 
+All QUIC frames are idempotent.  That is, the state of the connection is
+unchanged regardless of whether a frame is processed one time or multiple times,
+and a frame can be received and processed without error at any point between the
+time the frame was initially sent and the end of the connection.
+
 # Life of a Connection
 
 A QUIC connection is a single conversation between two QUIC endpoints.  QUIC's
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -906,10 +906,9 @@ explained in more detail as they are referenced later in the document.
 | 0x10 - 0x17 | STREAM            | {{frame-stream}}            |
 {: #frame-types title="Frame Types"}
 
-All QUIC frames are idempotent.  That is, the state of the connection is
-unchanged regardless of whether a frame is processed one time or multiple times,
-and a frame can be received and processed without error at any point between the
-time the frame was initially sent and the end of the connection.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -906,6 +906,11 @@ explained in more detail as they are referenced later in the document.
 | 0x10 - 0x17 | STREAM            | {{frame-stream}}            |
 {: #frame-types title="Frame Types"}
 
+All QUIC frames are idempotent.  That is, the state of the connection is
+unchanged regardless of whether a frame is processed one time or multiple times,
+and a frame can be received and processed without error at any point between the
```
ianswett: Not all frames are processed without errors, so this is a bit awkward

MikeBishop: Perhaps that it never produces an error on a subsequent processing that it didn't produce the first time?

draft-ietf-quic-transport.md
```diff
@@ -906,6 +906,11 @@ explained in more detail as they are referenced later in the document.
 | 0x10 - 0x17 | STREAM            | {{frame-stream}}            |
 {: #frame-types title="Frame Types"}
 
+All QUIC frames are idempotent.  That is, the state of the connection is
+unchanged regardless of whether a frame is processed one time or multiple times,
```
martinthomson: How about: A valid frame does not change the state of the connection as a result of being received more than once.

mikkelfj: It could affect the state in terms of congestion logic and timeouts.

janaiyengar: @mikkelfj, presumably you are talking about receiving ACK frames. It shouldn't affect the state of congestion control or timeouts, since it all relies on whether an ACK frame has new information or not.

martinthomson: I think that it's packets and (specifically) ACKs that affect congestion and loss recovery state.

martinthomson: A valid frame does not cause undesirable side effects or errors when received more than once.

janaiyengar: I like this last re"framing"

mikkelfj: Idle timeout?

PR #1450, opened 2018-06-15
Stream0 dt output for merge

ekr:
@martinthomson


draft-ietf-quic-recovery.md
```diff
@@ -107,7 +111,7 @@ important to the loss detection and congestion control machinery below.
 * Retransmittable packets are those that contain at least one
   retransmittable frame.
 
-* Crypto handshake data is sent on stream 0, and uses the reliability
+* Crypto handshake data is sent in CRYPTO_HS frames, and uses the reliability
```
chris-wood: 👍

draft-ietf-quic-tls.md
```diff
@@ -255,142 +205,169 @@ A simplified TLS 1.3 handshake with 0-RTT application data is shown in
...
+
+- CRYPTO_HS frames MAY appear in packets of any encryption level.
+- CONNECTION_CLOSE and CRYPTO_CLOSE MAY appear in packets of any
+  encryption level other than 0-RTT.
+- PADDING and PING frames MAY appear in packets of any encryption level.
+- ACK frames MAY appear in packets of any encryption level, but
+  MUST only acknowledge packets which appeared in that encryption
+  level.
+- STREAM frames MUST ONLY appear in the 0-RTT and 1-RTT levels.
+- All other frame types MUST only appear at the 1-RTT levels.
```
nibanks: Would it be reasonable to allow the BLOCKED frames in 0-RTT, assuming the previous connection set very small limits in it's TP? For instance, what if the previous connection set the initial maximum number of bidirectional streams to 1, and the client would like to open 2 streams in 0-RTT. Should it be allowed to open 1 stream and include a `STREAM_ID_BLOCKED` frame?

draft-ietf-quic-tls.md
```diff
@@ -667,73 +565,85 @@ Early data within the TLS connection MUST NOT be used.  As it is for other TLS
...
-client packet protection key would use HKDF-Expand with an `info` parameter of
-0x00200851554943206b6579.
+The keys for the Initial encryption level are computed based on
+the client's first Destination Connection Id, as described in
+{{initial-secrets}}.
 
+The keys for the remaining encryption level are computed in the same
+fashion as the corresponding TLS keys (see {{TLS13}}; Section 7),
+except that the label for HKDF-Expand-Label uses the prefix "quic "
+rather than "tls 13". The purpose of this change is to provide key
```
kazuho: `"tls 13"` should be `"tls13 "` (notice the position of the whitespace)

draft-ietf-quic-tls.md
```diff
@@ -1024,144 +795,90 @@ encrypted_pn = ChaCha20(pn_key, counter, nonce, packet_number)
...
-handshake messages that are carried in unprotected packets.  However, a server
-might need to retransmit its TLS handshake messages in response to receiving an
-unprotected packet that contains ACK frames.  A server MUST process ACK frames
-in unprotected packets until the TLS handshake is reported as complete, or it
-receives an ACK frame in a protected packet that acknowledges all of its
-handshake messages.
+Receiving and verifying the TLS Finished message is critical in
+ensuring the integrity of the TLS handshake.  A server MUST NOT use
+1-RTT protected packets from the client prior to verifying the client
+Finished message if its response depends on client authentication.
```
kazuho: This paragraph seems redundant. The PR already has the following text starting from line 843,
```
However,
a server MUST NOT process data from incoming 1-RTT protected packets
before verifying either the client Finished message or - in the case that the
server has chosen to use a pre-shared key - the pre-shared key binder (see
Section 4.2.8 of {{!TLS13}}).
```

draft-ietf-quic-transport.md
```diff
@@ -1118,19 +1227,83 @@ the cryptographic handshake provides QUIC with:
...
-Details of how TLS is integrated with QUIC is provided in more detail in
-{{QUIC-TLS}}.
+The CRYPTO_HS frame can be sent in different packet number spaces.
+CRYPTO_HS frames in each packet number space carry a separate sequence
+of handshake data starting from an offset of 0.
+
+## Example Handshake Flows
+
+Details of how TLS is integrated with QUIC are provided in {{QUIC-TLS}},
+but we provide some examples here.
```
MikeBishop: Document typically avoids the first person.  "...but some examples are provided here."

draft-ietf-quic-transport.md
```diff
@@ -1415,17 +1586,10 @@ A server can process an initial cryptographic handshake messages from a client
...
-causes a client to reset its transport state and to continue the connection
-attempt with new connection state while maintaining the state of the
-cryptographic handshake.
-
-A server MUST NOT send multiple Retry packets in response to a client handshake
-packet.  Thus, any cryptographic handshake message that is sent MUST fit within
-a single packet.
-
-In TLS, the Retry packet type is used to carry the HelloRetryRequest message.
+causes a client to restart the connection attempt and includes the token in the
```
MikeBishop: "include~s~"

draft-ietf-quic-transport.md
```diff
@@ -1459,68 +1623,69 @@ To send additional data prior to completing the cryptographic handshake, the
 server then needs to validate that the client owns the address that it claims.
 
 Source address validation is therefore performed during the establishment of a
-connection.  TLS provides the tools that support the feature, but basic
-validation is performed by the core transport protocol.
+connection, by the core transport protocol.
```
MikeBishop: "performed by the core transport protocol during the establishment of a connection."

draft-ietf-quic-transport.md
```diff
@@ -1459,68 +1623,69 @@ To send additional data prior to completing the cryptographic handshake, the
...
-A different type of token is needed when resuming.  Unlike the token that is
-created during a handshake, there might be some time between when the token is
-created and when the token is subsequently used.  Thus, a resumption token
-SHOULD include an expiration time.  It is also unlikely that the client port
-number is the same on two different connections; validating the port is
+The server uses the NEW_TOKEN frame {{frame-new-token}} to provide the client
+with an address validation token that can be used to validate future 0-RTT
+connections.  The client may then use this token to validate future 0-RTT
+connections by including it in the Initial packet's header.  The client MUST
+NOT use the token provided in a RETRY for future connections.
```
MikeBishop: Title case, not all-caps.

draft-ietf-quic-transport.md
```diff
@@ -1459,68 +1623,69 @@ To send additional data prior to completing the cryptographic handshake, the
...
+
+Unlike the token that is created for a Retry packet, there might be some time
+between when the token is created and when the token is subsequently used.
+Thus, a resumption token SHOULD include an expiration time.  The server may
```
MikeBishop: MAY?  Might?  Could?

draft-ietf-quic-transport.md
```diff
@@ -2929,6 +3070,34 @@ PATH_CHALLENGE frame previously sent by the endpoint, the endpoint MAY generate
 a connection error of type UNSOLICITED_PATH_RESPONSE.
 
 
+## NEW_TOKEN frame {#frame-new-token}
+
+An server sends a NEW_TOKEN frame (type=0x21) to provide the client a token to
```
MikeBishop: "A~n~"

draft-ietf-quic-transport.md
```diff
@@ -3016,6 +3185,66 @@ advised to bundle as few streams as necessary in outgoing packets without losing
...
+end, so CRYPTO_HS frames do not have a FIN bit.
+
+
+## CRYPTO_CLOSE Frame {#frame-crypto-close}
+
+The CRYPTO_CLOSE frame (type=0x20) is used to indicate connection failures
+caused by the crypto handshake. It uses the same format as the
+CONNECTION_CLOSE frame ({{frame-connection-close}}), except that the
+error codes are specific to the crypto protocol in use. For TLS 1.3,
+the error code is simply the one-byte TLS alert number.
```
MikeBishop: The error code in the other two frames is two bytes long.  Is this frame shorter (in which case the format is not the same), or is the MSB of the error code always zero?

draft-ietf-quic-transport.md
```diff
@@ -3064,6 +3293,10 @@ been lost.  In general, information is sent again when a packet containing that
 information is determined to be lost and sending ceases when a packet
 containing that information is acknowledged.
 
+* Data sent in CRYPTO_HS frames are retransmitted according to the rules in
+  {{QUIC-RECOVERY}}, until either all data has been ACKed or the crypto state
```
MikeBishop: "acknowledged"

draft-ietf-quic-transport.md
```diff
@@ -3299,8 +3532,7 @@ The two type bits from a Stream ID therefore identify streams as summarized in
 | 0x3      | Server-Initiated, Unidirectional |
 {: #stream-id-types title="Stream ID Types"}
 
-Stream ID 0 (0x0) is a client-initiated, bidirectional stream that is used for
-the cryptographic handshake.  Stream 0 MUST NOT be used for application data.
+The first bi-directional stream opened by the client is stream 0.
```
MikeBishop: For people who don't have the Stream 0 legacy (intended audience of this doc), is this statement even relevant?

draft-ietf-quic-transport.md
```diff
@@ -3893,6 +4102,14 @@ errors is not mandatory, but only because requiring that an endpoint generate
 these errors also means that the endpoint needs to maintain the final offset
 state for closed streams, which could mean a significant state commitment.
 
+## Flow control of CRYPTO_HS data {#flow-control-crypto}
+
+Data sent in CRYPTO_HS frames is not flow controlled in the same way as STREAM
+frames.  QUIC relies on the cryptographic protocol implementation to avoid
+excessive buffering of data, see {{QUIC-TLS}}.  The implementation SHOULD
+provide an interface to QUIC to tell it about its buffering limits so that there
+is no excessive buffering at multiple layers.
```
MikeBishop: no => not

draft-ietf-quic-transport.md
```diff
@@ -4313,6 +4531,15 @@ the range from 0xFE00 to 0xFFFF.
 
 Issue and pull request numbers are listed with a leading octothorp.
 
+## Since draft-ietf-quic-transport-12
+
+- Enable server to transition connections to a preferred address (#560,#1251).
+- No more stream 0.
```
MikeBishop: Stream 0 still exists.  Maybe "Stream 0 is no longer reserved for the cryptographic handshake.  Instead, CRYPTO_HS frames are used."

draft-ietf-quic-recovery.md
```diff
@@ -130,15 +145,17 @@ number for transmissions, and any data that is to be delivered to the receiving
...
+space, and directly encodes transmission order.  A higher QUIC packet
+number signifies that the packet was sent later, and a lower QUIC
+packet number signifies that the packet was sent earlier.  When a
+packet containing frames is deemed lost, QUIC rebundles the
+retransmittable data in a new packet with a new packet number,
+removing ambiguity about which packet is acknowledged when an ACK is
+received.  Consequently, more accurate RTT measurements can be made,
+spurious retransmissions are trivially detected, and mechanisms such
+as Fast Retransmit can be applied universally, based only on packet
+number.
```
MikeBishop: The only change here seems to be that it was rewrapped to a smaller width than the rest of the document.  Is this necessary?

draft-ietf-quic-recovery.md
```diff
@@ -280,23 +298,26 @@ connection's final smoothed RTT value as the resumed connection's initial RTT.
...
 computed and the alarm SHOULD be set for twice the newly computed smoothed RTT.
 
-Handshake data may be cancelled by handshake state transitions. In particular,
-all non-protected data SHOULD no longer be transmitted once packet protection
-is available.
+#### Retry
 
-(TODO: Work this section some more. Add text on client vs. server, and on
-stateless retry.)
+A RETRY packet causes the content of the client's Initial packet to be
```
MikeBishop: Packet types are title-cased, not all-caps.  Same on line 318.

draft-ietf-quic-recovery.md
```diff
@@ -398,29 +419,23 @@ A packet sent on an RTO alarm MUST NOT be blocked by the sender's congestion
...
-but MUST NOT excessively delay acknowledgements of packets containing
-non-ack frames.  Specifically, implementaions MUST attempt to
-enforce a maximum ack delay to avoid causing the peer spurious
-timeouts.  The default maximum ack delay in QUIC is 25ms.
+There are cases where one may be able to gain recovery information from
+acknowledgements of packets in another packet number space, but they rely
+on complex assumptions about the peer’s processing and acknowledgement
+algorithms.  Even those are unable to quickly recover from cases such as
+losing the client's Initial, but receiving the 0-RTT packets.  Below is
+an optimization using coalesced packets and implicit acknowledgements.
```
MikeBishop: I don't see the optimization using coalesced packets and implicit acknowledgements.  If it's not immediately below this statement, perhaps add a forward reference?

ekr: Oh, I meant to delete this graf too

draft-ietf-quic-tls.md
```diff
@@ -255,142 +205,169 @@ A simplified TLS 1.3 handshake with 0-RTT application data is shown in
...
+{: #schematic title="QUIC and TLS Interactions"}
 
+Unlike TLS over TCP, QUIC applications which want to send data do not
+send it through TLS "application_data" records. Rather, they send it
+as QUIC STREAM frames which are then carried in QUIC packets.
+
+
+# Carrying TLS Messages {#carrying-tls}
+
+QUIC carries TLS handshake data in CRYPTO_HS frames, each of which
```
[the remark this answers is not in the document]
ekr: We also have CRYPTO_CLOSE. It's also confusing to have HANDSHAKE frames in !Handshake packets.

I would really prefer not to change this at this time. I'm happy to bikeshed this later, but it's going to be much easier to s/CRYPTO_HS/<whatever>/ than any of the more generic alternatives, and so this doesn't need to block the PR.

draft-ietf-quic-tls.md
ekr: I am going to leave this up to MT's discretion.

You should open an issue for BLOCKED.

draft-ietf-quic-tls.md
```diff
@@ -667,73 +565,85 @@ Early data within the TLS connection MUST NOT be used.  As it is for other TLS
 application data, a server MUST treat receiving early data on the TLS connection
 as a connection error of type PROTOCOL_VIOLATION.
 
+## Rejecting 0-RTT
 
-## QUIC Key Expansion {#key-expansion}
+A server rejects 0-RTT by rejecting 0-RTT at the TLS layer.  This results in
+early exporter keys being unavailable, thereby preventing the use of 0-RTT for
+QUIC.
```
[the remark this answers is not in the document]
ekr: Yes, this just got missed in my edit. I can fix.

draft-ietf-quic-tls.md
```diff
@@ -667,73 +565,85 @@ Early data within the TLS connection MUST NOT be used.  As it is for other TLS
...
-~~~
-struct {
-   uint16 length = Length;
-   opaque label<6..255> = "QUIC " + Label;
-} QhkdfExpandInfo;
-~~~
+# QUIC Packet Protection {#packet-protection}
+
+As with TLS over TCP, QUIC encrypts packets with keys derived from the TLS
+handshake, using the AEAD algorithm negotiated by TLS.
```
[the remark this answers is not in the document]
ekr: I think this is fine as-is. The TLS record layer is never used alone with TCP. But I'll leave it to MT's discretion.

draft-ietf-quic-tls.md
```diff
@@ -667,73 +565,85 @@ Early data within the TLS connection MUST NOT be used.  As it is for other TLS
...
 
-### Handshake Secrets {#handshake-secrets}
+### Initial Secrets {#initial-secrets}
 
-Packets that carry the TLS handshake (Initial, Retry, and Handshake) are
-protected with a secret derived from the Destination Connection ID field from
-the client's Initial packet.  Specifically:
+Initial packets are protected with
+a secret derived from the Destination Connection ID field from the client's
+Initial packet.  Specifically:
```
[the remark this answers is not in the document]
ekr: Yes, that's correctl

draft-ietf-quic-tls.md
```diff
@@ -1024,144 +795,90 @@ encrypted_pn = ChaCha20(pn_key, counter, nonce, packet_number)
...
-are marked with a KEY_PHASE of 0.
+A client that receives an indication that its 0-RTT data has been accepted by a
+server can send 0-RTT data until it receives all of the server's handshake
+messages.  A client SHOULD stop sending 0-RTT data if it receives an indication
+that 0-RTT data has been rejected.
 
+A server MUST NOT use 0-RTT keys to protect packets.
 
-### Retransmission and Acknowledgment of Unprotected Packets
+: 0-RTT data can be acknowledged by the server as it receives it, but any
```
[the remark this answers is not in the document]
ekr: OK

draft-ietf-quic-tls.md
```diff
@@ -667,73 +565,85 @@ Early data within the TLS connection MUST NOT be used.  As it is for other TLS
...
+   HKDF-Expand-Label(initial_secret, "client in", Hash.length)
+server_initial_secret =
+   HKDF-Expand-Label(initial_secret, "server in", Hash.length)
 ~~~
 
 The hash function for HKDF when deriving handshake secrets and keys is SHA-256
-{{!SHA=DOI.10.6028/NIST.FIPS.180-4}}.  The connection ID used with QHKDF-Expand
-is the connection ID chosen by the client.
+{{!SHA=DOI.10.6028/NIST.FIPS.180-4}}.  The connection ID used with
+HKDF-Expand-Label is the connection ID chosen by the client.
```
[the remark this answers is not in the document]
MikeBishop: Okay, then this one doesn't apply.

draft-ietf-quic-recovery.md
ianswett: I moved Crypto Handshake Data below under generating acknowledgements.

draft-ietf-quic-tls.md
```diff
@@ -255,142 +205,169 @@ A simplified TLS 1.3 handshake with 0-RTT application data is shown in
...
+and QUIC CRYPTO_HS frames is that in QUIC multiple frames may appear
+in the same QUIC packet as long as they are associated with the
+same encryption level. For instance, an implementation might
+bundle a Handshake message and an ACK for some Handshake
+data into the same packet.
+
+In general, the rules for which data can appear in packets of which
+encryption level are the same in QUIC as in TLS over TCP:
+
+- CRYPTO_HS frames MAY appear in packets of any encryption level.
```
rpaulo: I'm not sure where to point this out, but shouldn't we have a restriction saying that a QUIC packet can only have *one* CRYPTO_HS frame?

draft-ietf-quic-tls.md
```diff
@@ -255,142 +205,169 @@ A simplified TLS 1.3 handshake with 0-RTT application data is shown in
...
+  MUST only acknowledge packets which appeared in that encryption
+  level.
+- STREAM frames MUST ONLY appear in the 0-RTT and 1-RTT levels.
+- All other frame types MUST only appear at the 1-RTT levels.
+
+Because packets may be reordered on the wire, QUIC uses the packet
+type to indicate which level a given packet was encrypted
+under, as shown in {{packet-types-levels}}. When multiple packets of
+different encryption levels need to be sent, endpoints SHOULD use
+coalesced packets to send them in the same UDP datagram.
```
rpaulo: I think that, up to this point, we were relying on the fact that typical UDP stacks give us a clear per-packet boundary and this changes that.  Do we now see implementations receiving a UDP packet and assuming it might contain more than one QUIC packet?
It seems like the only way to implement this is to start parsing the QUIC packet and then, if there are leftover bytes, interpret that as another QUIC packet.
It's not too hard but it seems like a significant change.

draft-ietf-quic-tls.md
ekr: Why? That's not, for instance, how Minq works.

draft-ietf-quic-tls.md
```diff
@@ -667,73 +565,85 @@ Early data within the TLS connection MUST NOT be used.  As it is for other TLS
...
-~~~
+## TLS Errors
 
-The HKDF-Expand function used by QHKDF-Expand uses the PRF hash function
-negotiated by TLS, except for handshake secrets and keys derived from them (see
-{{handshake-secrets}}).
+If TLS experiences an error, it MUST generate an appropriate alert
+as defined in {{TLS13}}; Section 6) and then provide it to QUIC,
+which sends the alert in a CRYPTO_CLOSE frame. All such alerts are
+"fatal".
```
rpaulo: "fatal" is under quotes but its meaning isn't explained.

draft-ietf-quic-tls.md
ekr: That's already documented under coalesced packets, I think: https://tools.ietf.org/html/draft-ietf-quic-transport-12#section-4.6

draft-ietf-quic-tls.md
```diff
@@ -873,26 +673,20 @@ from the packet number.
...
-The key, K, is either the client packet protection key (client_pp_key\<i>) or
-the server packet protection key (server_pp_key\<i>), derived as defined in
-{{key-expansion}}.
-
-The nonce, N, is formed by combining the packet protection IV (either
-client_pp_iv\<i\> or server_pp_iv\<i\>) with the packet number.  The 64 bits
-of the reconstructed QUIC packet number in network byte order is left-padded
-with zeros to the size of the IV.  The exclusive OR of the padded packet number
-and the IV forms the AEAD nonce.
+The key and iv for the packet are computed as described in {{encryption-keys}}.
```
rpaulo: IV

draft-ietf-quic-tls.md
```diff
@@ -873,26 +673,20 @@ from the packet number.
...
+The nonce, N, is formed by combining the packet protection IV with the
+packet number.  The 64 bits of the reconstructed QUIC packet number in
+network byte order is left-padded with zeros to the size of the IV.
```
rpaulo: "is left-padded" or "are left-padded"?

draft-ietf-quic-tls.md
rpaulo: Because it's like having two STREAM frames for the same stream ID.

draft-ietf-quic-tls.md
rpaulo: OK, got it.

draft-ietf-quic-tls.md
```diff
@@ -1608,14 +1017,13 @@ A small ClientHello that results in a large block of handshake messages from a
...
-Certificate caching {{?RFC7924}} can reduce the size of the server's handshake
-messages significantly.
-
-QUIC requires that the packet containing a ClientHello be padded to a minimum
-size.  A server is less likely to generate a packet reflection attack if the
-data it sends is a small multiple of this size.  A server SHOULD use a
-HelloRetryRequest if the size of the handshake messages it sends is likely to
-significantly exceed the size of the packet containing the ClientHello.
+QUIC includes three defenses against this attack. First, the packet
+containing a ClientHello be padded to a minimum size. Second, if
```
rpaulo: typo? is padded to?

draft-ietf-quic-tls.md
ekr: This also seems to be permitted for STREAM frames:

```
   Stream multiplexing is achieved by interleaving STREAM frames from
   multiple streams into one or more QUIC packets.  A single QUIC packet
   can include multiple STREAM frames from one or more streams.
```

MikeBishop: There's nothing wrong with having multiple STREAM frames from the same stream, and I don't see anything wrong with having multiple CRYPTO_HS frames for the same reason.  Presumably, they contain different regions of the handshake stream generated at different points by the TLS stack.

(If they claim to contain the same region of data but differ, that might be a good reason to blow up, of course....)

rpaulo: Right, so I get the case where more than one CRYPTO_HS/STREAM frame is present but at least we should mention what to do if they overlap.

ekr: I think we should, but not limit it to the case where they are in the same packet. The situation is the same if they are in two different packets.

rpaulo: If they are in two different packets then it's just a retransmission ;-)
I'm thinking we just need some text saying the receive is supposed to detect and handle overlapping CRYPTO_HS/STREAM frame even if they are in the same packet.  Maybe we already have that, but I don't see it.

ekr: Sure. Given that CRYPTO_HS has the same semantics in this respect as STREAM, I suggest a separate PR that adds it to the stream section.

draft-ietf-quic-transport.md
```diff
@@ -3016,6 +3185,66 @@ advised to bundle as few streams as necessary in outgoing packets without losing
...
+Length:
+
+: A variable-length integer specifying the length of the Crypto Data field in
+  this CRYPTO_HS frame.
+
+Crypto Data:
+
+: The cryptographic message data.
+
+There is a separate flow of cryptographic handshake data in each
```
martinthomson: Missing context: "CRYPTO_HS frames are functionally identical to STREAM frames, except that they do not bear a stream identifier; they are not flow controlled; and they do not carry markers for optional offset, optional, length, and the end of the stream."

draft-ietf-quic-transport.md
```diff
@@ -3016,6 +3185,66 @@ advised to bundle as few streams as necessary in outgoing packets without losing
...
+encryption level, each of which starts at an offset of 0. This implies
+that each encryption level is treated as a separate CRYPTO_HS stream
+of data.
+
+Unlike STREAM frames, which include a Stream ID indicating to which
+stream the data belongs, the CRYPTO_HS frame carries data for a single
+stream per encryption level. The stream does not have an explicit
+end, so CRYPTO_HS frames do not have a FIN bit.
```
martinthomson: What do we do about extra Handshake messages on old epochs?  Those could be added to every epoch indefinitely, if it weren't for the fact that TLS would choke.  But I'm not entirely sure that it would choke if the handshake message were chosen carefully.

draft-ietf-quic-transport.md
martinthomson: I'd double down and make this comprehensive.  We might have confusion about whether 0 is allowed if we don't say at least this much.

draft-ietf-quic-transport.md
```diff
@@ -3733,9 +3958,9 @@ Stream flow control, which prevents a single stream from consuming the entire
 receive buffer for a connection.
 
 A data receiver sends MAX_STREAM_DATA or MAX_DATA frames to the sender
-to advertise additional credit. MAX_STREAM_DATA frames send the maximum
-absolute byte offset of a stream, while MAX_DATA sends the maximum sum
-of the absolute byte offsets of all streams other than stream 0.
+to advertise additional credit. MAX_STREAM_DATA frames send the
+maximum absolute byte offset of a stream, while MAX_DATA sends the
+maximum of the sum of the absolute byte offsets of all streams.
```
martinthomson: 👍

ekr:
> At risk of this turning into a unicorn, here are my comments. I don't think that you need to address all of them, but I would appreciate you taking one pass through this so we don't have too many editorial landmines to worry about later. Most are nitty, so they should be easy to fix; others I'm happy to punt on.

I will do so directly. ETA tomorrow.

draft-ietf-quic-recovery.md
```diff
@@ -422,6 +439,14 @@ sending any ACK frames in response.  In this case they can determine
 whether an immediate or delayed acknowledgement should be generated
 after processing incoming packets.
 
+### Crypto Handshake Data
+
+In order to quickly complete the handshake and avoid spurious
+retransmissions due to handshake alarm timeouts, handshake packets
+should use a very short ack delay, such as 1ms.  ACK frames may be
```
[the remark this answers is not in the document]
ianswett: It's ack delay everywhere in this doc.

draft-ietf-quic-recovery.md
```diff
@@ -422,6 +439,14 @@ sending any ACK frames in response.  In this case they can determine
...
+sent immediately when the crypto stack indicates all data for that
+encryption level has been received.
+
```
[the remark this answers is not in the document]
ianswett: This part is only about sending ACKs, but good point on the EndOfEarlyData.  As spec'd, that'd be TLP'd before being RTO'd, FYI.

draft-ietf-quic-tls.md
kazuho: @huitema:
> I am busy implementing this, and I have a small issue with the "MAY appear in packets of any encryption level." How do I verify that the packets are sent as intended? For example, what if I receive a CRYPTO_HS frame with offset 500 in an handshake-protected, and another CRYPTO_HS frame with an overlapping offset in an application-data protected packet?

My assumption is that such check will be provided as an API or a guideline by the TLS stack, and that therefore it is unnecessary to clarify in the specification on how that should be done.

For example, the `ptls_handle_messages` function of [picotls](https://github.com/h2o/picotls/pull/141) accepts the epoch as the argument, and returns an error if a TLS messages was being received in an incorrect epoch.

ekr: @huitema: I don't understand what you are asking here, because those are different data namespaces.

So, for instance, SH will be sent with Handshake[Offset=0], and NST will be sent with Application[Offset=0]. They don't interact with each other.

The case that is an error is when data comes in at a given encryption level after the TLS state machine has moved on. For instance, if the server sends Handshake: [Offset=X, SFIN], Handshake: [Offset=X + len(SFIN), <something>]. This isn't valid because the TLS handshake is now reading at Application. As Kazuho says, this should generate an error, and the text requires that:

```
- If the packet is from a previously installed encryption level, it
  MUST not contain data which extends past the end of previously
  received data in that flow. Implementations MUST treat any
  violations of this requirement as a connection error of type
  PROTOCOL_VIOLATION.
```

draft-ietf-quic-tls.md
```diff
@@ -399,18 +376,34 @@ Before starting the handshake QUIC provides TLS with the transport parameters
...
-
-At the server, when TLS provides handshake octets, it also needs to indicate
-whether the octets contain a HelloRetryRequest.  A HelloRetryRequest MUST always
-be sent in a Retry packet, so the QUIC server needs to know whether the octets
-are a HelloRetryRequest.
+A QUIC server starts the process by providing TLS with the client's
+handshake octets.
+
+At any given time, an endpoint will have a current sending encryption
+level and receiving encryption level. Each encryption level is
```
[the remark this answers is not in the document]
martinthomson: I see.  The problem here is that I was thinking QUIC, and this really applies to TLS.  TLS assumes ordering, but QUIC doesn't.  That's something that needs to be improved here, but I don't think that we need to do that here.

draft-ietf-quic-transport.md
```diff
@@ -2849,43 +3014,19 @@ received packets in preference to packets received in the past.
...
-ACK frames that acknowledge protected packets MUST be carried in a packet that
-has an equivalent or greater level of packet protection.
-
-Packets that are protected with 1-RTT keys MUST be acknowledged in packets that
-are also protected with 1-RTT keys.
-
-A packet that is not protected and claims to acknowledge a packet number that
-was sent with packet protection is not valid.  An unprotected packet that
-carries acknowledgments for protected packets MUST be discarded in its entirety.
+ACK frames MUST only be carried in a packet that has the same packet
```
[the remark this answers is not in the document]
martinthomson: My point is that you can't cross-ACK because you aren't capable of acknowledging something in a different packet number space.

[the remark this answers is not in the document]
ekr: I guess this may be a terminology issue. Let me try to explain what I am saying again.
 
Say I receive the following packets 2:1, 2:2, 3:1, 3:3 where X:Y means "epoch X, PN Y". And then I screw up and ACK PN space 3 in epoch 2, so I send epoch=2: ACK[3, 1]. Technically, this means "Acknowledge PNs 3 and 1 in epoch 2", so it's not a cross-ACK from the receiver's perspective, but it *is* an ACK of an unsent packet, which at least at some times we have thought was grounds for tearing down a connection.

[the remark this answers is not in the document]
martinthomson: Yeah, I think that there are two statements to make, each of which differs from the one you chose.

1. An endpoint MUST NOT acknowledge packets you haven't received (and processed).
2. An ACK frame can only acknowledges packets from the same packet number space.

The conjunction of these two things is what you are getting at: if you receive a packet in one packet number space, you cannot acknowledge it in a different space because that would be acknowledging a different packet, one you might not have received.

draft-ietf-quic-recovery.md
```diff
@@ -280,23 +294,28 @@ connection's final smoothed RTT value as the resumed connection's initial RTT.
...
+handshake alarm without receiving an acknowledgement for a new packet,
+the sender SHOULD double the handshake timeout and set an alarm for this
+period.
 
-When the alarm fires, the sender MUST retransmit all unacknowledged handshake
-data, by calling RetransmitAllUnackedHandshakeData(). On each consecutive
-firing of the handshake alarm, the sender SHOULD double the handshake timeout
-and set an alarm for this period.
+When CRYPTO_HS frames are outstanding, the TLP and RTO timers are not active
+unless the CRYPTO_HS frames were sent at 1RTT encryption.
```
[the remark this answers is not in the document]
ianswett: I was intending it to be encryption level, since I thought it was necessary to send end of early data in a timely manner?  But maybe it doesn't matter that much? 

It's possible a signal that the handshake is complete would be preferable to monitoring what encryption level CRYPTO_HS is sent at, but I'll just make it consistent for now.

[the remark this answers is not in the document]
martinthomson: Sounds reasonable.  I just wanted to check.  And yeah, we probably need to send EndOfEarlyData quickly.  It blocks access to 1-RTT data just like Finished.

Issue #1452, opened 2018-06-15
CID ratcheting language is imprecise


MikeBishop:
As currently written, each endpoint MUST switch to a new CID when it sees its peer do so.  If taken literally by each peer, this will cause a change of CID by each peer every round trip; unless a sufficient stream of NCID frames is kept up, they'll exhaust the supply and #1276 would have the connection be closed.  That seems both unnecessarily fragile and not actually the intent of this requirement.

Intuitively, you want to switch when:
- You see the peer do so
- It's not in response to your own switch

Not marking this design, because I'm fairly sure this is the intent of the existing text, albeit not what it actually says.  I'm having trouble coming up with a concise way of wording the second piece of the condition, however.

Ironically, this would have been simple when CIDs had sequence numbers -- you MUST always use a CID with a sequence number >= the sequence number of the CID you're receiving from the peer.  But we weren't actually using them for that.  We could do something similar, though, saying you must count the number of times the peer has changed CIDs post-handshake and you must have changed at least as many times.

MikeBishop:
Packet loss makes this even harder, similar to the case in #1464.  Using the counter approach, if you don't un-count a CID where all packets using it were lost, you'll fail to change CIDs the next time the peer changes.  However, if you *do* un-count it, but any of the packets were declared lost in error, then you wind up in the same loop where your delayed packet causes the peer to change, which causes you to change, which causes the peer to change....

I'm starting to think we need the sequence numbers back to make this work, but with some additional text about probing packets on a different path being handled a bit differently.

PR #1453, opened 2018-06-15
Expand CID ratchet text

### LINKED ISSUE #1276: Exhausting a supply of connection IDs
What - if anything - do we want to say about connection migration in the face of a server that provides too few (including 0) connection IDs?  It's a fairly simple trade-off that we can explain.  I suspect that different clients will adopt different policies though.

And, should we provide a means to request more?

MikeBishop:
Fixes #1276, plus removes a stray reference to CIDs being used in sequential order.  Encountered #1452 while writing this; if we come to a resolution quickly I can incorporate that here as well.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1857,13 +1857,20 @@ congestion state (see {{migration-cc}}), so the port SHOULD only be changed
 infrequently.
 
 An endpoint that receives a successfully authenticated packet with a previously
-unused connection ID MUST use the next available connection ID for any packets
-it sends to that address.  To avoid changing connection IDs multiple times when
+unused connection ID MUST use a new connection ID for any future packets it
+sends to that address.  To avoid changing connection IDs multiple times when
 packets arrive out of order, endpoints MUST change only in response to a packet
 that increases the largest received packet number.  Failing to do this could
 allow for use of that connection ID to link activity on new paths.  There is no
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1857,13 +1857,20 @@ congestion state (see {{migration-cc}}), so the port SHOULD only be changed
 infrequently.
 
 An endpoint that receives a successfully authenticated packet with a previously
-unused connection ID MUST use the next available connection ID for any packets
-it sends to that address.  To avoid changing connection IDs multiple times when
+unused connection ID MUST use a new connection ID for any future packets it
+sends to that address.  To avoid changing connection IDs multiple times when
```
martinthomson: Isn't the point of this PR to cover the case you describe in the issue?  That is, you need to say "unless it is likely that the peer changed its connection ID in response to the endpoint changing."  Or something like that.  I don't know how to implement that other than to block switches when you change, unless there is a path change.  This is tricky.

MikeBishop: The point of the PR grew as I worked on it.  :-)  Initially, it was to replace "the next available" with "a new" now that sequence numbers are gone.  Then, while I was at it, I decided to pick up #1276 (what happens if you're out of CIDs?).  But then I ran into the fact that the existing text will lead you into that spiral.

PR #1460, opened 2018-06-19
ECN Proposal with only two counters

gloinul:
This is an evolution based on #1372 but using @martinthomson proposal of only two counters tracking only the CE and non-default ECT marking.


martinthomson:
we're making progress on the other, so let's close this

PR #1462, opened 2018-06-19
Enable SNI encryption

### LINKED ISSUE #1459: Soften SNI requirement
hq mandates use of the server_name extension rather too strongly.  

> The Server Name Indication (SNI) extension [RFC6066] MUST be included in the TLS handshake.

RFC 7540 is much weaker on this point.

> The TLS implementation MUST support the Server Name Indication (SNI) [TLS-EXT] extension to TLS.  HTTP/2 clients MUST indicate the target domain name when negotiating TLS.

If we are to allow SNI encryption, then we should adopt similar language.

janaiyengar:
Closes #1459.


ekr:
@janaiyengar: The commit message seems wrong here.

Issue #1464, opened 2018-06-20
Confusion about number of outstanding CIDs


MikeBishop:
Suppose an endpoint is probing an alternate path to keep it available for future migration.  (Probing periodically over cellular in case Wi-Fi goes away, for example.)  The endpoint has N connection IDs outstanding and chooses to use one of them on a probing packet, which gets lost.  When the endpoint retransmits the probe, it uses a different CID because it has been quiescent for a while.  This probe gets through.

At this point, the endpoint believes it has N-2 remaining CIDs to work with, while its peer believes it has N-1 CIDs still issued.  Now the text about making sure your peer has outstanding CIDs becomes pointless, because the peers can't accurately estimate.  Repeat this a couple times, and you might find yourself stuck behind "MUST NOT send additional packets until a NEW_CONNECTION_ID frame is received" because you've run out and the peer thinks you haven't.

ACK tracking would let the endpoint detect that this has happened (all packets using CID A have been declared lost, so peer doesn't know I burned it), but that doesn't permit the endpoint to *fix* the problem.  I think this ultimately leads us to a NEED_MORE_CIDS frame, unless anyone else has an idea.  (Not quite CID_BLOCKED, because once you're blocked it's too late.)

mikkelfj:
Somewhat OT, but it got me thinking - for long should a connection track a given path? Consider an drone traversing a futuristic public mesh wifi network there could be thousands of migrations over the lifetime of the connection.

MikeBishop:
Implementation specific.  I'm inclined to say that once you've lost the local address, you can forget all paths associated with it, unless you have some local reason to believe it will come back.

nibanks:
I do agree that it's implementation specific, but I still think it's something they spec should provide recommendations for.

For @mikkelfj's scenario, I'm more worried about the server side. It keeps giving out more CIDs and the client keeps using them. Should we have a way for the client to explicitly tell the server that a particular old CID is no longer in use? Otherwise the best I can see the server doing it having some timer (idle timeout period?) per CID to age it out or just a simple max number of simultaneous CIDs to allow at any one time, and as new CIDs are given out, if there is no more room, the oldest is automatically aged out.

kazuho:
I think that we need a NEED_MORE_CIDS frame.

[...]

PR #1465, opened 2018-06-20
Restore CID sequence numbers

### LINKED ISSUE #1452: CID ratcheting language is imprecise

MikeBishop:
After trying to find a better way to manage CID changes, probing, etc. I concluded that what we had before and removed is a superior place to start than anything else I'm coming up with.

The result of this PR is that, while we no longer have packet number gaps (so #1280 is no longer an issue), the sequence of CIDs issued by each endpoint is once again ordered.  On each path (local address), you know the sequence number of the CID you're currently using, and you can recognize the sequence number of the CID you're currently receiving (because you assigned it).

Fixes #1452 by requiring that each CID change on a path move to a CID with a greater sequence number; you move to the CID with the matching sequence number if it's usable, but if it's not, you move to something with a higher sequence number.  That will force your peer to meet you on that sequence number instead, and either way you land in equilibrium on each path.

This improves, but doesn't totally fix, the core problem on #1464.  Once you've moved to the CID with sequence number X, the peer knows you can't use CID X-1 on this path even if you've never used it; if you're on the last / next-to-last CID the peer issued, you need more, even if you have some old ones you could still use on a different path.  However, if you have the bad luck to use your last CID for a probe which gets dropped and you're no longer able to send on that path, you can still get stuck if your peer is only giving you one CID at a time.  This probably still requires a `REQUEST_CONNECTION_ID` frame.

Potentially fixes the side-issue discussed on #1464, too.  Because the CID sequence must advance per path, you can forget old CIDs you've issued once every path has either moved past that sequence number or been abandoned.  New paths must use a higher sequence number than ever used on any path.

However, there's a slight possibility that peers will have different concepts of abandoned (particularly for probing on backup paths that are never migrated to).  We could mint a `RETIRE_CONNECTION_ID` frame to tell the receiver that connection IDs with a sequence before X are no longer recognized by the sender.  Or it might be sufficient to send a probe packet back on the potentially-abandoned path with a newer sequence number; assuming this is received successfully, it will force the peer to update the CID used on that path as well (even if it doesn't send any packets back).

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1086,12 +1086,21 @@ address already in use.
 
 An endpoint MUST NOT send packets with a connection ID which has a lower
 sequence number than the highest sequence number of any connection ID ever sent
-or received on that local address.
+or received on that local address.  This ensures that when an endpoint migrates
+to a new path or changes connection ID on an existing path, the packets will use
+a new connection ID in both directions.
 
 Implementations SHOULD ensure that peers have a connection ID with a matching
 sequence number available when changing to a new connection ID.  An
 implementation could do this by always supplying a corresponding connection ID
-to a peer for each connection ID received from that peer.
+to a peer for each connection ID received from that peer.k
+
+While endpoints select connection IDs as appropriate for their implementation,
+the connection ID MUST NOT include the unprotected sequence number.  Endpoints
+need to be able to recover the sequence number associated with each connection
+ID they generate without relying on information available to unaffiliate
+parties. A connection ID that encodes an unencrypted sequence number could be
+used to correlate connection IDs across network paths.
 
 
 ## Matching Packets to Connections {#packet-handling}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2004,7 +2004,9 @@ new connection IDs using the NEW_CONNECTION_ID frame.
 An endpoint might need to send packets on multiple networks without receiving
 any response from its peer.  To ensure that the endpoint is not linkable across
 each of these changes, a new connection ID is needed for each network.  To
-support this, multiple NEW_CONNECTION_ID messages are needed.
+support this, multiple NEW_CONNECTION_ID messages are needed.  Each
+NEW_CONNECTION_ID is marked with a sequence number.  Connection IDs MUST be used
+in the order in which they are numbered.
 
 Upon changing networks an endpoint MUST use a previously unused connection ID
 provided by its peer.  This eliminates the use of the connection ID for linking
@@ -2769,6 +2771,8 @@ The NEW_CONNECTION_ID is as follows:
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                          Sequence (i)                       ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |   Length (8)  |          Connection ID (32..144)            ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                                                               |
@@ -2782,6 +2786,14 @@ The NEW_CONNECTION_ID is as follows:
 ~~~
 
 The fields are:
+
+Sequence:
+
+: A variable-length integer.  This value starts at 0 and increases by 1 for each
+  connection ID that is provided by the server.  The connection ID that is
+  assigned during the handshake is assumed to have a sequence of -1.  That is,
+  the value selected during the handshake comes immediately before the first
+  value that a server can send.
 
 Length:
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2007,9 +2007,9 @@ each of these changes, a new connection ID is needed for each network.  To
 support this, multiple NEW_CONNECTION_ID messages are needed.  Each
 NEW_CONNECTION_ID is marked with a sequence number.  Connection IDs MUST be used
 on only one local address, and each local address MUST advance to a connection
-ID with a later sequence number each time the connection ID changes.  Once a
-connection ID has been used, connection IDs with an earlier sequence number MUST
-NOT be used for packets with a greater packet number from that local address.
+ID with a later sequence number each time the connection ID changes.  Once a new
+connection ID is used, earlier connection IDs MUST NOT be used for new packets
+sent from the same local address.
 
 Upon changing networks an endpoint MUST use a previously unused connection ID
 provided by its peer.  This eliminates the use of the connection ID for linking
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2032,17 +2032,22 @@ congestion state (see {{migration-cc}}), so the port SHOULD only be changed
 infrequently.
 
 An endpoint that receives a successfully authenticated packet with a previously
-unused connection ID MUST advance to a connection ID with an equal or greater
-sequence number for any future packets it sends to that address.  Failing to do
-this could allow for use of that connection ID to link activity on new paths.
-There is no need to move to a new connection ID if the address of a peer changes
-without also changing the connection ID.  If no new connection IDs are
-available, the endpoint MUST NOT send additional packets until a
-NEW_CONNECTION_ID frame is received.  However, an implementation need not stop
-sending packets due to a gap in sequence numbers; it simply advances to the next
-currently-available connection ID it has received.  Connection IDs with earlier
-sequence numbers which arrive later MAY be retained for use on other local
-addresses or discarded.
+unused connection ID MUST advance to the connection ID whose sequence number is
+equal to the sequence number of the newly received connection ID for any future
+packets it sends to that address.  If that connection ID cannot be used (e.g.,
+it was used on a different path or has not been received), it MUST advance to a
+connection ID with a greater sequence number for any future packets it sends to
+that address.  Failing to do this could allow for use of that connection ID to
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1834,7 +1834,12 @@ new connection IDs using the NEW_CONNECTION_ID frame.
 An endpoint might need to send packets on multiple networks without receiving
 any response from its peer.  To ensure that the endpoint is not linkable across
 each of these changes, a new connection ID is needed for each network.  To
-support this, multiple NEW_CONNECTION_ID messages are needed.
+support this, multiple NEW_CONNECTION_ID messages are needed.  Each
+NEW_CONNECTION_ID is marked with a sequence number.  Connection IDs MUST be used
+on only one local address, and each local address MUST advance to a connection
+ID with a later sequence number each time the connection ID changes.  Once a
+connection ID has been used, connection IDs with an earlier sequence number MUST
+NOT be used for packets with a greater packet number from that local address.
```
nibanks: I'm confused as to how this works with probing a new path (but not moving to it). Assume I am currently on Path A using CID 1, then I probe Path B with CID 2. This text seems to state that I may not use CID 1 on Path A with any packet number greater than I used to probe Path B with  CID 2. Is that correct?

draft-ietf-quic-transport.md
```diff
@@ -2613,6 +2627,14 @@ The NEW_CONNECTION_ID is as follows:
 
 The fields are:
 
+Sequence:
+
+: A variable-length integer.  This value starts at 0 and increases by 1 for each
+  connection ID that is provided by the server.  The connection ID that is
+  assigned during the handshake is assumed to have a sequence of -1.  That is,
```
nibanks: Is there a real reason we can consider the handshake CID `0` and start from there? I'd rather not have to deal with negative numbers in my code.

draft-ietf-quic-transport.md
MikeBishop: At that point, you may not use a CID < 1 on path A, and may not use a CID < 2 on path B.  You also aren't allowed to use CID = 2 on path A, because it's been used on a different path.

draft-ietf-quic-transport.md
MikeBishop: Probably not.  This is just the existing text.  It's actually a little worse than that, because the Server's Preferred Address gives the client a CID to use when probing, so there can actually be two CIDs prior to the first one issued by frame.

draft-ietf-quic-transport.md
MikeBishop: Basically:
- For each path, the CID sequence number must advance, but the currently-in-use CID for each path is unrelated.  You could keep probing forever with CID2 on Path B even while Path A has moved up to CID 39, so long as the path isn't abandoned.
- Each CID can only be used on a single path.  If you used CID2 on Path B, you have to skip CID2 on Path A.

nibanks: Ok, maybe instead of "Once a connection ID has been used, connection IDs with an earlier sequence number MUST NOT be used for packets with a greater packet number from that local address." something like "Once a new connection ID is used, earlier connection IDs MUST NOT be used for new packets sent from the new local address".

draft-ietf-quic-transport.md
nibanks: Then maybe we define handshake as `0`, the server's preferred address as `1` (even if not present) and then new CIDs start from there (i.e. `2`). I don't see any reason to prevent gaps in the CID sequence number space either.

MikeBishop: The alternative being -2, -1, and starting from 0.  I think the logic there is the principle that we should try to avoid there being invalid values, and 0/1 would be invalid values still present in the frame.  If the frame can carry a 0, then 0 should be the first value value.

MikeBishop: This also creates an odd situation in that SPA prompts the client to probe, but the server won't have a corresponding CID at that sequence number, so it will have to use the first CID the client issued to it -- *if* the client issues one with/before the probe! -- which will force the client to advance again.

draft-ietf-quic-transport.md
```diff
@@ -1834,7 +1834,12 @@ new connection IDs using the NEW_CONNECTION_ID frame.
 An endpoint might need to send packets on multiple networks without receiving
 any response from its peer.  To ensure that the endpoint is not linkable across
 each of these changes, a new connection ID is needed for each network.  To
-support this, multiple NEW_CONNECTION_ID messages are needed.
+support this, multiple NEW_CONNECTION_ID messages are needed.  Each
+NEW_CONNECTION_ID is marked with a sequence number.  Connection IDs MUST be used
+on only one local address, and each local address MUST advance to a connection
```
martinthomson: Perhaps
... on only one local address.  An endpoint MUST use a connection ID with a higher sequence number if it changes to using a different local address or network interface.  Once an endpoint has used a connection ID for packets sent from a given local address, they MUST NOT use a connection ID with a lower sequence number with that address.

draft-ietf-quic-transport.md
martinthomson: There can be very many connection IDs during the handshake with this new Retry packet, so it's all special casing here.  Start at 2^16-N if you feel compelled to track things this way.

draft-ietf-quic-transport.md
```diff
@@ -1857,15 +1862,22 @@ congestion state (see {{migration-cc}}), so the port SHOULD only be changed
...
+the address of a peer changes without also changing the connection ID, or if the
+connection ID changes to the one with the sequence number equal to the
+endpoint's current connection ID.  If no new connection IDs are available, the
 endpoint MUST NOT send additional packets until a NEW_CONNECTION_ID frame is
-received.
+received.  However, an implementation need not stop sending packets due to a gap
+in sequence numbers; it simply advances to the next currently-available
+connection ID it has received.  Connection IDs with earlier sequence numbers
+which arrive later MAY be retained for use on other local addresses or
+discarded.
```
martinthomson: OMG wall of text :)

When this happens, it's probably cause to create a new section.

FWIW, I like the idea that once you use a connection ID on a given interface, you can't use a lower-numbered one.  That could extend to responses.  Once you see a connection ID on a given tuple, then you can't use a lower-numbered one.  That leads to some simple rules:

When you use a new interface/address you MUST use of a new connection ID.
When you use a new connection ID, you MUST use one with a higher sequence number than you have ever used before.
You MUST NOT use a connection ID with a lower sequence number than you have either sent or received on that path.

Formulating the text around that is what we need.

draft-ietf-quic-transport.md
martinthomson: Or, create a new section on connection ID use along the lines I describe below and cite that instead.

MikeBishop:
Moved to a separate section.  Doesn't actually change the line count any, but hopefully it's clearer.

draft-ietf-quic-transport.md
MikeBishop: What if we just eliminate this statement and let it be implementation-local?  The new text I added to the CID section says that "The series of
connection IDs issued by an endpoint is ordered, with the final connection ID
selected during the handshake coming first."

We could also fix this by removing the CID from SPA and simply saying that the server needs to send the client a CID it can use for the probe pretty promptly if it wants the client to do so.

martinthomson: I don't think that would help with the suggestion in #1468.  Insisting on a fixed starting point ensures that you can do things like matching.

Removing the connection ID from the preferred address would make it harder to get that right because you then need to wait for a NEW_CONNECTION_ID frame before migrating.

draft-ietf-quic-transport.md
```diff
@@ -932,17 +932,60 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+for the first time on a new local address, a new connection ID MUST be used with
+a higher sequence number than any connection ID previously used on any local
+address.  At any time, an endpoint MAY change to a new connection ID on a local
+address already in use.
+
+An endpoint MUST NOT send packets with a connection ID which has a lower
+sequence number than the highest sequence number of any connection ID ever sent
+or received on that local address.
+
+Implementations SHOULD ensure that peers have a connection ID with a matching
```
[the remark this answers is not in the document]
MikeBishop: But the trade-off is that lost packets can cause a peer to appear to violate this temporarily, so it's not realistically enforceable by a peer.  You can't "MUST ensure" something has arrived on the other side, only detect whether it has or not.  And the connection doesn't actually break; the peer will have to advance to a higher-order CID, which will force the endpoint to move off of the one that doesn't have a match (assuming it still hasn't arrived).

Actually, what a smart implementation could do is to bundle the NEW_CONNECTION_ID frame with the corresponding CID in every packet between when it changes to a new CID until *something* gets ACK'd.

draft-ietf-quic-transport.md
```diff
@@ -932,17 +932,60 @@ different IP or port at either endpoint, due to NAT rebinding or mobility, as
...
+
+Each connection ID MUST be used on only one local address. When packets are sent
+for the first time on a new local address, a new connection ID MUST be used with
+a higher sequence number than any connection ID previously used on any local
+address.  At any time, an endpoint MAY change to a new connection ID on a local
+address already in use.
+
+An endpoint MUST NOT send packets with a connection ID which has a lower
+sequence number than the highest sequence number of any connection ID ever sent
+or received on that local address.
```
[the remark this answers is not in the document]
MikeBishop: Good point.  I tried to put it mid-discussion, but it didn't flow, so right now it's tucked at the end.  Editorial discretion welcome.

draft-ietf-quic-transport.md
martinthomson: I've merged this, so let's open an issue to close this out.

PR #1472, opened 2018-06-22
Modify codepoints on decoder and request/push streams for consistency

### LINKED ISSUE #1471: QPACK instruction code points are inconsistent
HPACK/QPACK can encode N different instructions in N-1 bits.  Both HPACK and the QPACK encoder stream use the same style fore code points - each instruction has either one or zero '1' bits (eg: 1, 01, 001, 0001, 0000).  The QPACK decoder stream and request/push streams however use (0, 10, 11) and (1, 00, 011, 0100, 0101), each of which have instructions with more than one '1'.  The logic to decode the instruction is slightly simpler when using the HPACK form, eg:

```
if byte & shortest_instruction:
  ...
else if byte & next_shortest_instruction:
  ...
else:  # it's the all-zero instruction
  ...
```

This logic doesn't work on the other two streams right now, you need to mask and check for equality.

afrind:
The prefix lengths remain the same but the instruction bits now follow the rule that each instruction has at most one '1' bit.

Fixes #1471

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -524,7 +524,7 @@ server's header blocks and table updates.
 
 ### Table State Synchronize
 
-The Table State Synchronize instruction begins with the '10' two-bit pattern.
+The Table State Synchronize instruction begins with the '01' two-bit pattern.
 The instruction specifies the total number of dynamic table inserts and
 duplications since the last Table State Synchronize or Header Acknowledgement
 that increased the Largest Known Received dynamic table entry.  This is encoded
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -714,9 +714,9 @@ header field name matches the header field name of an entry stored in the static
 table or the dynamic table.
 
 If the entry is in the static table, or in the dynamic table with an absolute
-index less than or equal to Base Index, this representation starts with the '00'
+index less than or equal to Base Index, this representation starts with the '01'
 two-bit pattern.  If the entry is in the dynamic table with an absolute index
-greater than Base Index, the representation starts with the '0101' four-bit
+greater than Base Index, the representation starts with the '0000' four-bit
 pattern.
```
MikeBishop: This change is fine, but didn't we separate out the Post-Base things to their own headers?  Did we miss one?

afrind:
I'm not concerned about this change delaying interop dramatically.  Implicit acks landed last week and was a bigger change.

MikeBishop:
@martinthomson, that's perhaps technically true, but I don't think this pushes out interop noticeably.  The bit sequences are the same length they previously were, and were always arbitrary bit sequences.  This should be a trivial change to incorporate.

PR #1473, opened 2018-06-25
Extension frames

### LINKED ISSUE #58: Frame type extensions

martinthomson:
This is the outcome we discussed in Kista.

One tweak here is to avoid allowing variant variable-length integer encodings.  The definition of variable-length integers would allow for four different encodings of the core set of frame types, which would require recipients to modify their code to look at more than one octet.  For instance, ping is 0x07, 0x4007, 0x80000007, and 0xc000000000000007.  I think that it's easier for all involved not to have to worry about long encodings until they want to experiment.

I'm flexible on that point, as I am with respect to registration policies.

Closes #58, #1072, #1068.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -880,22 +880,25 @@ Protected payloads MUST contain at least one frame, and MAY contain multiple
 frames and multiple frame types.
 
 Frames MUST fit within a single QUIC packet and MUST NOT span a QUIC packet
-boundary. Each frame begins with a Frame Type byte, indicating its type,
-followed by additional type-dependent fields:
+boundary. Each frame begins with a Frame Type, indicating its type, followed by
+additional type-dependent fields:
 
 ~~~
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|   Type (8)    |           Type-Dependent Fields (*)         ...
+|                           Type (i)                          ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                   Type-Dependent Fields (*)                 ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ~~~
 {: #frame-layout title="Generic Frame Layout"}
 
-Frame types are listed in {{frame-types}}. Note that the Frame Type byte in
-STREAM frames is used to carry other frame-specific flags.  For all
-other frames, the Frame Type byte simply identifies the frame.  These frames are
-explained in more detail as they are referenced later in the document.
+The frame types defined in this specification are listed in {{frame-types}}.
+The Frame Type in STREAM frames is used to carry other frame-specific flags.
+For all other frames, the Frame Type field simply identifies the frame.  These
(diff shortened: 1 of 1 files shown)


mikkelfj:
I'd definitely prefer a few code points for var length for example 0xfd for a single octet following. 0xfe for two octest following, and 0xff for a varint following. This will use much less space in the average case.

mikkelfj:
Fair point.

Each length could be explicitly placed in its own namespace by mapping all frames to 64-bit including the length encoding. This makes PING unique and other values easy to index.

Another fun encoding would be an FNV1a-64 hash of the reverse org name to use as a base for extensions where the upper two bits clamped to 1. Each such name reserving 100 sequential values.

draft-ietf-quic-transport.md
```diff
@@ -921,6 +924,30 @@ explained in more detail as they are referenced later in the document.
...
 undesirable side effects or errors when received more than once.
 
+The Frame Type field uses a variable length integer encoding (see
+{{integer-encoding}}) with one exception.  To ensure simple and efficient
+implementations of frame parsing, a frame type MUST use the shortest possible
+encoding.  Though a two-, four- or eight-octet encoding of the frame types
+defined in this document is possible, the Frame Type field for these frames are
+encoded on a single octet.  For instance, though 0x4007 is a legitimate
+two-octet encoding for a variable-length integer with a value of 7, PING frames
+are always encoded as a single octet with the value 0x07.
```
MikeBishop: And what am I required to do if I encounter a 2+ byte encoding of 7?

draft-ietf-quic-transport.md
```diff
@@ -921,6 +924,30 @@ explained in more detail as they are referenced later in the document.
...
+
+## Extension Frames
+
+QUIC frames do not use a self-describing encoding.  An endpoint therefore needs
+to understand the syntax of all frame before it can successfully process a
```
MikeBishop: all frames / each frame

draft-ietf-quic-transport.md
```diff
@@ -921,6 +924,30 @@ explained in more detail as they are referenced later in the document.
...
+packet.  This allows for efficient encoding of frames, but it means that an
+endpoint cannot send a frame of a type that is unknown to its peer.
+
+An extension to QUIC that wishes to use a new type of frame MUST first ensure
+that a peer is able to understand the frame.  An endpoint can use a transport
+parameter to signal its willingness to receive a new type of frame, or even
+multiple types of frame with the one transport parameter.
```
MikeBishop: ...receive one or more extension frame types.

draft-ietf-quic-transport.md
ianswett: Agreed, I think we either need to say you MAY close the connection or this is a SHOULD not a MUST

draft-ietf-quic-transport.md
```diff
@@ -921,6 +924,30 @@ explained in more detail as they are referenced later in the document.
...
+
+## Extension Frames
+
+QUIC frames do not use a self-describing encoding.  An endpoint therefore needs
+to understand the syntax of all frame before it can successfully process a
+packet.  This allows for efficient encoding of frames, but it means that an
+endpoint cannot send a frame of a type that is unknown to its peer.
+
+An extension to QUIC that wishes to use a new type of frame MUST first ensure
+that a peer is able to understand the frame.  An endpoint can use a transport
```
ianswett: It would be nice to have an example here at some point, but since we don't have any extension frames, I guess that's for the future?

MikeBishop: There's always @larseggert's suggestion that BLOCKED et al. should be extensions.  That's been on hold while we wait for an extension mechanism to be defined, but it's a perfectly reasonable one.  Define a transport parameter `inform_if_blocked` and if that's set, the various BLOCKED frames are understood.

ianswett: That sounds like a reasonable suggestion to me.

PR #1476, opened 2018-06-26
Crypto cleanup

martinthomson:
This is two changes: a mechanical renaming of CRYPTO_HS to CRYPTO.  Then it's a reflowing of a bunch of paragraphs.  Both are mechanical in nature, though I did notice and correct a couple of typos in the process (which I'll have to dig out if we decide this is a bad idea).

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -111,7 +111,7 @@ important to the loss detection and congestion control machinery below.
 * Retransmittable packets are those that contain at least one
   retransmittable frame.
 
-* Crypto handshake data is sent in CRYPTO_HS frames, and uses the reliability
+* Crypto handshake data is sent in CRYPTO frames, and uses the reliability
   machinery of QUIC underneath.
 
 * ACK and ACK_ECN frames contain acknowledgment information. ACK_ECN frames
@@ -282,9 +282,9 @@ and Retransmission Timeout mechanisms.
 
 ### Crypto Handshake Timeout
 
-Data in CRYPTO_HS frames is critical to QUIC transport and crypto negotiation,
+Data in CRYPTO frames is critical to QUIC transport and crypto negotiation,
 so a more aggressive timeout is used to retransmit it.  Below, the word
-handshake packet is used to refer to packets containing CRYPTO_HS frames,
+handshake packet is used to refer to packets containing CRYPTO frames,
 not packets with the specific long header packet type Handshake.
 
 The initial handshake timeout SHOULD be set to twice the initial RTT.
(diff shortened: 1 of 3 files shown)


chris-wood:
My $0.02: I'm not a fan of CRYPTO_HS or CRYPTO. I'd much prefer them to be named HANDSHAKE. Pushback to that stemmed from the collision with packet type names, which I'm not convinced is confusing. (I'd expect handshake packets to carry handshake frames after all.)

mikkelfj:
>  I'd much prefer them to be named HANDSHAKE.

But some communication happens post handshake, such as rekeying.

ekr:
I also prefer HANDSHAKE.

ianswett:
I also prefer CRYPTO over HANDSHAKE FWIW

chris-wood:
Regarding post-handshake events -- I consider these part of the handshake path. They simply occur later in time. This really is a minor detail. I just don't like the name CRYPTO. I'm also not fond of SECURITY or SEC_STREAM.

janaiyengar:
Such reflow. Wow.

I prefer CRYPTO. But either is better than the current CRYPTO_HS.

All this reflow is going to rot rapidly, so if you want to keep discussing this name, I'd at least submit the reflow/typo changes and discuss the name separately.

draft-ietf-quic-recovery.md
```diff
@@ -255,11 +252,11 @@ than 1.25 * max(SRTT, latest_RTT) since when it was sent.
 Using max(SRTT, latest_RTT) protects from the two following cases:
 
 * the latest RTT sample is lower than the SRTT, perhaps due to reordering where
-packet whose ack triggered the Early Retransit process encountered a shorter
-path;
+  packet whose ack triggered the Early Retransit process encountered a shorter
```
MikeBishop: **the** packet

draft-ietf-quic-recovery.md
```diff
@@ -274,36 +271,35 @@ alarm for TCP as well, and this document incorporates this advancement.
...
 
 ### Crypto Handshake Timeout
 
-Data in CRYPTO_HS frames is critical to QUIC transport and crypto negotiation,
-so a more aggressive timeout is used to retransmit it.  Below, the word
-handshake packet is used to refer to packets containing CRYPTO_HS frames,
-not packets with the specific long header packet type Handshake.
+Data in CRYPTO frames is critical to QUIC transport and crypto negotiation, so a
+more aggressive timeout is used to retransmit it.  Below, the word handshake
+packet is used to refer to packets containing CRYPTO frames, not packets with
```
MikeBishop: (Unmodified text, but you touch it, you own it.)  "handshake packet" isn't actually *a* word.  Perhaps "the term," and put "handshake packet" in quotes to designate it as the term you're discussing for clarity?

PR #1478, opened 2018-06-26
tls: fix citation format

reschke:


martinthomson:
I've been switching these to "Section x.y of {{...}}" as I go.  I will do that instead.  I'll hold this PR open until that's done though (this would be better than not doing that so I'll merge this if I don't get time before we cut -13).

martinthomson:
Fixed as part of other changes.

PR #1480, opened 2018-06-26
Replace default with recommended

### LINKED ISSUE #1475: "Default" max ack delay
> The default maximum ack delay in QUIC is 25ms.

We can't set defaults in a spec.  We can recommend this as a sensible value though.

ianswett:
Fixes #1475

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -222,7 +222,7 @@ reordering of packets in the network.
 
 The RECOMMENDED initial value for kReorderingThreshold is 3.
 
-We derive this default from recommendations for TCP loss recovery {{?RFC5681}}
+We derive this recommendation from TCP loss recovery {{?RFC5681}}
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -423,7 +423,7 @@ QUIC SHOULD delay sending acknowledgements in response to packets,
 but MUST NOT excessively delay acknowledgements of packets containing
 non-ack frames.  Specifically, implementaions MUST attempt to
 enforce a maximum ack delay to avoid causing the peer spurious
-timeouts.  The default maximum ack delay in QUIC is 25ms.
+timeouts.  The RECOMMENDED maximum ack delay in QUIC is 25ms.
```
janaiyengar: Why is only this recommendation capitalized? I think we should use capitalized RECOMMEND everywhere below

ianswett: Done.

PR #1493, opened 2018-06-28
Add an appendix related to packet number decoding.

rpaulo:
This PR adds an appendix with some pseudo code that show how to decode packet numbers.
The explanation in the draft was a bit terse so I think some example code might help.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -946,6 +946,8 @@ The next expected packet is the highest received packet number plus one.  For
 example, if the highest successfully authenticated packet had a packet number of
 0xaa82f30e, then a packet containing a 14-bit value of 0x9b3 will be decoded as
 0xaa8309b3.
+Example pseudo-code for packet number decoding can be found in 
+{{sample-packet-number-decoding}}.
 
 The sender MUST use a packet number size able to represent more than twice as
 large a range than the difference between the largest acknowledged packet and
@@ -4777,6 +4779,36 @@ from 0xFF00 to 0xFFFF are reserved for Private Use {{!RFC8126}}.
 
 
 --- back
+
+# Sample code for packet number decoding {#sample-packet-number-decoding}
+
+The following pseudo-code shows how an implementation can decode packet
+numbers after packet protection has been removed.
+
+~~~
+rcv_nxt        = 0xaa82f30e + 1
+pktn           = 0x9b3
+pktn_nbits     = 14
+pktn_len       = 1 << pktn_nbits
+pktn_win       = pktn_len / 2          // 0x2000
+
+// The incoming packet number is between:
+//     [rcv_nxt - pktn_win ; rcv_nxt + pktn_win[
+//
+// However, we can't just perform the exclusive-or of rcv_nxt
+// (with the appropriate bits removed)  with pktn because that
+// would yield 0xaa82c9b3 which is lower than rcv_nxt - pktn_win.
+//
+// We need to find a packet number that fits the window,
+// so we add the window to the expected packet number and check
+// for overflow.
+rcv_nxt_masked  = (rcv_nxt + pktn_win) & ~(pktn_len - 1)
+decoded_pktn    = rcv_nxt_masked + pktn
+
+// Note that we also check for underflow.
+if decoded_pktn > rcv_nxt + pktn_win and decoded_pktn > pktn_len:
+       decoded_pktn = decoded_pktn - pktn_len
+~~~
 
 # Change Log
 


draft-ietf-quic-transport.md
```diff
@@ -4778,6 +4780,36 @@ from 0xFF00 to 0xFFFF are reserved for Private Use {{!RFC8126}}.
...
+
+~~~
+rcv_nxt        = 0xaa82f30e + 1
+pktn           = 0x9b3
+pktn_nbits     = 14
+pktn_len       = 1 << pktn_nbits
+pktn_win       = pktn_len / 2          // 0x2000
+
+// The incoming packet number is between:
+//     [rcv_nxt - pktn_win ; rcv_nxt + pktn_win[
```
martinthomson: I would instead say the incoming packet number is greater than (rcv_next - pktn_win) and less than or equal to (rcv_next + pktn_win).  The use of '[' here implies a different interpretation, and it's not so obvious.

draft-ietf-quic-transport.md
```diff
@@ -4778,6 +4780,36 @@ from 0xFF00 to 0xFFFF are reserved for Private Use {{!RFC8126}}.
 
 --- back
 
+# Sample code for packet number decoding {#sample-packet-number-decoding}
+
+The following pseudo-code shows how an implementation can decode packet
+numbers after packet protection has been removed.
+
+~~~
+rcv_nxt        = 0xaa82f30e + 1
```
martinthomson: Maybe start with a function declaration:
InterpretPacketNumber(largest_received_pn, truncated_pn, truncated_pn_bits):

then you can include comments on each line that show the result of the step, as you do for pktn_win.

draft-ietf-quic-transport.md
```diff
@@ -4778,6 +4780,36 @@ from 0xFF00 to 0xFFFF are reserved for Private Use {{!RFC8126}}.
...
+pktn           = 0x9b3
+pktn_nbits     = 14
+pktn_len       = 1 << pktn_nbits
+pktn_win       = pktn_len / 2          // 0x2000
+
+// The incoming packet number is between:
+//     [rcv_nxt - pktn_win ; rcv_nxt + pktn_win[
+//
+// However, we can't just perform the exclusive-or of rcv_nxt
```
martinthomson: Exclusive-or might be misleading.  Also, we're concerned about being outside the window on both sides.

Maybe, "However, we can't just remove the trailing bits from rcv_nxt and add the truncated packet number because that might be outside the window."

draft-ietf-quic-transport.md
```diff
@@ -4778,6 +4780,36 @@ from 0xFF00 to 0xFFFF are reserved for Private Use {{!RFC8126}}.
...
+// (with the appropriate bits removed)  with pktn because that
+// would yield 0xaa82c9b3 which is lower than rcv_nxt - pktn_win.
+//
+// We need to find a packet number that fits the window,
+// so we add the window to the expected packet number and check
+// for overflow.
```
martinthomson: Here we find the top of the window and use that as the basis for calculating the value.  Then, the value can only be too large.

draft-ietf-quic-transport.md
```diff
@@ -4778,6 +4780,36 @@ from 0xFF00 to 0xFFFF are reserved for Private Use {{!RFC8126}}.
...
+rcv_nxt_masked  = (rcv_nxt + pktn_win) & ~(pktn_len - 1)
```
martinthomson: I would add a temporary for rcv_max = rcv_nxt + pktn_win so that you can a) explain it better, and b) use it in two places.

draft-ietf-quic-transport.md
```diff
@@ -4778,6 +4780,36 @@ from 0xFF00 to 0xFFFF are reserved for Private Use {{!RFC8126}}.
...
+decoded_pktn    = rcv_nxt_masked + pktn
+
+// Note that we also check for underflow.
```
martinthomson: "Note the extra check for underflow when the next packet number is near zero."

maskit:
Although I'm +1 on adding some pseudo code and this code seems nice as an implementation, I'm not sure this is also nice as a pseudo code to help understanding.

Maybe comparing two candidate packet numbers and choosing one from them (like `candidate1 - expected < candidate2 - expected`) is more straightforward approach.

martinthomson:
Note that you can avoid one if statement with:

  decoded_pktn -= ((rcv_nxt + pktn_win) & (pktn_len - 1) + pktn) & (pktn_len - 1)

But no one wants to see that in a spec.

@maskit, are you suggesting the simpler code:

```python
expected = next + 1
range = 1 << nbits
mask = range - 1
candidate = (expected & mask) | truncated
if candidate <= expected - range/2:
  return candidate + range
if candidate > expected + range/2 and candidate > range:
  return candidate - range
return candidate
```

That might be better, I agree.

martinthomson:
Just to be clear, the challenge with assembling candidates is that you end up with 3, not 2.

maskit:
@martinthomson Yeah, something like that. It probably make the code redundant and longer, but we shouldn't be too clever. The point is straightforwardness.

draft-ietf-quic-transport.md
rpaulo: I wanted to use this appendix as the basis for the example above but I agree we can follow the status quo

PR #1501, opened 2018-06-29
Define NAT rebinding better

### LINKED ISSUE #594: 5-tuple is stable during handshake
We need to state this as an assumption.  This is particularly relevant in the case where we have version negotiation and a rebalancing of the connection-id-based routing on the server cluster.  We're opting for a hope-based solution here and assuming that the server will be the same (or at least run the same versions) and that the 5-tuple is enough to keep the routing the same.

### LINKED ISSUE #952: Define or cite NAT rebinding
The QUIC Transport and Management documents makes several references to the term "NAT rebinding". With my pedant hat on, I was surprised to find no concise definition or citation to an expanded definition. Basic Google searching didn't help much either.

Given that this came up during Singapore as something that is important to mitigate, it seem like it would be useful to define.

martinthomson:
The change to add a means of disabling migration also mostly fixed #594,
so close that too.

Closes #952, #594.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1978,15 +1978,20 @@ network.  This section describes the process by which an endpoint migrates to a
 new address.
 
 An endpoint MUST NOT initiate connection migration before the handshake is
-finished and the endpoint has 1-RTT keys.  An endpoint also MUST NOT initiate
-connection migration if the peer sent the `disable_migration` transport
-parameter during the handshake.  An endpoint which has sent this transport
-parameter, but detects that a peer has nonetheless migrated to a different
-network MAY treat this as a connection error of type INVALID_MIGRATION. However,
-note that not all changes of peer address are intentional migrations. The peer
-could experience an unintended change of address due to NAT rebinding; endpoints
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1978,15 +1978,20 @@ network.  This section describes the process by which an endpoint migrates to a
...
+
+An endpoint also MUST NOT initiate connection migration if the peer sent the
+`disable_migration` transport parameter during the handshake.  An endpoint which
+has sent this transport parameter, but detects that a peer has nonetheless
+migrated to a different network MAY treat this as a connection error of type
+INVALID_MIGRATION.
+
+Not all changes of peer address are intentional migrations. The peer could
+experience NAT rebinding: a change of address due a middlebox, usually a NAT,
+allocating a new outgoing IP address and port for a flow.  Endpoints SHOULD
```
ianswett: IP address and/or port?  Typically NAT rebinds are only port change.

MikeBishop: Typically, but not universally.  I agree with "and/or", or perhaps even making the levels differ more -- "allocating a new port or even a new outgoing IP address for a flow."

PR #1510, opened 2018-07-02
Add TP disable_migration to the registry

dtikhonov:


alagoutte:
Confirmed... found the same issue... https://github.com/alagoutte/base-drafts/tree/disable_migration_iana

PR #1516, opened 2018-07-03
Discuss headers in more detail

### LINKED ISSUE #1461: hq: Are h2 pseudo-header requirements applicable
I was reading through [RFC 7540 Section 8.1.2.1](https://tools.ietf.org/html/rfc7540#section-8.1.2.1) and was wondering if these requirements still hold true for HTTP/QUIC. Especially regarding pseudo-headers, from RFC 7540 

> All pseudo-header fields MUST appear in the header block before regular header fields.

[HTTP/QUIC Section 3.2](https://tools.ietf.org/html/draft-ietf-quic-http-12#section-3.2) states that the message exchange sequence starts with one header block containing RFC 7230 headers. The HEADERS frame description mostly delegates to QPACK, which doesn't seem to mention pseudo-headers.

I appreciate a lot of HTTP/2 behaviour is inherited and the tendancy is to document difference. Perhaps we just need a small statement to highlight H2-defined pseudo-headers.

PS. the only mention of pseudo-headers is in 3.2.2 in relation to the CONNECT method :)

MikeBishop:
Fixes #1461 by pulling in a bit of RFC7540 language and referencing lots more.  I think we've assumed all of this, but the document didn't actually say.

(If you disagree that this is what we already intended to specify, scream now.)

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -327,7 +327,9 @@ listing of registered HTTP headers, see the "Message Header Field" registry
 maintained at <https://www.iana.org/assignments/message-headers>.
 
 Just as in previous versions of HTTP, header field names are strings of ASCII
-characters that are compared in a case-insensitive fashion.  As in HTTP/2,
+characters that are compared in a case-insensitive fashion.  Properties of HTTP
+header names and values are discussed in more detail in Section 3.2 of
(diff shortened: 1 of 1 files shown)


mikkelfj:
What are the exact valid header name characters? Can space and control characters be included? Ditto values - us that ASCII only, or also UTF-8? A reference to RFC would be ok, but “like earlier HTTP versions” is vague.

PR #1519, opened 2018-07-04
Grease the reserved bits

### LINKED ISSUE #1511: Short header: reserved bits
The test says:

> Reserved: The sixth, seventh, and eighth bits (0x7) of octet 0 are reserved for experimentation.

But there's no guidance on what they should be set to, so I think it should be:

> Reserved: The sixth, seventh, and eighth bits (0x7) of octet 0 are reserved for experimentation and MUST be set to zero.

martinthomson:
Closes #1511.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -444,10 +444,11 @@ Google QUIC Demultipexing Bit:
 Reserved:
 
 : The sixth, seventh, and eighth bits (0x7) of octet 0 are reserved for
-  experimentation.  Endpoints MUST ignore these bits unless they are
-  participating in an experiment that uses these bits.  An endpoint not
-  actively using these bits SHOULD set the value randomly to protect against
-  unwanted inference about particular values.
+  experimentation.  Endpoints MUST ignore these bits on packets they receive
+  unless they are participating in an experiment that uses these bits.  An
+  endpoint not actively using these bits SHOULD set the value randomly on
+  packets they send to protect against unwanted inference about particular
+  values.
 
 Destination Connection ID:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -444,7 +444,10 @@ Google QUIC Demultipexing Bit:
 Reserved:
 
 : The sixth, seventh, and eighth bits (0x7) of octet 0 are reserved for
-  experimentation.
+  experimentation.  Endpoints can MUST ignore these bits unless they are
```
marten-seemann: c̶a̶n̶

gQUIC 44 is using the Long Header according to draft-11 (?), where the last two bits are used to encode the packet number length. So technically, the bits are not ignored there.

martinthomson: This is version-specific text, so I think that we're OK.  Even if qQUIC doesn't use the invariants (yet) to identify versions.

draft-ietf-quic-transport.md
```diff
@@ -444,7 +444,10 @@ Google QUIC Demultipexing Bit:
 Reserved:
 
 : The sixth, seventh, and eighth bits (0x7) of octet 0 are reserved for
-  experimentation.
+  experimentation.  Endpoints MUST ignore these bits unless they are
```
larseggert: s/ignore these bits/ignore these bits on receipt/

draft-ietf-quic-transport.md
```diff
@@ -444,7 +444,10 @@ Google QUIC Demultipexing Bit:
...
+  participating in an experiment that uses these bits.  An endpoint not
+  actively using these bits SHOULD set the value randomly to protect against
```
larseggert: s/should set the value randomly/should randomize these bits on transmit/

PR #1520, opened 2018-07-04
ACK_ECN type value is not incremented

ami-GS:
ACK_ECN type should be 0x1a if you expect the type to be incremented by 1. Or is this(0x20) expected value?


janaiyengar:
@martinthomson : I think this is fine, given that I don't expect that anyone has implemented the ACK_ECN frame yet. We can reduce later pain if we take this in now, I think. I'll merge.

PR #1526, opened 2018-07-05
Add extensibility section

### LINKED ISSUE #1524: Extensions in HTTP/QUIC
The text currently says that extensions can define stream types and settings.  It says that they'll need to register the frame types they define, but doesn't actually say they're permitted to do so.  (I guess we consider that implied.)  It doesn't permit them to define pseudoheaders (e.g. `:protocol`).

We probably need the equivalent language of 7540 which permits them to "alter any aspect of the protocol" subject to certain constraints, and spell out those constraints.

MikeBishop:
Fixes #1524 by largely cribbing the RFC7540 text.  I believe this is what we've all been assuming all along; if not, scream now.

(Assumes presence of, but doesn't build on, #1525.)

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1197,6 +1197,44 @@ HTTP_MALFORMED_FRAME (0x01XX):
   be indicated with the code (0x10D).
 
 
+# Extensions to HTTP/QUIC
+
+HTTP/QUIC permits extension of the protocol.  Within the limitations described
+in this section, protocol extensions can be used to provide additional services
+or alter any aspect of the protocol.  Extensions are effective only within the
+scope of a single HTTP/QUIC connection.
+
+This applies to the protocol elements defined in this document.  This does not
+affect the existing options for extending HTTP, such as defining new methods,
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -1197,6 +1197,44 @@ HTTP_MALFORMED_FRAME (0x01XX):
...
+be negotiated before being used.  For example, an extension that changes the
+layout of the HEADERS frame cannot be used until the peer has given a positive
+signal that this is acceptable. In this case, it could also be necessary to
+coordinate when the revised layout comes into effect.
+
+This document doesn't mandate a specific method for negotiating the use of an
+extension but notes that a setting ({{settings-parameters}}) could be used for
+that purpose.  If both peers set a value that indicates willingness to use the
+extension, then the extension can be used.  If a setting is used for extension
+negotiation, the initial value MUST be defined in such a fashion that the
```
martinthomson: s/initial/default/ - hq has a slightly different notion of how settings apply than h2.

PR #1529, opened 2018-07-06
HPACK errors should be QPACK

martinthomson:


MikeBishop:
Actually, these should probably have moved to QPACK along with the frame definitions.

PR #1532, opened 2018-07-06
Move HTTP_*PACK_COMPRESSION_FAILED to QPACK

MikeBishop:
Closes #1529; finishes what #1358 started.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -817,6 +817,15 @@ represented as an 8-bit prefix string literal.
 {: title="Literal Header Field Without Name Reference"}
 
 
+# Error Handling
+
+The following error code is defined for HTTP/QUIC to indicate all failures of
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -975,6 +984,21 @@ The entries in the following table are registered by this document.
...
+registry established in {{QUIC-HTTP}}.
+
+Name:
+: HTTP_QPACK_DECOMPRESSION_FAILED
+
+Code:
+: 0x06
+
+Description:
+: QPACK failed to decompress a frame and cannot continue.
```
afrind: I'm not sure 'frame' is the right term here.  This is basically any decoding error of any kind, which may occur outside of a frame.

MikeBishop: Good point -- just stale text.

PR #1533, opened 2018-07-06
Optimize definition IP addr in TP PreferredAddress

dtikhonov:
There are two reasons for this change.

First, having the IP address stored as an opaque string 4 to 255 octets
in length does not make it clear in what format the IP address is stored.
The choice of 4 and 16 bytes makes the format obvious.

Second, the longer of the two possibilities is 16 bytes.  There is no
reason to allow for longer values.  Only allowing 4 and 16 bytes make
the checks both stricter and simpler and implementations more efficient.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1445,18 +1445,10 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
    } TransportParameters;
 
    struct {
-     uint8[4] addr;
-   } IPv4addr;
-
-   struct {
-     uint8[16] addr;
-   } IPv6addr;
-
-   struct {
      enum { IPv4(4), IPv6(6), (15) } ipVersion;
      select (PreferredAddress.ipVersion) {
-       case IPv4: IPv4addr;
-       case IPv6: IPv6addr;
+       case IPv4: uint8 IPv4addr[4];
+       case IPv6: uint8 IPv6addr[16];
      };
      uint16 port;
      opaque connectionId<0..18>;
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1445,8 +1445,19 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
    } TransportParameters;
 
    struct {
+     uint8[4] addr;
+   } IPv4addr;
+
+   struct {
+     uint8[16] addr;
+   } IPv6addr;
+
+   struct {
      enum { IPv4(4), IPv6(6), (15) } ipVersion;
-     opaque ipAddress<4..2^8-1>;
+     select (PreferredAddress.ipVersion) {
+       case IPv4: IPv4addr;
+       case IPv6: IPv6addr;
+     };
      uint16 port;
      opaque connectionId<0..18>;
      opaque statelessResetToken[16];


dtikhonov:
The proposed change does more than simply make the IP address length implicit.  Tying length to the IP version makes it apparent what the format of the IP address is without any additional verbiage.  Such verbiage is not present in the current version of the Draft.

As it stands, one could imagine (however perversely) that the IP address is a NUL-terminated ASCII string, such as "::1" (hence the minimum of 4 bytes), "192.168.100.25", or "2001:0db8:85a3:0000:0000:8a2e:0370:7334".

mikkelfj:
Like http://doc.cat-v.org/plan_9/4th_edition/papers/net/?

draft-ietf-quic-transport.md
```diff
@@ -1444,9 +1444,20 @@ language from Section 3 of {{!I-D.ietf-tls-tls13}}.
       TransportParameter parameters<22..2^16-1>;
    } TransportParameters;
 
+   struct {
+     uint8[4] addr;
+   } IPv4addr;
+
+   struct {
+     uint8[16] addr;
+   } IPv6addr;
```
martinthomson: You can use a typedef for this:

```
typedef uint8[16] IPv6addr;
```

martinthomson:
You can't assume that you know the IP address.  The server might want to advertise a preferred address in a version other than the current one.  For instance, you reach the server over v4, but it would rather you use v6.  You can probably assume that people understand that v6 means 16 octets, but is that true for IPv10?

Yes, you can ignore the entire transport parameter if you hit an unknown version.  That's a little unusual for TLS though.

dtikhonov:
> You can probably assume that people understand that v6 means 16 octets, but is that true for IPv10?

This reasoning seems far-fetched.  By the same logic, we should not limit the IP version to 15 (see the enum maximum value), nor the port number to 65535.

If the idea is to be able to skip unknown IP address blobs, so to speak, then the format of the IPv4 and IPv6 blobs should be specified somehow -- either using the TLS presentation language or English.

draft-ietf-quic-transport.md
dtikhonov: There are no typedefs in the [TLS 1.3 draft](https://tools.ietf.org/html/draft-ietf-tls-tls13-28).

ekr: The way you say this in TLS is:

```
uint8 IPv6addr[16];
```

dtikhonov: Ah yes, this is much simpler, thanks.  Done in c2647e37ea6af229d201b542bb2a8f372f872440.

MikeBishop:
@dtikhonov, the IP version is limited to 15 because that's the space in which 4 and 6 are allocated.  https://www.iana.org/assignments/version-numbers/version-numbers.xhtml.  However, I'm sympathetic to the argument that it should be implicit because if you don't know the specified version, being able to figure out how long the address is seems pointless.

Or are you suggesting that a future version might have a variable-length address, for which an explicit length would be required?

dtikhonov:
I don't understand why the build [failed](https://circleci.com/gh/quicwg/base-drafts/6042?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).  Works for me when I `make` 😕 .

mikkelfj:
Just say that the encoding is big endian in same encoding as the UDP/4 and UDP/6 header formats with 4 and 16 bytes respetively.

MikeBishop:
Build failure isn't you.  One of the references in TLS has stopped working.  It probably works for you either because you're specifically building just this doc or because your .refcache contains a version of the reference that's sufficiently fresh that it's not trying to refresh.

dtikhonov:
There seems to be no support for this change.  I will close this PR lest a champion step forward.

PR #1537, opened 2018-07-08
Fix bit prefix verbiage to match figure

dtikhonov:


MikeBishop:
No, the existing text is correct and the change is incorrect.  The `H` bit is part of the string prefix, so the string prefix is one bit longer than the string's length field.  See the definition of prefixed strings in 5.1.2.

However, I hate how confusing that terminology is, so if you'd like to propose an improvement to that, *please*.  ;-)

dtikhonov:
Sure confused me!  I'll have to reread it.

martinthomson:
Right.  Perhaps it would be better to say "a string literal with an N-bit-prefixed length" or something like that.

dtikhonov:
Having read the document again, I think it is fine as-is. The prefix is described in the _String Literals_ section which comes before the operation definitions and the accompanying figures.

Closing.

PR #1539, opened 2018-07-09
Remove UNSOLICITED_PATH_RESPONSE

### LINKED ISSUE #1265: error handling for path_response
>   If the content of a PATH_RESPONSE frame does not match the content of
>   a PATH_CHALLENGE frame previously sent by the endpoint, the endpoint
  > MAY generate a connection error of type UNSOLICITED_PATH_RESPONSE.

I think this is a footgun - we should just say MUST ignore.

PATH_RESPONSE is not sent reliably but PATH_CHALLENGES are regenerated as needed. So you've got a non-deterministic system because you have to timeout things.

If we want to keep the error then I think these frames ought to be considered reliable so a deterministic call can be made.

But I think the better route is to just ignore a RESPONSE you're not going to do anything with.

martinthomson:
Giving this error a special code creates an emphasis on the error
condition that is entirely inappropriate.  This is a tricky error code
to generate correctly, but giving it a special number implies that it
needs special handling.  That's not really the case, so using
PROTOCOL_VIOLATION is plenty.

Closes #1265.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3293,7 +3293,7 @@ frame.  Its format is identical to the PATH_CHALLENGE frame
 
 If the content of a PATH_RESPONSE frame does not match the content of a
 PATH_CHALLENGE frame previously sent by the endpoint, the endpoint MAY generate
-a connection error of type UNSOLICITED_PATH_RESPONSE.
+a connection error of type PROTOCOL_VIOLATION.
 
 
 ## NEW_TOKEN frame {#frame-new-token}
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3293,7 +3293,7 @@ frame.  Its format is identical to the PATH_CHALLENGE frame
 
 If the content of a PATH_RESPONSE frame does not match the content of a
 PATH_CHALLENGE frame previously sent by the endpoint, the endpoint MAY generate
-a connection error of type UNSOLICITED_PATH_RESPONSE.
+a connection error of type PROTOCOL_VIOLATION.
```
ianswett: Should this specify the frame type as PATH_RESPONSE now that frame type is present in CONNECTION_CLOSE?

martinthomson: Yes, but I didn't think that it would be necessary to state that explicitly.  If we decide to do that, I'd want to do that everywhere.

ianswett: Agreed, we might want to state it everywhere.

PR #1540, opened 2018-07-09
PADDING counts towards bytes in flight

### LINKED ISSUE #1531: Only Update max_ack_delay for Retransmittable Packets
I came across this pseudocode while debugging an issue in my implementation related to it:

```
  UpdateRtt(latest_rtt, ack_delay):
    // min_rtt ignores ack delay.
    min_rtt = min(min_rtt, latest_rtt)
      latest_rtt -= ack_delay
      // Only save into max ack delay if it's used
      // for rtt calculation and is not ack only.
      if (!sent_packets[ack.largest_acked].ack_only)
        max_ack_delay = max(max_ack_delay, ack_delay)
```

Should this be ignoring all packets that don't contain any rentrasmittable frames, instead of just `ack_only` packets?

ianswett:
Clarifies the recovery draft in light of #1517 and fixes #1531

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -108,7 +108,7 @@ of frames contained in a packet affect recovery and congestion control logic:
 * All packets are acknowledged, though packets that contain only ACK,
   ACK_ECN, and PADDING frames are not acknowledged immediately.  Packets
   containing at least one frame besides ACK, ACK_ECN, and PADDING are referred
-  to as retransmittable below.
+  to as "retransmittable" below.
 
 * Long header packets that contain CRYPTO frames are critical to the
   performance of the QUIC handshake and use shorter timers for acknowledgement
@@ -117,7 +117,8 @@ of frames contained in a packet affect recovery and congestion control logic:
 * Packets that contain only ACK and ACK_ECN frames do not count toward
   congestion control limits and are not considered in-flight. Note that this
   means PADDING frames cause packets to contribute toward bytes in flight
-  without directly causing an acknowledgment to be sent.
+  without directly causing an acknowledgment to be sent.  The rest of this
+  document uses "ACK frames" to refer to both ACK and ACK_ECN frames.
 
 ## Relevant Differences Between QUIC and TCP
 
@@ -583,7 +584,7 @@ sent_packets:
 
 : An association of packet numbers to information about them, including a number
   field indicating the packet number, a time field indicating the time a packet
-  was sent, a boolean indicating whether the packet is ack only, a boolen
+  was sent, a boolean indicating whether the packet is ack only, a boolean
   indicating whether it counts towards bytes in flight, and a bytes
   field indicating the packet's size.  sent_packets is ordered by packet number,
   and packets remain in sent_packets until acknowledged or lost.  A sent_packets
@@ -625,7 +626,7 @@ are as follows:
 
 * packet_number: The packet number of the sent packet.
 
-* ack_only: A boolean that indicates whether a packet only contains an
+* ack_only: A boolean that indicates whether a packet contains only
   ACK or PADDING frame(s).  If true, it is still expected an ack will
   be received for this packet, but it is not retransmittable.
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1026,7 +1026,8 @@ ecn_ce_counter:
 bytes_in_flight:
 : The sum of the size in bytes of all sent packets that contain at least
   one retransmittable or PADDING frame, and have not been acked or declared
-  lost. The size does not include IP or UDP overhead.
+  lost. The size does not include IP or UDP overhead, but does include the
+  QUIC header and AEAD overhead.
   Packets only containing ACK frames do not count towards bytes_in_flight
   to ensure congestion control does not impede congestion feedback.
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -775,7 +775,7 @@ Pseudocode for SetLossDetectionAlarm follows:
 
 ~~~
  SetLossDetectionAlarm():
-    // Don't arm the alarm if there are no retransmittable packets
+    // Don't arm alarm if there are no retransmittable packets
     // in flight.
     if (bytes_in_flight == 0):
       loss_detection_alarm.cancel()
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -1018,7 +1025,7 @@ ecn_ce_counter:
 
 bytes_in_flight:
 : The sum of the size in bytes of all sent packets that contain at least
-  one retransmittable frame, and have not been acked or declared
+  one retransmittable or PADDING frame, and have not been acked or declared
   lost. The size does not include IP or UDP overhead.
```
mikkelfj: But presumably it does include AEAD tags?

draft-ietf-quic-recovery.md
```diff
@@ -110,6 +110,13 @@ important to the loss detection and congestion control machinery below.
 * Retransmittable packets are those that contain at least one retransmittable
   frame.
 
+* PADDING frames do cause packets containing them to count towards bytes in
+  flight, but do not instigate acknowledgement.
```
MikeBishop: Instigate feels like an unusual word choice here.  Not necessarily incorrect, but unusual.

martinthomson: What Mike said, but I think that it would be better to concentrate on the more important distinctions and perhaps rewrite this bit.  Your bullet list is now 6 items, which is too long anyway, so perhaps this could be reduced:

QUIC packets can contain multiple frames of different types.  The recovery mechanisms ensure that frames that need reliable delivery are acknowledged and sent in new packets as necessary.  The types of frames contained in a packet can affect recovery and congestion control logic:

* All packets are acknowledged, though packets that contain only ACK, ACK_ECN, and PADDING frames are not acknowledged immediately.

* Packets that contain CRYPTO frames are critical to the performance of the QUIC handshake and use shorter timers for acknowledgement and retransmission.

* Packets that contain only ACK and ACK_ECN frames do not count toward congestion control limits (see {{blah}}).  That is, they do not contribute toward bytes in flight.  Note that this means that PADDING frames cause packets to contribute toward bytes in flight without directly causing an acknowledgment to be sent. 

(This last point might be accomplished better with a forward reference, but I didn't see a nice place to put that, the definition of bytes in flight is buried a little.)

martinthomson: Bump for this comment.  I think that it might be a better framing still, after some adjustments.

ianswett:
I think there are more cleanups to be done, possibly by adding a terminology section for bytes_in_flight, etc, but I'd like to do that in a separate PR if it's a good idea.

draft-ietf-quic-recovery.md
```diff
@@ -100,30 +100,23 @@ This design obviates the need for disambiguating between transmissions and
...
+acknowledged or declared lost and sent in new packets as necessary. The types
+of frames contained in a packet affect recovery and congestion control logic:
+
+* All packets are acknowledged, though packets that contain only ACK,
+  ACK_ECN, and PADDING frames are not acknowledged immediately.  Packets
+  containing at least one frame besides ACK, ACK_ECN, and PADDING are referred
+  to as retransmittable below.
+
+* Packets that contain CRYPTO frames are critical to the performance of the
+  QUIC handshake and use shorter timers for acknowledgement and retransmission.
```
MikeBishop: Technically, this isn't (or need not be) true once you've reached 1-RTT.  Are NewSessionTicket messages treated with equally high priority?

draft-ietf-quic-recovery.md
```diff
@@ -110,8 +110,9 @@ of frames contained in a packet affect recovery and congestion control logic:
   containing at least one frame besides ACK, ACK_ECN, and PADDING are referred
   to as retransmittable below.
 
-* Packets that contain CRYPTO frames are critical to the performance of the
-  QUIC handshake and use shorter timers for acknowledgement and retransmission.
+* Long header ackets that contain CRYPTO frames are critical to the performance
```
MikeBishop: **p**ackets

draft-ietf-quic-recovery.md
```diff
@@ -100,23 +100,24 @@ This design obviates the need for disambiguating between transmissions and
...
-  frame.
+* Long header packets that contain CRYPTO frames are critical to the
+  performance of the QUIC handshake and use shorter timers for acknowledgement
+  and retransmission.
 
-* Cryptographic handshake data is sent in CRYPTO frames, and uses the
-  reliability machinery of QUIC underneath.
-
-* ACK and ACK_ECN frames contain acknowledgment information. ACK_ECN frames
-  additionally contain information about ECN codepoints seen by the peer.  (The
```
janaiyengar: This sentence in the parenthesis is necessary for subsequent uses of the term "ACK frames"

draft-ietf-quic-recovery.md
```diff
@@ -100,23 +100,24 @@ This design obviates the need for disambiguating between transmissions and
...
+acknowledged or declared lost and sent in new packets as necessary. The types
+of frames contained in a packet affect recovery and congestion control logic:
 
-* Retransmittable frames are those that count towards bytes in flight and need
-  acknowledgement.  The most common are STREAM frames, which typically contain
-  application data.
+* All packets are acknowledged, though packets that contain only ACK,
+  ACK_ECN, and PADDING frames are not acknowledged immediately.  Packets
+  containing at least one frame besides ACK, ACK_ECN, and PADDING are referred
+  to as retransmittable below.
```
janaiyengar: put "retransmittable" in quotes, for clarity

draft-ietf-quic-recovery.md
```diff
@@ -582,7 +583,8 @@ sent_packets:
 
 : An association of packet numbers to information about them, including a number
   field indicating the packet number, a time field indicating the time a packet
-  was sent, a boolean indicating whether the packet is ack only, and a bytes
+  was sent, a boolean indicating whether the packet is ack only, a boolen
```
janaiyengar: boolean

draft-ietf-quic-recovery.md
```diff
@@ -623,9 +625,12 @@ are as follows:
 
 * packet_number: The packet number of the sent packet.
 
-* is_ack_only: A boolean that indicates whether a packet only contains an
-  ACK frame.  If true, it is still expected an ack will be received for
-  this packet, but it is not retransmittable.
+* ack_only: A boolean that indicates whether a packet only contains an
```
janaiyengar: s/only contains an/contains only/

PR #1549, opened 2018-07-11
Fixup implicit stream opening

### LINKED ISSUE #1342: Implicit stream open
We've had this on and off, but it doesn't seem like we've ever concluded.

Arguments for both sides claim simplicity.  It all depends on perspective apparently.

The material difference is that an implicitly opened stream can be used to send.  e.g., if a server receives a STREAM frame for stream 8, then it can now send on stream 4.

martinthomson:
This was already in pretty good shape.  I added something to the state
machine and tweaked the critical text slightly.

Closes #1342.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3970,7 +3970,7 @@ Receiving a MAX_STREAM_DATA frame implies that the remote peer has opened the
 stream and is providing flow control credit.  A MAX_STREAM_DATA frame might
 arrive before a STREAM or STREAM_BLOCKED frame if packets are lost or reordered.
 
-Before creating a stream, all lower-numbered stream of the same type MUST be
+Before creating a stream, all lower-numbered streams of the same type MUST be
 created.  That means that receipt of a frame that would open a stream causes all
 lower-numbered streams of the same type to be opened in numeric order.  This
 ensures that the creation order for streams is consistent on both endpoints.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3971,8 +3971,10 @@ stream and is providing flow control credit.  A MAX_STREAM_DATA frame might
 arrive before a STREAM or STREAM_BLOCKED frame if packets are lost or reordered.
 
 In addition to the exchange of frames related to the stream, a stream is created
-before any higher-numbered stream of the same type can be created.  This ensures
-that the creation order for streams is consistent on both endpoints.
+before any higher-numbered stream of the same type can be created.  That is,
+receipt of a frame that would open a stream causes all lower-numbered streams of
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3791,9 +3791,9 @@ The two type bits from a Stream ID therefore identify streams as summarized in
 
 The first bi-directional stream opened by the client is stream 0.
 
-A QUIC endpoint MUST NOT reuse a Stream ID.  Streams can be used in any order.
-Streams that are used out of order result in opening all lower-numbered streams
-of the same type in the same direction.
+A QUIC endpoint MUST NOT reuse a Stream ID.  Streams of each type are created in
+numeric order.  Streams that are used out of order result in opening all
+lower-numbered streams of the same type in the same direction.
```
mikkelfj: It is a bit unclear what the difference is between "use", "create", "send", and "receive". I read such that "create" has to do with allocating a stream ID at or below the MAX ID but not necessarily sending. So a stream can be created silently and later some lower number stream can have its first frame sent - doing so accounts for a "use". For the receiver "use" either means "receive" or it means keep received frames buffered and pretend it is not there, but eventually "use" one.

I realize there is nothing enforceable about these distinctions, but it might affect buffering strategies.

draft-ietf-quic-transport.md
```diff
@@ -3969,6 +3970,10 @@ Receiving a MAX_STREAM_DATA frame implies that the remote peer has opened the
 stream and is providing flow control credit.  A MAX_STREAM_DATA frame might
 arrive before a STREAM or STREAM_BLOCKED frame if packets are lost or reordered.
 
+In addition to the exchange of frames related to the stream, a stream is created
```
ianswett: I find this paragraph a bit confusing.  Are you trying to say that if a frame for a new stream arrives, all lower numbered unopened streams of the same type must be opened?

ianswett: How about "Before creating a higher numbered stream, all lower numbered streams of the same type MUST be created." ?

draft-ietf-quic-transport.md
```diff
@@ -3970,11 +3970,10 @@ Receiving a MAX_STREAM_DATA frame implies that the remote peer has opened the
 stream and is providing flow control credit.  A MAX_STREAM_DATA frame might
 arrive before a STREAM or STREAM_BLOCKED frame if packets are lost or reordered.
 
-In addition to the exchange of frames related to the stream, a stream is created
-before any higher-numbered stream of the same type can be created.  That is,
-receipt of a frame that would open a stream causes all lower-numbered streams of
-the same type to be opened first.  This ensures that the creation order for
-streams is consistent on both endpoints.
+Before creating a stream, all lower-numbered stream of the same type MUST be
```
ianswett: streams

PR #1552, opened 2018-07-12
Allow Retry in response to 0-RTT packets

### LINKED ISSUE #1547: Retry in response to 0-RTT
It's entirely possible that a box that might be inclined to send a Retry packet could receive 0-RTT packets before an Initial packet.  Right now, that box has to drop the 0-RTT, it is prohibited from sending a Retry.

There doesn't seem to be a strong reason to prohibit sending Retry here.  Nor is there any strong reason to disallow it.  

In favour of the existing text are some weak arguments:
* Allowing Retry here means that the number of Retry packets in flight might exceed the number of Initial packets.  But we already have text about ignoring old and redundant Retry packets.
* A server that is sending Retry does so primarily to avoid taking on extra work and dropping the 0-RTT and waiting for an Initial is the easiest thing to do.

In favour of the change are equally weak arguments:
* This would allow Retry to be faster in cases where an Initial is reordered or dropped.
* A Retry sent in response to 0-RTT is indistinguishable from one sent in response to Initial packets.

Did I miss anything?

martinthomson:
Closes #1547.

(Note the target of this PR.  This is blocked on #1498.)

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -595,8 +595,9 @@ A Retry packet does not include a packet number and cannot be explictly
 acknowledged by a client.
 
 A server MUST NOT send a Retry in response to packets other than Initial
-or 0-RTT packets.  A server MAY discard 0-RTT packets and only send Retry in
-response to Initial packets.
+or 0-RTT packets.  A server MAY choose to only send Retry in response to Initial
+packets and discard or buffer 0-RTT packets corresponding to unvalidated client
+addresses.
 
 If the Original Destination Connection ID field does not match the Destination
 Connection ID from the most recent Initial packet it sent, clients MUST discard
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -594,7 +594,9 @@ of subsequent packets that it sends.
 A Retry packet does not include a packet number and cannot be explictly
 acknowledged by a client.
 
-A server MUST only send a Retry in response to a client Initial packet.
+A server MUST NOT send a Retry in response to packets other than Initial
+or 0-RTT packets.  A server MAY discard 0-RTT packets and only send Retry in
+response to Initial packets.
```
ekr: I would invert this last sentence:

"A server MAY choose to send Retry in response to Initial packets and discard or buffer 0-RTT packets corresponding to unvalidated client addresses."

martinthomson: Yes, that it better.  Thanks.

PR #1557, opened 2018-07-14
Add Definitions

ianswett:
Add a separate definitions section, similar to the transport doc.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -88,30 +88,30 @@ when, and only when, they appear in all capitals, as shown here.
 
 Definitions of terms that are used in this document:
 
-Sender:
+ACK frames:
 
-: The endpoint sending QUIC packets.
-
-Receiver:
-
-: The endpoint receiving QUIC packets and sending acknowledgements in
-  response.
+: ACK frames refer to both ACK and ACK_ECN frames in this
+  document.
 
 ACK-only:
 
-: Any packet containing only an ACK or ACK_ECN frame. The rest of this
-  document uses "ACK frames" to refer to both ACK and ACK_ECN frames.
+: Any packet containing only an ACK frame.
 
 In-flight:
 
-: Packets are considered in-flight because they have been sent
-  and neither acknowledged or considered lost, and they are not
+: Packets are considered in-flight when they have been sent
+  and neither acknowledged or declared lost, and they are not
   ACK-only.
 
-Retransmittable:
+Retransmittable Frames:
 
-: Packets that contain frames besides ACK, ACK_ECN, or PADDING elicit
-  an ACK from the receiver and are called retransmittable.
+: All frames besides ACK, ACK_ECN, or PADDING are considered
+  retransmittable.
+
+Retransmittable Packets:
+
+: Packets that contain retransmittable frames elicit an ACK from
+  the receiver and are called retransmittable packets.
 
 
 # Design of the QUIC Transmission Machinery
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -79,12 +79,39 @@ recovery, and where applicable, attributes the TCP equivalent in RFCs,
 Internet-drafts, academic papers, and/or TCP implementations.
 
 
-## Notational Conventions
+# Conventions and Definitions
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
 document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
 when, and only when, they appear in all capitals, as shown here.
+
+Definitions of terms that are used in this document:
+
+Sender:
+
+: The endpoint sending QUIC packets.
+
+Receiver:
+
+: The endpoint receiving QUIC packets and sending acknowledgements in
+  response.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -79,13 +79,40 @@ recovery, and where applicable, attributes the TCP equivalent in RFCs,
...
+
+Sender:
+
+: The endpoint sending QUIC packets.
+
+Receiver:
+
+: The endpoint receiving QUIC packets and sending acknowledgements in
+  response.
+
```
janaiyengar: I don't think we need the above ones. In fact they may be more confusing than useful (sender of acks, for instance)

draft-ietf-quic-recovery.md
```diff
@@ -79,13 +79,40 @@ recovery, and where applicable, attributes the TCP equivalent in RFCs,
...
+ACK-only:
```
janaiyengar: I think this includes PADDING. Also, you'll want to fix all uses in the doc (ack-only, ack only)

draft-ietf-quic-recovery.md
```diff
@@ -79,13 +79,40 @@ recovery, and where applicable, attributes the TCP equivalent in RFCs,
...
+
+: Any packet containing only an ACK or ACK_ECN frame. The rest of this
+  document uses "ACK frames" to refer to both ACK and ACK_ECN frames.
```
janaiyengar: This second sentence doesn't seem to fit in this definition. I would separate it out into a new definition for now.

draft-ietf-quic-recovery.md
```diff
@@ -79,13 +79,40 @@ recovery, and where applicable, attributes the TCP equivalent in RFCs,
...
+
+In-flight:
+
+: Packets are considered in-flight because they have been sent
+  and neither acknowledged or considered lost, and they are not
```
janaiyengar: s/or/nor/

draft-ietf-quic-recovery.md
```diff
@@ -79,13 +79,40 @@ recovery, and where applicable, attributes the TCP equivalent in RFCs,
...
+  response.
+
+ACK-only:
+
+: Any packet containing only an ACK or ACK_ECN frame. The rest of this
+  document uses "ACK frames" to refer to both ACK and ACK_ECN frames.
+
+In-flight:
+
+: Packets are considered in-flight because they have been sent
```
janaiyengar: s/because/when/

draft-ietf-quic-recovery.md
```diff
@@ -79,13 +79,40 @@ recovery, and where applicable, attributes the TCP equivalent in RFCs,
...
+  and neither acknowledged or considered lost, and they are not
+  ACK-only.
+
+Retransmittable:
+
+: Packets that contain frames besides ACK, ACK_ECN, or PADDING elicit
+  an ACK from the receiver and are called retransmittable.
```
janaiyengar: I think you should define retransmittable frames, and packets containing a retransmittable frame as a retransmittable packet.

draft-ietf-quic-recovery.md
ianswett: Hmm, I think it's or.  But I changed considered to declared.

draft-ietf-quic-recovery.md
```diff
@@ -79,13 +79,40 @@ recovery, and where applicable, attributes the TCP equivalent in RFCs,
...
+
+ACK frames:
+
+: ACK frames refer to both ACK and ACK_ECN frames in this
+  document.
+
+ACK-only:
+
+: Any packet containing only an ACK frame.
+
```
mikkelfj: Suggest being specific about ACK and ACK_ECN here

draft-ietf-quic-recovery.md
```diff
@@ -79,13 +79,40 @@ recovery, and where applicable, attributes the TCP equivalent in RFCs,
...
+Retransmittable Frames:
+
+: All frames besides ACK, ACK_ECN, or PADDING are considered
+  retransmittable.
+
+Retransmittable Packets:
+
+: Packets that contain retransmittable frames elicit an ACK from
+  the receiver and are called retransmittable packets.
```
martinthomson: Both packets and frames aren't classified as retransmittable or not.  It would be better if this used a different label.

janaiyengar: @martinthomson : I don't understand your comment. Can you rephrase?

martinthomson: The notion that a packet is retransmittable is a misnomer because it creates the impression that you can retransmit the packet.  Similarly, we don't retransmit frames either.  The distinction that is relevant to your uses of these terms is not whether the packet is able to be retransmitted, but whether the packet contains frames other than ACK/ACK_ECN/PADDING.

ianswett: I opened issue #1595 for splitting up the terms, but I'd prefer do that separately.

draft-ietf-quic-recovery.md
```diff
@@ -79,13 +79,40 @@ recovery, and where applicable, attributes the TCP equivalent in RFCs,
...
+  document.
+
+ACK-only:
+
+: Any packet containing only an ACK or ACK_ECN frame.
+
+In-flight:
+
+: Packets are considered in-flight when they have been sent
+  and neither acknowledged or declared lost, and they are not
```
MikeBishop: Late to the party, but:

Either/or, but Neither/nor

ianswett: Thanks Mike, I fixed it in master

PR #1555, opened 2018-07-12
Exception in ECN detection when ACK loss happens

gloinul:
Adding an exception to the ECN detection to handle ACK loss if that occurs. 

Fixes Issue #1481

PR #1565, opened 2018-07-14
Simplify receiver verification of ECN counters

### LINKED ISSUE #1481: Fragility in ECN counters with lost acks
The text for ECN says: "The increase in ECT(0) and ECT(1) counters MUST be no greater than the number of packets newly acknowledged that were sent with the corresponding codepoint."

As @gloinul observed in his PR and @martinthomson noted in his review, this can cause ECN to be disabled when acks are lost. A sender may never receive an ack for a packet that was in fact acknowledged, and might see an increase in ECN counters without a corresponding packet being newly acked. This should be addressed.

@martinthomson writes: "Say a marked packet gets through and gets acknowledged, but the acknowledgment is lost. And the recipient abandons the ACK. The counters will increase, but the number of newly acknowledged packets will be lower than the increase to the counters. That suggests a tweak, though it complicates things a little:

If the acknowledgments cover a range of packet numbers that extend past (or adjacent to) the largest acknowledged packet number that was previously acknowledged, then the counters must match precisely. If the acknowledgments do not extend to the largest previously acknowledged packet number, then the numbers MAY increase by the number of unacknowledged packets between the previously largest acknowledged and the newly lowest acknowledged.

That is, if an endpoint has seen an ACK_ECN for packet 10, then receives a new ACK_ECN for packet 20-30, it needs to allow for an increase in the counters for packets 11 through 19 (inclusive). It can then reset both the largest acknowledged and the values it tracks for each counter."

janaiyengar:
This mechanism modifies the response proposed in #1555. It isn't perfect, but it's straightforward, and addresses the fragility problem adequately I think.

Fixes #1481.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1794,9 +1794,10 @@ not support ECN.
 
 If an endpoint sets ECT codepoints on outgoing packets and encounters a
 retransmission timeout due to the absence of acknowledgments from the peer (see
-{{QUIC-RECOVERY}}), the endpoint MAY cease setting ECT codepoints in subsequent
-packets. Doing so allows the connection to traverse network elements that drop
-packets carrying ECT or CE codepoints in the IP header.
+{{QUIC-RECOVERY}}), or if an endpoint has reason to believe that a network
+element might be corrupting ECN codepoints, the endpoint MAY cease setting ECT
+codepoints in subsequent packets. Doing so allows the connection to traverse
+network elements that drop or corrupt ECN codepoints in the IP header.
 
 
 ## Proof of Source Address Ownership {#address-validation}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1755,7 +1755,8 @@ counters in subsequent (see {{processing-and-ack}}) ACK_ECN frames (see
 {{frame-ack-ecn}}).
 
 A packet detected by a receiver as a duplicate does not affect the receiver's
-local ECN codepoint counts to mitigate security concerns ({{security-ecn}}).
+local ECN codepoint counts; see ({{security-ecn}}) for relevant security
+concerns.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1771,20 +1772,18 @@ the expectation that either the network or the peer no longer supports ECN.
 To protect the connection from arbitrary corruption of ECN codepoints by the
 network, an endpoint verifies the following when an ACK_ECN frame is received:
 
-* The total increase in ECT(0), ECT(1), and CE counters reported in the ACK_ECN
-  frame MUST be equal to the total number of packets newly acknowledged in this
-  ACK_ECN frame.
-
-* The increase in ECT(0) and ECT(1) counters MUST be no greater than the number
+* The increase in ECT(0) and ECT(1) counters MUST be no lesser than the number
```
martinthomson: "no lesser" isn't a turn of phrase I'm comfortable with.  "MUST be at least the number of newly acknowledged packets"

draft-ietf-quic-transport.md
```diff
@@ -1771,20 +1772,18 @@ the expectation that either the network or the peer no longer supports ECN.
...
 
-* If the increase in the ECT(0), ECT(1) and CE counters are greater than
-  the number of newly acknowledged packets, and the received ACK's
-  smallest acknowledged is larger than the previous largest acknowledged,
-  then it's likely acknowledgements were missed, and the above comparison
-  MUST NOT be performed. Instead a new comparison point is stored by the
-  sender so that only changes after this point will be used in the future
-  comparisons.
+* The total increase in ECT(0), ECT(1), and CE counters reported in the ACK_ECN
+  frame MUST be no lesser than the total number of packets newly acknowledged in
```
martinthomson: ditto

janaiyengar:
I've taken in the suggestion. Merging.

PR #1581, opened 2018-07-18
Change early retransmit timer to 1/8th

### LINKED ISSUE #945: Early retransmit threshold is different from time-loss reordering threshold
delay_until_lost is currently described as 

> 9/8 * max(latest_rtt, smoothed_rtt) for early retransmit

however in the text it mentions that we should wait 1/4 of an rtt for early retransmit.

So it seems like the calculation should really be

`delay_until_lost = 1/4 * max(latest_rtt, smoothed_rtt)`

since effectively

> loss_time = unacked.time_sent + delay_until_lost

Similarly time loss detection should also be adjusted to not be 

> (1 + time_reordering_fraction) *  max(latest_rtt, smoothed_rtt)

but 

` time_reordering_fraction * max(latest_rtt, smoothed_rtt)`

ianswett:
Fixes #945


siyengar:
lgtm 👍 . Ship it

PR #1582, opened 2018-07-18
Remove packets from flight when RTO is verified

### LINKED ISSUE #935: RTO verified behavior
While implementing loss recovery from the draft, found some more weird behavior. When we get ACK for 2 packets sent after RTO, and RTO is verified, it seems like the psuedo code suggests that we use the reordering threshold again, however if we don't find a lost packet we would set the early retransmission timer again. This seems a bit weird that we would need to use the early retransmit timer after the RTO is verified and we could remove all the unacked packets < largest_sent_before_rto instead. Is this behavior intentional?

ianswett:
Fixes #935

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -729,7 +729,7 @@ Pseudocode for OnPacketAcked follows:
      // was spurious.  Otherwise, inform congestion control.
      if (rto_count > 0 &&
          acked_packet.packet_number > largest_sent_before_rto)
-       OnRetransmissionTimeoutVerified()
+       OnRetransmissionTimeoutVerified(acket_packet.packet_number)
      handshake_count = 0
      tlp_count = 0
      rto_count = 0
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -1147,11 +1146,17 @@ are detected lost.
 ### On Retransmission Timeout Verified
 
 QUIC decreases the congestion window to the minimum value once the
-retransmission timeout has been verified.
+retransmission timeout has been verified and removes any packets
+sent before the newly acknowledged RTO packet.
 
 ~~~
-   OnRetransmissionTimeoutVerified()
+   OnRetransmissionTimeoutVerified(packet_number)
```
siyengar: i think OnPacketAcked() needs to be changed as well too pass in largest_sent packet right?

siyengar:
this looks good to me

siyengar:
cannot merge this unfortunately even though i'm a member

Issue #1583, opened 2018-07-18
qpack: control stream is not defined


dtikhonov:
"_The_ control stream" (with the definite article) is mentioned several times in the draft, but is never defined.

The control stream must be the Encoder Stream.  Should we `s/control stream/Encoder Stream/g`?

afrind:
Yes, let's replace references to control stream with the encoder or decoder stream as appropriate.

afrind:
Fixed by #1588

PR #1588, opened 2018-07-20
Fix: replace "control stream" with "Encoder Stream"

dtikhonov:
"control stream" is not defined and in all instances it was used to refer to the Encoder Stream.

Fixes issue #1583.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -173,7 +173,7 @@ The relative index begins at zero and increases in the opposite direction from
 the absolute index.  Determining which entry has a relative index of "0" depends
 on the context of the reference.
 
-On the control stream, a relative index of "0" always refers to the most
+On the Encoder Stream, a relative index of "0" always refers to the most
 recently inserted value in the dynamic table.  Note that this means the entry
 referenced by a given relative index will change while interpreting instructions
 on the encoder stream.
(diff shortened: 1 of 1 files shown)


dtikhonov:
```
[dmitri@ubuntu-dmitri base-drafts]$ grep '[Ee]ncoder [Ss]tream' draft-ietf-quic-qpack.txt
     5.2.  QPACK Encoder Stream  . . . . . . . . . . . . . . . . . .  11
   Entries are added by instructions on the Encoder Stream (see
   while interpreting instructions on the encoder stream.
   o  The encoder stream is a unidirectional stream of type "0x48"
   All table updates occur on the encoder stream.  Request streams and
5.2.  QPACK Encoder Stream
   The contents of the encoder stream are an unframed sequence of the
         | QPACK Encoder Stream | 0x48 | Section 5     | Both   |
   o  Removed framing from encoder stream (#1361,#1467)
```

Outside of tables and headers, _Encoder Stream_ is used once, while _encoder stream_ is used three times.  We should pick one and go with it.  I, too, think that Encoder Stream looks odd (as this is not XVIII century), so I'd vote for the lowercase version.

P.S. Why does not GitHub allow one to insert polls...

martinthomson:
Only one of the uppercase instances is not a title.

Either way, this seems like a matter of editorial discretion. Alan can make a call.

afrind:
I think I prefer lowercase as well

Issue #1413, opened 2018-06-05
Recovery incorrectly declares packets lost at higher encryption levels


ianswett:
Currently the recovery draft indicates you should use fast retransmit to declare packets lost, even if there's no evidence the peer can decrypt the packet.

A simple example is if the first client initial is sent with 0RTT and then the Initial is lost; Then the client Initial is retransmitted and acked, 0RTT packets will be declared lost via early and/or fast retransmit.

If QUIC sticks with a single packet number space, then encryption level will have to be added to each outstanding packet we'll have to not declare packets at an encryption level lost until at least one packet at that level has been acked.

huitema:
Don't you need to mark the packet level anyhow? In my implementation, I mark packet in the retransmission queue with their original encryption type, such as CI/0RTT/HS/1RTT. I need that to implement the acknowledgement checks (HS cannot ack 0RTT), and I use that to avoid spurious retransmission of 0RTT packet, pretty much the way you describe.

mikkelfj:
Could that be delaying loss detection for peers that do not buffer packets they do not understand?

ianswett:
Yes, my implementation has encryption level as well and you need it somewhere in practice, but it's not documented in the recovery draft.

@mikkelfj Can you clarify the exact concern about delaying loss detection?

mikkelfj:
@ianswett My concern was mostly a question because I don't have the full overview, but the concern is for example that a server will not retransmit 1-RTT data for several roundtrips while it is waiting for client finished while the client is a small IoT device with limited buffering so it drops anything it cannot consume. I'm not stating this would be a problem, merely asking.

ianswett:
Yes, a server would not retransmit 1RTT data until the peer had the keys the ACK them.  As you said, a peer may drop packets it can't immediately process, so I don't know of a way to improve upon this case.

Issue #1018, opened 2017-12-16
Second flight DOS Attack with Client Clear Text packets.


huitema:
The spec says "Once the TLS handshake is complete, both peers MUST ignore unprotected packets.
 From that point onward, unprotected messages can be safely dropped." I wish that were true. But truthiness here depends of what you mean by "the TLS handshake is complete."

The current spec says that the second flight of the client is carried in a clear text handshake message. This was decided to align with TLS over TCP, because the proof of correctness is easier to establish if the transmission of the Finished message does not depend on the encryption key used for application messages. OK, maybe.

That rule means that if the clear text handshake message is lost, clients will repeat it. They will normally be using a different packet sequence number, so as to elicit a new acknowledgement -- because after all, the packet might have been received but the ACK may have been lost.

The problem is that third parties can also replay this "Finished" message. They can do that a number of times, using different packet sequence numbers, including numbers that the client might have used for its own data packets. Consider this:

1) Client sends Finished message with sequence number 5;
2) Adversary races Finished message with sequence number 6;
3) Client sends first HTTP command in 1-RTT message with sequence number 6.

This will probably cause the application to fail, as the server will "acknowledge" packet number 6 when it receives the duplicate "Finished" message, and ignore the 1-RTT packet because the packet sequence number has already be seen. 

In fact, the adversary does not need to race the first client packet. It could just as well use some higher numbers, or do that multiple times. If just one packet gets accepted, it will create a "hole" in some application stream. If the adversary feels lucky, it may not even need to place a copy of the Finished message there. Any old duplicate ACK might work just as well, as long that the server does not systematically reject the clear text messages.

The server will only be in a position to drop these incoming packets after receiving the "ACK of ACK" that shows that the acknowledgement of the finished packet has been received. That leaves a window of at least 1 RTT.

The simplest protection, IMHO, is to just start sending 1-RTT packets as soon as the keys have been negotiated, and to start rejecting clear text messages as soon as 1-RTT packets have been successfully received. That would require sending the second client flight using 1-RTT packets, instead of handshake packets.

There may be other solutions, such as using different sequence numbers for clear-text packets and protected packets. But that feels more complicated.

ianswett:
I would agree that sending 1-RTT packets as soon as they keys have been negotiated seems like the most robust solution.

martinthomson:
But that will also violate the conditions of the analysis of TLS.

[...]

PR #1591, opened 2018-07-20
Single Packet Number Space

### LINKED ISSUE #1579: Go Back to Single Packet Number Space
I think we should go back to a single packet number space, and allow for higher encryption levels/epochs to acknowledge packets sent/received in lower levels/epochs.

I have been working on the draft -13 updates for the last week or so, and having completely independent packet number spaces causes a lot of complexity. On the sender side, it requires separate loss detection logic per packet number space, though you still have one congestion control and recovery modules for the entire connection. Resolving the intersection of those two is a headache. On the receiver side, you need a separate tracker for all packets you have received in a given packet number space. At least in my implementation, that comes with a good bit more state than just a list/array of packet numbers. Then when it goes time to frame the acks for all these packet numbers, it's a mess to keep track of which types of packets need to be built, based on the data (both ACKs and CRYPTO) currently queued up.

The argument has been made, at least on the sender side, to use one contiguous set of packet numbers across all encryption levels/epochs. Yes, that does simply the loss detection logic somewhat, but doesn't help the receive and ACK side of things. Additionally, even if we were to mandate that behavior for everyone, the requirement that you ACK a packet in the same encryption level still causes a lot more complexity than the previous logic of always acknowledging in packets in the same (or greater) encryption level. It was a lot easier to implement the previous ACK model design, IMO.

Finally, for debugging purposes, it is really messy to see the same packet numbers, but at different encryption levels. It makes understanding exactly what's going on that much harder. It is a lot more natural to see packet numbers always increase for every packet send/received, independent of what encryption level is used.

Having all this extra complexity for just the handshake really seems like overkill. I understand and am all for the separate encryption levels, but I really think we should go back to a single packet number space and allow for ACK frames in higher encryption levels.

nibanks:
Fixes #1579 (see for more details on the problem).

- Moves back to a single packet number space across all encryption levels.
- Allows for a packet to be acknowledged in a greater than or equal encryption level.
- Initial packets cannot carry ACK frames, and therefore must be acknowledged in Handshake (or greater) encryption level.
- Handshake packets can acknowledge 0-RTT packets.

With these design changes, implementations do not have to track 3 separate packet number spaces (and all the overhead that goes with it). The number of QUIC packets (not UDP datagrams) required to be sent during the handshake is reduced (see Example Handshake Flows section). There is no Initial packet ACK spoofing since they must be acknowledged in a higher encryption level. 0-RTT packets can be acknowledged a bit quicker in the handshake process by using the Handshake encryption level; generally the first server ACK block will be able to ACK all Initial and 0-RTT packets it has received in one atomic operation.

Also makes sure #1018 and #1413 don't regress and are still solved.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -91,10 +91,8 @@ when, and only when, they appear in all capitals, as shown here.
 
 All transmissions in QUIC are sent with a packet-level header, which indicates
 the encryption level and includes a packet sequence number (referred to below as
-a packet number).  The encryption level indicates the packet number space, as
-described in {{QUIC-TRANSPORT}}.  Packet numbers never repeat within a packet
-number space for the lifetime of a connection.  Packet numbers monotonically
-increase within a space, preventing ambiguity.
+a packet number).  Packet numbers never repeat for the lifetime of a connection.
+Packet numbers monotonically increase within a space, preventing ambiguity.
 
 This design obviates the need for disambiguating between transmissions and
 retransmissions and eliminates significant complexity from QUIC's interpretation
@@ -126,15 +124,6 @@ Readers familiar with TCP's loss detection and congestion control will find
 algorithms here that parallel well-known TCP ones. Protocol differences between
 QUIC and TCP however contribute to algorithmic differences. We briefly describe
 these protocol differences below.
-
-### Separate Packet Number Spaces
-
-QUIC uses separate packet number spaces for each encryption level, except 0-RTT
-and all generations of 1-RTT keys use the same packet number space.  Separate
-packet number spaces ensures acknowledgement of packets sent with one level of
-encryption will not cause spurious retransmission of packets sent with a
-different encryption level.  Congestion control and RTT measurement are unified
-across packet number spaces.
 
 ### Monotonically Increasing Packet Numbers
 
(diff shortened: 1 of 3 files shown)


nibanks:
@huitema my understanding for the DOS attack, is that it is only possibly with ACKs in an Initial packet. This PR explicitly disallows ACKs in Initial packets. If I am incorrect, I'd appreciate a bit for explaination.

MikeBishop:
What happens when you receive, in a Handshake packet, an ACK for an Initial packet you never sent?  If you blow up, it becomes trivial to inject a packet and make the handshake fail.  If you don't, then the receiver will discard your packet with the same packet number because it has already received that number.

However, I think the first piece is true regardless of packet number spaces -- someone who sent additional Initial packets with your CID can cause your handshake to fail in interesting and novel ways.

nibanks:
@MikeBishop for that issue, I believe that if we don't send ACK frames in Initial packets and disallow PING / PADDING only packets, essentially requiring Initial packets to just contain CRYPTO frames (and PADDING), in both directions, if you get an ACK for a packet you didn't sent, then it would be OK to blow up. If someone is able to inject a packet with a CRYPTO frame that was accepted and acknowledged by your peer, then you are hosed anyways.

draft-ietf-quic-recovery.md
```diff
@@ -118,10 +118,8 @@ Retransmittable Packets:
 
 All transmissions in QUIC are sent with a packet-level header, which indicates
 the encryption level and includes a packet sequence number (referred to below as
-a packet number).  The encryption level indicates the packet number space, as
-described in {{QUIC-TRANSPORT}}.  Packet numbers never repeat within a packet
-number space for the lifetime of a connection.  Packet numbers monotonically
-increase within a space, preventing ambiguity.
+a packet number).  Packet numbers never repeat for the lifetime of a connection.
+Packet numbers monotonically increase within a space, preventing ambiguity.
```
MikeBishop: But you have no packet number spaces here....  I'm guessing this is a merge artifact.

MikeBishop: But you have no packet number spaces here....  I'm guessing this is a merge artifact.

martinthomson:
Closing per discussion in NYC.

Issue #1603, opened 2018-07-25
qpack: acking headers that do not reference the dynamic table is wasteful


dtikhonov:
There is no need to acknowledge header blocks that do not reference the dynamic table.  These acknowledgements do not contain any information useful to the encoder.  I propose that the draft be modified to state that

```
    A header is acknowledged only when the Largest Reference in
    its Header Data Prefix is not zero.
```

This change
1. saves bytes -- improving compression and potentially avoiding extra unneeded _STREAM_ frames and packets; and
1. allows the encoder to drop state for such header blocks as soon as they are encoded.

afrind:
Fixed by #1605

PR #1605, opened 2018-07-25
Only ack headers that have non-zero declared Largest Reference

dtikhonov:
Acknowledging headers that do not reference the dynamic table convey no useful information to the encoder.

I took the liberty to remove the part that a header block is "on a request or push stream."  This qualification is not required in the context of the changed sentence.  This information is also included in several places elsewhere in the document.

Resolves issue #1603.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -573,11 +573,11 @@ before using it.
 
 ### Header Acknowledgement
 
-After processing a header block on a request or push stream, the decoder emits a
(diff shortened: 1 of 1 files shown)


martinthomson:
I implemented this just now and it was simple enough.  And it saves a few octets in some critical cases.

dtikhonov:
I'd leave it as-is.

martinthomson:
I thought the same thing, but reached the same conclusion as @dtikhonov.

PR #1611, opened 2018-07-30
Use largest acknowledged consistently

ianswett:


martinthomson:
I find "largest acknowledged" to be easier to read.  And it avoids us using made up words.

ianswett:
The transport draft seems to consistently use acknowledged, so I changed recovery to use that.

Issue #1049, opened 2018-01-12
Specify start time for idle timeout


janaiyengar:
The spec should be more precise in exactly when the idle timeout period starts. Specifically, GQUIC experience has shown this to be slightly tricky, and there's value in articulating it in the spec.

PR #1614, opened 2018-07-31
Define idle timeout more precisely

### LINKED ISSUE #1429: Idle timeout interaction with RTO
When RTO timer is active, idle timeout only fires when idle timeout goes over RTO. This is because we specify that idle timeout is reset whenever a packet is sent or received. RTO will trigger packet transmission. We should do something better about this.

martinthomson:
This uses the definition that @ianswett provided, which is as good as any we have.  The primary measure here is receipt of new packets, though initiating new activity refreshes the timer as well (to avoid pathological timeout cases that are entirely under local control).

The point about asymmetry means that getting this time absolutely right isn't necessary.  The timeout is entirely under an endpoints own control.  We communicate this so that peers can be careful about sending near this boundary.

Closes #1429, #1049.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2377,18 +2377,18 @@ A connection that remains idle for longer than the advertised idle timeout (see
 {{transport-parameter-definitions}}) is closed.  A connection enters the
 draining state when the idle timeout expires.
 
-Endpoints use the value they advertise when determining an idle timeout.  The
-idle timeout starts from the last packet received, or the first packet sent
-after sending that packet.  The latter condition ensures that initiating new
-activity postpones a timeout.
+Each endpoint advertises their own idle timeout to their peer. The idle timeout
+starts from the last packet received, or the first packet sent after sending
+that packet.  The latter condition ensures that initiating new activity
+postpones a timeout.
 
-The value for an idle timeout can be asymmetric.  The value advertised by a peer
-is only used to determine whether the connection is live at a peer.  An endpoint
-that sends packets near the end of the idle timeout period of a peer risks
-having those packets discarded if its peer enters the draining state before the
-packets arrive.  If a peer could timeout within an RTO (see Section 4.3.3 of
-{{QUIC-RECOVERY}}), it is advisable to test for liveness before sending any data
-that cannot be retried safely.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2373,14 +2373,22 @@ source address.
...
+Endpoints use the value they advertise when determining an idle timeout.  The
+idle timeout starts from the last packet received, or the first packet sent
+after sending that packet.  The latter condition ensures that initiating new
+activity postpones a timeout.
+
+The value for an idle timeout can be asymmetric.  The value advertised by a peer
+is only used to determine whether the connection is live at a peer.  An endpoint
+that sends packets near the end of the idle timeout period of a peer risks
+having those packets discarded if its peer enters the draining state before the
+packets arrive.  If a peer could be in within an RTO (see Section 4.3.3 of
```
ianswett: How about "If a peer could idle timeout within an RTO(), it is advisable..."

draft-ietf-quic-transport.md
```diff
@@ -2377,18 +2377,18 @@ A connection that remains idle for longer than the advertised idle timeout (see
...
-is only used to determine whether the connection is live at a peer.  An endpoint
-that sends packets near the end of the idle timeout period of a peer risks
-having those packets discarded if its peer enters the draining state before the
-packets arrive.  If a peer could timeout within an RTO (see Section 4.3.3 of
-{{QUIC-RECOVERY}}), it is advisable to test for liveness before sending any data
-that cannot be retried safely.
+Each endpoint advertises their own idle timeout to their peer. The idle timeout
+starts from the last packet received, or the first packet sent after sending
+that packet.  The latter condition ensures that initiating new activity
+postpones a timeout.
```
mikkelfj: This is not very clear without more context:
What is "that" packet? Do you mean "The idle timeout is reset whenever a packet is received or sent."?
 Where is the timeout advertised, TP presumably?

mikkelfj: OK there is a reference to TP in earlier paragraph.

martinthomson: @mikkelfj, you picked up on a serious error here.  It's the first packet sent after the last received packet.  Fresh eyes FTW.

PR #1616, opened 2018-07-31
Don't amplify 0-RTT

### LINKED ISSUE #1309: Need to prevent amplification attack for any case where the server's response is big
S 4.4.3 says

```
If the server expects to generate more than three Handshake packets in response to an Initial packet, it SHOULD include a PATH_CHALLENGE frame in each Handshake packet that it sends. After receiving at least one valid PATH_RESPONSE frame, the server can send its remaining Handshake packets. Servers can instead perform address validation using a Retry packet; this requires less state on the server, but could involve additional computational effort depending on implementation choices.
```

However, there are two other cases where the server might generate >3 packets outside of the handshake:

* The application data protocol is a server speaks first protocol and it has a large banner it wants to send in 0.5 RTT
* The client is doing 0-RTT

There are ways to fix this, but it's not clear to me that the protocol actually documents them.

martinthomson:
No concrete recommendations, though it seems like kInitialWindow (which
caps at 14600) is probably a good backstop.  It might even be a good
recommendation, but I don't think that we need to concern ourselves with
finding a specific value in the same way we have for the handshake.

Closes #1309.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1896,6 +1896,13 @@ dynamically calculate the expiration time.  It is also unlikely that the client
 port number is the same on two different connections; validating the port is
 therefore unlikely to be successful.
 
+Even when a token is valid, a server SHOULD limit the amount of data that it
+sends toward a client in response to 0-RTT data.  Spoofing of source addresses
+and replays of 0-RTT handshake messages can cause tokens to appear to be valid
+when they are not.  No specific recommendation is made regarding the ratio of
+data sent to received, though it is unlikely to be necessary to send more than
+the initial congestion window.
+
 
 ### Address Validation Token Integrity {#token-integrity}
 


draft-ietf-quic-transport.md
```diff
@@ -1896,6 +1896,13 @@ dynamically calculate the expiration time.  It is also unlikely that the client
 port number is the same on two different connections; validating the port is
 therefore unlikely to be successful.
 
+Even when a token is valid, a server SHOULD limit the amount of data that it
+sends toward a client in response to 0-RTT data.  Spoofing of source addresses
+and replays of 0-RTT handshake messages can cause tokens to appear to be valid
+when they are not.  No specific recommendation is made regarding the ratio of
+data sent to received, though it is unlikely to be necessary to send more than
+the initial congestion window.
```
ekr: This text seems wrong. Tokens are either valid or they are not.

What specific attack scenario are you concerned with?

martinthomson: Tokens might be ignored (as we discussed, they might be old, or the keys might not be available).

This is your issue #1309.

ianswett: I think it's fairly likely it may be necessary to send more than the IW, but I think we should advise limiting the amount of data sent to the IW.  I think this just falls out of the recovery doc, since the recovery doc specifies an initial window of 10 packets, so I'm not completely sure this is necessary?

martinthomson: Yeah, this is why it is worded so loosely.  Without this advice, the risk seems to be limited to the server receiving a spoofed ACK for the ServerHello and sending that much more in response, but that's only one more packet than the IW, so it's hardly earth-shattering.

Given that this is largely dictated by congestion controller design (IW specifically), is this really a change we need to implement?  Should we instead close #1309 with no change?

ianswett: I'm fine with just closing #1309

martinthomson: @ekr?

PR #1621, opened 2018-08-01
Explicitly permit starting over

martinthomson:
Generating a new cryptographic handshake message is a fine response to
Retry or Version Negotiation.  It's sometimes necessary, but even when
it is not, it avoids having to worry about 0-RTT packet numbers.

Followup for #1498.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -806,7 +806,8 @@ A client MUST NOT reset the packet number it uses for 0-RTT packets.  The keys
 used to protect 0-RTT packets will not change as a result of responding to a
 Retry or Version Negotiation packet unless the client also regenerates the
 cryptographic handshake message.  Sending packets with the same packet number in
-that case would compromise the packet protection for all 0-RTT packets.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -806,7 +806,8 @@ A client MUST NOT reset the packet number it uses for 0-RTT packets.  The keys
 used to protect 0-RTT packets will not change as a result of responding to a
 Retry or Version Negotiation packet unless the client also regenerates the
 cryptographic handshake message.  Sending packets with the same packet number in
-that case would compromise the packet protection for all 0-RTT packets.
+that case could compromise the packet protection for all 0-RTT packets because
```
ekr: Why the change from "would" to "could". Could seems a lot more likely.

martinthomson: It's not guaranteed - if the exact same data is sent again, then there is no compromise.  And, from what I know of some implementations, that might even work.  But it's a long shot.  I'll tweak it to change emphasis.

PR #1624, opened 2018-08-01
Duplicates are dropped

### LINKED ISSUE #1405: Handling of duplicate packets
Packet duplication sometimes occur and it has a potential for affecting the ECN counter for what markings that arrive within a connection. Thus, it would be good if the requirements on how a QUIC implementation handle packets that are detected as duplicates. 

There also appear that there are current text in the transport draft that makes assumption about handling of duplicate packets:

6.10.4.1.  Detecting a Stateless Reset

   A client detects a potential stateless reset when a packet with a
   short header either cannot be decrypted or is marked as a duplicate
   packet.

The other mention of packet duplication, also pointing to an action that is this:

6.8.3.2.  Handling Address Spoofing by an On-path Attacker

   An on-path attacker could cause a spurious connection migration by
   copying and forwarding a packet with a spoofed address such that it
   arrives before the original packet.  The packet with the spoofed
   address will be seen to come from a migrating connection, and the
   original packet will be seen as a duplicate and dropped.

These two pieces appear to indicate that packet duplicates should:
   1. Be detected by receiver, at least within the window of PN between highest seen and some window into the past. 
  2. The packet marked as duplicate
  3. Duplicate are dropped, but only after having been somewhat processed (see 6.10.4.1).

To my understanding the smallest window a receiver could have for tracking packets, is what is in ACKs that hasn't been ACKed yet. But that could be extremely short into the past, which could be short in time, compared to the expected re-ordering or packet life time in the network. Independently of window size, if a packet would arrive with a PN that is prior to the current window the receiver have, the receiver has discard the state for the packet number, and thus can't determine if it was ACKed or not, and thus needs to be discarded without further processing. Is that correctly assumed. 

Thus, it would be good if these assumptions and needed functionality was collected in its own paragraphs somewhere in the specification. 

From an ECN perspective, the current thinking is that the ECN marking of a duplicate will simply be ignored. We want to avoid adding duplicates in the counter, as it will increase the uncertainty of the validation of the correct marking.

martinthomson:
It's a little surprising that we didn't say this already.

I thought about a mention of the security considerations in the TLS
draft with this change, but I think that's unnecessary at this point.

Closes #1405.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1066,12 +1066,12 @@ sending a packet with a number of 0x6b2d79 requires a packet number encoding
 with 14 bits or more; whereas the 30-bit packet number encoding is needed to
 send a packet with a number of 0x6bc107.
 
-An endpoint MUST discard packets with duplicate packet numbers.  After
-unprotecting a packet, the receiver MUST check that it has not already processed
-a packet with this PN and if so discard it. Duplicate suppression MUST happen
-after the packet has been unprotected for the reasons described in Section 9.3
-of {{QUIC-TLS}}. An efficient algorithm for duplicate suppression can be found
-in Section 3.4.3 of {{?RFC2406}}.
+A receiver MUST discard a newly unprotected packet unless it is certain that it
+has not processed another packet with the same packet number from the same
+packet number space. Duplicate suppression MUST happen after removing packet
+protection for the reasons described in Section 9.3 of {{QUIC-TLS}}. An
+efficient algorithm for duplicate suppression can be found in Section 3.4.3 of
+{{?RFC2406}}.
 
 A Version Negotiation packet ({{packet-version}}) does not include a packet
 number.  The Retry packet ({{packet-retry}}) has special rules for populating
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1066,11 +1066,12 @@ sending a packet with a number of 0x6b2d79 requires a packet number encoding
 with 14 bits or more; whereas the 30-bit packet number encoding is needed to
 send a packet with a number of 0x6bc107.
 
-An endpoint MUST discard packets with duplicate packet numbers.  That is, a
-packet is discarded if the reconstructed packet number is equal to the
-reconstructed packet number of a previously received and processed packet.  Note
-however that endpoints MUST attempt to remove packet protection before
-discarding packets (see Section 9.3 of {{QUIC-TLS}}).
+An endpoint MUST discard packets with duplicate packet numbers.  After
+unprotecting a packet, the receiver MUST check that it has not already processed
+a packet with this PN and if so discard it. Duplicate suppression MUST happen
+after the packet has been unprotected for the reasons described in Section 9.3
+of {{QUIC-TLS}}. An efficient algorithm for duplicate suppression can be found
+in Section 3.4.3 of {{?RFC2406}}.
 
 A Version Negotiation packet ({{packet-version}}) does not include a packet
 number.  The Retry packet ({{packet-retry}}) has special rules for populating
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -979,6 +979,7 @@ The connection ID can change over the lifetime of a connection, especially in
 response to connection migration ({{migration}}). NEW_CONNECTION_ID frames
 ({{frame-new-connection-id}}) are used to provide new connection ID values.
 
+
 ## Packet Numbers {#packet-numbers}
 
 The packet number is an integer in the range 0 to 2^62-1. The value is used in
@@ -1064,6 +1065,10 @@ For example, if an endpoint has received an acknowledgment for packet 0x6afa2f,
 sending a packet with a number of 0x6b2d79 requires a packet number encoding
 with 14 bits or more; whereas the 30-bit packet number encoding is needed to
 send a packet with a number of 0x6bc107.
+
+An endpoint MUST discard packets with duplicate packet numbers.  That is, a
+packet is discarded if the reconstructed packet number is equal to the
+reconstructed packet number of a previously received and processed packet.
 
 A Version Negotiation packet ({{packet-version}}) does not include a packet
 number.  The Retry packet ({{packet-retry}}) has special rules for populating
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1068,7 +1068,9 @@ send a packet with a number of 0x6bc107.
 
 An endpoint MUST discard packets with duplicate packet numbers.  That is, a
 packet is discarded if the reconstructed packet number is equal to the
-reconstructed packet number of a previously received and processed packet.
+reconstructed packet number of a previously received and processed packet.  Note
+however that endpoints MUST attempt to remove packet protection before
+discarding packets (see Section 9.3 of {{QUIC-TLS}}).
 
 A Version Negotiation packet ({{packet-version}}) does not include a packet
 number.  The Retry packet ({{packet-retry}}) has special rules for populating


marten-seemann:
We didn't say this because it's not necessary. There's no harm (other than a performance hit) in processing any frame more than once.

kazuho:
I agree with @marten-seemann here.

If we are to add some text, I think that we should have a reference to the TLS document that states how it should be done, i.e.
> If the recipient of a packet discards packets with duplicate packet numbers without attempting to remove packet protection they could reveal through timing side-channels that the packet number matches a received packet. For authentication to be free from side-channels, the entire process of packet number protection removal, packet number recovery, and packet protection removal MUST be applied together without timing and other side-channels.
> https://quicwg.org/base-drafts/draft-ietf-quic-tls.html#rfc.section.9.3

martinthomson:
Some of us assumed that this was how it had to work.  Mainly because it is hard to think of a good reason to allow acceptance.

Consider: what if the same packet number was used for different content?  Other than it completely breaking most AEADs, what was it that was acknowledged?

I can add a reference to Section 9.3 of the TLS doc.  I didn't think that it was necessary, but it's easy to do.

kazuho:
@martinthomson 
> Some of us assumed that this was how it had to work. Mainly because it is hard to think of a good reason to allow acceptance.

I prefer using MAY rather than MUST.

The reason is because there is no need to enforce every stack to be written that way. For example, a QUIC stack can do the following:
* discard a newly received packet if it contains a PN in the ack queue
* retire PNs from ack queue after some time

This would work completely fine, even though it has the chance of processing a packet with same PN more than once.

If it is a MUST, every stack will be required to have a bit vector and a minimum threshold to detect all duplicates and also other awfully reordered packets. I understand that it is possible to implement such a mechanism. However, I simply think that it is not necessary.

> Consider: what if the same packet number was used for different content? Other than it completely breaking most AEADs, what was it that was acknowledged?

My understanding is that AEAED is broken if the sender reuses a packet number for carrying different content. I think we do not need to consider how to handle a corrupt implementation in this respect.

> I can add a reference to Section 9.3 of the TLS doc. I didn't think that it was necessary, but it's easy to do.

Thank you for the addition!

kazuho:
> The protocol doesn't make any sense if you get (valid) duplicates and don't discard them.

I think we have different opinions here. My understanding is that processing of a “valid” duplicate (i.e, a packet that has the same PN and same payload as what have been observed earlier) is perfectly fine, because processing the payload of such packet does not cause any change in state. Processing of previously-observed frames are guaranteed to be idempotent.

mikkelfj:
I tend to agree with @kazuho but on the other hand, taking that position forces all future frames and extensions to be idempotent. I'm not sure that is a good tradeoff.

kazuho:
@mikkelfj I’d argue that any frame that cause a change in state needs to be retransmittable. We allow that by using different PNs.

In my view, what we are arguing about is if we need to have a special rule for the case of receiving a “retransmitted” frame using same PN.

mikkelfj:
> In my view, what we are arguing about is if we need to have a special rule for the case of receiving a “retransmitted” frame using same PN.

Which, as I see it, is the same as arguing if all frames must be idempotent. I do not disagree here.

ekr:
I'm with @mt here. It's not that hard to implement the now-conventional
anti-replay techniques, and if we don't do this, then we have to analyze
every single QUIC frame to guarantee it can't cause problems if replayed
(and moreover implementors need to enforce it). To take a simple example
from earlier, consider the flow control frames: if we require duplicate
suppression, then even if you miss the rule about things being
monotonically increasing, things are fine. However, if duplicates are
allowed, and you make that mistake, then attackers can just replay a flow
control frame and force you into a bad posture.

draft-ietf-quic-transport.md
```diff
@@ -1065,6 +1066,12 @@ sending a packet with a number of 0x6b2d79 requires a packet number encoding
 with 14 bits or more; whereas the 30-bit packet number encoding is needed to
 send a packet with a number of 0x6bc107.
 
+An endpoint MUST discard packets with duplicate packet numbers.  That is, a
+packet is discarded if the reconstructed packet number is equal to the
+reconstructed packet number of a previously received and processed packet.  Note
+however that endpoints MUST attempt to remove packet protection before
+discarding packets (see Section 9.3 of {{QUIC-TLS}}).
```
ekr: I think i would put this in the TLS document S 5.4 instead. That way it's in logical order and this requirement seems like less of an aside.

Perhaps:

```
After unprotecting a packet, the receiver MUST check that it has not already processed
a packet with this PN and if so discard it. Duplicate suppression MUST happen after
the packet has been unprotected for the reasons described in Section 9.3. An efficient
algorithm for duplicate suppression can be found in RFC 2406 Section 3.4.3.```

**Comment by kazuho** [comment]
@ekr 
> I'm with @mt here. It's not that hard to implement the now-conventional anti-replay techniques, and if we don't do this, then we have to analyze every single QUIC frame to guarantee it can't cause problems if replayed (and moreover implementors need to enforce it).

My understanding is that we need to analyze and guarantee that, regardless of if we require deduplication. This is because a peer can always replay a frame.

Taking the flow control frames as an example, the fact that endpoints need to handle out-of-order delivery means that they need to take care of receiving smaller maximum values that it has seen before. The requirement does not go away by requiring deduplication.

**Comment by martinthomson** [comment]
It's very different in my opinion to have rules about frame duplication (which we clearly need for any frame that might be retransmitted), and packet duplication.  A properly duplicate packet - one with identical content - is easy, but I don't want to have to think about what happens when a packet isn't genuinely identical.  And deduplication is easy to require because it's easy to implement.

This isn't a rulebook we can hit people with.  So if you process the same packet twice and no one hears it, then maybe it doesn't matter, but the "MUST"-language we put here is part of what determines how the protocol is analyzed.  If we don't say "no duplicates", we made that analysis harder and the results weaker.

**Comment by kazuho** [comment]
@martinthomson
> This isn't a rulebook we can hit people with. 

Ah. Thank you for the suggestion. That’s true. Then, I will not object against requiring deduplication if it helps analyse the protocol.

**Comment by janaiyengar** [comment]
I am sympathetic to Kazuho's argument that this language seems to require state that a receiver may not have. I think this is easily resolvable though. If a packet received is either a duplicate or if it is to the left of the receiver's "window" (the lowest packet number that the receiver is willing to receive and process), the receiver drops it. I think this handles the cases where the receiver may not be able to dedup, but the outcome is the same -- the receiver drops the packet.

File: draft-ietf-quic-transport.md [hunk 2]
```diff
@@ -1065,6 +1066,13 @@ sending a packet with a number of 0x6b2d79 requires a packet number encoding
 with 14 bits or more; whereas the 30-bit packet number encoding is needed to
 send a packet with a number of 0x6bc107.
 
+An endpoint MUST discard packets with duplicate packet numbers.  After
```
ianswett: How about combining the first two sentences, since they seem largely redundant? Maybe "An endpoint MUST discard packets with duplicate packet numbers after unprotecting the packet." or something similar.

MikeBishop:
I'm mostly in agreement with Kazuho from a theoretical perspective -- there is no harm in processing a true duplicate, and that's part of what the idempotency language was about.  My inclination is SHOULD discard, but MUST NOT send more than one packet with the same packet number.  If an attacker can spoof a packet into being valid, you've got bigger problems.

However, "MUST discard packets which can't be determined not to be a duplicate" is also workable.  There's no correctness loss either way.

draft-ietf-quic-transport.md
```diff
@@ -1065,6 +1066,13 @@ sending a packet with a number of 0x6b2d79 requires a packet number encoding
...
+unprotecting a packet, the receiver MUST check that it has not already processed
+a packet with this PN and if so discard it. Duplicate suppression MUST happen
```
MikeBishop: The receiver isn't necessarily able to confirm that for all packet numbers.  I think the language needs to be that if the receiver MUST discard if it can't verify that it hasn't processed this packet number before.

PR #1626, opened 2018-08-01
Consequences of bad ECN markings

### LINKED ISSUE #1426: Detect adversarial ECN reporting
In the ECN proposal, the sender relies on the recipient to send back an honest count of ECN marks. A client who wants the sender to overshoot the network capacity will be able to just avoid reporting the "congestion experienced" marks, pretending that there never is congestion.

This is similar to the "optimistic ack" problem that is mitigated by leaving holes in the sequence of packet numbers. If a bad client acks a hole, the sender notices it and can terminate the connection.

We should consider a similar mechanism for ECN. The sender normally marks packets as ECT(0) or ECT(1). It could also at some random intervals mark a packet as "CE", and then verify that the CE is reflected in the counters sent by the recipient.

The downside is that the sender-set CE may "mask" a congestion in the network. But if only a small fraction of the packets are used for this probing mechanism, it should not be too much of a problem.

martinthomson:
I put this in the recovery draft because it relates directly to the
congestion control response of peers.  Happy to move it if an argument
supporting it being in the transport doc can be made.

Closes #1426.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1209,6 +1209,27 @@ characteristics or application behavior.  Endpoints can use PADDING frames or
 bundle acknowledgments with other frames to reduce leaked information.
 
 
+## Misreporting ECN Markings
+
+An endpoint can misreport ECN markings to alter the congestion response of a
+peer.  Suppressing reports of ECN-CE markings could cause a peer to increase
+their send rate.  This increase could result in congestion and loss.
+
+An endpoint MAY attempt to detect suppression of reports by marking occasional
+packets that they send with ECN-CE.  If a packet marked with ECN-CE is not
+reported as having been marked when the packet is acknowledged, the endpoint
+SHOULD then disable ECN for that path.
+
+Reporting additional ECN-CE markings will cause a peer to reduce their sending
+rate, which is similar in effect to advertising reduced connection flow control
+limits and so no advantage is gained by doing so.
+
+Endpoints choose the congestion controller that they use.  Though congestion
+controllers ideally use reports of ECN markings as input, the exact response for
+each controller could be different.  Failure to correctly respond to information
+about ECN markings is therefore difficult to detect.
+
+
 # IANA Considerations
 
 This document has no IANA actions.  Yet.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -1209,6 +1209,27 @@ characteristics or application behavior.  Endpoints can use PADDING frames or
...
+SHOULD then disable ECN for that path.
+
+Reporting additional ECN-CE markings will cause a peer to reduce their sending
+rate, which is similar in effect to advertising reduced connection flow control
+limits and so no advantage is gained by doing so.
+
+Endpoints choose the congestion controller that they use.  Though congestion
+controllers ideally use reports of ECN markings as input, the exact response for
+each controller could be different.  Failure to correctly respond to information
+about ECN markings is therefore difficult to detect.
```
mirjak: Actually this is a bit weak. RFC3168 say
"Upon the receipt by an ECN-Capable transport of a single CE
packet, the congestion control algorithms followed at the end-
systems MUST be essentially the same as the congestion control
response to a single dropped packet."
(where RFC8311 adds a
"unless otherwise specified by an Experimental RFC in the IETF document stream"
to this last sentence, however, this is not really important here).

The text as currently written indicates that the congestion controller can choose to account for ECN information at all or not which is not the case.

mirjak: DCTCP is not applicable to the Internet. The point being is that if you mark your packets as ECN-enabled, you also must react to CE marks accordingly, you can't just ignore them. I guess we can remove the whole paragraph or just this one sentence in the middle.

huitema: RFC3168 does say that the response to an ECN mark must be the same as to a dropped packet. However, different congestion control algorithms may well respond to dropped packets differently. Receivers may detect that the sender's algorithm does not respond as TCP-Reno would, but then so what? It just means that the sender's congestion control algorithm is different from TCP-Reno.

martinthomson: Would this be better if it said "Though congestion controllers generally treat reports of ECN-CE markings as equivalent to loss [RFC8311], the exact response for each controller could be different." ?

mirjak: WFM.

martinthomson:
@ianswett I think that this is ready.  It's all yours now.

PR #1628, opened 2018-08-01
add code points for 1, 2 and 3 byte connection IDs

### LINKED ISSUE #1570: Connection ID lengths 1, 2 and 3 bytes can't be encoded
The connection IDs can be either 0 bytes or anything between 4 and 18 bytes (inclusive). Lengths of 1, 2 and 3 bytes are not possible.

There are several use cases that would benefit from using shorter connection IDs, especially in p2p applications. An endpoint could encode the length of the connection ID into the first few bits, and then starts using 1 byte connection IDs first, and successively moves to longer lengths only when handling more connections. For p2p hosts, in many cases 2 bytes would be sufficient for all use cases.

In the connection ID length field in the Long Header, we only have 4 bits to encode a range of 19 values (we want 0 byte connection IDs, and we want to support 18 byte connection IDs), so *some* values have to be left out. It seems there are few use cases for 13, 14 and 15 bytes.

We could achieve this by changing the definition of the length to:

>  An encoded length between 0 and 12 inclusive indicates that the connection ID is also 0 to 12 octets in length. Any larger value is increased by 3 to get the full length of the connection ID, producing a length between 16 and 18 octets inclusive. 

If people are supportive of this change, I'd be happy to write a PR.

marten-seemann:
This PR changes the invariants and the transport document to allow 1, 2 and 3 byte connection IDs as well.

Fixes #1570.

### DIFF
--- a/draft-ietf-quic-invariants.md
+++ b/draft-ietf-quic-invariants.md
@@ -163,10 +163,10 @@ The next octet contains the length in octets of the two Connection IDs (see
 {{connection-id}}) that follow.  Each length is encoded as a 4-bit unsigned
 integer.  The length of the Destination Connection ID (DCIL) occupies the high
 bits of the octet and the length of the Source Connection ID (SCIL) occupying
-the low bits of the octet.  An encoded length of 0 indicates that the connection
-ID is also 0 octets in length.  Non-zero encoded lengths are increased by 3 to
-get the full length of the connection ID; the final value is therefore either 0
-or between 4 and 18 octets in length (inclusive).  For example, an octet with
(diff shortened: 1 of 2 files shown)


mikkelfj:
Perhaps some advisory text on a CID lengths below 32 cannot be expected to be have the same uniqueness and unlinkability properties as longer CID's and should be considered for special purposes only, such as internal server P2P.

ekr:
As I noted in email, I think this really would need a new WG consensus call

ianswett:
Agreed, this needs a new WG consensus call.  I was giving my approval of the text and direction.

janaiyengar:
Closing, per discussion at interim.

Issue #158, opened 2017-01-16
Padding between frames


larseggert:
The section on padding frames currently says:
```
   The PADDING frame (type=0x00) pads a packet with 0x00 bytes.  When
   this frame is encountered, the rest of the packet is expected to be
   padding bytes.  The frame contains 0x00 bytes and extends to the end
   of the QUIC packet.  A PADDING frame has no additional fields.
```

I'd like to discuss whether a padding frame that can be inserted anywhere, instead of only as the last frame in a packet, makes sense.

I'm currently doing a [zero-copy implementation of QUIC](https://github.com/NTAP/quant) on top of [netmap](http://info.iet.unipi.it/~luigi/netmap/) via a [userspace IPv4/UDP stack](https://github.com/NTAP/warpcore), which due to static DMA mappings requires an app to place stream frame data directly into packet buffers at a (fixed) offset. The QUIC stack then inserts the common header and other frames before the stream frames, and needs to pad any unused space. Currently, this isn't possible.

In other words, I'd like something similar to the [No-Operation TCP option](https://tools.ietf.org/html/rfc793#page-18).

(And it's an open question whether going to great lengths to support zero-copy TX is actually all that useful, given that we're going to do crypto over the entire buffer anyway.)

martinthomson:
There is actually a very simple fix for this, though it increases the work required to handle a packet: you could say that a PADDING frame is empty.

I'm not convinced of the necessity though, is this because lengths occasionally need to grow?  Can you identify the cases where this happens?

FWIW, crypto shouldn't affect zero-copy because most AEAD implementations encrypt and decrypt in place.

larseggert:
Yes, and empty PADDING frame would work, but you need to parse each byte of the padding then, which is not so great. I was thinking more about about a frame type prefix like STREAM and ACK have, and using the remaining bits to encode a length.

I'm not sure I fully understand your comment about necessity. For an implementation where data is pre-placed into a packet buffer and the QUIC stack needs to prepend its headers. Because the buffers are statically DMA-mapped, the beginning of the buffer cannot be changed to account for the variable-length QUIC header from packet to packet. Therefore, the stack needs to leave some slack space for QUIC to use for its header, and pad the unused slack space until the beginning of the first STREAM frame (for which the app has placed the data already.)

Let me try and draw something:
    
                            also at fixed offset, but var. length
      fixed length headers  |        var. length
      v          v    v     v        v
    +----------+----+-----+--------+--------+--------+--------+
[...]

PR #1631, opened 2018-08-01
Clarify use of the term "supports ECN"

mirjak:
See issue #158: "support" should only be used for the path. I have left two occasions where it is used for both the path and the peer but I think that is okay because it only addresses the case where there is no support.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1834,26 +1834,31 @@ congestion in the network by setting a codepoint in the IP header of a packet
 instead of dropping it.  Endpoints react to congestion by reducing their sending
 rate in response, as described in {{QUIC-RECOVERY}}.
 
-To use ECN, QUIC endpoints first determine whether a path and peer support ECN
-marking. Verifying the path occurs at the beginning of a connection and when the
+To use ECN, QUIC endpoints first determine whether a path supports ECN
+marking and the peer is able to access the ECN codepoint in the IP header. 
+A network path regarded as not supporting ECN, if ECN marked packets get dropped
+or ECN marking are rewritten on the path.
+Verifying the path occurs at the beginning of a connection and when the
 connection migrates to a new path (see {{migration}}).
 
-Each endpoint independently verifies and enables ECN for the path from it to the
-peer.
+Each endpoint independently verifies and enables use of ECN by setting the IP
+header ECN codepoint to ECN Capable Transport (ECT) for the path from it to the
+peer. Even if ECN is not used on the path to the peer, the endpoint MUST provide
+ECN feedback information if the receive of a non-zero ECN IP codepoint is observed.
 
-To verify that both a path and the peer support ECN, an endpoint MUST set one of
-the ECN Capable Transport (ECT) codepoints -- ECT(0) or ECT(1) -- in the IP
-header {{!RFC8311}} of all outgoing packets.
+To verify both that a path supports ECN and the peer can provide ECN feedback, 
+an endpoint MUST set the ECT(0) codepoint in the IP header of all outgoing packets
+{{!RFC8311}}. 
 
 If an ECT codepoint set in the IP header is not corrupted by a network device,
 then a received packet contains either the codepoint sent by the peer or the
 Congestion Experienced (CE) codepoint set by a network device that is
 experiencing congestion.
 
-On receiving a packet with an ECT or CE codepoint, an endpoint that supports ECN
-increases the corresponding ECT(0), ECT(1), or CE count, and includes these
-counters in subsequent (see {{processing-and-ack}}) ACK_ECN frames (see
-{{frame-ack-ecn}}).
+On receiving a packet with an ECT or CE codepoint, an endpoint that can access the
+IP ECN codepoints increases the corresponding ECT(0), ECT(1), or CE count, and 
+includes these counters in subsequent (see {{processing-and-ack}}) ACK_ECN frames 
+(see {{frame-ack-ecn}}).
 
 A packet detected by a receiver as a duplicate does not affect the receiver's
 local ECN codepoint counts; see ({{security-ecn}}) for relevant security
@@ -1862,7 +1867,7 @@ concerns.
 If an endpoint receives a packet without an ECT or CE codepoint, it responds per
 {{processing-and-ack}} with an ACK frame.
 
-If an endpoint does not support ECN or does not have access to received ECN
+If an endpoint does not have access to received ECN
 codepoints, it acknowledges received packets per {{processing-and-ack}} with an
 ACK frame.
 


draft-ietf-quic-transport.md
```diff
@@ -1834,26 +1834,31 @@ congestion in the network by setting a codepoint in the IP header of a packet
 instead of dropping it.  Endpoints react to congestion by reducing their sending
 rate in response, as described in {{QUIC-RECOVERY}}.
 
-To use ECN, QUIC endpoints first determine whether a path and peer support ECN
-marking. Verifying the path occurs at the beginning of a connection and when the
+To use ECN, QUIC endpoints first determine whether a path supports ECN
+marking and the peer is able to access the ECN codepoint in the IP header. 
+A network path regarded as not supporting ECN, if ECN marked packets get dropped
```
martinthomson: remove the comma

draft-ietf-quic-transport.md
```diff
@@ -1834,26 +1834,31 @@ congestion in the network by setting a codepoint in the IP header of a packet
...
+or ECN marking are rewritten on the path.
+Verifying the path occurs at the beginning of a connection and when the
 connection migrates to a new path (see {{migration}}).
 
-Each endpoint independently verifies and enables ECN for the path from it to the
-peer.
+Each endpoint independently verifies and enables use of ECN by setting the IP
+header ECN codepoint to ECN Capable Transport (ECT) for the path from it to the
```
martinthomson: Which one?

draft-ietf-quic-transport.md
```diff
@@ -1834,26 +1834,31 @@ congestion in the network by setting a codepoint in the IP header of a packet
...
+peer. Even if ECN is not used on the path to the peer, the endpoint MUST provide
+ECN feedback information if the receive of a non-zero ECN IP codepoint is observed.
 
-To verify that both a path and the peer support ECN, an endpoint MUST set one of
-the ECN Capable Transport (ECT) codepoints -- ECT(0) or ECT(1) -- in the IP
-header {{!RFC8311}} of all outgoing packets.
+To verify both that a path supports ECN and the peer can provide ECN feedback, 
+an endpoint MUST set the ECT(0) codepoint in the IP header of all outgoing packets
+{{!RFC8311}}.
```
martinthomson: Trailing space.

draft-ietf-quic-transport.md
```diff
@@ -1834,26 +1834,31 @@ congestion in the network by setting a codepoint in the IP header of a packet
 instead of dropping it.  Endpoints react to congestion by reducing their sending
 rate in response, as described in {{QUIC-RECOVERY}}.
 
-To use ECN, QUIC endpoints first determine whether a path and peer support ECN
-marking. Verifying the path occurs at the beginning of a connection and when the
+To use ECN, QUIC endpoints first determine whether a path supports ECN
+marking and the peer is able to access the ECN codepoint in the IP header. 
+A network path regarded as not supporting ECN, if ECN marked packets get dropped
+or ECN marking are rewritten on the path.
+Verifying the path occurs at the beginning of a connection and when the
```
martinthomson: Passive voice.  "An endpoint verifies every path, both during connection establishment and when migrating to a new path (see {{migration}})."

draft-ietf-quic-transport.md
```diff
@@ -1834,26 +1834,32 @@ congestion in the network by setting a codepoint in the IP header of a packet
...
+A network path regarded as not supporting ECN if ECN marked packets get dropped
+or ECN marking are rewritten on the path.
+An endpoint verifies the path, both during connection establishment 
+and when migrating to a new path (see {{migration}}).
+
+Each endpoint independently verifies and enables use of ECN by setting the IP
+header ECN codepoint to ECN Capable Transport (ECT) for the path from it to the
+other peer. Even if ECN is not used on the path to the peer, the endpoint MUST
+provide ECN feedback information if the receive of a non-zero ECN IP codepoint
+is observed.
```
martinthomson: Grammar here could be improved: "an endpoint MUST provide ECN feedback information if it receives packets with non-zero ECN markings." ?

mirjak: Maybe just "the endpoint MUST provide feedback about ECN markings received (if accessible)." Is that better?

martinthomson: Yeah, concise and correct.  You can just drop the parenthetical: if the markings aren't accessible, the endpoint didn't get them.

janaiyengar:
I don't think this distinction is useful, and I don't understand it. An endpoint can choose to not support ECN, for whatever reason, one of them being that the ECN codepoints are not accessible. If it helps to clarify this, I'm happy to write something that clarifies what "endpoint support of ECN" means... I don't think we need more than a sentence.

draft-ietf-quic-transport.md
```diff
@@ -1834,26 +1834,31 @@ congestion in the network by setting a codepoint in the IP header of a packet
...
-Each endpoint independently verifies and enables ECN for the path from it to the
-peer.
-
-To verify that both a path and the peer support ECN, an endpoint MUST set one of
-the ECN Capable Transport (ECT) codepoints -- ECT(0) or ECT(1) -- in the IP
-header {{!RFC8311}} of all outgoing packets.
+To use ECN, QUIC endpoints first determine whether a path supports ECN
+marking and the peer is able to access the ECN codepoint in the IP header.
+A network path regarded as not supporting ECN if ECN marked packets get dropped
+or ECN marking are rewritten on the path. An endpoint verifies the path, both
```
MikeBishop: "marking**s** are"; "An endpoint verifies **each** path"

draft-ietf-quic-transport.md
```diff
@@ -1834,26 +1834,31 @@ congestion in the network by setting a codepoint in the IP header of a packet
...
+during connection establishment and when migrating to a new path (see
+{{migration}}).
+
+Each endpoint independently verifies and enables use of ECN by setting the IP
+header ECN codepoint to ECN Capable Transport (ECT) for the path from it to the
+other peer. Even if ECN is not used on the path to the peer,the endpoint MUST
```
MikeBishop: space

draft-ietf-quic-transport.md
```diff
@@ -1862,7 +1867,7 @@ concerns.
 If an endpoint receives a packet without an ECT or CE codepoint, it responds per
 {{processing-and-ack}} with an ACK frame.
 
-If an endpoint does not support ECN or does not have access to received ECN
+If an endpoint does not have access to received ECN
 codepoints, it acknowledges received packets per {{processing-and-ack}} with an
 ACK frame.
```
MikeBishop: Rewrap.

draft-ietf-quic-transport.md
mirjak: Actually the endpoint only verifies the forward path (the path from that endpoint to the other endpoint). Use of ECN of the "backward" path is verified by the other endpoint (given it's its forward path). Do we need to clarify this?

draft-ietf-quic-transport.md
```diff
@@ -1862,9 +1867,8 @@ concerns.
 If an endpoint receives a packet without an ECT or CE codepoint, it responds per
 {{processing-and-ack}} with an ACK frame.
 
-If an endpoint does not support ECN or does not have access to received ECN
-codepoints, it acknowledges received packets per {{processing-and-ack}} with an
-ACK frame.
+If an endpoint does not have access to received ECN codepoints, it acknowledges
```
MikeBishop: Rewrap, but without making the line long enough to produce a build break.

PR #1632, opened 2018-08-01
Fix: a connection is not identified by a *collection*

dtikhonov:
This small change contains several other improvements, each with a clear
purpose behind it.

    1. Replace the word "collection" with the word "set," as the former
       looks a lot like "connection."

    2. Remove the "unlikely to be unique" parenthetical clause: it is
       obvious that there can only be one unique zero-length connection ID.

    3. Drop the "assigned to it" part of the sentence, as it is unnecessary
       in this introductory paragraph.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1183,10 +1183,10 @@ endpoint, as described in {{termination}}.
 
 ## Connection ID
 
-Each connection is identified by a collection of identifiers assigned to it. A
-connection ID can be 0 octets in length (and thus unlikely to be unique), or
-between 4 and 18 octets (inclusive).  Connection IDs are selected independently
-in each direction.
+Each connection possesses a set of identifiers, any of which could be used to
+distinguish it from other connections.  A connection ID can be either 0 octets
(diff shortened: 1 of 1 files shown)


mikkelfj:
Connections ID's are not necessarily unique since they can be distinquished also by tuple. Before the asymmetric design unqueness could not be guaranteed as the ID could be peer selected. Now it can be made unique but it still doesn't have to be - and for closed connections a risk remains, especially if you have less than 32-bits as is being proposed here https://github.com/quicwg/base-drafts/pull/1628

MikeBishop:
I think we need to assume that CIDs are unique without regard to the client address, because otherwise you won't survive NAT rebinding.  The same will be true for server addresses if we allow server migration in a future version, so I'd like to lay the conceptual groundwork here.

I agree that a server could hypothetically reuse CIDs across its own addresses, but to me that seems more conceptually akin to multiple independent servers running on different ports.  I'm inclined to leave the uniqueness comments in, as @dtikhonov has done here.

I originally said "unlikely to be" unique, since IIRC we have at least one implementation that used zero-length addresses until there was a tuple overlap.  Zero-length can be unique if it's the only one.  :-)

PR #1633, opened 2018-08-02
Other congestion controllers

### LINKED ISSUE #1625: Choice of congestion controller is discretionary
I couldn't find explicit text in the recovery draft that says "this is one congestion controller, but implementations MAY use a different congestion controller at their discretion".  It's a pretty important point and probably belongs in the introduction.

ianswett:
Fixes #1625

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -946,7 +946,9 @@ QUIC hosts MUST NOT send packets if they would increase bytes_in_flight
 unless the packet is a probe packet sent after the TLP or RTO timer expires,
 as described in {{tlp}} and {{rto}}.
 
-Implementations MAY use other congestion control algorithms.
+Implementations MAY use other congestion control algorithms. The signals QUIC
+provides for congestion control are generic and are designed to support
+different algorithms.
 
 ## Explicit Congestion Notification {#congestion-ecn}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -936,15 +936,17 @@ both the median and mean min_rtt typically observed on the public internet.
...
 
 QUIC hosts MUST NOT send packets if they would increase bytes_in_flight
-(defined in {{vars-of-interest}}) beyond the available congestion window, unless
-the packet is a probe packet sent after the TLP or RTO timer expires, as
-described in {{tlp}} and {{rto}}.
+(defined in {{vars-of-interest}}) beyond the available congestion window,
+unless the packet is a probe packet sent after the TLP or RTO timer expires,
+as described in {{tlp}} and {{rto}}.
+
+Implementations MAY use other congestion control algorithms.
```
martinthomson: Do we also need to say that endpoints don't need to use the same congestion controller?  IOW, "Implementations MAY use a different congestion control algorithm, even within the same connection.  The signals that QUIC provides for congestion control are generic and can support different congestion control regimens."

ianswett:
Thanks Martin, I added your comment about the signals being generic.  Changing congestion controllers mid-connection is certainly possible, but I think it ends up being more confusing than clarifying.

martinthomson:
I never intended to imply a change mid-connection, just that the choice was unilateral.

PR #1636, opened 2018-08-07
Rules for discarding old keys

### LINKED ISSUE #1544: Document when you can delete keys
When you change epochs. Per Slack conversation.

martinthomson:
This is - at its core - simple.  When you are done with keys, set a
timer.  Working through all the ramifications and defining this
precisely takes more words than I'd originally anticipated.

Closes #1544.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -44,6 +44,23 @@ normative:
         org: Mozilla
         role: editor
 
+  QUIC-RECOVERY:
+    title: "QUIC Loss Detection and Congestion Control"
+    date: {DATE}
+    seriesinfo:
+      Internet-Draft: draft-ietf-quic-recovery-latest
+    author:
+      -
+        ins: J. Iyengar
+        name: Jana Iyengar
+        org: Fastly
+        role: editor
+      -
+        ins: I. Swett
+        name: Ian Swett
+        org: Google
+        role: editor
+
 informative:
 
   AEBounds:
@@ -73,23 +90,6 @@ informative:
         ins: M. Bishop
         name: Mike Bishop
         org: Microsoft
-        role: editor
-
-  QUIC-RECOVERY:
-    title: "QUIC Loss Detection and Congestion Control"
-    date: {DATE}
-    seriesinfo:
-      Internet-Draft: draft-ietf-quic-recovery-latest
-    author:
-      -
-        ins: J. Iyengar
-        name: Jana Iyengar
-        org: Fastly
-        role: editor
-      -
-        ins: I. Swett
-        name: Ian Swett
-        org: Google
         role: editor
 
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -619,6 +619,55 @@ The alert level of all TLS alerts is "fatal"; a TLS stack MUST NOT generate
...
+
+QUIC endpoints MUST retain keys until they are no longer needed.  If packets
+from a lower encryption level contain CRYPTO frames, frames that retransmit that
+data MUST be sent at the same encryption level.  Similarly, an endpoint
+generates acknowledgements for packets at the same encryption level as the
+packet being acknowledged.  Thus, it is possible that keys for a lower
+encryption level are needed for a short time after keys for a newer encryption
+level are available.
+
+An endpoint can only discard keys for a given encryption level only after it has
```
MikeBishop: Pick an only, but only one.

draft-ietf-quic-tls.md
```diff
@@ -619,6 +619,55 @@ The alert level of all TLS alerts is "fatal"; a TLS stack MUST NOT generate
...
+both received and acknowledged all CRYPTO frames for that encryption level and
+when it has had all CRYPTO frames for that encryption level acknowledged by its
+peer.  However, this does not guarantee that no further packets will need to be
+received or sent at that encryption level because a peer might not have received
+all the acknowledgements necessary to reach the same state.
```
MikeBishop: This paragraph seems duplicative of the preceding one.  Can they be consolidated?  Or if they are substantially different, the difference highlighted?

draft-ietf-quic-tls.md
```diff
@@ -619,6 +619,55 @@ The alert level of all TLS alerts is "fatal"; a TLS stack MUST NOT generate
...
+
+After all CRYPTO frames for a given encryption level have been either sent or
+received and the corresponding acknowledgments have been received or sent, an
```
MikeBishop: Given that each encryption level other than 0-RTT includes CRYPTO frames in both directions, I'd question the use of "either" here -- after all outgoing CRYPTO frames have been sent and all expected CRYPTO frames have been received, and the corresponding acknowledgements have also been sent and received.

draft-ietf-quic-tls.md
```diff
@@ -619,6 +619,55 @@ The alert level of all TLS alerts is "fatal"; a TLS stack MUST NOT generate
...
+endpoint starts a timer.  To limit the effect of packet loss around a change in
+keys, endpoints MUST retain packet protection keys for that encryption level for
+at least three times the current Retramsmission Timeout (RTO) interval as
+defined in {{QUIC-RECOVERY}}.  Retaining keys for this interval allows packets
+containing CRYPTO or ACK frames at that encryption level to be sent if packets
+are determined to be lost or new packets require acknowledgment.  While this
+timer is running, an endpoint MUST use the most recent packet protection keys
+for all packets, except to protect packets containing CRYPTO and ACK frames for
+the older encryption level.  These packets MAY also include PADDING frames.
```
MikeBishop: Talking about the "most recent packet protection keys" feels a little jarring here, since you're actually talking about encryption levels in this section.  This doesn't flow as clearly if you mix in key updates at the same time.

Perhaps "New data MUST be sent at the highest currently-available encryption level, but ACK frames and retransmissions of CRYPTO frames from prior encryption levels continue to be sent at that encryption level."

draft-ietf-quic-tls.md
```diff
@@ -619,6 +619,55 @@ The alert level of all TLS alerts is "fatal"; a TLS stack MUST NOT generate
...
+
+Once this timer expires, an endpoint MUST NOT either accept or generate new
+packets using those packet protection keys.  An endpoint can discard packet
+protection keys for that encryption level.
+
+An endpoint can update keys multiple times (see {{key-update}}) while this timer
+runs.  In that case, packets protected with the newest packet protection keys
+and packets sent two updates prior will appear to use the same keys.  After the
+handshake is complete, endpoints only need to maintain the two latest sets of
+packet protection keys and MAY discard older keys.
```
MikeBishop: Make the transition to discussing key updates more explicit, if you need to do it in this section.  Maybe, "Key updates in the 1-RTT encryption level do not require that keys from other encryption levels have already been discarded."  I'd be inclined to move the rest of the key update discussion to that section and leave it with that statement.  

Wherever that discussion lands, it feels like these two paragraphs need to merge.  Something like "Frequent key updates could create confusion about which generation of 1-RTT keys were used to protect the packet.  To mitigate this, key updates can be performed only once per RTT.  This means...."

draft-ietf-quic-tls.md
martinthomson: Yes, I think so.  The previous paragraph is intended to highlight the reasons why keys might be retained in a general sense.  The second paragraph is intended to be more precise about what the requirements are.  A small tweak to the lead-in of each should help emphasize the difference.

draft-ietf-quic-tls.md
martinthomson: The mitigation is a structural one - we don't say "don't update too often", instead it's more or less impossible to update that often (peers could update more often if they synchronized their efforts, but the floor is half an RTT).  I'll reword with these comments in mind though.

draft-ietf-quic-tls.md
```diff
@@ -625,47 +625,51 @@ After QUIC moves to a new encryption level, packet protection keys for previous
...
+the packet being acknowledged.  Thus, it is possible that keys for a lower
 encryption level are needed for a short time after keys for a newer encryption
 level are available.
 
-An endpoint can only discard keys for a given encryption level only after it has
-both received and acknowledged all CRYPTO frames for that encryption level and
-when it has had all CRYPTO frames for that encryption level acknowledged by its
+An endpoint cannot discard keys for a given encryption level unless it has both
+received and acknowledged all CRYPTO frames for that encryption level and when
+it all CRYPTO frames for that encryption level have been acknowledged by its
```
MikeBishop: "and when ~it~ all"

PR #1637, opened 2018-08-07
Discuss Application-Limited Sending

### LINKED ISSUE #1619: Discuss application-limited congestion control in Recovery
There's no mention of this in the draft, and there should be, both in regards to CWND increases and reductions in response to loss.

ianswett:
Fixes #1619

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1050,7 +1050,6 @@ after idle periods, such as those proposed for TCP in {{?RFC7661}}.
 The congestion window should not be increased in slow start or congestion
 avoidance when it is not fully utilized.  The congestion window could be
 under-utilized due to insufficient application data or flow control credit.
-limits.
 
 When the sender is pacing (see {{pacing}}) packets, the sender may be unable
 to use the full congestion window for a period of time after receiving
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1004,9 +1004,10 @@ under-utilized due to insufficient application data to send or flow control
 limits.
 
 When the sender is pacing (see {{pacing}}) packets, the sender may be unable
-to use the full congestion window for a period of time after receiving an
-ACK, due to pacing.  In this case, the sender should not consider themselves
-application limited and should allow the congestion window to increase.
+to use the full congestion window for a period of time after receiving
+acknowledgements, due to pacing.  In this case, the sender should not consider
+themselves application limited and should allow the congestion window to
+increase.
 
 ## Discarding Packet Number Space State
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1162,7 +1162,8 @@ acked_packet from sent_packets.
        // Do not increase congestion window in recovery period.
        return
      if (IsAppLimited())
-       // Do not increase congestion_window if application limited.
+       // Do not increase congestion_window if application
+       // limited.
        return
      if (congestion_window < ssthresh):
        // Slow start.
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1005,7 +1005,7 @@ limits.
 
 When the sender is pacing(see {{pacing}}) packets, the sender may be unable
 to use the full congestion window for a period of time after receiving an
-ACK due to pacing.  In this case, the sender should not consider themselves
+ACK, due to pacing.  In this case, the sender should not consider themselves
 application limited and should allow the congestion window to increase.
 
 ## Discarding Packet Number Space State
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -944,12 +944,6 @@ The recovery period limits congestion window reduction to once per round trip.
 During recovery, the congestion window remains unchanged irrespective of new
 losses or increases in the ECN-CE counter.
 
-## Application Limited Sending
-
-If the sender is sufficiently application limited that the congestion window is
-not fully utilized, the congestion window should not be increased in slow start
-or congestion avoidance.
-
 ## Probe Timeout
 
 Probe packets MUST NOT be blocked by the congestion controller.  A sender MUST
@@ -968,8 +962,7 @@ response of collapsing the congestion window on persistent congestion is
 functionally similar to a sender's response on a Retransmission Timeout (RTO) in
 TCP {{RFC5681}}.
 
-
-## Pacing
+## Pacing {#pacing}
 
 This document does not specify a pacer, but it is RECOMMENDED that a sender pace
 sending of all in-flight packets based on input from the congestion
@@ -1002,6 +995,18 @@ paces the sending of any packets in excess of the initial congestion window.
 
 A sender MAY implement alternate mechanisms to update its congestion window
 after idle periods, such as those proposed for TCP in {{?RFC7661}}.
+
+## Application Limited Sending
+
+The congestion window should not be increased in slow start or congestion
+avoidance when it is not fully utilized.  The congestion window could be
+under-utilized due to insufficient application data to send or flow control
+limits.
+
+When the sender is pacing(see {{pacing}}) packets, the sender may be unable
+to use the full congestion window for a period of time after receiving an
+ACK due to pacing.  In this case, the sender should not consider themselves
+application limited and should allow the congestion window to increase.
 
 ## Discarding Packet Number Space State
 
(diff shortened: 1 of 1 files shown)


janaiyengar:
Both app-limited sending and also receiver-window limited sending should be incorporated in the text here.

draft-ietf-quic-recovery.md
```diff
@@ -989,6 +988,13 @@ The recovery period limits congestion window reduction to once per round trip.
 During recovery, the congestion window remains unchanged irrespective of new
 losses or increases in the ECN-CE counter.
 
+## Application Limited Sending
+
+If the sender is sufficiently application limited that the congestion window is
+not fully utilized, the congestion window should not be increased in slow start
+or congestion avoidance.  Senders should consider themselves application limited
+if bytes in flight when receiving an ACK frame are more than a max datgram size
+less than the congestion window.
```
janaiyengar: This last sentence is a run-on and unnecessarily detailed. I would simply say something like "if the congestion window is not fully utilized." The detail of kMaxDatagramSize is a rounding thing, which is probably too much detail in this descriptive text.

draft-ietf-quic-recovery.md
```diff
@@ -1115,13 +1121,19 @@ acked_packet from sent_packets.
...
 
    OnPacketAckedCC(acked_packet):
      // Remove from bytes_in_flight.
      bytes_in_flight -= acked_packet.bytes
      if (InRecovery(acked_packet.packet_number)):
        // Do not increase congestion window in recovery period.
        return
+     if (IsAppLimited())
+       // Do not increase congestion_window if application limited.
+       return
```
nibanks: This gets into implementation specifics a bit, but I don't think the pseudocode can be as trivial as this.

Naively, if this function is called for two different packets in a row, without a corresponding send operation being performed between them, then only the first `OnPacketAckedCC` ever actually increases `congestion_window` because the previous call just decremented `bytes_in_flight` and incremented `congestion_window` guaranteeing that `IsAppLimited()` will always return false for future packets.

In a slightly more intelligent interpretation of this, where you actually preform this logic for all the bytes acknowledged in a whole ACK frame in a single call, you can still have the same problem if you end up processing two ACK frames back to back before you end up doing a send. I have some ideas on how to solve that (at least for my implementation) but I was wondering if this should be mentioned (if not pseudocoded) here.

draft-ietf-quic-recovery.md
ianswett: In that case, I think I should just remove that sentence?

draft-ietf-quic-recovery.md
```diff
@@ -1115,13 +1119,19 @@ acked_packet from sent_packets.
 ~~~
    InRecovery(packet_number):
      return packet_number <= end_of_recovery
+   
+   IsAppLimited():
+     bytes_in_flight >= congestion_window - kMaximumDatagramSize
```
MikeBishop: Isn't this backward?  Bytes in flight can't be greater than the congestion window, but if the bytes in flight are within one datagram of the window, you're clearly not limited.

draft-ietf-quic-recovery.md
```diff
@@ -1115,13 +1119,19 @@ acked_packet from sent_packets.
 ~~~
    InRecovery(packet_number):
      return packet_number <= end_of_recovery
+
```
MikeBishop: Build failure is trailing whitespace on this line.

draft-ietf-quic-recovery.md
nibanks: I agree this is backward. The `>=` should just be `<` or `<=`. But as far as tracking `IsAppLimited()` or whatever you want to call it, I would just suggest a modification to the code:

Instead of using `bytes_in_flight` here and the helper function `IsAppLimited`, introduce a new variable `window_full` that is only set in `OnPacketSent`. At the end of each call to that function, if `bytes_in_flight` is close enough to `congestion_window` set it to true, otherwise false. Then below, you use that variable to decide if the cwnd should be increased..

martinthomson: That seems better, yes.  When you say "close enough", I assume that you mean "so close that you couldn't send anything meaningful", something like headerlen+sizeof(cid)+16+sizeof(STREAM frame overhead), so maybe sizeof(cid)+32 ?

nibanks: I was specifically not defining "close enough". I am fine with a single datagram size or something smaller.

mirjak:
Actually as mentioned in #1619 one would need to also reduce the window if application-limited and the send window goes below the congestion window. See https://datatracker.ietf.org/doc/rfc7661/ for TCP.

draft-ietf-quic-recovery.md
ianswett: Agreed, getting this right is complex.  I am going to not define the IsAppLimited() method, but document what it's intent is.

draft-ietf-quic-recovery.md
```diff
@@ -999,6 +996,18 @@ paces the sending of any packets in excess of the initial congestion window.
...
 after idle periods, such as those proposed for TCP in {{?RFC7661}}.
 
+## Application Limited Sending
+
+The congestion window should not be increased in slow start or congestion
+avoidance when it is not fully utilized.  The congestion window could be
+under-utilized due to insufficient application data to send or flow control
+limits.
+
+When the sender is pacing(see {{pacing}}) packets, the sender may be unable
```
```suggestion
When the sender is pacing (see {{pacing}}) packets, the sender may be unable
```

draft-ietf-quic-recovery.md
```diff
@@ -999,6 +996,18 @@ paces the sending of any packets in excess of the initial congestion window.
...
+to use the full congestion window for a period of time after receiving an
+ACK, due to pacing.  In this case, the sender should not consider themselves
```
```suggestion
acknowledgment.  In this case, the sender should not consider themselves
```

draft-ietf-quic-recovery.md
```diff
@@ -999,6 +996,18 @@ paces the sending of any packets in excess of the initial congestion window.
...
+application limited and should allow the congestion window to increase.
```
martinthomson: I'm not really following the logic here.  I think that you need more words.

The reason that the congestion window is limited in this fashion is that without being exercised, there is no assurance that this much data can be sent.  If there is a sudden increase in demand from the application such that the inflated limit is now used, that results in using an untested limit, which might result in severe congestion.

draft-ietf-quic-recovery.md
ianswett: I think I mean ACK, as in the frame?

draft-ietf-quic-recovery.md
ianswett: SG, maybe @janaiyengar or @nibanks will have a suggestion?

nibanks: Maybe it's just a case of too much detail. I think folks should understand that just because a pacing interval has currently been used up, that's not considered application limited? So, maybe, just don't have this at all?

draft-ietf-quic-recovery.md
martinthomson: You need more than an ACK, you need one that contains fresh acknowledgments, right?

ianswett: True, changed to acknowledgements.

draft-ietf-quic-recovery.md
```diff
@@ -1047,6 +1045,19 @@ paces the sending of any packets in excess of the initial congestion window.
 A sender MAY implement alternate mechanisms to update its congestion window
 after idle periods, such as those proposed for TCP in {{?RFC7661}}.
 
+## Application Limited Sending
+
+The congestion window should not be increased in slow start or congestion
+avoidance when it is not fully utilized.  The congestion window could be
+under-utilized due to insufficient application data to send or flow control
```
```suggestion
under-utilized due to insufficient application data or flow control credit.
```

draft-ietf-quic-recovery.md
```diff
@@ -1047,6 +1045,19 @@ paces the sending of any packets in excess of the initial congestion window.
...
+limits.
```
```suggestion
```

draft-ietf-quic-recovery.md
```diff
@@ -1047,6 +1045,19 @@ paces the sending of any packets in excess of the initial congestion window.
...
+
+When the sender is pacing (see {{pacing}}) packets, the sender may be unable
+to use the full congestion window for a period of time after receiving
+acknowledgements, due to pacing.  In this case, the sender should not consider
+themselves application limited and should allow the congestion window to
+increase.
```
janaiyengar: Suggested rephrasing:  "A sender that uses pacing (see {{pacing}}) might delay sending of packets and might not fully utilize the congestion window due to this delay.  A sender should not consider itself application limited if it might have utilized the entire congestion window without pacing delay."

"When the sender paces packets (see {{pacing}}), it might not send 
to use the full congestion window for a period of time after receiving
acknowledgements, due to pacing.  In this case, the sender should not consider
themselves application limited and should allow the congestion window to
increase."

ianswett: Accepted with minor tweaks.

PR #1640, opened 2018-08-08
Fixed length CID or encode length somehow

### LINKED ISSUE #1639: Stateless Reset and CID Recovery
[6.13.4.2](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#rfc.section.6.13.4.2) discusses the need to request a non-zero-length CID if you want to be able to generate Stateless Resets:

> An endpoint that loses state can use the same method to generate a valid Stateless Reset Token. The connection ID comes from the packet that the endpoint receives.
> 
> This design relies on the peer always sending a connection ID in its packets so that the endpoint can use the connection ID from a packet to reset the connection. An endpoint that uses this design cannot allow its peers to send packets with a zero-length destination connection ID.

However, it doesn't discuss the fact that the length of the CID isn't identifiable from a short-header packet, and therefore even this algorithm works only if you either use a constant CID length or embed a length indicator in your CIDs.  This might be a worthwhile piece to mention.

martinthomson:
Closes #1639.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2679,7 +2679,9 @@ endpoint receives.
 This design relies on the peer always sending a connection ID in its packets so
 that the endpoint can use the connection ID from a packet to reset the
 connection.  An endpoint that uses this design cannot allow its peers to send
-packets with a zero-length destination connection ID.
+packets with a zero-length destination connection ID and need to either use a
+the same connection ID length for all connections or ensure that the connection
+ID encodes its own length in a common fixed portion.
 
 Revealing the Stateless Reset Token allows any entity to terminate the
 connection, so a value can only be used once.  This method for choosing the


draft-ietf-quic-transport.md
```diff
@@ -2679,7 +2679,9 @@ endpoint receives.
 This design relies on the peer always sending a connection ID in its packets so
 that the endpoint can use the connection ID from a packet to reset the
 connection.  An endpoint that uses this design cannot allow its peers to send
-packets with a zero-length destination connection ID.
+packets with a zero-length destination connection ID and need to either use a
+the same connection ID length for all connections or ensure that the connection
+ID encodes its own length in a common fixed portion.
```
mikkelfj: It is not necessary to have a fixed portion. You can use bit 7 of each octet to indicate if there is more.

draft-ietf-quic-transport.md
```diff
@@ -2679,7 +2679,9 @@ endpoint receives.
 This design relies on the peer always sending a connection ID in its packets so
 that the endpoint can use the connection ID from a packet to reset the
 connection.  An endpoint that uses this design cannot allow its peers to send
-packets with a zero-length destination connection ID.
+packets with a zero-length destination connection ID and need to either use a
```
MikeBishop: "use ~a~ the same"

draft-ietf-quic-transport.md
MikeBishop: Yeah, that might be a little more specific than is needed.  You could just say "...connection ID encodes its own length in a method the server can recover without state."

draft-ietf-quic-transport.md
ianswett: need -> needs

draft-ietf-quic-transport.md
ianswett: Agreed, you could even use trial decryption if you really wanted :)

draft-ietf-quic-transport.md
MikeBishop: Good catch.  I was reading "packets ... need", but it's actually "An endpoint ... needs"

draft-ietf-quic-transport.md
```diff
@@ -2679,7 +2679,9 @@ endpoint receives.
 This design relies on the peer always sending a connection ID in its packets so
 that the endpoint can use the connection ID from a packet to reset the
 connection.  An endpoint that uses this design cannot allow its peers to send
-packets with a zero-length destination connection ID.
+packets with a zero-length destination connection ID and need to either use
+the same connection ID length for all connections or encode the length of
+the connection ID such that it can be recovered without state.
```
ekr: Late to the party, but this is pretty hard to read. I believe that it is intended to say:

(1) must not use a zero-length CID
(2) must  have CID length be either (a) constant or (b) encoded in the CID itself

But the use of "cannot allow" might instead say you can't do both of these together instead. I would instead suggest.

"an endpoint that uses this design MUST either use the same connection ID length for all connections or encode the length of the connection ID such that it can be recovered without state. In addition, it MUST NOT provide zero-length destination connection IDs to its peers".

PR #1643, opened 2018-08-09
Specify behavior on incomplete requests

### LINKED ISSUE #1596: How should an HQ server respond to RST_STREAM?
If a client sends a RST_STREAM with code HTTP_REQUEST_CANCELLED, the server also needs to reset its end of the stream.  As I read the spec, I'm unsure what error code best suits this purpose.

1. STOPPING says it is reserved by the transport in response to STOP_SENDING, so that's not it.
2. HTTP_REQUEST_CANCELLED is allowed S->C, but only if no processing of the request has occurred.  What if part of the request has been processed? 

A corollary question is, if a server cancels a request with a reset_stream, and does not send a STOP_SENDING, is the client expected to reply with HTTP_REQUEST_CANCELLED, a different error code, or continue sending until told to stop?  CANCELLED seems ok here, but I dislike the asymmetry.

MikeBishop:
Fixes #1596 (and an oversight I caught in passing) by specifying server behavior if it reaches the end of readable data on a stream and hasn't parsed a full HTTP request yet.  There are a couple cases here, and I think this generically covers all the interesting ones.

- **Client sends FIN early.**  Server sees truncated request, but non-error end of stream.
- **Client sends RST_STREAM part-way through request.**  Server sees truncated request with read error.
- **Client sends a RST_STREAM (but not a STOP_SENDING) after sending a fully-formed request.**
    - **...and packet loss occurred.**  No retransmissions, so same as sending it part-way through request.
    - **...after the server has read the request and generated / begun generating a response.**  The server probably never realizes and responds merrily.  If you didn't want it, you should have sent a STOP_SENDING.
    - **...but before the server has read all the data out of the buffer.**  Depends on implementation; transport might discard unread data or deliver all the data anyway.  Either way, this is identical to one of the two previous cases.

We already discuss cancelling requests, so I added a suggestion that clients go ahead and include a STOP_SENDING if they reset their request stream.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -320,11 +320,11 @@ and cleanly closing its streams. Clients MUST NOT discard complete responses as
 a result of having their request terminated abruptly, though clients can always
 discard responses at their discretion for other reasons.
 
-Changes to the state of a request stream, including receiving a RST_STREAM
-(regardless of error code), do not affect the state of the server's response.
-Servers do not abort a response in progress solely due to a state change on the
-request stream. However, if the request stream terminates without containing a
-usable HTTP request, the server SHOULD abort its response with the error code
+Changes to the state of a request stream, including receiving a RST_STREAM with
+any error code, do not affect the state of the server's response. Servers do not
+abort a response in progress solely due to a state change on the request stream.
+However, if the request stream terminates without containing a usable HTTP
+request, the server SHOULD abort its response with the error code
 HTTP_INCOMPLETE_REQUEST.
 
 ### Header Formatting and Compression
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -322,10 +322,10 @@ discard responses at their discretion for other reasons.
 
 Changes to the state of a request stream, including receiving a RST_STREAM
 (regardless of error code), do not affect the state of the server's response.
-Servers MUST NOT abort a response in progress solely due to a state change on
-the request stream. However, if the request stream terminates without containing
-a fully-formed HTTP request, the server SHOULD abort its response with the error
-code HTTP_INCOMPLETE_REQUEST.
+Servers do not abort a response in progress solely due to a state change on the
+request stream. However, if the request stream terminates without containing a
+usable HTTP request, the server SHOULD abort its response with the error code
+HTTP_INCOMPLETE_REQUEST.
 
 ### Header Formatting and Compression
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -317,8 +318,14 @@ client abort transmission of a request without error by triggering a QUIC
...
 a result of having their request terminated abruptly, though clients can always
-discard responses at their discretion for other reasons.  Servers MUST NOT
-abort a response in progress as a result of receiving a solicited RST_STREAM.
+discard responses at their discretion for other reasons.
+
+Changes to the state of a request stream, including receiving a RST_STREAM
+(regardless of error code), do not affect the state of the server's response.
+Servers MUST NOT abort a response in progress solely due to a state change on
+the request stream. However, if the request stream terminates without containing
+a fully-formed HTTP request, the server SHOULD abort its response with the error
```
martinthomson: s/fully-formed/usable/ perhaps.

A server can often produce a response based purely on header fields.  That's the case that we specifically want to support with STOP_SENDING (a 404 on a POST).  This implies that the message needs to be complete, which isn't necessary.

draft-ietf-quic-http.md
```diff
@@ -317,8 +318,14 @@ client abort transmission of a request without error by triggering a QUIC
 STOP_SENDING with error code HTTP_EARLY_RESPONSE, sending a complete response,
 and cleanly closing its streams. Clients MUST NOT discard complete responses as
 a result of having their request terminated abruptly, though clients can always
-discard responses at their discretion for other reasons.  Servers MUST NOT
-abort a response in progress as a result of receiving a solicited RST_STREAM.
+discard responses at their discretion for other reasons.
+
+Changes to the state of a request stream, including receiving a RST_STREAM
+(regardless of error code), do not affect the state of the server's response.
+Servers MUST NOT abort a response in progress solely due to a state change on
```
martinthomson: I don't think that this is a normative requirement on reflection.  Maybe just "Servers do not abort a response..." or "Servers are not expected to abort a response..."

draft-ietf-quic-http.md
```diff
@@ -317,8 +318,14 @@ client abort transmission of a request without error by triggering a QUIC
 STOP_SENDING with error code HTTP_EARLY_RESPONSE, sending a complete response,
 and cleanly closing its streams. Clients MUST NOT discard complete responses as
 a result of having their request terminated abruptly, though clients can always
-discard responses at their discretion for other reasons.  Servers MUST NOT
-abort a response in progress as a result of receiving a solicited RST_STREAM.
+discard responses at their discretion for other reasons.
+
+Changes to the state of a request stream, including receiving a RST_STREAM
+(regardless of error code), do not affect the state of the server's response.
```
martinthomson: s/(regardless of error code)/with any error code/

PR #1648, opened 2018-08-10
Add another condition to idle timeout

siyengar:
When implementing the new idle timeout, I noticed that there might be one condition missing from this. 

Previously if you send the first packet in the connection or send the first packet after all your ACKs are received / packets lost, then your idle timeout will not be set.

This changes it so that when there are no outstanding packets and you send a packet, it resets the idle timeout.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2491,9 +2491,14 @@ A connection that remains idle for longer than the advertised idle timeout (see
 draining state when the idle timeout expires.
 
 Each endpoint advertises their own idle timeout to their peer. The idle timeout
-starts from the last packet received, or the first packet sent after the last
-received packet.  The latter condition ensures that initiating new activity
-postpones a timeout.
+is started if any of the following events occur:
+
+- A new packet is received
+- The first packet is sent after the last packet received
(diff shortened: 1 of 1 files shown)


martinthomson:
Is this new condition better than the one @ianswett suggested?  That is, should it replace that condition instead of adding a new one?

If you think about it, the last packet you send before an idle period is almost always sent with no other packets outstanding.  If you receive an ACK, then you send nothing.  Otherwise, you send an ACK after a short delay, which would generally fit with this new definition.

siyengar:
You have a good point. The only difference I see in these 2 conditions is that the previous one covers the case that you have a few packets outstanding and then you receive a packet. However it's probably not that useful since presumably if you have a few packets outstanding, you expect to get feedback about them still. So I think that we can actually replace the condition with this new one of outstandingpackets.size = 0. It's also much simpler to implement as well.

martinthomson:
What about outstanding ACK-only packets?  Those won't be declared lost.  Let's say that the last thing that happened was that an endpoint receives a packet containing something useful (e.g., STREAM).  It sends a packet with an ACK, then goes idle.  That ACK-only packet is technically still outstanding.  See #1652.

Issue #1644, opened 2018-08-09
QPACK improvement: wrap absolute index values


dtikhonov:
_What follows is a modified version of the [original proposal on the mailing list](https://www.ietf.org/mail-archive/web/quic/current/msg04515.html).  It includes changes to address a few problems identified by @martinthomson._

Abstract
--------

The practical and theoretical problems caused by the unbounded absolute index values in the current QPACK Internet Draft are discussed.  The proposal to impose a limit on the range of absolute index values by wrapping them is presented.

Background
----------

The current QPACK draft states that

> [the] first entry inserted has an absolute index of "1"; indices increase sequentially with each insertion.

No upper limit is specified.  This is reasonable, as it is impossible to run out of numbers.

Nevertheless, an absolute index whose value is ever-increasing creates significant problems.

Problem 1: Compression Impact
-----------------------------

A large absolute index value has a negative effect on compression.  The unadjusted value of the
absolute index is used in the _Largest Reference_ field in the _Header Data Prefix_.

Larger values mean that more bytes are necessary to encode them.  For example, using 8-bit prefix integer encoding, values larger than  254 encode to 2 bytes, larger than 382 to 3 bytes, and larger than 16638 to 4 bytes.

Relatively quickly, the overhead associated with communicating the absolute index doubles and then triples.

Problem 2: Theoretical Concerns
-------------------------------

Why should compression at _t(1)_ be worse than at _t(0)_?  An ability to operate at the optimum level in perpetuity is a good property for a protocol to have.

Problem 3: Implementation Efficiency
------------------------------------

A QPACK implementation has to place some limits on the value of the absolute index -- simply because it has to pick an integral type to store it.  For example, uint64_t is likely large enough for all uses of the protocol, but it is wider than uint32_t and may be more expensive to process.  On other platforms, uint16_t may be the fastest integer; the unbounded nature of the absolute index would rule out its use.

Such limit also requires an [extra check](https://github.com/litespeedtech/ls-qpack/blob/122d1163eaca359f5d79cd608722b1b19acad15f/src/lsqpack.c#L1110) to cease inserting entries into the dynamic table once the maximum value for the integer is reached.  (One can run out of numbers in practice...)

Solution: Have the Absolute Index Wrap Around
---------------------------------------------

The main insight of the proposal is that the dynamic table has a bounded number of entries, that both encoder and decoder know what this number is, and that both can use this information to place an absolute index value communicated to it by its peer at the correct place in the sequence.

[...]

PR #1650, opened 2018-08-10
Start indexing dynamic QPACK table beginning at zero

dtikhonov:
Based on discussion in [issue #1644](https://github.com/quicwg/base-drafts/issues/1644#issuecomment-411949968).

To avoid potential for confusion around zero the raw _Largest Reference_ value and zero the smallest valid absolute index, I introduced the term "specified:" the _Largest Reference_ value is either specified or not.


afrind:
Is this intended to stand on its own or only as part of the stack with #1644.  @martinthomson had a comment on that issue about what it means to be not-specified.  This field is required so how can it be not specified?

dtikhonov:
After discussing #1644 at length, I am no longer convinced that starting the dynamic table at 0 is better that 1.  I now think it makes no difference.  Closing.

PR #1652, opened 2018-08-13
Use the 'no outstanding' rule to defer idle timeout

martinthomson:
Closes #1648.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2499,9 +2499,11 @@ A connection that remains idle for longer than the advertised idle timeout (see
 draining state when the idle timeout expires.
 
 Each endpoint advertises their own idle timeout to their peer. The idle timeout
-starts from the last packet received, or the first packet sent after the last
-received packet.  The latter condition ensures that initiating new activity
-postpones a timeout.
+starts from the last packet received.  In order to ensure that initiating new
+activity postpones an idle timeout, an endpoint restarts this timer when sending
+a packet if there are no other packets outstanding.  This condition only applies
+to packets that contain other than ACK or PADDING frames; those packets could
+remain outstanding longer and prevent the timeout from being extended.
 
 The value for an idle timeout can be asymmetric.  The value advertised by an
 endpoint is only used to determine whether the connection is live at that
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2501,9 +2501,11 @@ draining state when the idle timeout expires.
 Each endpoint advertises their own idle timeout to their peer. The idle timeout
 starts from the last packet received.  In order to ensure that initiating new
 activity postpones an idle timeout, an endpoint restarts this timer when sending
-a packet if there are no other packets outstanding.  This condition only applies
-to packets that contain other than ACK or PADDING frames; those packets could
-remain outstanding longer and prevent the timeout from being extended.
+a packet.  An endpoint does not postpone the idle timeout if another packet has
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2499,9 +2499,11 @@ A connection that remains idle for longer than the advertised idle timeout (see
 draining state when the idle timeout expires.
 
 Each endpoint advertises their own idle timeout to their peer. The idle timeout
-starts from the last packet received, or the first packet sent after the last
-received packet.  The latter condition ensures that initiating new activity
-postpones a timeout.
+starts from the last packet received.  In order to ensure that initiating new
+activity postpones an idle timeout, an endpoint restarts this timer when sending
+a packet if there are no other packets outstanding.  This condition only applies
+to packets that contain other than ACK or PADDING frames; those packets could
```
ianswett: "that contain other" is awkward, how about "that contain frames other"

draft-ietf-quic-transport.md
```diff
@@ -2499,9 +2499,11 @@ A connection that remains idle for longer than the advertised idle timeout (see
...
+remain outstanding longer and prevent the timeout from being extended.
```
ianswett: How about a separate sentence "Packets that contain only ACK or PADDING frames may never be acknowledged, so they do not extend the timeout."

draft-ietf-quic-transport.md
```diff
@@ -2499,9 +2499,11 @@ A connection that remains idle for longer than the advertised idle timeout (see
 draining state when the idle timeout expires.
 
 Each endpoint advertises their own idle timeout to their peer. The idle timeout
-starts from the last packet received, or the first packet sent after the last
-received packet.  The latter condition ensures that initiating new activity
-postpones a timeout.
+starts from the last packet received.  In order to ensure that initiating new
+activity postpones an idle timeout, an endpoint restarts this timer when sending
+a packet if there are no other packets outstanding.  This condition only applies
```
ianswett: How about "In order to ensure that initiating new activity postpones an idle timeout, an endpoint restarts this timer when sending a packet containing frames besides ACK or PADDING and no other packets are outstanding."

draft-ietf-quic-transport.md
ianswett: This comment is obsolete if you make the above change or something similar.

ianswett:
On a high level, there's some remaining vagueness about what outstanding means.  Particularly now that PADDING frames count towards bytes in flight, but do not elicit an ACK.

mikkelfj:
Surely an ACK only packet does not prevent idle timeout from refreshing? It merely does not controbute to progress?

martinthomson:
Receiving an ACK-only packet will reset the timer.  Sending one might.  If there are no other outstanding packets, then an ACK is sent after receiving a packet and after waiting for any ACK delay.  That will refresh the timer.  However, you won't send another ACK-only packet until a packet is declared lost, which only happens when new packets arrive.

mikkelfj:
I'm just find the word "prevent" confusing here:

"Packets that contain only ACK or
 PADDING frames are not acknowledged until an endpoint has other frames to send,
 so they could prevent the timeout from being refreshed."

martinthomson:
An ACK- and PADDING-only packet would not get acknowledged and so remain unacknowledged indefinitely.  If the rule is "don't refresh if you have unacknowledged packets", then those packets would prevent the timeout from being restarted when a new packet is sent.

mikkelfj:
I see - but why don't you exclude those packets. Simplicity?

Issue #1592, opened 2018-07-20
Single Stream of Bytes for CRYPTO Frames


nibanks:
Currently, the spec states:

> There is a separate flow of cryptographic handshake data in each encryption level, each of which starts at an offset of 0. This implies that each encryption level is treated as a separate CRYPTO stream of data.

This means you must track set of bytes in each encryption level (Initial, 0-RTT, Handshake and 1-RTT) as a separate stream and manage them accordingly; separate outstanding sent byte ranges, recovery ranges, acked ranges.

Additionally, it makes it a little unclear where one 'stream' stops and another 'starts'. For instance, in the 0-RTT case, the client (is supposed to) sends EOED with 0-RTT key, immediately followed by (ideally in the same UDP datagram, but not required) FIN with the Handshake key. If the 0-RTT EOED is lost, but the FIN isn't it could look like the client just skipped EOED. Instead, if they were all in the same stream, with FIN's offset starting where EOED left off, it would be easier to track this.

For these reasons, I propose each CRYPTO stream in each encryption level's offset starts where the previous level stopped.

marten-seemann:
Wouldn’t that be equivalent to going back to a single CRYPTO stream, which changes encryption levels at various offsets?
If I remember correctly, this was one of the main pain points with the old design, so I’d be unhappy to revert this change.

nibanks:
I wouldn't say we had exactly that, as now we are much more explicit with encryption level (as there are 4 different ones). Why do you say that was a pain point? I feel that is a much cleaner and explicit design to implement.

ianswett:
I have the single stream approach implemented today, and it's ugly on the send side, because I need a very weird datastructure to determine what was sent at what encryption level and make sure there's no cross-encryption level bundling/etc.

With a stream per level, it's much clearer what's going on, and I can re-use most of the standard stream code.  I need to actually implement this, so maybe there are some issues I'm not anticipating, but I know the single stream approach is ugly.

nibanks:
Using a standard stream for this will bring in 4 times the overhead. What is your FC limit for each stream? What tracking variables do you store for each? Having one stream with 4 offsets (Initial is implicitly offset 0) is a much more efficient design, IMO. I also have the single stream approach implemented (because that's what I thought  we were supposed to do, before I reread the spec) and I have a much simpler implementation that using 4 different streams.

dtikhonov:
[...]

PR #1654, opened 2018-08-13
Separate key/secret availability from usage

### LINKED ISSUE #1651: Ordering of crypto stream packets
I am doing interop tests and self tests with the new crypto stream encoding (draft-13 + PR). I think we have an ordering problem. If the 0-RTT packet containing the EOED message is lost, the QUIC server may receive the Client Finished message before receiving the EOED. If it submits that message to TLS, the exchange will fail, because the EOED was not received. The server should keep the Client Finished message in memory and only submit it after EOED is received. But the server does not really know that without having some inside knowledge of the TLS state -- and one of the goal of the new Crypto Stream design was to avoid requiring that knowledge.

Of course, one solution is to explain the problem in details. The server knows that 0-RTT was negotiated, and should refrain from submitting the epoch 2 CS data until the epoch 1 data has been received. Note that this does not work in theory, because there is no FIN mark in the CS frames, so the server does not really know when epoch 1 data is finished, but it will work in practice if the server assumes that epoch 1 data is always sent as a single frame. That's a big wart in the stack. 

Another solution would be to send the EOED in epoch 2, because then the ordering would be natural. 

Another solution would be to make sure that the "offset" in the CS frame is computed from the beginning of the session, instead of from the beginning of the epoch. 

The client can minimize the risk by ensuring that EOED and Finished are always transmitted and retransmitted as coalesced segments. That's easy to do for the initial transmission, but not quite as natural for the retransmission, since different packet contexts will have different timers, sequence numbers, etc.

For the details, consider the following typical session exchanges:

1) Basic connection:
```
CHello (250 bytes) => Send CS frame offset 0, length 250 with Initial packet Epoch 0.
       <-- SHello (160 bytes) 
               => Send CS frame offset 0, length 160 with Initial packet, Epoch 0
       <-- server first flight (1500 bytes)
               => Send CS frame offset 0, length 1000 with Handshake packet, Epoch 2
               => Send CS frame offset 1000, length 500 with Handshake packet, Epoch 2
Client Finished =>  Send CS frame offset 0, length 46 with Handshake packet Epoch 2.
<process everything in order and life is good >
```
2) 0-RTT connection:
```
CHello (250 bytes) => Send CS frame offset 0, length 250 with Initial packet Epoch 0.
       <-- SHello (160 bytes) 
               => Send CS frame offset 0, length 160 with Initial packet, Epoch 0
       <-- server first flight (1500 bytes)
               => Send CS frame offset 0, length 1000 with Handshake packet, Epoch 2
               => Send CS frame offset 1000, length 500 with Handshake packet, Epoch 2
EOED =>Send CS frame offset 0, length 4 with 0-RTT packet Epoch 1.
Client Finished =>  Send CS frame offset 0, length 46 with Handshake packet Epoch 2.
<process everything in order and life is good >
```
2) 0-RTT connection with inconvenient packet loss :
```
CHello (250 bytes) => Send CS frame offset 0, length 250 with Initial packet Epoch 0.
       <-- SHello (160 bytes) 
               => Send CS frame offset 0, length 160 with Initial packet, Epoch 0
       <-- server first flight (1500 bytes)
               => Send CS frame offset 0, length 1000 with Handshake packet, Epoch 2
               => Send CS frame offset 1000, length 500 with Handshake packet, Epoch 2
Packet loss: EOED =>Send CS frame offset 0, length 4 with 0-RTT packet Epoch 1.
Client Finished =>  Send CS frame offset 0, length 46 with Handshake packet Epoch 2.
Retransmit: EOED =>Send CS frame offset 0, length 4 with 0-RTT packet Epoch 1.
<process everything in order and the exchange fails >
```

martinthomson:
As we discovered, conflating these two signals creates problems.  The
only problem we're aware of is at the server when it consumes 0-RTT, but
it's serious enough that some clarification is in order.

This makes it clear that there are two events of interest, then expands
on the particular problem scenario.

Closes #1651.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -440,10 +440,12 @@ Important:
 
 ### Encryption Level Changes
 
-At each change of encryption level in either direction, TLS provides QUIC with
-the new level and the encryption keys.  These events are not asynchronous; they
-always occur immediately after TLS is provided with new handshake octets, or
-after TLS produces handshake octets.
+As keys for new encryption levels become available, TLS provides QUIC with those
+keys.  Separately, as TLS starts using keys at a given encryption level, TLS
+indicates to QUIC that it is now reading or writing with keys at that encryption
+level.  These events are not asynchronous; they always occur immediately after
+TLS is provided with new handshake octets, or after TLS produces handshake
+octets.
 
 If 0-RTT is possible, it is ready after the client sends a TLS ClientHello
 message or the server receives that message.  After providing a QUIC client with
@@ -451,12 +453,18 @@ the first handshake octets, the TLS stack might signal the change to 0-RTT
 keys. On the server, after receiving handshake octets that contain a ClientHello
 message, a TLS server might signal that 0-RTT keys are available.
 
-Note that although TLS only uses one encryption level at a time, QUIC may use
-more than one level. For instance, after sending its Finished message (using a
-CRYPTO frame in Handshake encryption) may send STREAM data (in 1-RTT
-encryption). However, if the Finished is lost, the client would have to
-retransmit the Finished, in which case it would use Handshake encryption.
+Although TLS only uses one encryption level at a time, QUIC may use more than
+one level. For instance, after sending its Finished message (using a CRYPTO
+frame at the Handshake encryption level) an endpoint can send STREAM data (in
+1-RTT encryption). If the Finished message is lost, the endpoint uses the
+Handshake encryption level to retransmit the lost message.
 
+In particular, server implementations need to be able to read packets at the
+Handshake encryption level before the final TLS handshake message at the 0-RTT
+encryption level (EndOfEarlyData) is available.  Though the content of CRYPTO
+frames at the Handshake encryption level cannot be forwarded to TLS before
+EndOfEarlyData is processed, the client could send ACK frames that the server
+needs to process in order to detect lost Handshake packets.
 
 
 ### TLS Interface Summary


draft-ietf-quic-tls.md
```diff
@@ -440,23 +440,31 @@ Important:
...
+frame at the Handshake encryption level) an endpoint can send STREAM data (in
+1-RTT encryption). If the Finished message is lost, the endpoint uses the
+Handshake encryption level to retransmit the lost message.
+
+In particular, server implementations need to be able to read packets at the
+Handshake encryption level before the final TLS handshake message at the 0-RTT
+encryption level (EndOfEarlyData) is available.  Though the content of CRYPTO
+frames at the Handshake encryption level cannot be forwarded to TLS before
+EndOfEarlyData is processed, the client could send ACK frames that the server
+needs to process in order to detect lost Handshake packets.
```
mikkelfj: I am not sufficiently familiar with the handshake process to really have an opinion here. But while I understand most of the text, the last paragraph seems very dense,

huitema: Looks good. You may or may not want to hammer the point about ordering. Something like this:

Proper operation of TLS requires that handshake octets be delivered to the TLS receiver in the order in which they were produced by the sender but QUIC can only guarantee ordering within an encryption level. Due to possible reordering and losses it is possible that encryption level 1 messages sent in 0-RTT packets arrive after level 2 messages sent in Handshake packets. QUIC receivers need to make sure that handshake messages are only submitted to TLS when TLS is ready to accept messages at that encryption level.

martinthomson: Encryption levels are strictly ordered, so isn't it possible that QUIC can guarantee ordering in what it delivers to TLS?

The point you are making is well taken though; packets might arrive out of order and QUIC is responsible for ensuring that they are delivered to TLS in the correct order.

Take a look at the text I've added.  I think that hammers the point home hard enough.

huitema: There is no requirement that QUIC guarantees ordering. In fact, it will deliver data from different streams in different order than they were received -- this is essential to avoid head of line blocking. The only guaranteed ordering is within a stream, or within a crypto stream.

Your paragraph develops the need for QUIC to handle the different Crypto contexts in parallel. That's fine, but in my mind that's not the same requirement as delivery of crypto-stream in specified order. In fact, just ordering them is not sufficient, because the QUIC implementation does not actually know how long a particular crypto stream is. There is no end mark. So how does it logically deduce that level N is now done, and that it is OK to proceed with N+1?

huitema: I may be suffering from excess of caution, after spending some time debugging the out of order EOED issue. But yes, the explicit mention of out of order EOED and need to reorder is there, so I guess I the text is fine.

nibanks: One of the reasons I opened #1592 was to make it easier to know when one part of the CRYPTO stream ended and another started. It also ordering easier and more explicit, IMO. But everyone seemed to disagree at the time.

kazuho: @nibanks FWIW, I still prefer the current approach above #1592.

By having different crypto streams at every epoch, TLS stacks can check if the TLS messages were delivered using the correct epoch. The verification is crucial for TLS.

If we allow single "stream" (i.e. flow of octets) span across multiple epochs, it becomes much harder to have that kind of enforcement in the API design, and I would assume that at least some of the QUIC implementations will fail to implement correct verification (consider the fact that most of us failed to notice NSTs sent using a cleartext packet).

draft-ietf-quic-tls.md
```diff
@@ -457,7 +457,11 @@ Although TLS only uses one encryption level at a time, QUIC may use more than
 one level. For instance, after sending its Finished message (using a CRYPTO
 frame at the Handshake encryption level) an endpoint can send STREAM data (in
 1-RTT encryption). If the Finished message is lost, the endpoint uses the
-Handshake encryption level to retransmit the lost message.
+Handshake encryption level to retransmit the lost message.  Reordering or loss
+of packets can mean that QUIC will need to handle packets at multiple encryption
+levels.  During the handshake, this means potentially handling of packets at
```
MikeBishop: "potentially handling packets" or "handling of packets", but not both.

draft-ietf-quic-tls.md
```diff
@@ -440,23 +440,35 @@ Important:
...
-At each change of encryption level in either direction, TLS provides QUIC with
-the new level and the encryption keys.  These events are not asynchronous; they
-always occur immediately after TLS is provided with new handshake octets, or
-after TLS produces handshake octets.
+As keys for new encryption levels become available, TLS provides QUIC with those
+keys.  Separately, as TLS starts using keys at a given encryption level, TLS
+indicates to QUIC that it is now reading or writing with keys at that encryption
+level.  These events are not asynchronous; they always occur immediately after
+TLS is provided with new handshake octets, or after TLS produces handshake
+octets.
```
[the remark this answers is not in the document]
kazuho: That's not always the case. People sometimes offload public key crypto to different devices (even to different continents when private key leakage is a concern).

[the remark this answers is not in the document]
mikkelfj: @kazuho that may be so, but isn't TLS / QUIC currently stating that all interaction is synchronous?

PR #1656, opened 2018-08-13
Minor editorial fixes.

bencebeky:
"indexes" occurs once while "indices" occurs six times, reconcile.  Small grammar fixes.  Position vertical arrow to center of block.  Update figure captions (Post-Base Index only applies to Request Streams).  Capitalize "Largest Reference" as it is done dozens of places elsewhere.  Fix obsolete reference to "frames" on control streams.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -104,14 +104,14 @@ with substantially less head-of-line blocking under the same loss conditions.
 
 # Header Tables
 
-Like HPACK, QPACK uses two tables for associating header fields to indexes.  The
+Like HPACK, QPACK uses two tables for associating header fields to indices.  The
 static table (see {{table-static}}) is predefined and contains common header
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -261,7 +261,7 @@ immediately. A stream becomes unblocked when the greatest absolute index in the
 dynamic table becomes greater than or equal to the Largest Reference for all
 header blocks the decoder has started reading from the stream.  If a decoder
 encounters a header block where the actual largest reference is not equal to the
-largest reference declared in the prefix, it MAY treat this as a stream error of
+Largest Reference declared in the prefix, it MAY treat this as a stream error of
```
MikeBishop: I could go either way on this one....  I tend to view it similar to when `-transport` capitalizes "[Cc]onnection ID" -- "Largest Reference" is the field itself, while "the largest reference" means the greatest index referenced by the frame.

PR #1657, opened 2018-08-13
QPACK: wrap absolute index values

### LINKED ISSUE #1644: QPACK improvement: wrap absolute index values

dtikhonov:
Resolves #1644.

Includes #1650 and #1642.

Closes #1763.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -130,12 +130,15 @@ instructions on the encoder stream (see {{encoder-stream}}).
 The maximum size of the dynamic table can be modified by the encoder, subject to
 a decoder-controlled limit (see {{configuration}} and {{size-update}}).  The
 initial maximum size is determined by the corresponding setting when HTTP
-requests or responses are first permitted to be sent. For clients using 0-RTT
-data in HTTP/QUIC, this means that the table size is the remembered value of the
-setting, even if the server later specifies a larger maximum in its SETTINGS
-frame.  For HTTP/QUIC servers and HTTP/QUIC clients when 0-RTT is not attempted
-or is rejected, this means that the initial maximum table size is the value of
-the setting in the peer's SETTINGS frame.
+requests or responses are first permitted to be sent.  For HTTP/QUIC
+servers and HTTP/QUIC clients when 0-RTT is not attempted or is rejected,
+this means that the initial maximum table size is the value of the setting
+in the peer's SETTINGS frame.  For clients using 0-RTT data in HTTP/QUIC,
+this means that the table size is the remembered value of the setting,
+even if the server later specifies a larger maximum in its SETTINGS frame.
+If the server does specify a larger maximum in response to a successful 0-RTT
+request, the client encoder MUST recalculate its "MaxEntries" (see
+{{absolute-indexing}}).
 
 Before a new entry is added to the dynamic table, entries are evicted from the
 end of the dynamic table until the size of the dynamic table is less than or
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -129,13 +129,13 @@ instructions on the encoder stream (see {{encoder-stream}}).
 
 The maximum size of the dynamic table can be modified by the encoder, subject to
 a decoder-controlled limit (see {{configuration}} and {{size-update}}).  The
-initial maximum size is the applicable value of this limit when HTTP requests or
-responses are first permitted to be sent. For clients using 0-RTT data in
-HTTP/QUIC, this means that the table size is the remembered value of
-`SETTINGS_HEADER_TABLE_SIZE`, even if the server later specifies a larger
-maximum in its SETTINGS frame.  For HTTP/QUIC servers and HTTP/QUIC clients not
-using 0-RTT data, this means that the initial maximum table size is the value of
-`SETTINGS_HEADER_TABLE_SIZE` in the peer's SETTINGS frame.
+initial maximum size is determined by the corresponding setting when HTTP
+requests or responses are first permitted to be sent. For clients using 0-RTT
+data in HTTP/QUIC, this means that the table size is the remembered value of the
+setting, even if the server later specifies a larger maximum in its SETTINGS
+frame.  For HTTP/QUIC servers and HTTP/QUIC clients when 0-RTT is not attempted
+or is rejected, this means that the initial maximum table size is the value of
+the setting in the peer's SETTINGS frame.
 
 Before a new entry is added to the dynamic table, entries are evicted from the
 end of the dynamic table until the size of the dynamic table is less than or
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -166,7 +166,7 @@ setting a maximum size of 0, which can subsequently be restored.
 
 Each entry possesses both an absolute index which is fixed for the lifetime of
 that entry and a relative index which changes over time based on the context of
-the reference. The first entry inserted has an absolute index of "1"; indices
+the reference. The first entry inserted has an absolute index of "0"; indices
 increase sequentially with each insertion.
 
 The relative index begins at zero and increases in the opposite direction from
@@ -255,8 +255,8 @@ Each header block contains a Largest Reference which identifies the table state
 necessary for decoding. If the greatest absolute index in the dynamic table is
 less than the value of the Largest Reference, the stream is considered
 "blocked."  While blocked, header field data should remain in the blocked
-stream's flow control window.  When the Largest Reference is zero, the frame
-contains no references to the dynamic table and can always be processed
+stream's flow control window.  When the Largest Reference is not specified, the
+frame contains no references to the dynamic table and can always be processed
 immediately. A stream becomes unblocked when the greatest absolute index in the
 dynamic table becomes greater than or equal to the Largest Reference for all
 header blocks the decoder has started reading from the stream.  If a decoder
@@ -573,7 +573,7 @@ before using it.
 
 ### Header Acknowledgement
 
-After processing a header block whose declared Largest Reference is not zero,
+After processing a header block whose declared Largest Reference is specified,
 the decoder emits a Header Acknowledgement instruction on the decoder stream.
 The instruction begins with the '1' one-bit pattern and includes the request
 stream's stream ID, encoded as a 7-bit prefix integer.  It is used by the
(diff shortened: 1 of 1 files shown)


martinthomson:
Or "Table State" to match the "Table State Synchronize" instruction.

draft-ietf-quic-qpack.md
```diff
@@ -162,12 +177,45 @@ the dynamic table is less than or equal to the maximum size.
...
 Each entry possesses both an absolute index which is fixed for the lifetime of
 that entry and a relative index which changes over time based on the context of
-the reference. The first entry inserted has an absolute index of "1"; indices
-increase sequentially with each insertion.
+the reference.
+
+The absolute index takes on the values in the range
+
+~~~
+   [ 0, MaxEntries * 2 - 1 ]
```
martinthomson: Inclusive, I presume.

draft-ietf-quic-qpack.md
```diff
@@ -251,14 +299,33 @@ Because QUIC does not guarantee order between data on different streams, a
...
-less than the value of the Largest Reference, the stream is considered
-"blocked."  While blocked, header field data should remain in the blocked
-stream's flow control window.  When the Largest Reference is zero, the frame
-contains no references to the dynamic table and can always be processed
-immediately. A stream becomes unblocked when the greatest absolute index in the
-dynamic table becomes greater than or equal to the Largest Reference for all
+Each header block contains a Largest Reference field which identifies the
+table state necessary for decoding.
+
+When the Largest Reference value is not specified, the frame contains no
```
martinthomson: You need to explain what "not specified" means.  Here, that's an encoded value of 0.

draft-ietf-quic-qpack.md
```diff
@@ -251,14 +299,33 @@ Because QUIC does not guarantee order between data on different streams, a
...
+references to the dynamic table and can always be processed immediately.
+
+When it is specified, the Largest Reference value is compared with
+the absolute index value of the next entry to be inserted into the
+dynamic table.  That entry's absolute index value divides the absolute
```
martinthomson: This explanation, relying on the "next entry to be inserted" is awkward.  As I suggested, if you count the total number of inserts, you get something easier to manage:

```
0 -----
  First: Insert (i.e., Index 8)
1 -----
...
  Index: 1
8 -----
  Index: 0
9 -----  << Base offset = 2 (or -2, I forget which this is).
  Post-Base: 0
10 -----
  Post-Base: 1
11 -----  << Table state = 11
  This: Can't be referenced
12 -----
```

Then you can say that the signaled table state and the decoder's view of the table state might differ, but only by less than `MaxEntries`.  That means that if `SignaledTableState%(MaxEntries*2)` differs from `LocalTableState` by more than MaxEntries, then it needs to be wrapped to get the full table state.

Don't worry about describing these in terms of absolute values.  I think that's easier than having the variables be modulo `2MaxEntries`.

draft-ietf-quic-qpack.md
```diff
@@ -312,6 +379,11 @@ for example because the encoder or the decoder have chosen not to risk blocked
 streams, the decoder sends a Table State Synchronize instruction (see
 {{table-state-synchronize}}).
 
+### Insertion Throttling
+
+The encoder MUST NOT create -- either via insertion or duplication -- more
+than "MaxEntries" unacknowledged dynamic table entries.
```
martinthomson: This text prevents an encoder from sending the insertions, but it doesn't prevent it from attempting to reference them.  I would start by saying that an encoder *cannot* reference dynamic table entries that have an absolute index more than MaxEntries greater than the largest acknowledged value.

I would prefer to lean more strongly on different advice.  If we say that an encoder MUST NOT reference a table state that it has not sent at least once we get a stronger guarantee.  The text here is only correct in that it prevents the table from extending more than MaxEntries into the future, but it doesn't also cover the point regarding references to those entries.

I'd also like to get more context here.  Why?

dtikhonov: The reference restriction is specified in another place in the document:

> An encoder MUST ensure that a header block which references a dynamic table
entry is not received by the decoder after the referenced entry has already been
evicted.

This guarantees that the encoder won't reference an entry that's more than `MaxEntries` away.

A header block cannot reference entries more than `MaxEntries - 1` slots apart.  This holds both for past and future entries.  Why do we need more restrictions?

The advice that
> an encoder MUST NOT reference a table state that it has not sent at least once
seems like a stronger version of the warning about deadlocks (which does not exist in the current draft).

Point taken regarding the context; I'll add it.

dtikhonov:
> Or "Table State" to match the "Table State Synchronize" instruction.

`TSS` transmits deltas, whereas `Largest Reference` specifies an absolute position in the table.  How could they match?

draft-ietf-quic-qpack.md
dtikhonov: Yes, inclusive.  Is something more that using square brackets required?

draft-ietf-quic-qpack.md
dtikhonov: Will a forward reference do?  I'd like to avoid discussing the actual encoding of the _Largest Reference_ in two places.

draft-ietf-quic-qpack.md
dtikhonov: I used the "next entry to be inserted" because I see no other way to describe the initial state, when there is no "last inserted entry."  Perhaps we could add the concept of "cursor?"

The "total number of inserts" concept eludes me.  The diagram is confusing, as it combines both encoder instructions (`Insert`) and header block instructions (`Post-Base`).

1. If the "total number of inserts" is the number of inserts performed while encoding a given header block, how does the decoder know where to start?
1. If this number is added to the current absolute index, then it is the same as _Largest Reference_.
1. ?

martinthomson:
> TSS transmits deltas, whereas Largest Reference specifies an absolute position in the table. How could they match?

Table State Synchronize transmits deltas to the table state?

dtikhonov:
Yes, _the [TSS] instruction specifies the total number of dynamic table inserts and duplications **since the last** Table State Synchronize or Header Acknowledgement that increased the Largest Known Received dynamic table entry._  (Emphasis added).

The _Largest Reference_, on the other hand, communicates the exact state (or position) in the dynamic table.  Counting all the inserts will work, of course (modulo `MaxEntries * 2`), but it will defeat the purpose of the proposed change, as the encoded representation will now grow.

PR #1663, opened 2018-08-15
Limit delayed ack timer to 1 RTT

### LINKED ISSUE #1662: Restrict delayed ACKs
Currently the transport drafts says:
"Packets containing non-ACK frames MUST be acknowledged immediately or when a delayed ack timer expires."

However, I think the transport draft should set actually boundaries here and not rely on the delayed ack timer defined in another doc. RFC1122 for TCP e.g. says
"A TCP SHOULD implement a delayed ACK, but an ACK should not
be excessively delayed; in particular, the delay MUST be
less than 0.5 seconds, and in a stream of full-sized
segments there SHOULD be an ACK for at least every second
segment."

I guess we can now a days adapt the values a bit but should probably define similar boundaries in the transport draft.

ianswett:
Partly addresses #1662

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3409,8 +3409,8 @@ acknowledge packets containing only ACK frames when sending ACK frames in
 response to other packets.  Implementations MUST NOT send more than one packet
 containing only an ACK frame per received packet that contains frames other than
 an ACK frame.  Packets containing non-ACK frames MUST be acknowledged
-immediately or when a delayed ack timer expires. The delayed ack timer should
-never delay an ACK for longer than an RTT, which ensures an ACK frame is sent
+immediately or when a delayed ack timer expires. The delayed ack timer MUST
+NOT delay an ACK for longer than an RTT, which ensures an ACK frame is sent
 at least once per RTT if new packets needing acknowledgement were received.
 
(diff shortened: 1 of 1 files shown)


mirjak:
I guess it could still makes sense to also add a reference to the recovery draft there.

draft-ietf-quic-transport.md
```diff
@@ -3407,9 +3407,11 @@ Implementations MUST NOT generate packets that only contain ACK frames in
 response to packets which only contain ACK frames. However, they MUST
 acknowledge packets containing only ACK frames when sending ACK frames in
 response to other packets.  Implementations MUST NOT send more than one packet
-containing only ACK frames per received packet that contains frames other than
-ACK frames.  Packets containing non-ACK frames MUST be acknowledged immediately
-or when a delayed ack timer expires.
+containing only an ACK frame per received packet that contains frames other than
+an ACK frame.  Packets containing non-ACK frames MUST be acknowledged
+immediately or when a delayed ack timer expires. The delayed ack timer should
+never delay an ACK for longer than an RTT, which ensures an ACK frame is sent
```
martinthomson: "should never" or "MUST NOT"?

MikeBishop: Or "does not," if you're making an assertion about the outcome of this strategy.

PR #1665, opened 2018-08-15
Full (mostly) editorial pass

mirjak:
This PR includes a bunch of small changes after a full review of the doc. Some are nits, many are mostly editorial or clarifications, but a few changes might be more than that. Please provide feedback!

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -138,16 +138,14 @@ QUIC protocol for connection establishment, stream multiplexing, stream and
 connection-level flow control, connection migration, and data reliability.
 
 An accompanying document describes QUIC's loss detection and congestion control
-{{QUIC-RECOVERY}}. As these are sender-side only mechanisms they can be changed
-without negotiating a new QUIC version. This document specifies version
-0x00000001 of QUIC which uses TLS1.3 for key negotiation as described in
-{{QUIC-TLS}}. Key negotiation and encryption is described in a separate
-docuement to make changes to only this part easier in future versions.
+{{QUIC-RECOVERY}}. Each implementation of QUIC will unilaterally select a loss
+detection and congestion control strategy. A recommended approach is described
+in {{QUIC-RECOVERY}}. This document specifies version 0x00000001 of QUIC which
+uses TLS 1.3 for key negotiation as described in {{QUIC-TLS}}. Key negotiation
+and encryption is described in a separate docuement to make changes to only
+this part easier in future versions.
 
-All versions of QUIC MUST conform to the protocol invariants in
-{{QUIC-INVARIANTS}}.
-
-QUIC is a name, not an acronym.
+All versions of QUIC conform to the protocol invariants in {{QUIC-INVARIANTS}}.
 
 # Conventions and Definitions
 
@@ -190,14 +188,15 @@ QUIC packet:
 
 0-RTT:
 
-: QUIC packets containing application payload data that are encrypted with
-  a key derived from previous QUIC connection to the same endpoint.
+: QUIC packets encrypted with a key derived from previous QUIC connection
+  to the same endpoint.
 
 1-RTT:
 
 : QUIC packets that are encrypted with keys derived from the QUIC
   handshake.
 
+QUIC is a name, not an acronym.
 
 ## Notational Conventions
 
@@ -394,8 +393,8 @@ covers the both the Packet Number and Payload fields, both of which are
 confidentiality protected and initially of unknown length.  The size of the
 Payload field is learned once the packet number protection is removed.
 
-The length field is used to enable senders to coalesce multiple long-header
-packets and potentially one short-header packet into one UDP datagram.  See
+The length field enables senders to coalesce multiple long-header packets
+and potentially one short-header packet into one UDP datagram.  See
 {{packet-coalesce}} for more details.
 
 
@@ -481,8 +480,8 @@ Protected Payload:
 
 : Packets with a short header always include a 1-RTT protected payload.
 
-The header form and connection ID field, if present, of a short header packet
-areversion-independent.  The remaining fields are specific to the selected
+The header form and connection ID field of a short header packet
+are version-independent.  The remaining fields are specific to the selected
 QUIC version.  See {{QUIC-INVARIANTS}} for details on how packets from
 different versions of QUIC are interpreted.
 
@@ -869,8 +868,8 @@ endpoint; the Source Connection ID includes the connection ID that the
 sender of the packet wishes to use (see {{connection-id-encoding}}).
 
 The first Handshake packet sent by a server contains a packet number of 0.
-Handshake packets are their own packet number space. See section
-{{packet-numbers}} on packet number spaces. Packet numbers are incremented
+Handshake packets are their own packet number space. (See {{packet-numbers}}
+for more information on packet number spaces.) Packet numbers are incremented
 normally for other Handshake packets.
 
 Servers MUST NOT send more than three datagrams including Initial and Handshake
@@ -891,8 +890,8 @@ static handshake keys or the 0-RTT keys are sent with long headers; all packets
 protected with 1-RTT keys are sent with short headers.  The different packet
 types explicitly indicate the encryption level and therefore the keys that are
 used to remove packet protection.  0-RTT and 1-RTT protected packets share a
-single packet number space. Initial and Handshake packets both have their own
-packet number space.
+single packet number space, while Initial and Handshake packets each have their
+own packet number space.
 
 Packets protected with handshake keys only use packet protection to ensure that
 the sender of the packet is on the network path.  This packet protection is not
@@ -908,7 +907,7 @@ Packets protected with 0-RTT keys use a type value of 0x7C.  The connection ID
 fields for a 0-RTT packet MUST match the values used in the Initial packet
 ({{packet-initial}}).
 
-The client can send further 0-RTT packets after receiving an Initial
+The client can send additional 0-RTT packets after receiving an Initial
 ({{packet-initial}}) or Handshake ({{packet-handshake}}) packet, if that
 packet does not complete the handshake.  Even if the client receives a
 different connection ID in the Handshake packet, it MUST continue to
@@ -923,8 +922,8 @@ confidentiality protection that is applied after packet protection is applied
 each packet sent, see {{packet-numbers}} for details.
 
 The content of 0-RTT and 1-RTT packets is protected using authenticated
-encryption. {{QUIC-TLS}}vdescribes packet protection in detail.  After
-decryption, the plaintext consistsvof a sequence of frames, as described in
+encryption. {{QUIC-TLS}} describes packet protection in detail.  After
+decryption, the plaintext consists of a sequence of frames, as described in
 {{frames}}.
 
 
@@ -934,7 +933,7 @@ A sender can coalesce multiple QUIC packets (typically a Cryptographic Handshake
 packet and a Protected packet) into one UDP datagram.  This can reduce the
 number of UDP datagrams needed to send application data during the handshake and
 immediately afterwards. It is not necessary for senders to coalesce
-packets, though not doing it leads larger number of datagrams during the
+packets, though not doing so leads to a larger number of datagrams during the
 handshake. Receivers MUST be able to process coalesced packets.
 
 Senders SHOULD coalesce packets in order of increasing encryption levels
@@ -1132,10 +1131,10 @@ Stateless Reset do not contain frames.
 Protected payloads MUST contain at least one frame, and MAY contain multiple
 frames and multiple frame types.
 
-Frames MUST fit within a single QUIC packet, and therefore within a single
-UDP datagram, and MUST NOT span a QUIC packet
-boundary. Each frame begins with a Frame Type, indicating its type, followed by
-additional type-dependent fields:
+QUIC packets cannot span multiple datagrams. Therefore frames must fit within
+a single QUIC packet and cannot span a QUIC packet boundary. Each frame begins
+with a Frame Type, indicating its type, followed by additional type-dependent
+fields:
 
 ~~~
  0                   1                   2                   3
@@ -1204,8 +1203,8 @@ An extension to QUIC that wishes to use a new type of frame MUST first ensure
 that a peer is able to understand the frame.  An endpoint can use a transport
 parameter to signal its willingness to receive one or more extension frame types
 with the one transport parameter. Otherwise it is assumed that enpoints have
-knowledge about which (extension) frame types are supported by the otherwise, e.g.
-through out of band negotiation.
+knowledge about which (extension) frame types are supported by the otherwise,
+e.g. through out of band negotiation.
 
 An IANA registry is used to manage the assignment of frame types, see
 {{iana-frames}}.
@@ -1218,8 +1217,8 @@ connection establishment intertwines version negotiation with the cryptographic
 and transport handshakes to reduce connection establishment latency, as
 described in {{handshake}}.  Once established, a connection may migrate to a
 different IP address and/or port number at either endpoint, due to NAT rebinding
-or mobility, as described in {{migration}}.  Finally a connection may be terminated
-by either endpoint, as described in {{termination}}.
+or mobility, as described in {{migration}}.  Finally a connection may be
+terminated by either endpoint, as described in {{termination}}.
 
 ## Connection ID
 
@@ -1229,20 +1228,20 @@ in length, or between 4 and 18 octets (inclusive).  Connection IDs are selected
 independently in each direction.
 
 The primary function of a connection ID is to ensure packets can still be
-assotiated with the right connection at the receiver if changes in sender
-addressing at lower protocol layers (UDP, IP, and below) occur. The Connection ID
-is visibe to the path to also enable the network to deliver packets for a QUIC
-connection to the right endpoint after an address change if load balancing is used.
-Each endpoint selects connection IDs using an implementation-specific (and perhaps
-deployment-specific) method which will allow packets with that connection ID to
-be routed back to the endpoint and identified by the endpoint upon receipt.
+associated with the correct connection at the receiver if changes in sender
+addressing at lower protocol layers (UDP, IP, and below) occur. The Connection
+ID is visibe to the path to enable load balancers to deliver packets for a
+QUIC connection to the correct endpoint despite address changes. Each endpoint
+selects connection IDs using an implementation-specific (and perhaps
+deployment-specific) method which will allow packets with that connection ID
+to be routed back to the endpoint and identified by the endpoint upon receipt.
 
 A zero-length connection ID MAY be used when the connection ID is not needed for
-routing and the address/port tuple of packets is assumed to be sufficiently stable
-to associate them to a connection. An endpoint whose peer has selected a
-zero-length connection ID MUST continue to use a zero-length connection ID for the
-lifetime of the connection and cannot send packets from any other local address
-without breaking the connection.
+routing and the address/port tuple of packets is assumed to be sufficiently
+stable to associate them to a connection. An endpoint whose peer has selected a
+zero-length connection ID MUST continue to use a zero-length connection ID for
+the lifetime of the connection and cannot send packets from any other local
+address without breaking the connection.
 
 When an endpoint has requested a non-zero-length connection ID, it will issue a
 series of connection IDs over the lifetime of a connection. The series of
@@ -1328,8 +1327,9 @@ unknown version, but instead send a Version Negotiation packet, provided that
 the packet is sufficiently long.
 
 Servers MUST drop packets without sending a Version Negotiation packet when the
-received packet contains unsupported versions and is too small
-to be an Initial packet for some version supported by the server.
+received packet indicates an unsupported versions and is too small
+to be the first packet that would create a new connection for some version
+supported by the server.
 
 If the packet is an Initial packet fully conforming with this specification, the
 server proceeds with the handshake ({{handshake}}). This commits the server to
@@ -1352,9 +1352,9 @@ is present in the header.
 ## Version Negotiation
 
 Version negotiation ensures that client and server agree to a QUIC version
-that is mutually supported independent of the implemented version. A server
+that is mutually supported, independent of the implemented version. A server
 can send a Version Negotiation packet in response to each packet that might
-initiate a new connection and is sufficiently large, see also
+initiate a new connection and is sufficiently large; see 
 {{packet-handling}} for details.
 
 The size of the first packet sent by a client will determine whether a server
@@ -1599,8 +1599,8 @@ language from Section 3 of {{!TLS13=I-D.ietf-tls-tls13}}.
 
 The `extension_data` field of the quic_transport_parameters extension defined in
 {{QUIC-TLS}} contains a TransportParameters value.  TLS encoding rules are
-therefore used to encode the transport parameters. As show in
-{{figure-transport-parameters}} TransportParameters consists of either an
+therefore used to encode the transport parameters. As shown in
+{{figure-transport-parameters}}, TransportParameters consists of either an
 indication of the initial version used by the client or the negotiated version
 and supported versions provided by the server followed by a list of other
 transport parameters that contain a parameter ID and optional value prefixed
@@ -1612,7 +1612,8 @@ transport parameters declared by the peer are available.  Each endpoint
 validates the value provided by its peer.  In particular, version negotiation
 MUST be validated (see {{version-validation}}) before the connection
 establishment is considered properly complete. Each transport parameter consists
-of an 2-byte parameter ID and an optional parameter value.
+of an 2-byte parameter ID and a parameter value, prefixed by a 2-byte length
+field for the value field which might be zero.
 
 Definitions for each of the defined transport parameters are included in
 {{transport-parameter-definitions}}.  Any given parameter MUST appear
@@ -1688,7 +1689,7 @@ disable_migration (0x0009):
 : The endpoint does not support connection migration ({{migration}}). Peers MUST
   NOT send any packets, including probing packets ({{probing}}), from a local
   address other than that used to perform the handshake.  This parameter is a
-  zero-length value. Values other than 0 and 1 are invalid.
+  zero-length value.
 
 Either peer MAY advertise an initial value for the flow control on each type of
 stream on which they might receive data.  Each of the following transport
@@ -1746,7 +1747,7 @@ type TRANSPORT_PARAMETER_ERROR.
 ### Values of Transport Parameters for 0-RTT {#zerortt-parameters}
 
 A client that attempts to send 0-RTT data MUST remember the transport parameters
-provided by the server in the initial handshake.  The transport parameters that
+provided by the server in the preivous handshake.  The transport parameters that
 the server advertises during connection establishment apply to all connections
 that are resumed using the keying material established during that handshake.
 Remembered transport parameters apply to the new connection until the handshake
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -131,17 +131,23 @@ QUIC implements techniques learned from experience with TCP, SCTP and other
...
+including its signaling.  This allows the protocol to evolve by minimizing
+dependencies on upgrades to middleboxes.  This document describes the core QUIC
 protocol, including the conceptual design, wire format, and mechanisms of the
 QUIC protocol for connection establishment, stream multiplexing, stream and
 connection-level flow control, connection migration, and data reliability.
 
-Accompanying documents describe QUIC's loss detection and congestion control
-{{QUIC-RECOVERY}}, and the use of TLS 1.3 for key negotiation {{QUIC-TLS}}.
+An accompanying document describes QUIC's loss detection and congestion control
+{{QUIC-RECOVERY}}. As these are sender-side only mechanisms they can be changed
```
ianswett: I don't think the sentence "As these are sender-side..." is adding anything.  Is there something you think is missing here?

draft-ietf-quic-transport.md
```diff
@@ -182,7 +188,15 @@ QUIC packet:
 
 : A well-formed UDP payload that can be parsed by a QUIC receiver.
 
-QUIC is a name, not an acronym.
+0-RTT:
+
+: QUIC packets containing application payload data that are encrypted with
```
ianswett: Technically 0-RTT packets contain application data and some CRYPTO data, so I'd be inclined to just say "QUIC packets encrypted with a key derived from a previous QUIC connection to the same endpoint."

draft-ietf-quic-transport.md
```diff
@@ -293,8 +307,8 @@ Long headers are used for packets that are sent prior to the completion of
 version negotiation and establishment of 1-RTT keys. Once both conditions are
 met, a sender switches to sending packets using the short header
 ({{short-header}}).  The long form allows for special packets - such as the
-Version Negotiation packet - to be represented in this uniform fixed-length
-packet format. Packets that use the long header contain the following fields:
+Version Negotiation packet - to be represented in this uniform packet format.
```
ianswett: It's actually weird to use version negotiation as an example here, because there's no packet number for version negotiation.  And I think there's no length either?

draft-ietf-quic-transport.md
```diff
@@ -380,7 +394,8 @@ covers the both the Packet Number and Payload fields, both of which are
 confidentiality protected and initially of unknown length.  The size of the
 Payload field is learned once the packet number protection is removed.
 
-Senders can sometimes coalesce multiple packets into one UDP datagram.  See
+The length field is used to enable senders to coalesce multiple long-header
```
ianswett: nit: "is used to enable" -> enables

draft-ietf-quic-transport.md
```diff
@@ -466,16 +481,16 @@ Protected Payload:
 
 : Packets with a short header always include a 1-RTT protected payload.
 
-The header form and connection ID field of a short header packet are
-version-independent.  The remaining fields are specific to the selected QUIC
-version.  See {{QUIC-INVARIANTS}} for details on how packets from different
-versions of QUIC are interpreted.
+The header form and connection ID field, if present, of a short header packet
```
ianswett: A 0-byte CID is always present and the extra clause makes this read more awkwardly, so I'd prefer keeping this as is.

draft-ietf-quic-transport.md
```diff
@@ -889,23 +908,24 @@ Packets protected with 0-RTT keys use a type value of 0x7C.  The connection ID
 fields for a 0-RTT packet MUST match the values used in the Initial packet
 ({{packet-initial}}).
 
-The client can send 0-RTT packets after receiving an Initial
-{{packet-initial}} or Handshake ({{packet-handshake}}) packet, if that
+The client can send further 0-RTT packets after receiving an Initial
```
ianswett: I prefer it without further.  If you want to add a word, I'd suggest "more"

draft-ietf-quic-transport.md
```diff
@@ -889,23 +908,24 @@ Packets protected with 0-RTT keys use a type value of 0x7C.  The connection ID
...
+({{packet-initial}}) or Handshake ({{packet-handshake}}) packet, if that
 packet does not complete the handshake.  Even if the client receives a
 different connection ID in the Handshake packet, it MUST continue to
 use the same Destination Connection ID for 0-RTT packets, see
 {{connection-id-encoding}}.
 
-The version field for protected packets is the current QUIC version.
+The version field for protected packets contains the current QUIC version.
```
ianswett: I'm not sure this is correct?

draft-ietf-quic-transport.md
```diff
@@ -889,23 +908,24 @@ Packets protected with 0-RTT keys use a type value of 0x7C.  The connection ID
...
 
 The packet number field contains a packet number, which has additional
 confidentiality protection that is applied after packet protection is applied
 (see {{QUIC-TLS}} for details).  The underlying packet number increases with
 each packet sent, see {{packet-numbers}} for details.
 
-The payload is protected using authenticated encryption.  {{QUIC-TLS}} describes
-packet protection in detail.  After decryption, the plaintext consists of a
-sequence of frames, as described in {{frames}}.
+The content of 0-RTT and 1-RTT packets is protected using authenticated
```
ianswett: I believe everything but INITIAL and RETRY are authenticated, so I think this change is not an improvement.

draft-ietf-quic-transport.md
```diff
@@ -914,14 +934,13 @@ A sender can coalesce multiple QUIC packets (typically a Cryptographic Handshake
 packet and a Protected packet) into one UDP datagram.  This can reduce the
 number of UDP datagrams needed to send application data during the handshake and
 immediately afterwards. It is not necessary for senders to coalesce
-packets, though failing to do so will require sending a significantly
-larger number of datagrams during the handshake. Receivers MUST
-be able to process coalesced packets.
+packets, though not doing it leads larger number of datagrams during the
```
ianswett: "though not doing so leads to a larger number of datagrams.."

draft-ietf-quic-transport.md
```diff
@@ -1184,7 +1203,9 @@ endpoint cannot send a frame of a type that is unknown to its peer.
 An extension to QUIC that wishes to use a new type of frame MUST first ensure
 that a peer is able to understand the frame.  An endpoint can use a transport
 parameter to signal its willingness to receive one or more extension frame types
-with the one transport parameter.
+with the one transport parameter. Otherwise it is assumed that enpoints have
```
ianswett: I'm not sure we want to imply or endorse the use of out of band negotiation, so I'd stick with the existing text.

draft-ietf-quic-transport.md
```diff
@@ -4125,37 +4172,38 @@ instead, receive streams track the delivery of data to the application or
...
+Send             | | Recv  |                               |
+ MAX_STREAM_SIZE '>|       | Recv RST_STREAM               |
+                   +-------+-----------------------.       |
+                       |                           |       |
+                       | Recv STREAM + FIN         |       |
+                       v                           |       v
+                   +-------+                       |   +-------+
+                   | Size  |                       |   | Wait  |
+                   | Known + Recv RST_STREAM       |   | STOP  |
+                   +-------+---------------------->|   +-------+
```
nibanks: What is this `Wait STOP` state?

draft-ietf-quic-transport.md
nibanks: I think it is. `The version field … contains the`

draft-ietf-quic-transport.md
```diff
@@ -2560,11 +2594,16 @@ properly continue the connection.  An endpoint that wishes to communicate a
...
 
-To support this process, a token is sent by endpoints.  The token is carried in
-the NEW_CONNECTION_ID frame sent by either peer, and servers can specify the
-stateless_reset_token transport parameter during the handshake (clients cannot
-because their transport parameters don't have confidentiality protection).  This
-value is protected by encryption, so only client and server know this value.
+To support this process, a token is sent by the endpoints that can be recovered
+or regenerated by the endpoint based on the information available in each packet
+even if the connection state was lost. This provides a protect mechanism to
+identify Stateless Reset packets from an attacker.
```
MikeBishop: "protect mechanism" is missing something, but I'm not sure what.  It doesn't permit identifying SRs generated by an attacker, because they'll just look like random noise.  I'd drop this last sentence, though your expansion on our assumptions about the token might be useful.

draft-ietf-quic-transport.md
```diff
@@ -2773,7 +2812,7 @@ encoding.
 The PADDING frame (type=0x00) has no semantic value.  PADDING frames can be used
 to increase the size of a packet.  Padding can be used to increase an initial
 client packet to the minimum required size, or to provide protection against
-traffic analysis for protected packets.
+traffic analysis for protected packets. PADDING frames need to be acknowleged.
```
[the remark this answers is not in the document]
MikeBishop: In addition to what @nibanks said, the presence of PADDING frames does not cause a packet to elicit an acknowledgement.

draft-ietf-quic-transport.md
```diff
@@ -3415,7 +3454,10 @@ acknowledged again.
 Because ACK frames are not sent in response to ACK-only packets, a receiver that
 is only sending ACK frames will only receive acknowledgements for its packets
 if the sender includes them in packets with non-ACK frames.  A sender SHOULD
-bundle ACK frames with other frames when possible.
+bundle ACK frames with other frames when possible. A receiver that would only
+send ACK frames therefore can trigger an ACK from the sender by sending a
+frames that needs to be acknowledged. Knowing which ACKs have been received
```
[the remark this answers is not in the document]
MikeBishop: And "a frames".

draft-ietf-quic-transport.md
```diff
@@ -3700,10 +3742,10 @@ FIN bit.
 A sender bundles one or more frames in a QUIC packet (see {{frames}}).
 
 A sender SHOULD minimize per-packet bandwidth and computational costs by
-bundling as many frames as possible within a QUIC packet.  A sender MAY wait for
-a short period of time to bundle multiple frames before sending a packet that is
+bundling multiple frames within a QUIC packet.  A sender MAY wait for
+a short period of time for additional frames before sending a packet that is
 not maximally packed, to avoid sending out large numbers of small packets.  An
-implementation may use knowledge about application sending behavior or
+implementation MAY use knowledge about application sending behavior or
```
MikeBishop: I don't know that this needs to be normative.  If you want to avoid the lower-case may, perhaps "can"?

draft-ietf-quic-transport.md
```diff
@@ -3713,11 +3755,11 @@ latency.
 ## Packet Processing and Acknowledgment {#processing-and-ack}
 
 A packet MUST NOT be acknowledged until packet protection has been successfully
-removed and all frames contained in the packet have been processed.  Any stream
-state transitions triggered by the frame MUST have occurred.  For STREAM frames,
-this means the data has been enqueued in preparation to be received by the
-application protocol, but it does not require that data is delivered and
-consumed.
+removed and all frames contained in the packet have been processed, including
+any stream state transitions triggered by the frame, as described in section
```
MikeBishop: remove "section"

draft-ietf-quic-transport.md
```diff
@@ -3751,13 +3793,15 @@ been lost.  In general, information is sent again when a packet containing that
...
+* Data sent in CRYPTO frames is retransmitted if detected as lost, e.g.
+  according to the rules in section 4.3.1 of {{QUIC-RECOVERY}}, until either all
+  data has been acknowledged or the crypto state machine implicitly knows that the
+  peer received the data.
 
-* Application data sent in STREAM frames is retransmitted in new STREAM frames
-  unless the endpoint has sent a RST_STREAM for that stream.  Once an endpoint
-  sends a RST_STREAM frame, no further STREAM frames are needed.
+* Application data sent in STREAM frames is retransmitted if detected as lost,
+  e.g. according to the rules in {{QUIC-RECOVERY}}, in new STREAM frames unless
```
MikeBishop: Having "in new STREAM frames" after the comma reads oddly.  Perhaps this needs to become its own sentence?  "New STREAM frames containing the lost data are generated, unless the endpoint has sent a RST_STREAM for that stream."

draft-ietf-quic-transport.md
```diff
@@ -3808,7 +3852,8 @@ containing that information is acknowledged.
 * A liveness or path validation check using PATH_CHALLENGE frames is sent
   periodically until a matching PATH_RESPONSE frame is received or until there
   is no remaining need for liveness or path validation checking. PATH_CHALLENGE
-  frames include a different payload each time they are sent.
+  frames include a different payload each time they are sent and are therefore
+  not-retransmittable.
```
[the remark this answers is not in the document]
MikeBishop: No, you can't.  You can try a fresh challenge, but it's not supposed to be the same payload.

MikeBishop: I'm not sure this addition is useful.

draft-ietf-quic-transport.md
```diff
@@ -4037,36 +4082,38 @@ Note:
...
+Send STREAM /         .-|       |                       |
+     STREAM_BLOCKED / | | Send  | Send RST_STREAM       |
+Recv MAX_STREAM_DATA  '>|       |---------------------->|
+                        +-------+                       |
+                            |                           |
+                            | Send STREAM + FIN /       |
+                            | Recv STOP_SENDING         |
+                            v                           v
+                      .-+-------+                   +-------+
+    Recv STOP_SEDNING | | Data  | Send RST_STREAM   | Reset |
```
MikeBishop: SENDING

draft-ietf-quic-transport.md
```diff
@@ -4037,36 +4082,38 @@ Note:
...
+                      '>| Sent  +------------------>| Sent  |
+                        +-------+                   +-------+
+                            |                           |
+                            | Recv All ACKs             | Recv ACK
+                            v                           v
+                        +-------+                   +-------+
+                        | Data  |                   | Reset |
+                        | Recvd |                   | Recvd |
+                        +-------+                   +-------+
```
MikeBishop: I don't think that the self-transitions are needed.  Those events simply don't cause changes to the current stream state, just like many other events that aren't shown.

draft-ietf-quic-transport.md
```diff
@@ -4078,7 +4125,7 @@ in preparation for sending.
 
 The sending part of a bidirectional stream initiated by a peer (type 0 for a
 server, type 1 for a client) enters the "Ready" state if the receiving part
-enters the "Recv" state.
+enters the "Recv" state (see Figure {{fig-stream-recv-states}}).
```
MikeBishop: ~Figure~

draft-ietf-quic-transport.md
```diff
@@ -4125,37 +4172,38 @@ instead, receive streams track the delivery of data to the application or
...
+                       o
+                       | Recv STREAM / STREAM_BLOCKED / RST_STREAM
+                       | Create Bidirectional Stream (Sending)
+                       | Recv MAX_STREAM_DATA
+                       | Create Higher-Numbered Stream
+                       v
+                   +-------+ Send STOP_SENDING
+                 .-|       |-------------------------------.
+Send             | | Recv  |                               |
+ MAX_STREAM_SIZE '>|       | Recv RST_STREAM               |
```
MikeBishop: Same comment about self-transitions.

draft-ietf-quic-transport.md
MikeBishop: Presumably the state where incoming data is being discarded, but the solicited RST_STREAM hasn't been received yet.  We originally said that STOP_SENDING didn't change the state, just hinted to the other side that it should do so.

I don't object to making this a defined state, but if so, then the description needs to be added below as well.  Let's make this a separate PR.

draft-ietf-quic-transport.md
```diff
@@ -4400,9 +4448,9 @@ STREAM frames ensures that loss recovery, congestion control, and flow control
 operate effectively.
 
 CRYPTO frames SHOULD be prioritized over other streams prior to the completion
-of the cryptographic handshake.  This includes the retransmission of the second
-flight of client handshake messages, that is, the TLS Finished and any client
-authentication messages.
+of the cryptographic handshake.  In {{QUIC-TLS}} this includes the retransmission
+of the second flight of client handshake messages, that is the TLS Finished and
+any client authentication messages.
```
MikeBishop: Addition of QUIC-TLS reference seems unnecessary, and the removal of the comma is incorrect.

draft-ietf-quic-transport.md
```diff
@@ -4450,7 +4498,8 @@ with the Stream ID set appropriately. A receiver could use the current offset of
 data consumed to determine the flow control offset to be advertised.  A receiver
 MAY send MAX_STREAM_DATA frames in multiple packets in order to make sure that
 the sender receives an update before running out of flow control credit, even if
-one of the packets is lost.
+one of the packets is lost. As soons as the packet with the MAX_STREAM_DATA
+frame is acknowlegded, there is no need to send the same max stream limit agian.
```
MikeBishop: Was this not obvious before?

draft-ietf-quic-transport.md
```diff
@@ -4486,7 +4535,7 @@ stream before the RST_STREAM frame, and the receiver MUST use the final offset
 to account for all bytes sent on the stream in its connection level flow
 controller.
 
-### Response to a RST_STREAM
+## Response to a RST_STREAM
```
MikeBishop: These are subsections of "Other Considerations."  If you're going to make them stand-alone sections, I'd move them before that section.  However, I'm more inclined to leave them as-is pending @janaiyengar and @martinthomson's general rearranging of the draft.

draft-ietf-quic-transport.md
```diff
@@ -4495,7 +4544,7 @@ RST_STREAM an endpoint will choose to stop sending data in its own direction. If
 the sender of a RST_STREAM wishes to explicitly state that no future data will
 be processed, that endpoint MAY send a STOP_SENDING frame at the same time.
 
-### Data Limit Increments {#fc-credit}
+## Data Limit Increments {#fc-credit}
```
MikeBishop: Same.

draft-ietf-quic-transport.md
```diff
@@ -4524,7 +4573,7 @@ peer-initiated streams close.  A receiver MAY also advance the maximum stream ID
 based on current activity, system conditions, and other environmental factors.
 
 
-### Blocking on Flow Control {#blocking}
+## Blocking on Flow Control {#blocking}
```
MikeBishop: Same.

draft-ietf-quic-transport.md
mirjak: Yes, the point was also to provide a reason why this is in a different doc. I used Mike's wording.

draft-ietf-quic-transport.md
```diff
@@ -2290,20 +2323,20 @@ validation of the address of the spurious migration to be abandoned.
 ### Loss Detection and Congestion Control {#migration-cc}
 
 The capacity available on the new path might not be the same as the old path.
-Packets sent on the old path SHOULD NOT contribute to congestion control or RTT
-estimation for the new path.
-
-On confirming a peer's ownership of its new address, an endpoint SHOULD
+On confirming a peer's ownership of its new address, an endpoint MUST
```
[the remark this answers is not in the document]
mirjak: I think the previous text was just confusing and probably not correct.

draft-ietf-quic-transport.md
```diff
@@ -2435,14 +2468,15 @@ of three ways:
...
+close cleanly and that delayed or reordered packets are properly discarded and
+do not interfere with future connections on the same 5-tuple. These states
+SHOULD persist for three times the current Retransmission Timeout (RTO)
+interval as defined in {{QUIC-RECOVERY}}.
 
 An endpoint enters a closing period after initiating an immediate close
-({{immediate-close}}).  While closing, an endpoint MUST NOT send packets unless
-they contain a CONNECTION_CLOSE or APPLICATION_CLOSE frame (see
-{{immediate-close}} for details).
+({{immediate-close}}).  In the closing period, an endpoint MUST NOT send
```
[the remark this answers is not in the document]
mirjak: done

draft-ietf-quic-transport.md
```diff
@@ -2290,20 +2323,20 @@ validation of the address of the spurious migration to be abandoned.
...
-responding to congestion signals and reducing send rates appropriately.
+the new path unless it is reasonably sure that the actually network path did not
+change. For instance, a change in the client's port number is likely indicative
+of a rebinding in a middlebox and not a complete change in path.  This
+determination likely depends on heuristics, which could be imperfect; if the new
+path capacity is significantly reduced, ultimately this relies on the congestion
+controller responding to congestion signals and reducing send rates
+appropriately.
+
+If the congestion controller has been reset this means that packets sent on the
```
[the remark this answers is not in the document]
mirjak: done

draft-ietf-quic-transport.md
mirjak: What about: "This mechanism protects against forged Stateless Retry packets that could otherwise enable an off-path attacker to close a connection." Any better?

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
mirjak: Yes, you are right this wording is wrong in any case. However, I now realized that the doc is actually inconsitent about PADDING frames. In section 8.4.2  it say "PADDING frames generate acknowledgements". That might be a separate PR. Removed that sentence for now.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
mirjak: reworded

draft-ietf-quic-transport.md
mirjak: Was not sure here. Changed to can for unambiguity.

draft-ietf-quic-transport.md
mirjak: done

draft-ietf-quic-transport.md
mirjak: done

draft-ietf-quic-transport.md
mirjak: done

draft-ietf-quic-transport.md
mirjak: I added it because that fit with the text describing which messages can be send when.

draft-ietf-quic-transport.md
mirjak: i initially though about transition to "Data Recvd" on reception of STOP_SENDING but that would not capture that you are actually waiting for a RST_STREAM. 

The text is mostly there it only needs to name the new state if people agree.

draft-ietf-quic-transport.md
mirjak: this sentence is very TLS specfic and bundles the two doc more then needed if left as a general statement. 

Readded the comma...

draft-ietf-quic-transport.md
mirjak: the point here is that I am actually unhappy with the "in multiple packets". How much is multiple? i would assume something like 3 but others might think they are only on the safe side if they send it e.g. 20 times... I just wanted to make sure nobody gets the idea that it is most safe to just always send it. I would be more happier if you could just give a number instead of saying "in multiple packets".

draft-ietf-quic-transport.md
mirjak: Some of them were subsection, some were new section, however, i also do not understand why they were subsection. The way it was, was just really confusing and this was the easier fix.

draft-ietf-quic-transport.md
```diff
@@ -996,7 +1014,7 @@ response to connection migration ({{migration}}). NEW_CONNECTION_ID frames
 
 The packet number is an integer in the range 0 to 2^62-1. The value is used in
 determining the cryptographic nonce for packet encryption.  Each endpoint
-maintains a separate packet number for sending and receiving.
+maintains separate packet number spaces for sending and receiving.
```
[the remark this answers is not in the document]
nibanks: Maybe just keep the working from before and prepend `Per packet number space, `.

draft-ietf-quic-transport.md
MikeBishop: I think sending the current value in each packet is a reasonable strategy, albeit verbose if you have a lot of streams.  It's certainly a good, even recommended, optimization to suppress that if it hasn't changed and one of them has been ACK'd.  But I think the larger point is that it's okay to duplicate, and how much to duplicate is entirely under the implementation's control.

Also, "again."

draft-ietf-quic-transport.md
```diff
@@ -1443,15 +1463,16 @@ that meets the requirements of the cryptographic handshake protocol:
...
+address validation MUST also be sent within a single packet. This avoids having
+to reassemble a message from multiple packets.
 
-The first client packet of the cryptographic handshake protocol MUST fit within
+The first client message of the cryptographic handshake protocol MUST fit within
 a 1232 octet QUIC packet payload.  This includes overheads that reduce the space
 available to the cryptographic handshake protocol.
 
-The CRYPTO frame can be sent in different packet number spaces.  CRYPTO frames
+The CRYPTO frames can be sent in different packet number spaces.  CRYPTO frames
```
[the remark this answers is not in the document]
MikeBishop: The original text, "The CRYPTO frame" talks about the CRYPTO frame as a unique type.  "CRYPTO frames" talks about the class of frames of that type.  Either would work here.  "The CRYPTO frames" sounds like a discrete set of frames which are being partitioned into the spaces.

draft-ietf-quic-transport.md
```diff
@@ -1578,20 +1599,26 @@ language from Section 3 of {{!TLS13=I-D.ietf-tls-tls13}}.
 
 The `extension_data` field of the quic_transport_parameters extension defined in
 {{QUIC-TLS}} contains a TransportParameters value.  TLS encoding rules are
-therefore used to encode the transport parameters.
+therefore used to encode the transport parameters. As show in
+{{figure-transport-parameters}} TransportParameters consists of either an
```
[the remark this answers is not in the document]
MikeBishop: Also, "As shown"

draft-ietf-quic-transport.md
MikeBishop: That works.

draft-ietf-quic-transport.md
```diff
@@ -1363,12 +1387,8 @@ selects an acceptable protocol version from the list provided by the server.
...
 contents of the Initial packet the client sends might not change in
 response to version negotiation, a client MUST increase the packet number it
-uses on every packet it sends.  Packets MUST continue to use long headers and
-MUST include the new negotiated protocol version.
-
-The client MUST use the long header format and include its selected version on
-all packets until it has 1-RTT keys and it has received a packet from the server
-which is not a Version Negotiation packet.
+uses on every packet it sends. The client MUST use the long header format and
+include its selected version on all packets until it has 1-RTT keys.
```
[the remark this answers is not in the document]
martinthomson: @MikeBishop, your example is a highly simplified view of 0-RTT :)

draft-ietf-quic-transport.md
martinthomson: That sort of change has to be a separate PR.

draft-ietf-quic-transport.md
martinthomson: "soons" "acknowlegded"

The new text is redundant though.

martinthomson:
Closing this.  The recent reorganization makes it very hard to take any of these changes.

PR #1682, opened 2018-08-17
Pad to 38 if you want a stateless reset

### LINKED ISSUE #1681: Small packets don't trigger a stateless reset
Suppose that a server uses 4-bytes connection ID. The client wants to test whether the server is still present, and sends a "PING". The packet will include:
* First octet: 1 byte
* Dest CNX-ID: 4 bytes
* PN: 1 byte (PNE encrypted)
* Ping frame: 1 byte (encrypted)
* AEAD checksum: 16 bytes.
* Total: 23 bytes.

The server does not recognize the connection ID -- memory of that connection is lost. It will compose a stateless reset packet that include, per draft 14:
* First octet: 1 byte
* At least 20 bytes with randomized or unpredictable value
* 16 bytes Reset Secret.
* Total: at least 37 bytes.

But then the stateless reset becomes larger than the incoming packet, and draft-14 says that:  _An endpoint SHOULD NOT send a stateless reset that is significantly larger than the packet it receives.  Endpoints MUST discard packets that are too small to be valid QUIC packets.  With the set of AEAD functions defined in [QUIC-TLS], packets less than 19 octets long are never valid._

I understand that "significantly larger" is a weasel word that can give some leeway. But then, too much leeway will allow attackers to seed a game of "reset ping pong" between two servers. On the other hand, the "ping" example above is exactly what we want to address with stateless reset. So what gives?

martinthomson:
But there isn't much point in doing that if you aren't expecting a response, so ACK-only packets don't need the same treatment (unless, like @huitema, you are doing that for a measure of traffic analysis protection).

Closes #1681.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2733,8 +2733,8 @@ a small packet might result in Stateless Reset not being useful in detecting
 cases of broken connections where only very small packets are sent; such
 failures might only be detected by other means, such as timers.
 
-An endpoint that wants to improve the chances that it triggers a Stateless Reset
-when its peer loses state can pad all packets it sends to 38 octets.
+An endpoint can increase the odds that a packet will trigger a Stateless Reset
+if it cannot be processed by padding it to at least 38 octets.
 
 
 # Frame Types and Formats
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2734,8 +2734,7 @@ cases of broken connections where only very small packets are sent; such
 failures might only be detected by other means, such as timers.
 
 An endpoint that wants to improve the chances that it triggers a Stateless Reset
-when its peer loses state can pad all packets it sends - aside from those
-containing only ACK and PADDING frames - to 38 octets.
+when its peer loses state can pad all packets it sends to 38 octets.
 
 
 # Frame Types and Formats
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2733,6 +2733,10 @@ a small packet might result in Stateless Reset not being useful in detecting
 cases of broken connections where only very small packets are sent; such
 failures might only be detected by other means, such as timers.
 
+An endpoint that wants to improve the chances that it triggers a Stateless Reset
+when its peer loses state can pad all packets it sends - aside from those
+containing only ACK and PADDING frames - to 38 octets.
+
 
 # Frame Types and Formats
 


martinthomson:
The point with ACK-only is that you aren't expecting a response, so there is no point in asking for one.  Maybe it's best to say nothing on that point anyway.  I was not 100% on that bit.

draft-ietf-quic-transport.md
```diff
@@ -2733,6 +2733,9 @@ a small packet might result in Stateless Reset not being useful in detecting
 cases of broken connections where only very small packets are sent; such
 failures might only be detected by other means, such as timers.
 
+An endpoint that wants to improve the chances that it triggers a Stateless Reset
+when its peer loses state can pad all packets it sends to 38 octets.
```
ianswett: All packets seems like a lot of padding.  In practice, I'd only expect people to pad timer-initiated packets like PING for keepalive or TLP/RTO packets.

huitema: Yes. Mikkel pointed out that if you do that, all short packets are 38 bytes, except for stateless reset which are exactly 37, and therefore stand out. That's not a hugely useful property. What Ian suggests makes more sense, and in that case you want to pad to something unpredictable, for example between 38 and 256. But this is of course a game of smoke and mirrors, so you want the peer to also randomly pad the responses to ping and tlp/rto. Oh well...

draft-ietf-quic-transport.md
```diff
@@ -2733,8 +2733,8 @@ a small packet might result in Stateless Reset not being useful in detecting
 cases of broken connections where only very small packets are sent; such
 failures might only be detected by other means, such as timers.
 
-An endpoint that wants to improve the chances that it triggers a Stateless Reset
-when its peer loses state can pad all packets it sends to 38 octets.
+An endpoint can increase the odds that a packet will trigger a Stateless Reset
```
ianswett: Does increasing it above 38 not ensure it will trigger a Stateless Reset?  If so, I'd prefer "An endpoint can ensure that a packet will trigger..."

mikkelfj: I would hope that stateless reset is an optional feature and when it is on, resets can be send at senders discretion, e.g. implementing rate limiting.

martinthomson: @mikkelfj is right, this doesn't guarantee that you get a stateless reset.  The point here is that the odds of receiving one for packets below 38 octets is much lower.

martinthomson:
Good enough for me.

PR #1692, opened 2018-08-22
Remove redundant word

amirlivneh:


draft-ietf-quic-http.md
```diff
@@ -384,7 +384,7 @@ state), the proxy will set the FIN bit on its connection to the TCP server. When
 the proxy receives a packet with the FIN bit set, it will terminate the send
 stream that it sends to client. TCP connections which remain half-closed in a
 single direction are not invalid, but are often handled poorly by servers, so
-clients SHOULD NOT cause send a STREAM frame with a FIN bit for connections on
+clients SHOULD NOT send a STREAM frame with a FIN bit for connections on
```
MikeBishop: This probably needs more words, not fewer.  The HTTP layer doesn't directly send STREAM frames, which is why there's a "cause" in there.  Maybe "clients SHOULD NOT close a stream for sending for streams on which they still expect to receive data from the target of the CONNECT"?

PR #1703, opened 2018-08-27
Reconcile 'content' and 'payload'

### LINKED ISSUE #1676: "Payload" vs. "content"
@mirjak points out that payload and content are both used to refer to the body of QUIC packets.  Worth doing a pass to ensure that these are used consistently.  My first pass at definitions:

- "Payload" is the collection of frames in the encrypted body of the packet.
- "Content" is the conceptual set of information conveyed by those frames, but probably not something we reference very often in the spec.

martinthomson:
I did a complete pass through and it wasn't that bad.  As we agreed,
payload is the bytes that we protect, content is the more abstract
notion of what needs to be communicated (for which we use frames).  That
was already mostly right, apart from a few places.

Oh, and I didn't mention this, but I think that it's worth saying it.  This isn't in the definitions section.  We don't need to load that down with these little editorial conventions.  We do that and it becomes unusable.

Closes #1676.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -566,8 +566,8 @@ wishes to perform a stateless retry (see {{stateless-retry}}).
 
 A Retry packet (shown in {{retry-format}}) only uses the invariant portion of
 the long packet header {{QUIC-INVARIANTS}}; that is, the fields up to and
-including the Destination and Source Connection ID fields.  The Payload of a
-Retry packet is not protected.  Like Version Negotiation, a Retry packet
+including the Destination and Source Connection ID fields.  A Retry packet does
+not contain any protected fields.  Like Version Negotiation, a Retry packet
 contains the long header including the connection IDs, but omits the Length,
 Packet Number, and Payload fields.  These are replaced with:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1109,7 +1109,7 @@ Stateless Reset do not contain frames.
 |                          Frame N (*)                        ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ~~~
-{: #packet-frames title="Contents of Protected Payload"}
+{: #packet-frames title="QUIC Payload"}
 
 Protected payloads MUST contain at least one frame, and MAY contain multiple
```
MikeBishop: You removed the term "protected payload" from the title of the figure, but not from the following paragraph.

draft-ietf-quic-transport.md
```diff
@@ -566,8 +566,8 @@ wishes to perform a stateless retry (see {{stateless-retry}}).
 
 A Retry packet (shown in {{retry-format}}) only uses the invariant portion of
 the long packet header {{QUIC-INVARIANTS}}; that is, the fields up to and
-including the Destination and Source Connection ID fields.  The contents of the
-Retry packet are not protected.  Like Version Negotiation, a Retry packet
+including the Destination and Source Connection ID fields.  The Payload of a
```
MikeBishop: Could go either way, but I'd lean toward not capitalizing this usage, or saying "The Payload field"

martinthomson: I realized that we say that it has a Payload here, then say that it doesn't in the next sentence, so I changed this more.

PR #1704, opened 2018-08-27
Tokens are a mess

### LINKED ISSUE #1647: The server shouldn't use a token for much
The token that a server provides in Retry or NEW_TOKEN isn't authenticated.  It can be used multiple times and it isn't folded into the handshake.  This is an explicit limitation of the design.  Neither client nor server can therefore rely on its contents for much.  A server can ensure that the token is self-authenticating (by including a MAC or self encrypting it with an AEAD, for instance), but it shouldn't really use the information beyond its immediate purpose: address validation.

martinthomson:
We can't realistically authenticate all tokens with up to three Retry
packets and the NEW_TOKEN message.  So they aren't really safe for much.
Describe the limitations and advise a very narrow usage.

I really hope that this choice doesn't bite us later.  It doesn't feel
that good, but I don't know how to fix this without going back to just
one Retry (and even then a solution would be ugly).

Closes #1647.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -786,6 +786,14 @@ Note:
   the packet is discarded.  A server MAY encode tokens provided with NEW_TOKEN
   frames and Retry packets differently, and validate the latter more strictly.
 
+In a stateless design, a server can use encrypted and authenticated tokens to
+pass information to clients that the server can later recover and use to
+validate a client address.  Tokens are not integrated into the crypto graphic
+handshake and so they cannot be authenticated.  For instance, a client might be
+able to reuse a token.  To avoid attacks that exploit this property, a server
+can limit its use of tokens to only the information needed validate client
+addresses.
+
 
 ### Starting Packet Numbers
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -788,8 +788,8 @@ Note:
 
 In a stateless design, a server can use encrypted and authenticated tokens to
 pass information to clients that the server can later recover and use to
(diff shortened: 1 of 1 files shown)


marten-seemann:
I'm still not a fan of the multiple Retry mechanism, since it doesn't serve any purpose, except that it allows implementers to build an architecture that imposes unnecessary RTT penalties on clients.

If I remember correctly, we discussed this in Montreal, and decided to at least authenticate the token in a way that makes it impossible for middleboxes to perform a Retry. Do I remember this wrong?

mikkelfj:
I did raise an issue where I touched upon a checksum and @martinthomson more explicitly stated the issue of using unauthenticated tokens. This was probably before that meeting which I did not attend to. Later the issue was resolved with no action on checksum and no argument as to why that was decided. This is all from volatile memory. I think even a simple short checksum could go a long way to protect against some forms of attack.

ianswett:
I'm also still not a fan of multiple retries.  I think it's a case of us over-designing something for use cases we don't fully understand and making implementations a lot more complex.  I thought it would end up being complex, and it's ended up being much worse than I anticipated.

But I realize that's a wider issue that is out of scope of this PR, so I'll take this to the list.

draft-ietf-quic-transport.md
```diff
@@ -786,6 +786,14 @@ Note:
   the packet is discarded.  A server MAY encode tokens provided with NEW_TOKEN
   frames and Retry packets differently, and validate the latter more strictly.
 
+In a stateless design, a server can use encrypted and authenticated tokens to
+pass information to clients that the server can later recover and use to
+validate a client address.  Tokens are not integrated into the crypto graphic
```
MikeBishop: cryptographic

PR #1705, opened 2018-08-27
Cleanup Initial packet use

### LINKED ISSUE #1666: Initial packets are version-specific
Packet handling text talks about "Initial packets in any version supported by the server."  However, there's no guarantee that those versions will have a packet named Initial.  This should continue to use the language of "could potentially create a new connection."

martinthomson:
There were a few case where version negotiation talked about Initial
packets unnecessarily.  The only signal we can rely on is size.

Closes #1666.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1323,8 +1323,7 @@ from sending Handshake packets prior to receiving a server response, so servers
 SHOULD ignore any such packets.
 
 Servers MUST drop incoming packets under all other circumstances.  They SHOULD
-send a Stateless Reset ({{stateless-reset}}) if a connection ID is present in
-the header.
+send a Stateless Reset ({{stateless-reset}}) if they are able.
 
 ## Version Negotiation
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1317,14 +1317,14 @@ If a server isn't currently accepting any new connections, it SHOULD send a
...
+packets in anticipation of a late-arriving Initial Packet. Clients are forbidden
+from sending Handshake packets prior to receiving a server response, so servers
+SHOULD ignore any such packets.
 
-Servers MUST drop incoming packets under all other circumstances. They
-SHOULD send a Stateless Reset ({{stateless-reset}}) if a connection ID
-is present in the header.
+Servers MUST drop incoming packets under all other circumstances.  They SHOULD
+send a Stateless Reset ({{stateless-reset}}) if a connection ID is present in
+the header.
```
MikeBishop: I know the change here is pure rewrap, but I don't think this statement ever applies as stated.  The CID is always present, but potentially zero-length.  For short-header packets, the server can't tell what the CID length is.

draft-ietf-quic-transport.md
```diff
@@ -1334,10 +1334,10 @@ response to each packet that might initiate a new connection, see
 {{packet-handling}} for details.
 
 The size of the first packet sent by a client will determine whether a server
-sends a Version Negotiation packet. Clients that support multiple QUIC
-versions SHOULD pad their Initial packets to reflect the largest minimum
-Initial packet size of all their versions. This ensures that the server
-responds if there are any mutually supported versions.
+sends a Version Negotiation packet. Clients that support multiple QUIC versions
+SHOULD pad the first packet they send to the largest of the minimum packet size
```
MikeBishop: "largest... size" or "largest of the... sizes"

PR #1707, opened 2018-08-27
Headers and header fields

MikeBishop:
While we all use "headers" colloquially most the time, it seems that the "correct" usage that is followed by RFCs has "the header" being the thing that precedes the body and "header fields" as the individual name-value entries within the header.  This cleans those up within the HTTP document.

We also have "stream headers" in the unidirectional case, but I eliminated the term "frame header" for simplicity.  Do we actually need a name for the part of the frame layout that precedes the payload, or are references to the Length and Type fields sufficient?

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -295,9 +295,9 @@ of informational (1xx) HTTP responses (see {{!RFC7230}}, Section 3.2 and
 {{!RFC7231}}, Section 6.2).
 
 A server MAY interleave one or more PUSH_PROMISE frames (see
-{{frame-push-promise}}) with the frames of a response message indicating a
-pushed resource related to the response. These PUSH_PROMISE frames are not part
-of the response; see {{server-push}} for more details.
+{{frame-push-promise}}) with the frames of a response message. These
+PUSH_PROMISE frames are not part of the response; see {{server-push}} for more
+details.
 
 The "chunked" transfer encoding defined in Section 4.1 of {{!RFC7230}} MUST NOT
 be used.
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -585,11 +585,11 @@ a promise, the server push response is conveyed on a push stream.
 A push stream is indicated by a stream type of `0x50` (ASCII 'P'), followed by
 the Push ID of the promise that it fulfills, encoded as a variable-length
 integer. The remaining data on this stream consists of HTTP/QUIC frames, as
-defined in {{frames}}, and carries the response side of an HTTP message
-exchange as described in {{request-response}}. The request header of the
-exchange is carried by a PUSH_PROMISE frame (see {{frame-push-promise}})
-on the request stream which generated the push. Promised requests MUST
-conform to the requirements in Section 8.2 of {{!RFC7540}}.
+defined in {{frames}}, and carries the response side of an HTTP message exchange
+as described in {{request-response}}. The header of the request message is
+carried by a PUSH_PROMISE frame (see {{frame-push-promise}}) on the request
+stream which generated the push. Promised requests MUST conform to the
+requirements in Section 8.2 of {{!RFC7540}}.
 
 Only servers can push; if a server receives a client-initiated push stream,
 this MUST be treated as a stream error of type HTTP_WRONG_STREAM_DIRECTION.
@@ -983,8 +983,8 @@ Push ID:
   ({{frame-cancel-push}}), and PRIORITY frames ({{frame-priority}}).
 
 Header Block:
-: QPACK-compressed request header for the promised response.  See [QPACK] for
-  more details.
+: QPACK-compressed request header fields for the promised response.  See [QPACK]
+  for more details.
 
 A server MUST NOT use a Push ID that is larger than the client has provided in a
 MAX_PUSH_ID frame ({{frame-max-push-id}}).  A client MUST treat receipt of a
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -297,7 +297,7 @@ of informational (1xx) HTTP responses (see {{!RFC7230}}, Section 3.2 and
 PUSH_PROMISE frames (see {{frame-push-promise}}) MAY be interleaved with the
 frames of a response message indicating a pushed resource related to the
 response. These PUSH_PROMISE frames are not part of the response, but carry the
-headers of a separate HTTP request message.  See {{server-push}} for more
+headers of separate HTTP request messages.  See {{server-push}} for more
```
LPardue: I think this falls into the trap that HTTP/QUIC PUSH_PROMISE frames can be sent multiple times. Was the original text supposed to capture that design feature? The new text seems to lose that description (PUSH_PROMISE for different messages) but accommodating both might get wordy.

LPardue:
Did you miss this text, or intentionally skipped?

> In addition, prior to sending the message header block indicated above, a
 response may contain zero or more header blocks containing the message headers
 response may contain zero or more header blocks containing the message headers
 of informational (1xx) HTTP responses (see {{!RFC7230}}, Section 3.2 and
 of informational (1xx) HTTP responses (see {{!RFC7230}}, Section 3.2 and
 {{!RFC7231}}, Section 6.2).
{{!RFC7231}}, Section 6.2).

MikeBishop:
Intentionally skipped.  The (multiple) message headers of (multiple) informational responses.

LPardue:
The multiple types of "header" is logical but a little annoying when discussing things. We've been using the term "stream preamble" internally, this just helps disambiguate between different headers. IMO the annoyance comes from their functional relationship: QUIC STREAM frame headers and HTTP/QUIC frame headers are envelopes on every message. The unidirectional stream header is a function of time and carried in message body.

MikeBishop:
I don't object to using preamble or preface.

draft-ietf-quic-http.md
```diff
@@ -586,8 +586,8 @@ A push stream is indicated by a stream type of `0x50` (ASCII 'P'), followed by
 the Push ID of the promise that it fulfills, encoded as a variable-length
 integer. The remaining data on this stream consists of HTTP/QUIC frames, as
 defined in {{frames}}, and carries the response side of an HTTP message
-exchange as described in {{request-response}}. The request headers of the
-exchange are carried by a PUSH_PROMISE frame (see {{frame-push-promise}})
+exchange as described in {{request-response}}. The request header of the
```
martinthomson: "the header of the request message is carried" perhaps

draft-ietf-quic-http.md
```diff
@@ -983,7 +983,7 @@ Push ID:
   ({{frame-cancel-push}}), and PRIORITY frames ({{frame-priority}}).
 
 Header Block:
-: QPACK-compressed request headers for the promised response.  See [QPACK] for
+: QPACK-compressed request header for the promised response.  See [QPACK] for
```
martinthomson: I think that we can say header fields here.  QPACK doesn't compress the header, it only compresses a list of header fields.

draft-ietf-quic-http.md
MikeBishop: In context, I don't think it says anything about PUSH_PROMISE frames that are sent multiple times on different streams.  It's talking about a single response, and says that the PUSH_PROMISE frames carry headers of separate requests being pushed.  If the same PUSH_PROMISE shows up during another response, that's mostly independent.

Can you suggest wording you think would be clearer?

draft-ietf-quic-http.md
MikeBishop: I'd call semantics on the distinction, since the list of header fields is the header, but I think it reads better your way.

draft-ietf-quic-http.md
LPardue: My original point is that you could send several PUSH_PROMISE frames, for the same promised resource with the same PUSH_ID, on a single request stream.

Something simpler could be to take away all mention of PUSH_PROMISE headers in th:

A server MAY chose to interleave one or more PUSH_PROMISE frames (see {{frame-push-promise}}) with the response message. However, these are not part of the response.  Server push is described further in {{server-push}}.

MikeBishop: Works for me.  (But why *would* you send several PUSH_PROMISEs for the same resource on the same request stream?)

LPardue: I couldn't slip that passed you, could I? :)

I wouldn't expect conventional HTTP/QUIC servers to do this. However, the existing specification allows for this and I would hope that implementations can implement the required mechanics.

As for a reason, in the absence of a true QUIC transport-level FEC, we could imagine an application-level FEC that repeats important (and relatively cheap) metadata. Its true that today streams are reliable but some form of unreliable streams may benefit from a smarter repetition strategy. 

Independent PUSH_PROMISE frames may duplicate the same data. Request and response header messages could be FEC'd by purposefully sending duplicate STREAM frames (same offset and content) containing such metadata. 

Both forms of repetition could work together.

draft-ietf-quic-http.md
```diff
@@ -294,11 +294,10 @@ response may contain zero or more header blocks containing the message headers
...
 {{!RFC7231}}, Section 6.2).
 
-PUSH_PROMISE frames (see {{frame-push-promise}}) MAY be interleaved with the
-frames of a response message indicating a pushed resource related to the
-response. These PUSH_PROMISE frames are not part of the response, but carry the
-headers of separate HTTP request messages.  See {{server-push}} for more
-details.
+A server MAY interleave one or more PUSH_PROMISE frames (see
+{{frame-push-promise}}) with the frames of a response message indicating a
+pushed resource related to the response. These PUSH_PROMISE frames are not part
```
martinthomson: "pushed resource" isn't quite right, we push responses.  You could probably drop from "indicating" onwards and retain the meaning.

PR #1708, opened 2018-08-27
Use 'encryption' less often

martinthomson:
The term is protection in most cases.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1008,9 +1008,10 @@ Packet numbers are divided into 3 spaces in QUIC:
 As described in {{QUIC-TLS}}, each packet type uses different protection keys.
 
 Conceptually, a packet number space is the context in which a packet can be
-processed and ACKed.  Initial packets can only be sent with Initial encryption
-keys and ACKed in packets which are also Initial packets.  Similarly, Handshake
-packets can only be sent and acknowledged in Handshake packets.
+processed and acknowledged.  Initial packets can only be sent with Initial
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1005,18 +1005,17 @@ Packet numbers are divided into 3 spaces in QUIC:
...
-As described in {{QUIC-TLS}}, each packet type uses different encryption keys.
+As described in {{QUIC-TLS}}, each packet type uses different protection keys.
 
-Conceptually, a packet number space is the encryption context in which
-a packet can be processed and ACKed.  Initial packets can only be sent
-with Initial encryption keys and ACKed in packets which are also
-Initial packets.  Similarly, Handshake packets can only be sent and
-acknowledged in Handshake packets.
+Conceptually, a packet number space is the context in which a packet can be
+processed and ACKed.  Initial packets can only be sent with Initial encryption
```
MikeBishop: Should this also be protection?

draft-ietf-quic-transport.md
```diff
@@ -1005,18 +1005,17 @@ Packet numbers are divided into 3 spaces in QUIC:
...
+keys and ACKed in packets which are also Initial packets.  Similarly, Handshake
+packets can only be sent and acknowledged in Handshake packets.
```
MikeBishop: "Handshake packets can only be sent in Handshake packets" is either tautological or incorrect, and I'm uncertain which.  Either way, I don't think this says what you mean to say.

PR #1712, opened 2018-08-28
Discuss HTTP/QUIC Connection Closure

### LINKED ISSUE #1599: Discuss different paths to closure in HTTP
This is probably an expansion/relocation of the GOAWAY text; offline editorial feedback indicated that it was unclear from the HTTP document when to use GOAWAY vs. APPLICATION_CLOSE vs. CONNECTION_CLOSE, and readers shouldn't need to read the transport doc to figure it out.

### LINKED ISSUE #1717: GOAWAY with StreamID == 0
The current text says about the StreamID in the GOAWAY frame definition:

> This identifier ... MAY be zero if no requests were processed.

With the latest transport changes, stream 0 is now a valid bidirectional stream, so GOAWAY(0) actually means that stream 0 is accepted/could be opened.
This opens up a case where an endpoint cannot know for sure whether the remote endpoint has accepted stream 0 or not.

If you agree this is an issue, I see two options to fix this: 
one is to add a flag to the GOAWAY frame that identify this particular case; 
another one is to sacrifice one stream and slightly change the GOAWAY semantics to say that StreamID is the id of the last client-initiated stream plus one, i.e. the endpoint agree that any stream with ID >= StreamID was not accepted.

MikeBishop:
Fixes #1599.  Note that the MAX_PUSH_ID section is a red herring -- I moved so much text from the GOAWAY frame definition to this new section that GH thinks I moved the text in between instead.

Fixes #1717.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -243,6 +243,9 @@ can indicate that it is not authoritative for a request by sending a 421
 (Misdirected Request) status code in response to the request (see Section 9.1.2
 of {{!RFC7540}}).
 
+The considerations discussed in Section 9.1 of {{?RFC7540}} also apply to the
+management of HTTP/QUIC connections.
+
 # Stream Mapping and Usage {#stream-mapping}
 
 A QUIC stream provides reliable in-order delivery of bytes, but makes no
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -1042,17 +1045,89 @@ The GOAWAY frame applies to the connection, not a specific stream.  An endpoint
...
+be needed.  A gateway MAY use PING to maintain connections in anticipation of
+need rather than incur the latency cost of connection establishment to servers.
+Servers SHOULD NOT use PING frames to keep a connection open.
+
+## Connection Shutdown
+
+Even when a connection is not idle, either endpoint can decide to stop using the
+connection.  Clients do this by not sending additional requests on the
+connection; responses and pushed responses associated to previous requests will
+continue to completion.
```
martinthomson: This introduction doesn't mention that the intent is to be "graceful" or to close the connection without losing requests.  That's important, because it is also possible to just slam the connection closed.

draft-ietf-quic-http.md
```diff
@@ -1102,69 +1167,33 @@ allowing time for any in-flight requests (at least one round-trip time), the
 server MAY send another GOAWAY frame with an updated last Stream ID.  This
 ensures that a connection can be cleanly shut down without losing requests.
 
-Once all requests on streams at or below the identified stream number have been
-completed or cancelled, and all promised server push responses associated with
-those requests have been completed or cancelled, the connection can be closed
-using an Immediate Close (see {{QUIC-TRANSPORT}}).  An endpoint that completes a
-graceful shutdown SHOULD use the QUIC APPLICATION_CLOSE frame with the
-HTTP_NO_ERROR code.
+Once all accepted requests have been processed, the server MAY permit the
```
martinthomson: I think s/MAY/can here.  The normative language elsewhere covers this.

draft-ietf-quic-http.md
```diff
@@ -1102,69 +1167,33 @@ allowing time for any in-flight requests (at least one round-trip time), the
...
+connection to become idle, or MAY initiate an explicit closure of the
+connection.  An endpoint that completes a graceful shutdown SHOULD use the
+HTTP_NO_ERROR code when closing the connection.
 
+## Explicit Application Closure
```
martinthomson: Explicit or immediate?  Better to use the same terminology as the transport doc, methinks.  Explicit, taken to mean "direct", might be right, but it seems a little off.

draft-ietf-quic-http.md
```diff
@@ -1102,69 +1167,33 @@ allowing time for any in-flight requests (at least one round-trip time), the
...
-
+An HTTP/QUIC implementation can explicitly close the QUIC connection at any
+time. This results in sending a QUIC APPLICATION_CLOSE frame to the peer; the
+error code in this frame indicates to the peer why the connection is being
+closed.  See {{errors}} for error codes which can be used when closing a
+connection.
 
-# Connection Management
+Before closing the connection, a GOAWAY MAY be sent to allow the client to retry
+requests.  Including the GOAWAY frame in the same packet as the QUIC
```
martinthomson: "retry some requests" - the implication here is that GOAWAY is sent immediately before, which would limit its usefulness.

draft-ietf-quic-http.md
```diff
@@ -1102,69 +1167,33 @@ allowing time for any in-flight requests (at least one round-trip time), the
...
-that might not be needed.  A gateway MAY use PING to maintain connections in
-anticipation of need rather than incur the latency cost of connection
-establishment to servers.
+For various reasons, the QUIC transport could indicate to the application layer
+that the connection has terminated.  This might be due to an explicit closure
+by the peer, a transport-level error, or a change in network topology which
+interrupts connectivity.
 
+If a connection terminates without a GOAWAY frame, clients MUST assume
+that any request which was sent might have been processed.
```
martinthomson: Maybe mention "in whole or part" when it comes to requests, even a partially sent request can be processed.

draft-ietf-quic-http.md
```diff
@@ -1117,7 +1117,55 @@ connection.  Clients do this by not sending additional requests on the
...
+
+Servers SHOULD send a GOAWAY frame when the closing of a connection is known
+in advance, even if the advance notice is small, so that the remote peer can
+know whether a stream has been partially processed or not.  For example, if an
+HTTP client sends a POST at the same time that a server closes a QUIC
+connection, the client cannot know if the server started to process that POST
+request if the server does not send a GOAWAY frame to indicate what streams it
+might have acted on.
+
+An endpoint MAY send multiple GOAWAY frames if circumstances change. For
```
lnicco: I know this commit is just moving text around, but I wanted to note how the beginning of this paragraph seems to be for the most part the same as in HTTP2 and does not take into consideration that in HQ there is no error code in the GOAWAY frame. So perhaps the "sends GOAWAY without an error code" can be removed ?

In HQ the only reason I can think of for the endpoint to send multiple GOAWAY frames is to first signal the intention to drain the connection, so that the remote endpoint on receipt stops creating new streams, and then after 1RTT (or on confirmation that the GOAWAY was received) the local endpoint can actually stop accepting new streams and seal the last stream ID and send a new GOAWAY frame. The remote endpoint upon receipt can now reset streams it thinks are established but in truth are not, with HTTP_REQUEST_CANCELLED, as explained above.

Issue #1487, opened 2018-06-27
Guidance for flow control and ACK of extension frames


mikkelfj:
Given the past history, it is easy to overlook the complixity of ACK and flow control when adding new frames, so some guidance could be useful.

One option could be to define a few classes of behavior such as now flow no ACK, etc. and use an existing frame such as PING, PAD, etc. as example.

martinthomson:
Good idea.  We should document what an extension needs to consider in defining a new frame type.

PR #1713, opened 2018-08-28
ACK and PADDING

ianswett:
Neither ACK nor PADDING frames cause an ACK to be sent.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3418,14 +3418,15 @@ ACK Block (repeated):
 ### Sending ACK Frames
 
 Implementations MUST NOT generate packets that only contain ACK frames in
-response to packets which only contain ACK frames. However, they MUST
-acknowledge packets containing only ACK frames when sending ACK frames in
-response to other packets.  Implementations MUST NOT send more than one packet
-containing only an ACK frame per received packet that contains frames other than
-an ACK frame.  Packets containing non-ACK frames MUST be acknowledged
-immediately or when a delayed ack timer expires. The delayed ack timer MUST
-NOT delay an ACK for longer than an RTT, which ensures an ACK frame is sent
-at least once per RTT if new packets needing acknowledgement were received.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3418,14 +3418,15 @@ ACK Block (repeated):
...
-response to other packets.  Implementations MUST NOT send more than one packet
-containing only an ACK frame per received packet that contains frames other than
-an ACK frame.  Packets containing non-ACK frames MUST be acknowledged
-immediately or when a delayed ack timer expires. The delayed ack timer MUST
-NOT delay an ACK for longer than an RTT, which ensures an ACK frame is sent
-at least once per RTT if new packets needing acknowledgement were received.
+response to packets which only contain ACK and PADDING frames. However, they
+MUST acknowledge packets containing only ACK and PADDING frames when sending
+ACK frames in response to other packets.  Implementations MUST NOT send more
+than one packet containing only an ACK frame per received packet that contains
```
martinthomson: "containing only ACK or PADDING frames" ?

ianswett: To clarify, are you saying I should replace "ACK and PADDING" with "ACK or PADDING" everywhere or another suggestion?

martinthomson: Just the one instance here that you only have "ACK".

huitema: Should we mention that Challenge and Challenge response frames have the same treatment? What about extension frames that would be defined using the extension mechanism?

mikkelfj: @huitema there is already an issue open on extension frames https://github.com/quicwg/base-drafts/issues/1487

ianswett: I think only an ACK is most correct there.  If you really wanted to send ACK and PADDING, that would count towards bytes in flight, and then congestion control is involved, which provides a different rate limit.

martinthomson: SGTM

PR #1715, opened 2018-08-30
Add max_bytes_before_ack to transport

### LINKED ISSUE #1428: When the sender is using standard Reno congestion control, ack every ~2 packets
We should recommend acking every 2 packets or something similar when the sender is using Reno or another window based congestion control like Cubic.

Other congestion controllers like BBR are able to tolerate less frequent acking, but those controllers are not standardized and quite a bit more complex.

This makes me a bit sad, since I like acking less frequently, but the recovery doc should be self-consistent.

ianswett:
Possibly it needs to be mentioned in recovery as well?

Fixes #1428

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1671,7 +1671,7 @@ disable_migration (0x0009):
   NOT send any packets, including probing packets ({{probing}}), from a local
   address other than that used to perform the handshake.  This parameter is a
   zero-length value.
-  
+
 max_bytes_before_ack (0x000c):
 
 : An unsigned 32-bit integer that indicates the maximum number of bytes received
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3434,8 +3434,11 @@ NOT delay an ACK for longer than an RTT, which ensures an ACK frame is sent
 at least once per RTT if new packets needing acknowledgement were received.
 
 If the max_bytes_before_ack transport parameter has been received, the
-receiver SHOULD send an ACK frame immediately once that many octets of packets
-containing frames besides ACK or PADDING have been received.
+receiver SHOULD send an ACK frame immediately when at least that many octets
+of packets containing frames besides ACK or PADDING have been received since
+sending the previous ACK frame. If a receiver sends ACK frames less frequently,
+it risks causing the sender to send more slowly, which may lower the delivery
+rate at the receiver.
 
 To limit ACK blocks to those that have not yet been received by the sender, the
 receiver SHOULD track which ACK frames have been acknowledged by its peer.  Once
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1554,6 +1554,7 @@ language from Section 3 of {{!TLS13=RFC8446}}.
       disable_migration(9),
       initial_max_stream_data_bidi_remote(10),
       initial_max_stream_data_uni(11),
+      max_bytes_before_ack(12),
       (65535)
    } TransportParameterId;
 
@@ -1670,6 +1671,13 @@ disable_migration (0x0009):
   NOT send any packets, including probing packets ({{probing}}), from a local
   address other than that used to perform the handshake.  This parameter is a
   zero-length value.
+  
+max_bytes_before_ack (0x000c):
+
+: An unsigned 32-bit integer that indicates the maximum number of bytes received
+  before an ACK frame should be sent. The peer sending this value indicates
+  to the receiver the value it should use.  If this value is absent, the
+  receiver may use any conformant acknowledgement algorithm.
 
 Either peer MAY advertise an initial value for the flow control on each type of
 stream on which they might receive data.  Each of the following transport
@@ -3424,6 +3432,10 @@ an ACK frame.  Packets containing non-ACK frames MUST be acknowledged
 immediately or when a delayed ack timer expires. The delayed ack timer MUST
 NOT delay an ACK for longer than an RTT, which ensures an ACK frame is sent
 at least once per RTT if new packets needing acknowledgement were received.
+
+If the max_bytes_before_ack transport parameter has been received, the
+receiver SHOULD send an ACK frame immediately once that many octets of packets
+containing frames besides ACK or PADDING have been received.
 
 To limit ACK blocks to those that have not yet been received by the sender, the
 receiver SHOULD track which ACK frames have been acknowledged by its peer.  Once


draft-ietf-quic-transport.md
```diff
@@ -3425,6 +3433,10 @@ immediately or when a delayed ack timer expires. The delayed ack timer MUST
 NOT delay an ACK for longer than an RTT, which ensures an ACK frame is sent
 at least once per RTT if new packets needing acknowledgement were received.
 
+If the max_bytes_before_ack transport parameter has been received, the
+receiver SHOULD send an ACK frame immediately once that many octets of packets
+containing frames besides ACK or PADDING have been received.
```
martinthomson: What is the logic you are looking for?  Count the total size of received packets, unless they are ACK-only, then send an ACK when count>max_bytes_before_ack?

I would reframe this slightly in that case.  "The max_bytes_before_ack transport parameter can be used to request more frequent acknowledgments.  After sending a packet containing an ACK frame, a receiver counts the size of all packets it receives, omitting packets that only contain ACK or PADDING frames.  Once this count exceeds the value of the max_bytes_before_ack transport parameter it has received from a peer, the receiver MUST send an ACK frame immediately."

mikkelfj: IoT devices may receive data at a lower power cost than transmitting. Firing up the transmit radio could be costly. For these applications it makes more sense to transmit ACK's at timed intervals. They might skip an interval if to few bytes have been received.

ianswett: Yeah, I don't think we want MUST, but I'll add a caveat that if you don't do what the peer requests, you risk reduced performance.

draft-ietf-quic-transport.md
```diff
@@ -1671,6 +1672,13 @@ disable_migration (0x0009):
   address other than that used to perform the handshake.  This parameter is a
   zero-length value.
 
+max_bytes_before_ack (0x000c):
+
+: An unsigned 32-bit integer that indicates the maximum number of bytes received
+  before an ACK frame should be sent. The peer sending this value indicates
+  to the receiver the value it should use.  If this value is absent, the
+  receiver may use any conformant acknowledgement algorithm.
+
```
huitema: The current algorithms are either time based or packet based, requiring for example several ACKs per RTT, and specifying a maximum delay or a maximum number of packets before an ACK will be sent. What is the rationale for adding a byte based control? Also, given that the algorithms are time based or packet based, why not specify max_packets_before_ack, or max_ack_delay?

draft-ietf-quic-transport.md
```diff
@@ -3425,6 +3433,13 @@ immediately or when a delayed ack timer expires. The delayed ack timer MUST
 NOT delay an ACK for longer than an RTT, which ensures an ACK frame is sent
 at least once per RTT if new packets needing acknowledgement were received.
 
+If the max_bytes_before_ack transport parameter has been received, the
+receiver SHOULD send an ACK frame immediately when at least that many octets
+of packets containing frames besides ACK or PADDING have been received since
+sending the previous ACK frame. If a receiver sends ACK frames less frequently,
+it risks causing the sender to send more slowly, which may lower the delivery
+rate at the receiver.
+
```
huitema: This is not free. It comes in addition to existing logic which is already designed to prevent sender starvation. It requires at least one additional counter in the connection state, i.e., "max bytes sent at the last ACK time", and of course additional tests in the ACK sending logic.

cpaasch:
I am to some extend agreeing with what @huitema is saying. This is adding complexity at a late stage, without giving clear guidance,... as to how to use the max_bytes_before_ack:
* What value should a congestion-control announce to the peer ?
* How exactly should the receiver use the value? During slow-start, should it ACK more frequently to allow the sender to build a rate-estimation? Or should it always wait for the necessary bytes?

AFAICS, this can easily be added later on as an extension once it is clearer how to use it.

ianswett:
At the moment, we're providing no guidance at all, except to not delay ACKs more than 1RTT.

This may not be the right approach, but I think doing nothing here about these ack delay/aggregation issues are almost guaranteed to provide poor performance.

Jana and I have discussed this quite a bit in the past, and it should have made it into a PR sooner.  It's not intended as a last minute addition.

draft-ietf-quic-transport.md
ianswett: Agreed, but it is simple to implement.

draft-ietf-quic-transport.md
ianswett: Congestion control is byte-based, not packet based, hence the choice.  But I can see the packet vs byte based going either way.

In order to use max_ack_delay effectively, I think it's almost necessary to be able to update it during the connection.  Fraction of an RTT may not need to be updated during a connection, admittedly.  max_ack_delay also feeds into TLP and RTO timers, so making it fraction of an RTT could be a bit dangerous if one side has a poor estimate of RTT.

huitema:
@ianswett I agree that receivers could hamper senders' performance if they delay ACK too much. We should say that explicitly in the recovery draft. But sending ACKs too frequently also affects the senders' performance, and we should say that too. Bottom line, it is a balance.

We probably want to set default requirements for the number of packets before ACK and for the max ACK delay -- your arbitrary choices of 10 packets and 10 ms look as good as any, or maybe the min of that and 1/4th of the congestion window (for bytes/packets) or the RTT (for ACK delay), with something like at least two packets and 1ms as safe minimum. It would set a reasonable default. Anything else is dependent of the congestion control implementation at the sender, and probably requires cooperation between sender and receiver. That's what extensions are for.

But I do not think that adding this "byte limit" concept is useful, especially if there is no way to change it during the course of the connection.

ianswett:
As the receiver, you don't know the sender's congestion window, so fraction of RTT(ie: 1/4) is a more practical option I think.

2 packets and 1ms seem like reasonable minimum values.

huitema:
Actually, TCP implementation need to have at least a rough idea of the sender's congestion window. They use that to set the TCP flow control window. The equivalent in QUIC is the MAX DATA parameter; that too has to be set by the receiver using some kind of guess work. Granted, it is typically an overestimate, because receivers don't want to slow down the senders. But implementation could use that to derive kind of upper bound for the max byte value, or max packet value. If it gets fewer than 10 packets, then use that.

gloinul:
I think the first issue is that currently TPs are only negotiated at connection establishment. Which, makes it impossible to change them if one change MTUs due to PLPMTUD or a client side migration. 

I also agree that adding clarified recommendations on delayed ACK. I think one full RTT is a bit much delay when one have something to ACK. I am not certain I understand the full relation and precedens of the minimal, maximum etc. Here I think one should be careful to write up rules and analys their impact and what will happen in the various cases, with both really low RTTs (Backplane/loopback), as well as LANs and long fat pipes. In Kista we discussed that 1/4 of an RTT likely makes sense as an upper limit for sender stacks that accept delayed ACKs.

ianswett:
Based on today's discussion, I'm closing this.

PR #1730, opened 2018-09-09
Bundle new client connection ID and path challenge in migration probes

huitema:
During migration, the server will respond to the client's probe. If the client does not use a null length connection ID, the server needs to chose one of the destination connection ID specified by the client in "new connection ID" frames. There is however a synchronization issue. The client may have sent a new connection ID frame on another path, but that frame may not have yet been received when the probe arrives. The server is in a bind. It does not know whether the client sent a new connection ID or not. It does not know whether it should reuse an old ID and compromise privacy, or delay processing of the probe.

The synchronization issue is solved if the probe bundles NEW_CONNECTION_ID and PATH_CHALLENGE frames.


draft-ietf-quic-transport.md
```diff
@@ -2074,6 +2074,9 @@ path validation with other frames.  For instance, an endpoint may pad a packet
 carrying a PATH_CHALLENGE for PMTU discovery, or an endpoint may bundle a
 PATH_RESPONSE with its own PATH_CHALLENGE.
 
+When probing a new path, an endpoint may want to ensure that the response, if
+any, is sent to a specific connection ID. The endpoint achieves that by bundling
+NEW_CONNECTION_ID and PATH_CHALLENGE frames, in that order.
```
martinthomson: This implies that order is significant.  It also implies that bundling the two means that this specific connection ID will be used.  We've tried really hard to avoid special inter-frame dependencies like this.  PATH_CHALLENGE is special in the sense that the path it follows is relevant to its processing (which is a necessary evil), but this implies much more than that.  I think that rewording this to suggest that providing NEW_CONNECTION_ID ahead of PATH_CHALLENGE is useful (even necessary if the peer has run out, as they might have) and that having them in the same packet means that the NEW_CONNECTION_ID is guaranteed to be available.  Anything more than might give the wrong impression.

draft-ietf-quic-transport.md
```diff
@@ -2187,7 +2190,10 @@ usable for this connection.  Failure to validate a path does not cause the
 connection to end unless there are no valid alternative paths available.
 
 An endpoint uses a new connection ID for probes sent from a new local address,
-see {{migration-linkability}} for further discussion.
+see {{migration-linkability}} for further discussion. An endpoint that uses
+a new local address should include a NEW_CONNECTION_ID frame in the probe,
```
martinthomson: "SHOULD"? I don't think that we need to be this suggestive here, other than to note that lack of connection IDs will make probing - and responding to probes - impossible.

draft-ietf-quic-transport.md
```diff
@@ -2359,6 +2365,11 @@ genuine migrations.  Changing port number can cause a peer to reset its
 congestion state (see {{migration-cc}}), so the port SHOULD only be changed
 infrequently.
 
+Endpoints could also have their activity correlated if their peers keep using
+the same destination connection id after migration. Nodes that initiate a
```
martinthomson: s/id/ID

draft-ietf-quic-transport.md
huitema: Agreed. I was being terse. That said, migration results in a new pair of <local cid>, <remote cid>. It would be really nice if the nodes could chose the ID by which they accept packets on the new path.

draft-ietf-quic-transport.md
huitema: Note that there is another issue in the same section. The text says that a packet containing PATH_CHALLENGE, PATH_RESPONSE, and PADDING frames is a "probing frame". Adding the NEW_CONNECTION_ID makes that a "non probing packet". I am not sure I fully get the consequences.

draft-ietf-quic-transport.md
huitema: This is fixed. Please look at the new text.

draft-ietf-quic-transport.md
```diff
@@ -2359,6 +2366,11 @@ genuine migrations.  Changing port number can cause a peer to reset its
 congestion state (see {{migration-cc}}), so the port SHOULD only be changed
 infrequently.
 
+Endpoints could also have their activity correlated if their peers keep using
+the same destination connection ID after migration. Nodes that initiate a
+migration need to provide their peers with new connection IDs. The goal is
```
ianswett: need is non-normative, which feels odd to me, and there's no mention of zero-length connection IDs.

huitema: Yes. I rewrote that paragraph. Tell me what you think of the new version.

draft-ietf-quic-transport.md
```diff
@@ -2366,11 +2366,16 @@ genuine migrations.  Changing port number can cause a peer to reset its
 congestion state (see {{migration-cc}}), so the port SHOULD only be changed
 infrequently.
 
-Endpoints could also have their activity correlated if their peers keep using
-the same destination connection ID after migration. Nodes that initiate a
-migration need to provide their peers with new connection IDs. The goal is
+Endpoints that use connection IDs with length greater than zero
+could also have their activity correlated if their peers keep using
```
ianswett: nit: I think the "also" is unnecessary.

draft-ietf-quic-transport.md
```diff
@@ -2366,11 +2366,16 @@ genuine migrations.  Changing port number can cause a peer to reset its
...
+the same destination connection ID after migration. If a node receives
+packets through a newly used connection ID, it SHOULD select an
```
ianswett: I'm wondering whether this should be a MUST in this case.  Not only are you migrating, but the peer has changed CIDs, which seems like strong evidence it's trying to limit linkability?

mikkelfj: Yes, it would seem rather pointerless to change CID otherwise. Responding with an old CID is the same as the peer never changing CID. The only reason is running out of CID's and not being that sensitive to linkability. As we now, opting out voluntarily does not work well in crypto.

huitema:
I understand that this could be stronger, but I am hesitant to put stronger words there without consensus. There is an obvious trade-off between privacy and performance. For example, should a server wait to answer a probe until a new connection ID is available? Maybe we could discuss this in NYC.

draft-ietf-quic-transport.md
huitema: @mikkelfj the whole point of this PR is to say that and ensure that it works. Do you want to specify what the server should do if the client's probe arrives before a new connection ID is available? By your logic, we want to say that the server should wait before responding to the probe.

Then, do you want to enforce that? For example, suppose that a client sent a probe with a new CID, and receives a response with an already-in-use CID. Should it just shrug and continue, or signal a protocol violation?

huitema:
Note that if endpoints really want to avoid correlation, they don't want a pattern in which the client sends packets from new address to new CID, but each packet triggers a response to old address and old CID. That too would be reasonably easy to observe. Yet that's probably what you get if the server does not want to use the new path yet and keeps sending data on the old path. I think that bundling the New CID with the client's probe solves that, or at least removes one of the reasons for the server to wait.

draft-ietf-quic-transport.md
huitema: Responding on the old path is problematic too, because it causes correlation between new and old path.

draft-ietf-quic-transport.md
MikeBishop: The consequence is that you'd be initiating a change to the new path, and not just a probe.  I think we want to expand the definition of probing packets to include NEW_CONNECTION_ID as well.

PR #1732, opened 2018-09-09
Correct application of RFC2119

jugglinmike:
The term "MAY" is better suited to describing implementer action (rather
than implementer desire).


martinthomson:
The ghost of RFC 6919...

jugglinmike:
Ooo, I haven't seen that one before. Interesting!

PR #1734, opened 2018-09-09
Correct transitions described by diagram

jugglinmike:
Receiving the RST_STREAM frame may cause a transition from "Data Recvd"
to "Reset Recvd," but it cannot cause a transition in the other
direction. Update the diagram to accurately describe the condition which
may cause a transition in that direction.


martinthomson:
I had trouble parsing that myself recently, but forgot that I should just fix it.  Thanks.

PR #1735, opened 2018-09-10
Stateless reset is now available to both peers

### LINKED ISSUE #1728: When to send Stateless Reset
Section 6.2.2. Server Packet Handling, says, at the end:

> Servers MUST drop incoming packets under all other circumstances. They SHOULD send a Stateless Reset (Section 6.13.4) if they are able.

The last sentence is a bit dangerous.  A stateless reset should NOT be sent when the endpoint has access to the connection state and can't decrypt a packet.  Otherwise any random packet would kill a connection.  Saying "if they are able" might be interpreted incorrectly.  Can we just change it to "They SHOULD send a Stateless Reset (Section 6.13.4) when they no longer have access to the complete connection state."?

The way I understand it, a stateless reset token would be something that the endpoint saves in non-volatile memory which can be used to reset connections after a crash/reboot.

martinthomson:
Also, don't create the impression that stateless reset can be used when
a packet can't be decrypted.

Closes #1728.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1278,6 +1278,9 @@ IDs, QUIC processes the packet as part of that connection. Endpoints MUST drop
 packets with zero-length Destination Connection ID fields if they do not
 correspond to a single connection.
 
+Endpoints SHOULD send a Stateless Reset ({{stateless-reset}}) for any packets
+that cannot be attributed to an existing connection.
+
 
 ### Client Packet Handling {#client-pkt-handling}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1278,6 +1278,9 @@ IDs, QUIC processes the packet as part of that connection. Endpoints MUST drop
 packets with zero-length Destination Connection ID fields if they do not
 correspond to a single connection.
 
+Endpoints SHOULD send a Stateless Reset ({{stateless-reset}}) for any packets
```
ianswett: I think this should be restricted to short-header packets, correct?

rpaulo: We already have text in the Stateless Reset section saying an endpoint MAY send a stateless reset for a LH packet.  I can't really understand how an implementation would do that, though.

martinthomson: Yes, let's not reiterate that discussion.  Right now, it's basically impossible to except in a very narrow window during the handshake, but we agreed that there was no reason not to allow it.  If the long header is used after the handshake completes (for reasons we come up with later), then we don't have to worry about stateless reset.

PR #1739, opened 2018-09-11
Tighten error processing of NEW CONNECTION ID

huitema:
Generate an error if the same connection ID is announced twice, with different reset secret or different sequence number. Also prohibits repeating the connection ID used in handshake in a new connection ID frame.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3218,6 +3218,19 @@ value of the connection ID changed.  An endpoint that is sending packets with a
 zero-length Destination Connection ID MUST treat receipt of a NEW_CONNECTION_ID
 frame as a connection error of type PROTOCOL_VIOLATION.
 
+Transmission errors, time-outs and retransmission may cause the same
+NEW_CONNECTION_ID frame to be received multiple time. Additional
+transmissions SHOULD be ignored and MUST NOT cause a connection
+error.
+
+If an endpoint receives a NEW_CONNECTION_ID frame that repeats the same
+connection ID as a previous NEW_CONNECTION_ID framebut with a different
+Stateless Reset Token or a different Sequence, the endpoint MUST
+treat that receipt as a connection error of type PROTOCOL_VIOLATION.
+Similarly, if an endpoint receives a NEW_CONNECTION_ID frame that repeats
+the Source Connection ID used by the peer during the initial
+handshake, it MUST treat that receipt as a connection error of type
+PROTOCOL_VIOLATION.
 
 ## STOP_SENDING Frame {#frame-stop-sending}
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3224,7 +3224,7 @@ transmissions SHOULD be ignored and MUST NOT cause a connection
 error.
 
 If an endpoint receives a NEW_CONNECTION_ID frame that repeats the same
-connection ID as a previous NEW_CONNECTION_ID framebut with a different
+connection ID as a previous NEW_CONNECTION_ID frame but with a different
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3218,6 +3218,19 @@ value of the connection ID changed.  An endpoint that is sending packets with a
...
+Transmission errors, time-outs and retransmission may cause the same
+NEW_CONNECTION_ID frame to be received multiple time. Additional
+transmissions SHOULD be ignored and MUST NOT cause a connection
+error.
+
+If an endpoint receives a NEW_CONNECTION_ID frame that repeats the same
+connection ID as a previous NEW_CONNECTION_ID frame but with a different
+Stateless Reset Token or a different Sequence, the endpoint MUST
+treat that receipt as a connection error of type PROTOCOL_VIOLATION.
+Similarly, if an endpoint receives a NEW_CONNECTION_ID frame that repeats
```
mikkelfj: That MUST is hard to achieve if the context has been erased meanwhile. Surely the intention is not to log ID's forever?

draft-ietf-quic-transport.md
```diff
@@ -3218,6 +3218,19 @@ value of the connection ID changed.  An endpoint that is sending packets with a
 zero-length Destination Connection ID MUST treat receipt of a NEW_CONNECTION_ID
 frame as a connection error of type PROTOCOL_VIOLATION.
 
+Transmission errors, time-outs and retransmission may cause the same
+NEW_CONNECTION_ID frame to be received multiple time. Additional
```
martinthomson: s/time/times

draft-ietf-quic-transport.md
martinthomson: As @mikkelfj says, I think that we only need to say "MAY" for these and let nature take its course.  A "MAY treat as a protocol error" turns into a fairly firm requirement not to send duplicates.

huitema:
Just pushed a commit that replaces MUST by MAY. I am still pondering @martinthomson question whether it is _"reasonable to send NEW_CONNECTION_ID with the same connection ID and a different token just to refresh the stateless reset keys."_ There are pros and cons either way.

Sending an error is simpler, it does not require updating state. But asking to update the reset token may be a better specification, because it is more likely result in consistent implementations. The error will only be sent if it is detected, and it will not be detected if the implementation lost trace of the old CID. We are thus getting potentially inconsistent results, some implementations generating errors and some not.

We will get better consistency if we ask endpoints to just update the reset secret associated with the CID. The implementations that keep trace of the old CID would just perform the update. The implementations that lost trace would register a new CID with the new reset secret. Both types of implementation would end up with the same result, which might be more consistent with the idea that frames are idempotent.

But then, there is an ordering issue. QUIC does not guarantee that frames are delivered in order. Even if implementation track whether the previous frame has been acknowledged, there is always the risk of a delayed copy popping up later. In that case, we are playing reset-roulette. I suppose we could build machinery to guarantee the ordering of the frames, but asking implementations to "only use a connection ID once" feels much simpler. Hence my choice of just detecting an error if parameters change.

martinthomson:
> asking implementations to "only use a connection ID once" feels much simpler

My conclusion exactly. :)

huitema:
@mikkelfj are you referring to "_MUST NOT be an error to receive twice_" when you say "_This MUST actually reintroduces the problem of remembering old ID transmissions. It cannot ignore without remembering forever_"? Because there is no such requirement. You need to remember if you want to signal errors, but we are rpecisely telling you, "don't do that".

mikkelfj:
No, I made the comment seconds after Martin comitted a change, snd Martin comited a new change seconds after my comment. The MUST NOT in its current form is fine.

PR #1742, opened 2018-09-12
Connection ID DT output

erickinnear:
- Add new text around Connection IDs
- Add CONNECTION_ID_FINISHED frame
- Remove sequence number for Connection IDs

Short summary: 
An endpoint can issue as many CIDs to its peer as it chooses. It must always allow incoming packets on all outstanding CIDs. The only way to get rid of one from that "pool" is for the peer to indicate that the CID may be forgotten via the CONNECTION_ID_FINISHED frame. 

On the receiving side, you remember all the CIDs you get from your peer and can generally use any that you like at any time. If you get more than you think you need, you can ignore some. 
You're only allowed to use a CID on a single path, so you should probably retire any that you've moved away from via a _FINISHED frame. 

Receiving a _FINISHED frame indicates that you should generally "replace" the retired CID. There are a few recommendations in this area, mostly to ensure that it's straightforward to keep your peer from running out of CIDs to use.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1143,28 +1143,29 @@ For all other frames, the Frame Type field simply identifies the frame.  These
 frames are explained in more detail as they are referenced later in the
 document.
 
-| Type Value  | Frame Type Name   | Definition                  |
-|:------------|:------------------|:----------------------------|
-| 0x00        | PADDING           | {{frame-padding}}           |
-| 0x01        | RST_STREAM        | {{frame-rst-stream}}        |
-| 0x02        | CONNECTION_CLOSE  | {{frame-connection-close}}  |
-| 0x03        | APPLICATION_CLOSE | {{frame-application-close}} |
-| 0x04        | MAX_DATA          | {{frame-max-data}}          |
-| 0x05        | MAX_STREAM_DATA   | {{frame-max-stream-data}}   |
-| 0x06        | MAX_STREAM_ID     | {{frame-max-stream-id}}     |
-| 0x07        | PING              | {{frame-ping}}              |
-| 0x08        | BLOCKED           | {{frame-blocked}}           |
-| 0x09        | STREAM_BLOCKED    | {{frame-stream-blocked}}    |
-| 0x0a        | STREAM_ID_BLOCKED | {{frame-stream-id-blocked}} |
-| 0x0b        | NEW_CONNECTION_ID | {{frame-new-connection-id}} |
-| 0x0c        | STOP_SENDING      | {{frame-stop-sending}}      |
-| 0x0d        | ACK               | {{frame-ack}}               |
-| 0x0e        | PATH_CHALLENGE    | {{frame-path-challenge}}    |
-| 0x0f        | PATH_RESPONSE     | {{frame-path-response}}     |
-| 0x10 - 0x17 | STREAM            | {{frame-stream}}            |
-| 0x18        | CRYPTO            | {{frame-crypto}}            |
-| 0x19        | NEW_TOKEN         | {{frame-new-token}}         |
-| 0x1a        | ACK_ECN           | {{frame-ack-ecn}}           |
+| Type Value  | Frame Type Name        | Definition                       |
+|:------------|:-----------------------|:---------------------------------|
+| 0x00        | PADDING                | {{frame-padding}}                |
+| 0x01        | RST_STREAM             | {{frame-rst-stream}}             |
+| 0x02        | CONNECTION_CLOSE       | {{frame-connection-close}}       |
+| 0x03        | APPLICATION_CLOSE      | {{frame-application-close}}      |
+| 0x04        | MAX_DATA               | {{frame-max-data}}               |
+| 0x05        | MAX_STREAM_DATA        | {{frame-max-stream-data}}        |
+| 0x06        | MAX_STREAM_ID          | {{frame-max-stream-id}}          |
+| 0x07        | PING                   | {{frame-ping}}                   |
+| 0x08        | BLOCKED                | {{frame-blocked}}                |
+| 0x09        | STREAM_BLOCKED         | {{frame-stream-blocked}}         |
+| 0x0a        | STREAM_ID_BLOCKED      | {{frame-stream-id-blocked}}      |
+| 0x0b        | NEW_CONNECTION_ID      | {{frame-new-connection-id}}      |
+| 0x0c        | STOP_SENDING           | {{frame-stop-sending}}           |
+| 0x0d        | ACK                    | {{frame-ack}}                    |
+| 0x0e        | PATH_CHALLENGE         | {{frame-path-challenge}}         |
+| 0x0f        | PATH_RESPONSE          | {{frame-path-response}}          |
+| 0x10 - 0x17 | STREAM                 | {{frame-stream}}                 |
+| 0x18        | CRYPTO                 | {{frame-crypto}}                 |
+| 0x19        | NEW_TOKEN              | {{frame-new-token}}              |
+| 0x1a        | ACK_ECN                | {{frame-ack-ecn}}                |
+| 0x1b        | CONNECTION_ID_FINISHED | {{frame-connection-id-finished}} |
 {: #frame-types title="Frame Types"}
 
 All QUIC frames are idempotent.  That is, a valid frame does not cause
@@ -1224,41 +1225,77 @@ be routed back to the endpoint and identified by the endpoint upon receipt.
 
 A zero-length connection ID MAY be used when the connection ID is not needed for
 routing and the address/port tuple of packets is sufficient to associate them to
-a connection. An endpoint whose peer has selected a zero-length connection ID
+a connection.  An endpoint whose peer has selected a zero-length connection ID
 MUST continue to use a zero-length connection ID for the lifetime of the
 connection and MUST NOT send packets from any other local address.
 
-When an endpoint has requested a non-zero-length connection ID, it will issue a
-series of connection IDs over the lifetime of a connection. The series of
-connection IDs issued by an endpoint is ordered, with the final connection ID
-selected during the handshake coming first.  Additional connection IDs are
-provided using the NEW_CONNECTION_ID frame ({{frame-new-connection-id}}), each
-with a specified sequence number.  The series of connection IDs issued SHOULD be
-contiguous, but might not appear to be upon receipt due to reordering or loss.
+An endpoint whose peer has selected a non-zero-length connection ID will
+potentially send packets using different connection IDs over the lifetime of a
+connection. These connection IDs are supplied by the peer using the
+NEW_CONNECTION_ID frame ({{frame-new-connection-id}}).
 
-Each connection ID MUST be used on only one local address. When packets are sent
-for the first time on a new local address, a new connection ID MUST be used with
-a higher sequence number than any connection ID previously used on any local
-address.  At any time, an endpoint MAY change to a new connection ID on a local
-address already in use.
 
-An endpoint MUST NOT send packets with a connection ID which has a lower
-sequence number than the highest sequence number of any connection ID ever sent
-or received on that local address.  This ensures that when an endpoint migrates
-to a new path or changes connection ID on an existing path, the packets will use
-a new connection ID in both directions.
+### Issuing Connection IDs
 
-Implementations SHOULD ensure that peers have a connection ID with a matching
-sequence number available when changing to a new connection ID.  An
-implementation could do this by always supplying a corresponding connection ID
-to a peer for each connection ID received from that peer.
+An endpoint issues connection IDs to its peer for the peer to use when sending
+packets. This allows the endpoint control over the strategy used to interpret
+connection IDs of packets that it receives.  These connection IDs are
+communicated to the peer using NEW_CONNECTION_ID frames
+({{frame-new-connection-id}}).
 
-While endpoints select connection IDs as appropriate for their implementation,
-the connection ID MUST NOT include the unprotected sequence number.  Endpoints
-need to be able to recover the sequence number associated with each connection
-ID they generate without relying on information available to unaffiliated
-parties. A connection ID that encodes an unencrypted sequence number could be
-used to correlate connection IDs across network paths.
+The initial connection ID issued to the peer is sent by the endpoint as the
+Source Connection ID during the handshake, ensuring that the peer always begins
+the connection with at least one connection ID to use when sending.  An endpoint
+SHOULD supply its peer with additional connection IDs via NEW_CONNECTION_ID
+frames. While each endpoint can choose how many connection IDs to issue, a
+recommended number of outstanding connection IDs is eight.
+
+When an endpoint issues a connection ID, it MUST accept packets using this
+connection ID for the duration of the connection or until its peer invalidates
+the connection ID via a CONNECTION_ID_FINISHED frame
+({{frame-connection-id-finished}}).
+
+Implementations SHOULD ensure that peers have sufficient connection IDs
+available to reduce the possibility of peers exhausting their supply of
+available connection IDs.  An implementation could do this by always supplying a
+new connection ID for each connection ID retired with a CONNECTION_ID_FINISHED
+frame. When a receiver of a packet notices that its peer is now using a
+previously unused connection ID, it can choose to supply its peer with a new
+connection ID using a NEW_CONNECTION_ID frame to reduce the possibility of its
+peer running out of available connection IDs.
+
+An endpoint that receives a packet with a different remote address than
+previously used SHOULD also switch to sending with a different connection ID.
+This can help to ensure that different connection IDs will be used in both
+directions when an endpoint migrates to a new path or changes connection ID on
+an existing path.
+
+
+### Consuming Connection IDs
+
+If an endpoint's peer has selected a non-zero-length connection ID, the endpoint
+maintains a set of connection IDs received from the peer that it can use when
+sending packets.  All connection IDs issued by the peer are considered valid for
+use by the endpoint when sending packets until the connection ID is retired by
+the endpoint.  Endpoints can choose to stop using a given connection ID to send
+packets at any time and signal this to the issuing endpoint via a
+CONNECTION_ID_FINISHED frame.  This frame indicates the connection ID that is no
+longer in use and serves as a request for the peer to issue additional
+connection IDs via a NEW_CONNECTION_ID frame.
+
+An endpoint that retires a connection ID should retain knowledge of that
+connection ID for a reasonable time after sending the CONNECTION_ID_FINISHED
+frame, or until that frame is acknowledged.  A recommended time is three times
+the current Retransmission Timeout (RTO) interval as defined in
+{{QUIC-RECOVERY}}.  This prevents potential retransmissions of a
+NEW_CONNECTION_ID frame from overlapping with the CONNECTION_ID_FINISHED frame
+for the same connection ID.
+
+Additionally, each connection ID MUST be used on packets sent from only one
+local address.  At any time, an endpoint MAY change to a new connection ID on a
+local address already in use.  An endpoint that migrates to a new local address
+SHOULD retire all connection IDs that were used on the previous address
+using the CONNECTION_ID_FINISHED frame.
 
 
 ## Matching Packets to Connections {#packet-handling}
@@ -2606,6 +2643,9 @@ the NEW_CONNECTION_ID frame sent by either peer, and servers can specify the
 stateless_reset_token transport parameter during the handshake (clients cannot
 because their transport parameters don't have confidentiality protection).  This
 value is protected by encryption, so only client and server know this value.
+Tokens sent via NEW_CONNECTION_ID frames are invalidated when their associated
+connection ID is retired via a CONNECTION_ID_FINISHED frame
+({{frame-connection-id-finished}}).
 
 An endpoint that receives packets that it cannot process sends a packet in the
 following layout:
@@ -3192,13 +3232,11 @@ An endpoint sends a NEW_CONNECTION_ID frame (type=0x0b) to provide its peer with
 alternative connection IDs that can be used to break linkability when migrating
 connections (see {{migration-linkability}}).
 
-The NEW_CONNECTION_ID is as follows:
+The NEW_CONNECTION_ID frame is as follows:
 
 ~~~
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                          Sequence (i)                       ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |   Length (8)  |          Connection ID (32..144)            ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1224,41 +1225,77 @@ be routed back to the endpoint and identified by the endpoint upon receipt.
...
+frame. When a receiver of a packet notices that its peer is now using a
+previously unused connection ID, it can choose to supply its peer with a new
+connection ID using a NEW_CONNECTION_ID frame to reduce the possibility of its
+peer running out of available connection IDs.
+
+An endpoint that receives a packet with a different remote address than
+previously used SHOULD also switch to sending with a different connection ID.
+This can help to ensure that different connection IDs will be used in both
+directions when an endpoint migrates to a new path or changes connection ID on
+an existing path.
```
MikeBishop: Technically, it doesn't do anything for changes on the existing path, but that's okay because the session is linkable anyway if you're on an existing path.

MikeBishop:
One item to note is that we've adopted EKR's suggestion to use IP:port (collectively "address" as in the migration section) changes instead of CID changes as the trigger to change your own CID.  Here's why:
  - If you switch CIDs without changing addresses they're already linked to any observers, so it doesn't matter whether the remote side changes.
  - If you switch addresses without changing CIDs, they're already linked to any observers, so it doesn't matter whether the remote side changes.
  - If you switch addresses and CIDs at the same time, you have an opportunity not to be linked.

The case we actually need to cover is ensuring that, if an endpoint changes address and CID at the same time, the peer absolutely does not respond with the old CID. Any rule that produces this is sufficient, even if it also produces some gratuitous changes along the way.

Part of what led to the changes in -13 is that switching on CID change makes it hard to avoid looping. Switching on endpoint address covers the scenario we need without requiring a fixed ordering of CIDs.

mikkelfj:
@MikeBishop this make sense, but in symmetric settings, both sides could have NAT rebindings, so there could a case for preemptive CID change at either endpoint after idletime - but perhaps that is a separate issue.

MikeBishop:
That's already a recommendation in the existing text -- this is strictly talking about how to ensure that the CID coming back from the peer doesn't link, but also doesn't trigger a new CID rotation, which triggers a rotation, which triggers....

draft-ietf-quic-transport.md
```diff
@@ -3257,6 +3287,49 @@ the Source Connection ID used by the peer during the initial
...
+Length:
+
+: An 8-bit unsigned integer containing the length of the connection ID.
+
+Connection ID:
+
+: A connection ID of the specified length.
+
+Receipt of a CONNECTION_ID_FINISHED frame containing a connection ID that was
+not previously sent to the peer MAY be treated as a connection error of type
```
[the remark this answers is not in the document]
mikkelfj: Umm aren't you reversing direction of CONNECTION_ID_FINISHED. One end sends ID's, the other end finishes them so ACK shouldn't matter. On the MAY: if the endpoint tracks generated CID's via an algorithm such as hashing a counter. With relatively short CIDs there are possible collisions and you might not want to actually store the CIDs already generated.

draft-ietf-quic-transport.md
```diff
@@ -1224,41 +1225,77 @@ be routed back to the endpoint and identified by the endpoint upon receipt.
...
+connection IDs of packets that it receives.  These connection IDs are
+communicated to the peer using NEW_CONNECTION_ID frames
+({{frame-new-connection-id}}).
+
+The initial connection ID issued to the peer is sent by the endpoint as the
+Source Connection ID during the handshake, ensuring that the peer always begins
+the connection with at least one connection ID to use when sending.  An endpoint
+SHOULD supply its peer with additional connection IDs via NEW_CONNECTION_ID
+frames. While each endpoint can choose how many connection IDs to issue, a
+recommended number of outstanding connection IDs is eight.
```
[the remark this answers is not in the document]
martinthomson: I think that 8 is the right recommendation, but we should contextualize it more.  The primary consideration being that it is important to support a number of failed connection migrations on the way to a successful migration.

That context is later, so I think we should just move this recommendation.

draft-ietf-quic-transport.md
martinthomson: It's not always the case that use of an existing path implies linkability.  Though it might be a strong signal, if I'm behind a NAT with a large number of other people, I can imagine the NAT being used as an mixing point.  Changing connection ID in that context might re-establish anonymity within that set.  Especially if the NAT understands connection IDs and I am able to use one for packets I receive as well.  In many cases we think about, the NAT is small and the value of re-mixing low, but you can imagine building that sort of aggregator for privacy purposes.

martinthomson: A recent PR highlighted the risks associated with that.  That text now belongs here.  See [this caution](https://github.com/quicwg/base-drafts/commit/17ccc69ecd0c652bc8f35c5db5fb3a37b0f7c67c#diff-db016291106766877c4921a79f8596e0R2380), which suggests that this change adds duplicated text...

draft-ietf-quic-transport.md
```diff
@@ -1224,41 +1225,77 @@ be routed back to the endpoint and identified by the endpoint upon receipt.
...
+
+### Consuming Connection IDs
+
+If an endpoint's peer has selected a non-zero-length connection ID, the endpoint
+maintains a set of connection IDs received from the peer that it can use when
+sending packets.  All connection IDs issued by the peer are considered valid for
+use by the endpoint when sending packets until the connection ID is retired by
+the endpoint.  Endpoints can choose to stop using a given connection ID to send
+packets at any time and signal this to the issuing endpoint via a
+CONNECTION_ID_FINISHED frame.  This frame indicates the connection ID that is no
```
martinthomson: I don't like this name (never have).  But until I read this text again and saw it use "retire", I didn't have an alternative I liked enough to suggest.  RETIRE_CONNECTION_ID is better (not awesome, but certainly better).

Yes, that's opinion.  Also, don't bother changing it.  I am prepared to do that.

draft-ietf-quic-transport.md
```diff
@@ -1224,41 +1225,77 @@ be routed back to the endpoint and identified by the endpoint upon receipt.
...
+longer in use and serves as a request for the peer to issue additional
+connection IDs via a NEW_CONNECTION_ID frame.
```
martinthomson: from "and serves as a request ..." is a new concept and doesn't belong to this paragraph.  It needs a new paragraph explaining things more fully.

Rather than frame this as an implicit request, I would say that you address the text at the endpoint that issues the connection IDs.  That is, rely on the text in the previous section and don't mention that function here at all.

draft-ietf-quic-transport.md
```diff
@@ -1224,41 +1225,77 @@ be routed back to the endpoint and identified by the endpoint upon receipt.
...
+
+An endpoint that retires a connection ID should retain knowledge of that
+connection ID for a reasonable time after sending the CONNECTION_ID_FINISHED
```
martinthomson: s/reasonable time/period/

I think we can leave the use of the word "reasonable" to lawyers.  It implies judgment.

draft-ietf-quic-transport.md
```diff
@@ -1224,41 +1225,77 @@ be routed back to the endpoint and identified by the endpoint upon receipt.
...
+frame, or until that frame is acknowledged.  A recommended time is three times
```
martinthomson: The acknowledgment point here is a tricky one.  The risk you describe is that a retransmission of NEW_CONNECTION_ID(X) arrives after RETIRE_CONNECTION_ID(X) is sent, which makes it appear as though X is a wholly new connection ID.  If it is then used, it might trigger stateless reset.  Defense against that is to wait for the ACK of the ACK for NEW_CONNECTION_ID, or to start the timer from when NEW_CONNECTION_ID is received, not when RETIRE_CONNECTION_ID is sent.

However, based on comments below, I think that we should recommend retention of the connection ID at the peer that uses it for 3xRTO as well and NOT short that timer when the RETIRE_CONNECTION_ID frame is acknowledged.  The reason for that is not use of the ID (that should have stopped before sending the frame), but acceptance of the stateless reset.  Packets in flight before the connection ID is retired might generate a stateless reset.

The risk I'm more concerned about is the receipt of delayed packets with X after RETIRE_CONNECTION_ID(X) is *received*.  In which case the recipient of RETIRE_CONNECTION_ID needs to remember X and not generate a stateless reset for some period.

That keeps it simple: remember a connection ID (and its associated reset token) from the time of sending or receiving RETIRE_CONNECTION_ID plus 3xRTO.

draft-ietf-quic-transport.md
```diff
@@ -1224,41 +1225,77 @@ be routed back to the endpoint and identified by the endpoint upon receipt.
...
+the current Retransmission Timeout (RTO) interval as defined in
+{{QUIC-RECOVERY}}.  This prevents potential retransmissions of a
+NEW_CONNECTION_ID frame from overlapping with the CONNECTION_ID_FINISHED frame
+for the same connection ID.
+
+Additionally, each connection ID MUST be used on packets sent from only one
+local address.  At any time, an endpoint MAY change to a new connection ID on a
+local address already in use.  An endpoint that migrates to a new local address
+SHOULD retire all connection IDs that were used on the previous address
+using the CONNECTION_ID_FINISHED frame.
```
martinthomson: This isn't quite right.  The point here is that you should retire connection IDs when you stop using the path, not when you move to a new one.  This text doesn't set us up very well for multipath, for instance.  

Also, an endpoint might decide to retain the connection ID for a period in case it has to switch back.  Think about a mobile device that hits a wifi access point.  It might keep the connection ID it used on the cellular path for a time in case it remains mobile.  Switching connection ID on each network change might be better, but maybe only if significant time passes.

draft-ietf-quic-transport.md
```diff
@@ -3257,6 +3287,49 @@ the Source Connection ID used by the peer during the initial
...
+## CONNECTION_ID_FINISHED Frame {#frame-connection-id-finished}
+
+An endpoint sends a CONNECTION_ID_FINISHED frame (type=0x1b) to indicate that it
+will no longer use a connection ID that was issued by its peer.  This also
+serves as a request to the peer to send additional connection IDs for future use
+(see {{connection-id}}).  New connection IDs can be delivered via the
+NEW_CONNECTION_ID frame ({{frame-new-connection-id}}).
+
+Retiring a connection ID using the CONNECTION_ID_FINISHED frame invalidates any
+stateless reset tokens associated with that connection ID.
```
martinthomson: This implies that there are multiple tokens.

"Retiring a connection ID using the CONNECTION_ID_FINISHED frame invalidates the stateless reset token associated with that connection ID."

martinthomson: As an open issue, we should decide how long to allow the old token to be effective.  I would suggest that we use the same 3xRTO timer for that as well.  Then we have 3xRTO everywhere in this.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
martinthomson: Yeah, @mikkelfj got it.  You can retire a connection ID before NEW_CONNECTION_ID is acknowledged even.

As for the short connection IDs, I think that's a reason not to allow shorter values more than an argument against this text.  Yes, over time, you might cycle really old IDs back into circulation such that they get re-issued to a connection that had it in the past.  That would be valid, but allowing it would complicate the protocol.  The protocol here uses the connection ID as its own identifier, so for that to work properly, we need it to be unique.

draft-ietf-quic-transport.md
```diff
@@ -3257,6 +3287,49 @@ the Source Connection ID used by the peer during the initial
...
+Connection ID:
+
+: A connection ID of the specified length.
+
+Receipt of a CONNECTION_ID_FINISHED frame containing a connection ID that was
+not previously sent to the peer MAY be treated as a connection error of type
+PROTOCOL_VIOLATION.
+
+An endpoint MUST NOT send this frame if it is currently sending packets with a
+zero-length Destination Connection ID.  Changing the length of a connection ID
```
martinthomson: It cannot send this frame because it won't have any connection IDs to put in it.  I think that we can shorten this paragraph to: "An endpoint cannot send this frame if it was provided with a zero-length connection ID by its peer.  An endpoint that provides a zero-length connection ID MUST treat receipt [...]"

draft-ietf-quic-transport.md
```diff
@@ -3257,6 +3287,49 @@ the Source Connection ID used by the peer during the initial
...
+to or from zero-length makes it difficult to identify when the value of the
+connection ID changed.  An endpoint that is receiving packets with a zero-length
+Destination Connection ID MUST treat receipt of a CONNECTION_ID_FINISHED frame
+as a connection error of type PROTOCOL_VIOLATION.
+
+
```
martinthomson: We want to point out, purely for the purpose of completeness, that the connection ID provided during the handshake can be retired with this frame.

draft-ietf-quic-transport.md
```diff
@@ -1224,41 +1225,77 @@ be routed back to the endpoint and identified by the endpoint upon receipt.
...
+connection ID for the duration of the connection or until its peer invalidates
+the connection ID via a CONNECTION_ID_FINISHED frame
+({{frame-connection-id-finished}}).
+
+Implementations SHOULD ensure that peers have sufficient connection IDs
+available to reduce the possibility of peers exhausting their supply of
+available connection IDs.  An implementation could do this by always supplying a
+new connection ID for each connection ID retired with a CONNECTION_ID_FINISHED
+frame. When a receiver of a packet notices that its peer is now using a
+previously unused connection ID, it can choose to supply its peer with a new
```
martinthomson: s/can choose to/can/

draft-ietf-quic-transport.md
erickinnear: Agreed, let's pull that text into this section?

draft-ietf-quic-transport.md
erickinnear: Sure, RETIRE_CONNECTION_ID works just as well. 😄

draft-ietf-quic-transport.md
erickinnear: Yeah, we've got text in the previous section for the issuer, we could remove all reference to that from this section.

draft-ietf-quic-transport.md
erickinnear: Good catch!

draft-ietf-quic-transport.md
martinthomson: SGTM

draft-ietf-quic-transport.md
MikeBishop: I think it's correct as written, actually, though we could make it more explicit if needed.  If a packet with the old CID was delayed and the recipient drops the CID immediately upon receipt, you'll get a SR that you ought to ignore.  So unless we're *also* adding a 3xRTO requirement on the server to continue accepting the retired CID, the sender needs to immediately stop recognizing the SRT associated with the CID.  You might miss a SR from an old packet if the race lands wrong.  But the packet being sent now has a new, still-valid CID and will presumably elicit a current SR of its own, so that's okay.

draft-ietf-quic-transport.md
MikeBishop: Ah, GitHub was hiding this comment initially.  Remembering retired CIDs for 3xRTO on both sides (accepting SRs, but not generating them) is certainly another solution -- it's more state for the endpoints, but causes Stateless Reset to tear down more quickly.

draft-ietf-quic-transport.md
huitema: In practice, that recommendation amounts to the server initiating a migration upon receiving packets from a new client address, so that a given pair of CID is tied to a single pair of addresses. That can work, but we have to consider security issues. Addresses are outside the cryptographic envelope, attackers can take good packets and just replace the address by something else. That mean attackers can easily trick servers into starting migrations.

I am not sure of the balance between aggravated DOD risk and moderate privacy gain. The beans are already spilled in the packet from the client with the changed address and the constant CID. If we do nothing, the client will notice that its own address changed upon receiving the first reply from the server. It can then chose to start a migration. That seems more robust.

draft-ietf-quic-transport.md
huitema: @martinthomson do we still have a reset risk if the client retires the stateless reset value that it received with the retired connection ID?

draft-ietf-quic-transport.md
huitema: I would actually disable the reset token asap. The only down side of doing that is to fall back to timer based termination if the connection breaks. But we probably need more text on the behavior of the reset token. Assume that a path is defined by a pair of CID. At any given point, a node will have a "main path" on which it is sending data, and a set of paths in various alternate paths in different stages of validation. Let's look at the following scenarios:

1) Node sends a normal packet, on the main path, and receives back a reset.
2) Node sends a probe with a fresh CID and new addresses and receives back a reset
3) Node sends a probe response with a fresh CID in response to a probe and receives back a reset
4) Node has parked an old path that it does not actively use and receives a reset on that path.

Do we want to treat all these scenarios the same, as in tear down the connection? Or do we want to just accept resets on the path that is actively used?

draft-ietf-quic-transport.md
huitema: We want to say whether it can or cannot...

erickinnear: Added note to this effect in 1fdbecd

draft-ietf-quic-transport.md
erickinnear: Changed this for FINISHED, works more nicely with what's in NEW now.

draft-ietf-quic-transport.md
erickinnear: Changed in 1fdbecd

draft-ietf-quic-transport.md
erickinnear: Moved and merged with the text that's here in 1fdbecd

draft-ietf-quic-transport.md
erickinnear: Reworded this to indicate you should retire when you're done with the local address (what we've been saying for "path" so far). We can explain more about this if necessary, but it seems pretty evident that you may migrate back, for example, and it's no harm if you use a new CID when migrating back (if you've got enough, of course).

draft-ietf-quic-transport.md
erickinnear: Fixed

draft-ietf-quic-transport.md
erickinnear: Changed in 325c21e

draft-ietf-quic-transport.md
erickinnear: This is done

erickinnear:
Added `max_connection_ids` TP as discussed in NYC. This allows the consumer of CIDs to put a cap on how many the server will send.

erickinnear:
The remaining issue here is the re-introduction of a sequence number, which will work as follows: 
- Each CID comes with a sequence number, every CID that you send must have a sequence number higher than any you've previously sent
- This does *not* requiring anything to happen in lockstep as before, however this solves a big edge case in that you can now forget CIDs immediately instead of waiting for 3xRTO when you retire them, since any NEW_CONNECTION_ID frame crossing paths with the retirement frame will contain a sequence number. Instead, you just remember the highest sequence number NEW_CONNECTION_ID that you've received. 

For example: 
1. Peer sends NEW_CONNECTION_ID and you use it, then immediately retire it with RETIRE_CONNECTION_ID. 
2. Ack of NEW_CONNECTION_ID gets lost.
3. RETIRE_CONNECTION_ID frame is still in flight when the peer retransmits NEW_CONNECTION_ID.

If you've got a sequence number you can keep track of the highest sequence that you've gotten and, since this one will not advance the sequence number, you can safely ignore the duplicate NEW_CONNECTION_ID frame.

draft-ietf-quic-transport.md
```diff
@@ -1220,48 +1220,38 @@ connection. An endpoint whose peer has selected a zero-length connection ID MUST
...
-An endpoint issues connection IDs to its peer for the peer to use when sending
-packets. This allows the endpoint control over the strategy used to interpret
-connection IDs of packets that it receives.  These connection IDs are
-communicated to the peer using NEW_CONNECTION_ID frames
-({{frame-new-connection-id}}).
+The initial connection ID issued by an endpoint is the Source Connection ID
+during the handshake.  Subsequent connection IDs are communicated to the peer
+using NEW_CONNECTION_ID frames ({{frame-new-connection-id}}).  While each
+endpoint independently chooses how many connection IDs to issue, a recommended
+number of outstanding connection IDs is eight.
```
martinthomson: endpoints SHOULD provide and maintain at least 8 connection IDs......

draft-ietf-quic-transport.md
```diff
@@ -1220,48 +1220,38 @@ connection. An endpoint whose peer has selected a zero-length connection ID MUST
...
-Endpoints that use connection IDs with length greater than zero could have their
-activity correlated if their peers keep using the same destination connection ID
-after migration.  An endpoint that receives a packet with a different remote
-address or destination connection ID than previously used SHOULD also switch to
-sending with a connection ID that has not previously been used.  The goal is to
-ensure absence of correlation between the pairs of client and server connection
-ID used on different paths.  To fulfill this privacy requirement, endpoints that
+An endpoint SHOULD ensure that its peer has a sufficient number of available and
+unused connection IDs.  The endpoint can do this by always supplying a new
+connection ID when a connection ID is retired by its peer or when the endpoint
```
martinthomson: Maybe move the recommendation for 8 in between the two sentences here.

draft-ietf-quic-transport.md
```diff
@@ -1220,48 +1220,38 @@ connection. An endpoint whose peer has selected a zero-length connection ID MUST
...
+receives a packet with a previously unused connection ID.
+
+
+### Consuming Connection IDs
+
+An endpoint that receives a packet with a different remote address or
+destination connection ID than previously used SHOULD also switch to sending
+with a connection ID that has not previously been used.  The goal is to ensure
+absence of correlation between the pairs of client and server connection ID used
```
martinthomson: The goal is to ensure that traffic on different paths cannot be correlated.

draft-ietf-quic-transport.md
```diff
@@ -1242,46 +1242,37 @@ the connection ID via a RETIRE_CONNECTION_ID frame
...
 unused connection IDs.  The endpoint can do this by always supplying a new
 connection ID when a connection ID is retired by its peer or when the endpoint
-receives a packet with a previously unused connection ID.
-
-
-### Consuming Connection IDs
+receives a packet with a previously unused connection ID.  Endpoints that
+initiate migration and require non-zero-length connection IDs SHOULD provide
+their peers with new connection IDs before migration, or risk the peer closing
+the connection.
```
martinthomson: closing the connection...when they run out of connection IDs.

draft-ietf-quic-transport.md
```diff
@@ -1242,46 +1242,37 @@ the connection ID via a RETIRE_CONNECTION_ID frame
...
-initiate migration and use connection IDs with length greater than zero SHOULD
-provide their peers with new connection IDs before migration.
 
-Caution:
+### Consuming and Retiring Connection IDs
 
-: If both endpoints change connection ID in response to seeing a change in
-  connection ID from their peer, then this can trigger an infinite sequence
-  of changes.
+An endpoint can arbitrarily change the connection ID it uses for a peer to
```
martinthomson: s/arbitrarily//

draft-ietf-quic-transport.md
```diff
@@ -1242,46 +1242,37 @@ the connection ID via a RETIRE_CONNECTION_ID frame
...
-connection ID for a period of time after sending the RETIRE_CONNECTION_ID
-frame, or until that frame is acknowledged.  A recommended time is three times
-the current Retransmission Timeout (RTO) interval as defined in
-{{QUIC-RECOVERY}}.  This prevents potential retransmissions of a
-NEW_CONNECTION_ID frame from overlapping with the RETIRE_CONNECTION_ID frame
-for the same connection ID.
+connection ID for a period of time after sending the RETIRE_CONNECTION_ID frame,
+or until that frame is acknowledged.  A recommended time is three times the
+current retransmission timeout (RTO) interval as defined in {{QUIC-RECOVERY}}.
+This prevents confusion from receiving retransmissions of a NEW_CONNECTION_ID
```
martinthomson: This prevents a retransmission of a NEW_CONNECTION_ID from incorrectly renewing a previously retired connection ID.

PR #1743, opened 2018-09-13
Fix typo: s/PADDDING/PADDING/

dtikhonov:


martinthomson:
It was paddded, what's the problem?

PR #1750, opened 2018-09-16
Make connection ID appear random

### LINKED ISSUE #1740: Connections ID's are not required to appear random
When sending a new connection ID it is understood that this is to prevent linkability, but it is not required that the actual ID is hard to correlate to previous ID's.

It may be tempting to use a random router prefix and a local counter in the ID.

If a client migrates the client can choose a strong new ID, but if the server responds with a weak ID in the opposite direction, that isn't very helpful. Since the client cannot influence that, servers could be coerced into using weak ID's for monitoring purposes without actually violating the protocol.

At the same time it is necessary to encode useful routing information and therefore the CID must be easy to decode.

In conclusion, at least some guidance is needed on how to construct connection ID's.

mikkelfj:
Fixes #1740

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1215,6 +1215,9 @@ distinguish it from other connections.  A connection ID can be either 0 octets
 in length, or between 4 and 18 octets (inclusive).  Connection IDs are selected
 independently in each direction.
 
+A connection ID SHOULD appear random to external observers in order to prevent
+correlation across paths and connections.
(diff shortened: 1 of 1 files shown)


martinthomson:
The appearance of randomness is not the requirement we are looking for here.  The point is that it should not be possible to correlate or predict connection IDs for the same connection using information available to entities other than the one that generates the connection ID.

martinthomson:
Thanks!

PR #1752, opened 2018-09-16
Add protection against abusive packer number sequences

### LINKED ISSUE #1751: Protection against abusive packet numbers
There is no easy way for a peer to decided when there are too many packets lost based on the packet numbers - even if the criteria is deployment specific.

The sender does not need to introduce large gaps and by enforcing that, the peer can make an informed decision on the packet number range.

There is also no error to disconnect due to suspected abuse. A too large packet number gap could be for valid reasons, even if the loss is too high to proceed. A new error that can deal with both these conditions would be useful.

mikkelfj:
Fixes #1751

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1093,6 +1093,13 @@ protection for the reasons described in Section 9.3 of {{QUIC-TLS}}. An
 efficient algorithm for duplicate suppression can be found in Section 3.4.3 of
 {{?RFC2406}}.
 
+Packet numbers SHOULD be incremented by one. Parallel processing and optimistic
+ACK attack mitigation {{optimistic-ack-attack}} can lead to gaps. The difference
+between the largets sent and the largest acknowledged packet number
+should be no more than twice the actual number of packets sent. A peer MAY close
+a connection with the error BAD_LINK when the observed loss rate based on packet
+numbers is considered too high.
+
 A Version Negotiation packet ({{packet-version}}) does not include a packet
 number.  The Retry packet ({{packet-retry}}) has special rules for populating
 the packet number field.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1093,6 +1093,14 @@ protection for the reasons described in Section 9.3 of {{QUIC-TLS}}. An
 efficient algorithm for duplicate suppression can be found in Section 3.4.3 of
 {{?RFC2406}}.
 
+Packet numbers SHOULD be incremented by one. Parallel processing and optimistic
+ACK attack mitigation {{optimistic-ack-attack}} can lead to gaps. The difference
+between the largest sent and the largest acknowledged packet number
+SHOULD be no more than twice the actual number of packets sent.  The number of
+gaps and their sizes SHOULD be kept to a minimum.  A peer MAY close a connection
+with the error BAD_LINK when the observed gap count or loss rate based on packet
```
ianswett: If we add an error code, I'd like to be specific about what it means.  ie: too many gaps?   BAD_LINK doesn't tell me if the connection is too poor to continue with or if there are too many gaps, which are potentially quite different.

mikkelfj: We could add a note about "SHOULD add a reason phrase" when using the BAD_LINK error.

marten-seemann: I'm not a fan of adding a separate error here. There are a lot of places where an implementation might introduce some limit as a DoS defense (e.g. too many gaps in STREAM data), many of those implementation specific. BAD_LINK certainly isn't appropriate in many of those cases.

mikkelfj: Which error would you use then? Or would you silently drop? I considered that as well.

nibanks: I agree with @marten-seemann here. I think if an implementation has a specific limit/threshold, it should just silently drop.

mikkelfj: The reason I chose not to silently drop was the subsequent reset that would follow.

PR #1761, opened 2018-09-18
QPACK: don't send stream cancellation when SETTINGS_HEADER_TABLE_SIZE=0

afrind:
Fixes: #1747

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -645,7 +645,7 @@ of a stream generates a Stream Cancellation instruction on the decoder stream.
 Similarly, a decoder that abandons reading of a stream needs to signal this
 using the Stream Cancellation instruction.  This signals to the encoder that all
 references to the dynamic table on that stream are no longer outstanding.  A
-decoder with a maximum dynamic table size equal to zero MAY omit send Stream
+decoder with a maximum dynamic table size equal to zero MAY omit sending Stream
 Cancellations, because the encoder cannot have any dynamic table references.
 
 An encoder cannot infer from this instruction that any updates to the dynamic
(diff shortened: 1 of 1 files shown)


mikkelfj:
Could you be more precise than "does not"? Is it valid but unnecessary when size is 0?

afrind:
That was a bit vague.  The idea is to allow an implementation that doesn't implement the dynamic table to not even bother with the QPACK control streams. I guess SHOULD NOT is OK - I don't think there is harm in sending them.  An encoder needs to be ready to handle Stream Cancellations since it doesn't know what the decoder's max will be.

afrind:
Talked to some folks in person in NYC, and the recommendation is that the decoder MAY omit stream cancellations when its max dynamic table size is 0: eg encoders need to handle it either way.

draft-ietf-quic-qpack.md
```diff
@@ -639,12 +639,14 @@ Largest Known Received.
...
-abandons reading of a stream needs to signal this using the Stream Cancellation
-instruction.  This signals to the encoder that all references to the dynamic
-table on that stream are no longer outstanding.
+A stream that is reset might have multiple outstanding header blocks with
+dynamic table references.  A decoder that receives a stream reset before the end
+of a stream generates a Stream Cancellation instruction on the decoder stream.
+Similarly, a decoder that abandons reading of a stream needs to signal this
+using the Stream Cancellation instruction.  This signals to the encoder that all
+references to the dynamic table on that stream are no longer outstanding.  A
+decoder with a maximum dynamic table size equal to zero MAY omit send Stream
```
MikeBishop: "omit sending", probably.

PR #1769, opened 2018-09-19
Permit closing frames in all packets

### LINKED ISSUE #1766: Allowed Frame Types in different Contexts
The sections on Initial and Handshake packets list some packet types that are allowed. These statements should be strengthened to use MUST and MAY, and should specify behavior if we receive disallowed packets.

The informal consensus in NYC was that CONNECTION_CLOSE frames should be allowed in Initial Packets, so the text should be extended.

martinthomson:
Closes #1766.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -703,9 +703,9 @@ server may send multiple Initial packets.  The cryptographic key exchange could
 require multiple round trips or retransmissions of this data.
 
 The payload of an Initial packet includes a CRYPTO frame (or frames) containing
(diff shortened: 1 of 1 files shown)


martinthomson:
Oh, that's a good point.  I think that makes the use of APPLICATION_CLOSE in Initial hard (Handshake is OK).  I'll adjust.

martinduke:
I don't think this completely addresses #1766. There should probably be some MUSTs or MUST NOTs. Furthermore, we should explain what to do with invalid frames in packets. I imagine we should drop silently in the case of Initial and throw PROTOCOL_VIOLATION in the other contexts.

PR #1772, opened 2018-09-19
Fix spelling error

DaanDeMeyer:
Not entirely sure about this one but I think it should be "the flow control window"


DaanDeMeyer:
Is it ok to make separate pull requests for each spelling error I find or should I fix any other spelling errors in a single pull request?

PR #1775, opened 2018-09-19
Remove length restrictions on connection IDs

### LINKED ISSUE #1570: Connection ID lengths 1, 2 and 3 bytes can't be encoded

martinthomson:
Closes #1570.

### DIFF
--- a/draft-ietf-quic-invariants.md
+++ b/draft-ietf-quic-invariants.md
@@ -189,7 +189,7 @@ version-specific semantics are marked with an X.
 +-+-+-+-+-+-+-+-+
 |0|X X X X X X X|
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                Destination Connection ID (0..144)           ...
+|                 Destination Connection ID (*)               ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |X X X X X X X X X X X X X X X X X X X X X X X X X X X X X X  ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-invariants.md
```diff
@@ -189,7 +189,7 @@ version-specific semantics are marked with an X.
 +-+-+-+-+-+-+-+-+
 |0|X X X X X X X|
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                Destination Connection ID (0..144)           ...
+|                 Destination Connection ID (*)               ...
```
mikkelfj: What about DCIL and long headers?

martinthomson: This is the short header only.  See the section header.

mikkelfj: yes, but why are short header CID's arbitrary when long headers are not? Isn't the short header CID length the same as those in the long header?

mikkelfj: OK, I see - invariants, not transport - and negotatied CIDs can change length, just not in long headers.

PR #1781, opened 2018-09-20
Add max_ack_delay transport parameter and ACK every two packets

### LINKED ISSUE #981: Explicitly communicate and include ACK delay in RTO computations

ianswett:
Fixes #981 and #1428

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1559,6 +1559,7 @@ language from Section 3 of {{!TLS13=RFC8446}}.
       disable_migration(9),
       initial_max_stream_data_bidi_remote(10),
       initial_max_stream_data_uni(11),
+      max_ack_delay(12),
       (65535)
    } TransportParameterId;
 
@@ -1673,6 +1674,12 @@ disable_migration (0x0009):
   NOT send any packets, including probing packets ({{probing}}), from a local
   address other than that used to perform the handshake.  This parameter is a
   zero-length value.
+
+max_ack_delay (0x000c):
+
+: An 8 bit unsigned integer value indicating the maximum amount of time in
+  milliseconds an ACK frame will be delayed by a receiver.  if this value is
+  absent, a default of 25 is assumed.
 
 Either peer MAY advertise an initial value for the flow control on each type of
 stream on which they might receive data.  Each of the following transport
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1674,6 +1675,12 @@ disable_migration (0x0009):
   address other than that used to perform the handshake.  This parameter is a
   zero-length value.
 
+max_ack_delay (0x000c):
+
+: An 8 bit unsigned integer value indicating the maximum amount of time in
```
marten-seemann: duration?

draft-ietf-quic-transport.md
```diff
@@ -3461,9 +3468,14 @@ ACK frames in response to other packets.  Implementations MUST NOT send more
 than one packet containing only an ACK frame per received packet that contains
 frames other than ACK and PADDING frames.  Packets containing frames besides
 ACK and PADDING MUST be acknowledged immediately or when a delayed ack timer
-expires. The delayed ack timer MUST NOT delay an ACK for longer than an RTT or
-the alarm granularity.  This ensures an ACK frame is sent at least once per RTT
-if new packets needing acknowledgement were received.
+expires.
+
+The receiver's delayed ack timer MUST NOT delay an ACK for longer than an RTT
+or the value indicated by the `max_ack_delay` transport parameter.
```
martinthomson: you say that it might exceed `max_ack_delay` in the recovery PR, so which is it?

draft-ietf-quic-transport.md
```diff
@@ -3461,9 +3468,14 @@ ACK frames in response to other packets.  Implementations MUST NOT send more
...
+This ensures an ACK frame is sent at least once per RTT if new packets
+needing acknowledgement were received and ensures the sender has a predictable
+limit on the ack delay for timer based retransmissions.  ACKs should be sent
```
martinthomson: I think we're gonna need caps on that should.

ianswett:
Thanks Martin, tweaked based on your comments.

draft-ietf-quic-transport.md
huitema: How about, "the maximum ACK delay in milliseconds"?

ianswett: I think it's helpful to clarify what ACK delay is here, so I'd like to avoid using the same words in the parameter when possible.

huitema: I agree that it would be helpful. The only reference to ACK Delay in the transport draft is in the description of the ACK frame. The ACK generation is described in the recovery draft, using a parameter kDelayedAckTimeout. Max ACK Delay is also defined in the recovery draft, but the section 4.1.1 appears obsolete. In light of that, the proper phrase is probably "maximum time that the peer will wait before acknowledging a packet" plus a reference to the appropriate section of the recovery draft. This PR should probably also update the recovery draft.

martinthomson:
I don't think that we need microsecond precision here.  It just makes expressing the value harder.

PR #1782, opened 2018-09-20
No longer calculate max_ack_delay in recovery

### LINKED ISSUE #1438: Resending ACKs increases max_ack_delay unnecessarily
Subodh notes that when a receiver simply resends old ACK frames, the ack_delay field keeps on increasing, which unnecessarily inflates max_ack_delay. From email thread:

(snip)
My concern is this situation: a "receiver" had not seen the original ACK (the packet was lost), then the sender rtos and sends a large ack delay to compensate for the rto, the max filter for max ack delay would inflate the delay for future packet. 

if latest_rtt - min_rtt  > ack_delay:

   max_ack_delay = max(max_ack_delay, ack_delay)

latest_rtt in this case would be rto_timer + rtt because the sender sends an ack after an rto. 
(snip)

ianswett:
Instead use the peer's value from transport parameters.

Fixes #1438

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -599,10 +599,9 @@ min_rtt:
 
 max_ack_delay:
 : The value from the `max_ack_delay` transport parameter, which
-  indicates the maximum amount of time the receiver intends
-  to delay acks.  The actual ack_delay in a received ACK
-  frame may be larger due to late timers, reordering, or
-  lost ACKs.
+  indicates the maximum amount of time in milliseconds the receiver
+  intends to delay acks.  The actual ack_delay in a received ACK
+  frame may be larger due to late timers, reordering, or lost ACKs.
 
 reordering_threshold:
 : The largest packet number gap between the largest acknowledged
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -598,9 +598,10 @@ min_rtt:
 : The minimum RTT seen in the connection, ignoring ack delay.
 
 max_ack_delay:
-: The maximum ack delay in an incoming ACK frame for this connection.
-  Excludes ack delays for non-retransmittable packets and those
-  that create an RTT sample less than min_rtt.
+: The value from the `max_ack_delay` transport parameter, which
+  indicates the maximum amount of time in milliseconds the receiver
```
martinthomson: the maximum amount of time by which the receiver intends to delay acknowledgments, in milliseconds

ianswett:
Thanks Martin, any other issues?

martinthomson:
Nope, merge away (the CI issue is not your fault).

mikkelfj:
wow that is some commit name 5ea1606

PR #1787, opened 2018-09-21
Clarify OOO CRYPTO data handling

martinduke:
We should never pass up Initial CRYPTO payloads once we've passed up Handshake, etc. This would break the ordered delivery of data to TLS.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3710,6 +3710,10 @@ functionally identical to STREAM frames, except that they do not bear a stream
 identifier; they are not flow controlled; and they do not carry markers for
 optional offset, optional length, and the end of the stream.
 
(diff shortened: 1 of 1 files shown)


marten-seemann:
That won’t work, e.g. when ServerHello and EncryptedExtensions are reordered.

martinduke:
If I don't have the Server Hello, I can't decrypt EE to find the CRYPTO frame. I'll change the word to "processed", which I believe addresses your concern.

marten-seemann:
That's clearer now, however I think this is already described in the TLS document in the section "Sending and Receiving Handshake Messages", isn't it?

martinduke:
You're right. I'll leave it to the editors to determine if this is useful to also mention where I've put it. If not, please close this PR.

martinthomson:
Yeah, the text in the -tls draft is much more precise.  I worry that this might be misinterpreted.

Issue #1709, opened 2018-08-28
Reconsider multiple Retry packets


martinthomson:
As others have observed, this might not be so easy to get right.

martinthomson:
Closed by #1788.

PR #1788, opened 2018-09-21
Most the Retry-related fixes

### LINKED ISSUE #1784: Handshake failure after Retry
I found that there is a corner case that handshake fails after Retry.
The current specification allows server not to change its SCID in Retry packet.  In other words, server is allowed to set client chosen CID in SCID field of Retry packet.
The spec also allows client to create fresh TLS context each time it receives Retry packet.
If they are used, there is a case that handshake fails after Retry packet.
Here is the possible scenario.

1. client sends Initial to server with TLS context T0: Initial SCID=A DCID=B
2. client resends Initial to server with TLS context T0: Initial SCID=A DCID=B
3. client receives Retry from server: Retry SCID=B DCID=A ODCID=B
4. in response to Retry(3), client sends Initial to server with TLS context T1: Initial SCID=A DCID=B
5. client receives Retry from server (in response to client Initial(2)): Retry SCID=B DCID=A ODCID=B ODCID=B
6. in response to Retry(5), client sends Initial to server with TLS context T2: Initial SCID=A DCID=B
7. client receives Initial from server (in response to client Initial(4), thus using TLS context T1): Initial SCID=C DCID=A
8. client receives Handshake from server (which uses TLS context T1): Handshake SCID=C DCID=A

client cannot decrypt Handshake packet from server because client uses TLS context T2 meanwhile server uses TLS context T1.

martinthomson:
Except the one where we will authenticate it.  That comes afterwards.

This reduces the number of Retries to 1.

This also forces the server to change its connection ID when it generates a Retry.

Closes #1784, #1709.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -593,8 +593,9 @@ The server populates the Destination Connection ID with the connection ID that
 the client included in the Source Connection ID of the Initial packet.
 
 The server includes a connection ID of its choice in the Source Connection ID
-field.  The client MUST use this connection ID in the Destination Connection ID
-of subsequent packets that it sends.
+field.  This value MUST not be equal to the Destination Connection ID field of
+the packet sent by the client.  The client MUST use this connection ID in the
+Destination Connection ID of subsequent packets that it sends.
 
 A server MAY send Retry packets in response to Initial and 0-RTT packets.  A
 server can either discard or buffer 0-RTT packets that it receives.  A server
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -596,34 +596,36 @@ The server includes a connection ID of its choice in the Source Connection ID
 field.  The client MUST use this connection ID in the Destination Connection ID
 of subsequent packets that it sends.
 
-A Retry packet does not include a packet number and cannot be explicitly
-acknowledged by a client.
+A server MAY send Retry packets in response to Initial and 0-RTT packets.  A
+server can either discard or buffer 0-RTT packets that it receives.  A server
+can send multiple Retry packets as it receives Initial or 0-RTT packets.
 
-A server MUST NOT send a Retry in response to packets other than Initial
-or 0-RTT packets.  A server MAY choose to only send Retry in response to Initial
-packets and discard or buffer 0-RTT packets corresponding to unvalidated client
-addresses.
+A client MUST accept and process at most one Retry packet for each connection
+attempt.  After the client has received and processed an Initial or Retry packet
+from the server, it MUST discard any subsequent Retry packets that it receives.
 
-If the Original Destination Connection ID field does not match the Destination
-Connection ID from the most recent Initial packet it sent, clients MUST discard
-the packet.  This prevents an off-path attacker from injecting a Retry packet.
+Clients MUST discard Retry packets that contain an Original Destination
+Connection ID field that does not match the Destination Connection ID its
+Initial packet.  This prevents an off-path attacker from injecting a Retry
+packet.
 
 The client responds to a Retry packet with an Initial packet that includes the
 provided Retry Token to continue connection establishment.
+
+A client sets the Destination Connection ID field of this Initial packet to the
+value from the Source Connection ID in the Retry packet. Changing Destination
+Connection ID also results in a change to the keys used to protect the Initial
+packet. It also sets the Token field to the token provided in the Retry. The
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -593,37 +593,40 @@ The server populates the Destination Connection ID with the connection ID that
...
-addresses.
+A client MUST accept and process at most one Retry packet for each connection
+attempt.  After the client has received and processed an Initial or Retry packet
+from the server, it MUST discard any subsequent Retry packets that it receives.
 
-If the Original Destination Connection ID field does not match the Destination
-Connection ID from the most recent Initial packet it sent, clients MUST discard
-the packet.  This prevents an off-path attacker from injecting a Retry packet.
+Clients MUST discard Retry packets that contain an Original Destination
+Connection ID field that does not match the Destination Connection ID its
```
mikkelfj: of its

draft-ietf-quic-transport.md
```diff
@@ -739,9 +742,10 @@ Source Connection IDs during the handshake.
 
 On first receiving an Initial or Retry packet from the server, the client uses
 the Source Connection ID supplied by the server as the Destination Connection ID
-for subsequent packets.  Once a client has received an Initial packet from the
-server, it MUST discard any packet it receives with a different Source
-Connection ID.
+for subsequent packets.  That means that a client might change the Destination
+Connection ID twice during connection establishment.  Once a client has received
```
mikkelfj: only twice? Is there text to prevent a server from issuing a retry when a retry token is present in initial?

martinthomson: There is nothing stopping the server from sending multiple Retry packets (regardless of which flight they came from, whether they had token, or anything).  However, the client is only going to process one of them.

mikkelfj: right

draft-ietf-quic-transport.md
```diff
@@ -593,37 +593,40 @@ The server populates the Destination Connection ID with the connection ID that
...
-acknowledged by a client.
+A server MAY send Retry packets in response to Initial and 0-RTT packets.  A
+server can either discard or buffer 0-RTT packets that it receives.  A server
+can send multiple Retry packets as it receives Initial or 0-RTT packets.
 
-A server MUST NOT send a Retry in response to packets other than Initial
-or 0-RTT packets.  A server MAY choose to only send Retry in response to Initial
-packets and discard or buffer 0-RTT packets corresponding to unvalidated client
-addresses.
+A client MUST accept and process at most one Retry packet for each connection
```
ianswett: This MUST is a bit strong for me when thinking of man-on-the-side, especially since you don't respond to cases when the original dcid doesn't match, but it's Ok for now since mots is out of scope for v1.

draft-ietf-quic-transport.md
```diff
@@ -593,37 +593,40 @@ The server populates the Destination Connection ID with the connection ID that
 the client included in the Source Connection ID of the Initial packet.
 
 The server includes a connection ID of its choice in the Source Connection ID
-field.  The client MUST use this connection ID in the Destination Connection ID
-of subsequent packets that it sends.
+field.  This value MUST not be equal to the Destination Connection ID field of
```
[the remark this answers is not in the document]
martinthomson: Yes, this isn't obvious until you read #1784.  In that case, the reuse of a connection ID causes the client to be confused about what belongs to what.

The problem @tatsuhiro-t identified is that the client can respond to a Retry, then receive a valid Initial in response to a packet sent before that Retry, but it is one that is incompatible with its current TLS state.  This assumes that the client has changed its ClientHello somehow in response to the Retry, which we encourage, but don't mandate.

Forcing the change of connection ID means that a client that receives a Retry will drop anything sent by the server in response to any Initial packets sent before receiving the Retry.  So, if a retransmission of the first round of Initial packets gets a real response, that will just be dropped.

draft-ietf-quic-transport.md
mikkelfj: At some point it was suggested that a man-on-the-side defence could be to excercise the retry, but still watch out of an initial response within a timeout period. If an initial response arrives, the retry is cancelled. But I guess you could have bogus inital responses to valid retries ...

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
ianswett: Agreed, it seems like the right solution, can you call this and the fact we're limiting to 1 Retry in the PR description, because those are fairly substantial changes?

PR #1794, opened 2018-09-25
Include a token on all Initial packets

### LINKED ISSUE #1649: Subsequent Initial Packets with Token Field
I was looking at the spec and it looks like we never explicitly indicated if subsequent Initial packets sent from the client should continue to populate the Token field. by subsequent packets, I mean responses to HRR and ACK only packets.

For the end server, I would assume it doesn't need the token for subsequent Initial packets, since it likely already has the state needed saved. But for a DDoS mitigation device in path, what is its logic supposed to be for these additional Initial packets it sees, assuming it wants to maintain a stateless design?

The DDoS mitigation device cannot really distinguish between the first Initial packet and subsequent ones. If the client always includes the token, the device would be able to statelessly verify it for every Initial packet, but that breaks if the end server also did a Retry. We could instead just assume full length (>1200 bytes) Initial packets are the first ones, and only they need to be validated. The device would just assume shorter packets should be let through untouched. I'm just not sure if that is too brittle an assumption.

martinthomson:
Closes #1649.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -621,12 +621,11 @@ packet. It also sets the Token field to the token provided in the Retry. The
 client MUST NOT change the Source Connection ID because the server could include
 the connection ID as part of its token validation logic (see {{tokens}}).
 
-The Initial packet sent by the client in response to a Retry packet is subject
-to the same restrictions as the first Initial packet, with the exception of the
-value of the Destination Connection ID and Token fields, which are set as
-described here.  A client can either reuse the cryptographic handshake message
-or construct a new one at its discretion.  All subsequent Initial packets from
+All subsequent Initial packets from the client MUST use the connection ID and
+token values from the Retry packet.  All subsequent Initial packets from
 the client MUST use the connection ID and token values from the Retry packet.
+A client can either reuse the cryptographic handshake message or construct a
+new one at its discretion.
 
 A client MAY attempt 0-RTT after receiving a Retry packet by sending 0-RTT
 packets to the connection ID provided by the server.  A client that sends
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -622,8 +622,8 @@ client MUST NOT change the Source Connection ID because the server could include
 the connection ID as part of its token validation logic (see {{tokens}}).
 
 All subsequent Initial packets from the client MUST use the connection ID and
-token values from the Retry packet.  All subsequent Initial packets from
-the client MUST use the connection ID and token values from the Retry packet.
+token values from the Retry packet.  Aside from this, the Initial packet sent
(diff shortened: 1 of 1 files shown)


mikkelfj:
What if the Intial packet with token needs to be sent again because of suspected packet loss?
(There is a lot of new text in this area, so it could well be covered).

nibanks:
I agree with @mikkelfj here. If the client's first Initial (with a token) is suspected to be lost and the contents retransmitted, the token should continue to be included, right? Additionally, is there **harm** in including the token in all client Initial packets?

draft-ietf-quic-transport.md
```diff
@@ -621,10 +621,19 @@ packet. It also sets the Token field to the token provided in the Retry. The
 client MUST NOT change the Source Connection ID because the server could include
 the connection ID as part of its token validation logic (see {{tokens}}).
 
+The Initial packet sent by the client in response to a Retry packet is subject
+to the same restrictions as the first Initial packet, with the exception of the
+value of the Destination Connection ID and Token fields, which are set as
+described here.  A client can either reuse the cryptographic handshake message
+or construct a new one at its discretion.  Any subsequent Initial packets from
+the client MUST use the same connection ID values, and MUST NOT include the
```
ianswett: I thought the Retry now must change the CID?

martinthomson:
Token on every Initial seemed wrong, but the more I think about it, the less reason I have for rejecting it.  Does anyone have a reason to remove the token?

draft-ietf-quic-transport.md
martinthomson: Yeah, I meant "same as the one from the Retry"  clarified.

draft-ietf-quic-transport.md
```diff
@@ -621,10 +621,18 @@ packet. It also sets the Token field to the token provided in the Retry. The
 client MUST NOT change the Source Connection ID because the server could include
 the connection ID as part of its token validation logic (see {{tokens}}).
 
+The Initial packet sent by the client in response to a Retry packet is subject
+to the same restrictions as the first Initial packet, with the exception of the
+value of the Destination Connection ID and Token fields, which are set as
+described here.  A client can either reuse the cryptographic handshake message
```
ianswett: How about putting the last sentence in this paragraph first, then saying "Aside from this, the Initial packet sent by the client is subject to the same restrictions as the first Initial packet."

mikkelfj:
How does packet numbers relate to this?

How about requiring initial packets to be retransmitted verbatim and perhaps only allow token in the packet number 0? An endpoint can either store the packet for retransmission or reproduce the same AEAD on retransmission using the same packet number.

ianswett:
If you retransmit verbatim, then you can't tell if the original or the retransmission was received, so you lose the first RTT sample on the client side, which is unfortunate, since that's the first congestion control or recovery related info you get on the connection.

PR #1796, opened 2018-09-25
Ack Delay and TLP

ianswett:
Fixes some remaining issues now that QUIC recommends sending an ACK every two packets, mostly involving reverting to the TCP style TLP calculation.

I wrote this based on 'two full-sized packets' equalling a fixed 2400 bytes.  QUIC could send immediate acks for non-full-sized packets, but that's not what TCP does.  I could base it on the connection's packet size, but that requires the receiver to monitor the sender's max packet size, which made me a bit concerned?

Advice on how to ensure sender and receiver agree on packet size would be appreciated.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -828,12 +828,11 @@ Pseudocode for SetLossDetectionTimer follows:
       if (tlp_count < kMaxTLPs):
         // Tail Loss Probe
         if (bytes_in_flight < 2400):
-          // Less than one full-sized packet in flight.
+          // Less than two full-sized packets in flight.
           tlp_timeout = max(1.5 * smoothed_rtt
                               + max_ack_delay, kMinTLPTimeout)
         else:
-          tlp_timeout = max(1.5 * smoothed_rtt,
-                              kMinTLPTimeout)
+          tlp_timeout = max(2 * smoothed_rtt, kMinTLPTimeout)
         timeout = min(tlp_timeout, timeout)
 
     loss_detection_timer.set(
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -439,10 +439,10 @@ ack delay to avoid causing the peer spurious timeouts.  The maximum ack delay
 is communicated in the `max_ack_delay` transport parameter and the default
 value is 25ms.
 
-An acknowledgement SHOULD be sent for every second full-sized packet, as TCP
-does {{?RFC5681}}, or may be sent less frequently, as long as the delay does
-not exceed the maximum ack delay. QUIC recovery algorithms do not assume the
-peer generates an acknowledgement immediately when receiving a second full-sized
+An acknowledgement SHOULD be sent for every second full-sized packet, as TCP does
+{{?RFC5681}}, or may be sent less frequently, as long as the delay does not
+exceed the maximum ack delay. QUIC recovery algorithms do not assume the peer
+generates an acknowledgement immediately when receiving a second full-sized
 packet.
 
 Out-of-order packets SHOULD be acknowledged more quickly, in order to accelerate
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -352,11 +352,11 @@ receiver.
 The timer duration, or Probe Timeout (PTO), is set based on the following
 conditions:
 
-* If less than 2400 bytes are in flight, PTO SHOULD be scheduled for
-  max(1.5*SRTT+MaxAckDelay, kMinTLPTimeout).
+* If less than 2*kMaxDatagramSize bytes are in flight, PTO SHOULD be
+  scheduled for max(1.5*SRTT+MaxAckDelay, kMinTLPTimeout).
 
-* If at least 2400 bytes are in flight, PTO SHOULD be scheduled for
-  max(2*SRTT, kMinTLPTimeout).
+* If at least 2*kMaxDatagramSize bytes are in flight, PTO SHOULD be
+  scheduled for max(2*SRTT, kMinTLPTimeout).
 
 * If RTO ({{rto}}) is earlier, schedule a TLP in its place. That is,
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -834,8 +827,13 @@ Pseudocode for SetLossDetectionTimer follows:
       timeout = timeout * (2 ^ rto_count)
       if (tlp_count < kMaxTLPs):
         // Tail Loss Probe
-        tlp_timeout = max(1.5 * smoothed_rtt
-                           + max_ack_delay, kMinTLPTimeout)
+        if (bytes_in_flight < 2400):
+          // Less than one full-sized packet in flight.
+          tlp_timeout = max(1.5 * smoothed_rtt
```
nibanks: Should the comment here be `Less than two full-sized packets in flight`?

draft-ietf-quic-recovery.md
```diff
@@ -834,8 +827,13 @@ Pseudocode for SetLossDetectionTimer follows:
...
+                              + max_ack_delay, kMinTLPTimeout)
+        else:
+          tlp_timeout = max(1.5 * smoothed_rtt,
```
nibanks: I believe the `1.5` here should be `2` according to the text you have above.

draft-ietf-quic-recovery.md
```diff
@@ -365,7 +352,11 @@ receiver.
 The timer duration, or Probe Timeout (PTO), is set based on the following
 conditions:
 
-* PTO SHOULD be scheduled for max(1.5*SRTT+MaxAckDelay, kMinTLPTimeout)
+* If less than 2400 bytes are in flight, PTO SHOULD be scheduled for
```
nibanks: Having this somewhat magic number here just feels weird. This is the number of bytes 'we, locally' have in flight right? So don't we know our current sending MTU? Can't it just be two times that?

ianswett: Yes, but the receiver needs to know the current.  Should I reference kMaxDatagramSize and call that close enough?  I'm fairly concerned that if that changes over the course of the connection, the sender and receiver will not act predictably.

nibanks: So, is the receiver acknowledging based on byte count or packet count? I thought it was packet count. If so, then it seems to reason, the sender should do the same here.

ianswett: My understanding is that TCP waits for 2 full-sized packets, so I was trying to specify something similar to that in bytes.  The issue here becomes what constitutes a full-sized packet if the packet size changes?  Maybe keeping a max received packet size is reasonable?

We can add a packet count, but it's yet another variable to track and I don't think receiving two tiny packets would cause an immediate ACK in TCP, so I wouldn't expect it to in QUIC.

rpaulo:
Probably unrelated to this PR but should kMaxDatagramSize really be a constant?  If a sender has performed PMTUD then it could be higher or lower.  If we change it to a non-constant then it would also take into account max_packet_size from the TPs.

ianswett:
That is definitely a potential issue with this PR.

The nice aspect of calculating this in bytes is it's already something the sender has to track.  However, I suspect a slightly more robust approach is for the sender to count number of 'full-sized'(>=1200 byte) packets with a separate counter?  I'm happy to rewrite the PR indicating that, but would like to hear some input from you and others.

rpaulo:
Ian, what did you have in mind?  I wasn't suggesting to count full sized packets, just to allow a constant to be variable.

ianswett:
I realize kMaxDatagramSize is in the constants section, but I'm not sure it should be a constant throughout a connection, given the existence of PMTU, so it's inclusion there may be incorrect or at least misleading.

I was thinking of tracking the number of in flight packets that are >= 1200 bytes on the sender side in addition to the number of bytes in flight and then on the receiver side, only send an ack immediately if two packets >1200 bytes are received.  That would keep the concept of full-sized packets while also allow changing the MTU during a connection without any other issues.

As I mention it, I realize this has the unintended side effect that if the sender wants to delay acks, it can send a 1199 byte packet instead of 1200 byte one.  I'm not sure if that's a feature or a bug.

draft-ietf-quic-recovery.md
```diff
@@ -446,14 +433,15 @@ flight, since this packet adds network load without establishing packet loss.
...
-exceed the maximum ack delay.  QUIC recovery algorithms do not assume the peer
+excessively delay acknowledgements of packets containing frames other than ACK.
+Specifically, implementations MUST attempt to enforce a maximum
+ack delay to avoid causing the peer spurious timeouts.  The maximum ack delay
+is communicated in the `max_ack_delay` transport parameter and the default
+value is 25ms.
+
+An acknowledgement SHOULD be sent immediately upon receiving two or more
+retransmitable full-sized packets, as TCP does {{?RFC5681}}.  QUIC recovery
+algorithms, including tail loss probe{{tail-loss-probe}}, assume the peer
```
mikkelfj: How do you define full-sized packets when this depends on PMTU discovery?
The text also isn't very clear about receiving drips of full sized packets: can you ack after only one full packet if it exceeds max ACK delay? I'd say obviously yes, but I'm sure the text is very clear (before or ofter this change).

draft-ietf-quic-recovery.md
```diff
@@ -446,16 +433,16 @@ flight, since this packet adds network load without establishing packet loss.
...
-
-An acknowledgement MAY be sent for every second full-sized packet, as TCP does
-{{?RFC5681}}, or may be sent less frequently, as long as the delay does not
-exceed the maximum ack delay.  QUIC recovery algorithms do not assume the peer
-generates an acknowledgement immediately when receiving a second full-sized
-packet.
+excessively delay acknowledgements of packets containing frames other than ACK.
+Specifically, implementations MUST attempt to enforce a maximum
+ack delay to avoid causing the peer spurious timeouts.  The maximum ack delay
+is communicated in the `max_ack_delay` transport parameter and the default
```
janaiyengar: s/default/RECOMMENDED

PR #1798, opened 2018-09-25
Remove EOED v2

martinthomson:
Just so we have options.

As suggested by @kazuho.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -322,7 +322,7 @@ the connection can usually appear at any encryption level, whereas those
 associated with transferring data can only appear in the 0-RTT and 1-RTT
 encryption levels
 
-- CRYPTO frames MAY appear in packets of any encryption level.
+- CRYPTO frames MAY appear in packets of any encryption level except 0-RTT.
 
 - CONNECTION_CLOSE and APPLICATION_CLOSE MAY appear in packets of any encryption
   level other than 0-RTT.
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1115,12 +1115,15 @@ transport is not QUIC.
 
 ## Removing the EndOfEarlyData Message {#remove-eoed}
 
-Negotiating the quic_transport_parameters extension also modifies TLS to remove
-the EndOfEarlyData message.  QUIC does not rely on this message to mark the end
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -656,13 +655,14 @@ all the acknowledgements necessary to reach the same state.
...
 expected CRYPTO frames received, and all the corresponding acknowledgments have
-been received or sent, an endpoint starts a timer.  To limit the effect of
-packet loss around a change in keys, endpoints MUST retain packet protection
-keys for that encryption level for at least three times the current
-Retransmission Timeout (RTO) interval as defined in {{QUIC-RECOVERY}}.
-Retaining keys for this interval allows packets containing CRYPTO or ACK frames
-at that encryption level to be sent if packets are determined to be lost or new
-packets require acknowledgment.
+been received or sent, an endpoint starts a timer.  For 0-RTT keys, which do not
+carry CRYPTO frames, this timer starts when the first packets protected with
```
ianswett: The 0-RTT timer can be much more aggressive than 3 * RTO, since nothing now ever needs to be retransmitted at 0-RTT, it can always be sent with 1-RTT keys instead.

draft-ietf-quic-tls.md
```diff
@@ -656,13 +655,14 @@ all the acknowledgements necessary to reach the same state.
...
+1-RTT are sent or received.  To limit the effect of packet loss around a change
+in keys, endpoints MUST retain packet protection keys for that encryption level
+for at least three times the current Retransmission Timeout (RTO) interval as
+defined in {{QUIC-RECOVERY}}.  Retaining keys for this interval allows packets
```
ianswett: I think the key goal here is allowing retransmitted CRYPTO data to be acknowledged with the correct encryption level?

draft-ietf-quic-tls.md
martinthomson: I don't think that we need to special case that - as we agreed, there are optimizations that are possible, but we don't need to describe them.  The trick here is that the sender/client might stop using 0-RTT, but the receiver/server can't as easily.

draft-ietf-quic-tls.md
martinthomson: That's part of it, but the reordering thing matters more for key updates.  I'd prefer to fix this separately, if that's OK.

PR #1800, opened 2018-09-26
Minor nits

martinduke:
Added a few explicit statements about how TCP-TLS and QUIC-TLS differ. Synced with the transport draft about which frames are allowed in which contexts.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -161,6 +161,9 @@ series of typed TLS records. Records are individually cryptographically
 protected and then transmitted over a reliable transport (typically TCP) which
 provides sequencing and guaranteed delivery.
 
+Note that Change Cipher Spec records, optional in TLS over TCP, have no
+encoding for TLS over QUIC.
+
 The TLS authenticated key exchange occurs between two entities: client and
 server.  The client initiates the exchange and the server responds.  If the key
 exchange completes successfully, both client and server will agree on a secret.
@@ -259,8 +262,8 @@ QUIC also relies on TLS 1.3 for authentication and negotiation of parameters
 that are critical to security and performance.
 
 Rather than a strict layering, these two protocols are co-dependent: QUIC uses
-the TLS handshake; TLS uses the reliability and ordered delivery provided by
-QUIC streams.
+the TLS handshake; TLS uses the reliability, ordered delivery, and record
+layer provided by QUIC.
 
 At a high level, there are two main interactions between the TLS and QUIC
 components:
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -387,6 +392,10 @@ peer in CRYPTO frames. When TLS provides handshake octets to be sent, they are
 appended to the current flow and any packet that includes the CRYPTO frame is
 protected using keys from the corresponding encryption level.
 
+Unlike its operations with TCP, the TLS bytestream is never separately
+encrypted and decrypted. This operation only occurs on the QUIC packet as a
+whole.
```
martinthomson: I know what you are saying, but it's equally confusing because it's not clear what the TLS bytestream is.  For QUIC, it's the sequence of TLS handshake messages, but TLS has other concepts that could be considered to be a bytestream: the sequence of TLS records (unencrypted), or the sequence of protected TLS records could very much be considered to be a bytestream.

Maybe the right thing to say here is that "QUIC takes the unprotected content of TLS handshake records as the content of CRYPTO frames.  Encapsulation of these messages in protected TLS records does not happen.  QUIC assembles CRYPTO frames into QUIC packets, which are protected using QUIC packet protection."

But isn't that written down elsewhere?

draft-ietf-quic-tls.md
```diff
@@ -320,14 +322,17 @@ Each encryption level has a specific list of frames which may appear in it. The
 rules here generalize those of TLS, in that frames associated with establishing
 the connection can usually appear at any encryption level, whereas those
 associated with transferring data can only appear in the 0-RTT and 1-RTT
-encryption levels
+encryption levels.
```
martinthomson: I think you want a colon:

draft-ietf-quic-tls.md
```diff
@@ -161,6 +161,9 @@ series of typed TLS records. Records are individually cryptographically
 protected and then transmitted over a reliable transport (typically TCP) which
 provides sequencing and guaranteed delivery.
 
+Note that Change Cipher Spec records, optional in TLS over TCP, have no
+encoding for TLS over QUIC.
```
martinthomson: I would say simply "Change Cipher Spec records cannot be sent in QUIC."

draft-ietf-quic-tls.md
martinduke: I have spent some time looking through the quic-tls draft and cannot find it. That doesn't mean it's not there, but I'll use your language.

I am really looking for a clear statement that says "turn off TLS handshake encryption, dummy".

martinthomson:
Thanks, this is a nice improvement (and you are right, that bit about not using TLS record protection really was missing).

PR #1807, opened 2018-09-27
Simplify SETTINGS payload

### LINKED ISSUE #1556: HTTP SETTINGS: define setting content encoding
The current draft says:
> The payload of a SETTINGS frame consists of zero or more parameters,
   each consisting of an unsigned 16-bit setting identifier and a
   length-prefixed binary value.

This leaves the encoding of the setting content to the implementor and may be cause of ambiguity.
I believe we should define the encoding more explicitly for each setting, e.g. 
is `SETTINGS_NUM_PLACEHOLDERS` encoded as a fixed-length 4 or 2 (given its maximum value) bytes integer, or perhaps it can also be variable-length, e.g. can the value be encoded in just 1 byte if less than 256 ? 

Also, do we expect arbitrarily large setting contents ?
Implementors would likely enforce some sensible maximum values whenever the length is encoded as a quic-integer, do you think there is value in the RFC suggesting these maximum values ? 
This is probably a larger scope question, but in the settings frame seems more prominent

MikeBishop:
Fixes #1556 by making all setting values varints and removing length prefixes.  This has two ramifications, which are probably livable:
- An extension which needs a structured configuration message needs to define an extra frame/stream type to carry that message.
- An extension which used the full 32-bit space in HTTP/2 will spill over to a 64-bit varint, because in four bytes we only carry 30 bits of payload.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1469,6 +1469,12 @@ SETTINGS_MAX_FRAME_SIZE:
 SETTINGS_MAX_HEADER_LIST_SIZE:
 : See {{settings-parameters}}.
 
+In HTTP/QUIC, setting values are variable-length integers (6, 14, 30, or 62 bits
+long) rather than fixed-length 32-bit fields.  This will often produce a shorter
+encoding, but will produce a longer encoding for settings which use the full
+32-bit space.  Settings ported from HTTP/2 might choose to redefine the format
+of their settings to avoid using the 62-bit encoding.
+
 Settings need to be defined separately for HTTP/2 and HTTP/QUIC.  The IDs of
 settings defined in {{!RFC7540}} have been reserved for simplicity.  See
 {{iana-settings}}.
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1470,10 +1470,10 @@ SETTINGS_MAX_HEADER_LIST_SIZE:
 : See {{settings-parameters}}.
 
 In HTTP/QUIC, setting values are variable-length integers (6, 14, 30, or 62 bits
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -1478,6 +1469,12 @@ SETTINGS_MAX_FRAME_SIZE:
 SETTINGS_MAX_HEADER_LIST_SIZE:
 : See {{settings-parameters}}.
 
+In HTTP/QUIC, setting values are variable-length integers (6, 14, 30, or 62 bits
+long) rather than fixed-length 32-bit fields.  This will often produce a shorter
+encoding, but will produce a longer encoding for settings which use the full
+32-bit space.  Settings ported from HTTP/2 might choose to redefine the format
+of their settings to avoid using the 62-bit encoding.
```
ianswett: I'm not sure I understand the purpose of this additional text?  You lose 2 bits on 32 bit fields, but do we really expect that will require changes on anyone's part?

MikeBishop: Secondary Certs currently stuffs an exporter in the setting value (to confirm both peers are on the same TLS connection with no MitM), and would overflow to an eight-byte field.  It doesn't *require* changes, but it's advisable to either take advantage of the entire eight bytes or trim two bits off to still fit in four bytes.  Hence, "might choose."

ianswett: In that case, how about starting the paragraph with 'Setting values are ...... rather than fixed length 32-bit fields as in HTTP/2."  Currently you mention HTTP/QUIC in the HTTP/QUIC draft, and don't reference the thing you're comparing against(HTTP/2).

PR #1808, opened 2018-09-27
Error for missing SETTINGS frame

### LINKED ISSUE #1697: Error for not receiving SETTINGS frame as the first frame on a control stream
The paragraph says the a SETTING frame must be the first frame on a control stream but it does not specifies which error should be sent if that is not the case.

>  A SETTINGS frame MUST be sent as the first frame of either control stream (see {{stream-mapping}}) by each peer, and MUST NOT be sent subsequently or on any other stream. If an endpoint receives an SETTINGS frame on a different stream, the endpoint MUST respond with a connection error of type HTTP_WRONG_STREAM. If an endpoint receives a second SETTINGS frame, the endpoint MUST respond with a connection error of type HTTP_MALFORMED_FRAME.

MikeBishop:
Fixes #1697.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -556,9 +556,11 @@ The control stream is indicated by a stream type of `0x43` (ASCII 'C').  Data on
 this stream consists of HTTP/QUIC frames, as defined in {{frames}}.
 
 Each side MUST initiate a single control stream at the beginning of the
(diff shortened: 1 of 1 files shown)


martinthomson:
YAEC

mikkelfj:
Not quite as esoteric I'm afraid. I'd suggest Yet Another Error Code. That, or Yeti Animals Exist Covertly.

PR #1811, opened 2018-09-28
Extension frames and congestion/flow control

### LINKED ISSUE #1487: Guidance for flow control and ACK of extension frames

ianswett:
Fixes #1487

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1199,10 +1199,10 @@ that a peer is able to understand the frame.  An endpoint can use a transport
 parameter to signal its willingness to receive one or more extension frame types
 with the one transport parameter.
 
-Extension frames MUST be contestion controlled and MUST cause an ACK frame to
+Extension frames MUST be congestion controlled and MUST cause an ACK frame to
 be sent.  The exception is extension frames that replace or supplement the ACK
-frame.  Extension frames are not included in flow control unless specifically
-specified in the extension.
+frame.  Extension frames are not included in flow control unless specified
+in the extension.
 
 An IANA registry is used to manage the assignment of frame types, see
 {{iana-frames}}.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1199,6 +1199,11 @@ that a peer is able to understand the frame.  An endpoint can use a transport
 parameter to signal its willingness to receive one or more extension frame types
 with the one transport parameter.
 
+Extension frames MUST be contestion controlled and MUST cause an ACK frame to
+be sent.  The exception is extension frames that replace or supplement the ACK
+frame.  Extension frames are not included in flow control unless specifically
+specified in the extension.
+
 An IANA registry is used to manage the assignment of frame types, see
 {{iana-frames}}.
 


draft-ietf-quic-transport.md
```diff
@@ -1199,6 +1199,11 @@ that a peer is able to understand the frame.  An endpoint can use a transport
 parameter to signal its willingness to receive one or more extension frame types
 with the one transport parameter.
 
+Extension frames MUST be contestion controlled and MUST cause an ACK frame to
```
mikkelfj: congestion

draft-ietf-quic-transport.md
```diff
@@ -1199,6 +1199,11 @@ that a peer is able to understand the frame.  An endpoint can use a transport
...
+be sent.  The exception is extension frames that replace or supplement the ACK
+frame.  Extension frames are not included in flow control unless specifically
+specified in the extension.
```
LPardue: Specifically specified can probably be reduced to just specified

rpaulo:
The requirement that it MUST be ACK'ed seems too strong.  Doesn't that prevent us from doing a completely unreliable stream?  I am aware that the current DATAGRAM proposals require ACKs but I don't agree with this PR.

ianswett:
How would it prevent an unreliable stream?

The reason you don't acknowledge ACKs is that it can cause the connection to never go idle, or to take a long time to do so.    Additionally, ACK only packets don't count towards congestion control.

I don't want non-ACK frames to be excluded from congestion control unless doing so could cause a connection deadlock or some other major issue.  If a frame is congestion controlled, it should cause an ACK to be sent at some point(ie: at least once per RTT, but probably more), otherwise the congestion controller never gets any feedback.

Allowing a reduced rate of ACKs for frames which are not reliable is sensible, and is what the DATAGRAM proposal recommends.

rpaulo:
Yes, I'm not talking about ACKs for ACKs.

I believe there are two cases for unreliable streams:

1) Send and forget about the data itself but save metadata in memory: this is the case you're talking about and it's used to grow the cwnd when an ACK arrives.  It increases  bytes_in_flight.

2) Send and forget about everything: this is the more basic approach where sending doesn't elicit an ACK. I'm not sure if it should increase bytes_in_flight, but it should respect cwnd.  

Obviously 1) is better and I also prefer it for DATAGRAM streams.

I'm just wondering if the text isn't too restrictive since we're really talking about an extension.  Imagine an extension that improves the ACK frame but doesn't replace it (ACK_ECN comeback? :-)).

It's true that implementations can break these rules and implement private extensions that don't require ACKs.

P.S.: It looks like PADDING increases bytes_in_flight. This is confusing because bytes_in_flight will increase but nothing prevents a QUIC stack from just sending PADDING in one packet which would never be ACK'ed and bytes_in_flight won't ever be zero again. This would put us in perpetual loss detection mode.  Does that make sense?

MikeBishop:
To your PS, I think the lack of clarity is the difference between "gets ACK'd eventually" and "elicits an ACK".  If you receive a packet with only PADDING and ACK, that doesn't elicit an ACK immediately.  However, when you do eventually sent an ACK, that packet is included in the range of acknowledged packets, which enabled bytes_in_flight to return to zero.

ianswett:
I'm of the opinion that even though QUIC can enable 2), it's neither a typical use case nor the one we should encourage, since I don't think it's feasible to respect CWND if you don't increase bytes_in_flight.

There's a specific exception in the text for ACK frame replacements and supplements:  "The exception is extension frames that replace or supplement the ACK frame."

PADDING is weird because people requested it behave as it does.  Admittedly, if it didn't have it's current behavior of increasing bytes_in_flight without eliciting an ACK, it would be identical to the PING frame.  It's certainly a potential footgun and there's some text in recovery noting that, though it could be a little stronger: "Note that this means PADDING frames cause packets to contribute toward bytes in flight without directly causing an acknowledgment to be sent.

mikkelfj:
Can you deadlock on sending PADDING between endpoints? Perhaps a very slow death?

ianswett:
I'm not aware of any deadlocks, but you could end up stalled until the TLP fires.

rpaulo:
Oh, yes, it already says "supplements".  I've thought about this more and yes I agree it's fine the way it is right now.  Thanks.

PR #1819, opened 2018-10-01
Let Endpoints Ignore invalid Initial Packets

martinduke:
A little more flexibility to defeat injection attacks. This is the other half of #1786.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -737,8 +737,8 @@ subsequent to the first do not need to fit within a single UDP datagram.
 
 The contents of some Initial packets may, according to this specification, force
 connection termination. For example, they might contain forbidden frame types
-or a CONNECTION_CLOSE frame. As Initial packets are not protected, these might
-be injection attacks to terminate the connection.
+or a CONNECTION_CLOSE frame. As Initial packets are not protected, these could
+indicate injection attacks to terminate the connection.
 
 Endpoints MAY treat the receipt of such packets as a connection error, drop them
 without further processing, or wait for a short interval to see if a valid
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -718,9 +718,8 @@ A server sends its first Initial packet in response to a client Initial.  A
 server may send multiple Initial packets.  The cryptographic key exchange could
 require multiple round trips or retransmissions of this data.
 
-The payload of an Initial packet includes a CRYPTO frame (or frames) containing
-a cryptographic handshake message, ACK frames, or both.  PADDING and
-CONNECTION_CLOSE frames are also permitted.
+The payload of an Initial packet includes CRYPTO framed, ACK frames, or both.
+PADDING and CONNECTION_CLOSE frames are also permitted.
 
 The first packet sent by a client always includes a CRYPTO frame that contains
 the entirety of the first cryptographic handshake message.  This packet, and the
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -718,7 +718,7 @@ A server sends its first Initial packet in response to a client Initial.  A
 server may send multiple Initial packets.  The cryptographic key exchange could
 require multiple round trips or retransmissions of this data.
 
-The payload of an Initial packet includes CRYPTO framed, ACK frames, or both.
+The payload of an Initial packet includes CRYPTO frames, ACK frames, or both.
 PADDING and CONNECTION_CLOSE frames are also permitted.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -720,9 +720,7 @@ require multiple round trips or retransmissions of this data.
 
 The payload of an Initial packet includes a CRYPTO frame (or frames) containing
```
ianswett: I think "containing a cryptographic handshake message" section makes this sentence more confusing.  Can you remove it?

draft-ietf-quic-transport.md
```diff
@@ -736,6 +733,18 @@ and will contain a CRYPTO frame with an offset matching the size of the CRYPTO
 frame sent in the first Initial packet.  Cryptographic handshake messages
 subsequent to the first do not need to fit within a single UDP datagram.
 
+### Handling of Fatal Initial Packets
+
+The contents of some Initial packets may, according to this specification, force
```
MikeBishop: I'm not a big fan of a specification saying you can ignore the specification; I'd prefer consistency.  You removed the language above indicating that things should be fatal, so you're part-way there.

Perhaps something like "Receipt of an invalid Initial packet could indicate an injection attack rather than a peer failure.  Endpoints abort a connection attempt if no valid Initial packets are received after an appropriate timeout."  That gives enough latitude that the timeout could be immediate or a long observation for multiple possibilities.

martinduke: I really don't want to go everywhere it says we MUST terminate the connection and then add "unless it's an Initial and ..." Inevitably it's a bit clunky.

Your language is fine, but I'd like to preserve the option to just drop the packet without running any timer. It's a simpler course of action for the endpoint and I don't see a significant downside.

draft-ietf-quic-transport.md
```diff
@@ -736,6 +733,18 @@ and will contain a CRYPTO frame with an offset matching the size of the CRYPTO
...
+connection termination. For example, they might contain forbidden frame types
+or a CONNECTION_CLOSE frame. As Initial packets are not protected, these could
+indicate injection attacks to terminate the connection.
+
+Endpoints MAY treat the receipt of such packets as a connection error, drop them
```
ianswett: The more I think about this, I think this is basically an intractable problem unless one is willing to keep copies of the QUIC handshake state machinery.

1) If you start processing a packet, and process one frame, then the next frame is invalid, you'd have to rewind/discard the state machinery.  Is that what you're recommending?
2) You have text that indicates you should process Initial packets, but not treat them as errors once you have received a Handshake packet, but similar to my first concern, I think that's fairly impractical unless you keep copies of the handshake.

To be clear, I sort of hate implicit ACKs of Initial, but we keep making decisions, such as that only Handshake packets count as address confirmation, that make anything sent in Initial largely useless.  And now I have https://github.com/quicwg/base-drafts/pull/1862 to fix the rest of the deadlock case, but that would be fixed if we didn't send ACKs in Initial and relied upon implicit ACKs instead.

mikkelfj: So why is there such resistance against implicit ACKs in handshake, when the solve a concrete problem, as opposed to a being a whacky optimization?

ianswett: I believe we've also said previously we don't care about man on the side attacks, but I think it's clear we do, but we don't want to do a full analysis of them, which leaves us in a weird place.

martinduke: There is nothing you can really do about injection attacks during the Initial phase of the handshake. This PR is more about the 3RTT (or, for lazy implementations, infinite) time that the initial keys remain lying around for use. This radically increases the time where a connection is vulnerable (especially if the server uses the client-generated CID, so that Initial keys are easy to generate).

In that context, I'm not sure how you would meaningfully process a late-arriving Initial Packet. Say the attacker sends the server Initial Packet #2 well after the handshake, with ACK, CRYPTO, and CONNECTION_CLOSE frames. Since Initial exchange is actually done, there are no outstanding packets to ACK. Since I've already started handshake CRYPTO exchange, it's too late to deliver any new CRYPTO data. So it is basically impossible to use frames to alter connection state except to bring down the connection by generating an error.

If your code sends off an ACK before processing the frames, that is a problem, but that seems like a bad thing to do for other reasons.

martinduke:
I would argue that this PR doesn't specify anything, but merely authorizes those advanced techniques should someone elect to implement them. As indicated above, this is also about reducing the vulnerability when keeping Initial keys around for a conservatively long time.

But if this is the consensus, we can close the PR. Especially if I'm the only one interested in using this MAY in v1.

martinthomson:
Closing as agreed.

PR #1820, opened 2018-10-02
Drop HTTP settings' maxima -- they serve no purpose

dtikhonov:
The [lack of responses on the mailing list](https://www.ietf.org/mail-archive/web/quic/current/msg04758.html) convinced me that the maximum values serve no purpose.


MikeBishop:
The original intent was to bound them when the settings were arbitrary-length.  I agree that they're unneeded now that they're integers.

PR #1822, opened 2018-10-02
Track smallest_newly_acked to prevent spurious RTO

### LINKED ISSUE #1774: OnRetransmissionTimeoutVerified issue
Issue for what I described to @ianswett at dinner yesterday.

When `OnRetransmissionTimeoutVerified` is called from `OnPacketAcked` for the highest ACK in a block, all earlier packets are marked as lost, even if they are contained in that same block. That seems suboptimal.

ianswett:
Fixes #1774

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -530,6 +530,9 @@ kDelayedAckTimeout:
 kInitialRtt:
 : The RTT used before an RTT sample is taken. The RECOMMENDED value is 100ms.
 
+kInvalidPacketNumber:
+: An invalid packet number.  May be any packet number that is never sent.
+
 ### Variables of interest
 
 Variables required to implement the congestion control mechanisms
@@ -691,13 +694,13 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
       latest_rtt = now - sent_packets[ack.largest_acked].time
       UpdateRtt(latest_rtt, ack.ack_delay)
     // Find all newly acked packets and track the smallest
-    smallest_newly_acked = 2^62
+    smallest_newly_acked = kInvalidPacketNumber
     for acked_packet in DetermineNewlyAckedPackets():
       smallest_newly_acked =
         min(smallest_newly_acked, acked_packet.packet_number)
       OnPacketAcked(acked_packet.packet_number)
 
-    if smallest_newly_acked != 2^62
+    if smallest_newly_acked != kInvalidPacketNumber
       // If any packets sent prior to the RTO were acked, then
       // the RTO was spurious.  Otherwise, inform congestion control.
       if (rto_count > 0 &&
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -701,8 +701,8 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
       OnPacketAcked(acked_packet.packet_number)
 
     if smallest_newly_acked != kInvalidPacketNumber
-      // If any packets sent prior to the RTO were acked, then
-      // the RTO was spurious.  Otherwise, inform congestion control.
+      // If any packets sent prior to RTO were acked, then the
+      // RTO was spurious. Otherwise, inform congestion control.
       if (rto_count > 0 &&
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -690,9 +690,22 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
...
       UpdateRtt(latest_rtt, ack.ack_delay)
-    // Find all newly acked packets.
+    // Find all newly acked packets and track the smallest
+    smallest_newly_acked = 2^62
     for acked_packet in DetermineNewlyAckedPackets():
+      smallest_newly_acked =
+        min(smallest_newly_acked, acked_packet.packet_number)
       OnPacketAcked(acked_packet.packet_number)
+    
+    if smallest_newly_acked != 2^62
```
marten-seemann: I still think we should make 0 the sentinel value.

draft-ietf-quic-recovery.md
```diff
@@ -690,9 +690,22 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
     if (sent_packets[ack.largest_acked]):
       latest_rtt = now - sent_packets[ack.largest_acked].time
       UpdateRtt(latest_rtt, ack.ack_delay)
-    // Find all newly acked packets.
+    // Find all newly acked packets and track the smallest
+    smallest_newly_acked = 2^62
     for acked_packet in DetermineNewlyAckedPackets():
```
marten-seemann: Not an issue of this PR, but I noticed that `DetermineNewlyAckedPackets` is not defined. Do we need to fix that?

ianswett: I can comment what it does more clearly, but I don't want to define it.  There are multiple different ways to determine newly acked packets based on one's data structure choices, so I think it's very implementation dependent.

draft-ietf-quic-recovery.md
ianswett: I was going to, but 0 is a valid packet number.

marten-seemann: My point is that it shouldn't be, in order to avoid ugly code like this here, and at many other places.

ianswett: I'm not disagreeing, but I don't think this is the PR to change that?

ianswett:
Changed to a constant for readability.

draft-ietf-quic-recovery.md
```diff
@@ -690,10 +695,23 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
     if (sent_packets[ack.largest_acked]):
       latest_rtt = now - sent_packets[ack.largest_acked].time
       UpdateRtt(latest_rtt, ack.ack_delay)
-    // Find all newly acked packets.
+    // Find all newly acked packets and track the smallest
+    smallest_newly_acked = kInvalidPacketNumber
     for acked_packet in DetermineNewlyAckedPackets():
+      smallest_newly_acked =
+        min(smallest_newly_acked, acked_packet.packet_number)
```
janaiyengar: This assumes an ordering in the processing of the acked packets data structure. I would make this simpler:

// Find all newly acked packets in this ACK frame
newly_acked = DetermineNewlyAckedPackets(ack)

// Find the smallest packet that is newly acked in this ACK frame.
smallest_newly_acked = FindSmallestNewlyAcked(newly_acked)

You don't have to define either function, but add a comment that describes what they do.

draft-ietf-quic-recovery.md
```diff
@@ -530,6 +532,9 @@ kDelayedAckTimeout:
 kInitialRtt:
 : The RTT used before an RTT sample is taken. The RECOMMENDED value is 100ms.
 
+kInvalidPacketNumber:
+: An invalid packet number.  Any number larger than the largest packet number.
```
janaiyengar: not needed if you use my suggestion below.

draft-ietf-quic-recovery.md
ianswett: I don't believe it does assume ordering, that's why I'm using a min, but your suggestion works as well.

ianswett:
Pseudocode updated, PTAL

PR #1824, opened 2018-10-02
Handshake deadlock

### LINKED ISSUE #1764: Handshake loss recovery interacts poorly with amplification attack defense
Section 4.7 of the transport draft states:

>    Servers MUST NOT send more than three datagrams including Initial and
>    Handshake packets without receiving a packet from a verified source
>    address.  Source addresses can be verified through an address
>    validation token (delivered via a Retry packet or a NEW_TOKEN frame)
>    or by receiving any message from the client encrypted using the
>    Handshake keys.

As written, there are two scenarios where no one can transmit. If the client successfully sends the initial ACK to the server, and *either* the server's handshake flight or the client's handshake acks are lost, then (1) the client has nothing to do, and (2) the server does not have the flow control credit to retransmit the lost packets.

It would be nice to minimize the resource and amplification implications for the server. Ekr proposed the following conditions that apply when the path has not been verified:
1) The client initial must be padded to 1200 bytes, as today.
2) The client must retransmit its initial until it has a Handshake Crypto frame to send (i.e. it has received the entire server's first flight.)
3) Until the server gets an hs pkt (the path has been verified), it only retransmits in response to the Client Initial; this response is all server packets that have not been acked. With no retransmissions, the server will need to have some kind of timer to clean up the state if the connection is dead.

There are many other ways to solve this problem; this is an example.

ianswett:
Fixes #1764

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -882,11 +882,11 @@ a NEW_TOKEN frame) or by processing any message from the client encrypted using
 the Handshake keys.  This limit exists to mitigate amplification attacks.
 
 In order to prevent this limit causing a handshake deadlock, the client SHOULD
-send a packet as large as the Initial containing only PADDING if it has no
-other data to send and does not yet have the Handshake keys.  If the client
-has no data to send and the Handshake keys are available, it SHOULD send a
-packet with a single byte of padding.  Details on when to send these PADDING
-packets are in {{QUIC-RECOVERY}}.
+always send a packet upon a handshake timeout, as described in
+{{QUIC-RECOVERY}}.  If the client has no data to retransmit and does not have
+Handshake keys, it should send a packet as large as the Initial containing
+only PADDING.  If the client has Handshake keys, it SHOULD send a packet
+containing only PADDING.
 
 The payload of this packet contains CRYPTO frames and could contain PADDING, or
 ACK frames. Handshake packets MAY contain CONNECTION_CLOSE or APPLICATION_CLOSE
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -884,7 +884,7 @@ the Handshake keys.  This limit exists to mitigate amplification attacks.
 In order to prevent this limit causing a handshake deadlock, the client SHOULD
 send a packet as large as the Initial containing only PADDING if it has no
 other data to send and does not yet have the Handshake keys.  If the client
-has no data to send and the Hanshake keys are available, it SHOULD send a
+has no data to send and the Handshake keys are available, it SHOULD send a
 packet with a single byte of padding.  Details on when to send these PADDING
 packets are in {{QUIC-RECOVERY}}.
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -875,11 +875,17 @@ The first Handshake packet sent by a server contains a packet number of 0.
 Handshake packets are their own packet number space.  Packet numbers are
 incremented normally for other Handshake packets.
 
-Servers MUST NOT send more than three datagrams including Initial and Handshake
-packets without receiving a packet from a verified source address.  Source
-addresses can be verified through an address validation token
-(delivered via a Retry packet or a NEW_TOKEN frame) or by receiving
-any message from the client encrypted using the Handshake keys.
+Servers MUST NOT send more than three times as many bytes as the number of bytes
+received prior to verifying the client's address.  Source addresses can be
+verified through an address validation token (delivered via a Retry packet or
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -875,11 +875,18 @@ The first Handshake packet sent by a server contains a packet number of 0.
...
 
-Servers MUST NOT send more than three datagrams including Initial and Handshake
-packets without receiving a packet from a verified source address.  Source
-addresses can be verified through an address validation token
-(delivered via a Retry packet or a NEW_TOKEN frame) or by receiving
-any message from the client encrypted using the Handshake keys.
+Servers MUST NOT send more than three times as many bytes as the number of bytes
+received prior to verifying the client's address.  Source addresses can be
+verified through an address validation token (delivered via a Retry packet or
+a NEW_TOKEN frame) or by receiving any message from the client encrypted using
```
MikeBishop: s/receiving/processing/ -- it has to be valid; if you can unprotect it, then you're okay.

draft-ietf-quic-transport.md
```diff
@@ -875,11 +875,18 @@ The first Handshake packet sent by a server contains a packet number of 0.
...
+the Handshake keys.  This limit exists to mitigate amplification attacks.
+
+In order to prevent this limit causing a handshake deadlock, the client SHOULD
+send a packet as large as the Initial containing only PADDING if it has no
+other data to send and does not yet have the Handshake keys.  If the client
+has no data to send and the Hanshake keys are available, it SHOULD send a
```
MikeBishop: Han**d**shake

martinduke: ... send a Handshake packet ...

draft-ietf-quic-transport.md
```diff
@@ -875,11 +875,18 @@ The first Handshake packet sent by a server contains a packet number of 0.
...
+packet with a single byte of padding.  Details on when to send these PADDING
+packets are in {{QUIC-RECOVERY}}.
```
martinduke: I couldn't find these details after a quick skim.

draft-ietf-quic-transport.md
martinduke: I think valid resumption keys should also be sufficient.

draft-ietf-quic-transport.md
```diff
@@ -875,11 +875,18 @@ The first Handshake packet sent by a server contains a packet number of 0.
...
-addresses can be verified through an address validation token
-(delivered via a Retry packet or a NEW_TOKEN frame) or by receiving
-any message from the client encrypted using the Handshake keys.
+Servers MUST NOT send more than three times as many bytes as the number of bytes
+received prior to verifying the client's address.  Source addresses can be
+verified through an address validation token (delivered via a Retry packet or
+a NEW_TOKEN frame) or by receiving any message from the client encrypted using
+the Handshake keys.  This limit exists to mitigate amplification attacks.
+
+In order to prevent this limit causing a handshake deadlock, the client SHOULD
```
martinduke: Is SHOULD sufficient? I feel like this a MUST, or compliant implementations will never complete under certain otherwise recoverable loss scenarios. Perhaps the recovery draft is more strict?

draft-ietf-quic-transport.md
```diff
@@ -875,11 +875,18 @@ The first Handshake packet sent by a server contains a packet number of 0.
 Handshake packets are their own packet number space.  Packet numbers are
 incremented normally for other Handshake packets.
 
-Servers MUST NOT send more than three datagrams including Initial and Handshake
-packets without receiving a packet from a verified source address.  Source
-addresses can be verified through an address validation token
-(delivered via a Retry packet or a NEW_TOKEN frame) or by receiving
-any message from the client encrypted using the Handshake keys.
+Servers MUST NOT send more than three times as many bytes as the number of bytes
```
martinduke: I suggest a rephrase: 

Prior to verifying the client's address, servers MUST NOT send more than three times as many bytes as the number of bytes received.

draft-ietf-quic-transport.md
```diff
@@ -875,11 +875,18 @@ The first Handshake packet sent by a server contains a packet number of 0.
...
+received prior to verifying the client's address.  Source addresses can be
+verified through an address validation token (delivered via a Retry packet or
+a NEW_TOKEN frame) or by processing any message from the client encrypted using
```
martinduke: Why not valid resumption keys?

draft-ietf-quic-transport.md

ianswett:
@martinduke I was meaning to follow up with a PR to fix recovery, but if you have time, that would be very helpful.  If not, I'll try to send out one tomorrow.

PR #1841, opened 2018-10-08
Fix Frame # for RETIRE_CONNECTION_ID

martinduke:
keep sections in frame ID sequence too.


martinthomson:
0x1b == 0x0d, so no problem.

Issue #1836, opened 2018-10-05
"Handshake Packet" is Overloaded


martinduke:
In the base drafts I can think of three definitions of handshake packets:

1) Packets of type "handshake" (obviously)
2) Packets with Long Headers (Used to limit server send before address verification), not formally defined as such but confusingly in the "handshake packets" section even though it applies to Initial, Handshake, and (in theory) 1-RTT packets.
3) In the Recovery draft, is_handshake_packets refers to packets with long headers that also contain CRYPTO frames.

This isn't a satisfactory state of affairs. I propose we retain #1. For #2, I believe it's sufficient to simply put this somewhere else (I vote for the flow control section) and make it clear that it applies to packets of all types.

For #3, I suggest we name it "crypto packets", as this term doesn't exist elsewhere. I don't really care what the name is as long as it's unique.

I'm happy to file a PR if no one has any problems with what I've proposed.

ianswett:
For 3, crypto packets SGTM and a PR is always appreciated.  I'd suggest a separate PR from fixing 2, since it's a separate doc and they're not inter-dependent.

On 2, I'm less clear what you're referencing.  There's a section called Cryptographic Handshake Packets(https://tools.ietf.org/html/draft-ietf-quic-transport-15#section-4.5) that does a good job of not abbreviating to handshake packets.

martinduke:
@ianswett Yes, that's a good section. It's section 4.7 that has the server send limitations that apply to both Initial and Handshake packets. the PR I just filed moves the text to Section 4.5.

PR #1842, opened 2018-10-08
Rename is_handshake_packets in quic-recovery

martinduke:
Resolves part of #1836.


martinduke:
#1859 handles this more comprehensively.

PR #1843, opened 2018-10-08
Move server send limitations

### LINKED ISSUE #1836: "Handshake Packet" is Overloaded

martinduke:
This text was in the "Handshake Packet" section but it applies to Handshake and Initial Packets, so I moved it.
Together with PR #1842, this resolves #1836.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -664,6 +664,18 @@ packets are protected with connection- and version-specific keys
 provide confidentiality or integrity against on-path attackers, but
 provides some level of protection against off-path attackers.
 
+Servers MUST NOT send more than three times as many bytes as the number of bytes
+received prior to verifying the client's address.  Source addresses can be
+verified through an address validation token (delivered via a Retry packet or
+a NEW_TOKEN frame) or by processing any message from the client encrypted using
+the Handshake keys.  This limit exists to mitigate amplification attacks.
+
+In order to prevent this limit causing a handshake deadlock, the client SHOULD
+always send a packet upon a handshake timeout, as described in
+{{QUIC-RECOVERY}}.  If the client has no data to retransmit and does not have
+Handshake keys, it SHOULD send an Initial packet in a UDP datagram of at least
+1200 octets.  If the client has Handshake keys, it SHOULD send a Handshake
+packet.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -664,6 +664,18 @@ packets are protected with connection- and version-specific keys
...
 provides some level of protection against off-path attackers.
 
+Servers MUST NOT send more than three times as many bytes as the number of bytes
+received prior to verifying the client's address.  Source addresses can be
+verified through an address validation token (delivered via a Retry packet or
+a NEW_TOKEN frame) or by processing any message from the client encrypted using
+the Handshake keys.  This limit exists to mitigate amplification attacks.
+
+In order to prevent this limit causing a handshake deadlock, the client SHOULD
+always send a packet upon a handshake timeout, as described in
```
ianswett: Should this be "crypto handshake timeout" now?

janaiyengar:
Can you rebase this to #1857? Alternatively, wait a bit until that PR lands and then update this PR.

draft-ietf-quic-transport.md
mikkelfj: How exactly is number of bytes measures?

martinduke:
@janaiyengar I'm happy to either do this when you submit #1857, or you could just incorporate the change (with @ianswett 's nit) directly into that PR and close this one. It's up to you.

martinthomson:
Integrated already.

Issue #1840, opened 2018-10-08
When to ignore coalesced packets and when to terminate the connection


igorlord:
[QUIC Transport "Coalescing Packets" section](https://tools.ietf.org/html/draft-ietf-quic-transport-15#section-4.9) states:

> If one or more packets in a datagram cannot be processed yet (because the keys are not yet
> available) or processing fails (decryption failure, unknown type, etc.), the receiver MUST still
> attempt to process the remaining packets.

The "etc" needs to be clarified.

There are many conditions under which the receiver MUST (or SHOULD or MAY) terminate the connection.  It is not clear when the requirement to terminate the connection trumps the requirement to attempt to process remaining packets and when it does not.

MikeBishop:
My interpretation would be that fatal errors can only occur *while processing* a packet, which is to say that protection has been successfully removed and you're consuming frames out of the payload.  Errors that prevent you from doing any processing in the first place, however, should cause you to keep going to other packets.

In other words, for most purposes, handle the packets independently -- if they were in separate datagrams and one was garbage, you'd move to the next-received datagram.

martinthomson:
Closed as part of the reorg.

PR #1844, opened 2018-10-08
Clarify coalesced packet processing

igorlord:
Addressing #1840

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -961,9 +961,9 @@ complete.  Though the values of some fields in the packet header might be
 redundant, no fields are omitted.  The receiver of coalesced QUIC packets MUST
 individually process each QUIC packet and separately acknowledge them, as if
 they were received as the payload of separate UDP datagrams.  For example, if
-decryption fails or the packet is of an unknown type, the receiver MAY either
-discard or buffer the packet for later processing and MUST attempt to process
-the remaining packets.
+decryption fails (the keys are not yet available or any other reason) or the
+packet is of an unknown type, the receiver MAY either discard or buffer the
+packet for later processing and MUST attempt to process the remaining packets.
 
 Retry ({{packet-retry}}) and Version Negotiation ({{packet-version}}) packets
 cannot be coalesced.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -961,10 +961,9 @@ complete.  Though the values of some fields in the packet header might be
 redundant, no fields are omitted.  The receiver of coalesced QUIC packets MUST
 individually process each QUIC packet and separately acknowledge them, as if
 they were received as the payload of separate UDP datagrams.  For example, if
-one or more packets in a datagram cannot be decrypted (the keys are not yet
-available, decryption failure, unknown type), the receiver MAY either discard or
-buffer for packet for later processing and MUST attempt to process the remaining
-packets.
+decryption fails or the packet is of an unknown type, the receiver MAY either
+discard or buffer the packet for later processing and MUST attempt to process
+the remaining packets.
 
 Retry ({{packet-retry}}) and Version Negotiation ({{packet-version}}) packets
 cannot be coalesced.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -960,12 +960,11 @@ Every QUIC packet that is coalesced into a single UDP datagram is separate and
...
 individually process each QUIC packet and separately acknowledge them, as if
-they were received as the payload of different UDP datagrams.  If one or more
-packets in a datagram cannot be processed yet (because the keys are not yet
-available) or processing fails (decryption failure, unknown type, etc.), the
-receiver MUST still attempt to process the remaining packets.  The skipped
-packets MAY either be discarded or buffered for later processing, just as if the
-packets were received out-of-order in separate datagrams.
+they were received as the payload of separate UDP datagrams.  For example, if
+one or more packets in a datagram cannot be decrypted (the keys are not yet
+available, decryption failure, unknown type), the receiver MAY either discard or
```
ianswett: ", decryption fails, or the packet is of an unknown type"

draft-ietf-quic-transport.md
```diff
@@ -960,12 +960,11 @@ Every QUIC packet that is coalesced into a single UDP datagram is separate and
...
+buffer for packet for later processing and MUST attempt to process the remaining
```
ianswett: for packet -> the packet

draft-ietf-quic-transport.md
igorlord: Thanks.  Applied the suggestion.

draft-ietf-quic-transport.md
igorlord: Fixed

ianswett (approved):

mikkelfj (commented):
If no more long headers are expected, it would seem reasonable to drop the packet.

igorlord:
Long headers are useful for PMTUD to validate (and route) PTB ICMP messages. So we discussed in Seattle the idea of prepending long header packets to PMTUD Probe packets.

draft-ietf-quic-transport.md
```diff
@@ -960,12 +960,10 @@ Every QUIC packet that is coalesced into a single UDP datagram is separate and
...
 redundant, no fields are omitted.  The receiver of coalesced QUIC packets MUST
 individually process each QUIC packet and separately acknowledge them, as if
-they were received as the payload of different UDP datagrams.  If one or more
-packets in a datagram cannot be processed yet (because the keys are not yet
-available) or processing fails (decryption failure, unknown type, etc.), the
-receiver MUST still attempt to process the remaining packets.  The skipped
-packets MAY either be discarded or buffered for later processing, just as if the
-packets were received out-of-order in separate datagrams.
+they were received as the payload of separate UDP datagrams.  For example, if
+decryption fails or the packet is of an unknown type, the receiver MAY either
```
MikeBishop: This seems to lose the case this text was originally added to address, which is that you don't have the keys to process this packet type yet.  While these are non-limiting examples, that's probably a case worth calling out explicitly.

igorlord: Makes sense.  Updated the PR to mention this case specifically.

draft-ietf-quic-transport.md
```diff
@@ -960,12 +960,10 @@ Every QUIC packet that is coalesced into a single UDP datagram is separate and
...
 individually process each QUIC packet and separately acknowledge them, as if
-they were received as the payload of different UDP datagrams.  If one or more
-packets in a datagram cannot be processed yet (because the keys are not yet
-available) or processing fails (decryption failure, unknown type, etc.), the
-receiver MUST still attempt to process the remaining packets.  The skipped
-packets MAY either be discarded or buffered for later processing, just as if the
-packets were received out-of-order in separate datagrams.
+they were received as the payload of separate UDP datagrams.  For example, if
+decryption fails (the keys are not yet available or any other reason) or the
+packet is of an unknown type, the receiver MAY either discard or buffer the
```
janaiyengar: "the packet" does not make sense in this context. How about "If one or more packets in a datagram cannot be processed yet, because of decryption failure or any other reason, the receiver MAY either discard or buffer the packets for later processing. The receiver MUST still attempt to process the remaining packets in the UDP datagram."

martinthomson: That sounds good.  Want to port that over to #1857?

igorlord: I would omit "yet".  "Yet" implies knowing that it will be possible to decrypt the packet in the future.  It is hard to know the future.

martinthomson:
Merged by hand.

PR #1856, opened 2018-10-12
Parenthesize 2 * MaxEntries to prevent naive errors

dtikhonov:
Transcribing

```
  LargestReference mod 2*MaxEntries + 1
```

to C one of its offspring without parentheses around 2 * MaxEntries
will produce an incorrect result.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -718,7 +718,7 @@ safe to process the rest of the block.  If Largest Reference is greater than
 zero, the encoder transforms it as follows before encoding:
 
 ~~~
-   LargestReference = LargestReference mod (2*MaxEntries) + 1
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -718,15 +718,15 @@ safe to process the rest of the block.  If Largest Reference is greater than
 zero, the encoder transforms it as follows before encoding:
 
 ~~~
-   LargestReference = LargestReference mod 2*MaxEntries + 1
+   LargestReference = LargestReference mod (2*MaxEntries) + 1
```
martinthomson: Do you mean `LR % ((2*ME)+1)` or `(LR % (2*ME))+1`?  I think the latter, but I think that it's probably best to double-down on the parentheses here.

dtikhonov: I mean the latter.  I can add more parentheses.

PR #1859, opened 2018-10-15
Replace handshake with crypto

### LINKED ISSUE #1836: "Handshake Packet" is Overloaded

ianswett:
This replaces #1842 and tries to fix everything to consistently use crypto packets instead of handshake packets.

Fixes #1836.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -311,23 +311,23 @@ RTT.
 
 At the beginning, there are no prior RTT samples within a connection.  Resumed
 connections over the same network SHOULD use the previous connection's final
-smoothed RTT value as the resumed connection's initial RTT.
+smoothed RTT value as the resumed connection's initial RTT.  When an
+acknowledgement is received, a new RTT is computed and the timer
+SHOULD be set for twice the newly computed smoothed RTT.
 
 If no previous RTT is available, or if the network changes, the initial RTT
 SHOULD be set to 100ms.
 
-When CRYPTO frames are sent, the sender SHOULD set a timer for the crypto
-timeout period.  Upon timeout, the sender MUST retransmit all unacknowledged
-CRYPTO data by calling RetransmitAllUnackedCryptoData(). On each consecutive
-expiration of the crypto timer without receiving an acknowledgement for a new
-packet, the sender SHOULD double the crypto handshake timeout and set a timer
-for this period.
+When CRYPTO frames are sent in Initial or Handshake packets, the sender SHOULD
+set a timer for the crypto timeout period.  Upon timeout, the sender MUST
+retransmit all unacknowledged CRYPTO data by calling
+RetransmitAllUnackedCryptoData(). On each consecutive expiration of the
+crypto timer without receiving an acknowledgement for a new packet, the
+sender SHOULD double the crypto handshake timeout and set a timer for this
+period.
 
 When CRYPTO frames are outstanding, the TLP and RTO timers are not active unless
 the CRYPTO frames were sent at 1-RTT encryption.
-
-When an acknowledgement is received for a crypto packet, the new RTT is
-computed and the timer SHOULD be set for twice the newly computed smoothed RTT.
 
 #### Retry and Version Negotiation
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -303,9 +303,10 @@ Retransmission Timeout mechanisms.
 
 Data in CRYPTO frames is critical to QUIC transport and crypto negotiation, so a
 more aggressive timeout is used to retransmit it.  Below, the term "crypto
-packet" is used to refer to packets containing CRYPTO frames.
+packet" is used to refer to packets containing CRYPTO frames before the
+handshake is complete.
 
-The initial crypto timeout SHOULD be set to twice the initial RTT.
+The initial crypto handshake timeout SHOULD be set to twice the initial RTT.
 
 At the beginning, there are no prior RTT samples within a connection.  Resumed
 connections over the same network SHOULD use the previous connection's final
@@ -318,8 +319,8 @@ When CRYPTO frames are sent, the sender SHOULD set a timer for the crypto
 timeout period.  Upon timeout, the sender MUST retransmit all unacknowledged
 CRYPTO data by calling RetransmitAllUnackedCryptoData(). On each consecutive
 expiration of the crypto timer without receiving an acknowledgement for a new
-packet, the sender SHOULD double the crypto timeout and set a timer for this
-period.
+packet, the sender SHOULD double the crypto handshake timeout and set a timer
+for this period.
 
 When CRYPTO frames are outstanding, the TLP and RTO timers are not active unless
 the CRYPTO frames were sent at 1-RTT encryption.
@@ -766,7 +767,7 @@ below shows how the single timer is set.
 #### Crypto Handshake Timer
 
 When a connection has unacknowledged handshake data, the handshake timer is
-set and when it expires, all unacknowledgedd CRYPTO data is retransmitted.
+set and when it expires, all unacknowledged CRYPTO data is retransmitted.
 
 When stateless rejects are in use, the connection is considered immediately
 closed once a reject is sent, so no timer is set to retransmit the reject.
@@ -814,8 +815,8 @@ Pseudocode for SetLossDetectionTimer follows:
       timeout = timeout * (2 ^ crypto_count)
       loss_detection_timer.set(
         time_of_last_sent_crypto_packet + timeout)
-      return;
-    else if (loss_time != 0):
+      return
+    if (loss_time != 0):
       // Early retransmit timer or time loss detection.
       timeout = loss_time -
         time_of_last_sent_retransmittable_packet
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -302,11 +302,10 @@ Retransmission Timeout mechanisms.
 ### Crypto Handshake Timeout
 
 Data in CRYPTO frames is critical to QUIC transport and crypto negotiation, so a
-more aggressive timeout is used to retransmit it.  Below, the term "handshake
-packet" is used to refer to packets containing CRYPTO frames, not packets with
-the specific long header packet type Handshake.
+more aggressive timeout is used to retransmit it.  Below, the term "crypto
+packet" is used to refer to packets containing CRYPTO frames.
```
martinthomson: We should further qualify this: "Handshake or Initial packets containing CRYPTO frames", because CRYPTO frames after the handshake completes don't require special handling.

draft-ietf-quic-recovery.md
```diff
@@ -764,10 +763,10 @@ duration of the timer is based on the timer's mode, which is set in the packet
 and timer events further below.  The function SetLossDetectionTimer defined
 below shows how the single timer is set.
 
-#### Handshake Timer
+#### Crypto Handshake Timer
 
 When a connection has unacknowledged handshake data, the handshake timer is
-set and when it expires, all unacknowledgedd handshake data is retransmitted.
+set and when it expires, all unacknowledgedd CRYPTO data is retransmitted.
```
martinthomson: typo: dd

draft-ietf-quic-recovery.md
```diff
@@ -302,11 +302,10 @@ Retransmission Timeout mechanisms.
...
 
 Data in CRYPTO frames is critical to QUIC transport and crypto negotiation, so a
-more aggressive timeout is used to retransmit it.  Below, the term "handshake
-packet" is used to refer to packets containing CRYPTO frames, not packets with
-the specific long header packet type Handshake.
+more aggressive timeout is used to retransmit it.  Below, the term "crypto
+packet" is used to refer to packets containing CRYPTO frames.
 
-The initial handshake timeout SHOULD be set to twice the initial RTT.
+The initial crypto timeout SHOULD be set to twice the initial RTT.
```
martinthomson: Can we rename this?  I initially assumed that the handshake would be abandoned when this fires.  It's a crypto retransmission timeout (a term you already use).

draft-ietf-quic-recovery.md
```diff
@@ -805,16 +804,16 @@ Pseudocode for SetLossDetectionTimer follows:
...
         timeout = 2 * kInitialRtt
       else:
         timeout = 2 * smoothed_rtt
       timeout = max(timeout, kMinTLPTimeout)
-      timeout = timeout * (2 ^ handshake_count)
+      timeout = timeout * (2 ^ crypto_count)
       loss_detection_timer.set(
-        time_of_last_sent_handshake_packet + timeout)
+        time_of_last_sent_crypto_packet + timeout)
       return;
```
martinthomson: extra semi-colon

draft-ietf-quic-recovery.md
```diff
@@ -805,16 +804,16 @@ Pseudocode for SetLossDetectionTimer follows:
...
     else if (loss_time != 0):
```
martinthomson: No need for an else after a return

draft-ietf-quic-recovery.md
janaiyengar: +1

ianswett: Done, but is the term "handshake complete" well-defined?  I wanted to use long header packets, but I realized that the CFIN is head of line blocking :(

draft-ietf-quic-recovery.md
ianswett: I went with consistently using crypto handshake timeout, because I think it reads well and is consistent, but I can change it to something else if you prefer.

draft-ietf-quic-recovery.md
```diff
@@ -302,11 +302,11 @@ Retransmission Timeout mechanisms.
 ### Crypto Handshake Timeout
 
 Data in CRYPTO frames is critical to QUIC transport and crypto negotiation, so a
-more aggressive timeout is used to retransmit it.  Below, the term "handshake
-packet" is used to refer to packets containing CRYPTO frames, not packets with
-the specific long header packet type Handshake.
+more aggressive timeout is used to retransmit it.  Below, the term "crypto
+packet" is used to refer to packets containing CRYPTO frames that are part of
+the handshake.
```
martinthomson: Better to say "Initial and Handshake" explicitly.

draft-ietf-quic-recovery.md
```diff
@@ -302,11 +302,11 @@ Retransmission Timeout mechanisms.
...
 
-The initial handshake timeout SHOULD be set to twice the initial RTT.
+The initial crypto handshake timeout SHOULD be set to twice the initial RTT.
```
martinthomson: initial crypto **retransmission** timeout

draft-ietf-quic-recovery.md
```diff
@@ -302,32 +302,31 @@ Retransmission Timeout mechanisms.
...
+When CRYPTO frames are sent in Initial or Handshake packets, the sender SHOULD
+set a timer for the crypto timeout period.  Upon timeout, the sender MUST
+retransmit all unacknowledged CRYPTO data by calling
+RetransmitAllUnackedCryptoData(). On each consecutive expiration of the
+crypto timer without receiving an acknowledgement for a new packet, the
+sender SHOULD double the crypto handshake timeout and set a timer for this
+period.
 
 When CRYPTO frames are outstanding, the TLP and RTO timers are not active unless
 the CRYPTO frames were sent at 1-RTT encryption.
```
MikeBishop: Now that you have a definition for "crypto packets" above which excludes 1-RTT, you could probably simplify this by saying that "When crypto packets are outstanding, the TLS and RTO timers are not active."

draft-ietf-quic-recovery.md
```diff
@@ -302,32 +302,31 @@ Retransmission Timeout mechanisms.
...
-CRYPTO data by calling RetransmitAllUnackedHandshakeData(). On each
-consecutive expiration of the handshake timer without receiving an
-acknowledgement for a new packet, the sender SHOULD double the handshake timeout
-and set a timer for this period.
+smoothed RTT value as the resumed connection's initial RTT.  If no previous RTT
+is available, or if the network changes, the initial RTT SHOULD be set to 100ms.
+When an acknowledgement is received, a new RTT is computed and the timer
+SHOULD be set for twice the newly computed smoothed RTT.
+
+When CRYPTO frames are sent in Initial or Handshake packets, the sender SHOULD
```
MikeBishop: Likewise, you can condense to "When crypto packets are sent,"

draft-ietf-quic-recovery.md
```diff
@@ -558,7 +557,7 @@ largest_sent_before_rto:
 time_of_last_sent_retransmittable_packet:
 : The time the most recent retransmittable packet was sent.
 
-time_of_last_sent_handshake_packet:
+time_of_last_sent_crypto_packet:
 : The time the most recent packet containing a CRYPTO frame was sent.
```
MikeBishop: This includes 1-RTT as currently defined, but I don't think you want it to.

draft-ietf-quic-recovery.md
martinduke: TLS->TLP

draft-ietf-quic-recovery.md
```diff
@@ -302,32 +302,31 @@ Retransmission Timeout mechanisms.
 ### Crypto Handshake Timeout
 
 Data in CRYPTO frames is critical to QUIC transport and crypto negotiation, so a
-more aggressive timeout is used to retransmit it.  Below, the term "handshake
-packet" is used to refer to packets containing CRYPTO frames, not packets with
-the specific long header packet type Handshake.
+more aggressive timeout is used to retransmit it.  Below, the term "crypto
+packet" is used to refer to packets containing CRYPTO data sent in Initial or
```
martinduke: Maybe add this definition to "Conventions and Definitions" above?

draft-ietf-quic-recovery.md
```diff
@@ -764,10 +763,10 @@ duration of the timer is based on the timer's mode, which is set in the packet
 and timer events further below.  The function SetLossDetectionTimer defined
 below shows how the single timer is set.
 
-#### Handshake Timer
+#### Crypto Handshake Timer
 
 When a connection has unacknowledged handshake data, the handshake timer is
```
martinduke: "handshake data" -> "crypto handshake data"
"handshake timer" -> "crypto handshake timer"

PR #1862, opened 2018-10-17
Amplification limit in recovery

ianswett:
Follow up for Issue #1764

Also fixes some remaining inconsistencies from #1859

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -330,7 +330,7 @@ may be sent, then no alarm should be armed until bytes have been received from
 the client.
 
 Because the server may be blocked until more packets are received, the
-client MUST arm the crypto retransmission alarm even if there is no
+client MUST start the crypto retransmission timer even if there is no
 unacknowledged CRYPTO data.  If the timer expires and the client has no
 CRYPTO data to retransmit and does not have Handshake keys, it SHOULD send
 an Initial packet in a UDP datagram of at least 1200 octets.
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -323,9 +323,9 @@ timeout period.  Upon timeout, the sender MUST retransmit all unacknowledged
 CRYPTO data if possible.
 
 Until the server has validated the client's address on the path, the number of
-bytes it can send is limited, as specified in the {{QUIC-TRANSPORT}}.
-If not all unacknowledged CRYPTO data can be sent, then all CRYPTO data in
-Initial encryption should be retransmitted.  If no bytes may be sent,
+bytes it can send is limited, as specified in {{QUIC-TRANSPORT}}.
+If not all unacknowledged CRYPTO data can be sent, then all CRYPTO data sent
+in Initial packets should be retransmitted.  If no bytes may be sent,
 then no alarm should be armed until bytes have been received by the peer.
 
 On each consecutive expiration of the crypto timer without receiving an
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -320,10 +320,17 @@ SHOULD be set for twice the newly computed smoothed RTT.
 
 When crypto packets are sent, the sender SHOULD set a timer for the crypto
 timeout period.  Upon timeout, the sender MUST retransmit all unacknowledged
-CRYPTO data by calling RetransmitAllUnackedCryptoData(). On each consecutive
-expiration of the crypto timer without receiving an acknowledgement for a new
-packet, the sender SHOULD double the crypto handshake timeout and set a timer
-for this period.
+CRYPTO data if possible.
+
+If the server has not confirmed the path, it must not send more than 3
+times the number of received bytes, as specified in the {{QUIC-TRANSPORT}}.
+If not all unacknowledged CRYPTO data can be sent, then all CRYPTO data in
+Initial encryption should be retransmitted.  If no bytes may be sent,
+then no alarm should be armed until bytes have been received by the peer.
+
+On each consecutive expiration of the crypto timer without receiving an
+acknowledgement for a new packet, the sender SHOULD double the crypto
+handshake timeout and set a timer for this period.
 
 When crypto packets are outstanding, the TLP and RTO timers are not active.
 
@@ -763,37 +770,6 @@ duration of the timer is based on the timer's mode, which is set in the packet
 and timer events further below.  The function SetLossDetectionTimer defined
 below shows how the single timer is set.
 
-#### Crypto Handshake Timer
-
-When a connection has unacknowledged crypto packets, the crypto handshake timer
-is set and when it expires, all unacknowledged CRYPTO data is retransmitted.
-
-When stateless rejects are in use, the connection is considered immediately
-closed once a reject is sent, so no timer is set to retransmit the reject.
-
-Version negotiation packets are always stateless, and MUST be sent once per
-handshake packet that uses an unsupported QUIC version, and MAY be sent in
-response to 0-RTT packets.
-
-#### Tail Loss Probe and Retransmission Timer
-
-Tail loss probes {{?TLP}} and retransmission timeouts {{?RFC6298}}
-are timer based mechanisms to recover from cases when there are
-outstanding retransmittable packets, but an acknowledgement has
-not been received in a timely manner.
-
-The TLP and RTO timers are armed when there are no unacknowledged crypto
-packets.  The TLP timer is set until the max number of TLP packets have been
-sent, and then the RTO timer is set.
-
-#### Early Retransmit Timer
-
-Early retransmit {{?RFC5827}} is implemented with a 1/4 RTT timer. It is
-part of QUIC's time based loss detection, but is always enabled, even when
-only packet reordering loss detection is enabled.
-
-#### Pseudocode
-
 Pseudocode for SetLossDetectionTimer follows:
 
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -320,10 +320,17 @@ SHOULD be set for twice the newly computed smoothed RTT.
 
 When crypto packets are sent, the sender SHOULD set a timer for the crypto
 timeout period.  Upon timeout, the sender MUST retransmit all unacknowledged
-CRYPTO data by calling RetransmitAllUnackedCryptoData(). On each consecutive
-expiration of the crypto timer without receiving an acknowledgement for a new
-packet, the sender SHOULD double the crypto handshake timeout and set a timer
-for this period.
+CRYPTO data if possible.
+
+Until the server has confirmed the path, it must not send more than 3
```
martinthomson: rather than "confirmed the path", we should be more consistent and say "the server has validated the client's address on the path"

```suggestion
Until the server has validated the client's address on the path, it must not send more than 3
```

draft-ietf-quic-recovery.md
```diff
@@ -320,10 +320,17 @@ SHOULD be set for twice the newly computed smoothed RTT.
...
+times the number of received bytes, as specified in the {{QUIC-TRANSPORT}}.
```
martinthomson: Is this MUST NOT?  You might avoid duplicating the requirement by saying that the server is limited in the amount of data it can send, as specified in {{QUIC-TRANSPORT}}.

draft-ietf-quic-recovery.md
```diff
@@ -320,10 +320,17 @@ SHOULD be set for twice the newly computed smoothed RTT.
...
+If not all unacknowledged CRYPTO data can be sent, then all CRYPTO data in
+Initial encryption should be retransmitted.  If no bytes may be sent,
+then no alarm should be armed until bytes have been received by the peer.
+
+On each consecutive expiration of the crypto timer without receiving an
+acknowledgement for a new packet, the sender SHOULD double the crypto
+handshake timeout and set a timer for this period.
```
martinthomson: The timeouts are still inconsistently named.  We have "crypto handshake timeout" here, but "crypto timeout" on 321/322.  I think that "crypto retransmission timeout " is most correct.  You should fix the section title as well.

draft-ietf-quic-recovery.md
```diff
@@ -763,37 +770,6 @@ duration of the timer is based on the timer's mode, which is set in the packet
...
-#### Tail Loss Probe and Retransmission Timer
-
-Tail loss probes {{?TLP}} and retransmission timeouts {{?RFC6298}}
-are timer based mechanisms to recover from cases when there are
-outstanding retransmittable packets, but an acknowledgement has
-not been received in a timely manner.
-
-The TLP and RTO timers are armed when there are no unacknowledged crypto
-packets.  The TLP timer is set until the max number of TLP packets have been
-sent, and then the RTO timer is set.
```
martinthomson: Did you mean to delete all of this?

ianswett: Yes, I believe it's duplicative of the more detailed text above.

draft-ietf-quic-recovery.md
```diff
@@ -320,10 +320,17 @@ SHOULD be set for twice the newly computed smoothed RTT.
...
-CRYPTO data by calling RetransmitAllUnackedCryptoData(). On each consecutive
-expiration of the crypto timer without receiving an acknowledgement for a new
-packet, the sender SHOULD double the crypto handshake timeout and set a timer
-for this period.
+CRYPTO data if possible.
+
+Until the server has validated the client's address on the path, the number of
+bytes it can send is limited, as specified in the {{QUIC-TRANSPORT}}.
+If not all unacknowledged CRYPTO data can be sent, then all CRYPTO data in
+Initial encryption should be retransmitted.  If no bytes may be sent,
```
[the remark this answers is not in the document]
ianswett: Hmm, probably Initial packets.

draft-ietf-quic-recovery.md
```diff
@@ -320,10 +320,17 @@ SHOULD be set for twice the newly computed smoothed RTT.
...
 timeout period.  Upon timeout, the sender MUST retransmit all unacknowledged
-CRYPTO data by calling RetransmitAllUnackedCryptoData(). On each consecutive
-expiration of the crypto timer without receiving an acknowledgement for a new
-packet, the sender SHOULD double the crypto handshake timeout and set a timer
-for this period.
+CRYPTO data if possible.
+
+Until the server has validated the client's address on the path, the number of
+bytes it can send is limited, as specified in {{QUIC-TRANSPORT}}.
+If not all unacknowledged CRYPTO data can be sent, then all CRYPTO data sent
```
martinduke: "then all unacknowledged CRYPTO data sent"

draft-ietf-quic-recovery.md
```diff
@@ -318,15 +318,23 @@ is available, or if the network changes, the initial RTT SHOULD be set to 100ms.
...
-If not all unacknowledged CRYPTO data can be sent, then all CRYPTO data sent
-in Initial packets should be retransmitted.  If no bytes may be sent,
-then no alarm should be armed until bytes have been received by the peer.
+If not all unacknowledged CRYPTO data can be sent, then all unacknowledged
+CRYPTO data sent in Initial packets should be retransmitted.  If no bytes
+may be sent, then no alarm should be armed until bytes have been received from
+the client.
+
+Because the server may be blocked until more packets are received, the
+client MUST arm the crypto retransmission alarm even if there is no
```
martinthomson: timer, not alarm
```suggestion
client MUST start the crypto retransmission timer even if there is no
```

draft-ietf-quic-recovery.md
```diff
@@ -318,12 +318,27 @@ is available, or if the network changes, the initial RTT SHOULD be set to 100ms.
 When an acknowledgement is received, a new RTT is computed and the timer
 SHOULD be set for twice the newly computed smoothed RTT.
 
-When crypto packets are sent, the sender SHOULD set a timer for the crypto
+When crypto packets are sent, the sender MUST set a timer for the crypto
 timeout period.  Upon timeout, the sender MUST retransmit all unacknowledged
```
```suggestion
retransmission interval.  Upon timeout, the sender MUST retransmit all unacknowledged
```

draft-ietf-quic-recovery.md
```diff
@@ -318,12 +318,27 @@ is available, or if the network changes, the initial RTT SHOULD be set to 100ms.
...
-expiration of the crypto timer without receiving an acknowledgement for a new
-packet, the sender SHOULD double the crypto handshake timeout and set a timer
-for this period.
+CRYPTO data if possible.
+
+Until the server has validated the client's address on the path, the number of
+bytes it can send is limited, as specified in {{QUIC-TRANSPORT}}.
+If not all unacknowledged CRYPTO data can be sent, then all unacknowledged
+CRYPTO data sent in Initial packets should be retransmitted.  If no bytes
+may be sent, then no alarm should be armed until bytes have been received from
```
```suggestion
can be sent, then no alarm should be armed until bytes have been received from
```

draft-ietf-quic-recovery.md
```diff
@@ -318,12 +318,27 @@ is available, or if the network changes, the initial RTT SHOULD be set to 100ms.
...
+the client.
+
+Because the server may be blocked until more packets are received, the
```
```suggestion
Because the server could be blocked until more packets are received, the
```

draft-ietf-quic-recovery.md
```diff
@@ -318,12 +318,27 @@ is available, or if the network changes, the initial RTT SHOULD be set to 100ms.
...
 timeout period.  Upon timeout, the sender MUST retransmit all unacknowledged
-CRYPTO data by calling RetransmitAllUnackedCryptoData(). On each consecutive
-expiration of the crypto timer without receiving an acknowledgement for a new
-packet, the sender SHOULD double the crypto handshake timeout and set a timer
-for this period.
+CRYPTO data if possible.
+
+Until the server has validated the client's address on the path, the number of
+bytes it can send is limited, as specified in {{QUIC-TRANSPORT}}.
+If not all unacknowledged CRYPTO data can be sent, then all unacknowledged
```
martinthomson: I think that you mean to say instead

> If all CRYPTO data is acknowledged, and there is nothing for the client to send, then it sends another Initial packet.

What that packet contains clearly doesn't matter (though an ACK would be perfectly OK, as would repeating information that the server already has), so I don't think we need to stipulate that explicitly.  Otherwise, this is a little ambiguous - and maybe in conflict with what we did in #1843 and the next paragraph.

draft-ietf-quic-recovery.md
martinthomson: OK.  It seemed entirely unrelated, but I see we have text on both elsewhere.

mikkelfj:
With these MUST retransmit, has it been considered what a Slowloris attack could do to the handshake? I know that there is a general mention in some section.

draft-ietf-quic-recovery.md
ianswett: This doc uses timeout and timeout period fairly consistently, so I'd like to keep it as is.

draft-ietf-quic-recovery.md
ianswett: This paragraph talks about the server's behavior, not the client's?

ianswett:
@mikkelfj The slowloris attack is against a server, but the peer that must retransmit is the client in this case, so I don't think this particular text is relevant to slowloris?

mikkelfj:
I got the same thought after writing it, but I suspected the logic could be similar server side - I'm not sufficiently into that handshake details atm.

PR #1869, opened 2018-10-17
Improve description of max_ack_delay transport parameter

dtikhonov:
The sentence it replaces is awkward:

> An 8 bit unsigned integer value indicating the maximum amount of time in milliseconds by which _it_ will delay sending of acknowledgments.

What is _it_ -- the integer value?  Can an integer value delay sending of acknowledgements?

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1729,7 +1729,7 @@ disable_migration (0x0009):
 max_ack_delay (0x000c):
 
 : An 8 bit unsigned integer value indicating the maximum amount of time in
-  milliseconds by which the endpoint will delay sending of acknowledgments.
+  milliseconds by which the endpoint will delay sending acknowledgments.
   If this value is absent, a default of 25 milliseconds is assumed.
 
 Either peer MAY advertise an initial value for the flow control on each type of


draft-ietf-quic-transport.md
```diff
@@ -1728,9 +1728,9 @@ disable_migration (0x0009):
 
 max_ack_delay (0x000c):
 
-: An 8 bit unsigned integer value indicating the maximum amount of time in
-  milliseconds by which it will delay sending of acknowledgments.  If this
-  value is absent, a default of 25 milliseconds is assumed.
+: An 8-bit unsigned integer value to use as the maximum delay (in milliseconds)
```
ianswett: The intent of this transport param is for the receiver to communicate it's max ack delay to the sender, so I don't think the change from "indicating" to "to use as" is quite right.

How about "An 8-bit unsigned integer value indicating the maximum amount of time in milliseconds a receiver will delay sending an acknowledgement.  The receiver sends this value to the sender to set timeouts more accurately."?

I remember this was reworded a few times when I was writing it,

dtikhonov: Oh, that's right.  In this case, I how about "used as?"

> The maximum delay (in milliseconds) used to send acknowledgements encoded as an 8-bit unsigned integer.

Alternatively, I can close this PR and you can improve it as you see fit, @ianswett.

draft-ietf-quic-transport.md
```diff
@@ -1728,9 +1728,9 @@ disable_migration (0x0009):
 
 max_ack_delay (0x000c):
 
-: An 8 bit unsigned integer value indicating the maximum amount of time in
-  milliseconds by which it will delay sending of acknowledgments.  If this
-  value is absent, a default of 25 milliseconds is assumed.
+: The maximum delay (in milliseconds) used to send acknowledgements, encoded as
+  an 8-bit unsigned integer.  If this value is absent, a default of 25
```
martinthomson: Nor do you use a delay to send acknowledgments.  Why not simply s/it/the endpoint/ ?

draft-ietf-quic-transport.md
```diff
@@ -1729,8 +1729,8 @@ disable_migration (0x0009):
 max_ack_delay (0x000c):
 
 : An 8 bit unsigned integer value indicating the maximum amount of time in
-  milliseconds by which it will delay sending of acknowledgments.  If this
-  value is absent, a default of 25 milliseconds is assumed.
+  milliseconds by which the endpoint will delay sending of acknowledgments.
```
ianswett: nit: I think "sending acknowledgements." reads a bit better than "sending of acknowledgments."

PR #1886, opened 2018-10-19
Move and consolidate address validation

martinthomson:
The two disparate sections on address validation are better together.  That means deferring the description of Retry to after the connection establishment section, but I think that makes sense.  I've added a forward reference.

Aside from moves there is a tiny bit of new glue text in the top of the respective sections.  I had to tweak some of it to reflect the reorganization, and I noticed some duplication that I eliminated from the first two paragraphs of the path migration section.

I added a picture showing how Retry is added to the handshake flow, addressing a TODO.

I merged the sections on dealing with NEW_TOKEN and address validation for future connections (or should that be returning connections...).  That was just removing the section heading.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1575,7 +1575,7 @@ NEW_TOKEN frame (see {{validate-future}}).
 QUIC uses token-based address validation during connection establishment.  Any
 time the server wishes to validate a client address, it provides the client with
 a token.  As long as it is not possible for an attacker to generate a valid
-token for its address (see {{token-integrity}}) and the client is able to return
+token for its own address (see {{token-integrity}}) and the client is able to return
 that token, it proves to the server that it received the token.
 
 Upon receiving the client's Initial packet, the server can request address
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1530,34 +1530,31 @@ validation is performed both during connection establishment (see
 
 ## Address Validation During Connection Establishment {#validate-new}
 
-During the initial handshake, QUIC requires that clients send UDP datagrams with
-at least 1200 octets of payload until the server has completed address
-validation. A server can thereby send more data to an unproven address without
-increasing the amplification advantage gained by an attacker.
+Successful receipt of the first packet protected with Handshake keys confirms
+that a client received the Initial packet from the server.
 
-A server eventually confirms that a client has received its messages when the
-first Handshake-level message is received. This might be insufficient, either
-because the server wishes to avoid the computational cost of completing the
-handshake, or because the server wish to send significant quantities of data.
-This is especially important for 0-RTT, where the server might wish to provide
-application data traffic - such as a response to a request - in response to the
-data carried in the early data from the client.
+However, a server might wish to validate the client address before starting the
+handshake.  To send additional data prior to completing the cryptographic
+handshake, the server then needs to validate that the client owns the address
+that it claims.  QUIC therefore provides mechanisms for source address
+validation during connection establishment.
 
-To send additional data prior to completing the cryptographic handshake, the
-server then needs to validate that the client owns the address that it claims.
-QUIC therefore provides mechanisms for source address validation during
-connection establishment.
+Source addresses can be verified through an address validation token.  This
+token is delivered during connection establishment with a Retry packet (see
+{{validate-retry}}) or in a previous connection using the NEW_TOKEN frame (see
+{{validate-future}}).
 
-Source addresses can be verified through an address validation token (delivered
-via a Retry packet or a NEW_TOKEN frame) or by processing any message from the
-client encrypted using the Handshake keys.
+Prior to validating the client address, servers MUST NOT send more than three
+times as many bytes as the number of bytes they have received.  This limits the
+magnitude of any amplification attack that can be mounted using spoofed source
+addresses.
 
-Servers MUST NOT send more than three times as many bytes as the number of bytes
-received prior to verifying the that the client can receive data from the
-server.  This limit exists to mitigate amplification attacks.
+To ensure that the server is not overly constrained by this restriction, clients
+MUST send UDP datagrams with at least 1200 octets of payload until the server
+has completed address validation, see {{packet-size}}.
 
-In order to prevent this limit causing a handshake deadlock, the client SHOULD
-always send a packet upon a handshake timeout, as described in
+In order to prevent a handshake deadlock as a result of the server being unable
+to send, clients SHOULD send a packet upon a handshake timeout, as described in
 {{QUIC-RECOVERY}}.  If the client has no data to retransmit and does not have
 Handshake keys, it SHOULD send an Initial packet in a UDP datagram of at least
 1200 octets.  If the client has Handshake keys, it SHOULD send a Handshake
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1631,7 +1631,8 @@ dynamically calculate the expiration time.  It is also unlikely that the client
 port number is the same on two different connections; validating the port is
 therefore unlikely to be successful.
 
-A resumption token SHOULD be easily distinguishable from tokens that are sent in
+A resumption token SHOULD be constructed to be easily distinguishable from tokens
+that are sent in Retry packets as they are carried in the same field.
 Retry packets as they are carried in the same field.
 
 If the client has a token received in a NEW_TOKEN frame on a previous connection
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1496,193 @@ One way that a new format could be introduced is to define a TLS extension with
...
+During the initial handshake, QUIC requires that clients send UDP datagrams with
+at least 1200 octets of payload until the server has completed address
+validation. A server can thereby send more data to an unproven address without
+increasing the amplification advantage gained by an attacker.
+
+A server eventually confirms that a client has received its messages when the
+first Handshake-level message is received. This might be insufficient, either
+because the server wishes to avoid the computational cost of completing the
+handshake, or it might be that the size of the packets that are sent during the
+handshake is too large.  This is especially important for 0-RTT, where the
```
ianswett: is -> are

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1496,193 @@ One way that a new format could be introduced is to define a TLS extension with
...
+
+
+## Address Validation During Connection Establishment {#validate-new}
+
+During the initial handshake, QUIC requires that clients send UDP datagrams with
+at least 1200 octets of payload until the server has completed address
+validation. A server can thereby send more data to an unproven address without
+increasing the amplification advantage gained by an attacker.
+
+A server eventually confirms that a client has received its messages when the
```
ianswett: I'd drop eventually.

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1496,193 @@ One way that a new format could be introduced is to define a TLS extension with
...
+first Handshake-level message is received. This might be insufficient, either
```
ianswett: insufficient for what?

ianswett: Or maybe this whole paragraph should be removed, because it's really not easy to read for me and I'm not sure what it's attempting to add.

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1496,193 @@ One way that a new format could be introduced is to define a TLS extension with
...
+
+                                 Initial[0]: CRYPTO[SH] ACK[0]
+                       Handshake[0]: CRYPTO[EE, CERT, CV, FIN]
+                                 <- 1-RTT[0]: STREAM[1, "..."]
+~~~~
+{: #fig-retry title="Example Handshake Showing Retry"}
+
+A connection MAY be accepted without address validation - or with only limited
+validation - but a server SHOULD limit the data it sends toward an unvalidated
+address.  Successful completion of the cryptographic handshake implicitly
```
ianswett: I think this last sentence is slightly incorrect.  Given it's covered above, maybe remove it?  Also, the limit towards an unvalidated address is covered above, so if there's anything important being added here, maybe move it above?

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1496,193 @@ One way that a new format could be introduced is to define a TLS extension with
...
+provides proof that the client has received packets from the server.
+
+
+### Address Validation for Future Connections {#validation-future}
```
ianswett: Technically this is still token based address validation, so the previous heading was a bit misleading.  Possibly it should be renamed Retry based address validation?

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1496,193 @@ One way that a new format could be introduced is to define a TLS extension with
...
 
-A server that generates a response to an Initial packet without retaining
-connection state MUST use the Retry packet ({{packet-retry}}).  This packet
-causes a client to restart the connection attempt and includes the token in the
-new Initial packet ({{packet-initial}}) to prove source address ownership.
 
+### Address Validation Token Integrity {#token-integrity}
 
+There is no need for a single well-defined format for the token because the
+server that generates the token also consumes it.  A token could include
```
ianswett: If the token includes this information, should we say something like.  If any information is included, it SHOULD be encrypted to avoid spoofing or leaking of information?

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1496,193 @@ One way that a new format could be introduced is to define a TLS extension with
...
+information about the claimed client address (IP and port), a timestamp, and any
+other supplementary information the server will need to validate the token in
+the future.
+
+An address validation token MUST be difficult to guess.  Including a large
```
ianswett: This seems like the most critical requirement, so maybe it should be first?  Especially since the first and third paragraphs of this section flow together better than having this one in the middle.

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1513,191 @@ One way that a new format could be introduced is to define a TLS extension with
...
+{{validate-retry}}) or in a previous connection using the NEW_TOKEN frame (see
+{{validate-future}}).
+
+
+### Address Validation using Retry Packets {#validate-retry}
+
+QUIC uses token-based address validation during connection establishment.  Any
+time the server wishes to validate a client address, it provides the client with
+a token.  As long as it is not possible for an attacker to generate a valid
+token for its address (see {{token-integrity}}) and the client is able to return
```
```suggestion
token for its address (see {{token-integrity}}) and the client is able to return
token for its own address (see {{token-integrity}}) and the client is able to return
```

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1513,191 @@ One way that a new format could be introduced is to define a TLS extension with
...
+
+Unlike the token that is created for a Retry packet, there might be some time
+between when the token is created and when the token is subsequently used.
+Thus, a resumption token SHOULD include an expiration time.  The server MAY
+include either an explicit expiration time or an issued timestamp and
+dynamically calculate the expiration time.  It is also unlikely that the client
+port number is the same on two different connections; validating the port is
+therefore unlikely to be successful.
+
+A resumption token SHOULD be easily distinguishable from tokens that are sent in
```
```suggestion
A resumption token SHOULD be constructed to be easily distinguishable from tokens that
are sent in Retry packets as they are carried in the same field.
```

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1513,191 @@ One way that a new format could be introduced is to define a TLS extension with
...
+Retry packets as they are carried in the same field.
```
```suggestion
```

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1513,191 @@ One way that a new format could be introduced is to define a TLS extension with
...
 
-A server that generates a response to an Initial packet without retaining
-connection state MUST use the Retry packet ({{packet-retry}}).  This packet
-causes a client to restart the connection attempt and includes the token in the
-new Initial packet ({{packet-initial}}) to prove source address ownership.
+If the client has a token received in a NEW_TOKEN frame on a previous connection
+to what it believes to be the same server, it can include that value in the
+Token field of its Initial packet.
 
+A token allows a server to correlate activity between connections.
```
```suggestion
A token allows a server to correlate activity between the connection where the token
was issued and any connection where it is used.
```

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1513,191 @@ One way that a new format could be introduced is to define a TLS extension with
...
+Specifically, the connection where the token was issued, and any connection
```
```suggestion
```

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1513,191 @@ One way that a new format could be introduced is to define a TLS extension with
...
+where it is used.  Clients that want to break continuity of identity with a
```
```suggestion
Clients that want to break continuity of identity with a
```

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1513,191 @@ One way that a new format could be introduced is to define a TLS extension with
...
+server MAY discard tokens provided using the NEW_TOKEN frame.  Tokens obtained
+in Retry packets MUST NOT be discarded.
 
+A client SHOULD NOT reuse a token.  Reusing a token allows connections to be
```
```suggestion
A client SHOULD NOT use a token more than once.  Reusing a token allows connections to be
```

draft-ietf-quic-transport.md
```diff
@@ -1656,34 +1496,193 @@ One way that a new format could be introduced is to define a TLS extension with
...
+amplification attack.  In such an attack, a packet is sent to a server with
+spoofed source address information that identifies a victim.  If a server
+generates more or larger packets in response to that packet, the attacker can
+use the server to send more data toward the victim than it would be able to send
+on its own.
+
+The primary defense against amplification attack is verifying that an endpoint
+is able to receive packets at the transport address that it claims.  Address
+validation is performed both during connection establishment (see
+{{validate-new}}) and during connection migration (see {{migrate-validate}}).
```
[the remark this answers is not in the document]
martinthomson: In 7dc45966

PR #1888, opened 2018-10-19
Consolidate connection ID negotiation section with prior text

martinthomson:
The text here was mostly good, and I believe that this is a good place
for that text.  However, some of the text was expository in nature, and
I moved that up to the connection ID definition section.  There is some
new stuff in there, mostly just to point forward to the negotiation
section.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -927,16 +927,16 @@ this means the same connection ID MUST NOT be issued more than once on the same
 connection.
 
 Packets with long headers include at least two connection IDs, which are used to
-establish the connection IDs that are used for the connection when established,
+determine the connection IDs that are used for the connection when established,
 see {{negotiating-connection-ids}} for details.
 
 Packets with short headers ({{short-header}}) only include the Destination
 Connection ID and omit the explicit length.  The length of the Destination
-Connection ID field is expected to be known to endpoints.  Endpoints using a
-connection-ID based load balancer could agree with the load balancer on a fixed
-or minimum length and on an encoding for connection IDs.  This fixed portion
-could encode an explicit length, which allows the entire connection ID to vary
-in length and still be used by the load balancer.
+Connection ID field is expected to be known to endpoints.  Endpoints that use a
+load balancer that routes based on connection ID could agree with the load
+balancer on a fixed or minimum length and on an encoding for connection IDs.
+This fixed portion could encode an explicit length, which allows the entire
+connection ID to vary in length and still be used by the load balancer.
 
 A Version Negotiation ({{packet-version}}) packet echoes the connection IDs
 selected by the client, both to ensure correct routing toward the client and to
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -926,6 +926,23 @@ them with other connection IDs for the same connection.  As a trivial example,
 this means the same connection ID MUST NOT be issued more than once on the same
 connection.
 
+Packets with long headers include at least two connection IDs, which are used to
+establish the connection IDs that are used for the connection when established,
+see {{negotiating-connection-ids}} for details.
```
mikkelfj: Please consider rewording. The layers of establishment are confusing. Also, are the contained ID's used to establish themselves, or something else? If there are at least two, how many might that be in total, and why?

draft-ietf-quic-transport.md
```diff
@@ -926,6 +926,23 @@ them with other connection IDs for the same connection.  As a trivial example,
...
+
+Packets with short headers ({{short-header}}) only include the Destination
+Connection ID and omit the explicit length.  The length of the Destination
+Connection ID field is expected to be known to endpoints.  Endpoints using a
+connection-ID based load balancer could agree with the load balancer on a fixed
```
mikkelfj: connection-ID -> Connection ID

dtikhonov: I think what was meant here is "Connection ID-based load balancer" (dash between ID and based).  This would look odd because Connection ID is two words.  This can be reworded as "using a load balancer based on Connection IDs."

draft-ietf-quic-transport.md
MikeBishop: There can be three in a Retry.

ianswett: How about "Packets with long headers include two connection IDs in the header, which are used.."  If you want to clarify that Retry adds a third used for path verification, that would be better than saying "at least 2"

janaiyengar: "used" is used twice in the sentence, which doesn't reads well. Also agree with @ianswett that it is worth spelling out the Retry case instead of alluding to it.

draft-ietf-quic-transport.md
```diff
@@ -926,6 +926,23 @@ them with other connection IDs for the same connection.  As a trivial example,
...
 connection.
 
+Packets with long headers include at least two connection IDs, which are used to
+determine the connection IDs that are used for the connection when established,
+see {{negotiating-connection-ids}} for details.
+
+Packets with short headers ({{short-header}}) only include the Destination
+Connection ID and omit the explicit length.  The length of the Destination
+Connection ID field is expected to be known to endpoints.  Endpoints that use a
+load balancer that routes based on connection ID could agree with the load
```
janaiyengar: Run on sentence ("that ... that"). I prefer the earlier construction: "Endpoints using a connection-ID based load balancer could agree with the load	balancer on a fixed or minimum length and on an encoding for connection IDs."

draft-ietf-quic-transport.md
martinthomson: Not at this point in the document; this is too soon for that.  Tweaked.

draft-ietf-quic-transport.md
martinthomson: It has all the problems already highlighted, plus the precedence of the "or" and "and" toward the end is unclear.  I've tweaked again.

PR #1895, opened 2018-10-19
Minor edits to amplification attack text

martinduke:
I had a late review to PR #1824 because @ianswett had to get it into draft-15, and he encouraged me to file a PR. I believe he has no problem with at least one of these.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1198,7 +1198,7 @@ Prior to verifying the client's address, servers MUST NOT send more than three
 times as many bytes as the number of bytes received. Source addresses can be
 verified through an address validation token (delivered via a Retry packet or
 a NEW_TOKEN frame), by processing any message from the client encrypted using
-the Handshake keys, or by processing a TLS client hello with a valid
+the Handshake keys, or by processing a TLS ClientHello with a valid
 pre-shared key.  This limit exists to mitigate amplification attacks.
 
 In order to prevent this limit causing a handshake deadlock, the client MUST
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1189,17 +1189,18 @@ handshake is too large.  This is especially important for 0-RTT, where the
 server might wish to provide application data traffic - such as a response to a
 request - in response to the data carried in the early data from the client.
 
+A server can also verify the client address if the client's first Initial
+Packet contains a valid address validation token, previously delivered to the
+client in a RETRY packet or NEW_TOKEN frame.
+
 To send additional data prior to completing the cryptographic handshake, the
 server then needs to validate that the client owns the address that it claims.
 QUIC therefore performs source address validation during connection
 establishment.
 
 Prior to verifying the client's address, servers MUST NOT send more than three
-times as many bytes as the number of bytes received. Source addresses can be
-verified through an address validation token (delivered via a Retry packet or
-a NEW_TOKEN frame), by processing any message from the client encrypted using
-the Handshake keys, or by processing a TLS ClientHello with a valid
-pre-shared key.  This limit exists to mitigate amplification attacks.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1194,13 +1194,14 @@ server then needs to validate that the client owns the address that it claims.
...
 
-Servers MUST NOT send more than three times as many bytes as the number of bytes
-received prior to verifying the client's address.  Source addresses can be
+Prior to verifying the client's address, servers MUST NOT send more than three
+times as many bytes as the number of bytes received. Source addresses can be
 verified through an address validation token (delivered via a Retry packet or
-a NEW_TOKEN frame) or by processing any message from the client encrypted using
-the Handshake keys.  This limit exists to mitigate amplification attacks.
+a NEW_TOKEN frame), by processing any message from the client encrypted using
+the Handshake keys, or by processing a TLS client hello with a valid
```
marten-seemann: s/client hello/ClientHello

draft-ietf-quic-transport.md
```diff
@@ -1194,13 +1194,14 @@ server then needs to validate that the client owns the address that it claims.
...
-Servers MUST NOT send more than three times as many bytes as the number of bytes
-received prior to verifying the client's address.  Source addresses can be
+Prior to verifying the client's address, servers MUST NOT send more than three
+times as many bytes as the number of bytes received. Source addresses can be
 verified through an address validation token (delivered via a Retry packet or
-a NEW_TOKEN frame) or by processing any message from the client encrypted using
-the Handshake keys.  This limit exists to mitigate amplification attacks.
+a NEW_TOKEN frame), by processing any message from the client encrypted using
+the Handshake keys, or by processing a TLS ClientHello with a valid
+pre-shared key.  This limit exists to mitigate amplification attacks.
```
martinthomson: This last is not sufficient for address validation.

martinthomson: That isn't right either.  Willingness to complete the handshake is a decision the server makes, and it does not need to be based on a test of the liveness of the remote address.

It's possible that TLS resumption (0-RTT or not) could bundle a token in the session ticket that allows for address validation, as we used to do.  But we no longer require that to happen.

A server can accept 0-RTT from an unvalidated client address.  The amount that you can send is limited, but it's not unreasonable.  The server doesn't need to send a certificate, so it can probably send some responses.

ianswett: Given this is in the validation section, why not remove this entire sentence?

draft-ietf-quic-transport.md
```diff
@@ -1194,13 +1194,14 @@ server then needs to validate that the client owns the address that it claims.
 QUIC therefore performs source address validation during connection
 establishment.
 
-Servers MUST NOT send more than three times as many bytes as the number of bytes
-received prior to verifying the client's address.  Source addresses can be
+Prior to verifying the client's address, servers MUST NOT send more than three
```
ianswett: Q: Is this actually less awkward?  It's ok, but seems neutral to me.  Both sentences are a bit odd.

martinduke: "prior to verifying the client's address" could be construed as modifying "bytes received". When I remove all context for what I know about this, that's how I read the sentence. A new reader would have to reason a bit to understand what you mean, IMO.

if you're neutral I'll stick with it.

draft-ietf-quic-transport.md
martinduke: Ah, OK, it's my fundamental misunderstanding.

martinduke:
@martinthomson Where is your redo?

martinthomson:
I think that it was #1888.

martinduke:
@martinthomson You've covered the main points in the editor's draft. I'll close this one. I have one complaint, which I'll file a new PR for.

PR #1896, opened 2018-10-22
Move ECN section

martinthomson:
This moves the section under packetization and reliability.
I've also added subsections, splitting out the feedback system (which is
simple) with the verification stuff (which is less so).

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1856,7 +1856,7 @@ rate. Therefore, the endpoint resets its congestion controller, as described in
 {{migration-cc}}.
 
 The new path might not have the same ECN capability. Therefore, the endpoint
-verifies ECN capability as described in {{using-ecn}}.
+verifies ECN capability as described in {{ecn}}.
 
 Receiving acknowledgments for data sent on the new path serves as proof of the
 peer's reachability from the new address.  Note that since acknowledgments may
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2978,6 +2897,89 @@ The details of loss detection and congestion control are described in
...
+an endpoint MUST set the ECT(0) codepoint in the IP header of all outgoing
+packets {{!RFC8311}}.
+
+If an ECT codepoint set in the IP header is not corrupted by a network device,
+then a received packet contains either the codepoint sent by the peer or the
+Congestion Experienced (CE) codepoint set by a network device that is
+experiencing congestion.
+
+If a packet sent with an ECT codepoint is newly acknowledged by the peer in an
+ACK frame, the endpoint stops setting ECT codepoints in subsequent packets, with
```
```suggestion
ACK frame without ECN feedback, the endpoint stops setting ECT codepoints in
subsequent packets, with the expectation that either the network or the peer no
longer supports ECN.
```

draft-ietf-quic-transport.md
```diff
@@ -2978,6 +2897,89 @@ The details of loss detection and congestion control are described in
...
+the expectation that either the network or the peer no longer supports ECN.
```
```suggestion
```

draft-ietf-quic-transport.md
```diff
@@ -2978,6 +2897,89 @@ The details of loss detection and congestion control are described in
...
+
+If an endpoint receives a packet without an ECT or CE codepoint, it responds per
+{{processing-and-ack}} with an ACK frame.  If an endpoint does not have access
+to received ECN codepoints, it acknowledges received packets per
+{{processing-and-ack}} with an ACK frame.
+
+
+### ECN Verification
+
+Each endpoint independently verifies and enables use of ECN by setting the IP
```
janaiyengar: I think we should say something about when this verification can be done. It can be done any time during the connection, but it might make the most sense to do it along with the handshake. We may want to add some retransmission suggestions: if a handshake packet is detected as lost then send the retransmission without ECT bits. (Feel free to add a TODO here if you don't want to address this in this PR.)

I'm also wondering if we should add a fwd reference in the handshake section to here, with something like "An endpoint MAY combine the cryptographic handshake with ECN verification, as described in {{}}".

martinthomson: Happy to take other PRs as a follow-on.  Note that this is literally reordering text, so correcting any omission is strictly improvements on this.

PR #1899, opened 2018-10-23
The "quic " label is used for key update

### LINKED ISSUE #1845: Should the key update expansion use "quic " or "tls13 "?
I am looking at the details of the key update mechanism. The TLS draft says that "_A receiving endpoint detects an update when the KEY_PHASE bit doesn’t match what it is expecting. It creates a new secret (see Section 7.2 of [TLS13]) and the corresponding read key and IV._" That means using HKDF-Expand-Label.

I assume that we are going to use the prefix “quic “ rather than “tls13 “, since that's what we do at all the other places that use HKDF-Expand-Label. But the spec is silent. Maybe it should have a mention about the prefix for the label expansion?

martinthomson:
Closes #1845.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -992,7 +992,7 @@ An endpoint MUST NOT initiate more than one key update at a time.  A new key
 cannot be used until the endpoint has received and successfully decrypted a
 packet with a matching KEY_PHASE.
 
-A receiving endpoint detects an update when the KEY_PHASE bit doesn't match what
+A receiving endpoint detects an update when the KEY_PHASE bit does not match what
 it is expecting.  It creates a new secret (see Section 7.2 of {{!TLS13}}) and
 the corresponding read key and IV.  This uses the same variation on HKDF as
 defined in {{protection-keys}}; that is, the prefix "quic " is used in place of
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -994,7 +994,7 @@ packet with a matching KEY_PHASE.
 
 A receiving endpoint detects an update when the KEY_PHASE bit does not match what
 it is expecting.  It creates a new secret (see Section 7.2 of {{!TLS13}}) and
-the corresponding read key and IV.  This uses the same variation on HKDF as
+the corresponding read key and IV using the same variation on HKDF as
 defined in {{protection-keys}}; that is, the prefix "quic " is used in place of
 "tls13 ".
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -983,23 +983,31 @@ without necessarily needing to receive the first packet that triggered the
...
+This mechanism replaces the TLS KeyUpdate message.  Endpoints MUST NOT send a
+TLS KeyUpdate message.  Endpoints MUST treat the receipt of a TLS KeyUpdate
+message as a connection error of type 0x10a, equivalent to a fatal TLS alert of
+unexpected_message (see {{tls-errors}}).
+
 An endpoint MUST NOT initiate more than one key update at a time.  A new key
 cannot be used until the endpoint has received and successfully decrypted a
 packet with a matching KEY_PHASE.
 
 A receiving endpoint detects an update when the KEY_PHASE bit doesn't match what
```
```suggestion
A receiving endpoint detects an update when the KEY_PHASE bit does not match what
```

draft-ietf-quic-tls.md
```diff
@@ -983,23 +983,31 @@ without necessarily needing to receive the first packet that triggered the
...
-are also updated.  The next packet sent by the endpoint will then use the new
-keys.
-
-An endpoint doesn't need to send packets immediately when it detects that its
-peer has updated keys.  The next packet that it sends will simply use the new
-keys.  If an endpoint detects a second update before it has sent any packets
-with updated keys it indicates that its peer has updated keys twice without
-awaiting a reciprocal update.  An endpoint MUST treat consecutive key updates as
-a fatal error and abort the connection.
+the corresponding read key and IV.  This uses the same variation on HKDF as
```
```suggestion
the corresponding read key and IV using the same variation on HKDF as
```

draft-ietf-quic-tls.md
```diff
@@ -983,23 +983,31 @@ without necessarily needing to receive the first packet that triggered the
...
+defined in {{protection-keys}}; that is, the prefix "quic " is used in place of
+"tls13 ".
+
+If the packet can be decrypted and authenticated using the updated key and IV,
+then the keys the endpoint uses for packet protection are also updated.  The
+next packet sent by the endpoint will then use the new keys.
+
+An endpoint doesn't always need to send packets when it detects that its peer
```
```suggestion
An endpoint does not always need to send packets when it detects that its peer
```

draft-ietf-quic-tls.md
```diff
@@ -983,23 +983,31 @@ without necessarily needing to receive the first packet that triggered the
...
+has updated keys.  The next packet that it sends will simply use the new keys.
+If an endpoint detects a second update before it has sent any packets with
+updated keys it indicates that its peer has updated keys twice without awaiting
```
```suggestion
updated keys, it indicates that its peer has updated keys twice without awaiting
```

PR #1903, opened 2018-10-24
Fix type value for RETIRE_CONNECTION_ID frame

udippant:
The correct frame type value for RETIRE_CONNECTION_ID, as listed in page 63 (https://tools.ietf.org/html/draft-ietf-quic-transport-16#page-63) is 0x0d. However, in the section that describes the structure of RETIRE_CONNECTION_ID, it says 0x1b, which is conflicts with the type values for Ack Frame.


martinthomson:
Oops, already fixed this.  Thanks for the PR.

udippant:
Thanks for fixing. This typo did indeed confuse me for several minutes ;)

PR #1911, opened 2018-10-25
Don't amplify accidentally when closing

### LINKED ISSUE #1905: Limit retransmission of closing packet
@mikkelfj observes that we permit rate limiting of connection close frames, but they might be used to create an amplification attack if the closing endpoint is naive in its handling.  The fix is easy - the endpoint needs to remember a validated address for its peer along with the closing packet.  Packets from other remote addresses need to be dropped or only responded to if the size of the incoming packet is > 1/3 of the closing packet.  In most cases, this won't change anything.

martinthomson:
A minor point, and we might not even regret this, but it's not worth
risking it.

Closes #1905.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2230,9 +2230,9 @@ Note:
 
 New packets from unverified addresses could be used to create an amplification
 attack (see {{address-validation}}).  To avoid this, endpoints MUST either limit
-transmission of closing frames to validated addresses - which implies retaining
-knowledge of validated addresses - or to drop packets without response if the
-response would be more than three times larger than the received packet.
+transmission of closing frames to validated addresses or to drop packets without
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2228,6 +2228,12 @@ Note:
   control, which are not expected to be relevant for a closed connection.
   Retransmitting the final packet requires less state.
 
+New packets from unverified addresses could be used to create an amplification
+attack (see {{address-validation}}).  To avoid this, endpoints MUST either limit
+transmission of closing frames to validated addresses - which implies retaining
```
ianswett: I'd remove the "- which implies retraining knowledge of validated addresses -" because it makes the sentence a bit awkward and I think it's unnecessary.

PR #1912, opened 2018-10-25
Use octets rather than bytes throughout

### LINKED ISSUE #1863: Clarify amplification limit
On the transport text: "Servers MUST NOT send more than three times as many bytes as the number of bytes received prior to verifying the client's address. "

The text does not state what is meant by bytes received. Presumably this refers to the accumulated QUIC packet sizes excluding IP/UDP framing. Vneg and Retry is a bit vague, do you count those earlier attempts, if you can? Can you use information gained from earlier connections and if so, for how long.

Note: generally the term "octet" is being used, but there are other places where "byte" is used.

martinthomson:
I have a feeling that Ian isn't going to like this because it messes with the notion of "bytes in flight" that we use in -recovery.  I would prefer to be consistent though.  For the transport, "bytes in flight" only appeared once, but its use wasn't critical.

Also closes #1863.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -195,7 +195,7 @@ Endpoint:
 
 Stream:
 
-: A logical unidirectional or bidirectional channel of ordered bytes within a
+: A logical unidirectional or bidirectional channel of ordered octets within a
   QUIC connection.
 
 Connection:
@@ -753,11 +753,11 @@ MAY send MAX_STREAM_DATA frames in multiple packets in order to make sure that
 the sender receives an update before running out of flow control credit, even if
 one of the packets is lost.
 
-Connection flow control is a limit to the total bytes of stream data sent in
+Connection flow control is a limit to the total octets of stream data sent in
 STREAM frames on all streams.  A receiver advertises credit for a connection by
-sending a MAX_DATA frame.  A receiver maintains a cumulative sum of bytes
+sending a MAX_DATA frame.  A receiver maintains a cumulative sum of octets
 received on all contributing streams, which are used to check for flow control
-violations. A receiver might use a sum of bytes consumed on all streams to
+violations. A receiver might use a sum of octets consumed on all streams to
 determine the maximum data limit to be advertised.
 
 A receiver MAY advertise a larger offset at any point by sending MAX_STREAM_DATA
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -618,8 +617,8 @@ sent_packets:
 : An association of packet numbers to information about them, including a number
   field indicating the packet number, a time field indicating the time a packet
   was sent, a boolean indicating whether the packet is ack-only, a boolean
-  indicating whether it counts towards bytes in flight, and a bytes
-  field indicating the packet's size.  sent_packets is ordered by packet number,
+  indicating whether it counts towards bytes in flight, and a size field
```
mikkelfj: It says bytes in flight

ianswett:
Q: Why not standardize on bytes?  All the TCP RFCs use byte, and the word octet does not occur in any TCP RFCs I can find.

nibanks:
I would also prefer `bytes` over `octets`, if we are taking a vote.

dtikhonov:
I vote for **bytes**.

mikkelfj:
bytes are not quantifiable. They could be 6-bit (as if I care).

Why not also little endian now we are at it?

martinthomson:
Right now, we're inconsistent.  I don't believe that there is any problem along the lines @mikkelfj describes with using bytes - that ship sailed a long time ago - but HTTP/2 went with octets and we've been inconsistent.  I can easily do a search and replace in the other direction; and I don't personally care.

mikkelfj:
then bytes it is - I have never heard of bandwidth measured in MOps

MikeBishop:
@mikkelfj, see [here](https://www.sospc2424.ch/site/images/web4glossary/image_vignette2/Unite_mesure_Conversion-5557698e4fa77.jpg), for example.  Capacities are described in Go, Mo, etc. in French.  Bandwidth is usually still in bits, though.

Bytes have historically varied in size.  The C and C++ standards define a byte as being **at least** eight bits in size.  (See their FAQ [here](https://isocpp.org/wiki/faq/intrinsic-types#very-large-bytes).)  I'm very much in favor of retaining octets, because that's not a platform-specific term.

mikkelfj:
@MikeBishop yeah - I discovered the french connection subsequently but more often MOps means mega outputs pers second or mega operations per second or similar.

PR #1913, opened 2018-10-25
Clarify ECN counters

### LINKED ISSUE #1838: ECN counters and Coalescing of QUIC packets
I realize there exist a need to clarify the text regarding ECN on the relation between IP packets and QUIC packets. As QUIC tracks QUIC packets, and not IP packets, this relation when coaleing occurs needs to be clarified. What I think is needed are:
- In Section 6.8: Using Explicit Congestion Notification should be updated to be explicit about when it talks about IP packet and when about QUIC packets. 
- State explicitly that for coalesed QUIC packets, the IP packet's ECN field value is reported for each QUIC packet.

martinthomson:
This deals with the distinction between IP and QUIC packets better,
clarifies the interaction with packet coalescing, and makes it clear
that counts are ACK state and therefore separated by packet number
space.

Closes #1838.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2941,10 +2941,20 @@ A packet detected by a receiver as a duplicate does not affect the receiver's
 local ECN codepoint counts; see ({{security-ecn}}) for relevant security
 concerns.
 
-If an endpoint receives a packet without an ECT or CE codepoint, it responds per
-{{processing-and-ack}} with an ACK frame.  If an endpoint does not have access
-to received ECN codepoints, it acknowledges received packets per
-{{processing-and-ack}} with an ACK frame.
+If an endpoint receives a QUIC packet without an ECT or CE codepoint, it
+responds per {{processing-and-ack}} with an ACK frame without increasing any ECN
+counters.  Similarly, if an endpoint does not have access to received ECN
+codepoints, it does not increase ECN counters.
+
+Coalesced packets (see {{packet-coalesce}}) mean that several packets can share
+the same IP header.  ECN counters count each QUIC packet, not the enclosing IP
+packet or UDP datagram.
+
+Each packet number space maintains separate acknowledgement state and separate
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2941,10 +2941,20 @@ A packet detected by a receiver as a duplicate does not affect the receiver's
...
+
+Coalesced packets (see {{packet-coalesce}}) mean that several packets can share
+the same IP header.  ECN counters count each QUIC packet, not the enclosing IP
+packet or UDP datagram.
+
+Each packet number space maintains separate acknowledgement state and separate
+ECN counters.  For example, if one each of an Initial, 0-RTT, Handshake, and
+1-RTT QUIC packet are coalesced, the corresponding counters for the Initial and
+Handshake packet number space will be incremented by one and the counters for
+the 1-RTT packet number space will be increased by two.
```
ianswett: Oh, well that's weird.

martinthomson:
Thanks for reviewing @gloinul,

> I wonder if we have the potential of conflict on ECN codepoints to set when coalesing packets. I think the main issue is if the Handshake number space determines that ECN is not working, and then a 1-RTT packet has been generated but not sent yet, that made the assumption that the outgoing IP header shall be ECT.

Yes, this is a tricky one, but I think at some point we have to leave this sort of thing to implementations.  If you are coalescing and you have concluded differently for Handshake and 1-RTT, then you have to make a hard call.  An endpoint that receives mixed signals about bleaching might be better off assuming that there is bleaching on the path.  I think that we already have text that supports that conclusion.

BTW, for IP-layer fragmentation and reassembly where fragments come in with different markings, I am assuming here that there are rules about reassembly elsewhere that deal with this case and we don't need to concern ourselves with that in this document.

martinthomson:
Thanks Magnus!

draft-ietf-quic-transport.md
MikeBishop: Consistent with our mantra that coalesced packets are treated just as if each was received separately, though.

PR #1914, opened 2018-10-25
Stream description moved to section 2 of transport

LPardue:


martinthomson:
Fixed in #1910 more comprehensively.

martinthomson:
OBE.

PR #1915, opened 2018-10-25
There are *Two* control streams

LPardue:
:picardmeme:

A change is not needed globally, I think these two cases are good enough to help avoid confusion.


MikeBishop:
Fine by me.  There's only one in each direction, so I don't think the singular is *wrong*, but I agree this is probably more correct.

PR #1916, opened 2018-10-25
Relax 1200B requirement for Initial ACKs/Retx

martinduke:
IMO it's silly to pad out a packet that contains the client's initial and handshake acks, because if delivered it verifies the address.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1556,9 +1556,10 @@ times as many bytes as the number of bytes they have received.  This limits the
 magnitude of any amplification attack that can be mounted using spoofed source
 addresses.
 
-To ensure that the server is not overly constrained by this restriction, clients
-MUST send UDP datagrams with at least 1200 octets of payload until the server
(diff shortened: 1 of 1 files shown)


mikkelfj:
More generally, it would be good if the 1200B restraint could be entirely replaced by increased latency by sending a probe packet of sorts.

IoT devices might want to have a smaller tranmissions buffer.

I'm not sure how much TLS overhead can be reduced, but that could be a larger cost.

martinthomson:
I was going to edit this directly, but it was a bigger change.  Does https://github.com/martinduke/base-drafts/pull/1 work for you @martinduke ?

martinthomson:
@mikkelfj, in practice, there is a probe.  The server sends what it can, which - in most of the cases where it is limited - includes a Handshake packet.  That Handshake packet is the probe.  The acknowledgment of that Handshake packet is the response to that probe.  It's implicit, that's all.

martinduke:
My github-fu is weak, but I believe this is ready to merge to master.

PR #1923, opened 2018-10-26
Try to clarify interaction of PADDING and ACK

### LINKED ISSUE #1827: PLPMTUD section claims that PADDING frames generate ACKs
Section 8.4.2. of the transport draft (Special Considerations for Packetization Layer PMTU Discovery) says:
"The PADDING frame provides a useful option for PMTU probe packets. PADDING frames generate acknowledgements, but they need not be delivered reliably."

Meanwhile Section 8.1. of the transport draft (Packet Processing and Acknowledgment) contains language suggesting that PADDING frames do not generate acknowledgements. Also I believe this was the conclusion reached in older issues about PADDING frames generating ACKs so Section 8.4.2 is in error.

It may also be helpful, in Section 8.4.2, to state that PING+PADDING is useful for PMTU probe packets, as that will generate an ACK while allowing the packet to be inflated to whatever size necessary.

martinthomson:
The text wasn't wrong, but this is shorter and - I think - clearer.

Closes #1827.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2772,7 +2772,7 @@ there are packet gaps which precede the received packet.  The endpoint MUST
 however acknowledge packets containing only ACK or PADDING frames when sending
 ACK frames in response to other packets.
 
-While PADDING frames are only eventually acknowledged, they are still considered
+Packets containing PADDING frames are considered
 to be in flight for congestion control purposes {{QUIC-RECOVERY}}. Sending only
 PADDING frames might cause the sender to become limited by the congestion
 controller (as described in {{QUIC-RECOVERY}}) with no acknowledgments
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2772,13 +2772,13 @@ there are packet gaps which precede the received packet.  The endpoint MUST
 however acknowledge packets containing only ACK or PADDING frames when sending
 ACK frames in response to other packets.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2772,13 +2772,13 @@ there are packet gaps which precede the received packet.  The endpoint MUST
...
 ACK frames in response to other packets.
 
-While PADDING frames do not elicit an ACK frame from a receiver, they are
-considered to be in flight for congestion control purposes
-{{QUIC-RECOVERY}}. Sending only PADDING frames might cause the sender to become
-limited by the congestion controller (as described in {{QUIC-RECOVERY}}) with no
-acknowledgments forthcoming from the receiver. Therefore, a sender should ensure
-that other frames are sent in addition to PADDING frames to elicit
-acknowledgments from the receiver.
+While PADDING frames are only eventually acknowledged, they are still considered
```
nibanks: The word `only` here is awkward to me. I think it would be best removed.

```suggestion
While PADDING frames may never be acknowledged, they are still considered
```

martinthomson: This is hard.  never acknowledged isn't true either.  The point I'm trying to capture is that a packet containing PADDING doesn't cause the recipient to start their delayed acknowledgment timer.  Keep in mind that this isn't critical language.  Maybe instead.

```suggestion
Packets containing PADDING frames are still considered
```

PR #1925, opened 2018-10-27
define HTTP_UNEXPECTED_GOAWAY

### LINKED ISSUE #1918: Section 4.2.17 references HTTP_UNEXPECTED_GOAWAY which doesn't exist
HTTP_UNEXPECTED_GOAWAY is referenced in section 4.2.17 as the connection error to generate when getting an unexpected goaway, but this is not defined in section 8.

lnicco:
using the first available ID 0x0013

fixes #1918

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1337,6 +1337,9 @@ HTTP_EARLY_RESPONSE (0x0011):
 HTTP_MISSING_SETTINGS (0x0012):
 : No SETTINGS frame was received at the beginning of the control stream.
 
+HTTP_UNEXPECTED_GOAWAY (0x0013):
+: A GOAWAY frame was received from the client.
+
 HTTP_GENERAL_PROTOCOL_ERROR (0x00FF):
 : Peer violated protocol requirements in a way which doesn't match a more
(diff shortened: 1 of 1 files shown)


martinthomson:
My preference is not to create specific error codes for this, but instead to rely on a more generic error code, but I will let @MikeBishop make that call.

MikeBishop:
Yeah, I'd like to be slightly more generic in situations that have multiple different variants, though probably not as generic as Martin would have me be.

We have HTTP_WRONG_STREAM_DIRECTION for cases where unidirectional streams open in a direction their definition doesn't permit (e.g. clients initiating push streams); maybe we need a HTTP_WRONG_FRAME_DIRECTION for GOAWAY and MAX_PUSH_ID direction issues?

lnicco:
another interesting case is when a second SETTINGS frame gets received, right now we return MALFORMED_FRAME, which is a bit misleading since the frame could be perfectly formed, it's just showing at the wrong time in the wrong place :) 

> If an endpoint receives a second SETTINGS frame, the endpoint MUST respond with a connection error of type HTTP_MALFORMED_FRAME.

how about HTTP_UNEXPECTED_FRAME to catch all the above cases ?

MikeBishop:
That works.

PR #1926, opened 2018-10-29
Cite limitations on key usage

### LINKED ISSUE #1457: Clarify and strengthen key update requirements
From several discussions it appears that the importance of key updates is not well understand and the consequences can be fatal. The requirements are in place via numerous indirect links over TLS 1.3 spec and further documents.

Some crypto modes can handle a large number of packets safely while others break down statistically, including AES-GCM.

A solution could be to require key updates no later than after 2^32 packets and require a protocol error shutdown if peer does not rekey in time. While 2^32 may be early in some cases, it is not really a burden, and the alternative might be that implementations skip handling key updates.

See also discussion here:
https://github.com/quicwg/base-drafts/issues/1405#issuecomment-398022746

martinthomson:
Closes #1457.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1042,6 +1042,9 @@ compromise and an attack, or if the peer is incorrectly reverting to use of old
 keys.  Because the latter cannot be differentiated from an attack, an endpoint
 MUST immediately terminate the connection if it detects this condition.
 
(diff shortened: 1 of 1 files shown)


martinthomson:
This generally isn't a hard edge, so - as in TLS - you can simply cause your peer to update if you think they are running too close to the line.  I don't think that we need an error code in that case.  Nor do I want tons of text about how to manage this.

PR #1927, opened 2018-10-29
Flow control hazards

### LINKED ISSUE #1432: Length-prefixes and flow control
HTTP/QUIC has a particular example of a pattern that seems likely to be usable in many other contexts.  A block of data is preceded by a length field.  The receiver of that block might read the length field then await the arrival of the entire block before processing the block.  During that time, the block is left unread so that the data can sit in the receive buffer of the transport.  This ensures that the data counts towards the flow control limits of the receiver.  This can prevent the sender from adding more messages that would only consume additional memory at the receiver.

In an extreme case, this leads to a deadlock.  The receiver is waiting for the end of a block.  The sender cannot send the remainder of the block because it does not have enough flow control credit.

This impasse doesn't naturally resolve in this case.  Worse, even if flow control might allow a single stream to make progress because the receiver ensures that it provides credit for the largest possible message or more, multiple concurrent streams could each consume a small part of the available flow control credit without allowing any single one to make progress.

Ultimately, there isn't an easy fix for this.  We could recommend a new input for prioritization: favour the sending of data for streams that are in the middle of sending a chunk, but that requires the transport be informed of message boundaries.  Relying on RST_STREAM to cancel streams until the logjam clears might work in some cases, but we have some examples already of streams that cannot be safely cancelled.  So this will always be a danger.  We gain some insight into what is going on with BLOCKED and friends, but ultimately this is just *hard*.

Not sure how much advice we can give here, or whether saying anything is even needed.

martinthomson:
This describes some of the risks entailed by flow control.  I'm starting
to think that this sort of language is best moved to the applicability
statement, but I'd like other opinions on that.

I'm also not completely happy with how this turned out.  This is just
the first rough cut.  Concrete suggestions for how to improve this would
be greatly appreciated.

Closes #1432.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -924,7 +924,7 @@ means of holding a peer accountable for the state commitment for partially
 processed messages.
 
 Deadlocking can also occur if data on different streams is interdependent.  In
-this case, data on one stream arrives before the data on a second stream on
+Suppose that data on one stream arrives before the data on a second stream on
 which it depends.  A deadlock can occur if first stream cannot be read and that
 prevents flow control limits for the second stream being increased.  To reduce
 the likelihood of deadlock for interdependent data, implementations can ensure
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -898,7 +898,8 @@ design.
 
 Large messages can produce deadlocking if the recipient does not process the
 message incrementally.  If the message is larger than flow control credit
-available and the recipient does not release additional flow control credit, a
+available and the recipient does not release additional flow control credit
+until the entire message is received and delivered, a
 deadlock can occur.  This is possible even where stream flow control limits are
 not reached because connection flow control limits can be consumed by other
 streams.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -930,7 +930,8 @@ prevents the receiver from extending flow control credit for the second stream.
 To reduce
 the likelihood of deadlock for interdependent data, the application sender
 should ensure
-that data is not sent until the data it depends on has consumed both stream- and
+that dependent data is not sent until the data it depends on has been accounted
+for in both stream- and connection- level flow control credit.
 connection- level flow control credit.
 
 Some deadlocking scenarios might be resolved by cancelling affected streams with
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -884,6 +884,54 @@ provide an interface to QUIC to tell it about its buffering limits so that there
 is not excessive buffering at multiple layers.
 
 
+## Flow Control Deadlocks {#flow-control-deadlocks}
+
+Flow control provides a means of managing access to the limited buffers
+endpoints have for incoming data.  This mechanism limits the amount of data that
+can be in buffers in endpoints or in transit on the network.  However, there are
+several ways in which limits can produce conditions that can cause a connection
+to either perform suboptimally or deadlock.
+
+Implementations need to be aware of the possibility of deadlocking and that some
+conditions that lead to deadlocking cannot be completely avoided in a protocol
+design.
+
+Large messages can produce deadlocking if the recipient does not process the
+message incrementally.  If the message is larger than flow control credit
+available and the recipient does not release additional flow control credit, a
+deadlock can occur.  This is possible even where stream flow control limits are
+not reached because connection flow control limits can be consumed by other
+streams.
+
+A common implementation technique has flow control credit extended as data is
+read from receive buffers.  In that setting, a length-prefixed message format
+makes it easier for a recipient to leave data unread and thereby withhold flow
+control credit.  If flow control limits prevent the remainder of a message from
+being sent, a deadlock will result.  However, a length prefix might enable the
+detection of this sort of deadlocks.  Where protocols have messages that might
+be processed as a single unit, reserving flow control credit for the entire
+message atomically ensures that this style of deadlock is less likely.
+
+Even a message as small as two bytes could produce a deadlock.  Though the
+chances of deadlock are reduced for smaller messages, the only way to prevent
+deadlock is to read every byte as it becomes available.  However, incremental
+processing of messages can greatly increase complexity of implementations.
+Also, releasing flow control credit might mean that an endpoint might need other
+means of holding a peer accountable for the state commitment for partially
+processed messages.
+
+Deadlocking can also occur if data on different streams is interdependent.  In
+this case, data on one stream arrives before the data on a second stream on
+which it depends.  A deadlock can occur if first stream cannot be read and that
+prevents flow control limits for the second stream being increased.  To reduce
+the likelihood of deadlock for interdependent data, implementations can ensure
+that data is not sent until the data it depends on has consumed both stream- and
+connection- level flow control credit.
+
+Some deadlocking scenarios might be resolved by cancelling affected streams with
+STOP_SENDING or RST_STREAM, if the protocol permits that.
+
+
 ## Stream Limit Increment {#stream-limit-increment}
 
 An endpoint limits the number of concurrently active incoming streams by
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -904,7 +904,9 @@ deadlock can occur.  This is possible even where stream flow control limits are
 not reached because connection flow control limits can be consumed by other
 streams.
 
-A common implementation technique has flow control credit extended as data is
+A common flow control implementation technique is for a receiver to extend
+credit to the sender as a the data consumer reads data out of the receiver's
+buffers. In this setting, a length-prefixed message format
 read from receive buffers.  In that setting, a length-prefixed message format
 makes it easier for a recipient to leave data unread and thereby withhold flow
 control credit.  If flow control limits prevent the remainder of a message from
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -928,7 +928,8 @@ Suppose that data on one stream arrives before the data on a second stream on
 which it depends.  A deadlock can occur if first stream cannot be read and that
 prevents the receiver from extending flow control credit for the second stream.
 To reduce
-the likelihood of deadlock for interdependent data, implementations can ensure
+the likelihood of deadlock for interdependent data, the application sender
+should ensure
 that data is not sent until the data it depends on has consumed both stream- and
 connection- level flow control credit.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,54 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+
+Large messages can produce deadlocking if the recipient does not process the
+message incrementally.  If the message is larger than flow control credit
+available and the recipient does not release additional flow control credit, a
+deadlock can occur.  This is possible even where stream flow control limits are
+not reached because connection flow control limits can be consumed by other
+streams.
+
+A common implementation technique has flow control credit extended as data is
+read from receive buffers.  In that setting, a length-prefixed message format
```
ianswett: I'm confused about why this is so.  The only thing I can think of is if the payload is too big, you can reset/stop_sending the stream.

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,54 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+makes it easier for a recipient to leave data unread and thereby withhold flow
+control credit.  If flow control limits prevent the remainder of a message from
+being sent, a deadlock will result.  However, a length prefix might enable the
+detection of this sort of deadlocks.  Where protocols have messages that might
+be processed as a single unit, reserving flow control credit for the entire
```
ianswett: Connection level flow control can't be reduced, so are you suggesting it not be further increased until the message arrives?  I haven't implemented this sort of logic, but it seems very tricky to get right and done incorrectly, could result in deadlocks.

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,54 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+message atomically ensures that this style of deadlock is less likely.
+
+Even a message as small as two bytes could produce a deadlock.  Though the
```
ianswett: This paragraph isn't adding anything for me.

draft-ietf-quic-transport.md
MikeBishop: Concrete example:  You can't consume less than a full QPACK instruction on the encoder stream.  If you read far enough to see that you have a 64KB auth header coming at you, you might logically not continue reading until you've received those 64KB.  If there's not sufficient flow control for the sender to complete sending, nothing gets read and you deadlock.

That might be because the connection flow control never gets increased, or it might be because it got increased but allocated to other streams that have data to send.

draft-ietf-quic-transport.md
MikeBishop: I agree, except that the last sentence is important.

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,54 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+Deadlocking can also occur if data on different streams is interdependent.  In
+this case, data on one stream arrives before the data on a second stream on
+which it depends.  A deadlock can occur if first stream cannot be read and that
+prevents flow control limits for the second stream being increased.  To reduce
+the likelihood of deadlock for interdependent data, implementations can ensure
+that data is not sent until the data it depends on has consumed both stream- and
+connection- level flow control credit.
+
+Some deadlocking scenarios might be resolved by cancelling affected streams with
+STOP_SENDING or RST_STREAM, if the protocol permits that.
```
MikeBishop: I'd change "if the protocol permits that" to "but depending on the application protocol, this might result in terminating the connection."

draft-ietf-quic-transport.md
martinthomson: I'm not following your comment here.

draft-ietf-quic-transport.md
martinthomson: Suggestions sought.  I don't think that I can just leave the last sentence without the preface.

draft-ietf-quic-transport.md
dtikhonov: > Concrete example: You can't consume less than a full QPACK instruction on the encoder stream.

Yes, you can. See [lsqpack_dec_enc_in()](https://github.com/litespeedtech/ls-qpack/blob/5f17c82e8c0f3b2b2faa8984256ff355bc5dc657/src/lsqpack.c#L4186)

afrind: I think he's just giving an example of an implementation that cannot handle this.  You are correct that the protocol does not mandate this specific behavior.

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,55 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+
+Flow control provides a means of managing access to the limited buffers
+endpoints have for incoming data.  This mechanism limits the amount of data that
+can be in buffers in endpoints or in transit on the network.  However, there are
+several ways in which limits can produce conditions that can cause a connection
+to either perform suboptimally or deadlock.
+
+Implementations need to be aware of the possibility of deadlocking and that some
+conditions that lead to deadlocking cannot be completely avoided in a protocol
+design.
```
janaiyengar: I don't think this is accurate. The deadlock problem is not within a protocol here, but in how two protocol implementations (HTTP and QUIC) might interact.

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,55 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+
+Large messages can produce deadlocking if the recipient does not process the
+message incrementally.  If the message is larger than flow control credit
+available and the recipient does not release additional flow control credit, a
```
```suggestion
available and the recipient does not release additional flow control credit
until the entire message is received and delivered, a
```

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,55 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+deadlock can occur.  This is possible even where stream flow control limits are
+not reached because connection flow control limits can be consumed by other
+streams.
+
+A common implementation technique has flow control credit extended as data is
```
```suggestion
A common flow control implementation technique is for a receiver to extend
credit to the sender as a the data consumer reads data out of the receiver's
buffers. In this setting, a length-prefixed message format
```

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,55 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+read from receive buffers.  In that setting, a length-prefixed message format
```
```suggestion
```

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,55 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+makes it easier for a recipient to leave data unread and thereby withhold flow
```
```suggestion
makes it easier for the data consumer to leave data unread in the receiver's
buffers and thereby withhold flow
```

draft-ietf-quic-transport.md
janaiyengar: How about "A data consumer can read bytes out of a receiver's buffer as they become available to cause the receiver to extend flow control credit to the sender and reduce chances of a deadlock. However, releasing flow control credit might mean that the data consumer might need other means of holding a peer accountable for the state commitment for partially processed messages."

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,55 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+
+Even a message as small as two bytes could produce a deadlock.  Though the
+chances of deadlock are reduced for smaller messages, the only way to prevent
+deadlock is to read every byte as it becomes available.  However, incremental
+processing of messages can greatly increase complexity of implementations.
+Also, releasing flow control credit might mean that an endpoint might need other
+means of holding a peer accountable for the state commitment for partially
+processed messages.
+
+Deadlocking can also occur if data on different streams is interdependent.  In
```
```suggestion
Deadlocking can also occur if a data consumer introduces dependencies between
data sent and received on different streams.
```

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,55 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+this case, data on one stream arrives before the data on a second stream on
+which it depends.  A deadlock can occur if first stream cannot be read and that
+prevents flow control limits for the second stream being increased.  To reduce
```
```suggestion
prevents the receiver from extending flow control credit for the second stream.
To reduce
```

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,55 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+the likelihood of deadlock for interdependent data, implementations can ensure
```
```suggestion
the likelihood of deadlock for interdependent data, the application sender
should ensure
```

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,55 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+that data is not sent until the data it depends on has consumed both stream- and
```
```suggestion
that dependent data is not sent until the data it depends on has been accounted
for in both stream- and connection- level flow control credit.
```

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,55 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+connection- level flow control credit.
```
```suggestion
```

draft-ietf-quic-transport.md
```diff
@@ -884,6 +884,55 @@ provide an interface to QUIC to tell it about its buffering limits so that there
...
+Even a message as small as two bytes could produce a deadlock.  Though the
+chances of deadlock are reduced for smaller messages, the only way to prevent
+deadlock is to read every byte as it becomes available.  However, incremental
+processing of messages can greatly increase complexity of implementations.
+Also, releasing flow control credit might mean that an endpoint might need other
+means of holding a peer accountable for the state commitment for partially
+processed messages.
+
+Deadlocking can also occur if data on different streams is interdependent.  In
+this case, data on one stream arrives before the data on a second stream on
```
```suggestion
Suppose that data on one stream arrives before the data on a second stream on
```

draft-ietf-quic-transport.md
martinthomson: My point is that this is not a matter for protocol design, because no protocol can defend against this effectively in its design and that implementations need to handle the issue.  I'll reword.

... and note that this just adds more weight to moving this to applicability :)

draft-ietf-quic-transport.md
martinthomson: Thanks, that's much better.

draft-ietf-quic-transport.md
martinthomson: I think that I want to keep this abstract, because it's not always the consumer that establishes the dependencies.  QPACK does this intentionally.  It's true that a consumer withholding credit is what causes the deadlock, but I'd rather say that differently.

martinthomson:
Closing in favour of quicwg/ops-drafts#53

PR #1930, opened 2018-10-29
Replace end of recovery with recovery_start_time

### LINKED ISSUE #1928: Need to clarify whether end_of_recovery have to be tracked on each PN spaces separately
On section 5.8.2 and 5.8.6, the spec suggests that you should track the largest packet number sent when QUIC detects a loss, and the pseudo code checks the packet number in `CongestionEvent`. Since there are multiple packet number spaces, it seems like `end_of_recovery` should be a vector and largest packet numbers should be tracked separately on each PN spaces. If it was true, it should be clarified on 5.8.2, otherwise it's hard to interpret I think.

However, if we managed them separately, `congestion_window` which is shared with all PN spaces will be decreased on every congestion event on each PN spaces. Although it still works, the result (behavior) would be different from one packet number space scenario.

If this change is not intentional and we want to keep the original behavior, we probably need to introduce one more packet numbering system that would be shared with all PN spaces to track the **last** packet sent.

5.8.2.  Variables of interest
```
end_of_recovery:  The largest packet number sent when QUIC detects a
      loss.  When a larger packet is acknowledged, QUIC exits recovery.
```

5.8.6.  On New Congestion Event
```
   Invoked from ProcessECN and OnPacketsLost when a new congestion event
   is detected.  Starts a new recovery period and reduces the congestion
   window.

      CongestionEvent(packet_number):
        // Start a new congestion event if packet_number
        // is larger than the end of the previous recovery epoch.
        if (!InRecovery(packet_number)):
          end_of_recovery = largest_sent_packet
          congestion_window = max(congestion_window, kMinimumWindow)
          ssthresh = congestion_window

```

ianswett:
Fixes #1928

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1047,9 +1047,9 @@ bytes_in_flight:
 congestion_window:
 : Maximum number of bytes-in-flight that may be sent.
 
-end_of_recovery:
-: The largest packet number sent when QUIC detects a loss.  When a larger
-  packet is acknowledged, QUIC exits recovery.
+recovery_start_time:
+: The time when QUIC first detects a loss, causing it to enter recovery.
+  When a packet sent after this time is acknowledged, QUIC exits recovery.
 
 ssthresh:
 : Slow start threshold in bytes.  When the congestion window is below ssthresh,
@@ -1064,7 +1064,7 @@ variables as follows:
 ~~~
    congestion_window = kInitialWindow
    bytes_in_flight = 0
-   end_of_recovery = 0
+   start_time_of_recovery = 0
    ssthresh = infinite
    ecn_ce_counter = 0
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -1130,8 +1130,9 @@ Invoked when an ACK frame with an ECN section is received from the peer.
      if (ack.ce_counter > ecn_ce_counter):
        ecn_ce_counter = ack.ce_counter
        // Start a new congestion event if the last acknowledged
-       // packet is past the end of the previous recovery epoch.
-       CongestionEvent(ack.largest_acked_packet)
+       // packet was sent after the start of the previous
+       // recovery epoch.
+       CongestionEvent(sent_packets[ack.largest_acked].time)
```
maskit: I'm not sure if this is really an issue but `sent_packets` is a variable described in Loss Detection section. Other places except `OnRetransmissionTimeoutVerified ` don't use any of the variables described in Loss Detection section. This makes dependency for Loss Detection from Congestion Control.

If this function receive a value of `sent_packets[ack.largest_acked].time` as a parameter, we can remove the dependency.

ianswett: I'm happy to have congestion control depend upon loss detection, but it likely would be cleaner to pass in the time directly.

ianswett: Actually, we're using the ce counter as well.  I think I'll leave it as is unless you think there's a correctness issue?

maskit: > I'm happy to have congestion control depend upon loss detection, but it likely would be cleaner to pass in the time directly.

OK, I just thought other places might avoid it intentionally.

> Actually, we're using the ce counter as well.

If you mean `ack.ce_counter` that one is fine because it is in `ack`. I meant `ProcessECN(ack, largest_acked_sent_time)`

> I think I'll leave it as is unless you think there's a correctness issue?

I'd implement it differently but I'm ok with as-is.

PR #1931, opened 2018-10-30
The spin-bit should be some times disabled

huitema:
For privacy reasons, and in a probabilistic way to grease the spin-bit, and to provide a sufficient anonymity set to the privacy conscious folks.

### DIFF
--- a/draft-ietf-quic-spin-exp.md
+++ b/draft-ietf-quic-spin-exp.md
@@ -271,7 +271,7 @@ RTT, and will thus enable adversaries near the client to discover that
 the connection does not terminate at the visible destination address, and
 that the client is probably accessing a hidden server through a proxy.
 
-Clients that want to hide their use of a proxy or a relay will want to
+Endpoints that want to hide their use of a proxy or a relay will want to
 disable the spin bit. However, if only privacy sensitive clients ever
 disabled the spin bit, they would stick out. The probabilistic disabling
 behavior explained in {#disabling-the-spin-bit} ensures that other clients
--- a/draft-ietf-quic-spin-exp.md
+++ b/draft-ietf-quic-spin-exp.md
@@ -242,7 +242,7 @@ in the downstream direction, and vice versa.
 
 Implementations SHOULD allow users to disable the spin bit either globally,
 or on a per-connection basis. Even when the spin bit is not disabled by
-the user, implementations SHOULD disable the spin-bit on a randomly chosen
+the user, implementations SHOULD disable the spin bit on a randomly chosen
 fraction of connections. The random process SHOULD be designed so that
 on average the spin-bit is disabled for at least 1/8th of the connections.
 
--- a/draft-ietf-quic-spin-exp.md
+++ b/draft-ietf-quic-spin-exp.md
@@ -240,7 +240,7 @@ in the downstream direction, and vice versa.
 
 # Disabling the Spin Bit
 
-Implementation SHOULD allow users to disable the spin-bit either globally,
+Implementations SHOULD allow users to disable the spin bit either globally,
 or on a per-connection basis. Even when the spin-bit is not disabled by
 the user, implementations SHOULD disable the spin-bit on a randomly chosen
 fraction of connections. The random process SHOULD be designed so that
--- a/draft-ietf-quic-spin-exp.md
+++ b/draft-ietf-quic-spin-exp.md
@@ -267,7 +267,7 @@ on-path devices is in most cases negligible.
 There is however an exception, when parts of the path from client to server
 are hidden from observers. An example would be a server accessed through a
 proxy. The spin bit allows for measurement of the end-to-end
-RTT, and will thus enable adversaries near the client to discover that
+RTT, and will thus enable adversaries near the endpoint to discover that
 the connection does not terminate at the visible destination address, and
 that the client is probably accessing a hidden server through a proxy.
 
--- a/draft-ietf-quic-spin-exp.md
+++ b/draft-ietf-quic-spin-exp.md
@@ -238,7 +238,7 @@ and the observer and the client, respectively. It does this by measuring the
 delay between a spin edge observed in the upstream direction and that observed
 in the downstream direction, and vice versa.
 
-# Disabling the spin bit
+# Disabling the Spin Bit
 
 Implementation SHOULD allow users to disable the spin-bit either globally,
 or on a per-connection basis. Even when the spin-bit is not disabled by
--- a/draft-ietf-quic-spin-exp.md
+++ b/draft-ietf-quic-spin-exp.md
@@ -241,7 +241,7 @@ in the downstream direction, and vice versa.
 # Disabling the Spin Bit
 
 Implementations SHOULD allow users to disable the spin bit either globally,
-or on a per-connection basis. Even when the spin-bit is not disabled by
+or on a per-connection basis. Even when the spin bit is not disabled by
 the user, implementations SHOULD disable the spin-bit on a randomly chosen
 fraction of connections. The random process SHOULD be designed so that
 on average the spin-bit is disabled for at least 1/8th of the connections.
--- a/draft-ietf-quic-spin-exp.md
+++ b/draft-ietf-quic-spin-exp.md
@@ -241,7 +241,7 @@ in the downstream direction, and vice versa.
 # Disabling the Spin Bit
 
 Implementations SHOULD allow administrators of clients and servers to disable
-the spin bit either globally, or on a per-connection basis.
+the spin bit either globally or on a per-connection basis.
 Even when the spin bit is not disabled by the administrator implementations
 SHOULD disable the spin bit on a randomly chosen
 fraction of connections.  The selection process should be designed such that
@@ -274,7 +274,7 @@ the connection does not terminate at the visible destination address.
 Endpoints that want to hide their use of a proxy or a relay will want to
 disable the spin bit. However, if only privacy-sensitive clients or servers ever
 disabled the spin bit, they would stick out. The probabilistic disabling
-behavior explained in {#disabling-the-spin-bit} ensures that other endpoints
+behavior explained in {{disabling-the-spin-bit}} ensures that other endpoints
 will also disable the spin bit some of the time, thus hiding the
 privacy sensitive endpoints in a large anonymity set. It also provides
 for a minimal greasing of the spin bit, in order to mitigate risks of
@@ -288,7 +288,7 @@ ossification.
 
 ## Since draft-ietf-spin-exp-00
 
-Nothing yet.
+Adding section on disabling the spin bit and privacy considerations.
 
 # Acknowledgments
 {:numbered="false"}
--- a/draft-ietf-quic-spin-exp.md
+++ b/draft-ietf-quic-spin-exp.md
@@ -272,7 +272,7 @@ RTT, and will thus enable adversaries near the endpoint to discover that
 the connection does not terminate at the visible destination address.
 
 Endpoints that want to hide their use of a proxy or a relay will want to
-disable the spin bit. However, if only privacy sensitive clients or servers ever
+disable the spin bit. However, if only privacy-sensitive clients or servers ever
 disabled the spin bit, they would stick out. The probabilistic disabling
 behavior explained in {#disabling-the-spin-bit} ensures that other endpoints
 will also disable the spin bit some of the time, thus hiding the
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-spin-exp.md
```diff
@@ -238,6 +238,17 @@ and the observer and the client, respectively. It does this by measuring the
 delay between a spin edge observed in the upstream direction and that observed
 in the downstream direction, and vice versa.
 
+# Disabling the spin bit
```
martinthomson:
```suggestion
# Disabling the Spin Bit
```

draft-ietf-quic-spin-exp.md
```diff
@@ -238,6 +238,17 @@ and the observer and the client, respectively. It does this by measuring the
...
+
+Implementation SHOULD allow users to disable the spin-bit either globally,
```
martinthomson:
```suggestion
Implementations SHOULD allow users to disable the spin bit either globally,
```

draft-ietf-quic-spin-exp.md
```diff
@@ -238,6 +238,17 @@ and the observer and the client, respectively. It does this by measuring the
...
+or on a per-connection basis. Even when the spin-bit is not disabled by
```
martinthomson:
```suggestion
or on a per-connection basis. Even when the spin bit is not disabled by
```

draft-ietf-quic-spin-exp.md
```diff
@@ -238,6 +238,17 @@ and the observer and the client, respectively. It does this by measuring the
...
+the user, implementations SHOULD disable the spin-bit on a randomly chosen
```
martinthomson:
```suggestion
the user, implementations SHOULD disable the spin bit on a randomly chosen
```

draft-ietf-quic-spin-exp.md
```diff
@@ -238,6 +238,17 @@ and the observer and the client, respectively. It does this by measuring the
...
+fraction of connections. The random process SHOULD be designed so that
+on average the spin-bit is disabled for at least 1/8th of the connections.
```
martinthomson:
```suggestion
on average the spin bit is disabled for at least 1/8th of the connections.
```

draft-ietf-quic-spin-exp.md
```diff
@@ -238,6 +238,17 @@ and the observer and the client, respectively. It does this by measuring the
...
+
+When the spin-bit is disabled, peers MUST set the spin bit value to zero,
```
martinthomson:
```suggestion
When the spin bit is disabled, endpoints MUST set the spin bit value to zero,
```

Any sense in permitting different forms of disabling?  I mean, there are other ways to erode the signal.  Fixed to zero, fixed to one, random, alternating values on every packet.  I wouldn't prohibit any (or all) of these techniques from being used.

draft-ietf-quic-spin-exp.md
```diff
@@ -251,7 +262,23 @@ same as those for passive RTT measurement in general. It has been shown
...
+are hidden from observers. An example would be a server accessed through a
+proxy. The spin bit allows for measurement of the end-to-end
+RTT, and will thus enable adversaries near the client to discover that
+the connection does not terminate at the visible destination address, and
+that the client is probably accessing a hidden server through a proxy.
+
+Clients that want to hide their use of a proxy or a relay will want to
+disable the spin bit. However, if only privacy sensitive clients ever
+disabled the spin bit, they would stick out. The probabilistic disabling
+behavior explained in {#disabling-the-spin-bit} ensures that other clients
```
martinthomson:
```suggestion
behavior explained in {{disabling-the-spin-bit}} ensures that other endpoints
```

draft-ietf-quic-spin-exp.md
```diff
@@ -251,7 +262,23 @@ same as those for passive RTT measurement in general. It has been shown
...
+will also disabling the spin bit some of the time, thus hiding the
```
martinthomson:
```suggestion
will also disable the spin bit some of the time, thus hiding the
```

draft-ietf-quic-spin-exp.md
```diff
@@ -251,7 +262,23 @@ same as those for passive RTT measurement in general. It has been shown
...
+on-path devices is in most cases negligible.
+
+There is however an exception, when parts of the path from client to server
+are hidden from observers. An example would be a server accessed through a
+proxy. The spin bit allows for measurement of the end-to-end
+RTT, and will thus enable adversaries near the client to discover that
+the connection does not terminate at the visible destination address, and
+that the client is probably accessing a hidden server through a proxy.
+
+Clients that want to hide their use of a proxy or a relay will want to
```
martinthomson:
```suggestion
Endpoints that want to hide their use of a proxy or a relay will want to
```

I don't think that we can assume that it is only clients that will be affected by this.

draft-ietf-quic-spin-exp.md
```diff
@@ -251,7 +262,23 @@ same as those for passive RTT measurement in general. It has been shown
 {{PAM-RTT}} that RTT measurements do not provide more information for
 geolocation than is available in the most basic, freely-available IP address
 based location databases. The risk of exposure of per-flow network RTT to
-on-path devices is therefore negligible.
+on-path devices is in most cases negligible.
+
+There is however an exception, when parts of the path from client to server
+are hidden from observers. An example would be a server accessed through a
+proxy. The spin bit allows for measurement of the end-to-end
+RTT, and will thus enable adversaries near the client to discover that
```
martinthomson:
```suggestion
RTT, and will thus enable adversaries near the endpoint to discover that
```

draft-ietf-quic-spin-exp.md
```diff
@@ -251,7 +262,23 @@ same as those for passive RTT measurement in general. It has been shown
...
+the connection does not terminate at the visible destination address, and
```
martinthomson:
```suggestion
the connection does not terminate at the visible destination address.
```

draft-ietf-quic-spin-exp.md
```diff
@@ -251,7 +262,23 @@ same as those for passive RTT measurement in general. It has been shown
...
+that the client is probably accessing a hidden server through a proxy.
```
martinthomson:
```suggestion
```

draft-ietf-quic-spin-exp.md
martinthomson: I would rather say that the endpoint can set the value to any value then.  Providing a concrete suggestion means that anyone who diverges from that suggestion will stick out.  Not providing a concrete suggestion means that you might end up with the choice of method being used to identify an implementation, but that's probably a lost cause.

draft-ietf-quic-spin-exp.md
```diff
@@ -238,6 +238,17 @@ and the observer and the client, respectively. It does this by measuring the
 delay between a spin edge observed in the upstream direction and that observed
 in the downstream direction, and vice versa.
 
+# Disabling the Spin Bit
+
+Implementations SHOULD allow users to disable the spin bit either globally,
```
janaiyengar:
```suggestion
Implementations SHOULD allow users to disable the spin bit either globally
```

draft-ietf-quic-spin-exp.md
```diff
@@ -238,6 +238,17 @@ and the observer and the client, respectively. It does this by measuring the
...
+or on a per-connection basis. Even when the spin bit is not disabled by
+the user, implementations SHOULD disable the spin bit on a randomly chosen
+fraction of connections. The random process SHOULD be designed so that
```
janaiyengar:
```suggestion
fraction of connections.  The selection process should be designed such that
```

draft-ietf-quic-spin-exp.md
marten-seemann: For greasing it might be (slightly) better to just use random values for this bit, otherwise middleboxes might learn that this bit is either 0 all the time or contains a valid spin bit.

draft-ietf-quic-spin-exp.md
```diff
@@ -251,7 +262,23 @@ same as those for passive RTT measurement in general. It has been shown
...
+
+There is however an exception, when parts of the path from client to server
+are hidden from observers. An example would be a server accessed through a
+proxy. The spin bit allows for measurement of the end-to-end
+RTT, and will thus enable adversaries near the endpoint to discover that
+the connection does not terminate at the visible destination address, and
+that the client is probably accessing a hidden server through a proxy.
+
+Endpoints that want to hide their use of a proxy or a relay will want to
+disable the spin bit. However, if only privacy sensitive clients ever
```
janaiyengar:
```suggestion
disable the spin bit. However, if only privacy-sensitive clients ever
```

draft-ietf-quic-spin-exp.md
```diff
@@ -251,7 +262,23 @@ same as those for passive RTT measurement in general. It has been shown
...
+disabled the spin bit, they would stick out. The probabilistic disabling
+behavior explained in {#disabling-the-spin-bit} ensures that other clients
+will also disabling the spin bit some of the time, thus hiding the
+privacy sensitive clients in a large anonymity set. It also provides
```
janaiyengar:
```suggestion
privacy-sensitive clients in a large anonymity set. It also provides
```

draft-ietf-quic-spin-exp.md
```diff
@@ -251,7 +263,22 @@ same as those for passive RTT measurement in general. It has been shown
...
+on-path devices is in most cases negligible.
+
+There is however an exception, when parts of the path from client to server
+are hidden from observers. An example would be a server accessed through a
+proxy. The spin bit allows for measurement of the end-to-end
+RTT, and will thus enable adversaries near the endpoint to discover that
+the connection does not terminate at the visible destination address.
+
+Endpoints that want to hide their use of a proxy or a relay will want to
+disable the spin bit. However, if only privacy sensitive clients or servers ever
```
janaiyengar:
```suggestion
disable the spin bit. However, if only privacy-sensitive clients or servers ever
```

draft-ietf-quic-spin-exp.md
```diff
@@ -251,7 +263,22 @@ same as those for passive RTT measurement in general. It has been shown
...
+disabled the spin bit, they would stick out. The probabilistic disabling
+behavior explained in {#disabling-the-spin-bit} ensures that other endpoints
+will also disable the spin bit some of the time, thus hiding the
+privacy sensitive endpoints in a large anonymity set. It also provides
```
janaiyengar:
```suggestion
privacy-sensitive endpoints in a large anonymity set. It also provides
```

marten-seemann:
It seems like this PR is absolutely essential to fulfill our charter (assuming that we standardize the spin bit, that is), which requires us to design a protocol that has privacy properties as least as good as TLS1.3 over TCP.

I'm still wondering if this is good enough though. If 1/8 of connections opt out of the spin-bit, this will provide a large anonymity set, if peers using VPN are rare. 
Consider however a situation where 50% of the peers are using a VPN, and don't want to expose their RTT. 80% of all connections that disable the spin bit are then VPN clients, which would certainly make them stick out.

draft-ietf-quic-spin-exp.md
huitema: In order of complexity, I can see:

1) Always zero. Simple, people won't mess up.
2) Pick either 0 or 1 randomly at the beginning of the connection. Simple enough, somewhat better than always zero for preventing ossification.
3) Pick a random value each time. I don't like that too much. Implementation will mess up and generate a fingerprint.

I could specify flip a coin at the beginning of each connection if that's what you want.

marten-seemann: In New York we discussed the proposal to grease whatever bits are left in the first octet by just encrypting 0s. I assumed that peers opting out of the spin bit would use the same algorithm they use for greasing the other bits.

huitema: @marten-seemann Yes, encryption is better than just randomization, especially if we specify it completely so there are no differences between implementations.

mikkelfj:
It is easy to detect the same user after multiple visits, and also easy to detect an IP range as serving as a VPN, so it is this complexity really worthwhile?

draft-ietf-quic-spin-exp.md
martinthomson: You can't encrypt using the same technique because the value you choose will be input to the packet protection AEAD and so might alter the header encryption sampling.  If the intent is to appear random, a CSPRNG would be fine.  If the value can be predicted, then you risk undoing packet number encryption, promoting linkability, or something worse.

PR #1936, opened 2018-10-30
Spin bit observers can't see the packet number anymore

britram:
@tfpauly pointed out that there was stale text in -01 that referred to an observer using the packet number to reject lost and reordered edges. Now that PNE is in, this is nonsensical. This PR removes it and points to heuristic methods instead.

### DIFF
--- a/draft-ietf-quic-spin-exp.md
+++ b/draft-ietf-quic-spin-exp.md
@@ -199,16 +199,6 @@ can observe the time difference between edges (changes from 1 to 0 or 0 to 1)
 in the spin bit signal in a single direction to measure one sample of
 end-to-end RTT.
 
-An observer can store the largest observed packet number per flow, and reject
-edges that do not have a monotonically increasing packet number (greater than
-the largest observed packet number).  This will avoid detecting spurious edges
-caused by reordering events that include an edge, which would lead to very low
-RTT estimates if not ignored.
-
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-spin-exp.md
```diff
@@ -224,11 +214,12 @@ small amount of periodic application traffic, where that period is longer than
...
-RTT series can be used to reject bad RTT samples due to application or flow
-control limitation; for example, QoF {{TMA-QOF}} rejects component RTTs
-significantly higher than RTTs over the history of the flow. These heuristics
-may use the handshake RTT as an initial RTT estimate for a given flow.
+Simple heuristics based on the observed data rate per flow or changes in the RTT
+series can be used to reject bad RTT samples due to lost or reordered edges in
+the spin signal, as well as application or flow control limitation; for example,
+QoF {{TMA-QOF}} rejects component RTTs significantly higher than RTTs over the
+history of the flow. These heuristics may use the handshake RTT as an initial
+RTT estimate for a given flow.
```
mirjak: I would be careful here because you actually want to detect suddent changes. The question is rather is it only one outlier or did actually something happen.

britram: yep, but this is a common problem with any passive RTT scheme (or indeed, RTT estimation at an endpoint): the base data is noisy, so it takes a few samples to adjust to a new real value.

PR #1938, opened 2018-10-30
Errors for bad stream-related frames

### LINKED ISSUE #1933: Uneven stream identifier example
draft -16
> If an endpoint receives a frame for a stream that it expects to
   initiate (i.e., odd-numbered for the client or even-numbered for the
   server), but which it has not yet opened, it MUST close the
   connection with error code STREAM_STATE_ERROR.

I believe that should have been an even id for client. Also, the example could be move until after discussing the bidi/uni types and add "or if a stream frame is received for a unidirectional stream that is initiated locally" (I'm sure this text exists, but it feels natural to have those together).

martinthomson:
This fixes the even/odd problem from #1933, moves the paragraph as
suggested there, and adds a note about unidirectional streams.

Closes #1933.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -289,10 +289,6 @@ set to 1).  Separation of the stream identifiers ensures that client and server
 are able to open streams without the latency imposed by negotiating for an
 identifier.
 
-If an endpoint receives a frame for a stream that it expects to initiate (i.e.,
-odd-numbered for the client or even-numbered for the server), but which it has
-not yet opened, it MUST close the connection with error code STREAM_STATE_ERROR.
-
 The second least significant bit (0x2) of the Stream ID differentiates between
 unidirectional streams and bidirectional streams. Unidirectional streams always
 have this bit set to 1 and bidirectional streams have this bit set to 0.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -310,6 +306,13 @@ The two type bits from a Stream ID therefore identify streams as summarized in
 
 The first bidirectional stream opened by the client is stream 0.
 
+If an endpoint receives a frame for a stream that it expects to initiate (i.e.,
+an even-numbered for the client or an odd-numbered for the server), but which it
+has not yet opened, it MUST close the connection with error code
+STREAM_STATE_ERROR.  Similarly, receiving a STREAM or RST_STREAM frame for a
+unidirectional streams that an endpoint expects to initiate MUST be treated as a
```
```suggestion
unidirectional stream that an endpoint expects to initiate MUST be treated as a
```

```suggestion
unidirectional stream that an endpoint expects to initiate MUST be treated as a
```

(plural stream)

draft-ietf-quic-transport.md
```diff
@@ -310,6 +306,13 @@ The two type bits from a Stream ID therefore identify streams as summarized in
 
 The first bidirectional stream opened by the client is stream 0.
 
+If an endpoint receives a frame for a stream that it expects to initiate (i.e.,
+an even-numbered for the client or an odd-numbered for the server), but which it
+has not yet opened, it MUST close the connection with error code
+STREAM_STATE_ERROR.  Similarly, receiving a STREAM or RST_STREAM frame for a
```
marten-seemann: This also applies to STREAM_DATA_BLOCKED.

martinthomson: Yeah, I thought of that.  Now I'm regretting having this paragraph.  Maybe it's best to just remove it and rely on other text.

mikkelfj: Just add a note: Other misguided stream frames such as MAX... also closes the cinnection.

Issue #1724, opened 2018-09-05
Recovery "what to send section"?  / Pseudo code issue 


larseggert:
i think there is an issue in the pseudo code (but I may be wrong - I find the code much more difficult to follow than an actual textual explanation.)

In `OnAckReceived`, there is the following call sequence:
```
       DetectLostPackets(ack.largest_acked_packet)
       SetLossDetectionTimer()
```

`DetectLostPackets` calls `OnPacketsLost`, which reduces `in_flight` potentially to zero. Then in `SetLossDetectionTimer`, the timer is cancelled via `loss_detection_timer.cancel()`.

This causes no retransmissions to happen, at least when (blindly) implementing the pseudo code.

Related to that, it is not clear where exactly the pseudo code expects actual retransmissions to happen. There is code in `OnLossDetectionTimeout` (`RetransmitAllUnackedHandshakeData`, `SendOnePacket`, `SendTwoPackets`) but what those do is not explained. There is also a comment in `DetectLostPackets` ("Inform the congestion controller...") but it remains unclear when and how that would lead to retransmissions.

IMO, if we rely on pseudo code for this document, it needs a lot more work to make sure it actually does what is intended. (My preference would be to instead describe the concepts in text.)

mirjak:
I believe this is general issue with TLP which shows up as you are sending only with a low constant rate. We observed a similar problem in TCP...

mirjak:
However, +1 to more text explanation!

larseggert:
The pseudo code shouldn't have "general issues" :-)

ianswett:
There needs to be some more clarity here, and I believe the core issue is this document is trying to describe both how to quickly detect packets as lost and congestion control, but is avoiding what send a bit too much.

Given QUIC doesn't retransmit packets, and instead retransmits stream data/etc, just because a packet is lost doesn't mean it's payload needs to be retransmitted, and if it does, it may be bundled differently from the original transmission.

Would a section on "What to send" be helpful?  Or would it be better to work that into the existing text and/or pseudocode?

martinthomson:
Was this addressed in #1939?  The comment there says "part of" this only.  @ianswett, what is missing?

ianswett:
Between that PR an some subsequent cleanups, I think we're in good shape.

PR #1939, opened 2018-10-31
What packet information to track and send

### LINKED ISSUE #765: Complicated Retransmission Corner Cases
The basic case for retransmission is pretty straightforward. I detect a packet loss, so I take all the stuff in that packet and send it in a new packet. There has to be some sort of data structure in the PCB to track what's in the packet. I can clear that data structure once the packet is acked.

There are two ambiguities here:
(1) How long do I hold on to the packet data if lost? I can delete it immediately, hold it for _n_ retransmissions, or hold it until all acks and data in the packet are acked.

Deleting it immediately is very simple, but if the ack comes back for that packet later and the first retransmission is lost, I will needlessly resend that data.

Offline, Jana suggested keeping it for 1 retransmission. That is fine with me (minus the second item below) but I think some implementation guidance on when it is acceptable to throw away old packet data would be helpful.

(2) More importantly, fitting the whole retransmission into a single packet does not hold in the general case. There might be an MTU change; the updated ack might be bigger; optimizers like me might want to aggregate multiple small packets into a single retransmission; some ack frames might be supersets of other ack frames, so although there is no direct retransmission relationship acking one of those packets might obviate the ack frame of the other.

What all this implies is a many-to-many mapping where every time data or acks are acked, we have to clear that from all other packets that contain it. I started down this road before seeing that it was way too hairy to continue.

****

So we're at a point where the optimal wire performance -- never throw it away until all components are acked, allow mix-and-match of retransmissions, don't resend stuff if acked in some other packet -- is close to unimplementable. Some implementation guidance about what it's OK to not worry about, and perhaps some judicious prohibitions on particularly gnarly stuff, might help some.

ianswett:
Fixes #765 and some of #1724

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -529,7 +529,7 @@ packet is lost, it SHOULD be kept for an amount of time comparable
 to the maximum expected packet reordering, such as 1 RTT.
 
 Sent packets are tracked for each packet number space, and ACK
-processing only applies to a single space.  
+processing only applies to a single space.
 
 ### Fields
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -519,9 +519,8 @@ being transmitted are tracked in case of loss.
 
 If a packet containing retransmittable frames is lost, the QUIC transport
 needs to recover from that loss, such as by retransmitting the data,
-sending an updated frame, or abandoning the frame. In the case of
-STREAM data, the data should be retransmitted unless the send side of the
-stream is closed.
+sending an updated frame, or abandoning the frame.  For more information,
+see "Retransmission of Information" in {{QUIC-TRANSPORT}}.
 
 Packets MUST be tracked until acknowledged or lost.  After a packet is lost,
 it SHOULD be tracked for an amount of time comparable to the maximum
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -507,6 +507,59 @@ frames containing them could be lost. In this case, the loss recovery algorithm
 may cause spurious retransmits, but the sender will continue making forward
 progress.
 
+## Tracking Sent Packets {#tracking-sent-packets}
+
+QUIC stores information about every packet sent.  QUIC does not retransmit
+packets, but implementations SHOULD retransmit the data within a packet in a
+new QUIC packet with a new packet number when the  packet is lost.  When
+retransmitting that data, it may be bundled with new data if there is space
+or split into multiple QUIC packets as necessary.
+
+It's expected implementations will index this per-packet information by
+packet number and store the per-packet information detailed below for
+loss recovery and congestion control.  Additionally, implementations
+must ensure that any reliable frames being transmitted are tracked
+in case of loss.  If a packet containing reliable frames is lost,
+the QUIC transport needs to decide how to recovery from that loss.
+In the case of data contained in STREAM frames, the data should be
+retransmitted unless the send side of the stream is now closed.
+
+Packets remain in sent_packets until acknowledged or lost.  After a
+packet is lost, it SHOULD be kept for an amount of time comparable
+to the maximum expected packet reordering, such as 1 RTT.
+
+Sent packets are tracked for each packet number space, and ACK
+processing only applies to a single space.  
+
+### Fields
+
+packet_number
+: The packet number of the sent packet.
+
+retransmittable:
+: A boolean that indicates whether a packet is retransmittable.
+  If true, it is expected that an acknowledgement will be received,
+  though the peer could delay sending the ACK frame containing it
+  by up to the MaxAckDelay.
+
+in_flight:
+: A boolean that indicates whether the packet counts towards bytes in
+  flight.
+
+is_crypto_packet:
+: A boolean that indicates whether the packet contains
+  cryptographic handshake messages critical to the completion of the QUIC
+  handshake. In this version of QUIC, this includes any packet with the long
+  header that includes a CRYPTO frame.
+
+sent_bytes:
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -653,27 +698,8 @@ follows:
...
-
-* is_crypto_packet: A boolean that indicates whether the packet contains
-  cryptographic handshake messages critical to the completion of the QUIC
-  handshake. In this version of QUIC, this includes any packet with the long
-  header that includes a CRYPTO frame.
-
-* sent_bytes: The number of bytes sent in the packet, not including UDP or IP
-  overhead, but including QUIC framing overhead.
+After any packet is sent.  The parameters to OnPacketSent are described in detail
+above in {{sent-packets-fields}}.
```
nibanks: I believe the error is because `detail` needs to wrap to the next line.
```
After any packet is sent.  The parameters to OnPacketSent are described in
detail above in {{sent-packets-fields}}.
```

ianswett: Thanks Nick, fixed.

draft-ietf-quic-recovery.md
```diff
@@ -509,6 +509,58 @@ frames containing them could be lost. In this case, the loss recovery algorithm
...
+will index this per-packet information by packet number and store the
+per-packet fields detailed below for loss recovery and congestion control.
+Additionally, implementations MUST ensure that any retransmittable frames
+being transmitted are tracked in case of loss.
+
+If a packet containing retransmittable frames is lost, the QUIC transport
+needs to recover from that loss, such as by retransmitting the data,
+sending an updated frame, or abandoning the frame. In the case of
+STREAM data, the data should be retransmitted unless the send side of the
+stream is closed.
```
martinthomson: I don't know why you would call out STREAM specially here.  Why not just reference [Section 13.2 of transport](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#rfc.section.13.2).

ianswett: Great idea, is this the right way to reference a section of another doc?

draft-ietf-quic-recovery.md
```diff
@@ -509,6 +509,57 @@ frames containing them could be lost. In this case, the loss recovery algorithm
...
+QUIC stores information about every packet sent. It's expected implementations
+will index this per-packet information by packet number and store the
+per-packet fields detailed below for loss recovery and congestion control.
+Additionally, implementations MUST ensure that any retransmittable frames
+being transmitted are tracked in case of loss.
+
+If a packet containing retransmittable frames is lost, the QUIC transport
+needs to recover from that loss, such as by retransmitting the data,
+sending an updated frame, or abandoning the frame.  For more information,
+see "Retransmission of Information" in {{QUIC-TRANSPORT}}.
```
```suggestion
see Section 13.2 of {{QUIC-TRANSPORT}}.
```

draft-ietf-quic-recovery.md
```diff
@@ -509,6 +509,57 @@ frames containing them could be lost. In this case, the loss recovery algorithm
...
+packet_number
+: The packet number of the sent packet.
+
+retransmittable:
+: A boolean that indicates whether a packet is retransmittable.
+  If true, it is expected that an acknowledgement will be received,
+  though the peer could delay sending the ACK frame containing it
+  by up to the MaxAckDelay.
+
+in_flight:
```
martinduke: Not a big deal, but I found it easier to just store "flightsize" which is set to zero if it doesn't count, instead of having in_flight and send_bytes. If you find that more compact and understandable, feel free to change it.

draft-ietf-quic-recovery.md
```diff
@@ -509,6 +509,57 @@ frames containing them could be lost. In this case, the loss recovery algorithm
...
+
+If a packet containing retransmittable frames is lost, the QUIC transport
+needs to recover from that loss, such as by retransmitting the data,
+sending an updated frame, or abandoning the frame.  For more information,
+see Section 13.2 of {{QUIC-TRANSPORT}}.
+
+Packets MUST be tracked until acknowledged or lost.  After a packet is lost,
+it SHOULD be tracked for an amount of time comparable to the maximum
+expected packet reordering, such as 1 RTT.  This allows detection of
+spurious retransmissions and MAY allow avoiding an extra retransmissions if
```
martinduke: s/allow avoiding an/avoid

draft-ietf-quic-recovery.md
```diff
@@ -509,6 +509,57 @@ frames containing them could be lost. In this case, the loss recovery algorithm
...
 progress.
 
+## Tracking Sent Packets {#tracking-sent-packets}
+
+QUIC stores information about every packet sent. It's expected implementations
+will index this per-packet information by packet number and store the
+per-packet fields detailed below for loss recovery and congestion control.
+Additionally, implementations MUST ensure that any retransmittable frames
+being transmitted are tracked in case of loss.
+
```
martinduke: "Implementations MUST track the ack blocks in a packet in order to stop sending those ack blocks if the packet is 
acknowledged."

Perhaps this is not strictly loss-recovery-related, but this is the section where I'm going to go to implement struct sent_pkt.

draft-ietf-quic-recovery.md
```diff
@@ -509,6 +509,57 @@ frames containing them could be lost. In this case, the loss recovery algorithm
 may cause spurious retransmits, but the sender will continue making forward
 progress.
 
+## Tracking Sent Packets {#tracking-sent-packets}
+
+QUIC stores information about every packet sent. It's expected implementations
+will index this per-packet information by packet number and store the
```
martinduke: "by packet number and crypto context"?

martinthomson:
@ianswett for some reason this got merged with some other changes.  I'm going to open you a new PR.

martinthomson:
See #1956.

PR #1942, opened 2018-10-31
Change retransmittable to Ack-eliciting

### LINKED ISSUE #1595: Rename "retransmittable" frames and packets
The word retransmittable is used in the recovery draft for both frames and packets that cause an ACK frame to be sent in response to their receipt.

This is really misleading, because QUIC never retransmits packets(except maybe connection close) and only sometimes retransmits frames.

Ideally we'd have different words for the frames and the packets containing those frames, but at a minimum I want to stop using retransmittable.

Previous suggestions:
- Ackable - I didn't like this before, but it's strictly better than retransmittable.
- Reliable - Maybe a good term for packets, less good for frames?
- Needs Acknowledgement - Verbose, but correct.  Possibly abbreviated to Nack for maximum confusion :)
- Persistent

I'm want to change retransmittable to something else before the next drafts, so please suggest alternatives now and I'll write a PR for whatever people hate the least.

ianswett:
Ack-eliciting is slightly shorter than retransmittable and seems as accurate as anything we have.

Fixes #1595

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -98,16 +98,16 @@ In-flight:
   and neither acknowledged nor declared lost, and they are not
   ACK-only.
 
-Retransmittable Frames:
+Ack-eliciting Frames:
 
 : All frames besides ACK or PADDING are considered
-  retransmittable.
+  ack-eliciting.
 
-Retransmittable Packets:
+Ack-eliciting Packets:
 
-: Packets that contain retransmittable frames elicit an ACK from
+: Packets that contain ack-eliciting frames elicit an ACK from
   the receiver within the maximum ack delay and are called
-  retransmittable packets.
+  ack-eliciting packets.
 
 Crypto Packets:
 
@@ -133,7 +133,7 @@ acknowledged or declared lost and sent in new packets as necessary. The types
 of frames contained in a packet affect recovery and congestion control logic:
 
 * All packets are acknowledged, though packets that contain no
-  retransmittable frames are only acknowledged along with retransmittable
+  ack-eliciting frames are only acknowledged along with ack-eliciting
   packets.
 
 * Long header packets that contain CRYPTO frames are critical to the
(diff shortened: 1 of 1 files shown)


ianswett:
I'm not set on this being what we go with, but I think it reads reasonably well and I'm really unhappy with retransmittable, since I think it's just confusing.  If there's another alternative you'd like to see me create a PR for, feel free to suggest it.

dtikhonov:
Does the English language have a suffix that means that something _must_ be done ("ack`$foo`") rather than that _can_ be done ("ackable")?  What's `$foo`? 😖

dtikhonov:
Based on [this list of English suffixes](https://www.learnthat.org/pages/view/suffix.html), I propose that we use _ackent_ instead of ack-eliciting.  The suffix _ENT_ means "one who performs/causes."

Thus, instead of

> Packets that contain **ack-eliciting** frames elicit an ACK from the receiver within the maximum ack delay and are called **ack-eliciting** packets.

we will have

> Packets that contain **ackent** frames elicit an ACK from the receiver within the maximum ack delay and are called **ackent** packets.

mikkelfj:
resilient

> Packets that contain resilient frames elicit an ACK from the receiver within the maximum ack delay and are called resilient packets.

unforgettable
https://www.youtube.com/watch?v=2HSt7LWsm-A

> Packets that contain unforgettable frames elicit an ACK from the receiver within the maximum ack delay and are called unforgettable packets.

unmissable

> Packets that contain unmissable frames elicit an ACK from the receiver within the maximum ack delay and are called unmissable packets.

robust

> Packets that contain robust frames elicit an ACK from the receiver within the maximum ack delay and are called robust packets.

stateful

> Packets that contain stateful frames elicit an ACK from the receiver within the maximum ack delay and are called stateful packets.

dtikhonov:
@mikkelfj, my suggestion was in earnest.

mikkelfj:
Mine was only half in jest. I kind of like resilient.

PR #1957, opened 2018-11-03
Editorial cleanup of Recovery

ianswett:
Some small changes I noticed when re-reading the entire doc.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -171,7 +171,7 @@ and any application data is sent in one or more streams, with delivery order
 determined by stream offsets encoded within STREAM frames.
 
 QUIC's packet number is strictly increasing within a packet number space,
-and directly encodes transmission order.  A higher QUIC packet number
+and directly encodes transmission order.  A higher packet number
 signifies that the packet was sent later, and a lower QUIC packet number
 signifies that the packet was sent earlier.  When a packet containing
 retransmittable frames is deemed lost, QUIC rebundles necessary frames in a
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -238,8 +238,8 @@ Ack-based loss detection implements the spirit of TCP's Fast Retransmit
 {{?RFC5681}}, Early Retransmit {{?RFC5827}}, FACK, and SACK loss recovery
 {{?RFC6675}}. This section provides an overview of how these algorithms are
 implemented in QUIC.  Though both time-based loss detection and early retransmit
-use a timer, they are part of Ack-based detection because they do not use a timer
-to send probes, but rather to declare packets lost.
+use a timer, they are part of Ack-based detection because they do not use a
+timer to send probes, but rather to declare packets lost.
 
 ### Fast Retransmit
 
@@ -792,7 +792,8 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
 
 When a packet is acknowledged for the first time, the following OnPacketAcked
 function is called.  Note that a single ACK frame may newly acknowledge several
-packets. OnPacketAcked must be called once for each of these newly acked packets.
+packets. OnPacketAcked must be called once for each of these newly acked
+packets.
 
 OnPacketAcked takes one parameter, acked_packet, which is the struct detailed in
 {{sent-packets-fields}}.
@@ -889,9 +890,11 @@ Pseudocode for OnLossDetectionTimeout follows:
 
 DetectLostPackets is called every time an ACK is received and operates on
 the sent_packets for that packet number space. If the loss detection timer
-expires and the loss_time is set, the previous largest acknowledged packet is supplied.
+expires and the loss_time is set, the previous largest acknowledged packet
+is supplied.
 
-DetectLostPackets takes one parameter, acked, which is the largest acked packet.
+DetectLostPackets takes one parameter, largest_acked, which is the largest
+acked packet.
 
 Pseudocode for DetectLostPackets follows:
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -264,9 +264,15 @@ TCP-NCR {{?RFC4653}}, to improve QUIC's reordering resilience.
 
 QUIC implementations can use time-based loss detection to handle reordering
 based on time elapsed since the packet was sent.  This may be used either as
-a replacement for a packet reordering threshold or in addition to it.
-The RECOMMENDED time threshold, expressed as a fraction of the
-round-trip time (kTimeReorderingFraction), is 1/8.
+a replacement for a packet reordering threshold or in addition to it.  When
+a larger packet is acknowledged, if it was sent more than the threshold after
+any in flight packets, those packets are immediately declared lost.
+Otherwise, a timer is set for the the reordering threshold minus the time
+difference between the earliest in flight packet and the largest newly
+acknowledged packet.  Note that in some cases the timer could become longer
+when packets are acknowleged out of order. The RECOMMENDED time threshold,
+expressed as a fraction of the round-trip time (kTimeReorderingFraction),
+is 1/8.
 
 ### Early Retransmit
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -163,17 +163,18 @@ across packet number spaces.
...
-for transmissions, and any application data is sent in one or more streams,
-with delivery order determined by stream offsets encoded within STREAM frames.
-
-QUIC's packet number is strictly increasing, and directly encodes transmission
-order.  A higher QUIC packet number signifies that the packet was sent later,
-and a lower QUIC packet number signifies that the packet was sent earlier.  When
-a packet containing frames is deemed lost, QUIC rebundles necessary frames in a
+TCP conflates transmission order at the sender with delivery order at the
+receiver, which results in retransmissions of the same data carrying the same
+sequence number, and consequently to "retransmission ambiguity".  QUIC
```
```suggestion
sequence number, and consequently leads to "retransmission ambiguity".  QUIC
```

draft-ietf-quic-recovery.md
```diff
@@ -236,7 +237,9 @@ underestimation of min RTT, which in turn prevents underestimating smoothed RTT.
 Ack-based loss detection implements the spirit of TCP's Fast Retransmit
 {{?RFC5681}}, Early Retransmit {{?RFC5827}}, FACK, and SACK loss recovery
 {{?RFC6675}}. This section provides an overview of how these algorithms are
-implemented in QUIC.
+implemented in QUIC.  Though both time-based loss detection and early retransmit
+use a timer, they're part of Ack-based detection because they do not use a timer
```
```suggestion
use a timer, they are part of Ack-based detection because they do not use a timer
```

draft-ietf-quic-recovery.md
```diff
@@ -450,11 +458,10 @@ this packet adds network load without establishing packet loss.
...
 QUIC SHOULD delay sending acknowledgements in response to packets, but MUST NOT
-excessively delay acknowledgements of packets containing frames other than ACK.
-Specifically, implementations MUST attempt to enforce a maximum
-ack delay to avoid causing the peer spurious timeouts.  The maximum ack delay
-is communicated in the `max_ack_delay` transport parameter and the default
-value is 25ms.
+excessively delay acknowledgements of retransmittable packets. Specifically,
+implementations MUST attempt to enforce a maximum ack delay to avoid causing
+the peer spurious timeouts.  The maximum ack delay is communicated in the
+`max_ack_delay` transport parameter and the default value is 25ms.
```
mikkelfj: Just a stray thought: would some kind of hysteresis be of relevance? E.g. receiver should allow for 10% overrun of ack delay? Or perhaps this is already taken into account.

draft-ietf-quic-recovery.md
```diff
@@ -501,13 +508,13 @@ frame may be saved.  When a packet containing an ACK frame is acknowledged, the
...
-frames containing them could be lost. In this case, the loss recovery algorithm
-may cause spurious retransmits, but the sender will continue making forward
-progress.
+In cases without ACK frame loss, this algorithm allows for a minimum of 1 RTT
+of reordering. In cases with ACK frame loss and reordering, this approach does
+not guarantee that every acknowledgement is seen by the sender before it is no
+longer included in the ACK frame. Packets could be received out of order and
+all subsequent ACK frames containing them could be lost. In this case, the
+loss recovery algorithm may cause spurious retransmits, but the sender will
+continue making forward progress.
```
```suggestion
continue to make forward progress.
```

draft-ietf-quic-recovery.md
```diff
@@ -878,13 +881,9 @@ Pseudocode for OnLossDetectionTimeout follows:
...
 ### Detecting Lost Packets
 
-Packets in QUIC are only considered lost once a larger packet number in
-the same packet number space is acknowledged.  DetectLostPackets is called
-every time an ack is received and operates on the sent_packets for that
-packet number space.  If the loss detection timer expires and the loss_time
-is set, the previous largest acked packet is supplied.
-
-#### Pseudocode
+DetectLostPackets is called every time an ack is received and operates on
```
```suggestion
DetectLostPackets is called every time an ACK is received and operates on
```

draft-ietf-quic-recovery.md
```diff
@@ -878,13 +881,9 @@ Pseudocode for OnLossDetectionTimeout follows:
...
+the sent_packets for that packet number space. If the loss detection timer
+expires and the loss_time is set, the previous largest acked packet is supplied.
```
```suggestion
expires and the loss_time is set, the previous largest acknowledged packet is supplied.
```

draft-ietf-quic-recovery.md
mikkelfj: or ACK-based ?

draft-ietf-quic-recovery.md
ianswett: The current model is the receiver communicates it's max ack delay.  The receiver is in the best place to determine how late alarms are likely to fire, so I think they should include that in the value they send.  That being said, it's probably worth stating that specifically.  Possibly in the transport draft?

mikkelfj: Makes sense

draft-ietf-quic-recovery.md
ianswett: I think I prefere making, since it's more active?

draft-ietf-quic-recovery.md
ianswett: PR #1960 merged already fyi.

draft-ietf-quic-recovery.md
mikkelfj: To me it seems more like spoken language, but either will do.

draft-ietf-quic-recovery.md
mikkelfj: Yes I noticed, looks good, thanks.

draft-ietf-quic-recovery.md
```diff
@@ -163,17 +163,18 @@ across packet number spaces.
...
-a packet containing frames is deemed lost, QUIC rebundles necessary frames in a
+TCP conflates transmission order at the sender with delivery order at the
+receiver, which results in retransmissions of the same data carrying the same
+sequence number, and consequently leads to "retransmission ambiguity".  QUIC
+separates the two: QUIC uses a packet number to indicate transmission order,
+and any application data is sent in one or more streams, with delivery order
+determined by stream offsets encoded within STREAM frames.
+
+QUIC's packet number is strictly increasing within a packet number space,
+and directly encodes transmission order.  A higher QUIC packet number
```
```suggestion
and directly encodes transmission order.  A higher packet number
```

draft-ietf-quic-recovery.md
```diff
@@ -163,17 +163,18 @@ across packet number spaces.
...
+signifies that the packet was sent later, and a lower QUIC packet number
```
```suggestion
signifies that the packet was sent later, and a lower packet number
```

draft-ietf-quic-recovery.md
```diff
@@ -163,17 +163,18 @@ across packet number spaces.
...
+signifies that the packet was sent earlier.  When a packet containing
+retransmittable frames is deemed lost, QUIC rebundles necessary frames in a
```
```suggestion
retransmittable frames is detected lost, QUIC rebundles necessary frames in a
```

draft-ietf-quic-recovery.md
```diff
@@ -236,7 +237,9 @@ underestimation of min RTT, which in turn prevents underestimating smoothed RTT.
 Ack-based loss detection implements the spirit of TCP's Fast Retransmit
 {{?RFC5681}}, Early Retransmit {{?RFC5827}}, FACK, and SACK loss recovery
 {{?RFC6675}}. This section provides an overview of how these algorithms are
-implemented in QUIC.
+implemented in QUIC.  Though both time-based loss detection and early retransmit
+use a timer, they are part of Ack-based detection because they do not use a
```
```suggestion
use a timer, they are part of ack-based detection because they do not use a
```

martinthomson: the word is "acknowledgment", which I appreciate is long and hard to say and type, but we should use the whole word unless we're talking about the frame.

ianswett: Hmm, the naming of this is a good point, but I think that's best saved for a follow up change, since it's used in a good number of spots.  Also, I was thinking it's ACK like the frame, but I'm not sure we want to go to ACK-based loss detection everywhere?

PR #1960, opened 2018-11-03
Late alarms and max_ack_delay

ianswett:
Add some text on how to correctly specify max_ack_delay on systems where the alarm may fire late.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3929,7 +3929,7 @@ max_ack_delay (0x000b):
 
 : The maximum ACK delay is an integer value indicating the
   maximum amount of time in milliseconds by which the endpoint will delay
-  sending acknowledgments.  This value should include the receiver's expected
+  sending acknowledgments.  This value SHOULD include the receiver's expected
   delays in alarms firing.  For example, if a receiver sets a timer for 5ms
   and alarms commonly fire up to 1ms late, then it should send a max_ack_delay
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3929,8 +3929,11 @@ max_ack_delay (0x000b):
 
 : The maximum ACK delay is an integer value indicating the
   maximum amount of time in milliseconds by which the endpoint will delay
-  sending acknowledgments.  If this value is absent, a default of 25
-  milliseconds is assumed.
+  sending acknowledgments.  This value should include the receiver's expected
+  delays in alarms firing.  For example, if a receiver specifies a 5ms
+  max_ack_delay and alarms commonly fire up to 1ms late, then it should set
+  a timer for 4ms.  If this value is absent, a default of 25 milliseconds is
```
martinthomson: I would do this the other way around.  If a receiver sets ack delay timers at 5ms and alarms could fire up to 1ms late, then it should advertise a max_ack_delay of 6ms.

draft-ietf-quic-transport.md
```diff
@@ -3929,8 +3929,10 @@ max_ack_delay (0x000b):
 
 : The maximum ACK delay is an integer value indicating the
   maximum amount of time in milliseconds by which the endpoint will delay
-  sending acknowledgments.  If this value is absent, a default of 25
-  milliseconds is assumed.
+  sending acknowledgments.  This value should include the receiver's expected
```
```suggestion
  sending acknowledgments.  This value SHOULD include the receiver's expected
```

PR #1961, opened 2018-11-03
pt sm vwls n frm nms

martinthomson:


draft-ietf-quic-http.md
```diff
@@ -564,13 +564,13 @@ When a server receives this frame, it aborts sending the response for the
 identified server push.  If the server has not yet started to send the server
 push, it can use the receipt of a CANCEL_PUSH frame to avoid opening a push
 stream.  If the push stream has been opened by the server, the server SHOULD
-send a QUIC RST_STREAM frame on that stream and cease transmission of the
+send a QUIC RESET_STREAM frame on that stream and cease transmission of the
```
nibanks: I've been thinking, should all stream specific frames (i.e. frames that have an ID) start with `STREAM_`? In other words, this would be `STREAM_RESET` instead. Not sure exactly what to change `STOP_SENDING` to instead though...

martinthomson: I'm settling for mentions of STREAM. STOP_SENDING is definitely awkward, but since it doesn't affect stream state, maybe that's OK.

Issue #1212, opened 2018-03-15
confusing Fast Retransmit modes


marten-seemann:
The loss recovery spec describes two different (and mutually exclusive) Fast Retransmit modes: one detects reordering in packet number space, the other one in time space.

The Fast Retransmit section is not clear about the fact that implementations have to choose one or the other mode, and just mentions time reordering at the very end.

Furthermore, I find the pseudo code for `DetectLostPackets` pretty hard to read, since depending on the mode, different variables are set to infinite and then used in comparisons. It would probably be easier to provide two different definitions of this functions depending on the mode.

I'd be happy to write up a PR, if people agree that the current text needs improvement in that direction.

ianswett:
Agreed, though technically both of these could run at once, I think it's best to describe them as two mutually exclusive modes.

A PR to improve the text and pseudocode would be appreciated, thanks.

siyengar:
ya there's some places which use the delayUntilLost for for time space for the early retransmit threshold. It might be nice to separate those two out to show that early retransmit is always active.

pravb:
I don't think these are mutually exclusive modes. With TCP today we do both, if the SACK information suggests lost packets or RACK kicks in and marks packets lost. Time based detection cannot work for low latency connections due to lack of fine grained timers so you always need to augment with packet based detection. IMO an implementation should do both.

martinthomson:
@ianswett, you seem to think that this is editorial, but it is still labeled design.  It's clearly a hot issue.  Do you have a plan to fix it?

janaiyengar:
Closed by #1974

Issue #934, opened 2017-11-15
Early retransmit code


siyengar:
If we're in the early retransmit mode, a loss_time is set. If the timer is not yet fired, however we send a packet when the timer is armed, the largest_sent_packet changes. When the timer does fire before this new packet is acked, largest_sent_packet != largest_acked_packet, so delay_until_lost == infinite and none of the packets will be declared lost, however we would instead set a TLP alarm. 

Is this intentional? Seems a bit weird.

martinthomson:
Note: Questions are better asked on the mailing list.

siyengar:
Ya it was a bit fuzzy on whether this was a bug in the draft, so I raised this as an issue.

ianswett:
@siyengar You volunteered to write a PR, can you try fixing this issue you filed?

siyengar:
will do

martinthomson:
@siyengar, some time has passed.  Do you still intend to do this?

siyengar:
@martinthomson  put up https://github.com/quicwg/base-drafts/pull/2021. Sorry about the delay

janaiyengar:
I think this issue goes away with #1974.

janaiyengar:
@siyengar, does the merged #1974 resolve this? I believe it does.

PR #1226, opened 2018-03-18
split fast retransmit modes

marten-seemann:
Fixes #1212.

Requesting a review from @ianswett, @siyengar, @janaiyengar.

PR #1974, opened 2018-11-06
Recovery uses both time and packet thresholds

### LINKED ISSUE #1969: Remove kUsingTimeLossDetection from Recovery draft
@pravb Indicated declaring a packet lost if the reordering threshold(ie: 3) OR the time threshold has been reached, then a packet is declared lost.

I'd like to remove early retransmit entirely, because it's a bit complex and having QUIC operating in two alternate modes, but indicating both can be enabled at once in the text is very confusing.

ianswett:
Based on @pravb feedback at tcpm and my anecdotal experience, combining the two should work well.

Fixes #1969, #1212, #934

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -271,8 +271,12 @@ they do not use a timer to send probes, but rather to declare packets lost.
 ### Fast Retransmit
 
 An unacknowledged packet is marked as lost when an acknowledgment is received
-for a packet that was sent kPacketThreshold number of packets or
-kTimeThreshold amount of time after the unacknowledged packet.
+for a packet sent sufficiently after the unacknowledged packet.
+* The largest acknowledged packet is more than kPacketThreshold packets larger
+  than the unacknowleged packet.
+* The unacknowledged packet was sent more than
+  (1 + kTimeThreshold) * max(SRTT, latest_RTT) ago.
+
 Receipt of the acknowledgement indicates that a later packet was received,
 while the reordering threshold provides some tolerance for reordering of
 packets in the network.
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -240,11 +240,10 @@ underestimation of min RTT, which in turn prevents underestimating smoothed RTT.
 
 Ack-based loss detection implements the spirit of TCP's Fast Retransmit
 {{?RFC5681}}, Early Retransmit {{?RFC5827}}, FACK, SACK loss recovery
-{{?RFC6675}}, and RACK {{draft-ietf-tcpm-rack-04}}. This section provides an
-overview of how these algorithms are implemented in QUIC.  Though both
-time-based loss detection and early retransmit use a timer, they are part
-of ack-based detection because they do not use a timer to send probes,
-but rather to declare packets lost.
+{{?RFC6675}}, and RACK. This section provides an overview of how these
+algorithms are implemented in QUIC.  Though both time-based loss detection
+and early retransmit use a timer, they are part of ack-based detection because
+they do not use a timer to send probes, but rather to declare packets lost.
 
 ### Fast Retransmit
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -285,21 +285,6 @@ acknowledged packet.  Note that in some cases the timer could become longer when
 packets are acknowleged out of order. The RECOMMENDED time threshold, expressed
 as a fraction of the round-trip time (kTimeReorderingFraction), is 1/8.
 
-### Early Retransmit
-
-Unacknowledged packets close to the tail may have fewer than
-kReorderingThreshold retransmittable packets sent after them.  Loss of such
-packets cannot be detected via Packet Threshold Fast Retransmit. To enable
-ack-based loss detection of such packets, receipt of an acknowledgment for
-the last outstanding retransmittable packet triggers the Early Retransmit
-process, as follows.
-
-If there are unacknowledged in-flight packets still pending, they should
-be marked as lost. To compensate for the reduced reordering resilience, the
-sender SHOULD set a timer for a small period of time. If the unacknowledged
-in-flight packets are not acknowledged during this time, then these
-packets MUST be marked as lost.
-
 An endpoint SHOULD set the timer such that a packet is marked as lost no earlier
 than 1.125 * max(SRTT, latest_RTT) since when it was sent.
 
@@ -316,12 +301,6 @@ The 1.125 multiplier increases reordering resilience. Implementers MAY
 experiment with using other multipliers, bearing in mind that a lower multiplier
 reduces reordering resilience and increases spurious retransmissions, and a
 higher multiplier increases loss recovery delay.
-
-This mechanism is based on Early Retransmit for TCP {{?RFC5827}}. However,
-{{?RFC5827}} does not include the timer described above. Early Retransmit is
-prone to spurious retransmissions due to its reduced reordering resilence
-without the timer. This observation led Linux TCP implementers to implement a
-timer for TCP as well, and this document incorporates this advancement.
 
 ## Timeout Loss Detection
 
@@ -597,10 +576,6 @@ kTimeReorderingFraction:
 : Maximum reordering in time space before time based loss detection considers
   a packet lost.  In fraction of an RTT. The RECOMMENDED value is 1/8.
 
-kUsingTimeLossDetection:
-: Whether time based loss detection is in use.  If false, uses FACK style
-  loss detection. The RECOMMENDED value is false.
-
 kMinTLPTimeout:
 : Minimum time in the future a tail loss probe timer may be set for.
   The RECOMMENDED value is 10ms.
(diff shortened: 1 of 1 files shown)


marten-seemann:
This would make #1226 obsolete, right?

ianswett:
Yes @marten-seemann , the hope is to make #1226 obsolete, though I suspect this PR needs a bit more work as well.

draft-ietf-quic-recovery.md
```diff
@@ -274,19 +274,14 @@ TCP-NCR {{?RFC4653}}, to improve QUIC's reordering resilience.
...
+thresholds.  It is recommended this is is used in conjunction with packet
+threshold fast retransmit. The RECOMMENDED time threshold, expressed
 as a fraction of the round-trip time (kTimeReorderingFraction), is 1/8.
 
-An endpoint SHOULD set the timer such that a packet is marked as lost no earlier
-than 1.125 * max(SRTT, latest_RTT) since when it was sent.
+An endpoint SHOULD declare packets lost no earlier than
+1.125 * max(SRTT, latest_RTT) after when they were sent.  If packets sent prior
+to the largest acknowledged packet cannot yet be declared lost, then a timer
+SHOULD be set for the remainint time.
```
mikkelfj: remaining

draft-ietf-quic-recovery.md
```diff
@@ -312,16 +292,10 @@ Using max(SRTT, latest_RTT) protects from the two following cases:
...
-experiment with using other multipliers, bearing in mind that a lower multiplier
-reduces reordering resilience and increases spurious retransmissions, and a
-higher multiplier increases loss recovery delay.
-
-This mechanism is based on Early Retransmit for TCP {{?RFC5827}}. However,
-{{?RFC5827}} does not include the timer described above. Early Retransmit is
-prone to spurious retransmissions due to its reduced reordering resilence
-without the timer. This observation led Linux TCP implementers to implement a
-timer for TCP as well, and this document incorporates this advancement.
+Implementers MAY experiment with using other reordering fractoins, bearing
```
mikkelfj: fractions

siyengar:
i think this diff would have to change SetLossDetectionAlarm as well. Currently SetLossDetectionAlarm sets the loss alarm based on the last unacked packet

```
   loss_detection_timer.set(
      time_of_last_sent_retransmittable_packet + timeout)
```

However in the case of time reordering, we would have to set the time based on the least unacked packet instead.

draft-ietf-quic-recovery.md
```diff
@@ -239,20 +239,20 @@ underestimation of min RTT, which in turn prevents underestimating smoothed RTT.
...
 
 ### Fast Retransmit
 
 An unacknowledged packet is marked as lost when an acknowledgment is received
 for a packet that was sent a threshold number of packets (kReorderingThreshold)
-and/or a threshold amount of time after the unacknowledged packet. Receipt of
-the acknowledgement indicates that a later packet was received, while the
-reordering threshold provides some tolerance for reordering of packets in the
-network.
+or a threshold amount of time (kTimeReorderingFraction) after the
```
janaiyengar: how 'bout just kPacketThreshold and kTimeThreshold? The context makes clear what it's used for and exactly what the units are.

draft-ietf-quic-recovery.md
```diff
@@ -274,34 +274,14 @@ TCP-NCR {{?RFC4653}}, to improve QUIC's reordering resilience.
...
-threshold or in addition to it.
-
-When a larger packet is acknowledged, if it was sent more than the threshold
-after any in flight packets, those packets are immediately declared lost.
-Otherwise, a timer is set for the the reordering threshold minus the time
-difference between the earliest in flight packet and the largest newly
-acknowledged packet.  Note that in some cases the timer could become longer when
-packets are acknowleged out of order. The RECOMMENDED time threshold, expressed
+thresholds.  It is recommended this is is used in conjunction with packet
+threshold fast retransmit. The RECOMMENDED time threshold, expressed
```
janaiyengar: This sentence is mostly unclear ("in conjunction with packet threshold fast retransmit") and not really useful. I would drop it entirely.

draft-ietf-quic-recovery.md
```diff
@@ -274,34 +274,14 @@ TCP-NCR {{?RFC4653}}, to improve QUIC's reordering resilience.
...
-
-If there are unacknowledged in-flight packets still pending, they should
-be marked as lost. To compensate for the reduced reordering resilience, the
-sender SHOULD set a timer for a small period of time. If the unacknowledged
-in-flight packets are not acknowledged during this time, then these
-packets MUST be marked as lost.
-
-An endpoint SHOULD set the timer such that a packet is marked as lost no earlier
-than 1.125 * max(SRTT, latest_RTT) since when it was sent.
+An endpoint SHOULD declare packets lost no earlier than
```
janaiyengar: I don't think the early retransmit section should be removed. This is a separate mechanism, and it is the only thing that uses the time threshold. I think a different structure might be useful... I'd actually argue for removing the time threshold section and leaving ER as is, since the time threshold section doesn't add much.

draft-ietf-quic-recovery.md
```diff
@@ -312,16 +292,10 @@ Using max(SRTT, latest_RTT) protects from the two following cases:
 * the latest RTT sample is higher than the SRTT, perhaps due to a sustained
   increase in the actual RTT, but the smoothed SRTT has not yet caught up.
 
-The 1.125 multiplier increases reordering resilience. Implementers MAY
-experiment with using other multipliers, bearing in mind that a lower multiplier
-reduces reordering resilience and increases spurious retransmissions, and a
-higher multiplier increases loss recovery delay.
-
-This mechanism is based on Early Retransmit for TCP {{?RFC5827}}. However,
```
janaiyengar: This is still true. I would leave this in.

draft-ietf-quic-recovery.md
```diff
@@ -841,7 +807,7 @@ Pseudocode for SetLossDetectionTimer follows:
         time_of_last_sent_crypto_packet + timeout)
       return
     if (loss_time != 0):
-      // Early retransmit timer or time loss detection.
+      // Time threshold loss detection.
```
```suggestion
      // Early retransmit timer.
```

draft-ietf-quic-recovery.md
```diff
@@ -875,7 +841,7 @@ Pseudocode for OnLossDetectionTimeout follows:
        RetransmitUnackedCryptoData()
        crypto_count++
      else if (loss_time != 0):
-       // Early retransmit or Time Loss Detection
+       // Time threshold loss Detection
```
```suggestion
       // Early retransmit
```

janaiyengar:
FWIW, I'm in favor of this algorithmic change (remove time-based loss detection), but this change does not remove ER.  Early retransmit is exactly what the timer is used for, which gets armed if there are < 3 packets beyond the unacked packet. What we want is FR-with-packet-threshold and ER-with-time-threshold.

draft-ietf-quic-recovery.md
ianswett: I'm confused, so let's talk today.

draft-ietf-quic-recovery.md
```diff
@@ -907,14 +872,8 @@ Pseudocode for DetectLostPackets follows:
...
   lost_packets = {}
-  delay_until_lost = infinite
-  if (kUsingTimeLossDetection):
-    delay_until_lost =
-      (1 + time_reordering_fraction) *
-          max(latest_rtt, smoothed_rtt)
-  else if (largest_acked.packet_number == largest_sent_packet):
-    // Early retransmit timer.
-    delay_until_lost = 9/8 * max(latest_rtt, smoothed_rtt)
+  delay_until_lost = (1 + time_reordering_fraction) *
```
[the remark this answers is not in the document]
ianswett: This only is used if there are retransmittable packets smaller than the largest acked, unlike TLP which aims for tail losses.

draft-ietf-quic-recovery.md
```diff
@@ -239,20 +239,20 @@ underestimation of min RTT, which in turn prevents underestimating smoothed RTT.
...
 ### Fast Retransmit
 
 An unacknowledged packet is marked as lost when an acknowledgment is received
-for a packet that was sent a threshold number of packets (kReorderingThreshold)
-and/or a threshold amount of time after the unacknowledged packet. Receipt of
-the acknowledgement indicates that a later packet was received, while the
-reordering threshold provides some tolerance for reordering of packets in the
-network.
+for a packet that was sent kReorderingThreshold number of packets or
+kTimeReorderingFraction amount of time after the unacknowledged packet.
```
janaiyengar: suggestion: expand this into a bulleted list. "Fast Retransmit uses two thresholds to mark packets as lost..."

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
siyengar: not sure what you mean.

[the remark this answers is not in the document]
ianswett: loss_time will only be non-0 if there are unacknowledged retransmittable packets sent prior to the largest acked packet.  In that case, the loss threshold declares the unacknowledged packet lost.  If there are no unacked packets less than largest acked, then the TLP would be armed, and as you said, it has a longer timeout.

janaiyengar:
I was thinking about this as a fix to ER, but I'm now convinced that this is a better framing. I'm suggesting some editorial changes in #2080, but I'm quite happy with the direction this is heading in.

mikkelfj:
I'm late on this. But two questions:

1) Should not `delay_until_lost` include peers ack delay? If you have a very fast network a nearby peer may have larger delay than the RTT.

2) Should not the `delta` in pseudocode for detecting lost packets by packet number take gaps into consideration. Notably gaps are used for optimistic ACK protection.

PR #1975, opened 2018-11-06
Mention CUBIC as an example of an alternative congestion control mechanism

loganaden:
At least one known implementation is using BBR:
https://cs.chromium.org/chromium/src/net/third_party/quic/core/congestion_control/bbr_sender.cc?g=0

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -918,7 +918,12 @@ packet is a probe packet sent after the TLP or RTO timer expires, as described
 in {{tlp}} and {{rto}}.
 
 Implementations MAY use other congestion control algorithms, and endpoints MAY
-use different algorithms from one another. The signals QUIC provides for
+use different algorithms from one another. An example would be the
+experimental BBR Congestion Control Mechanism
+{{?I-D.cardwell-iccrg-bbr-congestion-control}} which minimises induced
+latency along delivery path.
+
+The signals QUIC provides for
 congestion control are generic and are designed to support different algorithms.
 
 ## Explicit Congestion Notification {#congestion-ecn}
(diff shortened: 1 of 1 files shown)


ekr:
I don't think it's helpful to specifically cite BBR here. There are a lot of alternatives.

martinthomson:
I don't think that you could reasonably say that the authors were unaware of BBR.  You might then conclude that not mentioning BBR was intentional.

loganaden:
@ianswett i understand better now. I can change the PR to mention Cubic instead.

loganaden:
@martinthomson understood.

draft-ietf-quic-recovery.md
```diff
@@ -918,7 +918,10 @@ packet is a probe packet sent after the TLP or RTO timer expires, as described
 in {{tlp}} and {{rto}}.
 
 Implementations MAY use other congestion control algorithms, and endpoints MAY
-use different algorithms from one another. The signals QUIC provides for
+use different algorithms from one another. An example would be the
```
```suggestion
use other algorithms, such as Cubic {{?RFC8312}}.  The signals QUIC provides for
```

draft-ietf-quic-recovery.md
```diff
@@ -918,7 +918,10 @@ packet is a probe packet sent after the TLP or RTO timer expires, as described
...
+experimental BBR Congestion Control Mechanism
```
```suggestion
```

draft-ietf-quic-recovery.md
```diff
@@ -918,7 +918,10 @@ packet is a probe packet sent after the TLP or RTO timer expires, as described
...
+{{?I-D.cardwell-iccrg-bbr-congestion-control}} which minimises induced
```
```suggestion
```

draft-ietf-quic-recovery.md
```diff
@@ -918,7 +918,10 @@ packet is a probe packet sent after the TLP or RTO timer expires, as described
...
+latency along delivery path. The signals QUIC provides for
```
```suggestion
```

loganaden:
@mikkelfj noted. I will work on updating the generic signal section if it goes through.

MikeBishop:
Belated, but I don't like this change.  The original text makes two points:
- You don't have to use the congestion control mechanism we describe here
- Each peer can choose independently

The changed text says the first point twice, and drops the second one.

PR #1988, opened 2018-11-10
Add spin bit to transport draft

britram:
This PR adds the spin bit to the short header in the transport draft, referring to draft-ietf-quic-spin-exp for details. This reference is temporary, while the details of the spin bit are worked out in that document. The intention is then to take the content of spin-exp and incorporate it into the transport and manageability drafts. 

It also changes the status of spin-exp to Standards Track, to avoid a downref from -transport to an Experimental draft.

Note that the details of which bit is the spin bit will change when the first octet discussion resolves; this PR is only meant to establish the reference.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3411,7 +3411,7 @@ Google QUIC Demultiplexing Bit:
 Spin Bit:
 
 : The sixth bit (0x4) of byte 0 is the Latency Spin Bit, set as described in
-{{!SPIN:I-D.ietf-quic-spin-exp}}.
+{{!SPIN=I-D.ietf-quic-spin-exp}}.
 
 Reserved:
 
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3408,9 +3408,14 @@ Google QUIC Demultiplexing Bit:
...
 
+Spin Bit:
+
+: The sixth bit (0x4) of byte 0 is the Latency Spin Bit, set as described in
+{{!SPIN=I-D.ietf-quic-spin-exp}}.
+
 Reserved:
 
-: The sixth, seventh, and eighth bits (0x7) of byte 0 are reserved for
+: The seventh and eighth bits (0x3) of byte 0 are reserved for
```
marten-seemann: I don't think we're reserving any more bits for experimentation at this point.

ferrieux: By un-reserving them you're shorting any hope for other measurement, like loss bits. DO you decide this kind of things single-handedly ?

martinthomson:
So this will change completely, so what the transport doc says doesn't really matter.  Merging.

PR #1992, opened 2018-11-12
Endpoint MAY discard CIDs

janaiyengar:
An endpoint should not assume that its peer will not drop any received CIDs.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -939,7 +939,7 @@ the connection ID via a RETIRE_CONNECTION_ID frame
 
 Endpoints store received connection IDs for future use.  An endpoint that
 receives excessive connection IDs MAY discard those it cannot store.  An
-endpoint that issues connection IDs should expect that its peer might not store
+endpoint that issues connection IDs cannot expect its peer to store
 and use all issued connection IDs.
 
 An endpoint SHOULD ensure that its peer has a sufficient number of available and
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -925,17 +925,22 @@ handshake.  The sequence number of the initial connection ID is 0.  If the
 preferred_address transport parameter is sent, the sequence number of the
 supplied connection ID is 1.
 
-Additional connection IDs are communicated to the
-peer using NEW_CONNECTION_ID frames ({{frame-new-connection-id}}).  The
-sequence number on each newly-issued connection ID MUST increase by 1. The
-connection ID randomly selected by the client in the Initial packet and any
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -925,18 +925,23 @@ handshake.  The sequence number of the initial connection ID is 0.  If the
...
+initial connection ID.
 
 When an endpoint issues a connection ID, it MUST accept packets that carry this
 connection ID for the duration of the connection or until its peer invalidates
 the connection ID via a RETIRE_CONNECTION_ID frame
 ({{frame-retire-connection-id}}).
 
+Endpoints store received connection IDs for future use.  An endpoint that
+receives excessive connection IDs MAY discard those it cannot store.  An
+endpoint that issues connection IDs should expect that its peer might not store
```
```suggestion
endpoint that issues connection IDs cannot expect its peer to store
```

mikkelfj:
I think this is messy - I know there have been lengthy discussions. But sender also may have to track state. It would be better to have a fixed TP length of say, 5 CID's and only replenish when one is retired.

mikkelfj:
If I understand @marten-seemann correctly, this is in line with what I propose, and which others have proposed earlier.

janaiyengar:
@marten-seemann @mikkelfj : Like I said on list, I think it's fine to drive that issue (please do file an issue first), and send a PR along (ASAP). I'm merely clarifying existing text.

draft-ietf-quic-transport.md
```diff
@@ -925,18 +925,23 @@ handshake.  The sequence number of the initial connection ID is 0.  If the
...
 
 When an endpoint issues a connection ID, it MUST accept packets that carry this
 connection ID for the duration of the connection or until its peer invalidates
 the connection ID via a RETIRE_CONNECTION_ID frame
 ({{frame-retire-connection-id}}).
 
+Endpoints store received connection IDs for future use.  An endpoint that
+receives excessive connection IDs MAY discard those it cannot store.  An
+endpoint that issues connection IDs cannot expect its peer to store and use all
+issued connection IDs.
```
ianswett: Belated question: when you discard them, it seems you should not send a RETIRE_CONNECTION_ID frame, otherwise you may just get more?

MikeBishop: The idea is that you can have an uncoordinated negotiation:  I'm willing to issue 64, you only need 8.  You take the 8 you want, let the remaining 56 fall on the floor, and then I'll replace the 8 you kept as you RETIRE them.  If you RETIREd the 56 you didn't need, I'd replace *those*, and we'd get into an endless RETIRE/NEW drop for no purpose.

PR #2000, opened 2018-11-13
Discard, but don't retire, excess connection IDs

martinthomson:
Followup for #1992 .


janaiyengar:
Thanks for fixing this up.

PR #2001, opened 2018-11-13
Refine minimum DCID length of Initial

martinduke:
This is meant to almost entirely mitigate #1810 by making sure that the reply to a VN packet ends up at the same server. It reduces the spurious downgrade attack signal to a corner case of a corner case.

It is much more lightweight than alternatives, if we're looking for a minimum-footprint change.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1319,9 +1319,8 @@ abandons the connection attempt and starts a new one. The initial Destination
 Connection ID is used to determine packet protection keys for Initial packets.
 
 A client SHOULD select a Destination Connection ID length long enough to fulfill
-the minimum for every QUIC version it supports. This makes it easier to detect
-version downgrade attacks, by increasing the chance Initial packets are routed
-to the same server.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1318,6 +1318,11 @@ packet is received from the server, the client MUST use the same value unless it
 abandons the connection attempt and starts a new one. The initial Destination
 Connection ID is used to determine packet protection keys for Initial packets.
 
+A client SHOULD select a Destination Connection ID length long enough to fulfill
+the minimum for every QUIC version it supports. This makes it easier to detect
+version downgrade attacks, by increasing the chance Initial packets are routed
+to the same server.
```
martinthomson: This second sentence worries me.  The mechanism we have for detecting version downgrade is robust without consistent routing.  The problem here is that it might be too robust and - during a server upgrade - the server might change its mind.  Having stable routing ensures that the server can maintain state during the cut-over and avoid connections being killed as a result.  Given the complexities involved, I would suggest that we limit this to just the first sentence.

martinduke:
can we merge this?

PR #2003, opened 2018-11-14
Interaction of FIN and message parsing

### LINKED ISSUE #1972: http deadlock with stream FIN choices
During interop, I couldn't interop with winquic because I never FIN my request stream, and winquic doesn't serve the response until the request FINs.

My conversation with @MikeBishop suggests that both of us are behaving legally, which is a problem. We should either force clients to FIN (allowing servers to wait for it) or not (not). I don't particularly care which.

MikeBishop:
Fixes #1972.

While the discussion on the issue wasn't conclusive in either direction, I think the presence of CONNECT and early response means servers will need to be processing HTTP messages as they arrive, rather than waiting for the FIN as an end-of-message signal.  Also, if the message is large, the alternative risks deadlock on a large POST.

This PR expands on the "stream state change isn't part of the message framing" language to clarify that both peers MUST be able to process complete HTTP messages regardless of whether the stream has been closed or not.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -874,13 +874,12 @@ After sending a request, a client SHOULD close the stream for sending; after
 sending a final response, the server SHOULD close the stream for sending. At
 this point, the QUIC stream is fully closed.
 
-Changes to the state of a request stream do not directly affect message
-processing.  Endpoints MUST process complete HTTP messages without relying on
-stream closure as an end-of-message signal.  For example, servers do not abort a
-response in progress solely due to a state change on the request stream.
-However, if a client stream terminates without containing a usable HTTP message,
-the server SHOULD abort its response with the error code
-HTTP_INCOMPLETE_REQUEST.
+When a stream is closed, this indicates the end of an HTTP message.
+Because some messages are large or unbounded, endpoints SHOULD begin processing
+partial HTTP messages once enough of the message has been received to make
+progress.  If a client stream terminates without enough of the HTTP message to
+provide a complete response, the server SHOULD abort its response with the error
+code HTTP_INCOMPLETE_REQUEST.
 
 A server can send a complete response prior to the client sending an entire
 request if the response does not depend on any portion of the request that has
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -870,16 +870,18 @@ response to the same request.  Non-final responses do not contain a payload body
 or trailers.
 
 An HTTP request/response exchange fully consumes a bidirectional QUIC stream.
-After sending a request, a client MUST close the stream for sending; after
-sending a final response, the server MUST close the stream for sending. At
-this point, the QUIC stream is fully closed.
+After sending a request, a client MUST close the stream for sending; clients
+MUST NOT make stream closure dependent on receiving a response to their request,
+unless using the CONNECT method (see {{the-connect-method}}). After sending a
+final response, the server MUST close the stream for sending. At this point, the
+QUIC stream is fully closed.
 
-When a stream is closed, this indicates the end of an HTTP message.
-Because some messages are large or unbounded, endpoints SHOULD begin processing
-partial HTTP messages once enough of the message has been received to make
-progress.  If a client stream terminates without enough of the HTTP message to
-provide a complete response, the server SHOULD abort its response with the error
-code HTTP_INCOMPLETE_REQUEST.
+When a stream is closed, this indicates the end of an HTTP message. Because some
+messages are large or unbounded, endpoints SHOULD begin processing partial HTTP
+messages once enough of the message has been received to make progress.  If a
+client stream terminates without enough of the HTTP message to provide a
+complete response, the server SHOULD abort its response with the error code
+HTTP_INCOMPLETE_REQUEST.
 
 A server can send a complete response prior to the client sending an entire
 request if the response does not depend on any portion of the request that has
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -870,11 +870,11 @@ response to the same request.  Non-final responses do not contain a payload body
 or trailers.
 
 An HTTP request/response exchange fully consumes a bidirectional QUIC stream.
-After sending a request, a client MUST close the stream for sending; clients
-MUST NOT make stream closure dependent on receiving a response to their request,
-unless using the CONNECT method (see {{the-connect-method}}). After sending a
-final response, the server MUST close the stream for sending. At this point, the
-QUIC stream is fully closed.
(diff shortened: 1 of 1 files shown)


RyanAtGoogle:
Maybe I'm not thinking about this the right way, but if a body does not have a content-length, then how is the end of the message determined, if not by the end of the stream?

MikeBishop:
That's true, and should have been obvious.  Ugh.

Let's take that point back to the issue; this PR might be DOA.

draft-ietf-quic-http.md
```diff
@@ -860,20 +860,27 @@ details.
...
-final response, the server closes the stream for sending and the QUIC stream is
-fully closed.  Requests and responses are considered complete when the
-corresponding QUIC stream is closed in the appropriate direction.
+After sending a request, a client SHOULD close the stream for sending; after
+sending a final response, the server SHOULD close the stream for sending. At
+this point, the QUIC stream is fully closed.
+
+Changes to the state of a request stream do not directly affect message
+processing.  Endpoints MUST process complete HTTP messages without relying on
+stream closure as an end-of-message signal.  For example, servers do not abort a
```
martinthomson: This is wrong.  The statement to make is:

The end of a stream (in either direction) indicates the end of a message.
Incomplete messages can be processed, and ideally endpoints do that after receiving a complete header block.

draft-ietf-quic-http.md
```diff
@@ -860,20 +860,26 @@ details.
...
 informational responses (1xx, see {{!RFC7231}}, Section 6.2) precede a final
 response to the same request.  Non-final responses do not contain a payload body
 or trailers.
 
 An HTTP request/response exchange fully consumes a bidirectional QUIC stream.
-After sending a request, a client closes the stream for sending; after sending a
-final response, the server closes the stream for sending and the QUIC stream is
-fully closed.  Requests and responses are considered complete when the
-corresponding QUIC stream is closed in the appropriate direction.
+After sending a request, a client SHOULD close the stream for sending; after
```
```suggestion
After sending a request, a client MUST close the stream for sending; after
```

draft-ietf-quic-http.md
```diff
@@ -860,20 +860,26 @@ details.
...
+sending a final response, the server SHOULD close the stream for sending. At
```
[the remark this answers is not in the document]
martinthomson: This is me agreeing with @kazuho.

```suggestion
sending a final response, the server MUST close the stream for sending. At
```

The nice thing is that we never really say "when".

MikeBishop:
@RyanAtGoogle and @martinduke, I'd ideally like your input before I merge this.

martinduke:
Thanks, this addresses my concern. I don't see how the client MUST has any force given that there's no time limit; perhaps it ought to be a SHOULD?

draft-ietf-quic-http.md
```diff
@@ -860,20 +860,26 @@ details.
...
 or trailers.
 
 An HTTP request/response exchange fully consumes a bidirectional QUIC stream.
-After sending a request, a client closes the stream for sending; after sending a
-final response, the server closes the stream for sending and the QUIC stream is
-fully closed.  Requests and responses are considered complete when the
-corresponding QUIC stream is closed in the appropriate direction.
+After sending a request, a client MUST close the stream for sending; after
+sending a final response, the server MUST close the stream for sending. At
+this point, the QUIC stream is fully closed.
```
martinthomson: To @martinduke's comment, maybe it is enough to say "A message is terminated when the stream is closed."  Instead of all these words.

martinthomson: Oh, and now I realize that this is what it said before...  Not sure whether this is an improvement then.

MikeBishop: Yeah, the original provoking issue was that he didn't FIN, the server didn't respond, and both are doing totally spec-compliant things.  With this language, it's at least clarified that while the server SHOULD process that request on receipt, it isn't required to until the stream closes, so the client needs to close promptly.

MikeBishop:
You're right that a MUST with no deadline is somewhat toothless, but I don't think it's SHOULD-level.  (If we were to fully expand it, you MUST close it eventually and SHOULD do so promptly.)

I think the force is actually in the *server's* SHOULD:  While the server SHOULD process messages before you FIN, it's not obligated to.  Therefore, if you want to ensure a response, you'll need to close promptly as a practical matter.

martinduke:
It sounds like there's a blanket rule except for one special flower - HTTP_CONNECT. So perhaps we should just call out the exception rather than dancing around it, e.g.

"With the exception of requests that contain the CONNECT method, senders use a FIN to indicate the end of an HTTP request or response.  Absent CONNECT, senders MUST NOT make the sending of FIN dependent on receipt of a response from the peer. 

Absent CONNECT, senders MUST NOT send additional data on the stream after the request or response.

However, due to the possible presence of a CONNECT method, and the opportunity to send HTTP_EARLY_DATA, servers SHOULD begin processing partial HTTP messages once enough of the message has been received to make progress.  If a client stream terminates without enough of the HTTP message to provide a complete response, the server SHOULD abort its response with the error code HTTP_INCOMPLETE_REQUEST."

This MUST strikes me as a testable condition.

MikeBishop:
Formally, the data transferred on the TCP connection is the body of the HTTP request, so it's not a special flower in that sense.  But it's special in that it's the only case where the server MUST process the headers before the end of the "message."

The requirement that the client MUST NOT make FIN dependent on receiving anything would be a good wording improvement.

martinduke:
Alright, how about this:

"Senders use a FIN to indicate the end of an HTTP request or response. Absent a CONNECT method in the request, senders MUST NOT make the sending of FIN dependent on receipt of a response from the peer.

However, due to the possible presence of a CONNECT method, and the opportunity to send HTTP_EARLY_DATA, servers SHOULD begin processing partial HTTP messages once enough of the message has been received to make progress. If a client stream terminates without enough of the HTTP message to provide a complete response, the server SHOULD abort its response with the error code HTTP_INCOMPLETE_REQUEST."

I'm open to making the server requirement a similar MUST -- "servers MUST NOT be dependent on a FIN to start processing..." if people think that would help avoid CONNECT deadlocks.

martinthomson:
Not seeing how early data (or HTTP_EARLY_DATA, whatever that is) has any bearing on this.  I would drop the first particle from the second paragraph and concentrate on the recommendations: process when you can, use HTTP_INCOMPLETE_REQUEST if you need to.

I don't think we need a MUST attached to the recommendations - if you don't follow the recommendations, then it's your loss.

martinduke:
> or HTTP_EARLY_DATA, whatever that is

Argh, I meant HTTP_EARLY_RESPONSE.

As for the final wording, I am happy with whatever you guys decide.

MikeBishop:
Did something along those lines -- see what you think.

PR #2004, opened 2018-11-14
Mike's recovery comments on Cubic

ianswett:
Mike's comments on PR #1975


loganaden:
I agree with this change. Thanks for catching this.

PR #2021, opened 2018-11-20
Account for loss_time in DetectLostPackets

siyengar:
Fixes https://github.com/quicwg/base-drafts/issues/934. If loss time is set, we should account for it in deciding whether or not the packet was lost. This helps in the case that a packet is sent while the early retransmit alarm was set.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -894,7 +894,6 @@ Pseudocode for DetectLostPackets follows:
 
 ~~~
 DetectLostPackets(largest_acked):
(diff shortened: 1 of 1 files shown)


janaiyengar:
I think this issue goes away with #1974.

siyengar:
@janaiyengar ya I think that would fix it, will comment on that

ianswett:
@siyengar I think this has been resolved?  If not, can we work it into the updated version of #2066 ?

siyengar:
This looks like it’s fixed now

PR #2023, opened 2018-11-20
Restarting from idle

### LINKED ISSUE #2007: congestion window should be reduced after RTO when no pacing
Otherwise, the sender can burst the entire cwnd out on a single ack.

ianswett:
When restarting from idle, limit the initial burst and pace or reduce the CWND if not pacing.

Fixes #2007

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1042,21 +1042,20 @@ in Linux (3.11 onwards).
 
 ## Restart after idle
 
-A connection is idle if bytes in flight is 0 and there is nothing
-retransmittable to send.  This occurs when the connection is application
-limited and after a verified retransmission timeout.  In order to limit the
-size of bursts sent into the network, the behavior when restarting from idle
-depends upon whether pacing is used.
+A connection is idle if there are no bytes in flight and there is no pending
+retransmittable data to send.  This can occur when the connection is
+application limited or after a retransmission timeout. In order to limit
+the size of bursts sent into the network, the behavior when restarting from
+idle depends upon whether pacing is used.
 
-If pacing is used, the connection should limit the initial burst of packets to
-no more than the initial congestion window and subsequent packets SHOULD be
-paced. The congestion window does not change while the connection is idle.
+If the sender uses pacing, the connection should limit the initial burst of
+packets to no more than the initial congestion window and subsequent packets
+SHOULD be paced. The congestion window does not change while the connection
+is idle.
 
-If pacing is not used, the congestion window SHOULD be reset to the minimum of
-the current congestion window and the initial congestion window.  If the
-slow start threshold is larger than the congestion window, the congestion window
-will grow back to the congestion window prior to idle via slow start.  This
-recommendation is based on Section 4.1 of {{?RFC5681}}.
+A sender that does not use pacing SHOULD reset its congestion window to the
+minimum of the current congestion window and the initial congestion window.
+This recommendation is based on Section 4.1 of {{?RFC5681}}.
 
 ## Pseudocode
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -1040,6 +1040,21 @@ As an example of a well-known and publicly available implementation of a flow
...
+If pacing is used, the connection should limit the initial burst of packets to
+no more than the initial congestion window and subsequent packets SHOULD be paced.
+The congestion window does not change while the connection is idle.
+
+If pacing is not used, the congestion window SHOULD be reset to the minimum of
+the current congestion window and the initial congestion window.  If the
+slow start threshold is larger than the congestion window, the congestion window
+will grow back to the congestion window prior to idle via slow start.
+
 ## Pseudocode
```
mikkelfj: How about slowly reducing the current window thorugh exponential decay down to initial window? Otherwise e.g. video might get quite chunky as buffers fill and the connection idles. But of course other connetions might have taken the bandwidth meanwhile, so it might not be a good idea after all, depending on decay speed.

mikkelfj: OK

draft-ietf-quic-recovery.md
```diff
@@ -1040,6 +1040,24 @@ As an example of a well-known and publicly available implementation of a flow
...
+
+If pacing is used, the connection should limit the initial burst of packets to
+no more than the initial congestion window and subsequent packets SHOULD be
+paced. The congestion window does not change while the connection is idle.
+
+If pacing is not used, the congestion window SHOULD be reset to the minimum of
+the current congestion window and the initial congestion window.  If the
+slow start threshold is larger than the congestion window, the congestion window
+will grow back to the congestion window prior to idle via slow start.  This
+recommendation is based on Section 4.1 of {{?RFC5681}}.
```
janaiyengar: Similarly, I would reduce this to a sentence and remove the sentence about slow start. I'm ok with leaving the ref to 5681 in here, but I would like to remove it in a later PR and replace it with more text explaining the rationale directly (instead of pointing to 5681). 

For now, how about "A sender that does not use pacing SHOULD reset its congestion window to the minimum of the current congestion window and the initial congestion window.  This recommendation is based on Section 4.1 of {{?RFC5681}}."

draft-ietf-quic-recovery.md
mikkelfj: FYI: I have not read this paper, except the abstract, but it seems to touch upon the current text (no my suggestion) related to pacing packets on idle restart https://www.isi.edu/~johnh/PAPERS/Visweswaraiah97b.pdf

mikkelfj: Again, only abstract read, ,but it speaks of decaying the cwnd on idle.
https://tools.ietf.org/html/rfc2861

ianswett: 5681 obsoletes 2581, but makes no mention of 2861.  I'll let @janaiyengar decide what to do here.

janaiyengar: RFC 2861 describes how to do this exponentially, but that was always considered too aggressive and resulted in folks turning it off in real deployments.  The new one is RFC 7661, which reduces the cwnd once to max(cwnd/2, IW) and ssthresh to max(ssthresh, 3/4 x cwnd).  We could use these reductions here, though 7661 is an experimental RFC, since it has gone through TCPM review.

draft-ietf-quic-recovery.md
```diff
@@ -1040,6 +1040,24 @@ As an example of a well-known and publicly available implementation of a flow
...
+
+A connection is idle if bytes in flight is 0 and there is nothing
+retransmittable to send.  This occurs when the connection is application
+limited and after a verified retransmission timeout.  In order to limit the
+size of bursts sent into the network, the behavior when restarting from idle
+depends upon whether pacing is used.
+
+If pacing is used, the connection should limit the initial burst of packets to
+no more than the initial congestion window and subsequent packets SHOULD be
+paced. The congestion window does not change while the connection is idle.
```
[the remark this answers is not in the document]
ianswett: If the goal is to reduce bursts, I think we need to advocate for limiting the burst of a paced connection to nothing larger than the restarting window of an unpaced connection.

PR #2024, opened 2018-11-20
Reset the handshake timer on Retry and VN

### LINKED ISSUE #1983: Loss recovery draft section 4.3.1.1 clarification
Section 4.3.1.1 of draft 16 says:
A Retry or Version Negotiation packet causes a client to send another
   Initial packet, effectively restarting the connection process.

Either packet indicates that the Initial was received but not
   processed.  Neither packet can be treated as an acknowledgment for
   the Initial, but they MAY be used to improve the RTT estimate.

Should we clear handshake_count when an Retry or Version Negotiation packet is received after Initial is sent (maybe multiple times)? Otherwise, the new Initial could be delayed extra if Initial is lost and needs retransmission.

ianswett:
Fixes #1983

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -375,8 +375,8 @@ packet, effectively restarting the connection process.
 
 Either packet indicates that the Initial was received but not processed.
 Neither packet can be treated as an acknowledgment for the Initial, but they MAY
-be used to improve the RTT estimate.  Whether or not the RTT estimate is
-updated, the handshake timer should be reset as though it has not yet fired.
+be used to improve the RTT estimate.  Initial packets that are sent in response
+to VN or Retry packets will reset the handshake timer.
 
 ### Tail Loss Probe {#tlp}
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -375,7 +375,8 @@ packet, effectively restarting the connection process.
 
 Either packet indicates that the Initial was received but not processed.
 Neither packet can be treated as an acknowledgment for the Initial, but they MAY
-be used to improve the RTT estimate.
+be used to improve the RTT estimate.  Whether or not the RTT estimate is
+updated, the handshake timer should be reset as though it has not yet fired.
 
 ### Tail Loss Probe {#tlp}
 


draft-ietf-quic-recovery.md
```diff
@@ -375,7 +375,8 @@ packet, effectively restarting the connection process.
 
 Either packet indicates that the Initial was received but not processed.
 Neither packet can be treated as an acknowledgment for the Initial, but they MAY
-be used to improve the RTT estimate.
+be used to improve the RTT estimate.  Whether or not the RTT estimate is
```
siyengar: hmm, wouldn't this automatically happen with the current. Receiving a VN or Retry will send a Handshake packet. Is this just editorial clarification?

ianswett: Yes, this is intended as editorial clarification to fix #1983

ianswett: Agreed, that wasn't what I intended to say, but that's how it came out.  I'll try to fix it, or if you want to make a suggestion, that'd be great.

siyengar: Oh my suggestion was "Initial packets that are sent in response to the VN or Retry packets will reset the handshake timer."

ianswett: Thanks, done

ianswett:
Good question, @marten-seemann I'm not sure it's necessary because of what you said, but there was a request for clarification.

Do you think we're better off keeping the text as-is and doing nothing?

mikkelfj:
@marten-seemann VN and Retry are essentially new connections, but they still participate in a single connection attempt which would have some state (e.g. Retry at most once) and an overall timeout. But perhaps it is simpler to assume defaults of new connections with the  few exceptions that apply mentioned explicitly.

marten-seemann:
@ianswett That's a good question. We say 

> Once the Version Negotiation packet is determined to be valid, the client then selects an acceptable protocol version from the list provided by the server. The client then attempts to create a connection using that version. Though the content of the Initial packet the client sends might not change in response to version negotiation, a client MUST increase the packet number it uses on every packet it sends.

I'm not sure why we changed this from initiating a completely new connection (including a new destination connection ID). I assume we want to ensure stable routing, if the server routes Initials based on the connection ID?

For Retry, we say 
> The client responds to a Retry packet with an Initial packet that includes the provided Retry Token to continue connection establishment.
> A client sets the Destination Connection ID field of this Initial packet to the value from the Source Connection ID in the Retry packet. Changing Destination Connection ID also results in a change to the keys used to protect the Initial packet. It also sets the Token field to the token provided in the Retry. The client MUST NOT change the Source Connection ID because the server could include the connection ID as part of its token validation logic

I can't find anything about resetting all connection state (except connection IDs and the packet number, and the RTT estimate) in the transport document. I think this is where this belongs, and this would include all state accumulated for recovery.

Issue #1968, opened 2018-11-06
Recovery draft unclear about when to stop sending an ACK for every packet when packets are missing


ianswett:
Stuart Cheshire asked when QUIC knows to stop sending immediate ACKs.  I don't believe the recovery draft is sufficiently clear on what the optimal behavior is here.

The draft indicates when to stop reporting old ACK ranges, but not when to stop sending immediate ACKs, so you could easily end up sending immediate ACKs for a full RTT.

PR #2025, opened 2018-11-20
Send an immediate ACK

ianswett:
Attempt to address #1968

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -228,7 +228,7 @@ an ACK immediately when receiving a second ack-eliciting packet.
 In order to accelerate loss recovery and reduce timeouts, the receiver SHOULD
 send an immediate ACK when it receives a new packet which is not one greater
 than the largest received packet number. A receiver MAY send immediate ACKs
-for the next few ack-eliciting packets that are received, but SHOULD not
+for the next few ack-eliciting packets that are received, but SHOULD NOT
 send an immediate ACK for more than 1/8 RTT after receiving an out-of-order
 packet.
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -228,8 +228,9 @@ an ACK immediately when receiving a second ack-eliciting packet.
 In order to accelerate loss recovery and reduce timeouts, the receiver SHOULD
 send an immediate ACK when it receives a new packet which is not one greater
 than the largest received packet number. A receiver MAY send immediate ACKs
-for the next few ack-eliciting packets that are received, but not for more
-than 1/8 RTT after receiving the out-of-order packet.
+for the next few ack-eliciting packets that are received, but SHOULD not
+send an immediate ACK for more than 1/8 RTT after receiving an out-of-order
+packet.
 
 Similarly, packets marked with the ECN Congestion Experienced (CE) codepoint in
 the IP header SHOULD be acknowledged immediately, to reduce the peer's response
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -222,9 +222,8 @@ the peer spurious timeouts.  The maximum ack delay is communicated in the
 `max_ack_delay` transport parameter and the default value is 25ms.
 
 An acknowledgement SHOULD be sent immediately upon receipt of a second
-packet but the delay SHOULD NOT exceed the maximum ack delay. QUIC recovery
-algorithms do not assume the peer generates an acknowledgement immediately when
-receiving a second packet.
+ack-eliciting packet. QUIC recovery algorithms do not assume the peer sends an
+an ACK immediately when receiving a second ack-eliciting packet.
 
 Out-of-order packets SHOULD be acknowledged more quickly, in order to accelerate
 loss recovery.  The receiver SHOULD send an immediate ACK when it receives a new
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -229,9 +229,8 @@ receiving a second packet.
 Out-of-order packets SHOULD be acknowledged more quickly, in order to accelerate
 loss recovery.  The receiver SHOULD send an immediate ACK when it receives a new
 packet which is not the next expected one. That is, its packet number is not one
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -483,11 +483,15 @@ the peer spurious timeouts.  The maximum ack delay is communicated in the
 An acknowledgement SHOULD be sent immediately upon receipt of a second
 packet but the delay SHOULD NOT exceed the maximum ack delay. QUIC recovery
 algorithms do not assume the peer generates an acknowledgement immediately when
```
mikkelfj: does immediately mean ASAP or any convenient time before ACK delay ?

draft-ietf-quic-recovery.md
```diff
@@ -483,11 +483,15 @@ the peer spurious timeouts.  The maximum ack delay is communicated in the
...
-receiving a second full-packet.
+receiving a second packet.
 
 Out-of-order packets SHOULD be acknowledged more quickly, in order to accelerate
 loss recovery.  The receiver SHOULD send an immediate ACK when it receives a new
-packet which is not one greater than the largest received packet number.
+packet which is not one greater than the largest received packet number.  The
```
janaiyengar: Suggestion: "The receiver SHOULD send an immediate ACK when it receives a new packet which is not the next expected one.  That is, its packet number is not one greater than the largest received packet number."

draft-ietf-quic-recovery.md
```diff
@@ -483,11 +483,15 @@ the peer spurious timeouts.  The maximum ack delay is communicated in the
...
+receiver MAY send an immediate ACK when a missing packet begins exceeding the
+peer's reordering threshold in packets to accelerate fast retransmit.  Time
```
janaiyengar: This is a problem, since the receiver does not know the peer's reordering threshold. Also, this sentence does not make sense "a missing packet begins exceeding ... threshold in packets"

draft-ietf-quic-recovery.md
```diff
@@ -483,11 +483,15 @@ the peer spurious timeouts.  The maximum ack delay is communicated in the
...
+threshold loss detection does not require any special ACK generation to
+accelerate fast retransmit, because it uses a timer.
```
janaiyengar: I think this text should be purely prescriptive, and only from the receiver's point of view, since the sender's is basically not known at the receiver. This also does not say when a receiver should stop immediate acking, which is important IMO.

Suggested replacement for text from "The receiver MAY" : "A receiver SHOULD immediately ack at least three subsequent packets after first receiving a packet out of order, after which it SHOULD resume delaying acknowledgements."

martinduke:
I think 'kPacketThreshold' is the ideal value of 'few' in my comment above. Maybe this should become an optional transport parameter?

draft-ietf-quic-recovery.md
```diff
@@ -224,11 +224,16 @@ the peer spurious timeouts.  The maximum ack delay is communicated in the
...
-receiving a second full-packet.
+receiving a second packet.
 
 Out-of-order packets SHOULD be acknowledged more quickly, in order to accelerate
 loss recovery.  The receiver SHOULD send an immediate ACK when it receives a new
-packet which is not one greater than the largest received packet number.
+packet which is not the next expected one. That is, its packet number is not one
+greater than the largest received packet number.  A receiver MAY immediately
+ack subsequent packets after first receiving a packet out of order. A receiver
+SHOULD NOT send an immediate ACK any time there is a gap in the ACK frame,
```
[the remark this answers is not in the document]
ianswett: Clarified.

martinduke:
@ianswett good point; I return to the position of my comment before last.

draft-ietf-quic-recovery.md
```diff
@@ -222,13 +222,17 @@ the peer spurious timeouts.  The maximum ack delay is communicated in the
...
 
 Out-of-order packets SHOULD be acknowledged more quickly, in order to accelerate
 loss recovery.  The receiver SHOULD send an immediate ACK when it receives a new
-packet which is not one greater than the largest received packet number.
+packet which is not the next expected one. That is, its packet number is not one
+greater than the largest received packet number.  A receiver MAY immediately
+acknowledge subsequent ack-eliciting packets after first receiving a packet out
+of order. A receiver SHOULD NOT send an immediate ACK any time there is a gap
+in the ACK frame being sent, because that will cause it to send an ACK in
+response to every packet for at least an RTT after a single packet loss.
```
janaiyengar: I don't understand what the text here is trying to say.  I'll suggest rewording for text "A receiver MAY ... single packet loss", see if this matches what you're getting at.

"A receiver MAY send acknowledgements immediately for the next few packets that are received, but the receiver SHOULD revert back to delaying acknowledgements after sending a few immediate acknowledgements."

draft-ietf-quic-recovery.md
```diff
@@ -222,13 +222,16 @@ the peer spurious timeouts.  The maximum ack delay is communicated in the
...
-packet but the delay SHOULD NOT exceed the maximum ack delay. QUIC recovery
-algorithms do not assume the peer generates an acknowledgement immediately when
-receiving a second full-packet.
+ack-eliciting packet. QUIC recovery algorithms do not assume the peer sends an
+an ACK immediately when receiving a second ack-eliciting packet.
 
 Out-of-order packets SHOULD be acknowledged more quickly, in order to accelerate
 loss recovery.  The receiver SHOULD send an immediate ACK when it receives a new
-packet which is not one greater than the largest received packet number.
+packet which is not the next expected one. That is, its packet number is not one
```
martinduke: Again, there are SHOULDs on top of SHOULDs here. How about this for a paragraph:

"Quicker acknowledgement of out-of-order packets may accelerate loss detection. Upon receipt of a packet number not one more than the previous maximum received, receivers SHOULD separately acknowledge it and the next few packets, but not more than five."

This is shorter, has only one SHOULD, and places a recommended limit that is less subjective than "a few".

ianswett: Thanks for the suggestions, I reworked the text to only have one SHOULD and mention the 1/8 RTT as the max amount of time to send immediate acks.

PR #2027, opened 2018-11-20
Move ack generation out of loss detection

janaiyengar:


janaiyengar:
Sure, happy to do it. I'm working on a couple of other PRs, I can redo #2025 after those.

PR #2028, opened 2018-11-20
Ignore loss of undecryptable packets

### LINKED ISSUE #1967: Loss events for reordered 0- and 0.5-RTT data
In the event that 0-or 0.5-RTT data arrives before the cryptographic handshake packets that would normally precede it.  That is, the Initial containing ClientHello might arrive after a few 0-RTT packets, or the 0.5-RTT packets might arrive before the server flight is completely received.

In that case, those packets will be declared lost.  That's fine from a loss-recovery perspective.  However, the congestion controller will treat this as a loss event.  We want to ensure that loss events are driven only by congestion.

Perhaps the congestion controller shouldn't trigger a loss event on packets that might be dropped in this way.  If we do this, we need to be careful about definition.

ianswett:
Fixes #1967

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -978,7 +978,6 @@ congestion window is less than ssthresh, which typically only occurs after an
 RTO. While in slow start, QUIC increases the congestion window by the number of
 bytes acknowledged when each acknowledgment is processed.
 
-
 ## Congestion Avoidance
 
 Slow start exits to congestion avoidance.  Congestion avoidance in NewReno
@@ -1000,6 +999,17 @@ The recovery period limits congestion window reduction to once per round trip.
 During recovery, the congestion window remains unchanged irrespective of new
 losses or increases in the ECN-CE counter.
 
+## Loss of protected packets during the handshake
+
+0RTT and 1RTT packets sent prior to handshake completion can arrive before
+the peer has keys to unprotect them.  In those cases, the peer may decide
+not to buffer the packets.  This will cause the packets to never be
+acknowledged and eventually declared lost, despite being delivered to
+the peer.  If the server rejects 0RTT, then the congestion controller
+SHOULD ignore the loss of 0RTT packets.  If any 0RTT or 1RTT packets sent
+prior to knowing the peer has keys to unprotect them are lost, the
+sender's congestion control MAY ignore the loss of those packets if it's
+believe they were received by the peer prior to having the correct keys.
 
 ## Tail Loss Probe
 


draft-ietf-quic-recovery.md
```diff
@@ -1000,6 +999,17 @@ The recovery period limits congestion window reduction to once per round trip.
 During recovery, the congestion window remains unchanged irrespective of new
 losses or increases in the ECN-CE counter.
 
+## Loss of protected packets during the handshake
+
+0RTT and 1RTT packets sent prior to handshake completion can arrive before
```
martinthomson: I think that you need to identify this as early RTT packets that are sent by an endpoint with new keys.  Also, expand on that: Handshake and 0-RTT from the client, and 1-RTT from either endpoint.

draft-ietf-quic-recovery.md
```diff
@@ -1000,6 +999,17 @@ The recovery period limits congestion window reduction to once per round trip.
 During recovery, the congestion window remains unchanged irrespective of new
 losses or increases in the ECN-CE counter.
 
+## Loss of protected packets during the handshake
```
martinthomson: Title Case Headings

```suggestion
## Reordering of Packets Sent with New Keys
```

draft-ietf-quic-recovery.md
```diff
@@ -1000,6 +999,17 @@ The recovery period limits congestion window reduction to once per round trip.
...
+
+0RTT and 1RTT packets sent prior to handshake completion can arrive before
+the peer has keys to unprotect them.  In those cases, the peer may decide
+not to buffer the packets.  This will cause the packets to never be
```
martinthomson: Use paragraphs:

```suggestion
not to buffer the packets.

Discarding packets for which there are no keys will result in the packets never being
```

draft-ietf-quic-recovery.md
```diff
@@ -1000,6 +999,17 @@ The recovery period limits congestion window reduction to once per round trip.
...
+acknowledged and eventually declared lost, despite being delivered to
+the peer.  If the server rejects 0RTT, then the congestion controller
```
martinthomson: Rejecting 0-RTT is a completely different concept that you should separate out.

draft-ietf-quic-recovery.md
```diff
@@ -1000,6 +999,17 @@ The recovery period limits congestion window reduction to once per round trip.
...
+SHOULD ignore the loss of 0RTT packets.  If any 0RTT or 1RTT packets sent
+prior to knowing the peer has keys to unprotect them are lost, the
+sender's congestion control MAY ignore the loss of those packets if it's
+believe they were received by the peer prior to having the correct keys.
```
```suggestion
believed they were received by the peer prior to having the correct keys.
```

Though I might instead say

>A congestion controller MAY choose to ignore loss signals from packets that might have been discarded due to packet protection keys being unavailable.

That's not complete, or a spec you could implement to, but it is less about belief and more about guesswork, which is what this is.

draft-ietf-quic-recovery.md
```diff
@@ -1000,6 +999,17 @@ The recovery period limits congestion window reduction to once per round trip.
...
 losses or increases in the ECN-CE counter.
 
+## Loss of protected packets during the handshake
+
+0RTT and 1RTT packets sent prior to handshake completion can arrive before
+the peer has keys to unprotect them.  In those cases, the peer may decide
+not to buffer the packets.  This will cause the packets to never be
+acknowledged and eventually declared lost, despite being delivered to
+the peer.  If the server rejects 0RTT, then the congestion controller
+SHOULD ignore the loss of 0RTT packets.  If any 0RTT or 1RTT packets sent
```
siyengar: should be consistent about 0-RTT, instead of using "0RTT", "0-RTT" might be better and similar for 1-RTT

draft-ietf-quic-recovery.md
```diff
@@ -1000,6 +999,17 @@ The recovery period limits congestion window reduction to once per round trip.
...
+prior to knowing the peer has keys to unprotect them are lost, the
+sender's congestion control MAY ignore the loss of those packets if it's
```
siyengar: this could use some psuedo-code changes as well. Happy to put them up in a separate PR if you want.

draft-ietf-quic-recovery.md
mikkelfj: What if an attacker deliberately sends random packets at high rate? It is not an unlikely attack. Could this shrink the cnwd to the point where valid communication becomes ineffective? If so, it is not a MAY but a MUST.

draft-ietf-quic-recovery.md
MikeBishop:
```suggestion
Handshake and 0-RTT packets sent by clients and 1-RTT packets sent by servers
prior to handshake completion can arrive before
```

draft-ietf-quic-recovery.md
MikeBishop:
```suggestion
the peer.  If the server rejects 0-RTT, then the congestion controller
```

draft-ietf-quic-recovery.md
MikeBishop:
```suggestion
SHOULD ignore the loss of 0-RTT packets.  If any 0-RTT or 1-RTT packets sent
```

draft-ietf-quic-recovery.md
MikeBishop: Traffic injection seems orthogonal to saying that there's a non-congestive form of loss that might get some grace during the handshake.  Injecting enough traffic to overwhelm the link and induce loss can happen at any point in the connection, because it's real loss from real congestion.

mikkelfj: But the question is if it can be done with far less resources here because it is not only taking bandwidth but also convincing and endpoint to back off. I don't know enough this to really know. It could also be the window is overestimated, which is not such a bad thing in this case.

janaiyengar:
I think this could be made a bit more specific:
- loss of Handshake packets and 0-RTT packets SHOULD NOT trigger a CC response.
- loss of 1-RTT packets triggers CC as normal, with one exception. A congestion controller MAY choose to ignore loss signals from packets that might have been discarded due to packet protection keys being unavailable.

martinthomson:
I like that formulation.  It recognizes the problem and recommends a simple solution (ignore for Handshake), but doesn't prohibit implementations from improving on that, or from choosing a simpler implementation that is slightly worse.  It's a corner case that doesn't need tons of spec machinery to support.

rpaulo:
I agree with Jana.  Implementations will likely buffer some packets during the handshake so the effect of ignoring cwnd reductions should be negligible.

martinthomson:
@ianswett, are we going to take #2045 instead of this, or do you want to try to fix this up?

martinthomson:
There are a few suggestions outstanding here.  @ianswett?

draft-ietf-quic-recovery.md
ianswett: Done in the other PR.

PR #2036, opened 2018-11-21
Rewrite Packet Size section

adventureloop:
Rewrite Packet Size section to make it clearer and to consider ICMP
handling and Datagram PLPMTUD.

- Add normative reference to DPLPMTUD
- Define MPS as the largest UDP Payload that can safely sent right now
- Improved desctription of IPv4 and IPv6 protocols.
- Discuss port numbers and connection ids as a method for verifying ICMP messages
- Separation of ICMP processing from the method: PMTUD or PLPMTUD
- Replace PLPMTUD considerations with DPLPMTUD considerations
- Specify that probes can made from Packets carrying PING and PADDING frames

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -60,7 +60,7 @@ normative:
         name: Sean Turner
         org: sn3rd
         role: editor
-  
+
   DPLPMTUD:
     title: "Datagram Packetization Layer Path MTU Discovery"
     date: {DATE}
@@ -2997,6 +2997,7 @@ the network path might be corrupting ECN codepoints, the endpoint MAY cease
 setting ECT codepoints in subsequent packets. Doing so allows the connection to
 traverse network elements that drop or corrupt ECN codepoints in the IP header.
 
+
 # Packet Size {#packet-size}
 
 The QUIC packet size includes the QUIC header and integrity check, but not the
@@ -3045,7 +3046,6 @@ path to a destination will support its desired message size without
 fragmentation.
 
 In the absence of these mechanisms, QUIC endpoints SHOULD NOT send IP packets
-<<<<<<< HEAD
 larger than 1280 bytes (assuming the minimum IP header size).  This results in
 a QUIC MPS of 1232 bytes for IPv6 and 1252 bytes for IPv4. A QUIC
 implementation MAY be more conservative in computing the QUIC MPS to allow for
@@ -5694,11 +5694,3 @@ Hamilton, Jana Iyengar, Fedor Kouranov, Charles Krasic, Jo Kulik, Adam Langley,
 Jim Roskind, Robbie Shade, Satyam Shekhar, Cherie Shi, Ian Swett, Raman Tenneti,
 Victor Vasiliev, Antonio Vicente, Patrik Westin, Alyssa Wilk, Dale Worley, Fan
 Yang, Dan Zhang, Daniel Ziegler.
-This new text attempts the following:
-- Fixes to wording to reflect the way things are described in other IETF 
-RFCs.
-- Clarity on how ICMP PTB messages are processed.
-- Improved desctription of IPv4 and IPv6 protocols.
-- Separation of ICMP processing from the method: PMTUD or PLPMTUD.
-- Alignment with what is expected by DPLPMTUD.
-
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3026,12 +3026,12 @@ The PMTU is the maximum size of the entire IP packet including the IP header,
 UDP header, and UDP payload.  The UDP payload includes the QUIC packet header,
 protected payload, and any authentication fields. This can depend upon the
 current path characteristics.  Therefore, the current largest UDP payload an
-implementation will send is referred to as QUIC Maximum Packet Size (MPS).
+implementation will send is referred to as QUIC maximum packet size.
 
 QUIC depends on a PMTU of at least 1280 bytes. This is the IPv6 minimum size
 {{!RFC8200}} and is also supported by most modern IPv4 networks.  All QUIC
-packets (except for PMTU probe packets) SHOULD be sized to fit within the MPS
-to avoid the packet being fragmented or dropped {{!RFC8805}.
+packets (except for PMTU probe packets) SHOULD be sized to fit within the 
+maximum packet size to avoid the packet being fragmented or dropped {{!RFC8805}.
 
 To optimize capacity efficiency, endpoints SHOULD use Datagram Packetization
 Layer PMTU Discovery ({{!DPLPMTUD=I-D.ietf-tsvwg-datagram-plpmtud}}), or
@@ -3041,17 +3041,18 @@ fragmentation.
 
 In the absence of these mechanisms, QUIC endpoints SHOULD NOT send IP packets
 larger than 1280 bytes. Assuming the minimum IP header size, this results in
-a QUIC MPS of 1232 bytes for IPv6 and 1252 bytes for IPv4. A QUIC
-implementation MAY be more conservative in computing the QUIC MPS to allow for
-unknown tunnel overheads or IP header options/extensions.
+a QUIC maximum packet size of 1232 bytes for IPv6 and 1252 bytes for IPv4. A QUIC
+implementation MAY be more conservative in computing the QUIC maximum packet 
+size to allow for unknown tunnel overheads or IP header options/extensions.
 
 Each pair of local and remote addresses could have a different PMTU.  QUIC
 implementations that implement any kind of PMTU discovery therefore SHOULD
-maintain a MPS for each combination of local and remote IP addresses.
+maintain a maximum packet size for each combination of local and remote IP 
+addresses.
 
 If a QUIC endpoint determines that the PMTU between any pair of local and
 remote IP addresses has fallen below the size needed to support the smallest
-allowed MPS, it MUST immediately cease sending QUIC packets on the
+allowed maximum packet size, it MUST immediately cease sending QUIC packets on the
 affected path.  This could result in termination of the connection if an
 alternative path cannot be found.
 
@@ -3099,8 +3100,8 @@ Further validation can also be provided:
 
 The endpoint SHOULD ignore all ICMP messages that are not validated or do not
 carry sufficient quoted packet payload to perform validation.  Any reduction in
-the QUIC MPS MAY be provisional until QUIC's loss detection algorithm
-determines that the quoted packet has actually been lost.
+the QUIC maximum packet size MAY be provisional until QUIC's loss detection 
+algorithm determines that the quoted packet has actually been lost.
 
 ## Considerations for Datagram Packetization Layer PMTU Discovery
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3030,7 +3030,7 @@ implementation will send is referred to as QUIC maximum packet size.
 
 QUIC depends on a PMTU of at least 1280 bytes. This is the IPv6 minimum size
 {{!RFC8200}} and is also supported by most modern IPv4 networks.  All QUIC
-packets (except for PMTU probe packets) SHOULD be sized to fit within the 
+packets (except for PMTU probe packets) SHOULD be sized to fit within the
 maximum packet size to avoid the packet being fragmented or dropped {{!RFC8805}.
 
 To optimize capacity efficiency, endpoints SHOULD use Datagram Packetization
@@ -3042,12 +3042,12 @@ fragmentation.
 In the absence of these mechanisms, QUIC endpoints SHOULD NOT send IP packets
 larger than 1280 bytes. Assuming the minimum IP header size, this results in
 a QUIC maximum packet size of 1232 bytes for IPv6 and 1252 bytes for IPv4. A QUIC
-implementation MAY be more conservative in computing the QUIC maximum packet 
+implementation MAY be more conservative in computing the QUIC maximum packet
 size to allow for unknown tunnel overheads or IP header options/extensions.
 
 Each pair of local and remote addresses could have a different PMTU.  QUIC
 implementations that implement any kind of PMTU discovery therefore SHOULD
-maintain a maximum packet size for each combination of local and remote IP 
+maintain a maximum packet size for each combination of local and remote IP
 addresses.
 
 If a QUIC endpoint determines that the PMTU between any pair of local and
@@ -3079,7 +3079,7 @@ packet as possible without the ICMPv6 packet exceeding 1280 bytes {{!RFC4443}}.
 The size of the quoted packet can actually be smaller, or the information
 unintelligible, for various reasons see Section 1.1 of {{!DPLPMTUD}}.
 
-When a randomized source port is used for a QUIC connection, this can provide 
+When a randomized source port is used for a QUIC connection, this can provide
 some protection from off path attacks that forge ICMP messages. The source port
 in a quoted packet can be checked for TCP {{!RFC6056}}
 and UDP transports {{!RFC8085}}, such as QUIC.  When used, a stack will only
@@ -3100,7 +3100,7 @@ Further validation can also be provided:
 
 The endpoint SHOULD ignore all ICMP messages that are not validated or do not
 carry sufficient quoted packet payload to perform validation.  Any reduction in
-the QUIC maximum packet size MAY be provisional until QUIC's loss detection 
+the QUIC maximum packet size MAY be provisional until QUIC's loss detection
 algorithm determines that the quoted packet has actually been lost.
 
 ## Considerations for Datagram Packetization Layer PMTU Discovery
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3026,107 @@ processed as valid.
...
-When implementing the algorithm in Section 7.2 of {{!PLPMTUD}}, the initial
-value of search_low SHOULD be consistent with the IPv6 minimum packet size.
-Paths that do not support this size cannot deliver Initial packets, and
-therefore are not QUIC-compliant.
-
-Section 7.3 of {{!PLPMTUD}} discusses trade-offs between small and large
-increases in the size of probe packets. As QUIC probe packets need not contain
-application data, aggressive increases in probe size carry fewer consequences.
-
+larger than 1280 bytes (assuming the minimum IP header size).  This results in
```
nibanks: I feel like this `(assuming the minimum IP header size)` should be moved to the end of the following sentence instead of here.

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3026,107 @@ processed as valid.
...
+allowed MPS, it MUST immediately cease sending QUIC packets on the
+affected path.  This could result in termination of the connection if an
+alternative path cannot be found.
+
+### Processing ICMP Messages to reduce the PMTU {#icmp-pmtud}
+
+PMTU discovery {{!RFC1191}} {{!RFC8201}} relies on reception of ICMP messages
+(e.g., IPv6 Packet Too Big, PTB, messages) that indicate when a packet is
+dropped because it is larger than the local router MTU. DPLPMTUD can also
+optionally utilise these messages.  This use of ICMP messages is
```
```suggestion
optionally utilize these messages.  This use of ICMP messages is
```

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3026,107 @@ processed as valid.
...
+potentially vulnerable to off-path attacks that successfully guess the IP
+address 3-tuple and reduce the PMTU to a bandwidth-inefficient value
+{{!RFC8201}}.
+
+QUIC endpoints SHOULD provide validation to protect from off-path injection of
+ICMP messages as specified in {{!RFC8201}} and Section 5.2 of {{!RFC8085}}.
+This uses the quoted packet supplied in the payload of an ICMP message, which,
+when present, can be used to associate the message with a corresponding
+transport connection {{!DPLPMTUD}}.
```
nibanks: Is it acceptable that a man-on-the-side could use received packets to spoof the ICMP messages?

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3026,107 @@ processed as valid.
...
+  packets outstanding, and can therefore be identified as spurious.
+
+* An endpoint could store additional information from the IP or UDP headers to
+  use for validation (for example, the IP ID or UDP checksum).
+
+The endpoint SHOULD ignore all ICMP messages that are not validated or do not
+carry sufficient quoted packet payload to perform validation.  Any reduction in
+the QUIC MPS MAY be provisional until QUIC's loss detection algorithm
+determines that the quoted packet has actually been lost.
+
```
nibanks: Since there should be enough quoted text to decode the packet number of the QUIC packet, you could save a hash/checksum of the first X bytes of the QUIC packet and validate that matches.

draft-ietf-quic-transport.md
dtikhonov: @nibanks, is there a consensus to utilize American English?

nibanks: No clue. I think in general it is used, but I have no idea if there is any consensus or requirement. It was just a suggestion. Feel free to ignore.

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3026,107 @@ processed as valid.
...
+current path characteristics.  Therefore, the current largest UDP payload an
+implementation will send is referred to as QUIC Maximum Packet Size (MPS).
 
-The Path Maximum Transmission Unit (PMTU) is the maximum size of the entire IP
-header, UDP header, and UDP payload. The UDP payload includes the QUIC packet
-header, protected payload, and any authentication fields.
+QUIC depends on a PMTU of at least 1280 bytes. This is the IPv6 minimum size
+{{!RFC8200}} and is also supported by most modern IPv4 networks.  All QUIC
+packets (except for PMTU probe packets) SHOULD be sized to fit within the MPS
+to avoid IP fragmentation or packet drop along the path {{!RFC8805}.
```
MikeBishop: "to avoid packet drop" feels like it's missing a word, or at least a letter.  "to avoid packet drops"?  Seems a little colloquial, but if it's accurate, maybe.  "to avoid the packet being fragmented or dropped", perhaps?

draft-ietf-quic-transport.md
MikeBishop: RFC7322:
> The RFC publication language is English.  Spelling may be either American or British, as long as an individual document is internally consistent.  Where both American and British English spelling are used within a document or cluster of documents, the text will be modified to be consistent with American English spelling.

draft-ietf-quic-transport.md
MikeBishop: That guards against an ICMP message that's wholly fabricated, but not against the man-on-the-side attack.  If the attacker legitimately saw a packet, the only way you detect the attack is when the peer acks the supposedly-too-big packet.

draft-ietf-quic-transport.md
MikeBishop: Unfortunately, the best security we have when trying to confirm that a message came from someone on path is that it saw the packet.  We should be wary of believing such unauthenticated messages blindly, however -- this might instead be a signal to re-do PMTUD, or as noted below, be a provisional signal until the packet is independently declared lost.

draft-ietf-quic-transport.md
```suggestion
larger than 1280 bytes. Assuming the minimum IP header size, this results in
```

draft-ietf-quic-transport.md
```suggestion
optionally use these messages.  This use of ICMP messages is
```

Don't overutilize "utilize".

FWIW, I think that for this document, we have chosen to use American English, even though that is the inferior variant.

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3020,108 @@ processed as valid.
...
+If a QUIC endpoint determines that the PMTU between any pair of local and
+remote IP addresses has fallen below the size needed to support the smallest
+allowed MPS, it MUST immediately cease sending QUIC packets on the
+affected path.  This could result in termination of the connection if an
+alternative path cannot be found.
+
+### Processing ICMP Messages to reduce the PMTU {#icmp-pmtud}
+
+PMTU discovery {{!RFC1191}} {{!RFC8201}} relies on reception of ICMP messages
+(e.g., IPv6 Packet Too Big, PTB, messages) that indicate when a packet is
```
```suggestion
(e.g., IPv6 Packet Too Big messages) that indicate when a packet is
```

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3020,108 @@ processed as valid.
...
+uses the quoted packet supplied in the payload of an ICMP message, which, when
+present, can be used to associate the message with a corresponding transport
+connection {{!DPLPMTUD}}.
+
+The IPv4 Router requirements {{!RFC1812} state that the quoted packet should
+contain as much of the original datagram as possible without the length of the
+ICMP datagram exceeding 576 bytes. IPv6 routers include as much of invoking
+packet as possible without the ICMPv6 packet exceeding 1280 bytes {{!RFC4443}}.
+The size of the quoted packet can actually be smaller, or the information
+unintelligible, for various reasons {{!DPLPMTUD}}.
```
martinthomson: A section reference would be useful here.

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3020,108 @@ processed as valid.
...
+
+When a randomized source port is used, this can provide some protection from
```
martinthomson: This could be read to imply that PMTUD probes could use a different source port.  That can have the effect of hitting a different path.  This probably needs to be qualified.

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3020,108 @@ processed as valid.
...
+determines that the quoted packet has actually been lost.
+
+## Considerations for Datagram Packetization Layer PMTU Discovery
+
+Section 6.4 of {{!DPLPMTUD}} provides considerations for
+implementing Datagram Packetization Layer PMTUD (DPLPMTUD) with QUIC.
+
+When implementing the algorithm in Section 5.3 of
+{{!DPLPMTUD}}, the initial value of BASE_PMTU SHOULD be
+consistent with the minimum QUIC packet size.
```
martinthomson: Saying what that is (1280, I assume, without reading DPLPMTUD) would be helpful.

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3020,108 @@ processed as valid.
...
+
+A PADDING frame can be used to generate PMTU probe packets. PADDING need not be
```
martinthomson: PADDING frames are not retransmitted if the packet they are contained in is lost.

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3020,108 @@ processed as valid.
...
+implementations that implement any kind of PMTU discovery therefore SHOULD
+maintain a MPS for each combination of local and remote IP addresses.
+
+If a QUIC endpoint determines that the PMTU between any pair of local and
+remote IP addresses has fallen below the size needed to support the smallest
+allowed MPS, it MUST immediately cease sending QUIC packets on the
+affected path.  This could result in termination of the connection if an
+alternative path cannot be found.
+
+### Processing ICMP Messages to reduce the PMTU {#icmp-pmtud}
```
```suggestion
### Processing ICMP Packet Too Big Messages {#icmp-pmtud}
```

or just

```suggestion
### ICMP Packet Too Big Messages {#icmp-pmtud}
```

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3026,107 @@ processed as valid.
...
 address of the client, see {{address-validation}}.
 
+## Path Maximum Transmission Unit (PMTU)
 
-## Path Maximum Transmission Unit
+The PMTU is the maximum size of the entire IP datagram including the IP header,
+UDP header, and UDP payload.  The UDP payload includes the QUIC packet header,
+protected payload, and any authentication fields. This can be depend upon the
+current path characteristics.  Therefore, the current largest UDP payload an
+implementation will send is referred to as QUIC Maximum Packet Size (MPS).
```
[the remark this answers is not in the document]
adventureloop: We use term MPS in DPLPMTUD, we have found that keeping the various packet size
values straight is quite hard. It is easier to reason about the sizes when
you seperate away the path size (PMTU), the estimate of the path size and the
size the transport can send (MPS).

I would not object to using 'maximum packet size' rather than an acronym.

draft-ietf-quic-transport.md
adventureloop: >  Assuming the minimum IP header size, this results in
> a QUIC MPS of 1232 bytes for IPv6 and 1252 bytes for IPv4. A QUIC
> implementation MAY be more conservative in computing the QUIC MPS to allow for
> unknown tunnel overheads or IP header options/extensions.

From above the QUIC minimum packet size would be 1252 or 1232. We avoided a number here to be consistent with the 'MAY be more conservative'

MikeBishop:
You need to trim trailing whitespace for the doc to build.

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3020,109 @@ processed as valid.
...
-IP addresses has fallen below 1280 bytes, it MUST immediately cease sending QUIC
-packets on the affected path.  This could result in termination of the
-connection if an alternative path cannot be found.
+QUIC endpoints SHOULD provide validation to protect from off-path injection of
+ICMP messages as specified in {{!RFC8201}} and Section 5.2 of {{!RFC8085}}. This
+uses the quoted packet supplied in the payload of an ICMP message, which, when
+present, can be used to associate the message with a corresponding transport
+connection {{!DPLPMTUD}}.
 
+The IPv4 Router requirements {{!RFC1812} state that the quoted packet should
```
martinduke: RFC 792 requires only 8 bytes of Transport header. While this quite old, it hasn't been formally obsoleted. I've always assumed endpoints needed to handle this case, although I don't have any data as to whether this still really happens in the internet. Hence the original text, which you've removed.

As a related data point, only about 4 years ago we saw ICMP messages where the MTU field was set to zero, which has been outdated for 25+ years. It's a wild world out there.

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3020,109 @@ processed as valid.
...
+implementing Datagram Packetization Layer PMTUD (DPLPMTUD) with QUIC.
 
-The PADDING frame provides a useful option for PMTU probe packets. PADDING
-frames generate acknowledgements, but they need not be delivered reliably. As a
-result, the loss of PADDING frames in probe packets does not require
-delay-inducing retransmission. However, PADDING frames do consume congestion
-window, which may delay the transmission of subsequent application data.
+When implementing the algorithm in Section 5.3 of
+{{!DPLPMTUD}}, the initial value of BASE_PMTU SHOULD be consistent with the
+minimum QUIC packet size (1232 bytes for IPv6 and 1252 bytes for IPv4).
```
martinduke: Isn't the BASE_PMTU simply 1280? Is that easier and clearer?

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3020,109 @@ processed as valid.
...
-value of search_low SHOULD be consistent with the IPv6 minimum packet size.
-Paths that do not support this size cannot deliver Initial packets, and
-therefore are not QUIC-compliant.
+A PADDING frame can be used to generate PMTU probe packets. PADDING need not be
+delivered reliably. As a result, the loss of PADDING frames in probe packets
+does not require delay-inducing retransmission. However, PADDING frames do
+consume congestion window, which could delay the transmission of subsequent
+application data.
 
-Section 7.3 of {{!PLPMTUD}} discusses trade-offs between small and large
```
martinduke: This text was meant to indicate a point in Section 7.3 that does not apply to QUIC.

draft-ietf-quic-transport.md
```diff
@@ -3022,85 +3020,109 @@ processed as valid.
...
-## Special Considerations for Packetization Layer PMTU Discovery
+Section 6.4 of {{!DPLPMTUD}} provides considerations for
+implementing Datagram Packetization Layer PMTUD (DPLPMTUD) with QUIC.
 
-The PADDING frame provides a useful option for PMTU probe packets. PADDING
-frames generate acknowledgements, but they need not be delivered reliably. As a
-result, the loss of PADDING frames in probe packets does not require
-delay-inducing retransmission. However, PADDING frames do consume congestion
-window, which may delay the transmission of subsequent application data.
+When implementing the algorithm in Section 5.3 of
```
martinduke: It appears that the rest of this section simply repeats section 6.4 of the DPLPMTUD draft, which seems unnecessary, except the bit about ICMP processing.

draft-ietf-quic-transport.md
```diff
@@ -2998,17 +2998,15 @@ The QUIC packet size includes the QUIC header and integrity check, but not the
 UDP or IP header.
 
 Clients MUST ensure that the first Initial packet they send is sent in a UDP
```
ianswett: nit: how about:
"Clients MUST ensure they send the first Initial packet in a UDP datagram that is at least 1200 bytes."

draft-ietf-quic-transport.md
```diff
@@ -3023,83 +3021,108 @@ The server MUST also limit the number of bytes it sends before validating the
 address of the client, see {{address-validation}}.
 
 
-## Path Maximum Transmission Unit
+## Path Maximum Transmission Unit (PMTU)
+
+The PMTU is the maximum size of the entire IP packet including the IP header,
+UDP header, and UDP payload.  The UDP payload includes the QUIC packet header,
+protected payload, and any authentication fields. This can depend upon the
```
```suggestion
protected payload, and any authentication fields. The PMTU can depend upon the
```

draft-ietf-quic-transport.md
```diff
@@ -3023,83 +3021,108 @@ The server MUST also limit the number of bytes it sends before validating the
...
+current path characteristics.  Therefore, the current largest UDP payload an
+implementation will send is referred to as QUIC maximum packet size.
```
```suggestion
implementation will send is referred to as the QUIC maximum packet size.
```

draft-ietf-quic-transport.md
```diff
@@ -3023,83 +3021,108 @@ The server MUST also limit the number of bytes it sends before validating the
...
+maintain a maximum packet size for each combination of local and remote IP
+addresses.
 
 If a QUIC endpoint determines that the PMTU between any pair of local and remote
-IP addresses has fallen below 1280 bytes, it MUST immediately cease sending QUIC
-packets on the affected path.  This could result in termination of the
-connection if an alternative path cannot be found.
+IP addresses has fallen below the size needed to support the smallest allowed
+maximum packet size, it MUST immediately cease sending QUIC packets on the
+affected path.  This could result in termination of the connection if an
```
ianswett: This recommendation results in a timeout in cases where it may be possible to immediately close the connection if no other paths are available. I think it'd be preferable to close the connection, or say it MAY close the connection?

martinthomson:
Thanks for this!  It's a large change, but well-written and a big improvement.

PR #2039, opened 2018-11-21
Connection state is reset on Retry or VN

janaiyengar:
Alternate fix to #1983.  Note that this is a PR against #2024.


janaiyengar:
Ugh. I'm making this more complicated that it needs to be. I'm closing this PR and opening one against master.

PR #2040, opened 2018-11-21
Reset connection state on retry and VN

### LINKED ISSUE #1983: Loss recovery draft section 4.3.1.1 clarification

janaiyengar:
This is an alternate patch to #2024.

Closes #1983.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -433,11 +433,11 @@ When crypto packets are outstanding, the TLP and RTO timers are not active.
 #### Retry and Version Negotiation
 
 A Retry or Version Negotiation packet causes a client to send another Initial
-packet, effectively restarting the connection process.
-
-Either packet indicates that the Initial was received but not processed.
-Neither packet can be treated as an acknowledgment for the Initial, but they MAY
-be used to improve the RTT estimate.
+packet, effectively restarting the connection process and resetting connection
+state, including resetting any pending timers.  Either packet indicates that the
+Initial was received but not processed.  Neither packet can be treated as an
+acknowledgment for the Initial, but they MAY be used to improve the RTT
+estimate.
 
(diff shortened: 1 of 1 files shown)


mikkelfj:
I like this, but since the state is not totally reset in case of Retry since you can Retry at most once, perhaps this should be clarified?

draft-ietf-quic-recovery.md
```diff
@@ -433,11 +433,11 @@ When crypto packets are outstanding, the TLP and RTO timers are not active.
...
-packet, effectively restarting the connection process.
-
-Either packet indicates that the Initial was received but not processed.
-Neither packet can be treated as an acknowledgment for the Initial, but they MAY
-be used to improve the RTT estimate.
+packet, effectively restarting the connection process and resetting connection
+state, including resetting any pending timers.  Either packet indicates that the
+Initial was received but not processed.  Neither packet can be treated as an
+acknowledgment for the Initial, but they MAY be used to improve the RTT
+estimate.
```
martinthomson: Here we're into dangerous waters again.  Resetting most of the state is fine, but you can't forget the connection ID and token give to you in Token, or the list of versions from the Version Negotiation.  Here, only RTT is mentioned.  If this were scoped to recovery and congestion state, that might be better than having to reiterate what is carried over.

PR #2042, opened 2018-11-22
QPACK: update language to use HTTP/3

LPardue:


mikkelfj:
Wouldn't HPACK/3, as opposed to HPACK3, be the more correct term?

PR #2043, opened 2018-11-22
transport: fix broken references to -recovery

reschke:


MikeBishop:
See https://github.com/cabo/kramdown-rfc2629/issues/57 for exactly that reason.  I'd like to be able to use the named anchors across documents and just pick up the correct sections as we go.  We'll definitely have to do a full pass of all cross-document links before publication, to be certain.

PR #2044, opened 2018-11-22
It's just TLS

martinthomson:
For the most part, we don't need to identify TLS 1.3 specifically.  TLS
on its own is often sufficient.  This avoids creating the impression
that we don't support TLS 1.4.  I hope that TLS 1.4 will be compatible
with QUIC, but we'll leave that determination for later.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -112,16 +112,16 @@ code and issues list for this draft can be found at
 
 # Introduction
 
-This document describes how QUIC {{QUIC-TRANSPORT}} is secured using TLS version
-1.3 {{!TLS13=RFC8446}}.  TLS 1.3 provides critical
-latency improvements for connection establishment over previous versions.
-Absent packet loss, most new connections can be established and secured within a
-single round trip; on subsequent connections between the same client and server,
-the client can often send application data immediately, that is, using a zero
-round trip setup.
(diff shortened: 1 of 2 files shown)


mikkelfj:
That is not very precise:

You have document that describes how to use with 1.3, implying one could use any other version. Presumably this happens by detecting the TLS version in the handshake so the same QUIC version could support multiple TLS versions. It could also mean different QUIC versions support different TLS versions. Further, it could mean that TLS 1.0 is supported, but not specified. This opens to downgrade attacks such as loss of perfect forward secrecy. It also makes it hard to decipher the spec.

I think it is fine to more relaxed about TLS as a building block, especially with an eye towards the feature, and perhaps also the past since TLS 1.2 might be more IoT friendly. However, a specific QUIC version needs to define a specific TLS version or set of TLS versions that are all mandatory.

mikkelfj:
To be fair the text does so TLS means 1.3 or newer, but I still think it is hard to ensure compatibility with random future TLS versions, 20 years out. There might not even by a HKDF function and header protection might not make sense in terms of available TLS algorithms.

martinthomson:
@mikkelfj, please read https://quicwg.org/base-drafts/draft-ietf-quic-tls.html#tls-version  We're not committing to support of arbitrary versions.  TLS <=1.2 is explicitly forbidden; TLS >1.3 only works if it meets the API requirements in the TLS doc.

PR #2045, opened 2018-11-23
Discard Initial keys as soon as possible

### LINKED ISSUE #1951: Disconnect with Initial Injection
Suppose that a connection has progressed and the server has sent an Initial packet with a Server Hello to the client. At that point, we expect that all following packets will be protected by the handshake key or by the application key, preventing 3rd parties from interfering with the connection. 

Suppose now that a third party sends to the server a spoofed initial packet, containing for example a malformed acknowledgement. The server will process the packet, find the malformed frame, generate a protocol violation error and close the connection. Which means that the 3rd party is able to interfere with the connection after the handshake keys have already been exchanged.

Is that an attack we should worry about?

martinthomson:
Closes #1951.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -44,24 +44,6 @@ normative:
         org: Mozilla
         role: editor
 
-  QUIC-TLS:
-    title: "Using Transport Layer Security (TLS) to Secure QUIC"
-    date: {DATE}
-    seriesinfo:
-      Internet-Draft: draft-ietf-quic-tls-latest
-    author:
-      -
-        ins: M. Thomson
-        name: Martin Thomson
-        org: Mozilla
-        role: editor
-      -
-        ins: S. Turner
-        name: Sean Turner
-        org: sn3rd
-        role: editor
-
-
 informative:
 
 
@@ -458,11 +440,12 @@ be used to improve the RTT estimate.
 
 #### Discarding Initial State {#discard-initial}
 
-As described in Section 4.10 of {{QUIC-TLS}}, endpoints stop sending and
+As described in Section 17.5.1 of {{QUIC-TRANSPORT}}, endpoints stop sending and
 receiving Initial packets once they start exchanging Handshake packets.  At this
 point, all loss recovery state for the Initial packet number space is also
-discarded.  After discarding state, new Initial packets will not be sent and no
-additional packets will be declared lost.
+discarded. Packets that are in flight for the packet number space are not
+declared as either acknowledged or lost.  After discarding state, new Initial
+packets will not be sent.
 
 ### Tail Loss Probe {#tlp}
 
@@ -1084,14 +1067,14 @@ A sender that does not use pacing SHOULD reset its congestion window to the
 minimum of the current congestion window and the initial congestion window.
 This recommendation is based on Section 4.1 of {{?RFC5681}}.
 
-## In-Flight Packet Accounting
+## Discarding Packet Number Space State
 
-When keys for an packet number space are discarded (see {{QUIC-TLS}}), any
-packets sent with those keys are removed from the count of bytes in flight.  No
-loss events will occur for these packets, as a result of discarding loss
-recovery state (see {{discard-initial}}).  Note that it is expected that keys
-are discarded after those packets would be declared lost, but Initial secrets
-are destroyed earlier.
+When keys for an packet number space are discarded, any packets sent with those
+keys are removed from the count of bytes in flight.  No loss events will occur
+any in-flight packets from that space, as a result of discarding loss recovery
+state (see {{discard-initial}}).  Note that it is expected that keys are
+discarded after those packets would be declared lost, but Initial secrets are
+destroyed earlier.
 
 ## Pseudocode
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -44,6 +44,23 @@ normative:
         org: Mozilla
         role: editor
 
+  QUIC-TLS:
+    title: "Using Transport Layer Security (TLS) to Secure QUIC"
+    date: {DATE}
+    seriesinfo:
+      Internet-Draft: draft-ietf-quic-tls-latest
+    author:
+      -
+        ins: M. Thomson
+        name: Martin Thomson
+        org: Mozilla
+        role: editor
+      -
+        ins: S. Turner
+        name: Sean Turner
+        org: sn3rd
+        role: editor
+
 
 informative:
 
@@ -438,6 +455,13 @@ packet, effectively restarting the connection process.
 Either packet indicates that the Initial was received but not processed.
 Neither packet can be treated as an acknowledgment for the Initial, but they MAY
 be used to improve the RTT estimate.
+
+#### Discarding Initial State
+
+As described in Section 4.10 of {{QUIC-TLS}}, endpoints stop sending and
+receiving Initial packets once they start exchanging Handshake packets.  At this
+point, all loss recovery state for the Initial encryption level is also
+discarded.
 
 ### Tail Loss Probe {#tlp}
 
@@ -1058,6 +1082,14 @@ is idle.
 A sender that does not use pacing SHOULD reset its congestion window to the
 minimum of the current congestion window and the initial congestion window.
 This recommendation is based on Section 4.1 of {{?RFC5681}}.
+
+## In-Flight Packet Accounting
+
+When keys for an encryption level are discarded (see {{QUIC-TLS}}), any packets
+sent with those keys are removed from the count of bytes in flight.  No loss
+events will occur for these packets.  Note that it is expected that keys are
+discarded after those packets would be declared lost, but Initial secrets are
+destroyed earlier.
 
 ## Pseudocode
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -256,7 +256,7 @@ delayed acknowledgement should be generated after processing incoming packets.
 In order to quickly complete the handshake and avoid spurious retransmissions
 due to crypto retransmission timeouts, crypto packets SHOULD use a very short
 ack delay, such as 1ms.  ACK frames MAY be sent immediately when the crypto
-stack indicates all data for that encryption level has been received.
+stack indicates all data for that packet number space has been received.
 
 ## ACK Ranges
 
@@ -456,12 +456,13 @@ Either packet indicates that the Initial was received but not processed.
 Neither packet can be treated as an acknowledgment for the Initial, but they MAY
 be used to improve the RTT estimate.
 
-#### Discarding Initial State
+#### Discarding Initial State {#discard-initial}
 
 As described in Section 4.10 of {{QUIC-TLS}}, endpoints stop sending and
 receiving Initial packets once they start exchanging Handshake packets.  At this
-point, all loss recovery state for the Initial encryption level is also
-discarded.
+point, all loss recovery state for the Initial packet number space is also
+discarded.  After discarding state, new Initial packets will not be sent and no
+additional packets will be declared lost.
 
 ### Tail Loss Probe {#tlp}
 
@@ -1085,11 +1086,12 @@ This recommendation is based on Section 4.1 of {{?RFC5681}}.
 
 ## In-Flight Packet Accounting
 
-When keys for an encryption level are discarded (see {{QUIC-TLS}}), any packets
-sent with those keys are removed from the count of bytes in flight.  No loss
-events will occur for these packets.  Note that it is expected that keys are
-discarded after those packets would be declared lost, but Initial secrets are
-destroyed earlier.
+When keys for an packet number space are discarded (see {{QUIC-TLS}}), any
+packets sent with those keys are removed from the count of bytes in flight.  No
+loss events will occur for these packets, as a result of discarding loss
+recovery state (see {{discard-initial}}).  Note that it is expected that keys
+are discarded after those packets would be declared lost, but Initial secrets
+are destroyed earlier.
 
 ## Pseudocode
 
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -641,7 +641,8 @@ alerts at the "warning" level.
 
 After QUIC moves to a new encryption level, packet protection keys for previous
 encryption levels can be discarded.  This occurs several times during the
-handshake, as well as when keys are updated (see {{key-update}}).
+handshake, as well as when keys are updated (see {{key-update}}).  Initial
+packet protection keys are treated specially, see {{discard-initial}}.
 
 Packet protection keys are not discarded immediately when new keys are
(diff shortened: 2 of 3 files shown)


draft-ietf-quic-tls.md
```diff
@@ -691,6 +692,24 @@ will be marked as lost before this, as they leave a gap in the sequence of
...
+authenticated, meaning that an attacker could spoof packets with the intent to
+disrupt a connection.  To limit these attacks, Initial packet protection keys
+can be discarded more aggressively than other keys.
+
+The successful use of Handshake packets indicates that no more Initial packets
+need to be exchanged, as these keys can only be produced after receiving all
+CRYPTO frames from Initial packets.  Thus, a client MUST discard Initial keys
+when it first sends a Handshake packet and a server MUST discard Initial keys
+when it first successfully processes a Handshake packet.  Endpoints MUST NOT
+send Initial packets after this point.
```
marten-seemann: Should we say that this means that the packet triggering this will never be acknowledged?

draft-ietf-quic-recovery.md
```diff
@@ -439,6 +456,13 @@ Either packet indicates that the Initial was received but not processed.
 Neither packet can be treated as an acknowledgment for the Initial, but they MAY
 be used to improve the RTT estimate.
 
+#### Discarding Initial State
+
+As described in Section 4.10 of {{QUIC-TLS}}, endpoints stop sending and
+receiving Initial packets once they start exchanging Handshake packets.  At this
+point, all loss recovery state for the Initial encryption level is also
+discarded.
```
ekr: I'm not sure I understand what this mean; it seems like there is one set of variables. Do you mean you reset all the variables to the initial values.

draft-ietf-quic-recovery.md
```diff
@@ -1059,6 +1083,14 @@ A sender that does not use pacing SHOULD reset its congestion window to the
 minimum of the current congestion window and the initial congestion window.
 This recommendation is based on Section 4.1 of {{?RFC5681}}.
 
+## In-Flight Packet Accounting
+
+When keys for an encryption level are discarded (see {{QUIC-TLS}}), any packets
+sent with those keys are removed from the count of bytes in flight.  No loss
+events will occur for these packets.  Note that it is expected that keys are
+discarded after those packets would be declared lost, but Initial secrets are
+destroyed earlier.
```
ekr: I'm not sure I understand this text.

draft-ietf-quic-tls.md
ekr: I thought the sense of the discussion in BKK was *not* to do this. It's a weird kind of implicit ACKIng. This probably needs list and issue discussion, not just merging at this time.

ekr: Yes, and it's not clear to me it's worth making *this* change to defend against this attack. As I said, I don't recall there being consensus to make this change.

rpaulo: I agree with @marten-seemann.  What this means is that implementations won't even bother sending any ACKs.  Perhaps the recommendation should be to discard the keys once the ACK arrives.

marten-seemann: @rpaulo To clarify, I'm saying we should not send an ACK, and we should discard the keys as soon as possible.
Delaying to discard the keys until the ACK arrives is exactly what makes the attack possible in the first place: If we rely on ACKs, there's the problem that the ACK might get lost. Then the peer has to retransmit the last Initial packet, just to trigger the ACK (although it already knows that the packet was received, because new TLS keys became available). In the end, you'll end up generating and accepting Initial packets for **a lot** longer than the 3-way handshake in which they are actually needed, which is a big window for an attacker to inject a packet that will kill the connection.

rpaulo: Okay, I see what you mean.  I think that from a transport protocol perspective, not sending an ACK isn't good since you just ignored a possible RTT estimation.  You could estimate the RTT from the new handshake packet, I suppose, but that's not how the transport part works.

I'm not sure I understand the attack since to spoof the initial packet,  the third party must be doing a man in the middle attack.  In that case, why not do it early and abuse the stateless reset or retry?

mikkelfj: It is not a man in the middle attack, it is a man on the side attack. The initial packet can be decrypted and a response can be constructed, but at the next encryption level that option is gone. This is why the initial is special. 

If it is possible to forge a reset from the initial packet, something is wrong in the design. It is not possible when a new connection id is produced, but I'm not sure if there the CID is fixed zero length. If there is an attack here, the reset token ought to moved further into the handshake.

kazuho:
@ekr 
> I thought the sense of the discussion in BKK was _not_ to do this. It's a weird kind of implicit ACKIng.

I share the view that it is a variant of implicit ACKing, something that we decided to not adopt in Kista. OTOH, it _is_ an improvement in terms of preventing man-on-the-side attacks that try to disrupt connection establishment.

IMO, the proposed approach is trivial to implement for everybody. In addition to that, it simplifies QUIC stacks that implement protection against Initial injection, because they are no longer required to continue responding to Initial packets (that might contain rogue information).

Considering the aspects, as stated in https://github.com/quicwg/base-drafts/issues/1951#issuecomment-438112817, I am now in favor of adopting this.

draft-ietf-quic-tls.md
huitema: I agree with the spirit of the text: we must defend against this attack, and the proper way has to be some form of implicit ACK. But I would like to see the implicit acknowledgement mechanism specified in the transport draft, because it affects the transport state machine. Obviously, bad things would happen if one side decided to stop send sending Initial packets while the other side was still waiting for ACK of its own packets, and repeating them. That means we must be crystal clear on when exactly to perform this "initial cut-off": on the client side, when the Handshake secret is successfully received and the client starts being able receiving handshake packets; on the server side, when the first handshake packet is received from the server.

huitema: I also looked at how to implement a similar protection without this kind of cut-off, and it just does not work. Suppose for example that the client receives a repeated Initial packet after the cut-off. In the post cutoff defensive mode, it can decide to ignore the packet content, since it cannot possibly learn anything interesting, but it would still need to send an ACK, otherwise the peer will keep repeating 3 or 4 times and then give up on the connection. But then, if the packet from the "server" was forged, the ACK from the client will be treated as a protocol violation, and the server will close the connection. Catch 22, no good way to get out of that.

huitema:
@ianswett the main problem with "no ACK" is that we will need special handling of retransmissions on the server. The server would need to trigger retransmission of Initial packets upon reception of a duplicate CH, and that's clearly some new ad hoc code. 

Also, what if the server does not like any of the key shares proposed in the Client Hello? Client sends CH1, server replies with HRR, client sends a modified CH2, server sends SH. If the client repeats CH1, the server should repeat HRR; if the client repeats CH2, the server needs to repeat the CH. I assume this could work if the client always coalesced an ACK with CH2, but this coalescing is optional, not mandated.

huitema:
@ianswett Timer-based repetition of initial packets is an amplification: one packet in, N packets repeated until the server gives up. So it is complicated. Let's not try to specify ACK logic in this specific PR. It really belongs in its own PR, and as you note there will be many details to get right.

ianswett:
@huitema I was trying to clarify that timer based repetition of the crypto data, including Initial, is already present in the recovery draft.

draft-ietf-quic-transport.md
```diff
@@ -3592,6 +3594,21 @@ and will contain a CRYPTO frame with an offset matching the size of the CRYPTO
...
+
+### Abandoning Initial Packets {#discard-initial}
+
+A client stops both sending and accepting Initial packets when it sends its
+first Handshake packet.  A server stops sending and accepting Initial packets
+when it receives its first Handshake packet.  Though packets might still be in
+flight or awaiting acknowledgment, no further Initial packets need to be
+exchanged beyond this point.  Initial packet protection keys are discarded (see
+Section 4.10 of {{QUIC-TLS}}) along with any loss recovery and congestion
+control state (see Sections 5.3.1.2 and 6.9 of {{QUIC-RECOVERY}}).
```
MikeBishop: Not my area, but I'm not sure implies the right thing to say that any congestion control state gets discarded here.  The congestion controller doesn't care what encryption level things are at, it just cares about bytes.

Perhaps it would be worth being more concrete:  "discarded(...), and all Initial packets are considered acknowledged for loss recovery and congestion control purposes (see...)."  This would also make the following paragraph unnecessary.

janaiyengar: +1. @MikeBishop 's proposed text also explicitly articulates that this is an implicit acknowledgement, something that we've been shy to do before.

martinthomson: I don't think that it is the case.  We don't really say anything from a loss recovery perspective - the packets are not declared lost.  The congestion controller is just told to remove the bytes from in flight - they don't cause the CWND to increase.  So it's not really acknowledgment.

As the changes to the recovery draft state: loss recovery state is just thrown out: you won't be retransmitting anything at the encryption level.  So saying anything more isn't really helpful there.  The effect on the congestion controller is intentionally neutral: don't increase CWND as though the packets got through, but don't keep the bytes on the books as though the packets were lost.

Or have I screwed this up?

MikeBishop: Okay, I see that reading of the -recovery text.  But that's really weird -- we *do* know that these packets have arrived, so why define a special operation for "forgetting" packets rather than just marking them arrived as normal and be done?

If one component were trying to game the other -- for example, by sending many Initial packets with redundant data -- then you wouldn't want to let the packetizer force many packets to be "acknowledged" when only one actually needed to get through to produce the effect.  But the ways to game that I see offhand are intra-implementation, and I think we can say that's not an attack of concern.

janaiyengar: Good points. I'd say the conservative approach is the one @martinthomson suggests. I wonder if it might be better to be more explicit about it though, but I'll save that for another PR I think.

martinthomson:
Just a note of warning.  @ianswett is going to review this within 24 hours and I expect to merge it shortly afterwards.

ekr:
I continue to be unconvinced this is the right answer, and I'd prefer you not merge it. If you elect to do so in any case, I am going to want to talk about it in Tokyo.

ianswett:
@ekr Clarification question: Do you think this is an issue we should just not address in QUIC v1 or there are other solutions not in this PR?

ekr:
I'm not sure we need to address it in QUICv1 at all. Rather than piecemeal
fixes, I'd like to start with a clear agreement about what the proposed
threat model is for man on the side attacks. Once we've got consensus on
that, we can then fix all the attacks that are implicated by that threat
model.

draft-ietf-quic-transport.md
```diff
@@ -3592,6 +3594,21 @@ and will contain a CRYPTO frame with an offset matching the size of the CRYPTO
 frame sent in the first Initial packet.  Cryptographic handshake messages
 subsequent to the first do not need to fit within a single UDP datagram.
 
+
+### Abandoning Initial Packets {#discard-initial}
+
+A client stops both sending and accepting Initial packets when it sends its
```
ianswett: This draft uses packet processing, and accepting for things like "accepting" new connections, so I think processing is better here and on 3601

draft-ietf-quic-recovery.md
```diff
@@ -1059,6 +1084,15 @@ A sender that does not use pacing SHOULD reset its congestion window to the
 minimum of the current congestion window and the initial congestion window.
 This recommendation is based on Section 4.1 of {{?RFC5681}}.
 
+## In-Flight Packet Accounting
+
+When keys for an packet number space are discarded (see {{QUIC-TLS}}), any
+packets sent with those keys are removed from the count of bytes in flight.  No
```
ianswett: clarification: any in-flight packets sent with those keys

draft-ietf-quic-recovery.md
```diff
@@ -439,6 +456,14 @@ Either packet indicates that the Initial was received but not processed.
 Neither packet can be treated as an acknowledgment for the Initial, but they MAY
 be used to improve the RTT estimate.
 
+#### Discarding Initial State {#discard-initial}
+
+As described in Section 4.10 of {{QUIC-TLS}}, endpoints stop sending and
```
ianswett: I think it makes more sense to reference the transport draft, now that it has the required text.  Also, it's a bit odd to have the recovery draft have a direct dependency on TLS.

Most of this paragraph seems to apply to the Handshake encryption level as well, correct?

draft-ietf-quic-recovery.md
```diff
@@ -1059,6 +1084,15 @@ A sender that does not use pacing SHOULD reset its congestion window to the
 minimum of the current congestion window and the initial congestion window.
 This recommendation is based on Section 4.1 of {{?RFC5681}}.
 
+## In-Flight Packet Accounting
```
ianswett: I would call this section something else like "Discarding Keys" or "Abandoning Encryption Levels"

draft-ietf-quic-recovery.md
```diff
@@ -439,6 +456,14 @@ Either packet indicates that the Initial was received but not processed.
 Neither packet can be treated as an acknowledgment for the Initial, but they MAY
 be used to improve the RTT estimate.
 
+#### Discarding Initial State {#discard-initial}
+
+As described in Section 4.10 of {{QUIC-TLS}}, endpoints stop sending and
+receiving Initial packets once they start exchanging Handshake packets.  At this
+point, all loss recovery state for the Initial packet number space is also
+discarded.  After discarding state, new Initial packets will not be sent and no
```
```suggestion
discarded without declaring them acknowledged or lost.  After discarding state, new Initial packets will not be sent.
```

ianswett:
One thought.  There are two parts of this PR.
 1) What to do when you drop keys, which the spec currently requires in some cases and others may want to do(implicit acking).
 2) Add the requirement that all implementations support implicit acking.

We really need the first part, even if we discuss the second part further.

martinthomson:
I've been avoiding using "implicit ACK" to describe this, because it's at a different layer.  ACKs deal with packets, but this is really looking at the packet number space.

The parts of this that relate to what to do with those dangling packets from an old packet number space are only really necessary here because we are moving the point where state is discarded earlier.  For other packet number spaces, they are long quiescent by the time any discarding happens, so the advice we'd provide would be largely academic.

huitema:
I am looking at the arguments on this thread, and I think EKR and Kazuho are making very similar points:

1) The current state is that packet injection attacks are possible until 3 RTT after handshake is complete, or about 4 RTT from the time the Client Hello is sent.

2) This PR (#2045) would reduce the window from 4 RTT to 1 RTT.

3) But it is reasonably easy for attackers to inject packets within 0.5 RTT from the point the Client Hello is sent.

4) Thus we get a piecemeal defense that adds some complexity and risk but does not actually solve the problem.

I feel sad.

martinthomson:
Yep, that seems like a reasonable summation.  But I hold hopes for the out-of-band configuration that you and Kazuho are working on.  I don't think that it is worth holding v1 for though.  Especially since it imposes some deployment constraints that could be onerous.

PR #2050, opened 2018-11-26
s/non-existent/not yet created/

### LINKED ISSUE #2048: Receipt of STOP_SENDING on already-closed stream
As of today, the draft states the following:

> Receiving a STOP_SENDING frame for a locally-initiated send stream that is “Ready” or non-existent MUST be treated as a connection error of type PROTOCOL_VIOLATION.

Due to packet loss, it is possible for a _STOP_SENDING_ frame to arrive after the stream initiator closes the stream.  In this case, it should not be an error.

martinthomson:
This is the only relevant condition.

Note that, as stated, it is valid to receive the frame for a stream that has been closed.

Closes #2048.


mikkelfj:
thanks

PR #2055, opened 2018-11-26
fix packet number decoding example

marten-seemann:
I stumbled upon this when I tried to build a unit test for my packet number decoding from the example.

I translated the pseudo code from the appendix to Python to double check: https://repl.it/@marten_seemann/QUIC-Packet-Number-Decoding


martinthomson:
Ugh, I ran the calculations myself, but obviously messed up somewhere.  The 2f value is clearly closer.

PR #2066, opened 2018-11-29
Refactor DetectLostPackets

martinthomson:
This had a couple of problems that I think this addresses, and some that
it doesn't.

The first is that it isn't clear what is being iterated over and in what
order.  I think that the point here is to iterate over sent_packets,
starting with the oldest.  Correct me if that is wrong.  In any case,
this change doesn't assume an iteration order.  The only reason the
iteration order is important is in setting loss_time, which needs to be
the earliest time (assuming that I'm right).

This simplifies the function, by setting thresholds at the top and doing
a simple comparison.

I've added a note about loss_time potentially being in the past.

The problem that remains is that this appeared to iterate only over
packets that have a packet number less than the largest acknowledged.
I've added that condition to the loop, but I don't think that it's
right.  I think that it's just redundant - and while an implementation
might stop its iteration at the largest acknowledged to save cycles,
this function will operate the same without the extra check.

It's also not clear whether the greater than comparisons here were
correct.  If you assume that firing of the timer cannot take zero time,
this is never an issue, but with discrete intervals on time values,
that's not always going to happen.  As setup, this code could be called
at exactly loss_time for a packet, in which case that packet will not be
declared lost.  I think that this wants >= in the time comparison for
that reason.

Finally, should the early retransmit timer be configurable?  Should it
be set based on kTimeReorderingThreshold?

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -867,26 +867,39 @@ Pseudocode for DetectLostPackets follows:
 DetectLostPackets(largest_acked):
   loss_time = 0
   lost_packets = {}
-  delay_until_lost = kTimeThreshold *
-                     max(latest_rtt, smoothed_rtt)
-  foreach (unacked < largest_acked.packet_number):
-    time_since_sent = now() - unacked.time_sent
-    delta = largest_acked.packet_number - unacked.packet_number
-    if (time_since_sent > delay_until_lost ||
-        delta > kPacketThreshold):
+  loss_delay = kTimeThreshold * max(latest_rtt, smoothed_rtt)
+  lost_packet_send_time = now()
+
+  // Packets sent before this time are deemed lost.
+  lost_send_time -= loss_delay
+
+  // Packets with packet numbers before this are deemed lost.
+  lost_pn = largest_acked.packet_number - reordering_threshold
+
+  foreach unacked:
+    if unacked.packet_number > largest_acked.packet_number:
+      continue
+
+    if (unacked.time_sent > lost_send_time ||
+        unacked.packet_number > lost_pn):
       sent_packets.remove(unacked.packet_number)
       if (unacked.retransmittable):
         lost_packets.insert(unacked)
-    else if (loss_time == 0 && delay_until_lost != infinite):
-      loss_time = now() + delay_until_lost - time_since_sent
+    else:
+      loss_time = min(loss_time, unacked.time_sent + loss_delay)
 
   // Inform the congestion controller of lost packets and
-  // lets it decide whether to retransmit immediately.
+  // let it decide whether to retransmit immediately.
   if (!lost_packets.empty()):
     OnPacketsLost(lost_packets)
 ~~~
 
+This algorithm results in loss_time being set to the earliest time that the
+earliest packet was sent.  As a result loss_time could be in the past.  Timers
+set based on a loss_time that has already passed need to fire immediately.
+
 ## Discussion
+
 The majority of constants were derived from best common practices among widely
 deployed TCP implementations on the internet.  Exceptions follow.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -913,32 +913,43 @@ Pseudocode for DetectLostPackets follows:
...
+                     max(latest_rtt, smoothed_rtt)
   else if (largest_acked.packet_number == largest_sent_packet):
     // Early retransmit timer.
-    delay_until_lost = 9/8 * max(latest_rtt, smoothed_rtt)
-  foreach (unacked < largest_acked.packet_number):
-    time_since_sent = now() - unacked.time_sent
-    delta = largest_acked.packet_number - unacked.packet_number
-    if (time_since_sent > delay_until_lost ||
-        delta > reordering_threshold):
+    loss_delay = 9/8 * max(latest_rtt, smoothed_rtt)
```
janaiyengar: This can be (1 + time_reordering_fraction)

draft-ietf-quic-recovery.md
```diff
@@ -913,32 +913,43 @@ Pseudocode for DetectLostPackets follows:
...
+  // Packets sent before this time are deemed lost.
+  lost_send_time -= loss_delay
+  lost_pn = largest_acked.packet_number - reordering_threshold
+
+  foreach unacked:
+    if unacked.packet_number > largest_acked.packet_number:
+      continue
+
+    if (unacked.time_sent > lost_send_time ||
+        unacked.packet_number > lost_pn):
```
janaiyengar: This should be
```
    if (lost_send_time > unacked.time_sent || lost_pn > unacked.packet_number)
```

draft-ietf-quic-recovery.md
```diff
@@ -913,32 +913,43 @@ Pseudocode for DetectLostPackets follows:
...
+      loss_time = min(loss_time, unacked.time_sent + loss_delay)
 
   // Inform the congestion controller of lost packets and
   // lets it decide whether to retransmit immediately.
   if (!lost_packets.empty()):
     OnPacketsLost(lost_packets)
 ~~~
 
+This algorithm results in loss_time being set to the earliest time that the
+earliest packet was sent.  As a result loss_time could be in the past.  Timers
```
janaiyengar: This algorithm results in loss_time being set based on the earliest packet that is still in flight. As a result ...

draft-ietf-quic-recovery.md
```diff
@@ -913,32 +913,42 @@ Pseudocode for DetectLostPackets follows:
...
+    loss_delay = (1 + time_reordering_fraction) *
+                     max(latest_rtt, smoothed_rtt)
+    lost_send_time -= loss_delay
+
+  foreach unacked in sent_packets:
+    if unacked.packet_number > largest_acked.packet_number:
+      continue
+
+    if (unacked.time_sent < lost_send_time ||
+        unacked.packet_number < lost_pn):
```
```suggestion
        unacked.packet_number <= lost_pn):
```

mikkelfj:
How about gaps in PN sequence (optimistic ACK defense)

draft-ietf-quic-recovery.md
```diff
@@ -867,26 +867,41 @@ Pseudocode for DetectLostPackets follows:
...
 
   // Inform the congestion controller of lost packets and
-  // lets it decide whether to retransmit immediately.
+  // let it decide whether to retransmit immediately.
   if (!lost_packets.empty()):
     OnPacketsLost(lost_packets)
 ~~~
 
+This algorithm results in loss_time being set based on the earliest packet that
+is still in flight.  Timers set based on a loss_time that has already passed
```
ianswett: I think this comment about timers with timeouts that have already passed is more widely applicable, so maybe move it closer to SetLossDetectionTimer()?

martinthomson:
Thanks for cleaning this up Jana.  I wasn't sure if the timers comment was scoped specifically to this particular instance, so I didn't make it a more general statement.  @ianswett, do you think that you could move that for me?

ianswett:
Thanks @martinthomson, I updated to what I believe is the right location.  Feel free to update further.

janaiyengar:
@mikkelfj: the loss recovery algorithm does not handle arbitrary gaps in sequence space. This can be handled with some complexity, which is probably worth writing a sentence or two about. Mind filing an issue?

PR #2069, opened 2018-11-29
Don't let the PMTU drop below 1280

### LINKED ISSUE #2063: ICMP DOS
I read through the recent PR #2036 which cleans up the packet size discussion. The text does touch upon ICMP denial of service attacks but I'm afraid the protection is too fragile because it is required that transmission stops on a path where PMTU is suspected to have been decreased. It is not explicitly stated (as far as I can tell) that ICMP packet too big is used for this decision, but it is used in PMTU discovery.

There is no major problem with attacks on discovery, but if ICMP can later close a connection, that is a big problem.

There is protection against off path attackers, but it is easy for a man on the side attack to generate ICMP messages that would pass validation. This only requires one compromised machine on a LAN.

This issue also relates to the ongoing discussion on protection against the initial handshake by man on the side stopping handshake by generating invalid early handshake packets, only here it can happen long after the connection is established.

A possible mitigation could be that a path should not be dropped based in ICMP messages alone but only after one or two new new PMTU probes. However, these new probes could also be vulnerable to forged ICMP messages.

martinthomson:
The loss of packets on the path will be the signal that we use here.

Closes #2063.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3068,6 +3068,9 @@ these messages.  This use of ICMP messages is potentially vulnerable to off-path
 attacks that successfully guess the IP address 3-tuple and reduce the PMTU to a
 bandwidth-inefficient value.
 
+An endpoint MUST ignore an ICMP message that claims the PMTU has decreased below
+1280 bytes.
+
 QUIC endpoints SHOULD provide validation to protect from off-path injection of
 ICMP messages as specified in {{!RFC8201}} and Section 5.2 of {{!RFC8085}}. This
 uses the quoted packet supplied in the payload of an ICMP message, which, when


draft-ietf-quic-transport.md
```diff
@@ -3100,9 +3100,11 @@ Further validation can also be provided:
   use for validation (for example, the IP ID or UDP checksum).
 
 The endpoint SHOULD ignore all ICMP messages that are not validated or do not
-carry sufficient quoted packet payload to perform validation.  Any reduction in
-the QUIC maximum packet size MAY be provisional until QUIC's loss detection
-algorithm determines that the quoted packet has actually been lost.
+carry sufficient quoted packet payload to perform validation.  ICMP messages
+that suggest that the PMTU has decreased below 1280 bytes SHOULD also be
```
marten-seemann: Why only a SHOULD? We say that QUIC depends on a PMTU of at least 1280 bytes, so I think a MUST would be justified here.

martinthomson: Oops, forgot to checkin my second pass.  I had exactly the same thought.

PR #2070, opened 2018-11-29
Remove paragraph talking about retaining DCID

dtikhonov:
The paragraph in question became what it is now in 391641ed60351d5685d5121b261059862dfd8e63.

Before CID sequence numbers, the endpoint that retired CIDs might have wanted to remember the CIDs it retired for some time to deduplicate _NEW_CONNECTION_ID_ frames.  Now this is no longer necessary and this paragraph is an oddity.


martinthomson:
Yep, thanks for this.

PR #2072, opened 2018-11-29
DUPLICATE_PUSH instead of multiple PUSH_PROMISE frames

### LINKED ISSUE #1864: Push promise history to detect conflicting header field values
draft-ietf-quic-http-15 has:

```
   []                           A client MUST treat receipt of a
   PUSH_PROMISE with conflicting header field values for the same Push
   ID as a connection error of type HTTP_MALFORMED_FRAME.

   Allowing duplicate references to the same Push ID is primarily to
   reduce duplication caused by concurrent requests.  A server SHOULD
   avoid reusing a Push ID over a long period.  Clients are likely to
   consume server push responses and not retain them for reuse over
   time.  Clients that see a PUSH_PROMISE that uses a Push ID that they
   have since consumed and discarded are forced to ignore the
   PUSH_PROMISE.
```

The two paragraphs contradict each other.  The former mandates that a client maintain a history of all push promises in order to detect possible header field conflicts.  The latter suggests that a client does not have to do that.  Which one is it?

The requirement to maintain a history of all push requests is unreasonable.  I suggest that the _MUST_ in the first paragraph above be replaced with a _MAY_.

### LINKED ISSUE #1959: what is the default dependent stream of pushed streams
in http/2 the default dependency of the pushed stream was it's associated stream. However in h3 we now have push ids and not associated streams. Should we have a default priority here?

MikeBishop:
Fixes #1864, by only sending the request headers once and having a separate frame to pick them up by reference.  ~Also would have fixed #1959 along the way, as an added bonus.~ (Separated into new PR so I can fix some other PRIORITY things.)

However, this is not an unmitigated improvement.  The point of the ordering requirement for PUSH_PROMISE frames is that the client knows that a push is coming before they see references to a dependent resource and initiate their own requests.  If reordering (or blocking QPACK references) places a DUPLICATE_PUSH ahead of the corresponding PUSH_PROMISE, knowing that the resource is on the way can no longer be guaranteed.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -723,37 +723,20 @@ The payload consists of:
 Push ID:
 : A variable-length integer that identifies the server push operation.  A Push
   ID is used in push stream headers ({{server-push}}), CANCEL_PUSH frames
-  ({{frame-cancel-push}}), and PRIORITY frames ({{frame-priority}}).
+  ({{frame-cancel-push}}), DUPLICATE_PUSH frames ({{frame-duplicate-push}}), and
+  PRIORITY frames ({{frame-priority}}).
 
 Header Block:
 : QPACK-compressed request header fields for the promised response.  See [QPACK]
   for more details.
 
 A server MUST NOT use a Push ID that is larger than the client has provided in a
-MAX_PUSH_ID frame ({{frame-max-push-id}}).  A client MUST treat receipt of a
-PUSH_PROMISE that contains a larger Push ID than the client has advertised as a
-connection error of type HTTP_MALFORMED_FRAME.
+MAX_PUSH_ID frame ({{frame-max-push-id}}) and MUST NOT use the same Push ID in
+multiple PUSH_PROMISE frames.  A client MUST treat receipt of a PUSH_PROMISE
+that contains a larger Push ID than the client has advertised or a Push ID which
+has already been promised as a connection error of type HTTP_MALFORMED_FRAME.
 
-A server MAY use the same Push ID in multiple PUSH_PROMISE frames.  This allows
-the server to use the same server push in response to multiple concurrent
-requests.  Referencing the same server push ensures that a PUSH_PROMISE can be
-made in relation to every response in which server push might be needed without
-duplicating pushes.
-
-A server that uses the same Push ID in multiple PUSH_PROMISE frames MUST include
-the same header fields each time.  The bytes of the header block MAY be
-different due to differing encoding, but the header fields and their values MUST
-be identical.  Note that ordering of header fields is significant.  A client
-MUST treat receipt of a PUSH_PROMISE with conflicting header field values for
-the same Push ID as a connection error of type HTTP_MALFORMED_FRAME.
-
-Allowing duplicate references to the same Push ID is primarily to reduce
-duplication caused by concurrent requests.  A server SHOULD avoid reusing a Push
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -825,6 +808,46 @@ MUST be treated as a connection error of type HTTP_MALFORMED_FRAME.
...
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                          Push ID (i)                        ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+~~~~~~~~~~
+{: #fig-duplicate-push title="DUPLICATE_PUSH frame payload"}
+
+The DUPLICATE_PUSH frame carries a single variable-length integer that
+identifies the Push ID of a resource that the server has previously promised
+(see {{frame-push-promise}}).  A server MUST treat a DUPLICATE_PUSH frame
```
martinthomson: You should point out that DUPLICATE_PUSH might appear before PUSH_PROMISE.  I know it's obvious, but it's also very important and so should be written down.

draft-ietf-quic-http.md
```diff
@@ -1007,11 +1030,16 @@ HTTP/3 uses a priority scheme similar to that described in {{!RFC7540}}, Section
...
-connection form a dependency tree. The structure of the dependency tree changes
-as PRIORITY frames add, remove, or change the dependency links between requests.
+connection form a dependency tree.
 
-The PRIORITY frame {{frame-priority}} identifies a prioritized element. The
-elements which can be prioritized are:
+When a client request is first sent or a placeholder first allocated, the
+element is dependent on the root of the priority tree. Pushed streams are
+initially dependent on the client request on which the PUSH_PROMISE frame was
+sent. In all cases, elements are assigned an initial weight of 16.
```
martinthomson: This is a nice side-effect.

MikeBishop: Moved to #2074 so I could do more priority-related changes without blocking on this, but I agree.  :-)

draft-ietf-quic-http.md
```diff
@@ -1106,6 +1131,13 @@ PUSH_PROMISE in relation to certain parts of the response is important (see
 Section 8.2.1 of {{!RFC7540}}).  Promised requests MUST conform to the
 requirements in Section 8.2 of {{!RFC7540}}.
 
+The same server push can be associated with additional client requests using a
+DUPLICATE_PUSH frame (see {{frame-duplicate-push}}).  Ordering of a
+DUPLICATE_PUSH in relation to certain parts of the response is similarly
+important.  Due to reordering, DUPLICATE_PUSH frames can arrive before the
+corresponding PUSH_PROMISE frame, in which case the request headers of the push
+might not be immediately available.
```
dtikhonov: Why "might not?"  If _DUPLICATE_PUSH_ arrives before the _PUSH_PROMISE_ it references, the request headers **won't** be available.

MikeBishop: I was thinking conditionally, "might not be available because DP could arrive before PP," but since this is inside the hypothetical, you're correct.

PR #2077, opened 2018-11-30
Do no define connection as a conversation

dtikhonov:
The target audience is not likely to be confused by the use of the term _connection_.  On the other hand, _conversation_ is not defined in this document; it is a superfluous addition.

_Encryption context_ is similarly not defined and the only reference to it removed in this change.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -192,13 +192,9 @@ QUIC packet:
 
 Endpoint:
 
-: An entity that can participate in a QUIC conversation by generating,
+: An entity that can participate in a QUIC connection by generating,
   receiving, and processing QUIC packets. There are only two types of endpoint
   in QUIC: client and server.
-
-Connection:
-
-: A conversation between two QUIC endpoints with a single encryption context.
 
 Client:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -865,11 +861,10 @@ longer if the peer chooses to not send STREAMS_BLOCKED frames.
...
 
-A QUIC connection is a single conversation between two QUIC endpoints.  QUIC's
-connection establishment combines version negotiation with the cryptographic
-and transport handshakes to reduce connection establishment latency, as
-described in {{handshake}}.  Once established, a connection may migrate to a
-different IP or port at either endpoint as
+QUIC's connection establishment combines version negotiation with the
+cryptographic and transport handshakes to reduce connection establishment
+latency, as described in {{handshake}}.  Once established, a connection
+may migrate to a different IP or port at either endpoint as
```
marten-seemann: We only specify client-side migration.

dtikhonov: This is reflowed verbiage: I only made changes related to the "connection is a conversation" bit.

draft-ietf-quic-transport.md
```diff
@@ -192,14 +192,10 @@ QUIC packet:
 
 Endpoint:
 
-: An entity that can participate in a QUIC conversation by generating,
+: An entity that can participate in a QUIC connection by generating,
```
```suggestion
: An entity that participates in a QUIC connection by generating,
```

dtikhonov: I thought about `s/can participate/participates/` but then I realized that during the handshake the endpoints send and receive QUIC packets before the connection is established -- perhaps that's why it was written this way.  I also want to limit this change only to the connection/conversation business.

draft-ietf-quic-transport.md
MikeBishop: @marten-seemann, I think this is an artifact of the effort to specify the migration primitives generically, even if only clients can migrate freely in v1.  Given the SPA mechanism, this is literally true as written, but more constrained that one might infer from this text.  Feel free to open a separate editorial issue for this, but I don't think we should make @dtikhonov fix it in his unrelated PR.

PR #2078, opened 2018-11-30
No RTT for non-retransmittable packets

ianswett:
Only take RTT samples when the largest acked is a retransmittable packet.  Otherwise, the ack delay may be unlimited.

Came up in discussion of #2060

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -284,12 +284,12 @@ frame.  For more information, see Section 13.2 of {{QUIC-TRANSPORT}}.
 ## Computing the RTT estimate
 
 RTT is calculated when an ACK frame arrives by computing the difference between
-the current time and the time the largest newly acked packet was sent.  If no
-packets are newly acknowledged, RTT cannot be calculated. When RTT is
-calculated, the ack delay field from the ACK frame SHOULD be subtracted from the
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -284,12 +284,12 @@ frame.  For more information, see Section 13.2 of {{QUIC-TRANSPORT}}.
...
 RTT is calculated when an ACK frame arrives by computing the difference between
-the current time and the time the largest newly acked packet was sent.  If no
-packets are newly acknowledged, RTT cannot be calculated. When RTT is
-calculated, the ack delay field from the ACK frame SHOULD be subtracted from the
-RTT as long as the result is larger than the Min RTT.  If the result is smaller
-than the min_rtt, the RTT should be used, but the ack delay field should be
-ignored.
+the current time and the time the largest acked packet was sent.  If the largest
+acked is not newly acknowledged or not retransmittable, RTT cannot be
+calculated. When RTT is calculated, the ack delay field from the ACK frame
```
janaiyengar: I would make this normative. "An RTT sample MUST NOT be taken for a packet that is not newly acknowledged or not retransmittable."

ianswett: Thanks, updated

PR #2080, opened 2018-12-01
Editorial rework

janaiyengar:
Editorial changes suggested against #1974

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -311,9 +311,8 @@ A packet is declared lost under the following conditions:
   packet ({{packet-threshold}}), or it was sent long enough in the past
   ({{time-threshold}}).
 
-The acknowledgement of a later packet indicates that a packet sent later was
-delivered, while the packet and time thresholds provide some tolerance for
-packet reordering.
+The acknowledgement indicates that a packet sent later was delivered, while the
+packet and time thresholds provide some tolerance for packet reordering.
 
 Spuriously declaring packets lost leads to unnecessary retransmissions and may
 result in degraded performance due to the actions of the congestion controller
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -570,7 +570,7 @@ kMaxTLPs:
   The RECOMMENDED value is 2.
 
 kPacketThreshold:
-: Maximum reordering in packets  before packet threshold loss detection
+: Maximum reordering in packets before packet threshold loss detection
   considers a packet lost. The RECOMMENDED value is 3.
 
 kTimeThreshold:
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -876,7 +876,8 @@ Pseudocode for DetectLostPackets follows:
 DetectLostPackets(largest_acked):
   loss_time = 0
   lost_packets = {}
-  delay_until_lost = time_threshold * max(latest_rtt, smoothed_rtt)
+  delay_until_lost = time_threshold *
+                     max(latest_rtt, smoothed_rtt)
   foreach (unacked < largest_acked.packet_number):
     time_since_sent = now() - unacked.time_sent
     delta = largest_acked.packet_number - unacked.packet_number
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -325,51 +299,50 @@ underestimation of min RTT, which in turn prevents underestimating smoothed RTT.
...
-sent.  If packets sent prior to the largest acknowledged packet cannot yet
-be declared lost, then a timer SHOULD be set for the remaining time.
+{{?RFC6675}}, and RACK {{?RACK=I-D.ietf-tcpm-rack}}. This section provides an
+overview of how these algorithms are implemented in QUIC.
+
+When an ACK frame is received that newly acknowledges a packet, a QUIC sender
+uses two thresholds to determine if prior packets should be declared lost.  A
+packet is declared lost under the following conditions:
+
+* it has a smaller packet number than the newly acknowledged packet,
```
ianswett: I would prefer merging the first two bullets because I think it might flow better.
ie: "The packet is retransmittable, unacknowledged, and was sent prior to an acknowledged packet."

The removal of the mention of packet number was to capture the concept a bit more and we talk about packet number in the next bullet.

Also, we typically capitalize the first letter after the *, which I think looks better.

draft-ietf-quic-recovery.md
```diff
@@ -596,13 +570,14 @@ kMaxTLPs:
   The RECOMMENDED value is 2.
 
 kPacketThreshold:
-: Maximum reordering in packet number space before FACK style loss detection
+: Maximum reordering in packets  before packet threshold loss detection
```
```suggestion
: Maximum reordering in packets before packet threshold loss detection
```

draft-ietf-quic-recovery.md
```diff
@@ -325,51 +299,50 @@ underestimation of min RTT, which in turn prevents underestimating smoothed RTT.
...
-of the round-trip time (kTimeThreshold), is 1/8.
-
-An endpoint SHOULD declare packets lost no earlier than
-(1 + kTimeThreshold) * max(SRTT, latest_RTT) after when they were
-sent.  If packets sent prior to the largest acknowledged packet cannot yet
-be declared lost, then a timer SHOULD be set for the remaining time.
+{{?RFC6675}}, and RACK {{?RACK=I-D.ietf-tcpm-rack}}. This section provides an
+overview of how these algorithms are implemented in QUIC.
+
+When an ACK frame is received that newly acknowledges a packet, a QUIC sender
```
ianswett: I'm now thinking we should drop newly because we also trigger loss on a timer.  If there's something subtle we're trying to say there, we should probably make it more explicit?

At some point, we discussed basing loss detection on the largest recently acknowledged packet(ie: in the last ACK), but I don't think that's what the pseudocode does and I don't think that's what the current gQUIC code does.  If we want to discuss that, maybe add a paragraph that the algorithm may be enhanced by doing so and it may reduce spurious retransmits in some cases?

draft-ietf-quic-recovery.md
```diff
@@ -325,51 +299,50 @@ underestimation of min RTT, which in turn prevents underestimating smoothed RTT.
...
+The RECOMMENDED initial value for the packet reordering threshold
+(kPacketThreshold) is 3, based on TCP loss detection {{?RFC5681}} {{?RFC6675}}.
+
+Some networks may exhibit higher degrees of reordering, causing a sender to
+detect spurious losses.  Implementers MAY use algorithms developed for TCP, such
+as TCP-NCR {{?RFC4653}}, to improve QUIC's reordering resilience.
+
+### Time Threshold {#time-threshold}
+
+Ack-based loss detection uses a time threshold to determine how much reordering
```
ianswett: I find changing this sentence to start with "Ack-based" is confusing, because there's also the packet number threshold.

How about dropping this first sentence entirely, since I think it's said better above?

draft-ietf-quic-recovery.md
janaiyengar: This is a good point. In fact, I think we may want to drop the text that says this is triggered on an receipt of an ACK only, because that's not necessary. Sure, that's when we check it (and when the ER timer fires), but that's not required.  I've changed the text above to reflect this.

draft-ietf-quic-recovery.md
janaiyengar: Done, and moved this para below. The ordering is better now I think.

PR #2088, opened 2018-12-01
Server Ignore VN Packets

nibanks:
To prevent the possibility of two servers being tricked into sending endless VN packets between themselves, a server should never respond to a VN packet.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1108,7 +1108,9 @@ is a mutually supported version.
 
 If the version selected by the client is not acceptable to the server, the
 server responds with a Version Negotiation packet (see {{packet-version}}).
-This includes a list of versions that the server will accept.
+This includes a list of versions that the server will accept.  A server MUST NOT
+send a Version Negotiation packet in response to receiving a Version Negotiation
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1108,7 +1108,9 @@ is a mutually supported version.
 
 If the version selected by the client is not acceptable to the server, the
 server responds with a Version Negotiation packet (see {{packet-version}}).
-This includes a list of versions that the server will accept.
+This includes a list of versions that the server will accept.  A server MUST NOT
+send a Version Negotiation packet in response to receiving a Version Negotiation
+packet because it could trigger and endless back and forth between two servers.
```
```suggestion
packet.
```

MikeBishop:
@ianswett, this is a follow-up from an old issue.  The text previously said that you send VN in response to a version you don't speak, and for some implementations, version 0 is a version they don't speak, so....

This effectively clarifies that you **do** speak version 0, and if you're not a client attempting a connection, the appropriate action in that version is to drop it.

ianswett:
Thanks for the explanation Mike, the idea of a QUIC implementation that doesn't support version 0 is a bit weird, but that's fine.

PR #2095, opened 2018-12-04
Unnecessary variables in recovery

janaiyengar:


janaiyengar:
I don't think so, since we aren't doing that in the draft. We're not requiring anyone to implement the code as is. And if they want to do this adaptively, implementers will have to depart from the draft anyways. This change makes the draft tighter (and more readable, IMO).

PR #2096, opened 2018-12-04
NEW_CONNECTION_ID: use sequence number for one CID only

dtikhonov:


mikkelfj:
How would know that the sequence number is from a different CID? If you have this concern, wouldn't it be better to advice not to send a sequence number for different CID, and keep a range check receiver side on the assumption that the sequence number is specific to the given CID domain?

dtikhonov:
As long as we are checking for for consistency, it makes sense to check that sequence number has not been reused.  Sequence number reuse means that there is no longer a one-to-one correspondence between CIDs and sequence numbers.

The check is opportunistic (thus it's a **MAY** treat it as an error): if the peer has a record from previous _NEW_CONNECTION_ID_ `(seqno: A, cid: ABC)` but it receives `(seqno: A, cid: XYZ)`, it is an error.

mikkelfj:
I'm not opposed to checking - I did write range check - only that I don't see how you can conclude a sequence number is wrong based on its assumed incorrect origin.

dtikhonov:
I don't understand what you mean by "domain" and "origin" -- there is only one origin of new CIDs and that's the peer.

mikkelfj:
Never mind, I way into multi-path ...

dtikhonov:
> Unless you're keeping a full history of CIDs for the life of the connection

One can check against the current set (I do).  It's probably too costly to keep a full history.

dtikhonov:
>  I might flip around the statements to put the two that are logically similar together

The proposed wording is similar to your list, @MikeBishop, but with CID and seqno swapped:

- The same CID with different values (seqno or RT)
- A repeated seqno (new CID)

I don't think one is better than the other.

PR #2099, opened 2018-12-05
Update draft-ietf-quic-recovery.md

ianswett:
Fixes part of #2060

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -267,6 +267,26 @@ all subsequent ACK frames containing them could be lost. In this case, the
 loss recovery algorithm may cause spurious retransmits, but the sender will
 continue making forward progress.
 
+# Computing the RTT estimate
+
+RTT is calculated when an ACK frame arrives by computing the difference between
+the current time and the time the largest acked packet was sent.  If the
+largest acked packet is not newly acknowledged, RTT cannot be calculated.
+
+When RTT is calculated, the ack delay field from the ACK frame SHOULD be limited
+to the max_ack_delay specified by the peer.  Limiting ack_delay to max_ack_delay
+ensures a peer specifying an extremely small max_ack_delay doesn't cause more
+spurious timeouts than a peer that correctly specifies max_ack_delay. It SHOULD
+be subtracted from the RTT as long as the result is larger than the Min RTT.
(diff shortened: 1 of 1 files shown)


janaiyengar:
@kazuho pointed out offline (and I agree) that we need to exempt packets containing only non-retransmittable (argh, terminology) frames, since an ACK may be sent after an arbitrarily long period in this case.

draft-ietf-quic-recovery.md
```diff
@@ -267,6 +267,26 @@ all subsequent ACK frames containing them could be lost. In this case, the
...
+
+RTT is calculated when an ACK frame arrives by computing the difference between
+the current time and the time the largest acked packet was sent.  If the
+largest acked packet is not newly acknowledged, RTT cannot be calculated.
+
+When RTT is calculated, the ack delay field from the ACK frame SHOULD be limited
+to the max_ack_delay specified by the peer.  Limiting ack_delay to max_ack_delay
+ensures a peer specifying an extremely small max_ack_delay doesn't cause more
+spurious timeouts than a peer that correctly specifies max_ack_delay. It SHOULD
+be subtracted from the RTT as long as the result is larger than the Min RTT.
```
janaiyengar: s/Min RTT/min_rtt

draft-ietf-quic-recovery.md
```diff
@@ -267,6 +267,26 @@ all subsequent ACK frames containing them could be lost. In this case, the
...
+If the result is smaller than the min_rtt, the RTT should be used, but the
+ack delay field should be ignored.
```
janaiyengar: I wouldn't add this in before we resolve #2060.

draft-ietf-quic-recovery.md
```diff
@@ -267,6 +267,26 @@ all subsequent ACK frames containing them could be lost. In this case, the
...
+
+Like TCP, QUIC calculates both smoothed RTT and RTT variance similar to those
+specified in {{?RFC6298}}.
+
+Min RTT is the minimum RTT measured over the connection, prior to adjusting by
```
janaiyengar: s/Min RTT/min_rtt

draft-ietf-quic-recovery.md
ianswett: This is just moving text from below, FYI

PR #2102, opened 2018-12-06
Fix QPACK-04 changelog: add two non-editorial changes

dtikhonov:


draft-ietf-quic-qpack.md
```diff
@@ -1254,7 +1254,9 @@ return controlBuffer, prefixBuffer + streamBuffer
 
 ## Since draft-ietf-quic-qpack-03
 
-Substantial editorial reorganization; no technical changes.
+- Avoid having multiple ways to encode Delta Base Index of zero (#2005)
```
martinthomson: Based on the tags on drafts, this change is in -05, not -04.

PR #2104, opened 2018-12-06
Make in_flight the criterion for declaring loss

martinduke:
If a packet contains just ACKs and PADDING, it is not retransmittable but must be declared lost to free up the congestion window space.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -304,7 +304,7 @@ Ack-based loss detection implements the spirit of TCP's Fast Retransmit
 {{?RFC6675}}, and RACK {{?RACK=I-D.ietf-tcpm-rack}}. This section provides an
 overview of how these algorithms are implemented in QUIC.
 
-A packet is declared lost if it meets both of the following conditions:
+A packet is declared lost if it meets all the following conditions:
 
 * The packet is unacknowledged, in-flight, and was sent prior to an
   acknowledged packet.
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -302,10 +302,9 @@ Ack-based loss detection implements the spirit of TCP's Fast Retransmit
 {{?RFC6675}}, and RACK {{?RACK=I-D.ietf-tcpm-rack}}. This section provides an
 overview of how these algorithms are implemented in QUIC.
 
-A packet is declared lost under the following conditions:
+A packet is declared lost if it meets both of the following conditions:
 
-* The packet is unacknowledged, retransmittable, and was sent prior to an
-  acknowledged packet.
+* The packet is in-flight, and was sent prior to an acknowledged packet.
 
 * Either its packet number is kPacketThreshold smaller than an acknowledged
   packet ({{packet-threshold}}), or it was sent long enough in the past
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -302,10 +302,9 @@ Ack-based loss detection implements the spirit of TCP's Fast Retransmit
 {{?RFC6675}}, and RACK {{?RACK=I-D.ietf-tcpm-rack}}. This section provides an
 overview of how these algorithms are implemented in QUIC.
 
-A packet is declared lost under the following conditions:
+A packet is declared lost if it meets both of the following conditions:
 
-* The packet is unacknowledged, retransmittable, and was sent prior to an
-  acknowledged packet.
+* The packet is in-flight, and was sent prior to an acknowledged packet.
```
ianswett: Did you intend to drop "unacknowledged"?

janaiyengar:
I don't think this is correct. We mark something as lost because we expected to hear an ACK back but didn't. That's true of ack-eliciting (née retransmittable) but not of things that are considered in flight.  PADDING is the reason -- it is considered in flight, but doesn't elicit an ACK, so a sender that pads ACK-only packets should add a PING frame in there to elicit an ack for this reason, so that the bytes in flight can be recovered when the packet is lost.

janaiyengar:
> I think this is an excellent reason to move the text about "Generating Acknowlegements" into transport, because having it in two places increases the chances of inconsistencies.

Agreed. It's on my todo for next week.

That's about what to do with acking PADDING packets. The question I'm raising is whether loss detection is tied to being in flight or to eliciting an ACK. I'm arguing that loss detection is premised on bytes counting towards in-flight, but loss detection logic triggered by receiving an ACK (which is why we use max_ack_delay in the computations).  The bytes being in flight are a necessary condition, not sufficient.

Loss detection marks a packet as lost when no ACK is received for it. This is predicated on an ACK being sent by the receiver, delayed or not, if the packet is received, which is only true for ack-eliciting packets.

PADDING of ACK-only frames is a weird special case that causes inflight to go up without being ack-eliciting. We had agreed in Montreal that if someone did this they will need to either add a PING frame in there or handle special conditions.

mikkelfj:
How about a sender self-acknowledge packets that are not ack-eliciting, for example packets older than largest acked or older than triggering TLP? Once these packets are ack'ed or self-ack'ed, they are retired from the congestion window.

Or just keep the packets out of the congestion window if that is at all feasible. I'm not really keen on the above complexity if it can be avoided.

nibanks:
I feel like this is somewhat related to my issue (#2015) related to calling `OnPacketsLost` for all lost packets, not just retransmittable/ack-eliciting ones. If you know the packet was lost, then process it as such. Now, what you do it response to it being lost is where you would then put the `in_flight` check for removing the `bytes_in`flight` and whatever other logic.

ianswett:
Hi Martin, I resolved the conflicts with the recently merged Ack-eliciting change for you.  Feel free to make further modifications.

If we decide to change this, we should ensure the text is clear as well and offers an explanation for the outcome of this discussion.

janaiyengar:
I wanted to ensure that we don't set the timer based on in-flight. This PR doesn't do that... and this PR is correct. I'll merge.

PR #2107, opened 2018-12-10
packet size is header + payload

igorlord:


dtikhonov:
The current wording is fine.  The sentence is not meant to be exclusive; rather, it is to clarify which headers are included into the packet size calculation and which aren't.

igorlord:
@dtikhonov 

> The QUIC packet size includes the QUIC header and integrity check, but not [...]

The current wording is not making sense to me.  Words _"integrity check"_ do not appear anywhere else anywhere else in any of the QUIC drafts.  The _"integrity check"_ is not a concept that is taking up bytes to be accounted for in _"QUIC packet size"_.

Sections 17.2 (_Long Header Packet_) and 17.3 (_Short Header Packet_) of the Transport draft define QUIC packets to contain the Header and Payload parts.

dtikhonov:
You're right about "integrity check."  What's probably meant is the AEAD tag.

draft-ietf-quic-transport.md
```diff
@@ -3047,8 +3047,8 @@ traverse network elements that drop or corrupt ECN codepoints in the IP header.
 
 # Packet Size {#packet-size}
 
-The QUIC packet size includes the QUIC header and integrity check, but not the
-UDP or IP header.
+The QUIC packet size includes the QUIC header and payload, but not the UDP or IP
```
```suggestion
The QUIC packet size includes the QUIC header and protected payload, but not the UDP or IP
```

PR #2108, opened 2018-12-10
Clarification and minor addition to ICMP PTB text

igorlord:


draft-ietf-quic-transport.md
```diff
@@ -3116,8 +3116,8 @@ PMTU discovery {{!RFC1191}} {{!RFC8201}} relies on reception of ICMP messages
 (e.g., IPv6 Packet Too Big messages) that indicate when a packet is dropped
 because it is larger than the local router MTU. DPLPMTUD can also optionally use
 these messages.  This use of ICMP messages is potentially vulnerable to off-path
-attacks that successfully guess the IP address 3-tuple and reduce the PMTU to a
-bandwidth-inefficient value.
+attacks that successfully guess the connection's 4-tuple and reduce the PMTU to
```
```suggestion
attacks that successfully guess the addresses used on the path and reduce the PMTU to
```

igorlord: Do you think that "addresses" implies UDP port numbers as well as IP addresses?

martinthomson: We've used it elsewhere in that way, so yes.

martinthomson:
@igorlord, can you open an issue for this?  It's a design-level change in that it causes implementations to change behaviour, so we'll want to track that.

PR #2111, opened 2018-12-11
Use "Insert Count" rather than "Largest Reference"

### LINKED ISSUE #2110: Largest Reference isn't the right concept
Counting by reference implies something about the indexing system that is unnecessary.

For instance, with one entry in the table that is used, the current system insists on us signaling a Largest Reference of 1.  But most people would treat that entry as being at an index of 0.

The thing that we're sending is the number of inserts.  Describing the field in those terms clarifies things considerably.

martinthomson:
The construction of largest reference currently implies the use of
1-based indexing, which is unnatural.  Using the insert count is much
more natural, even if it means the encoding is a little strange and a
tiny bit inefficient (it goes 0, 2, 3, ..., Max-1, 1, 2, ...).

I've gone through and put the references on the boundary between
records, which makes explanations much easier for post-base indexing in
my view.  I've renamed Base Index (which was never really an index) to
Base, which matches that change.  Similarly, renaming Largest Known
Received to Known Received Count makes it match better.

There are also examples added to Insert Count and Base sections.  These
definitions are pretty damned opaque without them.

Closes #2110.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -800,9 +800,15 @@ HEADERS and PUSH_PROMISE frames on request and push streams reference the
 dynamic table in a particular state without modifying it.  Frames on these
 streams emit the headers for an HTTP request or response.
 
-### Header Data Prefix {#header-prefix}
+### Header Block Prefix {#header-prefix}
 
-Header data is prefixed with two integers, `Required Insert Count` and `Base`.
+Each header block is prefixed with two integers.  The Required Insert Count
+encoded as an integer with an 8-bit prefix after the encoding described in
+{{ric}}).  The Base is encoded as sign-and-modulus integer, using a single sign
+bit and a value with a 7-bit prefix (see {{base}}).
+
+These two values are followed by instructions for compressed headers.  The
+entire block is expected to be framed by the using protocol.
 
 ~~~~~~~~~~  drawing
   0   1   2   3   4   5   6   7
@@ -816,7 +822,7 @@ Header data is prefixed with two integers, `Required Insert Count` and `Base`.
 ~~~~~~~~~~
 {:#fig-base-index title="Frame Payload"}
 
-#### Required Insert Count
+#### Required Insert Count {#ric}
 
 Required Insert Count identifies the state of the dynamic table needed to
 process the header block successfully.  Blocking decoders use the Required
@@ -868,7 +874,7 @@ For example, if the dynamic table is 100 bytes, then the Required Insert Count
 will be encoded modulo 6.  If a decoder has received 10 inserts, then an encoded
 value of 3 indicates that the Required Insert Count is 9 for the header block.
 
-#### Base
+#### Base {#base}
 
 `Base` is used to resolve references in the dynamic table as described in
 {{relative-indexing}}.
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -250,19 +250,24 @@ Because QUIC does not guarantee order between data on different streams, a
 header block might reference an entry in the dynamic table that has not yet been
 received.
 
-Each header block contains the total number of entries added to the dynamic
-table - the Required Insert Count - which identifies the table state necessary
-for decoding (see {{header-prefix}}). If the Required Insert Count is greater
-than the number of dynamic table entries received, the stream is considered
-"blocked."  While blocked, header field data SHOULD remain in the blocked
-stream's flow control window.  When the Required Insert Count is zero, the frame
-contains no references to the dynamic table and can always be processed
-immediately. A stream becomes unblocked when the number of entries inserted in
-the dynamic table becomes greater than or equal to the Required Insert Count for
-all header blocks the decoder has started reading from the stream.  If the
-decoder encounters a header block where the largest Absolute Index used not
-equal to the largest value permitted by the Required Insert Count, it MAY treat
-this as a stream error of type HTTP_QPACK_DECOMPRESSION_FAILED.
+Each header block contains a Required Insert Count, the lowest possible value
+for the Insert Count with which the header block can be decoded. For a header
+block with no references to the dynamic table, the Required Insert Count is
+zero.
+
+When the Required Insert Count is zero, the frame contains no references to the
+dynamic table and can always be processed immediately.
+
+If the Required Insert Count is greater than the number of dynamic table entries
+received, the stream is considered "blocked."  While blocked, header field data
+SHOULD remain in the blocked stream's flow control window.  A stream becomes
+unblocked when the Insert Count becomes greater than or equal to the Required
+Insert Count for all header blocks the decoder has started reading from the
+stream.
+
+If the decoder encounters a header block where the largest Absolute Index used
+is not equal to the largest value permitted by the Required Insert Count, it MAY
+treat this as a stream error of type HTTP_QPACK_DECOMPRESSION_FAILED.
 
 The SETTINGS_QPACK_BLOCKED_STREAMS setting (see {{configuration}}) specifies an
 upper bound on the number of streams which can be blocked. An encoder MUST limit
@@ -812,15 +817,16 @@ Header data is prefixed with two integers, `Required Insert Count` and `Base`.
 ~~~~~~~~~~
 {:#fig-base-index title="Frame Payload"}
 
-#### Insert Count
+#### Required Insert Count
 
-`Insert Count` identifies the state of the dynamic table needed to process the
-header block successfully.  Blocking decoders use the Insert Count to determine
-when it is safe to process the rest of the block.  If Insert Count is greater
-than zero, the encoder transforms it as follows before encoding:
+Required Insert Count identifies the state of the dynamic table needed to
+process the header block successfully.  Blocking decoders use the Required
+Insert Count to determine when it is safe to process the rest of the block.  If
+Required Insert Count is greater than zero, the encoder transforms it as follows
+before encoding:
 
 ~~~
-   EncodedInsertCount = (InsertCount mod (2 * MaxEntries)) + 1
+   EncodedInsertCount = (ReqInsertCount mod (2 * MaxEntries)) + 1
 ~~~
 
 Here `MaxEntries` is the maximum number of entries that the dynamic table can
@@ -840,28 +846,28 @@ algorithm:
 
 ~~~
    if EncodedInsertCount == 0:
-      InsertCount = 0
+      ReqInsertCount = 0
    else:
       InsertCount = EncodedInsertCount - 1
       CurrentWrapped = TotalNumberOfInserts mod (2 * MaxEntries)
 
       if CurrentWrapped >= InsertCount + MaxEntries:
          # Insert Count wrapped around 1 extra time
-         InsertCount += 2 * MaxEntries
+         ReqInsertCount += 2 * MaxEntries
       else if CurrentWrapped + MaxEntries < InsertCount:
          # Decoder wrapped around 1 extra time
          CurrentWrapped += 2 * MaxEntries
 
-      InsertCount += TotalNumberOfInserts - CurrentWrapped
+      ReqInsertCount += TotalNumberOfInserts - CurrentWrapped
 ~~~
 
 Where TotalNumberOfInserts is the total number of inserts into the decoder's
 dynamic table.  This encoding limits the length of the prefix on long-lived
 connections.
 
-For example, if the dynamic table is 100 bytes, then the Insert Count will be
-encoded modulo 6.  If a decoder has received 10 inserts, then an encoded value
-of 3 indicates that the InsertCount is 9 for the header block.
+For example, if the dynamic table is 100 bytes, then the Required Insert Count
+will be encoded modulo 6.  If a decoder has received 10 inserts, then an encoded
+value of 3 indicates that the Required Insert Count is 9 for the header block.
 
 #### Base
 
@@ -876,31 +882,32 @@ indicates that the Base is less than the Insert Count.  That is:
 
 ~~~
    if S == 0:
-      Base = InsertCount + DeltaBase
+      Base = ReqInsertCount + DeltaBase
    else:
-      Base = InsertCount - DeltaBase - 1
+      Base = ReqInsertCount - DeltaBase - 1
 ~~~
 
 A single-pass encoder determines the Base before encoding a header block.  If
 the encoder inserted entries in the dynamic table while encoding the header
-block, Insert Count will be greater than the Base, so the encoded difference is
-negative and the sign bit is set to 1.  If the header block did not reference
-the most recent entry in the table and did not insert any new entries, Base will
-be greater than the Insert Count, so the delta will be positive and the sign bit
-is set to 0.
+block, Required Insert Count will be greater than the Base, so the encoded
+difference is negative and the sign bit is set to 1.  If the header block did
+not reference the most recent entry in the table and did not insert any new
+entries, Base will be greater than the Required Insert Count, so the delta will
+be positive and the sign bit is set to 0.
 
 An encoder that produces table updates before encoding a header block might set
-Insert Count and Base to the same value.  In such case, both the sign bit and
-the Delta Base will be set to zero.
+Required Insert Count and Base to the same value.  In such case, both the sign
+bit and the Delta Base will be set to zero.
 
 A header block that does not reference the dynamic table can use any value for
-Base; setting both Insert Count and Delta Base to zero is the most efficient
-encoding.
+Base; setting both Required Insert Count and Delta Base to zero is the most
+efficient encoding.
 
-For example, with an Insert Count of 9, a decoder receives a S bit of 1 and a
-Delta Base of 2.  This sets the Base to 6 and enables post-base indexing for
-three entries.  In this example, a regular index of 1 refers to the 5th entry
-that was added to the table; a post-base index of 1 refers to the 8th entry.
+For example, with an Required Insert Count of 9, a decoder receives a S bit of 1
+and a Delta Base of 2.  This sets the Base to 6 and enables post-base indexing
+for three entries.  In this example, a regular index of 1 refers to the 5th
+entry that was added to the table; a post-base index of 1 refers to the 8th
+entry.
 
 
 ### Indexed Header Field
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -713,21 +713,21 @@ The Table State Synchronize instruction begins with the '00' two-bit pattern.
 The instruction specifies the total number of dynamic table inserts and
 duplications since the last Table State Synchronize or Header Acknowledgement
 that increased the Known Received Count for the dynamic table (see
-{{known-received-count}}).  This is encoded as a 6-bit prefix integer. The
-encoder uses this value to determine which table entries might cause a stream to
-become blocked, as described in {{state-synchronization}}.
+{{known-received-count}}).  The Insert Count Delta is encoded as a 6-bit prefix
+integer. The encoder uses this value to determine which table entries might
+cause a stream to become blocked, as described in {{state-synchronization}}.
 
 ~~~~~~~~~~ drawing
   0   1   2   3   4   5   6   7
 +---+---+---+---+---+---+---+---+
-| 0 | 0 |   Insert Count (6+)   |
+| 0 | 0 |Insert Count Delta (6+)|
 +---+---+-----------------------+
 ~~~~~~~~~~
 {:#fig-size-sync title="Table State Synchronize"}
 
-An encoder that receives an Insert Count equal to zero or one that increases
-Known Received Count beyond what the encoder has sent MUST treat this as a
-connection error of type `HTTP_QPACK_DECODER_STREAM_ERROR`.
+An encoder that receives an Insert Count Delta equal to zero or one that
+increases the Known Received Count beyond what the encoder has sent MUST treat
+this as a connection error of type `HTTP_QPACK_DECODER_STREAM_ERROR`.
 
 ### Header Acknowledgement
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -249,18 +250,19 @@ Because QUIC does not guarantee order between data on different streams, a
...
-zero, the frame contains no references to the dynamic table and can always be
-processed immediately. A stream becomes unblocked when the greatest absolute
-index in the dynamic table becomes greater than or equal to the Largest
-Reference for all header blocks the decoder has started reading from the stream.
-If the decoder encounters a header block where the actual largest reference is
-not equal to the Largest Reference declared in the prefix, it MAY treat this as
-a stream error of type HTTP_QPACK_DECOMPRESSION_FAILED.
+Each header block contains the total number of entries added to the dynamic
+table - the Insert Count - which identifies the table state necessary for
+decoding (see {{header-prefix}}). If the Insert Count is greater than the number
```
afrind: I think this will be somewhat confusing, because the state of the table at the encoder at the time of encoding (insert count) can be different than the table state (insert count) required to decode a particular block.  Maybe if you called it 'Required Insert Count' in the header prefix that would make it clear that an encoder shouldn't just blindly use the actual Insert Count?

draft-ietf-quic-qpack.md
```diff
@@ -456,31 +458,36 @@ referenced by a given relative index will change while interpreting instructions
...
 Because frames from request streams can be delivered out of order with
-instructions on the encoder stream, relative indices are relative to the Base
-Index at the beginning of the header block (see {{header-prefix}}). The Base
-Index is an absolute index. When interpreting the rest of the frame, the entry
-identified by Base Index has a relative index of zero.  The relative indices of
-entries do not change while interpreting headers on a request or push stream.
+instructions on the encoder stream, relative indices are relative to the Base at
+the beginning of the header block (see {{header-prefix}}). The Base is encoded
+as a value relative to the Insert Count, so it can be expressed as the number of
+entries inserted. The dynamic table entries up to the Base can be referenced.
```
afrind: What about post-base?

afrind: > The Base is encoded as a value relative to the Insert Count, so it can be expressed as the number entries inserted

I'm having a hard time understanding why you are trying to say here.  That Base in the number of entries inserted?  Or the encoding of Base is the number of entries inserted?

draft-ietf-quic-qpack.md
```diff
@@ -456,31 +458,36 @@ referenced by a given relative index will change while interpreting instructions
...
+The most recently inserted entry within the limit set by the Base has a relative
+index of 0.
+
+Though new entries could be added while processing a header block, the relative
+indices of entries do not change.
```
afrind: How are new entries added while processing a header block?  Do you mean on the encoder side?  Maybe encoding would be better than processing then?

draft-ietf-quic-qpack.md
```diff
@@ -279,14 +281,14 @@ table entries which have been acknowledged, but this could mean using
 literals. Since literals make the header block larger, this can result in the
 encoder becoming blocked on congestion or flow control limits.
 
-### Largest Known Received
+### Known Received Count
 
 In order to identify which dynamic table entries can be safely used
 without a stream becoming blocked, the encoder tracks the absolute index of the
-decoder's Largest Known Received entry.
+decoder's Known Received Count entry.
```
afrind: What is the decoder's Known Received Count entry?  That doesn't really make sense to me, maybe just this sentence needs rewording.

What the encoder is tracking is the state of the decoder.  The state of either the encoder or the decoder can be uniquely identified by the number of inserts that it has performed on its dynamic table.  

Would it be helpful to call the a table's 'State' or 'Version'?  Or maybe change it from "Known Received Count" to "Decoder's Known Insert Count"?

draft-ietf-quic-qpack.md
```diff
@@ -456,31 +458,36 @@ referenced by a given relative index will change while interpreting instructions
...
-instructions on the encoder stream, relative indices are relative to the Base
-Index at the beginning of the header block (see {{header-prefix}}). The Base
-Index is an absolute index. When interpreting the rest of the frame, the entry
-identified by Base Index has a relative index of zero.  The relative indices of
-entries do not change while interpreting headers on a request or push stream.
+instructions on the encoder stream, relative indices are relative to the Base at
+the beginning of the header block (see {{header-prefix}}). The Base is encoded
+as a value relative to the Insert Count, so it can be expressed as the number of
+entries inserted. The dynamic table entries up to the Base can be referenced.
+The most recently inserted entry within the limit set by the Base has a relative
```
afrind: > The most recently inserted entry with the limit set by the Base

I think this is where calling it Base rather than Base Index becomes awkward.  Base Index tells you which entry in the table is relative index 0.

draft-ietf-quic-qpack.md
MikeBishop: In theory, you could process instructions on the encoder stream in parallel with interpreting a header block.  However, more likely this refers to entries getting inserted during encoding, and this is the wrong terminology.

draft-ietf-quic-qpack.md
martinthomson: I failed to update this text.  I will do that.

draft-ietf-quic-qpack.md
martinthomson: Yeah, I tried to fix this without changing it too much and that turned out badly.  I've a rewrite for you.

draft-ietf-quic-qpack.md
martinthomson: I was referring to parallel processing.  My implementation allows for that possibility.

draft-ietf-quic-qpack.md
```diff
@@ -279,14 +281,15 @@ table entries which have been acknowledged, but this could mean using
...
 
-### Largest Known Received
+### Known Received Count
 
-In order to identify which dynamic table entries can be safely used
-without a stream becoming blocked, the encoder tracks the absolute index of the
-decoder's Largest Known Received entry.
+In order to identify which dynamic table entries can be safely used without a
+stream becoming blocked, the encoder tracks the number of entries received by
+the decoder.  TheKnown Received Count tracks the total number of acknowledged
```
afrind: There's a missing space here

draft-ietf-quic-qpack.md
```diff
@@ -249,18 +250,19 @@ Because QUIC does not guarantee order between data on different streams, a
...
-stream is considered "blocked."  While blocked, header field data SHOULD remain
-in the blocked stream's flow control window.  When the Largest Reference is
-zero, the frame contains no references to the dynamic table and can always be
-processed immediately. A stream becomes unblocked when the greatest absolute
-index in the dynamic table becomes greater than or equal to the Largest
-Reference for all header blocks the decoder has started reading from the stream.
-If the decoder encounters a header block where the actual largest reference is
-not equal to the Largest Reference declared in the prefix, it MAY treat this as
-a stream error of type HTTP_QPACK_DECOMPRESSION_FAILED.
+Each header block contains the total number of entries added to the dynamic
```
bencebeky: Maybe "Each header block contains a Required Insert Count, the lowest possible value for the Insert Count with which the header block can be decoded.  For a header block with no references to the dynamic table, the Required Insert Count is zero."?

draft-ietf-quic-qpack.md
```diff
@@ -444,31 +447,39 @@ referenced by a given relative index will change while interpreting instructions
...
 {: title="Example Dynamic Table Indexing - Control Stream"}
 
 Because frames from request streams can be delivered out of order with
-instructions on the encoder stream, relative indices are relative to the Base
-Index at the beginning of the header block (see {{header-prefix}}). The Base
-Index is an absolute index. When interpreting the rest of the frame, the entry
-identified by Base Index has a relative index of zero.  The relative indices of
-entries do not change while interpreting headers on a request or push stream.
+instructions on the encoder stream, relative indices are relative to the Base at
+the beginning of the header block (see {{header-prefix}}). The Base is encoded
```
bencebeky: "Unlike on the encoder stream, relative indices on push and request streams are relative to the Base at the beginning of the header block (see bla).  This allows for the mapping from relative indices to absolute indices to not be affected by dynamic table insertions on the encoder stream, regardless of what order the decoder processes request stream and encoder stream instructions.", and scratch the paragraph "Though new entries..." below.  Processing order encompasses on-the-wire reordering as well as the parallel processing case, and specifically talking about "mapping" makes it unnecessary to mention blocking in this sentence.

draft-ietf-quic-qpack.md
```diff
@@ -923,10 +948,9 @@ header field name matches the header field name of an entry stored in the static
 table or the dynamic table.
 
 If the entry is in the static table, or in the dynamic table with an absolute
-index less than or equal to Base Index, this representation starts with the '01'
-two-bit pattern.  If the entry is in the dynamic table with an absolute index
-greater than Base Index, the representation starts with the '0000' four-bit
-pattern.
+index less than the Base, this representation starts with the '01' two-bit
```
bencebeky: Are you changing one-based absolute indexing to zero-based in this PR?  Then update 3.2.4 Absolute Indexing, #indexing.  Or maybe do this in a separate PR?

draft-ietf-quic-qpack.md
```diff
@@ -950,16 +974,16 @@ values that are not to be put at risk by compressing them (see Section 7.1 of
...
-reference is to the static (S=1) or dynamic (S=0) table.
+less than the Base, the header field name is represented using the relative
+index of that entry, which is represented as an integer with a 4-bit prefix (see
+Section 5.1 of [RFC7541]). The `S` bit indicates whether the reference is to the
+static (S=1) or dynamic (S=0) table.
 
 ### Literal Header Field With Post-Base Name Reference
 
-For entries in the dynamic table with an absolute index greater than Base Index,
+For entries in the dynamic table with an absolute index greater than the Base,
```
bencebeky: If you change "less than or equal to" to "less than" above, maybe "greate than" here should be changed to "greater than or equal to"?

draft-ietf-quic-qpack.md
martinthomson: Thanks, that prompted me to restructure this (massive) paragraph.

draft-ietf-quic-qpack.md
```diff
@@ -249,18 +250,19 @@ Because QUIC does not guarantee order between data on different streams, a
...
+table - the Required Insert Count - which identifies the table state necessary
+for decoding (see {{header-prefix}}). If the Required Insert Count is greater
+than the number of dynamic table entries received, the stream is considered
+"blocked."  While blocked, header field data SHOULD remain in the blocked
+stream's flow control window.  When the Required Insert Count is zero, the frame
+contains no references to the dynamic table and can always be processed
+immediately. A stream becomes unblocked when the number of entries inserted in
+the dynamic table becomes greater than or equal to the Required Insert Count for
+all header blocks the decoder has started reading from the stream.  If the
+decoder encounters a header block where the largest Absolute Index used not
```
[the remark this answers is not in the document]
martinthomson: I think that I'm OK with this as is.  Repeating the requirement is OK if that requirement is tricky.

draft-ietf-quic-qpack.md
martinthomson: Thanks.  That's better.

draft-ietf-quic-qpack.md
```diff
@@ -784,29 +798,29 @@ streams emit the headers for an HTTP request or response.
 
 ### Header Data Prefix {#header-prefix}
 
-Header data is prefixed with two integers, `Largest Reference` and `Base Index`.
+Header data is prefixed with two integers, `Required Insert Count` and `Base`.
```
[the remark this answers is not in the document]
martinthomson: Yeah, that is a good idea.  The prose here was a little light.

draft-ietf-quic-qpack.md
```diff
@@ -784,29 +798,29 @@ streams emit the headers for an HTTP request or response.
...
 
 ~~~~~~~~~~  drawing
   0   1   2   3   4   5   6   7
 +---+---+---+---+---+---+---+---+
-|     Largest Reference (8+)    |
+|   Required Insert Count (8+)  |
```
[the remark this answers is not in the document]
martinthomson: I think that I'll stick with Required Insert Count at this level.

draft-ietf-quic-qpack.md
```diff
@@ -784,29 +798,29 @@ streams emit the headers for an HTTP request or response.
...
-| S |   Delta Base Index (7+)   |
+| S |      Delta Base (7+)      |
 +---+---------------------------+
 |      Compressed Headers     ...
 +-------------------------------+
 ~~~~~~~~~~
 {:#fig-base-index title="Frame Payload"}
 
-#### Largest Reference
+#### Insert Count
```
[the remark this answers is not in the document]
martinthomson: Done.  Thanks.

draft-ietf-quic-qpack.md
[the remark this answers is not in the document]
MikeBishop: The guidance that I've previously received is that illustrations are helpful, but not normative, and the text therefore has to convey everything that's required to implement.

draft-ietf-quic-qpack.md
```diff
@@ -444,33 +452,40 @@ referenced by a given relative index will change while interpreting instructions
...
-    +---+---------------+-----------+
-    | n |      ...      |   d + 1   |  Absolute Index
-    + - +---------------+ - - - - - +
-    | 0 |      ...      | n - d - 1 |  Relative Index
-    +---+---------------+-----------+
-      ^                       |
-      |                       V
-Insertion Point         Dropping Point
+      +---+---------------+-----------+
+      | n |      ...      |     d     |  Absolute Index
```
MikeBishop: If Absolute Index is zero-based, then the Absolute Index of the most recently inserted item would be n-1.

draft-ietf-quic-qpack.md
```diff
@@ -782,31 +800,41 @@ HEADERS and PUSH_PROMISE frames on request and push streams reference the
 dynamic table in a particular state without modifying it.  Frames on these
 streams emit the headers for an HTTP request or response.
 
-### Header Data Prefix {#header-prefix}
+### Header Block Prefix {#header-prefix}
 
-Header data is prefixed with two integers, `Largest Reference` and `Base Index`.
+Each header block is prefixed with two integers.  The Required Insert Count
+encoded as an integer with an 8-bit prefix after the encoding described in
```
MikeBishop: **is** encoded, presumably.

draft-ietf-quic-qpack.md
```diff
@@ -821,62 +849,74 @@ have.  The smallest entry has empty name and value strings and has the size of
...
          CurrentWrapped += 2 * MaxEntries
 
-      LargestReference += TotalNumberOfInserts - CurrentWrapped
+      ReqInsertCount += TotalNumberOfInserts - CurrentWrapped
 ~~~
 
-TotalNumberOfInserts is the total number of inserts into the decoder's
-dynamic table.  This encoding limits the length of the prefix on
-long-lived connections.
+Where TotalNumberOfInserts is the total number of inserts into the decoder's
```
MikeBishop: I know that the actual intent is to continue the previous sentence (...using the following algorithm, where....), but this reads as a dangling phrase.  If you want the "where," maybe move the whole thing above the algorithm?

draft-ietf-quic-qpack.md
```diff
@@ -821,62 +849,74 @@ have.  The smallest entry has empty name and value strings and has the size of
...
 An encoder that produces table updates before encoding a header block might set
-Largest Reference and Base Index to the same value.  In such case, both the sign
-bit and the Delta Base Index will be set to zero.
+Required Insert Count and Base to the same value.  In such case, both the sign
+bit and the Delta Base will be set to zero.
 
 A header block that does not reference the dynamic table can use any value for
-Base Index; setting both Largest Reference and Base Index to zero is the most
+Base; setting both Required Insert Count and Delta Base to zero is the most
 efficient encoding.
```
MikeBishop: NYT but isn't it required to set a correct RIC?

draft-ietf-quic-qpack.md
```diff
@@ -444,33 +452,40 @@ referenced by a given relative index will change while interpreting instructions
...
+Unlike on the encoder stream, relative indices on push and request streams are
+relative to the Base at the beginning of the header block (see
+{{header-prefix}}). This ensures that references are stable even if the dynamic
+table is updated while decoding a header block.
+
+The Base is encoded as a value relative to the Required Insert Count. The Base
+identifies which dynamic table entries can be referenced using relative
+indexing, starting with 0 at the last entry added.
+
+Post-base references are used for entries inserted after base, starting at 0 for
```
MikeBishop: "Base" is capitalized elsewhere

draft-ietf-quic-qpack.md
```diff
@@ -444,33 +452,40 @@ referenced by a given relative index will change while interpreting instructions
...
 
 ~~~~~ drawing
-             Base Index
-                 |
-                 V
+ Required
+  Insert
+  Count      Base
+    |         |
+    V         V
```
afrind: I would still have these point at the boxes (where the numbers are) rather than the lines.

afrind: Also, not sure we need Required Insert Count in this drawing.  The section is describing relative indexing so it's not really adding anything?

draft-ietf-quic-qpack.md
```diff
@@ -782,31 +800,41 @@ HEADERS and PUSH_PROMISE frames on request and push streams reference the
...
-#### Largest Reference
 
-`Largest Reference` identifies the largest absolute dynamic index referenced in
-the block.  Blocking decoders use the Largest Reference to determine when it is
-safe to process the rest of the block.  If Largest Reference is greater than
-zero, the encoder transforms it as follows before encoding:
+#### Required Insert Count {#ric}
+
+Required Insert Count identifies the state of the dynamic table needed to
+process the header block successfully.  Blocking decoders use the Required
```
afrind: Does 'successfully' add much here?

draft-ietf-quic-qpack.md
```diff
@@ -923,10 +961,9 @@ header field name matches the header field name of an entry stored in the static
 table or the dynamic table.
 
 If the entry is in the static table, or in the dynamic table with an absolute
-index less than or equal to Base Index, this representation starts with the '01'
-two-bit pattern.  If the entry is in the dynamic table with an absolute index
-greater than Base Index, the representation starts with the '0000' four-bit
-pattern.
+index less than the Base, this representation starts with the '01' two-bit
+pattern.  If the entry is in the dynamic table with an absolute index greater
+than the Base, the representation starts with the '0000' four-bit pattern.
```
afrind: Something is off here because it doesn't say what to do if Base and the Absolute Index of the entry are equal.  If Base is the Insert Count when encoding started, then we need to use post-base indexing if the entry is greater than or equal to base.

draft-ietf-quic-qpack.md
```diff
@@ -950,16 +987,16 @@ values that are not to be put at risk by compressing them (see Section 7.1 of
...
+Section 5.1 of [RFC7541]). The `S` bit indicates whether the reference is to the
+static (S=1) or dynamic (S=0) table.
 
 ### Literal Header Field With Post-Base Name Reference
 
-For entries in the dynamic table with an absolute index greater than Base Index,
-the header field name is represented using the post-base index of that entry
-(see {{post-base-indexing}}) encoded as an integer with a 3-bit prefix.
+For entries in the dynamic table with an absolute index greater than or equal to
+the Base, the header field name is represented using the post-base index of that
```
afrind: or equal?  per above.

draft-ietf-quic-qpack.md
```diff
@@ -444,33 +452,40 @@ referenced by a given relative index will change while interpreting instructions
...
-                 |
-                 V
+ Required
+  Insert
+  Count      Base
+    |         |
+    V         V
     +---+-----+-----+-----+-------+
-    | n | n-1 | n-2 | ... |  d+1  |  Absolute Index
+    | n | n-1 | n-2 | ... |   d   |  Absolute Index
```
afrind: If the absolute index is 0 based, then shouldn't the most recent entry be 'n - 1'?.  Eg: if Insert Count is 1 then the entry's absolute index is 0.

draft-ietf-quic-qpack.md
```diff
@@ -480,21 +495,23 @@ d = count of entries dropped
...
 ~~~~~ drawing
-             Base Index
-                 |
-                 V
+             Base
+              |
+              V
     +---+-----+-----+-----+-----+
-    | n | n-1 | n-2 | ... | d+1 |  Absolute Index
+    | n | n-1 | n-2 | ... |  d  |  Absolute Index
```
afrind: Same here

draft-ietf-quic-qpack.md
```diff
@@ -708,17 +726,17 @@ become blocked, as described in {{state-synchronization}}.
 {:#fig-size-sync title="Table State Synchronize"}
```
afrind: The above table still uses 'Insert Count', which might confuse some readers now that we've defined this term above to mean something else.  What this instruction sends is a Insert Count Delta, or something?

martinthomson: Yep, that's a good idea, though it pushes the limits of what fits in the little box :)  I wanted Insert Count Increment, but that very much doesn't fit.

martinthomson: Incidentally, does this work?  https://github.com/quicwg/base-drafts/compare/insert-count...insert-count-increment?expand=1

afrind: Insert Count Increment is much more specific than Table State Synchronize.  The description is good.  Sadly, it doesn't have the same ring to it, but I suppose clarity is more important.  I don't really like Delta, Increment is better.  We could also keep TSS and just put Increment or IC Increment or something in the little box.

martinthomson: "Synchronize" is fancy word, yeah :)  But accurate >> fancy.  I'll make the proposed change.

PR #2113, opened 2018-12-11
Fix to ECN section regarding validation

gloinul:
An issue with validation was identified. The first bullet of validation is intended to check so that the ECT code point (0 or 1) are not changed to the other. However, when an ACK includes new CE marks the comparison must deduct those CE marks to not wrongly report a mismatch. 

Also fixed a sentence indicating that is unclear about what is dropped. ECN field value of not-ect could rarely result in that the IP packet to be dropped.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3043,7 +3043,8 @@ retransmission timeout due to the absence of acknowledgments from the peer (see
 {{QUIC-RECOVERY}}), or if an endpoint has reason to believe that an element on
 the network path might be corrupting ECN codepoints, the endpoint MAY cease
 setting ECT codepoints in subsequent packets. Doing so allows the connection to
-traverse network elements that drop or corrupt ECN codepoints in the IP header.
+traverse network elements that drop IP packets with ECT or CE markings or
+corrupt ECN codepoints in the IP header.
 
 
 # Packet Size {#packet-size}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3028,9 +3028,13 @@ frame is received:
   Not-ECT.
 
 If the sender does not have state to determine if a particular packet number is
-newly acknowledged or not, then the comparison SHOULD NOT be performed. By the
-end of the verification process the local reference counter must have be updated
-to include the newly acknowledged packets.
+newly acknowledged or not, then the verifications using this acknowledgement
+cannot be performed. If an acknowledgement arrive with a packet number that
+isn't the highest to arrived so far, i.e. some reordering has occured, then the
+verification may fail erronously. To prevent this the sender MUST NOT perform
+ECN verification using older acknwoeldgements. By the end of the verification
+process the local reference counter must have be updated to include the newly
+acknowledged packets.
 
 An endpoint could miss acknowledgements for a packet when ACK frames are lost.
 It is therefore possible for the total increase in ECT(0), ECT(1), and CE
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3019,18 +3019,18 @@ frame is received:
 
 * The increase in ECT(0) and ECT(1) counters MUST be at least the number of QUIC
   packets newly acknowledged that were sent with the corresponding codepoint
-  minus the increase in the CE counter. This detects network remarking
-  between ECT(0) and ECT(1).
+  minus the increase in the CE counter. This detects network remarking between
+  ECT(0) and ECT(1).
 
 * The total increase in ECT(0), ECT(1), and CE counters reported in the ACK
   frame MUST be at least the total number of QUIC packets newly acknowledged in
-  this ACK frame. This detects if the network remark ECT(0), ECT(1) or CE to
+  this ACK frame. This detects if the network changes ECT(0), ECT(1) or CE to
   Not-ECT.
 
-If a sender receives an ACK that contains no new acknowledgments, for example
-due to reordering of the ACKs, then ECN counter comparison SHOULD NOT be
-performed. If the sender does not have state to determine if a particular PSN
-is newly acknowledged or not, then the comparison SHOULD NOT be performed.
+If the sender does not have state to determine if a particular packet number is
+newly acknowledged or not, then the comparison SHOULD NOT be performed. By the
+end of the verification process the local reference counter must have be updated
+to include the newly acknowledged packets.
 
 An endpoint could miss acknowledgements for a packet when ACK frames are lost.
 It is therefore possible for the total increase in ECT(0), ECT(1), and CE
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3019,11 +3019,13 @@ frame is received:
 
 * The increase in ECT(0) and ECT(1) counters MUST be at least the number of QUIC
   packets newly acknowledged that were sent with the corresponding codepoint
-  minus the increase in the CE counter.
+  minus the increase in the CE counter. Detects network remarking
+  between ECT(0) and ECT(1).
 
 * The total increase in ECT(0), ECT(1), and CE counters reported in the ACK
   frame MUST be at least the total number of QUIC packets newly acknowledged in
-  this ACK frame.
+  this ACK frame. Detects if the network remark ECT(0), ECT(1) or CE to
+  Not-ECT.
 
 An endpoint could miss acknowledgements for a packet when ACK frames are lost.
 It is therefore possible for the total increase in ECT(0), ECT(1), and CE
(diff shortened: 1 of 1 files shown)


mirjak:
I think it could actually be good to at one short sentence to each bulletpoint to explain what the test is checking for; e.g. as you said on slack "This is to ensure that packets sent as ECT(0) are not arriving as ECT(1)."

gloinul:
Yes, I think that is reasonable to clarify the intention here. I will add that.

Ralith:
I think the verification procedure defined here still spuriously fails when the receiver sends a pair of ACK such that each ACK newly acknowledges at least one packet not acknowledged by the other (for example, due to an implementation aggressively limiting the number of ACK blocks sent), and the ACKs are reordered such when the sender receives them, it sees the one with higher ECN counters first. By my reading, the apparent reduction in ECN would cause verification to fail.

Ignoring ECN blocks from ACKs whose largest acked PN is not higher than that of every previous ACK would resolve this problem by ensuring that reordered ECN blocks are always ignored, and subsumes the "no new acknowledgements" test.

draft-ietf-quic-transport.md
```diff
@@ -3018,11 +3018,19 @@ elements on the network path, an endpoint verifies the following when an ACK
...
 * The total increase in ECT(0), ECT(1), and CE counters reported in the ACK
   frame MUST be at least the total number of QUIC packets newly acknowledged in
-  this ACK frame.
+  this ACK frame. Detects if the network remark ECT(0), ECT(1) or CE to
+  Not-ECT.
+
+If a sender receives an ACK that contains no new acknowledgments, for example
+due to reordering of the ACKs, then ECN counter comparison SHOULD NOT be
+performed. Also if sender do not have state to determine if a particular PSN
+is newly acknowledge or not, then the comparison SHOULD NOT be performed.
```
Ralith: The implication here seems to be that an implementation should retain extra state to distinguish between a packet that is deemed lost but not yet acknowledged and a packet that was previously acknowledged. Is that intended?

If these two cases are instead conflated, e.g. by forgetting about sent packets once either acknowledged or deemed lost, then at worst the number of newly acknowledged packets in an ACK would sometimes be under-counted, and a verification with a higher rate of false positives seems preferable to going to extra effort to decide to skip verification outright.

draft-ietf-quic-transport.md
```diff
@@ -3019,18 +3019,18 @@ frame is received:
...
-  this ACK frame. Detects if the network remark ECT(0), ECT(1) or CE to
+  this ACK frame. This detects if the network remark ECT(0), ECT(1) or CE to
   Not-ECT.
 
 If a sender receives an ACK that contains no new acknowledgments, for example
 due to reordering of the ACKs, then ECN counter comparison SHOULD NOT be
-performed. Also if sender do not have state to determine if a particular PSN
-is newly acknowledge or not, then the comparison SHOULD NOT be performed.
+performed. If the sender does not have state to determine if a particular PSN
+is newly acknowledged or not, then the comparison SHOULD NOT be performed.
```
mikkelfj: Please explain PSN - is it Packet Serial Number? In that case "packet", "packet number", or "PN" is more aligned with the rest of the text, I suppose.

ianswett: I think it's supposed to be packet sequence number?  But the text always uses packet number and abbreviates it to PN, so if that's what it is, it should be changed.

draft-ietf-quic-transport.md
gloinul: Partly, sender need to track if a packet has been acknowledged or not. But, my assumption is that the sender will have a window where such information will be discarded when likely not relevant anymore. So if everything has been ACKed up to PSN=48. Then cleaning up your state and discarding all below 48 is fine. When the reorderd ACK that indicate reception of PSN 44 and 45 can then be determined to be behind this window and ignored. The second sentence is trying to state that for this example case this is fine to do.

draft-ietf-quic-transport.md
gloinul: Sorry, yes I will try to use PN in my comments.

gloinul: This uses packet number now.

draft-ietf-quic-transport.md
```diff
@@ -3018,11 +3018,19 @@ elements on the network path, an endpoint verifies the following when an ACK
...
+  ECT(0) and ECT(1).
 
 * The total increase in ECT(0), ECT(1), and CE counters reported in the ACK
   frame MUST be at least the total number of QUIC packets newly acknowledged in
-  this ACK frame.
+  this ACK frame. This detects if the network changes ECT(0), ECT(1) or CE to
+  Not-ECT.
+
+If the sender does not have state to determine if a particular packet number is
+newly acknowledged or not, then the comparison SHOULD NOT be performed. By the
```
larseggert: Is this really a SHOULD NOT? If the sender doesn't have that state, it **CANNOT** perform that check. That's not a 2119 keyword, unfortunately :-)

gloinul:
Benjamin Saunders did point out that if one have ACKs with none overlapping ACK ranges and they get reordered there is still an issue with performing the comparison in some cases. Example:

Receiver sent: 
ACK:1  PN:1,2,3,4 ECT(0)=4, ECT(1)=0, CE=0
ACK:2  PN:5,6,7,8 ECT(0)=8, ECT(1)=0, CE=0
ACK:3  PN:9,10,11 ECT(0)=11, ECT(1)=0, CE=0

Sender receives:
ACK:1  PN:1,2,3,4 ECT(0)=4, ECT(1)=0, CE=0
ACK:3  PN:9,10,11 ECT(0)=11, ECT(1)=0, CE=0
ACK:2  PN:5,6,7,8 ECT(0)=8, ECT(1)=0, CE=0

Local Reference Counters: 0,0,0
1: Comparison ECT(0)_diff=4, Newly ACKed=4 (with ECT(0)) OK
Local Reference Counters: 4,0,0
3: Verification ECT(0)_diff=7, Newly ACKEd=3
Local Reference Counters: 7,0,0 (Based on ACKed packets)
Local Reference Counters: 11,0,0 (Latest Received ACK)

2: Verification ECT(0)_diff=1 or -3 (Based on ACKed, vs Latest Received).
Newly ACKed=4
Comparison fails if run.  

So the question is how to deal with this. From my perspective there are multiple implementation strategies that could solve this. So my question is should we mandate any particular or simply try to illustrate the issue. I think the most straight forward and which I don't so far sees any flaws with are:

- Don't perform verification with ACK frames that on arrivial is not highest PN seen. 

Opinions?

gloinul:
So now there is a proposal for how resolve Benjamin's comment. Please review.

draft-ietf-quic-transport.md
```diff
@@ -3018,11 +3018,23 @@ elements on the network path, an endpoint verifies the following when an ACK
...
-  this ACK frame.
+  this ACK frame. This detects if the network changes ECT(0), ECT(1) or CE to
+  Not-ECT.
+
+If the sender does not have state to determine if a particular packet number is
+newly acknowledged or not, then the verifications using this acknowledgement
+cannot be performed. If an acknowledgement arrive with a packet number that
+isn't the highest to arrived so far, i.e. some reordering has occured, then the
+verification may fail erronously. To prevent this the sender MUST NOT perform
+ECN verification using older acknwoeldgements. By the end of the verification
```
Ralith:
```suggestion
ECN verification using older acknowledgements. By the end of the verification
```

martinthomson:
Yes, we should not validate (or increase the sender-side view of where the counters are) if the largest acknowledged packet does not increase.

draft-ietf-quic-transport.md
```diff
@@ -3018,11 +3018,23 @@ elements on the network path, an endpoint verifies the following when an ACK
 frame is received:
```
martinthomson: I think that we need a little restructuring here.

> To reduce the risk of bad ECN markings affecting the operation of an endpoint, an endpoint verifies the counts it receives when it receives new acknowledgements:
> 
> * \[bullets]
>
> This validation is only performed if the ACK frame increases the largest received packet number.  Reordered acknowledgments could have lower counter values and might not be successfully validated as a result.
>
> These counts might be inflated if acknowledgments are never received for packets that were successfully delivered.  If validation succeeds, an endpoint MUST increase its expected counter values to those it receives.

Note that I almost didn't use MUST on that last clause, because there is no interop loss from doing that; it only means that validation is less likely to catch bad markings.

I also removed the "If the sender does not have state to determine if a packet is newly acknowledged..."  Such a packet - such as a packet that is declared lost - will never be considered to be newly acknowledged.

Would you like this as a PR?

gloinul: I think the restructuring is fine. On a high level I agree with your removal of the if you are missing state. However, considering some of the discussion on the quicdev slack I do wonder if something along those lines needs to be included to ensure that people are not interpreting this as requiring one to maintain state further into history than what one otherwise would.

martinthomson: Thanks for that.

The pseudocode in the recovery draft drops packets from the "in flight" structure when they are declared lost and only treats packets as "newly acknowledged" if they are in that structure.  I think that makes it clear enough.

Issue #1017, opened 2017-12-15
Eliminating the fixed minimum RTO


janaiyengar:
TL;DR: QUIC currently uses a minimum RTO of 200ms. Given that (i) the cost of a spurious RTO is substantially reduced in QUIC, (ii) we have the ability to track the receiver's delayed ack timeout as discussed in #981, and (iii) QUIC is strongly leaning towards being latency-sensitive, we should consider eliminating the fixed minimum RTO for QUIC in its RTO computations. Argument follows.

[RFC 6298](https://tools.ietf.org/html/rfc6298) for TCP conservatively uses a 1 second minimum RTO, and the QUIC recovery draft uses 200 ms since that's what Linux TCP does. A fixed minimum RTO is used in TCP (and by inheritance, in QUIC) to protect a sender from the cost of spurious RTOs. See [RFC 6298, rule 2.4 in section 2](https://tools.ietf.org/html/rfc6298#section-2): "Research suggests that a large minimum RTO is needed to keep TCP conservative and avoid spurious retransmissions [AP99]".

Spurious RTOs can occur due to several reasons, for instance (i) delayed ack timeout at the receiver is not always sampled in the RTT measurements, so minRTO provides protection at the sender, or (ii) the RTO computation uses EWMAs, which always lag, causing spurious RTOs when the underlying network RTT increases. 

AIUI, the primary reason to avoid spurious RTOs in RFC 6298 is the cost of an RTO firing, which hurts sender performance  by reducing the congestion window to 1 MSS. (There are secondary costs of additional load on the network, but that's not debilitating, since TLP demonstrates that this load is manageable.) 

QUIC's loss recovery does not suffer this problem because the sender handles spurious RTO detection as part of its loss recovery logic. Specifically, QUIC does not reduce the congestion window to 1 on an RTO, but reduces it if the RTO is verified to be real, when an ACK is received for the retransmission but not the transmission. Alternatively, if an ACK for an earlier packet is received after an RTO fires, the RTO is considered spurious and does not cause the sender to reduce its congestion window. See [Section 3.3.2 of the draft-ietf-quic-recovery]((https://quicwg.github.io/base-drafts/draft-ietf-quic-recovery.html#rfc.section.3.3.2)).

[...]

PR #2114, opened 2018-12-11
Unify TLP and RTO into Probe Timeout

### LINKED ISSUE #2166: TLP and RTO arbitrariness
There is some arbitrariness around TLP, RTO, and crypto timeouts  as they are described in the draft. 

- TLP computation includes smoothed_rtt but not rttvar, where RTO includes both. There's no clear rationale for not using rttvar in TLP.  This is what the TLP draft (individual, expired) says and what Linux TCP implements, but there's little justification beyond that. rttvar should be included here.
- minTLP is 10ms and minRTO is 200ms. This choice of 10ms is a bit arbitrary, but is based on what Linux TCP implements.
- The crypto timeout uses minTLP, which does not seem intuitive.

We should include rttvar in TLP, and we should either justify the 10ms min timeout or use the same 200ms timeout for TLP.  We should also use a different min for the crypto timeout than min TLP.

janaiyengar:
Closes #2166, #2168, #1017.

This PR builds on #1974 and simplifies QUIC's timeout-based loss detection mechanisms. Unfortunately, this PR cuts through text in many places, so I'm summarizing the key changes here.  The compiled version of the draft, [here](https://quicwg.org/base-drafts/pto/draft-ietf-quic-recovery.html), might be easier to read than the diffs.

The high-order description is that TLP and RTO are combined in this PR into a single "Probe Timeout" (PTO). Consecutive PTOs are exponentially longer, just like RTO used to be.  The rationale for calling this a "Probe" Timeout is that loss detection in QUIC only happens when an ack is received, and this timer and timeout only cause probe packets to be sent out to elicit acks. This unification is made possible by the following constituent changes.

- (Issue #1017) There was general agreement in #1017 that adding a max_ack_delay should allow us to remove the minRTO. max_ack_delay has since been added to QUIC in a transport parameter, which allows us to include it in the PTO computation and remove minRTO/minTLP.

- Timer granularity is added for both crypto and PTO timeouts, ensuring that even when the computed network RTT might be too small and represented as 0, timeouts are set for at least 1 clock tick. This is in keeping with use of granularity in TCP (RFC 6298). A min granularity of 1ms is recommended.

- (Issue #2166) In line with the resolution above for removing min RTO, min TLP and min crypto timeout values are now removed.

- (Issue #2168) PTO validation has been removed.  PTOs cause congestion window collapse to min cwnd only after 3 consecutive PTOs to retain parity with current behavior, where an RTO collapses the window after 2 TLPs.

One side-effect to note is that a single timer will start exponentially backing off right from the first PTO, as against TLP and RTO where the TLP timers did not back off at all, and the RTO timer started exponentially backing off after the TLPs. I can restore this by adjusting the backoff expression in SetLossDetectionTimer(), but I'm not convinced this is necessary.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -371,8 +371,7 @@ increases loss detection delay.
 
 Timeout loss detection recovers from losses that cannot be handled by
 ack-based loss detection.  It uses a single timer which switches between
-a crypto retransmission timer, a Tail Loss Probe timer and Retransmission
-Timeout mechanisms.
+a crypto retransmission timer and a Probe timer.
 
 ### Crypto Retransmission Timeout
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -433,7 +433,7 @@ flight but an acknowledgement is not received within the expected period of
 time.  A PTO enables a connection to recover from loss of tail packets or acks.
 The PTO algorithm used in QUIC implements the reliability functions of Tail Loss
 Probe {{?TLP=I-D.dukkipati-tcpm-tcp-loss-probe}}, RTO {{?RFC5681}} and F-RTO
-algorithms for TCP {{?RFC5682}}, and the computation is based on TCP's
+algorithms for TCP {{?RFC5682}}, and the timeout computation is based on TCP's
 retransmission timeout period {{?RFC6298}}.
 
 #### Computing PTO
@@ -475,14 +475,15 @@ consecutive PTO expiration due to a single packet loss.
 Consecutive PTO periods increase exponentially, and as a result, connection
 recovery latency increases exponentially as packets continue to be dropped in
 the network.  Sending two packets on PTO expiration increases resilience to
-packet drop, thus reducing the probability of consecutive PTO events.
+packet drops, thus reducing the probability of consecutive PTO events.
 
 Probe packets sent on a PTO MUST be ack-eliciting.  A probe packet SHOULD carry
 new data when possible.  A probe packet MAY carry retransmitted unacknowledged
 data when new data is unavailable, when flow control does not permit new data to
-be sent, or to opportunistically reduce loss recovery delay.  Implementers MAY
-use alternate strategies for determining the content of probe packets, including
-sending new or retransmitted data based on the application's priorities.
+be sent, or to opportunistically reduce loss recovery delay.  Implementations
+MAY use alternate strategies for determining the content of probe packets,
+including sending new or retransmitted data based on the application's
+priorities.
 
 
 #### Loss Detection {#pto-loss}
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -338,8 +338,8 @@ The RECOMMENDED initial value for the packet reordering threshold
 {{?RFC5681}} {{?RFC6675}}.
 
 Some networks may exhibit higher degrees of reordering, causing a sender to
-detect spurious losses.  Implementers MAY increase the packet threshold under
-such conditions.
+detect spurious losses.  Implementers MAY use algorithms developed for TCP, such
+as TCP-NCR {{?RFC4653}}, to improve QUIC's reordering resilience.
 
 ### Time Threshold {#time-threshold}
 
@@ -352,7 +352,7 @@ lost, then a timer SHOULD be set for the remaining time.
 The RECOMMENDED time threshold (kTimeThreshold), expressed as a round-trip time
 multiplier, is 9/8.
 
-Using max(smoothed_rtt, latest_rtt) protects from the two following cases:
+Using max(SRTT, latest_RTT) protects from the two following cases:
 
 * the latest RTT sample is lower than the SRTT, perhaps due to reordering where
   packet whose ack triggered the Early Retransmit process encountered a shorter
@@ -371,7 +371,7 @@ increases loss detection delay.
 
 Timeout loss detection recovers from losses that cannot be handled by
 ack-based loss detection.  It uses a single timer which switches between
-a crypto retransmission timer and a Probe timer.
+a crypto retransmission timer and a probe timer.
 
 ### Crypto Retransmission Timeout
 
@@ -428,48 +428,47 @@ a subsequent connection attempt to the server.
 ### Probe Timeout {#pto}
 
 A Probe Timeout (PTO) triggers a probe packet when ack-eliciting data is in
-flight but an acknowledgement does not seem forthcoming.  A PTO enables a
-connection to recover from loss of tail packets or acks, where acks of
-subsequent packets are not available to trigger ack-based loss detection.  The
-PTO algorithm used in QUIC implements the reliability functions of Tail Loss
-Probe {{?TLP=I-D.dukkipati-tcpm-tcp-loss-probe}}, RTO {{?RFC5681}} and F-RTO
-algorithms for TCP {{?RFC5682}}, and the computation is based on TCP's
+flight but an acknowledgement is not received within the expected period of
+time.  A PTO enables a connection to recover from loss of tail packets or acks,
+where acks of subsequent packets are not available to trigger ack-based loss
+detection.  The PTO algorithm used in QUIC implements the reliability functions
+of Tail Loss Probe {{?TLP=I-D.dukkipati-tcpm-tcp-loss-probe}}, RTO {{?RFC5681}}
+and F-RTO algorithms for TCP {{?RFC5682}}, and the computation is based on TCP's
 retransmission timeout period {{?RFC6298}}.
 
 #### Computing PTO
 
-When the final ack-eliciting packet before quiescence is transmitted, the sender
-schedules a timer for the PTO period as follows:
+When an ack-eliciting packet is transmitted, the sender schedules a timer for
+the PTO period as follows:
 
 PTO = max(smoothed_rtt + 4*rttvar + max_ack_delay, kGranularity)
 
 kGranularity, smoothed_rtt, rttvar, and max_ack_delay are defined in
-{{ld-vars-of-interest}}.
+{{ld-consts-of-interest}} and {{ld-vars-of-interest}}.
 
 The PTO period is the amount of time that a sender ought to wait for an
-acknowledgement for a sent packet to be received.  This time period includes the
-estimated network roundtrip-time (smoothed_rtt), the variance in the estimate
-(4*rttvar), and max_ack_delay, to account for the maximum time by which a
-receiver might delay sending an acknowledgement.
+acknowledgement of a sent packet.  This time period includes the estimated
+network roundtrip-time (smoothed_rtt), the variance in the estimate (4*rttvar),
+and max_ack_delay, to account for the maximum time by which a receiver might
+delay sending an acknowledgement.
 
-There is no requirement on the clock granularity. If the PTO computation results
-in a value of zero, a sender MUST set the PTO value to kGranularity, to avoid
-the timer expiring immediately.
-
-A PTO timer is set on an ack-eliciting tail packet.  A sender may not know that
-a packet being sent is a tail packet and may have to adjust the timer every time
-an ack-eliciting packet is sent.
+There is no requirement on the clock granularity. The PTO value MUST be set to
+at least kGranularity, to avoid the timer expiring immediately.
 
 When a PTO timer expires, the PTO period MUST be set to twice its current value.
 This exponential reduction in the sender's rate is important because the PTOs
 might be caused by loss of packets or acknowledgements due to severe congestion.
 
+A sender computes its PTO timer every time an ack-eliciting packet is sent. A
+sender might choose to optimize this by setting the timer fewer times if it
+knows that more ack-eliciting packets will be sent within a short period of
+time.
 
 #### Sending Probe Packets
 
 When a PTO timer expires, the sender MUST send one ack-eliciting packet as a
 probe. A sender MAY send up to two ack-eliciting packets, to avoid an expensive
-consecutive PTO expiration due to packet loss.
+consecutive PTO expiration due to a single packet loss.
 
 Consecutive PTO periods increase exponentially, and as a result, connection
 recovery latency increases exponentially as packets continue to be dropped in
@@ -478,13 +477,11 @@ packet drop in both directions, thus reducing the probability of consecutive PTO
 events.
 
 Probe packets sent on a PTO MUST be ack-eliciting.  A probe packet SHOULD carry
-new data when possible.  Implementers MAY use alternate strategies for
-determining the content of probe packets.  An implementation could use new data
-in probe packets to maximize throughput.  An implementation could retransmit
-unacknowledged data when new data is unavailable, when flow control does not
-permit new data to be sent, or to opportunistically reduce loss recovery delay.
-Alternatively, an implementation could send new or retransmitted data based on
-the application's priorities.
+new data when possible.  A probe packet MAY carry retransmitted unacknowledged
+data when new data is unavailable, when flow control does not permit new data to
+be sent, or to opportunistically reduce loss recovery delay.  Implementers MAY
+use alternate strategies for determining the content of probe packets, including
+sending new or retransmitted data based on the application's priorities.
 
 
 #### Loss Detection {#pto-loss}
@@ -554,7 +551,7 @@ time:
 
 ## Pseudocode
 
-### Constants of interest
+### Constants of interest {#ld-consts-of-interest}
 
 Constants used in loss recovery are based on a combination of RFCs, papers, and
 common practice.  Some may need to be changed or negotiated in order to better
@@ -652,7 +649,7 @@ follows:
    smoothed_rtt = 0
    rttvar = 0
    min_rtt = infinite
-   largest_sent_before_rto = 0
+   largest_sent_before_pto = 0
    time_of_last_sent_ack_eliciting_packet = 0
    time_of_last_sent_crypto_packet = 0
    largest_sent_packet = 0
@@ -706,10 +703,10 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
       // Find the smallest newly acknowledged packet
       smallest_newly_acked =
         FindSmallestNewlyAcked(newly_acked_packets)
-      // If any packets sent prior to RTO were acked, then the
-      // RTO was spurious. Otherwise, inform congestion control.
-      if (rto_count > 0 &&
-            smallest_newly_acked > largest_sent_before_rto):
+      // If any packets sent prior to PTO were acked, then the
+      // PTO was spurious. Otherwise, inform congestion control.
+      if (pto_count > 0 &&
+            smallest_newly_acked > largest_sent_before_pto):
         OnProbeTimeoutVerified(smallest_newly_acked)
       crypto_count = 0
       pto_count = 0
@@ -795,7 +792,7 @@ Pseudocode for SetLossDetectionTimer follows:
       loss_detection_timer.set(loss_time)
       return
 
-    // PTO timer. Calculate PTO duration
+    // Calculate PTO duration
     timeout =
       smoothed_rtt + 4 * rttvar + max_ack_delay
     timeout = max(timeout, granularity)
@@ -920,7 +917,7 @@ experiment with other response functions.
 QUIC begins every connection in slow start and exits slow start upon loss or
 upon increase in the ECN-CE counter. QUIC re-enters slow start anytime the
 congestion window is less than ssthresh, which typically only occurs after an
-RTO. While in slow start, QUIC increases the congestion window by the number of
+PTO. While in slow start, QUIC increases the congestion window by the number of
 bytes acknowledged when each acknowledgment is processed.
 
 
@@ -951,9 +948,9 @@ losses or increases in the ECN-CE counter.
 Probe packets MUST NOT be blocked by the congestion controller.  A sender MUST
 however count these packets as being additionally in flight, since these packets
 adds network load without establishing packet loss.  Note that sending probe
-packets might cause the sender's estimated bytes in flight to exceed the
-sender's congestion window until an acknowledgement is received that establishes
-loss or delivery of packets.
+packets might cause the sender's bytes in flight to exceed the congestion window
+until an acknowledgement is received that establishes loss or delivery of
+packets.
 
 A PTO expiration is classified as spurious or valid when an ACK frame is
 received that newly acknowledges packets in flight, see {{pto-loss}}.  On a
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -768,7 +768,7 @@ Pseudocode for SetLossDetectionTimer follows:
         timeout = 2 * kInitialRtt
       else:
         timeout = 2 * smoothed_rtt
-      timeout = max(timeout, granularity)
+      timeout = max(timeout, kGranularity)
       timeout = timeout * (2 ^ crypto_count)
       loss_detection_timer.set(
         time_of_last_sent_crypto_packet + timeout)
@@ -781,7 +781,7 @@ Pseudocode for SetLossDetectionTimer follows:
     // Calculate PTO duration
     timeout =
       smoothed_rtt + 4 * rttvar + max_ack_delay
-    timeout = max(timeout, granularity)
+    timeout = max(timeout, kGranularity)
     timeout = timeout * (2 ^ pto_count)
 
     loss_detection_timer.set(
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -432,9 +432,9 @@ A Probe Timeout (PTO) triggers a probe packet when ack-eliciting data is in
 flight but an acknowledgement is not received within the expected period of
 time.  A PTO enables a connection to recover from loss of tail packets or acks.
 The PTO algorithm used in QUIC implements the reliability functions of Tail Loss
-Probe {{?TLP=I-D.dukkipati-tcpm-tcp-loss-probe}}, RTO {{?RFC5681}} and F-RTO
-algorithms for TCP {{?RFC5682}}, and the timeout computation is based on TCP's
-retransmission timeout period {{?RFC6298}}.
+Probe {{?TLP=I-D.dukkipati-tcpm-tcp-loss-probe}} {{?RACK}}, RTO {{?RFC5681}} and
+F-RTO algorithms for TCP {{?RFC5682}}, and the timeout computation is based on
+TCP's retransmission timeout period {{?RFC6298}}.
 
 #### Computing PTO
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -409,7 +409,7 @@ On each consecutive expiration of the crypto timer without receiving an
 acknowledgement for a new packet, the sender SHOULD double the crypto
 retransmission timeout and set a timer for this period.
 
-When crypto packets are in flight, the TLP and RTO timers are not active.
+When crypto packets are in flight, the PTO timer is not active.
 
 #### Retry and Version Negotiation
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -726,10 +700,9 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
       // RTO was spurious. Otherwise, inform congestion control.
       if (rto_count > 0 &&
             smallest_newly_acked > largest_sent_before_rto):
```
siyengar: largest_sent_before_rto is still used here

draft-ietf-quic-recovery.md
```diff
@@ -1183,14 +1142,14 @@ are detected lost.
...
-QUIC decreases the congestion window to the minimum value once the
-retransmission timeout has been verified and removes any packets
-sent before the newly acknowledged RTO packet.
+QUIC decreases the congestion window to the minimum value once the probe timeout
+has been verified and removes any packets sent before the newly acknowledged
+packet.
 
 ~~~
-   OnRetransmissionTimeoutVerified(packet_number)
+   OnProbeTimeoutVerified(packet_number)
```
siyengar: This looks like we will set the probetimeout verified even if this happens

    1, 2, 3 [lost], 4 [probe], 5 [probe]
    -> [ack 4]
    -> cwnd = mincwnd

in the first probe case it seems like the probe would be aggressive so it would be bad to drop the cwnd down.

even if it not aggressive right now, in the future I think we are going to tune timeouts constant per application (realtime vs interactive). This means that if we tune timeout constants too aggressively, it will result in having this reaction.

draft-ietf-quic-recovery.md
```diff
@@ -361,8 +361,7 @@ increases loss detection delay.
 
 Timeout loss detection recovers from losses that cannot be handled by
 ack-based loss detection.  It uses a single timer which switches between
-a crypto retransmission timer, a Tail Loss Probe timer and Retransmission
-Timeout mechanisms.
+a crypto retransmission timer and a Probe timer.
```
nibanks: You don't capitalize `Probe` in other places. It looks out of place here.

draft-ietf-quic-recovery.md
nibanks: As well as `rto_count`

draft-ietf-quic-recovery.md
```diff
@@ -812,17 +785,11 @@ Pseudocode for SetLossDetectionTimer follows:
       loss_detection_timer.set(loss_time)
       return
 
-    // RTO or TLP timer
-    // Calculate RTO duration
+    // PTO timer. Calculate PTO duration
```
nibanks: Nit: I feel like having either `PTO timer` or `Calculate PTO duration` would be cleaner. Having both seems a bit overdone.

draft-ietf-quic-recovery.md
```diff
@@ -973,24 +936,20 @@ During recovery, the congestion window remains unchanged irrespective of new
...
+adds network load without establishing packet loss.  Note that sending probe
+packets might cause the sender's estimated bytes in flight to exceed the
+sender's congestion window until an acknowledgement is received that establishes
+loss or delivery of packets.
 
-## Retransmission Timeout
+A PTO expiration is classified as spurious or valid when an ACK frame is
+received that newly acknowledges packets in flight, see {{pto-loss}}.  On a
+valid PTO, the congestion window MUST be reduced to the minimum congestion
+window and slow start is re-entered.
```
nibanks: Just to confirm my understanding, this is a difference from TLP before, right? Before this change, if TLP triggered and you sent a probe packet that elicited an ACK for just that packet, you wouldn't go back into slow start, but now an ACK for only the first probe packet would trigger slow start. Correct?

draft-ietf-quic-recovery.md
```diff
@@ -328,8 +328,8 @@ The RECOMMENDED initial value for the packet reordering threshold
 (kPacketThreshold) is 3, based on TCP loss detection {{?RFC5681}} {{?RFC6675}}.
 
 Some networks may exhibit higher degrees of reordering, causing a sender to
-detect spurious losses.  Implementers MAY use algorithms developed for TCP, such
-as TCP-NCR {{?RFC4653}}, to improve QUIC's reordering resilience.
+detect spurious losses.  Implementers MAY increase the packet threshold under
+such conditions.
```
ianswett: This change seems unrelated to the rest of the change.

draft-ietf-quic-recovery.md
```diff
@@ -415,100 +414,88 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
+Probe {{?TLP=I-D.dukkipati-tcpm-tcp-loss-probe}}, RTO {{?RFC5681}} and F-RTO
+algorithms for TCP {{?RFC5682}}, and the computation is based on TCP's
+retransmission timeout period {{?RFC6298}}.
 
-The timer duration, or Probe Timeout (PTO), is set based on the following
-conditions:
+#### Computing PTO
 
-* PTO SHOULD be scheduled for max(1.5*SRTT+MaxAckDelay, kMinTLPTimeout)
+When the final ack-eliciting packet before quiescence is transmitted, the sender
```
ianswett: Final is an optimization and we don't define quiescence
```suggestion
When an ack-eliciting packet is sent, the sender
```

draft-ietf-quic-recovery.md
```diff
@@ -577,21 +560,18 @@ kTimeThreshold:
   considers a packet lost. Specified as an RTT multiplier. The RECOMMENDED
   value is 9/8.
 
-kMinTLPTimeout:
-: Minimum time in the future a tail loss probe timer may be set for.
-  The RECOMMENDED value is 10ms.
+kGranularity:
```
ianswett: It feels odd to define this here, but use it three times up above.  I'd suggest discussing the idea that you're using clock granularity above where you first use it.

draft-ietf-quic-recovery.md
```diff
@@ -415,100 +414,88 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
-QUIC includes MaxAckDelay in all probe timeouts, because it assumes the ack
-delay may come into play, regardless of the number of packets outstanding.
-TCP's TLP assumes if at least 2 packets are outstanding, acks will not be
-delayed.
+kGranularity, smoothed_rtt, rttvar, and max_ack_delay are defined in
+{{ld-vars-of-interest}}.
 
-A PTO value of at least 1.5*SRTT ensures that the ACK is overdue.  The 1.5 is
-based on {{?TLP}}, but implementations MAY experiment with other constants.
+The PTO period is the amount of time that a sender ought to wait for an
```
```suggestion
The PTO period is the amount of time that a sender SHOULD wait for an
```

draft-ietf-quic-recovery.md
```diff
@@ -415,100 +414,88 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
+acknowledgement for a sent packet to be received.  This time period includes the
```
```suggestion
acknowledgement of a sent packet.  This time period includes the
```

draft-ietf-quic-recovery.md
```diff
@@ -415,100 +414,88 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
+estimated network roundtrip-time (smoothed_rtt), the variance in the estimate
```
ianswett: Using variance here is potentially confusing, since it's not true variance and it's 4*rttvar.

I'm having a hard time coming up with suggestions, but I'll think about it.

draft-ietf-quic-recovery.md
```diff
@@ -415,100 +414,88 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
+(4*rttvar), and max_ack_delay, to account for the maximum time by which a
+receiver might delay sending an acknowledgement.
 
-To reduce latency, it is RECOMMENDED that the sender set and allow the TLP timer
-to fire twice before setting an RTO timer. In other words, when the TLP timer
-expires the first time, a TLP packet is sent, and it is RECOMMENDED that the TLP
-timer be scheduled for a second time. When the TLP timer expires the second
-time, a second TLP packet is sent, and an RTO timer SHOULD be scheduled {{rto}}.
+There is no requirement on the clock granularity. If the PTO computation results
+in a value of zero, a sender MUST set the PTO value to kGranularity, to avoid
```
ianswett: How about "The PTO must be set to at least kGranularity,"

draft-ietf-quic-recovery.md
```diff
@@ -415,100 +414,88 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
+the timer expiring immediately.
 
-A TLP packet SHOULD carry new data when possible. If new data is unavailable or
-new data cannot be sent due to flow control, a TLP packet MAY retransmit
-unacknowledged data to potentially reduce recovery time. Since a TLP timer is
-used to send a probe into the network prior to establishing any packet loss,
-prior unacknowledged packets SHOULD NOT be marked as lost when a TLP timer
-expires.
+A PTO timer is set on an ack-eliciting tail packet.  A sender may not know that
```
ianswett: Drop this first sentence and move the second sentence up with "When the final ack-eliciting packet"

draft-ietf-quic-recovery.md
```diff
@@ -415,100 +414,88 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
-When the last TLP packet is sent, a timer is set for the RTO period. When
-this timer expires, the sender sends two packets, to evoke acknowledgements from
-the receiver, and restarts the RTO timer.
+#### Sending Probe Packets
 
-Similar to TCP {{?RFC6298}}, the RTO period is set based on the following
-conditions:
+When a PTO timer expires, the sender MUST send one ack-eliciting packet as a
+probe. A sender MAY send up to two ack-eliciting packets, to avoid an expensive
+consecutive PTO expiration due to packet loss.
```
```suggestion
consecutive PTO expiration due to a single packet loss.
```

draft-ietf-quic-recovery.md
```diff
@@ -415,100 +414,88 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
+Consecutive PTO periods increase exponentially, and as a result, connection
+recovery latency increases exponentially as packets continue to be dropped in
+the network.  Sending two packets on PTO expiration increases resilience to
+packet drop in both directions, thus reducing the probability of consecutive PTO
+events.
 
-* When an RTO timer expires, the RTO period is doubled.
+Probe packets sent on a PTO MUST be ack-eliciting.  A probe packet SHOULD carry
+new data when possible.  Implementers MAY use alternate strategies for
+determining the content of probe packets.  An implementation could use new data
```
ianswett: From "An implementation could" on, this paragraph seems quite speculative, and the first sentence is redundant to the above setence "A probe packet SHOULD carry new data when possible."

draft-ietf-quic-recovery.md
```diff
@@ -415,100 +414,88 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
+the first of one or more consecutive PTO expirations, the PTOs are considered
+spurious.  Loss detection proceeds as dictated by packet and time threshold
+mechanisms, see {{ack-loss-detection}}.
 
-A packet sent on an RTO timer MUST NOT be blocked by the sender's congestion
-controller. A sender MUST however count these packets as being in flight, since
-this packet adds network load without establishing packet loss.
+If an ACK frame is received that newly acknowledges only those packets that were
+sent after a PTO timer expiration, all unacknowledged packets with lower packet
+numbers MUST be marked as lost.
```
ianswett: This will happen naturally due to loss detection I believe.

draft-ietf-quic-recovery.md
```diff
@@ -973,24 +936,20 @@ During recovery, the congestion window remains unchanged irrespective of new
...
-A TLP packet MUST NOT be blocked by the sender's congestion controller. The
-sender MUST however count TLP packets against bytes in flight, since a TLP adds
-network load without establishing packet loss.
+## Probe Timeout
 
-Acknowledgement or loss of tail loss probes are treated like any other packet.
+Probe packets MUST NOT be blocked by the congestion controller.  A sender MUST
+however count these packets as being additionally in flight, since these packets
+adds network load without establishing packet loss.  Note that sending probe
+packets might cause the sender's estimated bytes in flight to exceed the
```
```suggestion
packets might cause the sender's bytes in flight to exceed the
```

draft-ietf-quic-recovery.md
```diff
@@ -973,24 +936,20 @@ During recovery, the congestion window remains unchanged irrespective of new
...
+sender's congestion window until an acknowledgement is received that establishes
```
```suggestion
congestion window until an acknowledgement is received that establishes
```

ianswett: Also, I think this would read better as "until packets are acknowledged or lost."

draft-ietf-quic-recovery.md
ianswett: Agreed, this is a potentially substantial change.

draft-ietf-quic-recovery.md
ianswett: I'd suggest we keep a "you can not do the CWND reduction on the first or 2nd PTO" to keep behavior more similar to what QUIC has now.

draft-ietf-quic-recovery.md
```diff
@@ -297,7 +297,7 @@ If a packet is lost, the QUIC transport needs to recover from that loss, such
 as by retransmitting the data, sending an updated frame, or abandoning the
 frame.  For more information, see Section 13.2 of {{QUIC-TRANSPORT}}.
 
-## Ack-based Detection
+## Ack-based Detection {#ack-loss-detection}
```
```suggestion
## ACK-based Detection {#ack-loss-detection}
```

draft-ietf-quic-recovery.md
```diff
@@ -415,100 +414,85 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
-conditions:
+#### Computing PTO
 
-* PTO SHOULD be scheduled for max(1.5*SRTT+MaxAckDelay, kMinTLPTimeout)
+When an ack-eliciting packet is transmitted, the sender schedules a timer for
+the PTO period as follows:
 
-* If RTO ({{rto}}) is earlier, schedule a TLP in its place. That is,
-  PTO SHOULD be scheduled for min(RTO, PTO).
+PTO = max(smoothed_rtt + 4*rttvar + max_ack_delay, kGranularity)
```
```suggestion
~~~
PTO = max(smoothed_rtt + 4*rttvar + max_ack_delay, kGranularity)
~~~
```

draft-ietf-quic-recovery.md
```diff
@@ -844,16 +808,12 @@ Pseudocode for OnLossDetectionTimeout follows:
...
-     else if (tlp_count < kMaxTLPs):
-       // Tail Loss Probe.
-       SendOnePacket()
-       tlp_count++
      else:
-       // RTO.
-       if (rto_count == 0)
-         largest_sent_before_rto = largest_sent_packet
+       // PTO.
+       if (pto_count == 0)
```
```suggestion
       if (pto_count == 0):
```

martinthomson:
Does this have an issue?

draft-ietf-quic-recovery.md
```diff
@@ -415,100 +414,85 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
+packet drop in both directions, thus reducing the probability of consecutive PTO
+events.
 
-* When an RTO timer expires, the RTO period is doubled.
+Probe packets sent on a PTO MUST be ack-eliciting.  A probe packet SHOULD carry
+new data when possible.  A probe packet MAY carry retransmitted unacknowledged
+data when new data is unavailable, when flow control does not permit new data to
+be sent, or to opportunistically reduce loss recovery delay.  Implementers MAY
+use alternate strategies for determining the content of probe packets, including
+sending new or retransmitted data based on the application's priorities.
```
ianswett: Thanks, this is much better!

draft-ietf-quic-recovery.md
```diff
@@ -415,100 +414,85 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
-considered a strong enough signal of packet loss, so does not result in an
-immediate change to congestion window or recovery state. An RTO timer expires
-only when there's a prolonged period of network silence, which could be caused
-by a change in the network RTT.
+#### Loss Detection {#pto-loss}
 
-QUIC also diverges from TCP by including MaxAckDelay in the RTO period. Since
-QUIC corrects for this delay in its SRTT and RTTVAR computations, it is
-necessary to add this delay explicitly in the TLP and RTO computation.
+Delivery or loss of packets in flight is established when an acknowledgement is
```
ianswett: Should this be "an ACK"?  We typically talk about an acknowledgement as referencing a single packet.

draft-ietf-quic-recovery.md
```diff
@@ -415,100 +414,85 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
+received that newly acknowledges one or more packets.  An endpoint might receive
+an acknowledgement after multiple consecutive PTO periods.
```
ianswett: I feel like you're trying to say something more with this last sentence, but I'm not sure what?

draft-ietf-quic-recovery.md
```diff
@@ -802,7 +772,7 @@ Pseudocode for SetLossDetectionTimer follows:
         timeout = 2 * kInitialRtt
       else:
         timeout = 2 * smoothed_rtt
-      timeout = max(timeout, kMinTLPTimeout)
+      timeout = max(timeout, granularity)
```
```suggestion
      timeout = max(timeout, kGranularity)
```

draft-ietf-quic-recovery.md
```diff
@@ -812,17 +782,11 @@ Pseudocode for SetLossDetectionTimer follows:
...
     timeout =
       smoothed_rtt + 4 * rttvar + max_ack_delay
-    timeout = max(timeout, kMinRTOTimeout)
-    timeout = timeout * (2 ^ rto_count)
-    if (tlp_count < kMaxTLPs):
-      // Tail Loss Probe
-      tlp_timeout = max(1.5 * smoothed_rtt
-                        + max_ack_delay, kMinTLPTimeout)
-      timeout = min(tlp_timeout, timeout)
+    timeout = max(timeout, granularity)
```
```suggestion
    timeout = max(timeout, kGranularity)
```

draft-ietf-quic-recovery.md
ianswett: If you have a lot of tails(ie: HTTP traffic), then it doesn't take too many packet losses to end up hitting this.  TLP was targeted partially at cases like that, and I think you'll find changing it will produce measurable performance regressions in web traffic.  We can do something else in a follow-up, but as I said before, I'd prefer text saying SHOULD consider an acknowledgement of the first PTO as standard packet loss, not extreme congestion, and treat it as such.

draft-ietf-quic-recovery.md
```diff
@@ -722,14 +695,13 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
...
         FindSmallestNewlyAcked(newly_acked_packets)
-      // If any packets sent prior to RTO were acked, then the
-      // RTO was spurious. Otherwise, inform congestion control.
-      if (rto_count > 0 &&
-            smallest_newly_acked > largest_sent_before_rto):
-        OnRetransmissionTimeoutVerified(smallest_newly_acked)
+      // If any packets sent prior to PTO were acked, then the
+      // PTO was spurious. Otherwise, inform congestion control.
+      if (pto_count > 0 &&
+            smallest_newly_acked > largest_sent_before_pto):
```
ianswett: You could define a kNumCongestivePtoPackets here and add it here and then define it above to be the number of PTOs you will treat as standard packet loss, and not extreme congestion.

martinthomson: What is "extreme congestion" ?

janaiyengar: @martinthomson : @ianswett is trying to capture the difference in congestion response that the current loss recovery spec has for TLP vs. RTO. The idea is to say that for a couple of PTOs, there will be a smaller reduction in cwnd, and after that, the cwnd will be collapsed. I can write that up.

draft-ietf-quic-recovery.md
janaiyengar: I'm going to simplify this further. I think that having a verification is unnecessary; the behavior ends up being the same whether a PTO is verified or not. I can retain this notion of reducing the cwnd differently for the first 2 PTOs and then collapsing it to min_cwnd after that.

janaiyengar:
I've updated the description text and the PR quite a bit based on the comments. Specifically, I've removed the verification of a timeout, since the behavior with and without it was the same. I've also added a clause that allows for the cwnd to be collapsed to min cwnd only after 2 PTOs, to retain parity with TLP/RTO behavior.

I'll open an issue to discuss the issues this PR addresses. We should have discussion there.

draft-ietf-quic-recovery.md
ianswett: If you want to use granularity, then you need to define it above.  Otherwise, use kGranularity as suggested.

draft-ietf-quic-recovery.md
```diff
@@ -983,24 +926,18 @@ During recovery, the congestion window remains unchanged irrespective of new
...
 
-When retransmissions are sent due to a retransmission timeout timer, no change
-is made to the congestion window until the next acknowledgement arrives.  The
-retransmission timeout is considered spurious when this acknowledgement
-acknowledges packets sent prior to the first retransmission timeout.  The
-retransmission timeout is considered valid when this acknowledgement
-acknowledges no packets sent prior to the first retransmission timeout.  In this
-case, the congestion window MUST be reduced to the minimum congestion window and
-slow start is re-entered.
+If two consecutive PTOs have occurred (pto_count is at least 3), the network is
```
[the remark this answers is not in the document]
ianswett: +1, I'm confused.  Also, this should be more about the number of PTOs that were lost than the number sent.   ie: I could send 3 PTOs, then the network returns to normal, no packets were lost, and the connection continues.

draft-ietf-quic-recovery.md
```diff
@@ -425,100 +425,79 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
-since acks of subsequent packets are needed to trigger ack-based detection. To
-ameliorate this weakness of tail packets, the sender schedules a timer when the
-last ack-eliciting packet before quiescence is transmitted. Upon timeout,
-a Tail Loss Probe (TLP) packet is sent to evoke an acknowledgement from the
-receiver.
+A Probe Timeout (PTO) triggers a probe packet when ack-eliciting data is in
+flight but an acknowledgement is not received within the expected period of
+time.  A PTO enables a connection to recover from loss of tail packets or acks.
+The PTO algorithm used in QUIC implements the reliability functions of Tail Loss
+Probe {{?TLP=I-D.dukkipati-tcpm-tcp-loss-probe}}, RTO {{?RFC5681}} and F-RTO
```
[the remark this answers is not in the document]
janaiyengar: I think the function we're talking about is TLP, though I'm happy to add and cite RACK separately somewhere. I'll do that separately.

draft-ietf-quic-recovery.md
```diff
@@ -587,21 +562,18 @@ kTimeThreshold:
...
-kMinRTOTimeout:
-:  Minimum time in the future an RTO timer may be set for. The RECOMMENDED
-   value is 200ms.
+: Timer granularity. This is a system-dependent value.  However, implementations
+  SHOULD use a value no smaller than 1ms.
 
 kDelayedAckTimeout:
 : The length of the peer's delayed ack timer. The RECOMMENDED value is 25ms.
 
 kInitialRtt:
```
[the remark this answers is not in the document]
janaiyengar: Would you mind opening a separate issue for this? I'd argue for 100ms, but I don't want to do it in this PR. FWIW, I think the cost of an early timeout is not high, so having this be aggressive seems fine to me.

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
janaiyengar: This is based on the number of timeouts that happened, not based on packets sent.  We want cwnd collapse at the equivalent of the first RTO's occurrence. This is the third timeout (2 TLP timeouts and then 1 RTO), which means pto_count is at least 3. Or am I doing this math wrong?

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
ianswett: Agreed, that's a separable issue, which was previously raised and then I believe closed.

draft-ietf-quic-recovery.md
```diff
@@ -1157,6 +1095,9 @@ window.
        congestion_window *= kLossReductionFactor
        congestion_window = max(congestion_window, kMinimumWindow)
        ssthresh = congestion_window
+       // Collapse congestion window if persistent congestion
+       if (pto_count > 2):
```
ianswett: 2 is a magic number.  Previously we said you SHOULD have 2 TLPs, but you could have fewer.  I think it makes sense to define this as a constant and explain its reason for existing there?

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
kazuho: To me it seems that "pto_count is at least 3" is correct, however "If two consecutive PTOs have occurred" is incorrect and needs to be changed to "If three...".

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
pravb: TLP is an expired draft in 2013 and RACK is standards track and active. Linux, FreeBSD and Windows are all implementing the RACK draft and TLP as specified in the RACK draft. I would just cite the RACK draft. There is a chance it may got into last call within the next two IETFs.

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
pravb: Ok I will open a separate issue. This will cause unnecessary timeouts for CI and 0-RTT data on many Internet paths. And pretty much always on satellite links. If we want to pick a magic number then let's pick 200 msec or 300 msec (the historical minrto values for Linux and Windows respectively).

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
pravb: IIRC Ian was going to get data on whether we need 2 TLPs or just one is sufficient. FWIW we only do one TLP in Windows TCP. Yes at the least please fix as Kazuho says: "if three consecutive PTOs have occurred"

draft-ietf-quic-recovery.md
pravb: I commented on the thread above this. Yes we should settle on one or two TLPs. I personally prefer one but unfortunately I don't have data to say if two is better since Windows TCP has been doing just one TLP.

draft-ietf-quic-recovery.md
```diff
@@ -425,100 +425,80 @@ from when the first Initial was sent to when a Retry or a Version Negotiation
...
-A PTO value of at least 1.5*SRTT ensures that the ACK is overdue.  The 1.5 is
-based on {{?TLP}}, but implementations MAY experiment with other constants.
-
-To reduce latency, it is RECOMMENDED that the sender set and allow the TLP timer
-to fire twice before setting an RTO timer. In other words, when the TLP timer
-expires the first time, a TLP packet is sent, and it is RECOMMENDED that the TLP
-timer be scheduled for a second time. When the TLP timer expires the second
-time, a second TLP packet is sent, and an RTO timer SHOULD be scheduled {{rto}}.
+~~~
+PTO = max(smoothed_rtt + 4*rttvar + max_ack_delay, kGranularity)
```
pravb: Sorry just noticed another difference. If peer advertises max_ack_delay as 0 (which seems to be legal), then first PTO becomes Srtt + 4 * RttVar, whereas in TCP the first PTO is 2* Srtt. If variance is low this may be too aggressive? Should we also do PTO = max (PTO, 2 *Srtt) ?

janaiyengar: I'll push back against this. I like the simplicity of having a single computation that includes rttvar. What is the rationale in TCP for using 2xSRTT when RTTVAR is available? If variance is low, that still seems like the right thing to do. Why inflate the PTO when you don't need to?

FWIW, the first PTO is akin to the first TLP. The second PTO would be backed off, so you will automatically get 2 x (SRTT + 4xRTTVAR).

draft-ietf-quic-recovery.md
janaiyengar: @siyengar : In this case, by the time 5 is acked, 3 and 4 should be marked as lost by the time-threshold loss detection mechanism, since they're going to be > 1.125 x SRTT in the past.

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
mikkelfj: Here are 10 ping packets from Denmark to China

10 packets transmitted, 10 packets received, 0.0% packet loss
round-trip min/avg/max/stddev = 211.218/350.768/593.271/149.383 ms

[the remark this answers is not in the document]
mikkelfj: And here from Denmark to New York, which is basically a direct sea cable link.

10 packets transmitted, 10 packets received, 0.0% packet loss
round-trip min/avg/max/stddev = 103.982/104.505/105.343/0.475 ms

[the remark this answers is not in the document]
mikkelfj: And Denmark New Zealand

10 packets transmitted, 10 packets received, 0.0% packet loss
round-trip min/avg/max/stddev = 324.042/454.764/672.469/122.684 ms

draft-ietf-quic-recovery.md
```diff
@@ -1065,6 +1004,16 @@ kLossReductionFactor:
...
   The RECOMMENDED value is 0.5.
 
+kPersistentCongestion:
+: Number of consecutive PTOs after which network is considered to be
+  experiencing persistent congestion.  The rationale for this threshold is to
+  enable a sender to use initial PTOs for aggressive probing, similar to Tail
+  Loss Probe (TLP) in TCP {{TLP}} {{RACK}}.  Once the number of consecutive PTOs
+  reaches this threshold - that is, persistent congestion is established - the
+  sender responds by collapsing its congestion window, similar to a
+  Retransmission Timeout (RTO) in TCP {{RFC5681}}.  The RECOMMENDED value is 3,
```
ianswett: The pseudocode below decreases CWND to kMinimumWindow, which is 2 packets.  I think sticking with that makes sense, at least for now.  It also requires less text to justify.

draft-ietf-quic-recovery.md
pravb: Quoting RACK draft:

Aiming for a PTO value of 2*SRTT allows a sender to wait long enough
   to know that an ACK is overdue.  Under normal circumstances, i.e. no
   losses, an ACK typically arrives in one SRTT.  But choosing PTO to be
   exactly an SRTT is likely to generate spurious probes given that
   network delay variance and even end-system timings can easily push an
   ACK to be above an SRTT.  We chose PTO to be the next integral
   multiple of SRTT.

   Similarly, network delay variations and end-system processing
   latencies and timer granularities can easily delay ACKs beyond
   2*SRTT, so senders SHOULD add at least 2ms to a computed PTO value
   (and MAY add more if the sending host OS timer granularity is more
   coarse than 1ms).

marten-seemann:
I think I found the answer to my question:
Since in #1974 we make time-threshold based loss recovery mandatory (before that you could choose between packet-threshold and time-threshold), we don't need the RTO verified anymore (see #2168) anyway. This is independent of this PR, this PR just happens to fix multiple issues at the same time.

Sorry for the noise.

draft-ietf-quic-recovery.md
janaiyengar: The recommendation is 3 for kPersistentCongestion, not for the value of cwnd. I'll change the text a bit to clarify.

draft-ietf-quic-recovery.md
janaiyengar: @pravb: Thanks for the context -- I should have just looked at the RACK draft. So, to be clear, the RACK draft does not use this 2ms when WCDelAckT is used, that is, when it accounts for the worst case delayed ack timer. This is what we expect max_ack_delay to do for us, but as you note, a receiver might send 0 here. 

I think we should put the onus of this onto the receiver.  A receiver should send the worst case ack delay in max_ack_delay. A receiver that doesn't do that shoots itself in the foot by causing spurious retransmissions at the sender. We may want to add some strong wording about how a receiver should choose max ack delay, I've filed #2186 for that.

Sound reasonable?

janaiyengar:
There's exactly one question that's pending to be resolved now in this PR: @pravb's question about using 2SRTT for initial PTO. I'm happy to open a new issue for that specific question if Praveen thinks we should continue that discussion, but I'd like to merge this soon, as (i) I'm hearing pushback on a detail or two but not the overall PR, which we can handle in separate issues, and (ii) this PR is fast approaching github's limit on comments (hence the unicorn tag).

draft-ietf-quic-recovery.md
pravb: That's reasonable. So this is how we'd justify the deviation from TCP:
1. Include RttVar component in the computation for in network variability. 
2. Require the receiver to include the perturbation for system processing delays. So 2186 needs to add a SHOULD for adding a perturbation value on top of the actual max ack delay. I'll comment there.

PR #2123, opened 2018-12-12
Strengthen 2119 language around tokens.

ekr:
Rationale:

1. We should recommend that people use tokens.
2. For privacy reasons we should just forbid reuse
3. Saying "SHOULD" attempt to validate seems weak. Why isn't this a
MUST? This is the weakest of these changes, I think.


ekr:
Ugh. I see I didn't read through to the end. Closing

PR #2127, opened 2018-12-12
HTTP: make few minor stylistic improvements

dtikhonov:


draft-ietf-quic-http.md
```diff
@@ -1079,9 +1079,9 @@ priority.
 ### Placeholders
 
 In HTTP/2, certain implementations used closed or unused streams as placeholders
-in describing the relative priority of requests.  However, this created
-confusion as servers could not reliably identify which elements of the priority
-tree could safely be discarded. Clients could potentially reference closed
+in describing the relative priority of requests.  This created
+confusion as servers could not identify reliably which elements of the priority
```
MikeBishop: I think this is placed correctly, and I feel like the change reads awkwardly.

draft-ietf-quic-http.md
```diff
@@ -1079,9 +1079,9 @@ priority.
...
+tree could be safely discarded. Clients could potentially reference closed
```
MikeBishop: I think you're right that this ought to move, but I think "could be discarded safely" reads better than either.

PR #2134, opened 2018-12-12
Reorder some text around dynamic table size

bencebeky:
This is a purely editorial PR reordering text.  The first commit reorders paragraphs and adds a section title, the second commit reorders sentences and moves around paragraph boundaries, the third commit is some clarification of text.

This PR does not at all address the issue I raised at https://github.com/quicwg/base-drafts/pull/2115#issuecomment-446721184.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -384,15 +384,14 @@ without any Huffman encoding applied.
 
 The encoder decides how to update the dynamic table and as such can control how
 much memory is used by the dynamic table.  To limit the memory requirements of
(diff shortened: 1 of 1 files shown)


bencebeky:
I just swapped the Eviction section with the Maximum Table Size one.  This makes sense because eviction talks about maximum table size (soon to be renamed table capacity or something similar), so it should come first, then Maximum Table Size talks about the limit on this value.

When working on https://github.com/quicwg/base-drafts/pull/2115 I made this change so I decided to factor out the swap of these sections into this PR which is already about reordering, to keep https://github.com/quicwg/base-drafts/pull/2115 more narrowly focused.

PR #2140, opened 2018-12-13
Clarify OnAckReceived pseudocode

Ralith:
I ran into a few points of confusion in the existing code:
- The time to update and scope of `largest_acked_packet` is not explicit
- `ProcessECN` references elements of `sent_packets` after they are removed by `OnPacketAcked`
- `DetectLostPackets` is passed an undefined variable

The fixes are trivial, except for the last: it isn't clear to me whether `DetectLostPackets` was intended to be called with `ack.largest_acked` or `largest_acked_packet`. These only differ when ACKs are reordered, and because the prior invocation of `DetectLostPackets` with a higher `largest_acked` would have already deemed lost all packets that the later invocation might deem lost, I don't think the behavior differs even then. Hence, removing the argument and consistently referencing `largest_acked_packet` seems to be the simpler option.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -634,7 +634,7 @@ largest_sent_packet:
 : The packet number of the most recently sent packet.
 
 largest_acked_packet:
-: The largest packet number acknowledged in an ACK frame.
+: The largest packet number acknowledged in the packet number space so far.
 
 latest_rtt:
 : The most recent RTT measurement made when receiving an ack for
@@ -682,6 +682,7 @@ follows:
    time_of_last_sent_ack_eliciting_packet = 0
    time_of_last_sent_crypto_packet = 0
    largest_sent_packet = 0
+   largest_acked_packet = 0
 ~~~
 
 ### On Sending a Packet
@@ -716,6 +717,9 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
 
 ~~~
   OnAckReceived(ack):
+    largest_acked_packet = max(largest_acked_packet,
+                               ack.largest_acked)
+
     // If the largest acknowledged is newly acked and
     // ack-eliciting, update the RTT.
     if (sent_packets[ack.largest_acked] &&
@@ -741,7 +745,7 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
       tlp_count = 0
       rto_count = 0
 
-    DetectLostPackets(ack.acked_packet)
+    DetectLostPackets()
     SetLossDetectionTimer()
 
     // Process ECN information if present.
@@ -853,7 +857,7 @@ Pseudocode for OnLossDetectionTimeout follows:
        crypto_count++
      else if (loss_time != 0):
        // Time threshold loss Detection
-       DetectLostPackets(largest_acked_packet)
+       DetectLostPackets()
      else if (tlp_count < kMaxTLPs):
        // Tail Loss Probe.
        SendOnePacket()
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -720,9 +720,14 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
     // ack-eliciting, update the RTT.
     if (sent_packets[ack.largest_acked] &&
         sent_packets[ack.largest_acked].ack_eliciting):
+      largest_acked_packet = ack.largest_acked
```
martinthomson: This doesn't work because the largest_acked in the ACK might be newly acknowledged, but not the largest ever acknowledged.

```suggestion
      if ack.largest_acked > largest_acked_packet:
         largest_acked_packet = ack.largest_acked
```

draft-ietf-quic-recovery.md
```diff
@@ -720,9 +720,14 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
...
       latest_rtt = now - sent_packets[ack.largest_acked].time
       UpdateRtt(latest_rtt, ack.ack_delay)
 
+    // Process ECN information if present.
+    if (ACK frame contains ECN information):
+       ProcessECN(ack)
```
martinthomson: This is a separate change.  I think that you should make this a separate PR.

Ralith:
Split the `ProcessECN` tweak into https://github.com/quicwg/base-drafts/pull/2142, corrected the `largest_acked_packet` update logic, and added initialization for `largest_acked_packet`.

draft-ietf-quic-recovery.md
```diff
@@ -634,7 +634,7 @@ largest_sent_packet:
 : The packet number of the most recently sent packet.
 
 largest_acked_packet:
-: The largest packet number acknowledged in an ACK frame.
+: The largest packet number acknowledged by any ACK frame.
```
```suggestion
: The largest packet number acknowledged in the connection so far.
```

martinthomson: Isn't this in the packet number space?

martinthomson: I think that you mean "packet number space" rather than connection.

draft-ietf-quic-recovery.md
```diff
@@ -716,6 +717,9 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
 
 ~~~
   OnAckReceived(ack):
+    largest_acked_packet = max(largest_acked_packet,
```
ianswett: I think this accurately describes the text we have, but I think we should add some text that you MAY use the largest acked from the most recent ACK, which could reduce spurious retransmits when one packet arrives far out of order?

ianswett: To clarify, I'm ok doing this in a follow up PR.

Ralith: > you MAY use the largest acked from the most recent ACK, which could reduce spurious retransmits when one packet arrives far out of order

Any reason not to make this the sole recommendation? I had some trouble coming up with a case where that might occur, which is why I didn't take that route originally, but I think I see it now: when a reordered ACK newly acknowledges a packet such that the newly acknowledged packet is > ack.largest_acked - kPacketThreshold, but <= largest_acked_packet - kPacketThreshold.

Ralith: Er, wait, that's nonsense. Still having trouble seeing the case.

ianswett: It's the case when one packet is far out of order, and then subsequent packets arrive in order.  ie: 100 is received, then 1, 2, 3, 4, 5, etc.  I don't fully understand the cause, but I've seen small UDP packets jump ahead of large ones in a number of traces.

But I suspect Jana and I will merge this as is and we can add follow-up text about that issue if we think it's valuable.

PR #2176, opened 2018-12-14
Ekr editorial 17 tls

ekr:


martinthomson:
This is now sufficient rotten that I think we will have to start over.

PR #2181, opened 2018-12-14
Pad to at least 1200 bytes

### LINKED ISSUE #2167: How long is Initial
the TLS draft says:

```
QUIC requires that the first Initial packet from a client contain an entire
cryptographic handshake message, which for TLS is the ClientHello.  Though a
packet larger than 1200 bytes might be supported by the path, a client improves
the likelihood that a packet is accepted if it ensures that the first
ClientHello message is small enough to stay within this limit.
```

This clearly implies that you can have an Initial > 1200 bytes.

However, transport S 7.1 says

```
   Clients MUST pad UDP datagrams that contain only Initial packets to
   1200 bytes.  Once a client has received an acknowledgment for a
   Handshake packet it MAY send smaller datagrams.  Sending padded
   datagrams ensures that the server is not overly constrained by the
   amplification restriction.
```

But you can't pad to 1200 bytes if you are over 1200 bytes. so which one is it?

martinthomson:
Closes #2167.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1539,10 +1539,10 @@ magnitude of any amplification attack that can be mounted using spoofed source
 addresses.  In determining this limit, servers only count the size of
 successfully processed packets.
 
-Clients MUST pad UDP datagrams that contain only Initial packets to 1200 bytes.
-Once a client has received an acknowledgment for a Handshake packet it MAY send
-smaller datagrams.  Sending padded datagrams ensures that the server is not
-overly constrained by the amplification restriction.
+Clients MUST pad UDP datagrams that contain only Initial packets to at least
+1200 bytes.  Once a client has received an acknowledgment for a Handshake packet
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1539,10 +1539,10 @@ magnitude of any amplification attack that can be mounted using spoofed source
 addresses.  In determining this limit, servers only count the size of
 successfully processed packets.
 
-Clients MUST pad UDP datagrams that contain only Initial packets to 1200 bytes.
-Once a client has received an acknowledgment for a Handshake packet it MAY send
-smaller datagrams.  Sending padded datagrams ensures that the server is not
-overly constrained by the amplification restriction.
+Clients MUST pad UDP datagrams that contain only Initial packets to at least
```
marten-seemann: Technically, we're padding the QUIC packet, not the UDP datagram.

janaiyengar: You could pad the datagram by sending a coalesced packet with only padding frames.

PR #2182, opened 2018-12-14
Keep the packet number after Retry

### LINKED ISSUE #2179: If you don't reset DCID you will repeat keys.
```
Note that if you don't reset DCID, you repeat keys


A client sets the Destination Connection ID field of this Initial packet to the
value from the Source Connection ID in the Retry packet. Changing Destination
Connection ID also results in a change to the keys used to protect the Initial
packet. It also sets the Token field to the token provided in the Retry. The
client MUST NOT change the Source Connection ID because the server could include
the connection ID as part of its token validation logic (see
{{validate-future}}).
```

Correct me if I'm wrong, but if the server chooses a DCID that is the same as the client's DCID (which, I believe it can do), then you would get the same keys. Arguably, this doesn't matter b/c any attacker who is on-path can derive the keys anyway, but...

martinthomson:
Closes #2179.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3859,8 +3859,10 @@ the connection ID as part of its token validation logic (see
 {{validate-future}}).
 
 The next Initial packet from the client uses the connection ID and token values
-from the Retry packet (see {{negotiating-connection-ids}}).  Aside from this,
-the Initial packet sent by the client is subject to the same restrictions as the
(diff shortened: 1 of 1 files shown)


marten-seemann:
This seems unnecessarily complex. Why don’t we just require the server to generate a new connection ID for a Retry, and then just start at packet number 0 again?

martinthomson:
Oh, stupid me....

> The server includes a connection ID of its choice in the Source Connection ID field. This value MUST not be equal to the Destination Connection ID field of the packet sent by the client.

PR #2183, opened 2018-12-14
The first Initial needs to fit into a single IP packet

martinthomson:
Not sure if this needs to address IP fragmentation.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3081,14 +3081,14 @@ corrupt ECN codepoints in the IP header.
 The QUIC packet size includes the QUIC header and protected payload, but not the
 UDP or IP header.
 
-Clients MUST ensure they send the first Initial packet in a UDP datagram that is
-at least 1200 bytes.  The payload of a UDP datagram carrying the Initial packet
-MUST be expanded to at least 1200 bytes, by adding PADDING frames to the Initial
-packet and/or by combining the Initial packet with a 0-RTT packet (see
-{{packet-coalesce}}).  Sending a UDP datagram of this size ensures that the
-network path supports a reasonable Maximum Transmission Unit (MTU), and helps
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3081,14 +3081,14 @@ corrupt ECN codepoints in the IP header.
...
 
-Clients MUST ensure they send the first Initial packet in a UDP datagram that is
-at least 1200 bytes.  The payload of a UDP datagram carrying the Initial packet
-MUST be expanded to at least 1200 bytes, by adding PADDING frames to the Initial
-packet and/or by combining the Initial packet with a 0-RTT packet (see
-{{packet-coalesce}}).  Sending a UDP datagram of this size ensures that the
-network path supports a reasonable Maximum Transmission Unit (MTU), and helps
-reduce the amplitude of amplification attacks caused by server responses toward
-an unverified client address, see {{address-validation}}.
+Clients MUST ensure they send the first Initial packet in single IP packet.  The
```
janaiyengar: Should this say something about the first Initial packet (and the first Initial packet after a Retry)?

```suggestion
Clients MUST ensure they send the first Initial packet in single IP datagram.  The
```

janaiyengar:
I'm fine with this. I don't think it's worth getting into a discussion about IP fragmentation here.

draft-ietf-quic-transport.md
mikkelfj: Isn't there a broader requirement to never fragment IP packets/datagrams when doing QUIC?

martinthomson: We've been using IP packet, UDP datagram, and QUIC packet.  Not great, but the IP RFCs use packet.

martinthomson:
Yes, this also applies after Retry, and this is the best place to add that text.  There's a space now if we wanted to say "don't use IP fragmentation", but I'm happy with this.

PR #2187, opened 2018-12-14
Allow alternate restarts after idle

### LINKED ISSUE #2138: How long idle is idle?
[Section 7.8 of -recovery](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#restart-after-idle) specifies rules for what a sender does when it starts sending after an idle period.  But surely this doesn't apply when the idle period is short.  Reading that text, an idle period of any duration would trigger this requirement, but I don't think that is the intent.  The period should be defined.

janaiyengar:
Closes #2138.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -964,9 +964,9 @@ in Linux (3.11 onwards).
 
 ## Sending data after an idle period
 
-A sender is considered idle if it has no bytes in flight.  A sender can become
-idle when it is application limited and has no data to send, or when it marks
-packets as lost (and therefore not in flight).
+A sender becomes idle if it has no bytes in flight.  A sender can become idle
+when it is application limited and has no data to send, or when it marks packets
+as lost (and therefore not in flight).
 
 A sender's congestion window MUST not increase while it is idle.
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -964,11 +964,8 @@ in Linux (3.11 onwards).
 
 ## Sending data after an idle period
 
-A sender becomes idle if it has no bytes in flight.  A sender can become idle
-when it is application limited and has no data to send, or when it marks packets
-as lost (and therefore not in flight).
-
-A sender's congestion window MUST not increase while it is idle.
+A sender becomes idle if it ceases to send data and has no bytes in flight.  A
+sender's congestion window MUST not increase while it is idle.
 
 When sending data after becoming idle, a sender MUST reset its congestion window
 to the initial congestion window (see Section 4.1 of {{?RFC5681}}), unless it
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -969,14 +969,17 @@ application limited or after a probe timeout. In order to limit
 the size of bursts sent into the network, the behavior when restarting from
 idle depends upon whether pacing is used.
 
-If the sender uses pacing, the connection should limit the initial burst of
-packets to no more than the initial congestion window and subsequent packets
-SHOULD be paced. The congestion window does not change while the connection
-is idle.
+If the sender uses pacing, the congestion window MUST not increase and does not
+need to be reduced while the connection is idle.  When restarting from idle, a
+paced connection SHOULD limit the initial burst of packets to no more than the
+initial congestion window and subsequent packets SHOULD be paced.
 
 A sender that does not use pacing SHOULD reset its congestion window to the
-minimum of the current congestion window and the initial congestion window.
-This recommendation is based on Section 4.1 of {{?RFC5681}}.
+smaller of the current congestion window and the initial congestion window, as
+recommended for TCP (see Section 4.1 of {{?RFC5681}}).
+
+A sender MAY implement alternate mechanisms to update its congestion window
+after idle periods, such as those proposed for TCP in {{?RFC7661}}.
 
 ## Pseudocode
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -961,22 +961,24 @@ As an example of a well-known and publicly available implementation of a flow
 pacer, implementers are referred to the Fair Queue packet scheduler (fq qdisc)
 in Linux (3.11 onwards).
 
-## Restart after idle
+## Sending data after an idle period
 
-A connection is idle if there are no bytes in flight and there is no pending
-ack-eliciting data to send.  This can occur when the connection is
-application limited or after a probe timeout. In order to limit
-the size of bursts sent into the network, the behavior when restarting from
-idle depends upon whether pacing is used.
+A sender is considered idle if it has no bytes in flight and no pending
+ack-eliciting data to send.  A sender can become idle when it is application
+limited or when it encounters a retransmission timeout.
 
-If the sender uses pacing, the congestion window MUST not increase and does not
-need to be reduced while the connection is idle.  When restarting from idle, a
-paced connection SHOULD limit the initial burst of packets to no more than the
-initial congestion window and subsequent packets SHOULD be paced.
+A sender's congestion window MUST not increase while it is idle.
 
-A sender that does not use pacing SHOULD reset its congestion window to the
-smaller of the current congestion window and the initial congestion window, as
-recommended for TCP (see Section 4.1 of {{?RFC5681}}).
+When a sender starts sending data after an idle period, it's sending rate SHOULD
+be decided as follows.
+
+  - If the sender uses pacing, it does not need to reduce its congestion
+    window. It SHOULD however pace the congestion window and MAY allow an
+    initial burst no larger than the initial congestion window.
+
+  - If the sender does not use pacing, it SHOULD reset its congestion window to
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -1030,14 +1030,17 @@ application limited or after a retransmission timeout. In order to limit
 the size of bursts sent into the network, the behavior when restarting from
 idle depends upon whether pacing is used.
 
-If the sender uses pacing, the connection should limit the initial burst of
-packets to no more than the initial congestion window and subsequent packets
-SHOULD be paced. The congestion window does not change while the connection
-is idle.
+If the sender uses pacing, the congestion window MUST not increase and does not
```
ianswett: Possibly the "MUST not increase CWND" part should move a paragraph up, since it's really not pacing specific?

draft-ietf-quic-recovery.md
```diff
@@ -1030,14 +1030,17 @@ application limited or after a retransmission timeout. In order to limit
...
+initial congestion window and subsequent packets SHOULD be paced.
 
 A sender that does not use pacing SHOULD reset its congestion window to the
-minimum of the current congestion window and the initial congestion window.
-This recommendation is based on Section 4.1 of {{?RFC5681}}.
+smaller of the current congestion window and the initial congestion window, as
+recommended for TCP (see Section 4.1 of {{?RFC5681}}).
+
+A sender MAY implement alternate mechanisms to update its congestion window
+after idle periods, such as those proposed for TCP in {{?RFC7661}}.
```
ianswett: To be clear, we're not really expecting anyone to implement this, right?

mikkelfj: As formulated this does not permit bursts up to a decayed cwnd because init cwnd is required if not pacing.

janaiyengar: We do want to limit bursts to init cwnd. Beyond that, instead of describing alt mechanisms here, I think RFC 7661 offers a good and well thought out set of considerations.

janaiyengar:
I've restructured the section a bit more.

draft-ietf-quic-recovery.md
mikkelfj: Right, I suspected as much, just wanted to be sure.

janaiyengar:
@ianswett: Take a look again -- I've made some changes.

draft-ietf-quic-recovery.md
```diff
@@ -961,22 +961,27 @@ As an example of a well-known and publicly available implementation of a flow
...
-This recommendation is based on Section 4.1 of {{?RFC5681}}.
+## Sending data after an idle period
+
+A sender is considered idle if it has no bytes in flight and no pending
+ack-eliciting data to send.  A sender can become idle when it is application
+limited or when it encounters a retransmission timeout.
+
+A sender's congestion window MUST not increase while it is idle.
+
+When a sender starts sending data after an idle period, it's sending rate SHOULD
```
```suggestion
When a sender starts sending data after an idle period, its sending rate SHOULD
```

draft-ietf-quic-recovery.md
```diff
@@ -961,22 +961,27 @@ As an example of a well-known and publicly available implementation of a flow
...
+be decided as follows.
```
```suggestion
be decided as follows:
```

draft-ietf-quic-recovery.md
```diff
@@ -961,22 +961,27 @@ As an example of a well-known and publicly available implementation of a flow
...
+
+  - If the sender uses pacing, it does not need to reduce its congestion
+    window. It SHOULD however pace the congestion window and MAY allow an
+    initial burst no larger than the initial congestion window.
+
+  - If the sender does not use pacing, it SHOULD reset its congestion window to
+    the smaller of the current congestion window and the initial congestion
+    window, as recommended for TCP (see Section 4.1 of {{?RFC5681}}).
```
martinthomson: Maybe ignore my suggestions and use this instead:

> When sending data after becoming idle, a sender MUST reset its congestion window to the initial congestion window (see blah), unless it paces the sending of packets.  A sender MAY retain its congestion window if it paces the sending of any packets in excess of the initial congestion window.

I realize that this changes the emphasis slightly, but I think that it's clearer and more succinct.

I still think that this is an abrupt transition, but I guess the answer to any complaint there is simply "pacing".

draft-ietf-quic-recovery.md
```diff
@@ -961,22 +961,27 @@ As an example of a well-known and publicly available implementation of a flow
...
+limited or when it encounters a retransmission timeout.
+
+A sender's congestion window MUST not increase while it is idle.
+
+When a sender starts sending data after an idle period, it's sending rate SHOULD
+be decided as follows.
+
+  - If the sender uses pacing, it does not need to reduce its congestion
+    window. It SHOULD however pace the congestion window and MAY allow an
+    initial burst no larger than the initial congestion window.
```
mikkelfj: I don't understand this. "MAY allow initial burst no larger", does this it MAY also allow larger?

mikkelfj: better

draft-ietf-quic-recovery.md
mikkelfj: There is a degenerate case where traffic arrives in blocks just under init cwnd at a high rate. The application can then burst at a high frequency.

draft-ietf-quic-recovery.md
```diff
@@ -961,22 +961,27 @@ As an example of a well-known and publicly available implementation of a flow
...
-is idle.
-
-A sender that does not use pacing SHOULD reset its congestion window to the
-minimum of the current congestion window and the initial congestion window.
-This recommendation is based on Section 4.1 of {{?RFC5681}}.
+## Sending data after an idle period
+
+A sender is considered idle if it has no bytes in flight and no pending
+ack-eliciting data to send.  A sender can become idle when it is application
+limited or when it encounters a retransmission timeout.
```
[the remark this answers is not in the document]
janaiyengar: I've replaced retransmission with probe. 
Actually I think this entire second point is wrong now. A sender does not become idle on a PTO, it might become idle once loss is established, when an ACK is received.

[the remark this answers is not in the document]
janaiyengar: Also, idle is when there's nothing in flight. I don't think we need the "pending data to send" bit.

draft-ietf-quic-recovery.md
janaiyengar: I like this better. I'm using this.

janaiyengar:
Ok, comments incorporated. I like @martinthomson's text, so I'm using that.

draft-ietf-quic-recovery.md
```diff
@@ -961,22 +961,22 @@ As an example of a well-known and publicly available implementation of a flow
...
-SHOULD be paced. The congestion window does not change while the connection
-is idle.
-
-A sender that does not use pacing SHOULD reset its congestion window to the
-minimum of the current congestion window and the initial congestion window.
-This recommendation is based on Section 4.1 of {{?RFC5681}}.
+
+## Sending data after an idle period
+
+A sender is considered idle if it has no bytes in flight.  A sender can become
```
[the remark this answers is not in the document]
martinthomson: Or

```suggestion
 A sender becomes idle if it has no bytes in flight.  A sender can become
```

draft-ietf-quic-recovery.md
```diff
@@ -961,22 +961,22 @@ As an example of a well-known and publicly available implementation of a flow
...
+idle when it is application limited and has no data to send, or when it marks
+packets as lost (and therefore not in flight).
```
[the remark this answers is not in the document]
janaiyengar: The text that was there was probably leftover from a while ago... at any rate there is no PTO case. There is nothing marked as lost on a PTO, and bytes in flight does not change on a PTO (modulo the two packets that are sent in response).  So a sender cannot become idle on a PTO.

draft-ietf-quic-recovery.md
```diff
@@ -961,22 +961,22 @@ As an example of a well-known and publicly available implementation of a flow
...
-is idle.
-
-A sender that does not use pacing SHOULD reset its congestion window to the
-minimum of the current congestion window and the initial congestion window.
-This recommendation is based on Section 4.1 of {{?RFC5681}}.
+
+## Sending data after an idle period
+
+A sender becomes idle if it has no bytes in flight.  A sender can become idle
+when it is application limited and has no data to send, or when it marks packets
```
```suggestion
 when it is application limited and has no data to send, or when it marks all packets
```

draft-ietf-quic-recovery.md
```diff
@@ -961,22 +961,22 @@ As an example of a well-known and publicly available implementation of a flow
...
+as lost (and therefore not in flight).
```
```suggestion
 as lost, and therefore not in flight.
```

ianswett: Can this also happen when all packets are acknowledged?  I'm not sure what the preferred behavior is in that case.  But the current text now properly captures the "my last PTO was acked and everything else was declared lost."

janaiyengar: Yes, that is a fine case ... this is what we've always had, the RTO text was incorrect. Basically the sender goes idle as soon as bytes in flight goes to 0. Maybe that's all we need to say. I've removed more text to make this clearer.

PR #2189, opened 2018-12-15
Editorial fixes to ECN text

janaiyengar:
Follow-up to #2113, and some other consistency checks.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2982,11 +2982,11 @@ rewritten on the path. An endpoint verifies the path, both during connection
 establishment and when migrating to a new path (see {{migration}}).
 
 
-### ECN Counters
+### ECN Counts
 
 On receiving a QUIC packet with an ECT or CE codepoint, an endpoint that can
 access the ECN codepoints from the enclosing IP packet increases the
-corresponding ECT(0), ECT(1), or CE count, and includes these counters in
+corresponding ECT(0), ECT(1), or CE count, and includes these counts in
 subsequent ACK frames (see {{processing-and-ack}} and {{frame-ack}}).
 
 A packet detected by a receiver as a duplicate does not affect the receiver's
@@ -2995,8 +2995,8 @@ concerns.
 
 If an endpoint receives a QUIC packet without an ECT or CE codepoint in the IP
 packet header, it responds per {{processing-and-ack}} with an ACK frame without
-increasing any ECN counters.  Similarly, if an endpoint does not have access to
-received ECN codepoints, it does not increase ECN counters.
+increasing any ECN counts.  Similarly, if an endpoint does not have access to
+received ECN codepoints, it does not increase ECN counts.
 
 Coalesced packets (see {{packet-coalesce}}) mean that several packets can share
 the same IP header.  The ECN counter for the ECN codepoint received in the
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3031,33 +3031,28 @@ an ACK frame without ECN feedback, the endpoint stops setting ECT codepoints in
 subsequent IP packets, with the expectation that either the network path or the
 peer no longer supports ECN.
 
-To reduce the risk of non-standard compliant ECN markings affecting the
-operation of an endpoint, an endpoint verifies the counts it receives when it
-receives new acknowledgements:
+Network devices that corrupt or apply non-standard ECN markings might result in
+reduced throughput or other undesirable side-effects.  To reduce this risk, an
+endpoint uses the following steps to verify the counts it receives in an ACK
+frame.  Note that the counts MUST NOT be verified if the ACK frame does not
```
```suggestion
frame.  Counts MUST NOT be verified if the ACK frame does not
```

Now we don't say why this is any more.  Is that OK?

There's a perverse situation in which this results in another problem with this logic.

That situation is where the largest acknowledged is lower on packet that is sent later.  AND that packet acknowledges more packets so that the counts are higher.  AND that packet is reordered ahead of packets with higher largest acknowledged.  This results in the verification failing on the second packet to arrive (counts are lower because it's an older ACK, but the largest acknowledged is higher).

A fix is to never decrease the largest acknowledged when sending an ACK frame.  Another is to disable validation if the packet containing the ACK is not higher-numbered than the last packet containing an ACK.  The question is whether we care enough to recommend either fix...

@gloinul, WDYT?

gloinul: To avoid all issues let us check so that the ACK frames are not reordered. For example rewrite this check to: "Counts MUST NOT be verified if the ACK frame arrived in an QUIC packet whose packet number is not the highest received so far."

draft-ietf-quic-transport.md
```diff
@@ -3031,33 +3031,28 @@ an ACK frame without ECN feedback, the endpoint stops setting ECT codepoints in
...
-* The increase in ECT(0) and ECT(1) counters MUST be at least the number of QUIC
-  packets newly acknowledged that were sent with the corresponding codepoint
-  minus the increase in the CE counter. This detects network remarking between
-  ECT(0) and ECT(1).
-
-* The total increase in ECT(0), ECT(1), and CE counters reported in the ACK
-  frame MUST be at least the total number of QUIC packets newly acknowledged in
-  this ACK frame. This detects if the network changes ECT(0), ECT(1) or CE to
+* The total increase in ECT(0), ECT(1), and CE counts MUST be no smaller than
+  the total number of QUIC packets newly acknowledged in this ACK frame.  This
```
gloinul: I suggest that you add the below (new after "frame").
"... in this ACK frame that was marked as ECT."
The reasons is that if a sender for some reason don't mark all packets, which is not required they need to check what was sent.

draft-ietf-quic-transport.md
martinthomson: Creating any coupling between frames and packets will be annoying.  We've resisted attempts to do that in the past.

There are two conditions necessary: first that counters never decrease when put in packets with higher numbers, second: that no check is performed if the packet number is lower.  I don't think that we have the first condition either (though it is strongly implied, we don't insist upon it for other frame types).

janaiyengar: We can add a requirement that a receiver MUST NOT decrease the largest acked. I cannot see a reason (let alone a good one) why a receiver would do this. QUIC does not allow reneging of acked data, so decreasing the largest acked is strictly bad. That resolves this issue, right?

ianswett: Hmm, this seems like a strong requirement we should think through a bit more.  At some point, I was thinking some very constrained implementations might just ACK the N most recent packets, and with reordering, that could cause largest acked to decrease.  Maybe that's not a compelling use case, but we should be sure we're ok with never decreasing it.

janaiyengar: I'd like to do this in a separate PR. This PR is editorial. While adding a requirement on largest acked shouldn't surprise anyone, it's probably best to make it a separate thing.

janaiyengar:
I've addressed the comments, and am merging. I'm happy to address the issue raised about largest acked in a separate issue (#2205).

PR #2200, opened 2018-12-17
Clarify In-flight

ianswett:
I'm not sure this is necessary, but I've had at least 2 people be unsure about whether a packet is still in flight if all the stream data in it is cancelled, so maybe it's worth stating this explicitly?

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -102,7 +102,8 @@ In-flight:
 
 : Packets are considered in-flight when they have been sent
   and neither acknowledged nor declared lost, and they are not
-  ACK-only.
+  ACK-only.  Packets are still considered in-flight even if
+  their payloads no longer need to be delivered.
 
 Ack-eliciting Frames:
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -102,8 +102,9 @@ In-flight:
 
 : Packets are considered in-flight when they have been sent
   and neither acknowledged nor declared lost, and they are not
-  ACK-only.  Packets are still considered in-flight even if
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -102,7 +102,8 @@ In-flight:
 
 : Packets are considered in-flight when they have been sent
   and neither acknowledged nor declared lost, and they are not
-  ACK-only.
+  ACK-only.  Packets are still considered in-flight even if
+  their payloads no longer need to be delivered.
```
```suggestion
  their payloads no longer need to be delivered in order to maintain the congestion window.
```

ianswett: I'm confused about what you're trying to add here?

ianswett:
I updated this to be a clarifying example, which I think is more helpful than re-iterating the previous sentence.  PTAL.

draft-ietf-quic-recovery.md
```diff
@@ -102,7 +102,9 @@ In-flight:
 
 : Packets are considered in-flight when they have been sent
   and neither acknowledged nor declared lost, and they are not
-  ACK-only.
+  ACK-only.  For example, packets containing STREAM frames are
```
martinthomson: I think that the "for example" is missing context.  Perhaps:

> : Packets are considered in-flight when they have been sent and neither acknowledged nor declared lost.  ACK-only packets are not considered in-flight, though all other packets are, including packets that contain data that would not otherwise be sent again, such as a packet containing only STREAM frames for a reset stream.

ianswett: That works for me.  @janaiyengar what do you think?

ianswett:
I fixed the pseudo-code issue in #2219 so I'm going to close this, since I think it's doing more harm than good.

PR #2203, opened 2018-12-18
Rearrange Packet Types section and references

### LINKED ISSUE #2157: Retry is not a long header
```
A Retry packet uses a long packet header with a type value of 0x3. It carries
an address validation token created by the server. It is used by a server that
wishes to perform a stateless retry (see {{validate-handshake}}).
```

This is not correct b/c the long header has length and PN.

### LINKED ISSUE #2165: References in Section 12 are awkward
> Packets that carry the long header are Initial Section 17.5, Retry Section 17.7, Handshake Section 17.6, and 0-RTT Protected packets Section 12.1.

These references should be delimited with parentheses, as is done in the other paragraphs:

> QUIC endpoints communicate by exchanging packets. Packets are carried in UDP datagrams (see Section 12.2) and have confidentiality and integrity protection (see Section 12.1).
>
> This version of QUIC uses the long packet header (see Section 17.2) during connection establishment and the short header (see Section 17.3) once 1-RTT keys have been established.

It also seems odd that our references are to subsections (5, 7, 6) rather than matching the order of the actual sections, and that three of the references are to 17.x while one is to 12.1.  Can we make these a little more symmetric?

MikeBishop:
Fixes #2165, fixes #2157, and probably fixes some others as well.  Trying to address one thing wound up pulling on some other pieces as well, so this is larger than I intended.

Key changes are:
- Added discussion of 0-RTT Protected Packets, which were mysteriously absent, to Section 17
- Moved discussion of starting packet numbers for each space from the Initial Packet definition to the existing Packet Numbers section (17.1)
- Moved Short Header packets to end of 17 (from 17.3 to 17.8) to get all the long-header packet types together
- Removed the non-invariant fields (Length, Packet Number, Payload) from the definition of the Long Header and made them type-specific fields for those packets which contain them; this reduces explanation for packet types which don't
- Reordered forward references in Section 10 so the references in each paragraph are to sequential sections, with some text updates to accommodate the new order

If a comparison of the output doc is more useful than the Markdown diff, try [the output diff](https://tools.ietf.org/rfcdiff?url1=https://quicwg.github.io/base-drafts/draft-ietf-quic-transport.txt&url2=https://quicwg.github.io/base-drafts/transport/packet-types/draft-ietf-quic-transport.txt) or the [resulting document](https://quicwg.github.io/base-drafts/transport/packet-types/draft-ietf-quic-transport.html).

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3821,7 +3821,7 @@ Handshake protection keys are discarded.
 
 ## Short Header Packet {#short-header}
 
-~~~~~
+~~~
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2602,10 +2602,17 @@ each packet sent, see {{packet-numbers}} for details.
 
 ## Coalescing Packets {#packet-coalesce}
 
-A sender can coalesce multiple QUIC packets into one UDP datagram.  This can
-reduce the number of UDP datagrams needed to complete the cryptographic
-handshake and starting sending data.  Receivers MUST be able to process
-coalesced packets.
+Initial ({{packet-initial}}), 0-RTT Protected ({{packet-0rtt}}), and Handshake
+({{packet-handshake}}) packets contain a Length field, which determines the end
+of the packet.  The Length field covers both the Packet Number and Payload
+fields, both of which are confidentiality protected and initially of unknown
+length. The length of the Payload field is learned once header protection is
+removed.
+
+Using the Length field, a sender can coalesce multiple QUIC packets into one UDP
+datagram.  This can reduce the number of UDP datagrams needed to complete the
+cryptographic handshake and starting sending data.  Receivers MUST be able to
+process coalesced packets.
 
 Coalescing packets in order of increasing encryption levels (Initial, 0-RTT,
 Handshake, 1-RTT) makes it more likely the receiver will be able to process all
@@ -2627,7 +2634,7 @@ packet for later processing and MUST attempt to process the remaining packets.
 
 Retry packets ({{packet-retry}}), Version Negotiation packets
 ({{packet-version}}), and packets with a short header cannot be followed by
-other packets in the same UDP datagram.
+other packets in the same UDP datagram, as they do not contain a Length field.
 
 
 ## Packet Numbers {#packet-numbers}
@@ -2658,7 +2665,7 @@ level and can only be acknowledged in Handshake packets.
 
 This enforces cryptographic separation between the data sent in the different
 packet sequence number spaces.  Each packet number space starts at packet number
-0.  Subsequent packets sent in the same packet number space MUST increase the
+1.  Subsequent packets sent in the same packet number space MUST increase the
 packet number by at least one.
 
 0-RTT and 1-RTT data exist in the same packet number space to make loss recovery
@@ -3396,12 +3403,6 @@ the connection.
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                 Source Connection ID (0/32..144)            ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                           Length (i)                        ...
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                    Packet Number (8/16/24/32)               ...
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                          Payload (*)                        ...
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ~~~~~
 {: #fig-long-header title="Long Header Packet Format"}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2658,7 +2663,7 @@ level and can only be acknowledged in Handshake packets.
 
 This enforces cryptographic separation between the data sent in the different
 packet sequence number spaces.  Each packet number space starts at packet number
-0.  Subsequent packets sent in the same packet number space MUST increase the
+1.  Subsequent packets sent in the same packet number space MUST increase the
```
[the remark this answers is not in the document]
MikeBishop: Bah, my editor sometimes does this and I don't catch it.  Let me correct that.

draft-ietf-quic-transport.md
```diff
@@ -2602,10 +2600,17 @@ each packet sent, see {{packet-numbers}} for details.
...
+length. The length of the Payload field is learned once header protection is
+removed.
+
+Using the Length field, a sender can coalesce multiple QUIC packets into one UDP
+datagram.  This can reduce the number of UDP datagrams needed to complete the
+cryptographic handshake and starting sending data.  Receivers MUST be able to
+process coalesced packets.
 
 Coalescing packets in order of increasing encryption levels (Initial, 0-RTT,
 Handshake, 1-RTT) makes it more likely the receiver will be able to process all
```
ianswett: 2 lines below: ", so it can only be the last packed included in a UDP datagram."

ianswett:
This adds a lot of duplicate text.  Can we avoid that, so we don't make a change in one spot and miss the other 3 spots?

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
janaiyengar: Hah! 0 _is_ special :-)

MikeBishop:
I think the way we would get there is to make Initial/Handshake/0-RTT subsections of one section that defines some common fields after the long header.  The "longer" header?

MikeBishop:
Done -- dubbed it the "Extended Long Header."

draft-ietf-quic-transport.md
```diff
@@ -3433,136 +3477,32 @@ Source Connection ID:
...
-
-: The payload of the packet.
-
 The following packet types are defined:
 
 | Type | Name                          | Section                     |
 |-----:|:------------------------------|:----------------------------|
 |  0x0 | Initial                       | {{packet-initial}}          |
-|  0x1 | 0-RTT Protected               | {{packet-protected}}        |
+|  0x1 | 0-RTT Protected               | {{packet-0rtt}}        |
```
martinthomson: I think that this is right, but counting is hard.

```suggestion
|  0x1 | 0-RTT Protected               | {{packet-0rtt}}             |
```

draft-ietf-quic-transport.md
```diff
@@ -3719,57 +3781,19 @@ Any data in CRYPTO frames is discarded - and no longer retransmitted - when
...
-the same keys, avoiding any risk of key and nonce reuse; this also prevents
-0-RTT packets from previous handshake attempts from being accepted as part of
-the connection.
+A 0-RTT Protected packet uses extended long headers with a type value of 0x1.
+It is used to carry "early" data from the client to the server as part of the
+first flight, prior to handshake completion.  As part of the TLS handshake, the
+server can accept or reject this early data.
 
+See Sections 2.3 of {{!TLS13}} for a discussion of 0-RTT data and its
+limitations.
```
martinthomson: You should move the 0-RTT packet numbers section to here and remove the TODO.

draft-ietf-quic-transport.md
```diff
@@ -3338,6 +3342,52 @@ then a packet containing a 16-bit value of 0x9b32 will be decoded as 0xa82f9b32.
 Example pseudo-code for packet number decoding can be found in
 {{sample-packet-number-decoding}}.
 
+### Starting Packet Numbers
```
martinthomson: I don't think that we need this still, but absent analysis, this can be moved under the Initial packet section.

draft-ietf-quic-transport.md
```diff
@@ -3795,119 +3819,84 @@ Handshake encryption level is discarded - and no longer retransmitted - when
...
 
-## Retry Packet {#packet-retry}
-
-A Retry packet uses a long packet header with a type value of 0x3. It carries
-an address validation token created by the server. It is used by a server that
-wishes to perform a stateless retry (see {{validate-handshake}}).
+## Short Header Packet {#short-header}
 
-~~~
+~~~~~
```
```suggestion
~~~
```

draft-ietf-quic-transport.md
```diff
@@ -3795,119 +3819,84 @@ Handshake encryption level is discarded - and no longer retransmitted - when
 Handshake protection keys are discarded.
 
 
-## Retry Packet {#packet-retry}
-
-A Retry packet uses a long packet header with a type value of 0x3. It carries
-an address validation token created by the server. It is used by a server that
-wishes to perform a stateless retry (see {{validate-handshake}}).
+## Short Header Packet {#short-header}
```
martinthomson: I'm assuming that the diff here looks terrible because it's moved text, not changed text.

draft-ietf-quic-transport.md
```diff
@@ -3674,7 +3707,37 @@ Token Length:
...
+: The packet number field is 1 to 4 bytes long. The packet number has
+  confidentiality protection separate from packet protection, as described in
+  Section 5.4 of {{QUIC-TLS}}. The length of the packet number field is encoded
+  in the plaintext packet number. See {{packet-encoding}} for details.
+
+Payload:
+
+: The payload of the packet.
+
+### Initial Packet {#packet-initial}
```
martinthomson: I think that we should use h2 for this.  The section is a mix of "here is a common format" and "here is a specific instance of a common format", but I think that's tolerable.  We would then have something like:

```
17. Packet Formats
    17.1. Packet Number Encoding and Decoding
    17.2. Long Header Packets <-- note the extra 's'
    17.3. Version Negotiation Packet
    17.4. Retry Packet
    17.5. Extended Long Header Packets
    17.6. Initial Packet
        17.6.1. Abandoning Initial Packets
        17.6.2. Starting Packet Numbers
    17.7. 0-RTT Protected Packet <-- extra 'Packet'
        17.7.1. 0-RTT Packet Numbers
    17.8. Handshake Packet
    17.9. Short Header Packets <-- extra 's'
```

Does that work?

MikeBishop: I understand the desire not to move subsections below h3, but the mix of "format" and "instance" really just doesn't flow for me.  But it does kind of match the rest of the document with the "grouped sections" motif, so maybe it's at least consistent.

(It's already contorted enough that VN is a long header packet, but not from this version and therefore absent from the type table.)

draft-ietf-quic-transport.md
MikeBishop: Yep.

draft-ietf-quic-transport.md
```diff
@@ -2553,20 +2553,18 @@ order to guarantee consistent state between endpoints.
...
-datagrams (see {{packet-coalesce}}) and have confidentiality and integrity
-protection (see {{packet-protected}}).
+QUIC endpoints communicate by exchanging packets. Packets have confidentiality
+and integrity protection (see {{packet-protected}}) and are carried in UDP
+datagrams (see {{packet-coalesce}}).
 
 This version of QUIC uses the long packet header (see {{long-header}}) during
-connection establishment and the short header (see {{short-header}}) once 1-RTT
-keys have been established.
+connection establishment.  Packets that carry the long header are Initial
```
```suggestion
 connection establishment.  Packets that start with the long header are Initial
```

ianswett: Or just "with the long header" as you say below.

draft-ietf-quic-transport.md
```diff
@@ -2602,15 +2600,21 @@ each packet sent, see {{packet-numbers}} for details.
 
 ## Coalescing Packets {#packet-coalesce}
 
-A sender can coalesce multiple QUIC packets into one UDP datagram.  This can
-reduce the number of UDP datagrams needed to complete the cryptographic
-handshake and starting sending data.  Receivers MUST be able to process
-coalesced packets.
+Packets using the Extended Long Header ({{extended-long-header}}) contain a
```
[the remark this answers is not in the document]
MikeBishop: In this format, they're the subsections of the ELH section.  We'd probably have to be more explicit if we take the non-indented layout @martinthomson proposes.

draft-ietf-quic-transport.md
```diff
@@ -3674,7 +3707,37 @@ Token Length:
 
 Token:
 
-: The value of the token.
+: The value of the token that was previously provided in a Retry packet or
```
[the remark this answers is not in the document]
MikeBishop: You're right, this fell in during the ELH combination.  So Initial packets *are* different because of the Token fields, which means that only 0-RTT and Handshake have the same ELH format.  Are we sure it's worth not simply having an explicit list of fields for those two?  That makes me more inclined to just revert the ELH change entirely and make each long-header packet type stand alone.

draft-ietf-quic-transport.md
```diff
@@ -2609,15 +2605,22 @@ details.
 
 ## Coalescing Packets {#packet-coalesce}
 
-A sender can coalesce multiple QUIC packets into one UDP datagram.  This can
-reduce the number of UDP datagrams needed to complete the cryptographic
-handshake and starting sending data.  Receivers MUST be able to process
-coalesced packets.
+Initial ({{packet-initial}}), 0-RTT ({{packet-0rtt}}), and Handshake
+({{packet-handshake}}) packets contain a Length field, which determines the end
```
DavidSchinazi: I think we might want to either rename the Length field, or at least clarify that it's in a different place based on the Long Header Type.

DavidSchinazi:
The comments in this PR discuss an "Extended Long Header" but I don't see it in the actual diff. Was it removed? I tried to address this in #2295 but I'm happy to have it be fixed here with an explicit extended long header.

MikeBishop:
Yes -- because it's subtly different in the different packets, I changed it to an overarching description of fields which occur in multiple packet types, but indicate which fields are present in each packet type's section.

draft-ietf-quic-transport.md
```diff
@@ -3451,136 +3436,64 @@ Source Connection ID:
...
-: The next bit (0x40) of byte 0 is set to 1.  Packets containing a zero value
-  for this bit are not valid packets in this version and MUST be discarded.
-
-Spin Bit (S):
-
-: The sixth bit (0x20) of byte 0 is the Latency Spin Bit, set as described in
-  {{!SPIN=I-D.ietf-quic-spin-exp}}.
+packet type.  While type-specific semantics for this version are described in
+the following sections, several long-header packets in this version of QUIC
+contain these additional fields:
```
martinthomson: Is there a picture to accompany this, or are you relying on the Initial/Handshake/0-RTT pictures?

draft-ietf-quic-transport.md
```diff
@@ -3736,23 +3651,37 @@ control state (see Sections 5.3.1.2 and 6.9 of {{QUIC-RECOVERY}}).
...
+Number Length bits.  It is used to carry "early" data from the client to the
+server as part of the first flight, prior to handshake completion. As part of
+the TLS handshake, the server can accept or reject this early data.
 
-### 0-RTT Packet Numbers {#retry-0rtt-pn}
-
-<!-- TODO: Thus far, we have not really addressed 0-RTT at all. We don't even
-really say much about what it is for. This really belongs in a section for that,
-probably right after the example handshake exchanges. -->
+See Sections 2.3 of {{!TLS13}} for a discussion of 0-RTT data and its
```
```suggestion
See Section 2.3 of {{!TLS13}} for a discussion of 0-RTT data and its
```

draft-ietf-quic-transport.md
```diff
@@ -3736,23 +3651,37 @@ control state (see Sections 5.3.1.2 and 6.9 of {{QUIC-RECOVERY}}).
...
+limitations.
 
+~~~
++-+-+-+-+-+-+-+-+
+|1|1| 0 |R R|P P|
```
```suggestion
|1|1| 1 |R R|P P|
```

draft-ietf-quic-transport.md
```diff
@@ -3785,11 +3714,33 @@ the same keys, avoiding any risk of key and nonce reuse; this also prevents
...
-used to carry acknowledgments and cryptographic handshake messages from the
-server and client.
+A Handshake packet uses long headers with a type value of 0x2, followed by the
+Length and Packet Number fields.  The first byte contains the Reserved and
+Packet Number Length bits.  It is used to carry acknowledgments and
+cryptographic handshake messages from the server and client.
+
+~~~
++-+-+-+-+-+-+-+-+
+|1|1| 0 |R R|P P|
```
```suggestion
|1|1| 2 |R R|P P|
```

draft-ietf-quic-transport.md
MikeBishop: Replying on the other pictures.

PR #2209, opened 2018-12-19
run DetectLostPackets before resetting crypto_count and pto_count

### LINKED ISSUE #2208: persistent congestion can never happen
The collapse of the congestion window (which used to happen after an RTO was verified) happens in `CongestionEvent`, if `pto_count > kPersistentCongestionThreshold`.
`CongestionEvent` is called from `OnPacketsLost`, which is called from `DetectLostPackets`, which in the case of a tail is only called when a new ACK is received from `OnAckReceived`. However, `OnAckReceived` resets the `pto_count` to 0 *before* calling `DetectLostPackets`, so the condition that leads to the collapse of the congestion window can never occur.

marten-seemann:
I'm not sure if this is correct, but I think this would fix #2208.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -711,10 +711,11 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
     for acked_packet in newly_acked_packets:
       OnPacketAcked(acked_packet.packet_number)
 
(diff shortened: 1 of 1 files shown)


janaiyengar:
Nice catch -- thanks @marten-seemann !

PR #2210, opened 2018-12-19
fix MUST not in idle period

marten-seemann:
I think "MUST NOT" was the intention of #2187, not "MUST not".


martinthomson:
@ianswett, you can just merge these.

Issue #2103, opened 2018-12-06
Guidelines for Retransmitting Lost Data


martinduke:
The recovery draft is oddly silent on when to send data that has actually been marked as lost. I'm not sure if this is constructive ambiguity or just an oversight.

Consider the following case with 1 stream, which is an entirely typical burst error scenario.
1. Segments A through J are sent in packets 1-10. There is no other data.
2. TLP timeout, send A in pkt 11.
3. TLP timeout, send pkt 12. **Do I send segment A or B** -- A is inflight twice so B is more likely to be new to the receiver, though if A is truly lost we won't make progress.
4. RTO timeout, send pkts 13 and 14. Similar to (3), **do we send C & D, or A & B**?
5. Receive ack of 14. Packets 1-12 are declared lost. Whatever was in pkt 13 is therefore the only outstanding data, and only the contents of pkt 14 are deleted from the queue. **Should I include the payload of pkt 13 in the outgoing queue,** or wait for pkt 13 to be declared lost?

Put another way, in the recovery draft we often prioritize segments that have no copies in-flight. Does that extend to prioritizing segments that are in-flight once over those that are out there twice? Does declaring a packet lost mean we should simply retx its segments if unacked, or do we consider that there are still outstanding copies?

janaiyengar:
Said differently, the draft says _when_ a sender can send, but not _what_.

There are rules of thumb we used to have, but I would argue now that this doesn't require strong recommendations besides that the sender decide based on local priorities. That said, I think it'd be good to have a paragraph that discusses this point more explicitly and the tradeoffs, so I'll leave this issue open.  It is not an oversight that we don't say _what_ should be transmitted, but there should be text clarifying it.

ianswett:
In the "Loss Detection" section, we reference the transport draft:

"If a packet is lost, the QUIC transport needs to recover from that loss, such as by retransmitting the data, sending an updated frame, or abandoning the frame. For more information, see Section 13.2 of {{QUIC-TRANSPORT}}."

We could certainly add some non-normative text about the considerations, particularly in regards to things like TLP, where sending the same payload in two consecutive TLPs seems a bit unwise, unless that's all that's outstanding.

ianswett:
The current text, which was written after this issue was filed is:
[...]

PR #2212, opened 2018-12-20
Proposed change to #1965

janaiyengar:
A proposal: I think I prefer to mark things as lost instead of suggesting that a sender SHOULD send something when there's really nothing to be sent.  I was starting to rephrase #1965 to say "SHOULD send ack-eliciting packet, but MAY mark as lost", and I needed to explain how a sender makes this choice. I realized that if it's been long enough (PTO is long enough) and there's nothing left to send, then maybe it is sensible to mark anything pending as lost. So, I changed my suggestion to that, which is this PR (against #1965).

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -498,12 +498,12 @@ priorities.
 
 When a PTO timer expires, new or previously-sent data may not available to send,
 and data may still be in flight.  Under these conditions, a sender MUST mark any
-packets still in flight as lost.  This can happen for example when data is sent on
-a stream which is then reset by a sender, and a PTO timer expires after this stream
-is reset.  A sender can be blocked from sending new data in the future if data is
-left in flight with no PTO armed.  Marking any in-flight packets as lost allows the
-sender to recover any congestion window space that is consumed by the packets still
-in flight.
+packets still in flight as lost.  This can happen for example when data is sent
+on a stream which is then reset by a sender, and a PTO timer expires after this
+stream is reset.  A sender can be blocked from sending new data in the future if
+data is left in flight with no PTO armed.  Marking any in-flight packets as lost
+allows the sender to recover any congestion window space that is consumed by the
+packets still in flight.
 
 
 #### Loss Detection {#pto-loss}
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -497,7 +497,7 @@ including sending new or retransmitted data based on the application's
 priorities.
 
 When a PTO timer expires, new or previously-sent data may not be available to
-send, and packets may still be in flight.  A sender can be blocked from sending
+send and packets may still be in flight.  A sender can be blocked from sending
 new data in the future if packets are left in flight.  Under these conditions, a
 sender SHOULD mark any packets still in flight as lost.  If a sender wishes to
 establish delivery of packets still in flight, it MAY send an ack-eliciting
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -497,13 +497,14 @@ including sending new or retransmitted data based on the application's
 priorities.
 
 When a PTO timer expires, new or previously-sent data may not available to send,
-and data may still be in flight.  Under these conditions, a sender MUST mark any
-packets still in flight as lost.  This can happen for example when data is sent
-on a stream which is then reset by a sender, and a PTO timer expires after this
-stream is reset.  A sender can be blocked from sending new data in the future if
-data is left in flight with no PTO armed.  Marking any in-flight packets as lost
-allows the sender to recover any congestion window space that is consumed by the
-packets still in flight.
+and data may still be in flight.  A sender can be blocked from sending new data
+in the future if data is left in flight with no PTO armed.  Under these
+conditions, a sender SHOULD mark any packets still in flight as lost.  Marking
+any in-flight packets as lost allows the sender to recover any congestion window
+space consumed by them.  Marking all packets as lost might be
+bandwidth-inefficient if a large amount of data is in flight.  A sender MAY
+optimize for bandwidth by sending an ack-eliciting packet and re-arming the PTO
+timer instead of marking all in-flight packets as lost.
 
 
 #### Loss Detection {#pto-loss}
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -496,11 +496,14 @@ MAY use alternate strategies for determining the content of probe packets,
...
-If no new data or unacknowledged data is available to send, an ack-eliciting
-packet SHOULD be sent.  Sending a retransmittable frame ensures that any in
-flight packets are acknowledged or declared lost in a timely manner.  If no
-ack-eliciting packet is sent, any packets currently in flight should be
-declared lost to avoid repeatedly arming and firing the PTO timer.
+When a PTO timer expires, new or previously-sent data may not available to send,
+and data may still be in flight.  Under these conditions, a sender MUST mark any
+packets still in flight as lost.  This can happen for example when data is sent
+on a stream which is then reset by a sender, and a PTO timer expires after this
+stream is reset.  A sender can be blocked from sending new data in the future if
```
nibanks: If the stream is reset, there is still ack-eliciting data pending, the RESET_STREAM frame. When that frame is sent, a new PTO would be sent.

Does this case that would trigger your scenaior: if you sent X (could be s STREAM frame or something else) and then a PTO fires and you retransmit X, but shortly after you receive an ACK for the first transmission. But then PTO fires for the retransmission.

janaiyengar: Hm, good point. So how can we end up in this situation then? Do we have any example?

janaiyengar:
@ianswett : If we say SHOULD, we need to say under what conditions an implementation might depart from the SHOULD. Specifically, in this case, we need to be able to say why an implementation might not declare all lost. I couldn't think of a good enough reason to not do this, which is why I went with a MUST.

I can remove the example, but if we can't come up with an example, this isn't a problem. @martinduke : did you have a better example?

martinduke:
I like this change, in principle.

I think the current editor's draft suffers from a little conceptual muddle. In TCP, losses, retransmissions, and congestion responses are all tightly coupled. For good reason, the draft backs off on the congestion response bit, and also delays losses, but not always retransmissions. As I indicated in #2103, separating retransmissions and losses creates some ambiguity. I like the idea of just declaring stuff lost. It's conceptually cleaner.

That said, declaring everything lost, rather than the things we're retransmitting, seems like overkill.

martinduke:
Correction: I like the change that means that all retransmitted packets are declared lost first. However, your change reopens the deadlock condition that @ianswett was trying to fix. That is, PTO, retransmission, ack of original packet, PTO. In this case there really is absolutely nothing to send, and nothing to declare lost. My impression was that @ianswett's change would encourage you to send a PING or something. That seems to have been lost in this PR.

martinduke:
To summarize, here is my preferred logic for the lossdetect timer:

if (new data available)
     send it in a new packet
else 
     declare in-flight packets lost and try to retransmit them, until you find one that still actually has contents
    if we didn't find any
           send a ping.

janaiyengar:
@martinduke : This PR has the outcome you are looking for. Basically, on a PTO, if you have something to retransmit, you will. If you have nothing to retransmit, Ian's PR suggests that you basically send a PING to elicit an ACK.  This PR calls them lost instead, so inflight is down to 0 immediately (if inflight is down to 0, you are not blocked anymore when you need to send new app data).

Basically, this PR says:
```
    if (new or unacked data available)
        send it in a new packet
    else
        mark unacked data as lost (give up on this data)
```
At the end, you either have a packet in flight (and a PTO armed), or you have nothing in flight (and no PTO armed).  The point is that you don't need to send anything if you don't have anything to send. You mark what was inflight as lost because it's actually reasonable to do so (it's been long enough).

In terms of #2103 , let's have that discussion there. I don't think there's conceptual muddling here. It's possible that I'm missing it entirely, but I don't see the muddle (yet :-)).

draft-ietf-quic-recovery.md
janaiyengar: @martinduke ^

mikkelfj:
Overall this makes sense, but what if you are on a fat pipe sending lots of data, and peer ACK's infrequently, and the last ACK is lost when the sender finishes sending. Then you have to retransmit a lot of data.

draft-ietf-quic-recovery.md
```diff
@@ -496,11 +496,14 @@ MAY use alternate strategies for determining the content of probe packets,
 including sending new or retransmitted data based on the application's
 priorities.
 
-If no new data or unacknowledged data is available to send, an ack-eliciting
-packet SHOULD be sent.  Sending a retransmittable frame ensures that any in
-flight packets are acknowledged or declared lost in a timely manner.  If no
-ack-eliciting packet is sent, any packets currently in flight should be
-declared lost to avoid repeatedly arming and firing the PTO timer.
+When a PTO timer expires, new or previously-sent data may not available to send,
```
```suggestion
When a PTO timer expires, new or previously-sent data may not be available to send,
```

draft-ietf-quic-recovery.md
```diff
@@ -496,11 +496,15 @@ MAY use alternate strategies for determining the content of probe packets,
...
+and data may still be in flight.  A sender can be blocked from sending new data
```
```suggestion
 and packets may still be in flight.  A sender can be blocked from sending new data
```

draft-ietf-quic-recovery.md
```diff
@@ -496,11 +496,15 @@ MAY use alternate strategies for determining the content of probe packets,
...
+in the future if data is left in flight with no PTO armed.  Under these
```
```suggestion
 in the future if packets are left in flight with no PTO armed.  Under these
```

draft-ietf-quic-recovery.md
```diff
@@ -496,11 +496,12 @@ MAY use alternate strategies for determining the content of probe packets,
 including sending new or retransmitted data based on the application's
 priorities.
 
-If no new data or unacknowledged data is available to send, an ack-eliciting
-packet SHOULD be sent.  Sending a retransmittable frame ensures that any in
-flight packets are acknowledged or declared lost in a timely manner.  If no
-ack-eliciting packet is sent, any packets currently in flight should be
-declared lost to avoid repeatedly arming and firing the PTO timer.
+When a PTO timer expires, new or previously-sent data may not be available to
+send, and data may still be in flight.  A sender can be blocked from sending new
```
```suggestion
 send, and packets may still be in flight.  A sender can be blocked from sending new
```

martinthomson:
For future reference: a helpful title ensures that creating changelogs is easier.

PR #2215, opened 2018-12-20
Don't verify ECN counts for out of order ACKs

### LINKED ISSUE #2198: More ECN verification failure modes
(From #2189)

There's a perverse situation in which ECN verification results in another problem.

That situation is where the largest acknowledged is lower on packet that is sent later. AND that packet acknowledges more packets so that the counts are higher. AND that packet is reordered ahead of packets with higher largest acknowledged. This results in the verification failing on the second packet to arrive (counts are lower because it's an older ACK, but the largest acknowledged is higher).

A fix is to never decrease the largest acknowledged when sending an ACK frame. Another is to disable validation if the packet containing the ACK is not higher-numbered than the last packet containing an ACK. The question is whether we care enough to recommend either fix...

@gloinul, WDYT?

ianswett:
Fixes #2198

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3045,9 +3045,7 @@ peer no longer supports ECN.
 Network devices that corrupt or apply non-standard ECN markings might result in
 reduced throughput or other undesirable side-effects.  To reduce this risk, an
 endpoint uses the following steps to verify the counts it receives in an ACK
-frame.  Counts MUST NOT be verified if the ACK frame is being processed in a
-different order than it was sent, as indicated by the packet number on the
-packets containing the ACK frames.
+frame.
 
 * The total increase in ECT(0), ECT(1), and CE counts MUST be no smaller than
   the total number of QUIC packets sent with an ECT codepoint that are newly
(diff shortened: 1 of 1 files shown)


ianswett:
There's probably slightly better phrasing for the packet number clause, so feel free to suggest something.

ianswett:
Jana's talked me through this and now I realize that this is broken in a larger sense.  As we've previously discussed, any time an acknowledgement is dropped, ECN validation MAY fail.

janaiyengar:
After more discussion, I agree that this does fix the problem. I think the text could be moved to a better location, but I'm fine with this for now.

mikkelfj:
"Processed in different order than it was sent" is a bit vague. If there is reordering you don't know which frame is the one that is out of order. There needs to be a clear selection criteria such as not processing a frame in packet with a PN lower than a previously processed.

janaiyengar:
@ianswett: I've reworded and moved the text. See what you think.

draft-ietf-quic-transport.md
```diff
@@ -3065,6 +3064,10 @@ to be greater than the number of packets acknowledged in an ACK frame.  When
 this happens, and if verification succeeds, the local reference counts MUST be
 increased to match the counts in the ACK frame.
 
+Processing counts out of order can result in verification failure.  An endpoint
+MUST NOT perform this verification if the ACK frame is received in a packet with
+packet number lower than a previously received ACK frame.
```
martinthomson: With the proposed change, add:

> Verifying based on ACK frames that arrive out of order can result in disabling ECN unnecessarily.

draft-ietf-quic-transport.md
```diff
@@ -3065,6 +3064,10 @@ to be greater than the number of packets acknowledged in an ACK frame.  When
 this happens, and if verification succeeds, the local reference counts MUST be
 increased to match the counts in the ACK frame.
 
+Processing counts out of order can result in verification failure.  An endpoint
+MUST NOT perform this verification if the ACK frame is received in a packet with
```
```suggestion
SHOULD NOT perform this verification if the ACK frame is received in a packet with
```

janaiyengar:
@martinthomson : I've incorporated your comments.

ianswett:
This still LGTM.

martinthomson:
Thanks for considering my input.

PR #2218, opened 2018-12-20
Clarify packet number representation

### LINKED ISSUE #2217: Packet Number encoding in draft-17
In Section 12.3, the transport specs says: 

> Where present, packet numbers are encoded as a variable-length integer (see Section 16). 

But in both 17.2 (Long Header) and 17.3 (Short Header), Packet Number field in the packet is noted as the following format:
> Packet Number (8/16/24/32)

which suggest it isn't variable-length integers.

Can we get some clarification regarding this? Thank you!

janaiyengar:
Closes #2217.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2639,7 +2639,7 @@ other packets in the same UDP datagram.
 
 ## Packet Numbers {#packet-numbers}
 
-The packet number is an integer in the range 0 to (2^62)-1.  This number is used
+The packet number is an integer in the range 0 to 2^62-1.  This number is used
 in determining the cryptographic nonce for packet protection.  Each endpoint
 maintains a separate packet number for sending and receiving.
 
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3297,9 +3300,10 @@ value of fields.
 
 ## Packet Number Encoding and Decoding {#packet-encoding}
 
-Packet numbers in long and short packet headers are encoded in 1 to 4 bytes.
-The number of bits required to represent the packet number is reduced by
-including the least significant bits of the packet number.
+Packet numbers are integers in the range 0 to 2^62-1 ({{packet-numbers}}).  When
+present in long or short packet headers, they are encoded in 1 to 4 bytes.  The
```
siyengar: do we have to say here that they are big endian encoded?

janaiyengar: We say that at the top, [here](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#packet-formats).

Issue #2232, opened 2018-12-20
Should receipt of multiple promises really be an error?


ekr:
S 4.2.6.

```
A server MUST NOT use a Push ID that is larger than the client has provided in a
MAX_PUSH_ID frame ({{frame-max-push-id}}) and MUST NOT use the same Push ID in
multiple PUSH_PROMISE frames.  A client MUST treat receipt of a PUSH_PROMISE
that contains a larger Push ID than the client has advertised or a Push ID which
has already been promised as a connection error of type HTTP_MALFORMED_FRAME.
```

Note that this doesn't match the convention we are using in QUIC transport
that things are idempotent and duplicates are fine.

martinthomson:
I think that this is OK being different due to the substrate being fundamentally different (i.e., these are reliably delivered as opposed to packets which are inherently unreliable).  But I'm happy to hear other arguments.

LPardue:
This changed recently for the -17 draft as a result of the discussion on https://github.com/quicwg/base-drafts/issues/1864.

In short, the old language introduced a requirement on the client to keep track of all PUSH_PROMISE  frames with same Push ID in order to verify that the header block was identical across them all. This was pretty ugly. 

-17 introduce the DUPLICATE_PUSH frame, which achieves the same functionality as multiple promises, whithout the client overhead. And as a result, this requirement to specifically disallow mutliple PUSH_PROMISEs was brought it.

Issue #1351, opened 2018-05-11
Reorder HTTP frame fields to put Type first


djc:
This would make parsing easier to deal with. I don't think there are adverse effects?

I'm proposing standardizing this:

```
    0                   1                   2                   3
    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |    Type (8)   |   Flags (8)   |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |       Length (i)                        ...
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |       Frame Payload (*)     ...
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```

instead of this:

```
    0                   1                   2                   3
    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |                           Length (i)                        ...
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |    Type (8)   |   Flags (8)   |       Frame Payload (*)     ...
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```

dtikhonov:
Why is one easier than the other?

djc:
With type first, you can directly dispatch to a type-specific parser. Additionally, it would be more similar to how QUIC packets work (header type in the first bit) and how QUIC frames work (type in the first byte).

dtikhonov:
Yes, but you still need to parse the length -- to skip unknown frames.  Even with a dispatch table, it is better to parse the length in one place rather than in each type-specific function.

djc:
[...]

PR #2235, opened 2018-12-20
Make HTTP frames a TLV format. This has several advantages:

ekr:
I know this is a late change, but I think it's an improvement....

- You can switch on the frame type before you even read a
  variable-length value
- You can have special-case (negotiated) frame types with no lengths

If this has already been discussed and rejected somewhere, maybe
I missed it.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -402,21 +402,22 @@ All frames have the following format:
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                           Length (i)                        ...
+|    Type (8)   |               Length (i)
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|    Type (8)   |               Frame Payload (*)             ...
+|                     Frame Payload (*)             ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ~~~~~~~~~~
 {: #fig-frame title="HTTP/3 frame format"}
 
 A frame includes the following fields:
 
+  Type:
+  : An 8-bit type for the frame.
+
+
   Length:
   : A variable-length integer that describes the length of the Frame Payload.
     This length does not include the Type field.
-
-  Type:
-  : An 8-bit type for the frame.
 
   Frame Payload:
   : A payload, the semantics of which are determined by the Type field.


ekr:
@DavidSchinazi

mikkelfj:
Me thinks it makes sense. As to 0-length special DATA frames (until end), I also think it could be a length-less separate type instead of 0-length. The downside is that you can't iterate frames without understanding all types, but that is no different from transport.

ekr:
I think the way you would make iterating all frames possible is by requiring explicit opt-in (or standards or something) for any types that didn't have the TLV format.

DavidSchinazi:
I like this. You could imagine reserving a part of the type space (e.g. 0-32) for non-TLV frames, and if you ever receive something in that range that you don't understand you fail the connection. But if you receive a TLV type you don't understand, you can safely ignore it since you know how long it is.

Minor: I'd make the type a variable-length integer

ekr:
I would be fine with type being a VLI.

dtikhonov:
We've had this discussion haven't we.

mikkelfj:
L=0 definitely was discussed in the context of whether only DATA frames or all frames could be 0 length. But that is a different issue.

martinthomson:
This has been discussed a few times.  We discussed this in h2 and explicitly made the opposite choice.  Though that was a while back, and I don't know if it was well documented.

ianswett:
This seems like a good change to me, I was rather confused when I realized the length came before the type.  If we're really going to drink the varint kool-aid, maybe the type should be a varint too?

kazuho:
I'm fine with moving frame type in front of the length.

OTOH, I am not sure if I prefer using varints for frame types.

IIRC, using varints for frame types of the transport, with the requirement that the minimal-sized encoding MUST be used, was based on the fact that it reserves more code points to future extensions, at the same time not making existing implementations complex. The reason the change did not make existing implementations complex was because new frame types cannot be used without negotiation.

The situation is different in H3.

H3 allows frames to be used without negotiation. In fact, we allow endpoints to grease the codepoints.

That means that every implementation would be required to support frame types using varints, even though in practice only the single-octet versions would be used.

To me that sounds like an unnecessary complexity.

I am also skeptical if we would ever run out of H3 frame types even if they are kept 8 bits, considering how few we have used in H2.

dtikhonov:
I don't see an advantage to having type come before length.  I would be against special-casing some frames to have no length, which is the only scenario where this change would be necessary for.

HTTP/2 frames have length before type; why change?  Seems gratuitous.

mikkelfj:
Nearly all binary self-describing formats introduced these days use some form of TLV because it offers the a good size/flexibility tradeoff, because the L and/or V part can omitted or specialised. For example there can be compact integers up to 4 bits in length using only a type field and a separate var length type. That is not in itself an argument for using it in HTTP/3.

LPardue:
I'm not sure about omitting length. I can see there could be benefit in stating that some frame types, that are part of H3 core, which contain only varints could be optimised by not including the length. 

However, from an implementation perspective it seems simpler to have code that can read frames out of the stream using the frame length.

dtikhonov:
The implementation is easy either way, which is why I ask: "why change?" The current frame format derives from HTTP/2, which makes me inclined to leave it as-is.

ekr:
On Mon, Dec 24, 2018 at 7:07 AM Dmitri Tikhonov <notifications@github.com>
wrote:

> The implementation is easy either way, which is why I ask: "why change?"
>
I listed the arguments for this in the original comments on the issue.

dtikhonov:
@ekr, I addressed the original comments above. The proposed change is only advantageous if we want to be able to special-case some frame types to exclude length. We should get a consensus on that first.

ekr:
No, I don't think we need consensus on that first, because (for instance) someone might want a private extension that permitted excluding the length, which the current design makes very difficult. TLV is just inherently more flexible. And so far the only argument I've heard against it is "why change".

dtikhonov:
Excluding the length has diminished benefit if it takes 8 bytes to encode the frame type (I am referring to #2253, with which this proposal is linked).

I am not against future-proofing the format, but both this and the varint type proposal seem very theoretical: _there might be someone who may want..._ etc.  For example, there have only been [two extensions](https://www.iana.org/assignments/http2-parameters/http2-parameters.xhtml#frame-type) to the HTTP/2 frames.  Has extending or experimenting with HTTP/2 frames ever been a real problem?

ianswett:
After thinking about this more, I think this is a very important change and making it will cause HTTP/3 align better with the QUIC transport.

In retrospect, PR#2098(Allow DATA to extend to the end of the stream) violated what appears to be a core design choice of HTTP/2, that all frames explicitly specify a length, so they're easily skippable.  Admittedly, skipping DATA doesn't make much sense, but it is still inconsistent.

To me, that's one of many signals that TLV is a better approach.  I was intending to create an extension that was DATA without length, but I realized that this new frame type would have a length byte, but it would be ignored or specified as a particular incorrect value(ie: 0), neither of which are ideal.

dtikhonov:
>  this is a very important change and making it will cause HTTP/3 align better with the QUIC transport.

1. What is meant by "better alignment?"
1. Why is it important?  QUIC and HTTP/3 are different protocols, we touched on this a bit in #2232.

ianswett:
@dtikhonov By better alignment, I meant more consistent design choices between QUIC transport and HTTP/3.

Previously HTTP over QUIC adopted varints because it was straightforward and QUIC transport had recently adopted them.  Moving HTTP/3 to TLV seems like a similar change.

More importantly, I'd like to argue that TLV is fundamentally better than LTV.  If someone can provide an example in HTTP/2 where LTV has shown clear value over TLV, I'd be interested in hearing about it.

dtikhonov:
> Previously HTTP over QUIC adopted varints because it was straightforward and QUIC transport had recently adopted them. Moving HTTP/3 to TLV seems like a similar change.

One does not follow from the other; TLV vs LTV and fixed-size vs varint are two different, orthogonal questions.

> If someone can provide an example in HTTP/2 where LTV has shown clear value over TLV, I'd be interested in hearing about it.

I'd be curious to know why HTTP/2 chose one over the other in the first place.

ekr:
On Mon, Dec 24, 2018 at 11:11 AM Dmitri Tikhonov <notifications@github.com>
wrote:

> Previously HTTP over QUIC adopted varints because it was straightforward
> and QUIC transport had recently adopted them. Moving HTTP/3 to TLV seems
> like a similar change.
>
> One does not follow from the other; TLV vs LTV and fixed-size vs varint
> are two different, orthogonal questions.
>

This is true at some level, but once we've changed anything, we might as
well ask what the best design is.

ianswett:
The varint question is completely orthogonal, sorry to bring it up here. @martinthomson filed a separate issue for that: #2253

DavidSchinazi:
We could have h3 use both TV and TLV: for example we say that
- all even types are TV, and must be negotiated before use (or in base spec)
- all odd types are TLV, and can be used without negotiation and ignored on receipt if unknown

This allows efficient encoding while preserving extensibility.

mikkelfj:
let's not over-engineer more than we already have

janaiyengar:
I wasn't in the room where it happened, but I can imagine one reason for doing LTV in HTTP/2 might have been that you need to read out frames from the underlying byte-stream before demuxing and parsing them. A frame reader only needs Length before reading out the entire frame and passing it on to the parser. QUIC is different in that the entire frame is present when the length is read.

I think this is a simple enough change, and there are straightforward benefits. We expect to see some changes with more implementation experience in the HTTP mapping, so I'd like to push for this PR.

@ekr -- this discussion should be happening on an issue. Can you open an issue ?

dtikhonov:
@janaiyengar writes, in part (emphasis added):
> QUIC is different in that the _entire frame is present_ when the length is read.

I don't understand what you mean by that.

LPardue:
Yeah I don't follow either. The h3 layer is presented an in-order byte stream for each stream but it still needs to parse out H3 frames. While there are some loose assurances on frame ordering I don't know it QUIC is so different from H2, even with separate streams

janaiyengar:
@dtikhonov, @LPardue: Urgh, yes, sorry, I was not thinking straight. I was incorrect.

LTV or TLV, I don't have a strong preference either way.

martinthomson:
I interpreted @janaiyengar's comment to mean that when you read a QUIC frame it's in a datagram, so it's probably all sitting in memory.  With a streamed frame, you might not have all of the frame in place.  You might only have the length.

LPardue:
Yeah I think we are all in agreement on that.

MikeBishop:
The previous discussion was #1351.  The conclusion there was that the order doesn't matter a great deal (under the assumption that you need both), and so consistency with HTTP/2 was the tie-break.

If there's a desire to enable length-less frames, that's an argument for TLV.  However, I don't think switching on the type value first is terribly compelling -- there doesn't seem to be a great difference between:
- passing your frame parser the type and expecting it to tell you how many bytes of the stream it processed and whether the frame was valid
- passing your frame parser the type and length and expecting it to tell you if the frame was mis-sized

janaiyengar:
@martinthomson 
> I interpreted @janaiyengar's comment to mean that when you read a QUIC frame it's in a datagram, so it's probably all sitting in memory. With a streamed frame, you might not have all of the frame in place. You might only have the length.

Yes, that's exactly what I meant, but this discussion is about HTTP frames, which is what I had confused in my mind.

draft-ietf-quic-http.md
```diff
@@ -402,22 +402,23 @@ All frames have the following format:
...
 
 A frame includes the following fields:
 
+  Type:
+  : An 8-bit type for the frame.
+
+
   Length:
   : A variable-length integer that describes the length of the Frame Payload.
     This length does not include the Type field.
```
mikkelfj: With reordering the "Length" description no longer makes sense. If anything, length should state that it does not include itself.

PR #2236, opened 2018-12-20
Omit size on ODCIL

### LINKED ISSUE #2211: "ODCIL(4" looks like a typo
Section 17.7. Retry Packet has "ODCIL(4".

I understand it doesn't fit in the diagram but it looks ugly.

martinthomson:
Which is probably less bad than the current.

Closes #2211.


dtikhonov:
What if we simply `s/ODCIL/ODCL/` to recover a byte for the closing parenthesis?

martinthomson:
@dtikhonov, I considered that.  We haven't put sizes on the other fields in this byte, so this seems tolerable.

PR #2238, opened 2018-12-21
s/RTO/PTO/g and consequences

### LINKED ISSUE #2206: RTO->PTO in transport
1789:Retransmittion Timeout (RTO) as defined in {{QUIC-RECOVERY}} is RECOMMENDED.
2182:(RTO) interval as defined in {{QUIC-RECOVERY}}.
2253:draining state before the packets arrive.  If a peer could timeout within an RTO

martinthomson:
Closes #2206.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1790,8 +1790,13 @@ abandons its attempt to validate the path.
 
 Endpoints SHOULD abandon path validation based on a timer. When setting this
 timer, implementations are cautioned that the new path could have a longer
-round-trip time than the original.  A value of three times the current
-Retransmittion Timeout (RTO) as defined in {{QUIC-RECOVERY}} is RECOMMENDED.
+round-trip time than the original.  A value of three times the larger of the
+current Probe Timeout (PTO) or the initial timeout (that is, kInitialRtt*2) as
+defined in {{QUIC-RECOVERY}} is RECOMMENDED.  That is:
+
+~~~
+   validation_timeout = max(3*PTO, 6*kInitialRtt)
+~~~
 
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1790,8 +1790,13 @@ abandons its attempt to validate the path.
 
 Endpoints SHOULD abandon path validation based on a timer. When setting this
 timer, implementations are cautioned that the new path could have a longer
-round-trip time than the original.  A value of three times the current
-Retransmittion Timeout (RTO) as defined in {{QUIC-RECOVERY}} is RECOMMENDED.
+round-trip time than the original.  A value of three times the larger of the
+current Probe Timeout (PTO) or the initial timeout (that is, kInitialRtt*2) as
```
ianswett: I think you might be conflating one type of path validation, which is prior to handshake completion, with post-handshake path validation?  Or is there an intent to this text I am missing?

martinthomson: Yes.  The point - as discussed - was that the PTO is based on the RTT of the old path and so might have little relationship with the RTT of the new path.  If the new RTT is much longer, we might time out unnecessarily.

When we had a minimum RTO, that was our safety net.  But PTO has no minimum, so we have no such safety net.  The initial timeout is therefore our stand-in for RTT.

ianswett: Thanks, after re-reading, that makes sense.

```suggestion
 current Probe Timeout (PTO) or the initial timeout (that is, 2*kInitialRtt) as
```

martinthomson: grr, leading spaces on suggestions are the worst

PR #2240, opened 2018-12-21
Adjust autotuning recommendations

### LINKED ISSUE #2083: Recommend autotuning for flow control
```
This document leaves when and how many bytes to advertise in a MAX_STREAM_DATA
or MAX_DATA frame to implementations, but offers a few considerations.  These
frames contribute to connection overhead.  Therefore frequently sending frames
with small changes is undesirable.  At the same time, larger increments to
limits are necessary to avoid blocking if updates are less frequent, requiring
larger resource commitments at the receiver.  Thus there is a trade-off between
resource commitment and overhead when determining how large a limit is
advertised.

A receiver MAY use an autotuning mechanism to tune the frequency and amount of
advertised additional credit based on a round-trip time estimate and the rate at
which the receiving application consumes data, similar to common TCP
implementations.
```

In light of the first paragraph which says that I can do whatever I want, what is the purpose of this MAY? It seems like you should either recommend that people do this or just remove the normative language.

martinthomson:
This doesn't need normative text.

This adds @ianswett's recommendation about only sending flow control
frames when there are other frames to send.  Though I added another note
about the peer being explicitly blocked, which I think is a useful
transition to the next paragraph.

Closes #2083.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -766,9 +766,9 @@ advertised.
 A receiver can use an autotuning mechanism to tune the frequency and amount of
 advertised additional credit based on a round-trip time estimate and the rate at
 which the receiving application consumes data, similar to common TCP
-implementations.  Sending frames related to flow control only when there are
-other frames to send or when a peer is blocked ensures that flow control doesn't
-cause extra packets to be sent.
+implementations.  As an optimization, sending frames related to flow control
+only when there are other frames to send or when a peer is blocked ensures that
+flow control doesn't cause extra packets to be sent.
 
 If a sender runs out of flow control credit, it will be unable to send new data
 and is considered blocked.  It is generally considered best to not let the
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -763,10 +763,12 @@ larger resource commitments at the receiver.  Thus there is a trade-off between
 resource commitment and overhead when determining how large a limit is
 advertised.
 
-A receiver MAY use an autotuning mechanism to tune the frequency and amount of
+A receiver can use an autotuning mechanism to tune the frequency and amount of
 advertised additional credit based on a round-trip time estimate and the rate at
 which the receiving application consumes data, similar to common TCP
-implementations.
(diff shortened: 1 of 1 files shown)


martinthomson:
I don't think so.  Flow control credits can go out with acknowledgments.  The advice is there so that you don't have the MAX_DATA frame go out immediately, with a delayed ACK going out 20ms later.  They can both go out together.

marten-seemann:
Bundling frames definitely makes sense, but I don’t think we need the “or when a peer is blocked” part.

martinthomson:
Your assertion being that a blocked peer will already have to live with being blocked for a round trip, so the gain isn't worth documenting?

marten-seemann:
I’m saying that you shouldn’t let your peer become blocked under normal circumstances.
I just noticed that in the corner case where the peer is blocked because the application is consuming data too slowly, it actually makes sense to not bundle the flow control update. So I guess the text is fine as is.

draft-ietf-quic-transport.md
```diff
@@ -763,10 +763,12 @@ larger resource commitments at the receiver.  Thus there is a trade-off between
 resource commitment and overhead when determining how large a limit is
 advertised.
 
-A receiver MAY use an autotuning mechanism to tune the frequency and amount of
+A receiver can use an autotuning mechanism to tune the frequency and amount of
 advertised additional credit based on a round-trip time estimate and the rate at
 which the receiving application consumes data, similar to common TCP
-implementations.
+implementations.  Sending frames related to flow control only when there are
+other frames to send or when a peer is blocked ensures that flow control doesn't
```
janaiyengar: This sentence is very vague, and I'm not sure I find it useful. What issue is it tackling?

ianswett: It's an optimization I've suggested a few times to ensure sending more frequent updates doesn't have substantial overhead.

janaiyengar: Suggestion: "As an optimization, sending frames ..."

janaiyengar:
Looks fine, one suggestion.

PR #2241, opened 2018-12-21
Remove PATH_RESPONSE limits

### LINKED ISSUE #2129: Limits on PATH_CHALLENGE/RESPONSE rate
```On receiving a PATH_CHALLENGE frame, an endpoint MUST respond immediately by
echoing the data contained in the PATH_CHALLENGE frame in a PATH_RESPONSE frame.
However, because a PATH_CHALLENGE might be sent from a spoofed address, an
endpoint MUST limit the rate at which it sends PATH_RESPONSE frames and MAY
silently discard PATH_CHALLENGE frames that would cause it to respond at a
higher rate.
```

It's pretty hard to understand what the normative force of this is, given that it doesn't explain what the limit is. I think there's actually a simple rule we should propose, which is one PATH_CHALLENGE frame per packet (with others to be ignored). Given that we require duplicate suppression at the packet level, that will have an amplification factor of 1:1, which you can get by a variety of other mechanisms.

martinthomson:
Implementations can do this anyway, so we're not getting much from the
recommendation.  And we already have good limits on what can be sent
(including PATH_RESPONSE) in the migration section.

Closes #2129.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1748,10 +1748,6 @@ it can associate the peer's response with the corresponding PATH_CHALLENGE.
 
 On receiving a PATH_CHALLENGE frame, an endpoint MUST respond immediately by
 echoing the data contained in the PATH_CHALLENGE frame in a PATH_RESPONSE frame.
-However, because a PATH_CHALLENGE might be sent from a spoofed address, an
-endpoint MUST limit the rate at which it sends PATH_RESPONSE frames and MAY
-silently discard PATH_CHALLENGE frames that would cause it to respond at a
-higher rate.
 
 To ensure that packets can be both sent to and received from the peer, the
 PATH_RESPONSE MUST be sent on the same path as the triggering PATH_CHALLENGE.


mikkelfj:
> An endpoint MAY bundle PATH_CHALLENGE and PATH_RESPONSE frames that are used for path validation with other frames. In particular, an endpoint may pad a packet carrying a PATH_CHALLENGE for PMTU discovery, or an endpoint may bundle a PATH_RESPONSE with its own PATH_CHALLENGE.

This can lead to amplification attack if not careful.

janaiyengar:
@mikkelfj: how can this lead to amplification?

mikkelfj:
> @mikkelfj: how can this lead to amplification?

If an endpoint seeing a challenge uses the opportunity to discover if its preferred MTU is supported by padding the response (as opposed to the challenge) even if the challenge packet is small. Since this might fail, it like will also send a small response after a delay or after a new challenge if it doesn't get an ACK before that.

A defence could be to ensure that challenges are no larger than, say twice, response.

mikkelfj:
An endpoint may also optimistically bundle all kinds of other data in a response. Even with 1200 bytes, this can be much larger than a small challenge packet without considering PMTU discovery at all.

mikkelfj:
I think I raised this issue quite a while ago this is the reason the limit was introduced, although I didn't formulate the exact counter measure.

martinthomson:
The counter measure we have against amplification is the initial congestion window.  We won't send more than that.  That's higher than the x3 we have during the handshake, but I think that's a reasonable balance, considering that the attacker in this context has to spend a considerable amount of effort.

Yes, this is a policy question, but I'm happy with it.  (Though we should probably highlight this when asking for review.)

PR #2242, opened 2018-12-21
Use sending and receiving parts for stream states

### LINKED ISSUE #2081: Stream terminology
The text in 3.1 and 3.2 is inconsistent about using "sending part" and "send stream". Because bidi streams aren't a pair of uni streams, this is confusing. It would help to either clarify or use "sending part" or something.

martinthomson:
I had partially settled on "send stream" and "received stream", but on
reflection these are inaccurate labels and using "parts" is better, even
if it is a little more clumsy at times.

Closes #2081.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -371,7 +371,7 @@ Note:
   with an implementation that implements these states.
 
 
-## Send Stream States {#stream-send-states}
+## Sending Stream States {#stream-send-states}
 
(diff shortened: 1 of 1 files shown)


RyanAtGoogle:
It you find "parts" to be clumsy would "side" be any cleaner?

martinthomson:
:) "sending side" or "sending part" is just clumsy.  Nothing to be done without minting a new term, which I will not do because it's just not that many instances to worry about.

PR #2244, opened 2018-12-21
Congestion window reduced on ACK

### LINKED ISSUE #2243: description of persistent congestion threshold is wrong
The text says
> If a threshold number of consecutive PTOs have occurred (pto_count is more than kPersistentCongestionThreshold, see Section 7.9.1), the network is considered to be experiencing persistent congestion, and the sender’s congestion window MUST be reduced to the minimum congestion window.

Similar text exists for the description of kPersistentCongestionThreshold.

We actually only collapse the congestion window when receiving an ACK for a probe packet, and only if this ACK leads us to declare any packet(s) lost.
Specifically, this means that pto_count may grow larger than kPersistentCongestionThreshold, if no acknowledgements are received.

janaiyengar:
Closes #2243.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1036,10 +1036,11 @@ kLossReductionFactor:
 kPersistentCongestionThreshold:
 : Number of consecutive PTOs after which network might be considered to be
   experiencing persistent congestion.  The rationale for this threshold is to
-  enable a sender to use initial PTOs for aggressive probing, similar to Tail
-  Loss Probe (TLP) in TCP {{TLP}} {{RACK}}.  The RECOMMENDED value for
-  kPersistentCongestionThreshold is 2, which is equivalent to having two TLPs
-  before an RTO in TCP.
+  enable a sender to use initial PTOs for aggressive probing, as TCP does with
+  Tail Loss Probe (TLP) {{TLP}} {{RACK}}, before establishing persistent
+  congestion, as TCP does with a Retransmission Timeout (RTO) {{?RFC5681}}.  The
+  RECOMMENDED value for kPersistentCongestionThreshold is 2, which is equivalent
+  to having two TLPs before an RTO in TCP.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -1023,15 +1027,13 @@ kLossReductionFactor:
   The RECOMMENDED value is 0.5.
 
 kPersistentCongestionThreshold:
-: Number of consecutive PTOs after which network is considered to be
+: Number of consecutive PTOs after which network might be considered to be
```
```suggestion
: Number of consecutive PTOs after which the network is considered to be
```

I prefer "is" instead of "might be" given you already have considered.

draft-ietf-quic-recovery.md
```diff
@@ -1023,15 +1027,13 @@ kLossReductionFactor:
...
-  reaches this threshold - that is, persistent congestion is established - the
-  sender responds by collapsing its congestion window to kMinimumWindow, similar
-  to a Retransmission Timeout (RTO) in TCP {{RFC5681}}.  The RECOMMENDED value
-  for kPersistentCongestionThreshold is 2, which is equivalent to having two
-  TLPs before an RTO in TCP.
+  enable a sender to use initial PTOs for aggressive probing, as TCP does with
+  Tail Loss Probe (TLP) {{TLP}} {{RACK}}, before establishing persistent
+  congestion, as TCP does with a Retransmission Timeout (RTO) {{?RFC5681}}.  The
+  RECOMMENDED value for kPersistentCongestionThreshold is 2, which is equivalent
+  to having two TLPs before an RTO in TCP.
```
```suggestion
  to two TLPs before an RTO in TCP.
```

draft-ietf-quic-recovery.md
janaiyengar: I don't want to say "is" because that implies that persistent congestion has happened. As Marten notes, that is only done when the ACK is received.

janaiyengar:
I've reworded and I think the wording is better now. I'll merge, but do comment if you think the wording needs fixing.

PR #2248, opened 2018-12-23
Alternative Largest Reference Decoding algorithm.

bencebeky:
A Largest Reference Decoding algorithm with two fewer branches (but one more modulo operation).

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -825,17 +825,11 @@ The decoder reconstructs the Largest Reference using the following algorithm:
 
 ~~~
    if LargestReference > 0:
-      LargestReference -= 1
-      CurrentWrapped = TotalNumberOfInserts mod (2 * MaxEntries)
-
-      if CurrentWrapped >= LargestReference + MaxEntries:
-         # Largest Reference wrapped around 1 extra time
-         LargestReference += 2 * MaxEntries
-      else if CurrentWrapped + MaxEntries < LargestReference
-         # Decoder wrapped around 1 extra time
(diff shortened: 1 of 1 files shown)


bencebeky:
This algorithm should be safe from overflow and underflow when using signed arithmetic, as long as the same width integers are used for SETTINGS_TABLE_SIZE and Largest Reference on the wire and TotalInsertionCount, because MaxEntries is bounded from above by the maximum representable number divided by 32.

When using unsigned integers, one can add 3 * MaxEntries instead of 1 * Maxentries in the first line to avoid underflow in the second line, and can break up the last line into two, checking before doing the subtraction.  A sane decoder would need to do some checking anyway to make sure that the resulting Largest Reference is at least 1.

bencebeky:
> The problem with code like this is that it is indecipherable. It's short, but I don't think that it improves readability much. Not that this code is ever readable...

I appreciate your concern.  Indeed if one is focused on the mechanics of how the decoder or Largest Reference wraps around and what that does to different quantities, the current code is easier to read.

On the other hand, if you focus on the requirement that "Largest Reference is the unique integer that is congruent with wire Largest Reference - 1 modulo 2 * MaxEntries and falls in the interval [TotalNumberOfInserts - MaxEntries, TotalNumberOfInserts + MaxEntries)", then the proposed code is easier to read.

So I believe readability is largely dependent on the mental model that readers have in their heads.

bencebeky:
Since this PR is not clearly an improvement, and it did not gain much traction, I am abandoning it so that we can focus on more important issues.

PR #2249, opened 2018-12-23
Invalid Largest Reference is a stream error.

bencebeky:
If Largest Reference does not figure, then the decoded Base Index will probably be bogus, resulting in corrupt entries.  I think it's better to close the connection with an error.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -864,13 +864,13 @@ decoder's dynamic table:
    if EncodedInsertCount == 0:
       ReqInsertCount = 0
    else:
-      InsertCount = EncodedInsertCount - 1
+      ReqInsertCount = EncodedInsertCount - 1
       CurrentWrapped = TotalNumberOfInserts mod (2 * MaxEntries)
 
-      if CurrentWrapped >= InsertCount + MaxEntries:
+      if CurrentWrapped >= ReqInsertCount + MaxEntries:
          # Insert Count wrapped around 1 extra time
          ReqInsertCount += 2 * MaxEntries
-      else if CurrentWrapped + MaxEntries < InsertCount:
+      else if CurrentWrapped + MaxEntries < ReqInsertCount:
          # Decoder wrapped around 1 extra time
          CurrentWrapped += 2 * MaxEntries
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -825,22 +825,32 @@ The decoder reconstructs the Largest Reference using the following algorithm:
...
          LargestReference += 2 * MaxEntries
-      else if CurrentWrapped + MaxEntries < LargestReference
+      else if CurrentWrapped + MaxEntries < LargestReference:
          # Decoder wrapped around 1 extra time
          CurrentWrapped += 2 * MaxEntries
 
-      LargestReference += TotalNumberOfInserts - CurrentWrapped
+      LargestReference += TotalNumberOfInserts
+
+      if LargestReference <= CurrentWrapped:
```
bencebeky: I put the error checking here because I am using unsigned integers.  If the majority of implementors are expected to use signed integers (after all, 2**62 required by the spec can comfortably be represented using 64 bit signed integers), then the original `LargestReference += TotalNumberOfInserts - CurrentWrapped` line can be kept and the error checking woul look like:

```
   if LargestReference <= 0:
      Error
```

martinthomson:
This will conflict with #2248.

bencebeky:
> This will conflict with #2248.

No worries, I do not expect all of my PRs to be approved (even though I think they are all improvements).  I'm just drafting them to start a discussion, because for small things like this it is much easier to talk about an actual PR then to try describing the changes.

If there is consensus to merge either this or #2248, I'm happy to rebase the other one.

MikeBishop:
I'm confused -- if LR is invalid, the HEADERS frame can't be interpreted, I agree.  It's unclear why a failure to decode anything on a request stream could constitute an encoder stream error.

bencebeky:
> I'm confused -- if LR is invalid, the HEADERS frame can't be interpreted, I agree. It's unclear why a failure to decode anything on a request stream could constitute an encoder stream error.

Good point.  Let's make it a stream error (with the usual code HTTP_QPACK_DECOMPRESSION_FAILED) then.

bencebeky:
I rebased on #2111, incorporated the suggestion by @MikeBishop above to make it a stream error instead of a connection error, and reconciled that sometimes InsertCount and sometimes ReqInsertCount was used in the decoding algorithm.

afrind:
overtaken by events

PR #2250, opened 2018-12-23
Call Largest Reference on the wire something else.

bencebeky:
It moderately bothers me that the term "Largest Reference" is used for the actual concept as well as the transformed on-the-wire value.  Using separate terms might make implementing less error-prone.


bencebeky:
The spec currently implies that 0 is sent on the wire as 0, and that a 0 wire value is decoded as 0, but it's very implicit.  Another advantage of this PR is to cover this case explicitly.

martinthomson:
OBE.

bencebeky:
This PR was obsoleted by #2111.

PR #2251, opened 2018-12-23
Explicitly talk about Largest Reference zero.

bencebeky:
At two places, explicitly spell out that Largest Reference is zero if there are no dynamic entries; also remove "Largest Reference" from a sentence that currently implies that it could be set to any value.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -142,7 +142,8 @@ Base Index:
 
 Largest Reference:
 
-: The largest absolute index of an entry referenced in a header block.
+: The largest absolute dynamic table index referenced in a header block, or zero
+  if the header block does not reference any dynamic table entries.
 
 QPACK is a name, not an acronym.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -874,8 +876,7 @@ Largest Reference and Base Index to the same value.  In such case, both the sign
 bit and the Delta Base Index will be set to zero.
 
 A header block that does not reference the dynamic table can use any value for
-Base Index; setting both Largest Reference and Base Index to zero is the most
```
bencebeky: I propose removing "Largest Reference" from this sentence because setting it to zero is not a choice.

Optionally one can start the paragraph by "A header block that does not reference the dynamic table MUST use zero for Largest Reference but can use any value...".

martinthomson: You can encode a Largest Reference of zero.  That's the problem with having an overloaded concept.  But in practice both can be zero.  0 on the wire means that there are no references; 0 the abstract notion is a sentinel value only, so maybe that's what is confusing.

bencebeky: I think the word "can" is misleading in your sentence "You can encode a Largest Reference of zero", because whether the encoded Largest Reference is zero is not a choice.  I have the same problem with the current wording of the spec, that's exactly why I'm proposing this change.

In my mind, 0 for the abstract notion means that there are no references.  (Is this what you mean by "sentinel value"?)  The spec already calls this out at least in one place, and I'm proposing more clarifications at issue #2251.

In my mind, 0 on the wire does not mean anything.  It has to be decoded.  It just happens so that 0 is mapped to 0, and it just happens to that the current wording of the spec is that 0 is not transcribed, but what that actually means is that 0 is transcribed to 0.

In any case, we having this discussion here is an indication that the text needs some kind of improvement.

afrind:
I think this is overtaken by #2111.  Largest Reference is dead.  Long live Required Insert Count.

PR #2197, opened 2018-12-17
Reorganize ACK section

martinthomson:
I was planning to make only two changes here, but there were a few
editorial itches I needed to scratch, so this does:

1. renames ACK Blocks to ACK Ranges

2. moves the First ACK Range field to the top level

3. renames the ECN Section to ECN Counters and adds a description of
   that field

4. uses the field names more consistently (s/Gaps/Gap fields/ and so
   forth)

5. makes figure and list intros more consistent

PR #2254, opened 2018-12-24
Change ACK Range to ACK Block

### LINKED ISSUE #2178: Rename ACK block something else?
An ACK frame currently consists of:

- A header
- An ACK block count (N)
- A field called ACK blocks which consists of
   - A single field called "First ack block" (if N > 0)
   - N-1 repeats of a pair of fields called respectively "gap" and "Additional ACK block"

(Please correct me if I have misread this).

It's a bit weird to call these fields "blocks" because really they are the length of something, rather than a block. Perhaps "ACK Range"?

ianswett:
Fixes #2178

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2853,7 +2853,7 @@ needing acknowledgement are received.  The sender can use the receiver's
 Strategies and implications of the frequency of generating acknowledgments are
 discussed in more detail in {{QUIC-RECOVERY}}.
 
-To limit ACK Blocks to those that have not yet been received by the sender,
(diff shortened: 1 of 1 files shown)


martinthomson:
The title of this PR is precisely backwards.  It's also a strict subset of #2197, which I'd encourage you to look at instead.

janaiyengar:
Let's work with #2197.

ianswett:
I'd rather have more small changes.  I find #2197 large and hard to fully understand the implications of.

ianswett:
Ok, well #2197 was merged, so this is obsolete.

PR #2256, opened 2018-12-26
Initial maximum table size is zero.

bencebeky:
Remove ambigous or confusing phrases "corresponding setting when HTTP requests or responses are first permitted to be sent" and "the value of the setting in the peer’s SETTINGS frame"; define initial maximum table size to be zero.


martinthomson:
This text is right, but I fear that the design is bad.  Oh well.

afrind:
I think we need to include something about the SETTING in there.  #2257 or #2330

PR #2257, opened 2018-12-26
Initial maximum table size is SETTINGS_HEADER_TABLE_SIZE.

bencebeky:
Remove ambigous or confusing phrases "corresponding setting when HTTP requests or responses are first permitted to be sent" and "the value of the setting in the peer’s SETTINGS frame"; define initial maximum table size to be value of SETTINGS_HEADER_TABLE_SIZE sent by the decoder.


martinthomson:
I think that I prefer this to #2257.  At least it is consistent.

bencebeky:
I belive that #2330 is a cleaner and less error-prone alternative, therefore I withdraw this PR in favor of #2330.

PR #2271, opened 2018-12-28
stateless reset packet minimum size is 39 bytes

dtikhonov:


martinthomson:
Gah.  These numbers will haunt me...

PR #2277, opened 2018-12-31
s/created/received/ stream for CANCEL_PUSH

### LINKED ISSUE #2234: Why is CANCEL_PUSH limited to being used before the stream was created?
```
The CANCEL_PUSH frame (type=0x3) is used to request cancellation of a server
push prior to the push stream being created.  The CANCEL_PUSH frame identifies a
server push by Push ID (see {{frame-push-promise}}), encoded as a
variable-length integer.

When a server receives this frame, it aborts sending the response for the
send a QUIC RESET_STREAM frame on that stream and cease transmission of the
response.
```

Why is this limited to being prior to the push stream being created, given
that you specify how the server handles it if it has created the stream.

martinthomson:
Not sure if this is enough, but it enacts Brad's suggested fix.

Closes #2234.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -573,8 +573,8 @@ HTTP_UNEXPECTED_FRAME.
 ### CANCEL_PUSH {#frame-cancel-push}
 
 The CANCEL_PUSH frame (type=0x3) is used to request cancellation of a server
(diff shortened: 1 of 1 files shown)


bslassey:
Yup, that's what I had in mind.

PR #2280, opened 2018-12-31
Clarify Actions on nonzero Reserved Bits

martinduke:
As currently written, any garbage packet, after removing header protection, will have nonzero reserved bits and bring the connection down. Not good!

Similarly, it is perverse to force me to actually decrypt the full packet to see if it turns out to be a misbehaving end host and comply with the other possible reading of this requirement. Let the receiver just drop the bad packet after header decryption and carry on.


martinthomson:
This requirement exists because if you only remove header protection then throw the packet away, you create a side channel that allows an attacker some leverage on your header protection keys.  [The TLS draft](https://quicwg.org/base-drafts/draft-ietf-quic-tls.html#rfc.section.9.3) says:

>  For authentication to be free from side-channels, the entire process of header protection removal, packet number recovery, and packet protection removal MUST be applied together without timing and other side-channels.

Yes, it's perverse, but there are lots of cases where extra work is done to avoid creating side channels.  This ain't static RSA and the attack here isn't a Bleichenbacker attack, but there are good reasons.  And if you are suggesting that this sort of timing can't be observed, that's proven not to be the case before (see [NetSpectre](https://misc0110.net/web/files/netspectre.pdf)).

As for a DoS vulnerability, there are two bits, so you will be doing the work for 1 in four packets anyway.  And it's not hard to find packets with valid sequences.

martinduke:
Ah, so the intent was to apply it after full packet decryption is complete. I'll change the PR to clarify this.

draft-ietf-quic-transport.md
```diff
@@ -3402,9 +3402,10 @@ Reserved Bits (R):
 
 : The next two bits (those with a mask of 0x0c) of byte 0 are reserved.  These
   bits are protected using header protection (see Section 5.4 of {{QUIC-TLS}}).
-  The value included prior to protection MUST be set to 0.  An endpoint MUST
-  treat receipt of a packet that has a non-zero value for these bits after
-  removing protection as a connection error of type PROTOCOL_VIOLATION.
+  The value included prior to protection MUST be set to 0.  An endpoint MUST treat
+  receipt of a packet that has a non-zero value for these bits after removing
+  packet (not just header) protection as a connection error of type
```
marten-seemann: I don't think this is very clear to read. Do we really need to say anything, given that the timing side channel is described in detail in the TLS document?

martinduke: Oof. I didn't recall that section of the TLS draft. Thanks for pointing that out.

There are a couple of dangers here:
1) The implementer will not check "security considerations" in the tls draft when implementing first byte validation in the transport draft.
2) Naively, the implementation will choose to discard the packet and abort the connection after removing header protection, which opens up a DoS vector, even if goes through the motions of decrypting the packet anyway.

This is coming down to an editorial matter, and I agree my sentence is somewhat clunky. I'm open to other wording, but I don't think the current draft is satisfactory.

martinthomson: Let's start with an issue then, shall we?

martinduke: Aren't we down to wordsmithing at this point?

janaiyengar: Isn't "packet protection" enough? I don't think you need to say "not just header"

janaiyengar:
Not relevant to this PR, but @martinthomson : there are other side-channels in QUIC sadly. One could simply (at a high level) observe whether an ACK is sent or not, for instance.

martinduke:
I think my change makes it clearer. If people think this is a bad change, I'll kill the PR.

draft-ietf-quic-transport.md
martinthomson: It's both.  My suggestion would be:

```suggestion
  packet and header protection as a connection error of type
```

draft-ietf-quic-transport.md
```diff
@@ -3529,8 +3530,8 @@ Reserved Bits (R):
   bits are protected using header protection (see Section 5.4 of
   {{QUIC-TLS}}).  The value included prior to protection MUST be set to 0.  An
   endpoint MUST treat receipt of a packet that has a non-zero value for these
-  bits after removing protection as a connection error of type
-  PROTOCOL_VIOLATION.
+  bits after removing packet (not just header) protection as a connection error
```
```suggestion
  bits after removing packet and header protection as a connection error
```

draft-ietf-quic-transport.md
ianswett: @martinthomson 's suggestion is good, but I'm confused why you need to remove packet protection to know the bits are not set to 0.  Isn't removing header protection sufficient?

martinduke:
See MT's first comment. We need consistent-time crypto results.

martinduke:
@ianswett yeah, this PR is trying to make it clear, because I also find it non-obvious (and buried in the security considerations section of the TLS draft).

While this PR makes it crystal clear what is supposed to happen, I can easily see a developer thinking they're smarter than the spec and discarding after header protection. I wouldn't mind some explanatory text, but one thing at a time.

DavidSchinazi:
Out of curiosity, why are we validating these bits in the first place? Can't we just say receivers MUST ignore them? That removes timing attacks and also allows for future extensibility. I get that these are not inside the full encryption boundary but the associated data + header protection boundary should be good enough to prevent middleboxes from looking at or messing with these bits.

marten-seemann:
@DavidSchinazi We had this discussion before in #2022, and we decided to require a fixed value.

kazuho:
@DavidSchinazi I think it is a good practice to forbid the endpoints using the reserved bits without negotiation, because the bits could mean different things between them. They should negotiate the use of bits (that would allow then to accept non-zero values), then use them.

ianswett:
@kazuho Given these are within the encrypted envelope, if one side thinks they mean something and the other doesn't or vice versa, there are sure to be problems, so any usage would have to be negotiated as you say.

@marten-seemann I couldn't find that in #2022.  My reading of that discussion was header protection is better than just doing packet number encryption.

I don't have a strong opinion here, but I'm also not seeing much reason to specify what the values need to be and saying they can be anything seems simpler.  Saying they have to be 0 invites people to short-cut the full decryption and drop packets that have other values immediately.  To be clear, I probably would have done that if I hadn't read this issue.

kazuho:
@ianswett 
> Saying they have to be 0 invites people to short-cut the full decryption and drop packets that have other values immediately.

Honestly speaking, I am not sure short-cutting in case the zero check fails would lead to side channel vulnerability, considering the fact that a failed check immediately ends up in connection termination. The only information an attacker can obtain is that the sender failed to set the bits correctly.

Anyways, I do not think that removing the zero-check dissolves the vulnerability to the side-channel attack, because any use of the bits protected by header protection prior to AEAD-decrypting the payload is forbidden. The real attack exists on the PN field, where an attacker can flip the bits of the encrypted PN and use the timing difference of the receiver that short-cuts AEAD decryption based on the value of the PN to infer what the value of the unprotected PN was.

Therefore, I think that we should spend our effort to be clear about when the unprotected header bits can be used, rather than trying to fixing a use-case that does not have a high impact.

ianswett:
In that case, let's land this and move onto more important issues.

martinduke:
@ianswett My previous reply was too glib. One problem with applying the check right after removing header protection is that the consequences are a connection error -- so a garbage packet has a 3 in 4 chance of taking down the connection.

There are three equilbria here:
1) Simply discard the packet after header protection. As @marten-seemann pointed out, it's not clear why this would provide any sort of intelligence to an attacker, as there is no response.
2) After packet protection, it's a PROTOCOL_VIOLATION. This is the current state of the draft with this PR serving to clarify a bit.
3) Not having any requirement for the reserved bits. There is a difference in opinion over whether this is a side channel issue or not, or if this should be done only with some sort of negotiation. I'm not the one to articulate this positon.

Which is all to say that I should have taken @martinthomson 's advice a week ago and filed and issue.

martinduke:
OK, I'm happy with this now. I added some explanatory text to say "no, really, don't cut corners here," which I believe addresses the concern that @ianswett and I have.

martinduke:
Also, please reclassify this PR as "editorial"

martinthomson:
Thanks for persisting Martin.  This looks good.

martinthomson:
And here is me fixing the issue myself.  I merged this, minus the last change manually.

PR #2281, opened 2019-01-01
Include handshake anti-deadlock logic in pseudocode

Ralith:
Recovery §6.2.1 instructs:
> Because the server could be blocked until more packets are received, the client MUST start the crypto retransmission timer even if there is no unacknowledged CRYPTO data. If the timer expires and the client has no CRYPTO data to retransmit and does not have Handshake keys, it SHOULD send an Initial packet in a UDP datagram of at least 1200 bytes. If the client has Handshake keys, it SHOULD send a Handshake packet.

but this was not reflected in the pseudocode. I think these changes reflect the intended behavior.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1135,10 +1135,10 @@ OnLossDetectionTimeout():
   else if (endpoint is client without 1-RTT keys):
     // Send anti-deadlock packet: padded to earn more anti-
     // amplification credit, unpadded to prove address ownership.
-    if (will send Initial packet):
-       SendOnePaddedPacket()
+    if (has Handshake keys):
+       SendOneHandshakePacket()
      else:
-       SendOnePacket()
+       SendOnePaddedInitialPacket()
     crypto_count++
   else:
     // PTO
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1078,19 +1078,14 @@ GetEarliestLossTime():
   return time, space
 
 SetLossDetectionTimer():
-  // Don't arm timer if there are no ack-eliciting packets
-  // in flight.
-  if (no ack-eliciting packets in flight):
-    loss_detection_timer.cancel()
-    return
-
   loss_time, _ = GetEarliestLossTime()
   if (loss_time != 0):
     // Time threshold loss detection.
     loss_detection_timer.update(loss_time)
     return
 
-  if (crypto packets are in flight):
+  if (crypto packets are in flight
+      || endpoint is client without 1-RTT keys):
     // Crypto retransmission timer.
     if (smoothed_rtt == 0):
       timeout = 2 * kInitialRtt
@@ -1100,6 +1095,12 @@ SetLossDetectionTimer():
     timeout = timeout * (2 ^ crypto_count)
     loss_detection_timer.update(
       time_of_last_sent_crypto_packet + timeout)
+    return
+
+  // Don't arm timer if there are no ack-eliciting packets
+  // in flight.
+  if (no ack-eliciting packets in flight):
+    loss_detection_timer.cancel()
     return
 
   // Calculate PTO duration
@@ -1130,6 +1131,14 @@ OnLossDetectionTimeout():
   else if (crypto packets are in flight):
     // Crypto retransmission timeout.
     RetransmitUnackedCryptoData()
+    crypto_count++
+  else if (endpoint is client without 1-RTT keys):
+    // Send anti-deadlock packet: padded to earn more anti-
+    // amplification credit, unpadded to prove address ownership.
+    if (will send Initial packet):
+       SendOnePaddedPacket()
+     else:
+       SendOnePacket()
     crypto_count++
   else:
     // PTO


draft-ietf-quic-recovery.md
```diff
@@ -823,6 +824,9 @@ Pseudocode for OnLossDetectionTimeout follows:
        // Crypto retransmission timeout.
        RetransmitUnackedCryptoData()
        crypto_count++
+     else if (handshaking client)
+       SendOnePacket()
```
martinthomson: Is this right?
```suggestion
        SendPaddedInitialPacket()
```

Or is it this?
```suggestion
        SendPaddedPacket()
```

martinthomson:
The x3 rule does refer to simple bytes as opposed to bytes in flight as calculated for the purposes of congestion control.  So PADDING is completely sufficient for the client.

My point was that this extra "code" exists only for the purposes of driving that x3 rule, and is only necessary when the client is sending Initial packets.  If it is sending Handshake packets, there is no need because any Handshake packet of any size is sufficient to validate the client address and disable the x3 rule.

(BTW, this wasn't directly a comment on this PR, but on the cited text as well as the code.)

Ralith:
> BTW, this wasn't directly a comment on this PR, but on the cited text as well as the code.

Right, but I think the currently-required behavior is justified. Consider this case:
1. Client sends Initial w/ CRYPTO
2. Server sends Initial w/ CRYPTO, ACK, followed by Handshake flight; Handshake flight is lost in full
3. Client establishes Handshake keys; sends Initial w/ ACK
4. Server Handshake packets continue to be lost until the 3x budget is spent.

Because the client's crypto packets have all been ACKed, and it has received no further ack-eliciting packets, it has nothing to send.

If the client's second Initial was also lost, meaning the budget is only about 3600 bytes, this can happen almost immediately; a single Quinn server Handshake flight is larger than that.

martinthomson:
The client can get Handshake keys in this case, but it won't be sending Handshake packets until it has all of the server Handshake packets, so I think we're still in the original state.  This is still the case that there are outstanding Initial packets.

If the client were to be sending Handshake packets, this rule - as written - would still apply, but my point is that it isn't necessary.

Ralith:
> it won't be sending Handshake packets until it has all of the server Handshake packets

Is that accurate? Transport §17.6 has:
> Once a client has received a Handshake packet from a server, it uses Handshake packets to send subsequent cryptographic handshake messages and acknowledgments to the server.

It seems that client might reasonably send Handshake-level ACKs (for example) prior to receiving the server's full flight even under ideal network conditions.

> This is still the case that there are outstanding Initial packets.

Ah, it wasn't clear to me that the client's second, ack-only, Initial packet qualified as "in flight"/"outstanding" here. Even then, the client discards the Initial space after first sending a Handshake packet, which might not be ack-eliciting, and might become lost, so a padding-only packet might still be necessary.

martinthomson:
Right, I had forgotten.  Handshake ACKs are sent earlier.  Either way, once that starts, the risk of deadlocking is gone.

> Ah, it wasn't clear to me that the client's second, ack-only, Initial packet qualified as "in flight"/"outstanding" here. 

Good point.  The current text says "When crypto packets are sent, the sender MUST set a timer for the crypto timeout period."  And crypto packets don't include ACK-only Initial packets.

That suggests a bigger problem with that text.  I would suggest then that the condition is "Initial packets outstanding and no packet from a higher encryption level ever sent".

Ralith:
On review, the client certainly doesn't need to pad Handshake level anti-deadlock packets to 1200 bytes, though it must send them so long as it hasn't seen an ACK or sent anything that it would otherwise be required to retransmit.

draft-ietf-quic-recovery.md
```diff
@@ -778,13 +778,7 @@ Pseudocode for SetLossDetectionTimer follows:
...
 ~~~
  SetLossDetectionTimer():
-    // Don't arm timer if there are no ack-eliciting packets
-    // in flight.
-    if (no ack-eliciting packets in flight):
-      loss_detection_timer.cancel()
-      return
-
-    if (crypto packets are in flight):
+    if (crypto packets are in flight || handshaking client):
```
ianswett: I agree the pseudocode doesn't match the text, but I think the term "handshaking client" is not specific enough a definition to use in the pseudocode.

Ralith: Yeah, I wasn't sure what to go with. Should I introduce a new defined variable?

Ralith:
Updated to use an explicit variable, and to use unpadded packets at Handshake level.

draft-ietf-quic-recovery.md
```diff
@@ -699,6 +703,9 @@ Pseudocode for OnAckReceived and UpdateRtt follow:
     largest_acked_packet = max(largest_acked_packet,
                                ack.largest_acked)
 
+    if (from Handshake packet):
```
```suggestion
    if (ack from Handshake packet):
```

ianswett: Can you add a short comment like "Once the client receives an ACK in a Handshake packet, it knows the server considers its address verified."?

draft-ietf-quic-recovery.md
```diff
@@ -823,6 +831,12 @@ Pseudocode for OnLossDetectionTimeout follows:
        // Crypto retransmission timeout.
        RetransmitUnackedCryptoData()
        crypto_count++
+     else if (unverified_client):
+       if (handshake keys available)
+         SendOnePacket()
+       else:
+         SendPaddedPacket()
```
```suggestion
         SendOnePaddedPacket()
```

draft-ietf-quic-recovery.md
Ralith: > The client can know that it is unverified if it is sending Initial packets and not Handshake (or greater).

That's not true, though: the client doesn't know it's been validated until it's seen an ACK using Handshake keys (or greater). The server needs to see a Handshake packet to validate the client, and until the client's seen such an ACK it can't be sure that any Handshake packets it might have sent so far weren't lost.

> As for the test in SetLossDetectionTimer() or LossDetectionTimeout(), how is it possible for a crypto packet to be in flight, where this condition would be true?

I'm not sure I understand the question. The tests in those functions are to ensure that a client sends packets to either earn anti-amplification credit or validate its address even when it has no information to send.

Ralith:
Added @ianswett's tweaks and relaxed the validation check slightly, since a client could conceivably receive a 1-RTT ACK before a Handshake ACK due to even a single packet's worth of reordering.

draft-ietf-quic-recovery.md
Ralith: Are you suggesting that we ditch `unverified_client` in favor of checking whether the endpoint is a client that doesn't yet have 1-RTT? I think that works as a simplification, if I'm understanding correctly.

martinthomson: I think that works.  For the most part "do I have crypto packets in flight" covers the rest of it.  Though "in flight" might be wrong, and it might be better as "are any crypto packets that I have sent, but are unacknowledged".

The cost is that the client might keep this timer well past the point it is needed when acknowledgments are lost, but the simplification is pretty big.

Ralith:
Rebased and switched to the simplified logic as suggested.

draft-ietf-quic-recovery.md
```diff
@@ -1122,15 +1123,21 @@ Pseudocode for OnLossDetectionTimeout follows:
...
-    DetectLostPackets(pn_space)
   // Retransmit crypto data if no packets were lost
   // and there are still crypto packets in flight.
-  else if (crypto packets are in flight):
+  if (crypto packets are in flight):
     // Crypto retransmission timeout.
     RetransmitUnackedCryptoData()
     crypto_count++
+  else if (endpoint is client without 1-RTT keys):
+    if (handshake keys available)
```
```suggestion
    if (handshake keys available):
```

Unfortunately, this is still incorrect.  As we discussed on https://github.com/quicwg/base-drafts/issues/2519#issuecomment-472671420 there is a requirement to pad any time the client needs to send an Initial packet (it's a little more complicated in practice, but this is what the resolution boils down to).

draft-ietf-quic-recovery.md
```diff
@@ -1122,15 +1123,21 @@ Pseudocode for OnLossDetectionTimeout follows:
...
   loss_time, pn_space = GetEarliestLossTime()
-  if (loss_time != 0):
-    // Time threshold loss Detection
-    DetectLostPackets(pn_space)
   // Retransmit crypto data if no packets were lost
   // and there are still crypto packets in flight.
-  else if (crypto packets are in flight):
+  if (crypto packets are in flight):
     // Crypto retransmission timeout.
     RetransmitUnackedCryptoData()
```
martinthomson: I assume that RetransmitUnackedCryptoData() is going to catch any need for padding...

draft-ietf-quic-recovery.md
Ralith: If the availability of Handshake keys isn't a strong enough condition for that, what should be used instead? I'd be inclined to just fall back on `SendOnePacket` and let the nontrivial padding logic be implied by the text elsewhere, but that's where we started for this fragment.

martinthomson: If you need to send an Initial packet, pad.

draft-ietf-quic-recovery.md
```diff
@@ -1122,15 +1123,21 @@ Pseudocode for OnLossDetectionTimeout follows:
...
     crypto_count++
+  else if (endpoint is client without 1-RTT keys):
+    if (will send Initial packet):
```
ianswett: I'd say
"if (has Handshake keys):
  SendOneHandshakePacket()
 else
  SendOnePaddedInitialPacket()"

draft-ietf-quic-recovery.md
```diff
@@ -1078,6 +1078,19 @@ GetEarliestLossTime():
   return time, space
 
 SetLossDetectionTimer():
+  if (crypto packets are in flight
```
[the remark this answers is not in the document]
Ralith: Thanks, I was confused about the intended behavior here.

draft-ietf-quic-recovery.md
Ralith: @martinthomson, is this consistent with your earlier suggestions?

martinthomson: No. The important factor is not whether you have handshake keys, but whether your peer believes your address to be validated.

ianswett: To be clear, my suggestion started at line 1133.  The intent was to change "Will send Initial packet" to "has Handshake keys", which I believe is clearer.

martinthomson: "Will send Initial packet" is correct.

ianswett: @martinthomson Why?  In particular, do you think it is more correct?

This is already in a conditional that indicate the 1RTT keys aren't present.

@Ralith please add a comment above this section as well, to explain why it's here, since it may not be obvious.

janaiyengar: I think I understand @martinthomson's point -- you want to keep sending initial packets until you know that the server has received and processed a handshake packet, which is when the server considers the client validated. This would happen when the client receives an ACK in a Handshake packet (for a Handshake packet) or has the 1-RTT keys.

That said, I agree with @ianswett that "will send Initial packet" is an odd predicate. I think we can say (has 1-RTT keys or has received an ACK in a Handshake packet). I could be wrong, but I think this captures the condition we're looking for.

Ralith: That sounds like the approach that I originally had here, which was [argued against](https://github.com/quicwg/base-drafts/pull/2281/files/8088a86a41852b071d3fee371357da4cf56e3a12#discussion_r248056667).

ianswett: Martin didn't want to allow for ACK of handshake stopping the anti-amplification retransmissions, which I think is fine, though a bit overly strict.  So all of this is within the "if (endpoint is client without 1-RTT keys)" conditional, which is our proxy for: Should we keep trying to prevent a deadlock?

To be clear on what I'm suggesting:

else if (endpoint is client without 1-RTT keys):
  // The client must continue sending packets until it's
  // sure the server has validated the path.
  if (has Handshake keys):
    SendOneHandshakePacket()
  else
    SendOnePaddedInitialPacket()

janaiyengar: I see. Your proposal seems good to me. @martinthomson: can you clarify what you think is missing?

martinthomson: Striking out the bit that we've agreed to remove, the text from the transport draft is:

> Clients MUST ensure that UDP datagrams containing Initial packets are sized to at least 1200 bytes, adding padding to packets in the datagram as necessary. ~~Once a client has received an acknowledgment for a Handshake packet it MAY send smaller datagrams.~~ Sending padded datagrams ensures that the server is not overly constrained by the amplification restriction.

This contradicts that text.  It's true that a server that received a Handshake packet is no longer subject to the anti-DoS limit on bytes transmitted, but it's a bit weird that we have an algorithm that directly contradicts a MUST-level requirement in the other document.

janaiyengar: @martinthomson: AIUI, you'd remember this better since it was your PR, but wasn't the bit we agreed to remove was about special-casing padding of Initial packets? My reading of @ianswett 's proposed text doesn't contradict that.

PR #2295, opened 2019-01-04
Move some long header fields to more appropriate definitions

DavidSchinazi:
The current definition of the long header defines fields that are not used by half of the long packet types (Initial packets have a token length and token, and retry packets do not have a length). This change moves those fields to the appropriate long packet types for correctness.


martinthomson:
My plan here is to close this with #2203.  If there are other changes you want to add, please open another PR once #2203 lands.

DavidSchinazi:
Works for me.

PR #2296, opened 2019-01-04
Allow servers to send a preferred address of each address family

### LINKED ISSUE #2122: QUIC connection migration and IPv6 only NAT64/DNS64 Networks
Here's one issue that I believe should be addressed in the final version.

Client device has both cellular and wifi interfaces.
Server has dual stack deployment.

1. Client is connected to cellular network that has IPv6 only NAT64/DNS64 network. There's no WiFi connectivity.

2. Client issues DNS queries (A and AAAA) for the server. Given that server is dual stack, DNS64 returns actual AAAA and A record responses.

3. Client can't assume that the A and AAAA records both necessarily are hosted by the same node or would eventually be directed to the same node.

4. Client runs happy eyeballs (or a version of it) and ends up establishing QUIC connection over IPv6.

5. WiFi comes up but with only IPv4 access connectivity.

6. Client should be allowed to migrate to WiFi (either because it is less expensive or cellular goes away), however client doesn't have information IPv4 address hosted by the server.

It seems we control frames should be defined (like ADD_ADDR and REMOVE_ADDR) to communicate the address availability between two nodes anytime during a connection. Wanted to get more feedback on this before we send out a diff to the draft.

DavidSchinazi:
This change allows servers to send a preferred address of each address family. The previous definition of preferred_address only allowed sending one address family, and this had the disadvantage of picking the non-deterministic address family used by the Initial if the client is performing Happy Eyeballs. Clients may now use the best address family based on their network attachment.

Fixes #2122.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -4103,13 +4103,14 @@ preferred_address (0x000d):
   of this transport parameter is the PreferredAddress struct shown in
   {{fig-preffered-address}}.  This transport parameter is only sent by a server.
   Servers MAY choose to only send a preferred address of one address family by
-  sending a zero-length address for the other family.
+  sending an all-zero address and port (0.0.0.0:0 or ::.0) for the other family.
 
 ~~~
    struct {
-     opaque ipv4Address<0..2^8-1>;
-     opaque ipv6Address<0..2^8-1>;
-     uint16 port;
+     opaque ipv4Address[4];
+     uint16 ipv4Port;
+     opaque ipv6Address[16];
+     uint16 ipv6Port;
      opaque connectionId<0..18>;
      opaque statelessResetToken[16];
    } PreferredAddress;
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2173,6 +2173,9 @@ receiving a probe packet from a different address.  Servers MUST NOT send more
 than a minimum congestion window's worth of non-probing packets to the new
 address before path validation is complete.
 
+When a client wishes to migrate to a new local address that is of a different
+address family from the currently used server address, the client can attempt
+migration with the server's preferred address from the matching address family.
 
 # Connection Termination {#termination}
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2119,9 +2119,10 @@ transport parameter in the TLS handshake.
 Servers MAY communicate a preferred address of each address family (IPv4 and
 IPv6) to allow clients to pick the one most suited to their network attachment.
 
-Once the handshake is finished, the client SHOULD initiate path validation (see
-{{migrate-validate}}) of the server's preferred address using the connection ID
-provided in the preferred_address transport parameter.
+Once the handshake is finished, the client SHOULD select one of the two
+server's preferred addresses and initiate path validation (see
+{{migrate-validate}}) of that address using the connection ID provided in the
+preferred_address transport parameter.
 
 If path validation succeeds, the client SHOULD immediately begin sending all
 future packets to the new server address using the new connection ID and
@@ -4101,7 +4102,7 @@ preferred_address (0x000d):
 : The server's preferred address is used to effect a change in server address at
   the end of the handshake, as described in {{preferred-address}}.  The format
   of this transport parameter is the PreferredAddress struct shown in
-  {{fig-preffered-address}}.  This transport parameter is only sent by a server.
+  {{fig-preferred-address}}.  This transport parameter is only sent by a server.
   Servers MAY choose to only send a preferred address of one address family by
   sending an all-zero address and port (0.0.0.0:0 or ::.0) for the other family.
 
@@ -4115,7 +4116,7 @@ preferred_address (0x000d):
      opaque statelessResetToken[16];
    } PreferredAddress;
 ~~~
-{: #fig-preffered-address title="Preferred Address format"}
+{: #fig-preferred-address title="Preferred Address format"}
 
 If present, transport parameters that set initial flow control limits
 (initial_max_stream_data_bidi_local, initial_max_stream_data_bidi_remote, and
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2173,9 +2173,9 @@ receiving a probe packet from a different address.  Servers MUST NOT send more
 than a minimum congestion window's worth of non-probing packets to the new
 address before path validation is complete.
 
-When a client wishes to migrate to a new local address that is of a different
-address family from the currently used server address, the client can attempt
-migration with the server's preferred address from the matching address family.
+A client that migrates to a new address SHOULD use a preferred address from the
+same address family for the server.
+
 
 # Connection Termination {#termination}
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2116,6 +2116,9 @@ packets.
 A server conveys a preferred address by including the preferred_address
 transport parameter in the TLS handshake.
 
+Servers MAY communicate a preferred address of each address family (IPv4 and
+IPv6) to allow clients to pick the one most suited to their network attachment.
+
 Once the handshake is finished, the client SHOULD initiate path validation (see
 {{migrate-validate}}) of the server's preferred address using the connection ID
 provided in the preferred_address transport parameter.
@@ -4099,11 +4102,13 @@ preferred_address (0x000d):
   the end of the handshake, as described in {{preferred-address}}.  The format
   of this transport parameter is the PreferredAddress struct shown in
   {{fig-preffered-address}}.  This transport parameter is only sent by a server.
+  Servers MAY choose to only send a preferred address of one address family by
+  sending a zero-length address for the other family.
 
 ~~~
    struct {
-     enum { IPv4(4), IPv6(6), (15) } ipVersion;
-     opaque ipAddress<4..2^8-1>;
+     opaque ipv4Address<0..2^8-1>;
+     opaque ipv6Address<0..2^8-1>;
      uint16 port;
      opaque connectionId<0..18>;
      opaque statelessResetToken[16];


draft-ietf-quic-transport.md
```diff
@@ -2116,6 +2116,9 @@ packets.
 A server conveys a preferred address by including the preferred_address
 transport parameter in the TLS handshake.
 
+Servers MAY communicate a preferred address of each address family (IPv4 and
```
janaiyengar: I think I would prefer to call this an "alternate" address.

ianswett: If you're speaking of "preferred" address, this is consistent with the rest of the text.  If you want to change the name, I'd suggest opening a separate issue.

I'm happy with preferred, since if it's just alternate, why would the client change?  Alternate implies no preference, when in fact the server does prefer the other address.

draft-ietf-quic-transport.md
```diff
@@ -4099,11 +4102,13 @@ preferred_address (0x000d):
...
   of this transport parameter is the PreferredAddress struct shown in
   {{fig-preffered-address}}.  This transport parameter is only sent by a server.
+  Servers MAY choose to only send a preferred address of one address family by
+  sending a zero-length address for the other family.
 
 ~~~
    struct {
-     enum { IPv4(4), IPv6(6), (15) } ipVersion;
-     opaque ipAddress<4..2^8-1>;
+     opaque ipv4Address<0..2^8-1>;
```
ianswett: Aren't these two fields now fixed sizes?  If so, any reason not to specify those?

martinthomson: Yep.  I think that the invalid address (0.0.0.0 or ::) is probably OK here.

draft-ietf-quic-transport.md
martinthomson: "preferred" seems right.  Yes, it is an alternative, but the fact that it's a preferred alternative is what is important.

draft-ietf-quic-transport.md
kazuho: IIUC the text says that the value should be a zero-length byte array if the address type is not used, rather than using an invalid address.

Therefore I think the correct notation is:
```
opaque ipv4Address<0..4>;
opaque ipv6Address<0..16>;
```

DavidSchinazi:
@kazuho I went with this approach to minimize changes from the current design. Having a list that allows different ports and future versions of IP are fine by me, but there seemed to be more pushback when that was proposed in #2122. As long as we have a way of communicating one IPv4 address *and* one IPv6 address I'm happy.

draft-ietf-quic-transport.md
ianswett: @kazuho suggestion SGTM.

draft-ietf-quic-transport.md
```diff
@@ -4099,11 +4102,13 @@ preferred_address (0x000d):
...
+     opaque ipv6Address<0..2^8-1>;
      uint16 port;
```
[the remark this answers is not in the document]
ianswett: I'm fine with only supporting v4 and v6.

rpaulo:
We could allow multiple PreferredAddress TPs and avoid the zero-length address problem.

kazuho:
@DavidSchinazi Thank you for the clarification. I think I'm fine with that.

@rpaulo The PreferredAddress structure includes Connection ID and Stateless Reset Token. I think we should avoid requiring the server to send them twice when it is constrained by the 3-packet rule.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
martinthomson: And I'm ok with having the same port across versions.  Which all suggests a smaller change:
```
ipv4Address[4];
ipv6Address[16];
uint16 port;
```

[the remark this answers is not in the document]
marten-seemann: Why would the ipv4 and ipv6 port numbers be the same? That seems like a special case derived from the HTTP use case.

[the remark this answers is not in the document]
ianswett: If there's a real use case for separate port numbers, I can see adding that flexibility, but otherwise I prefer @martinthomson 's suggestion.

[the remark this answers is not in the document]
marten-seemann: p2p applications might not assign any meaning to ports, but just use a random port number. There's no reason why those values would be the same for IPv4 and IPv6.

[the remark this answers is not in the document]
mikkelfj: Yes, this seems like an unnecessary constraint. Especially since the mechanism can be used to advertise non-standard ports. Random ports could also be a first  line of defence against DoS attacks and they could be used to move endpoints dynamically over time. Forcing two different interfaces to be in sync isn't ideal for that.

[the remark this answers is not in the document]
kazuho: This might be a dumb question, but would P2P applications use the feature, considering that the intent is to remap clients connecting to an anycast address to a unicast address?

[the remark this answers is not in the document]
mikkelfj: I think the intent is to deal with anycast, but I could see it being used defensively against attacks.

As for P2P there is a strong case for both ends to use a known fixed port. However, I could also see this as a way to load balance multiple connections to multiple threads or processes.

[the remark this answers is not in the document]
DavidSchinazi: @marten-seemann most implementations (on the operating systems I'm familiar with) will use a single dual-stack listening socket for all address families so unless you go out of your way you will end up with the same port. That's true of HTTP servers and P2P apps.

That said, if there is consensus that separate ports are useful, I'm OK adding separate ports to this PR.

DavidSchinazi:
I've simplified the wire-format using @martinthomson's suggestion with separate port numbers

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
dtikhonov: I'd err on the side of making it as generic as possible: that is, separate ports.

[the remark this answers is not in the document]
DavidSchinazi: The latest version has separate ports, see the diff.

draft-ietf-quic-transport.md
ianswett: I realize, looking back on the issue, why Jana mentioned alternate.  There are two separate cases: Connection migration and Server's Preferred Address.

In one case, you want to make sure connection migration doesn't fail when moving from v6 to v4 only, but you don't want them to do anything after the handshake.  In the other case, you want them to migrate to a more stable IP after the handshake.

If the server omits the address family you connected on, but supplies a preferred address on the other address family, do you try to migrate?

I suspect we need to change the structure from what's in this PR, and if not, we need to nail down the critical use cases and make sure they all work.

ianswett: Re-reading this PR, I think its workable as is, but this might read better if this paragraph was merged into the next paragraph and the next paragraph was updated to something like: "one of the servers preferred addresses using..."

DavidSchinazi:
@ianswett I think migration and preferred address are separate parts of the protocol. The goal of this PR is to improve preferred address. It allows the client to pick the address family it transfers to right after the handshake. If the server only provides you with a preferred address family you don't have access to, you can simply ignore it - and that's not new from this PR. I've slightly tweaked the text to clarify how clients respond to preferred_address. Does this address your concerns?

draft-ietf-quic-transport.md
```diff
@@ -2116,9 +2116,13 @@ packets.
 A server conveys a preferred address by including the preferred_address
 transport parameter in the TLS handshake.
 
-Once the handshake is finished, the client SHOULD initiate path validation (see
-{{migrate-validate}}) of the server's preferred address using the connection ID
-provided in the preferred_address transport parameter.
+Servers MAY communicate a preferred address of each address family (IPv4 and
+IPv6) to allow clients to pick the one most suited to their network attachment.
```
martinthomson: After all our discussion, the question I have here is whether this preferred address might be used for a migating client.  Say I use v6 now, but then migrate to a network that doesn't have v6.  Is it acceptable for me to use the v6 preferred address during that migration?  The text here is silent on that point.

DavidSchinazi: @martinthomson I'm thinking we may want more deployment experience before we decide if we want to recommend or prohibit that.

martinthomson: I'm only asking if it is *allowed*.

DavidSchinazi: It is currently allowed - there is no text prohibiting it.

draft-ietf-quic-transport.md
janaiyengar: Yes, this ambiguity comes from the fact that this PR solves a problem that is slightly different than the one solved by the PreferredAddress TP.  I think you're all right that "preferred" is better than "alternate", but there's a gap.

I think this can be addressed by just a bit of text stating that it is up to the client to decide which among the preferred addresses it uses.  If the client has access to both address families, it's up to the client to decide whether it wants to migrate or not. The addresses are preferred by the server, but it's up to the client to decide if and when it moves to those.

draft-ietf-quic-transport.md
janaiyengar: We may want to explicitly state this. See my comment above.

DavidSchinazi: Good point, I've added some text to the _Interaction of Client Migration and Preferred Address_ section.

janaiyengar:
I'm good with this PR. @martinthomson -- can you see if your concern is addressed?

draft-ietf-quic-transport.md
```diff
@@ -2169,6 +2173,9 @@ receiving a probe packet from a different address.  Servers MUST NOT send more
 than a minimum congestion window's worth of non-probing packets to the new
 address before path validation is complete.
 
+When a client wishes to migrate to a new local address that is of a different
+address family from the currently used server address, the client can attempt
```
martinthomson: This can be simpler.  A client that migrates to a new address SHOULD use a preferred address from the same address family for the server. 

I wonder if we should forbid doing a new address lookup for the name or if we should stay silent on that point...

janaiyengar: +1 to the SHOULD.  

Let's not forbid lookups.  It's an interesting question whether other mechanisms of learning the server's addresses are useful or not; I think more deployment experience would be useful here.

DavidSchinazi: Agreed. I've incorporated @martinthomson's text. I agree with Jana that we shouldn't forbid lookups yet.

MikeBishop: I'm okay with the SHOULD.  The caveats on our network are:
- A lookup by address from a different network will lead you to a different machine.  Unless we then have it route back to the machine you were previously talking to, it will ignore / Reset you.  Of course, if it can't route, it shouldn't have the same Reset Token key, so you'll probably then drop the Reset....
- Using the SPA from a different network will lead you back to the same host, which might not be a good choice for you to keep using.  You'll likely get a prompt GOAWAY in HTTP/3.

RyanAtGoogle: On our networks as well, lookups are unlikely to route to the same machine. Given that there is no reason to expect a lookup to successfully route to the same machine (even on the SAME network), forbidding it seems like the right choice.

DavidSchinazi: I think the DNS lookup issue is orthogonal to this PR. @martinthomson can you file a new issue to discuss it there?

DavidSchinazi:
@martinthomson is there anything left for me to do so this can get merged in?

martinthomson:
You could wait until I had time to check it, that's all.

PR #2305, opened 2019-01-07
use the STREAM_STATE_ERROR for stream-related errors

marten-seemann:
So far, we didn't use the STREAM_STATE_ERROR at all


martinthomson:
WFM.

PR #2307, opened 2019-01-07
allow PING frames in Initial and Handshake packets

marten-seemann:
Furthermore, specify the error code to send when encountering invalid frames at the Handshake encryption level.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3801,7 +3801,7 @@ Handshake packet sent by a server contains a packet number of 0.
 The payload of this packet type contains CRYPTO frames and could contain
 PADDING, PING and ACK frames. Handshake packets MAY contain CONNECTION_CLOSE
 frames.  Endpoints MUST treat receipt of Handshake packets with other frames as
-a connection error.
+a connection error of type PROTOCOL_VIOLATION.
 
 Like Initial packets (see {{discard-initial}}), data in CRYPTO frames at the
 Handshake encryption level is discarded - and no longer retransmitted - when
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3700,7 +3700,7 @@ server may send multiple Initial packets.  The cryptographic key exchange could
 require multiple round trips or retransmissions of this data.
 
 The payload of an Initial packet includes a CRYPTO frame (or frames) containing
-a cryptographic handshake message, ACK frames, or both.  PADDING and
+a cryptographic handshake message, ACK frames, or both.  PADDING, PING and
```
ianswett: Not your text, but the first sentence implies an Initial packet has to contain either an ACK frame and/or a CRYPO frame, but I don't think that's true.  I'd suggest merging these sentences and say:
"The payload of an Initial packet includes a combination of one or more CRYPTO, ACK, PADDING, PING or CONNECTION_CLOSE frames."

And I'd update the paragraph below to use the same structure so they're parallel, because right now they're subtly different in not very interesting ways.

mikkelfj:
The less there is in the handshake the more robust the handshake is across QUIC versions, which may be beneficial since it is likely a separate code path from general packet handling.

martinthomson:
Is there an issue that this addresses?

marten-seemann:
I didn't open an issue, because I thought this was straightforward. In my implementation I use PINGs quite often for loss recovery purposes. Just sending a single PING frame is for example a valid way to generate a probe packet. There's no security implication of allowing PING frames.

martinthomson:
If we're going to resolve this, it needs an issue.  Because we're discussing issues there.  This one seems contentious.

PR #2313, opened 2019-01-07
Proposal to make Version Negotiation more like Retry to punt VN to QUICv2

DavidSchinazi:
We've been talking about Version Negotiation (VN) for a while, as the current design isn't perfect yet. In particular, it is vulnerable to subtle deadlocks and downgrades and requires a full RTT.

I have a proposal for the WG that would effectively punt this issue, allowing us to focus on other QUICv1 deliverables until we actually need VN and have deployment experience.

I'm working with the following assumptions:
- when we'll have clients and servers supporting both QUICv1 and QUICv2, we will not be willing to pay a full RTT to negotiate which version to use
- when we standardize QUICv1, we will not know what QUICv2 will look like, so cannot design compatible VN in v1 and be sure it will work with v2
- compatible VN can rely on transport parameters, so it will be possible to standardize it later as an extension to QUICv1

The proposal is to make VN packets a fatal error that effectively kills the first connection, and have them transmit a new connection ID to allow clients to create a new connection that will hit the same server (or not, based on server preference). In terms of wire format, this makes VN more like Retry by making the server send a new connection ID to the client.

(Note that this was mostly inspired by EKR's proposals.)


draft-ietf-quic-transport.md
```diff
@@ -3604,23 +3503,55 @@ The layout of a Version Negotiation packet is:
...
+A Version Negotiation packet contains the following fields:
 
-The server MUST include the value from the Source Connection ID field of the
-packet it receives in the Destination Connection ID field.  The value for Source
-Connection ID MUST be copied from the Destination Connection ID of the received
-packet, which is initially randomly selected by a client.  Echoing both
-connection IDs gives clients some assurance that the server received the packet
-and that the Version Negotiation packet was not generated by an off-path
-attacker.
+Reserved Bits (R):
```
ianswett: There is a lot of added text here.  Is that because the long header packets are currently not correctly documented?

DavidSchinazi: The added text here is to add new fields to VN packets. This PR adds ODCIL and Original Destination Connection ID to allow the server to send a new source connection ID (similar to how Retry packets work today). The PR also adds a Number of Supported Versions field to allow future extensibility.

MikeBishop:
My issue with this general approach is that VN is currently part of the invariants.  (Or otherwise framed, VN is the only packet defined by QUIC version zero.)  If we want to support versions which aren't mutually intelligible, it basically has to be version-independent.  If you're only writing in the Transport document, the PR probably doesn't have the necessary contents yet.

Further, I think we as a WG need to return to our basic principles in VN and agree on those first -- we have too many designs being tossed around fueled by different assumptions.

DavidSchinazi:
I've now reverted the part of this proposal that changed the invariants. It now simply removes VN from v1.

draft-ietf-quic-transport.md
```diff
@@ -1143,40 +1125,30 @@ expectation that it will eventually receive an Initial packet.
 
 ## Handling Version Negotiation Packets {#handle-vn}
 
-When the client receives a Version Negotiation packet, it first checks that the
-Destination and Source Connection ID fields match the Source and Destination
-Connection ID fields in a packet that the client sent.  If this check fails, the
-packet MUST be discarded.
+How a client reacts to receiving a Version Negotiation packet is left as future
+work defined by future versions of QUIC.  Future versions of QUIC that define
```
martinthomson: This isn't right.  This version needs to define what to do.

I recommend having clients abandon a connection attempt in the absence of better advice.

draft-ietf-quic-transport.md
```diff
@@ -1143,40 +1125,30 @@ expectation that it will eventually receive an Initial packet.
...
+\[\[RFC editor: please remove this section before publication.]]
 
-A client MUST NOT change the version it uses unless it is in response to a
-Version Negotiation packet from the server.  Once a client receives a packet
-from the server which is not a Version Negotiation packet, it MUST discard other
-Version Negotiation packets on the same connection.  Similarly, a client MUST
-ignore a Version Negotiation packet if it has already received and acted on a
-Version Negotiation packet.
+When a draft implementation receives a Version Negotiation packet, it MAY use
+it to attempt a new connection with one of the supported versions.
```
```suggestion
it to attempt a new connection with one of the versions listed in the packet.
```

draft-ietf-quic-transport.md
```diff
@@ -4010,14 +3903,6 @@ language from Section 3 of {{!TLS13=RFC8446}}.
    } TransportParameter;
 
    struct {
-      select (Handshake.msg_type) {
-         case client_hello:
-            QuicVersion initial_version;
```
martinthomson: You missed the definition of QuicVersion at the top of this block.

draft-ietf-quic-transport.md
```diff
@@ -4010,14 +3903,6 @@ language from Section 3 of {{!TLS13=RFC8446}}.
...
-
-         case encrypted_extensions:
-            QuicVersion negotiated_version;
-            QuicVersion supported_versions<4..2^8-4>;
-      };
       TransportParameter parameters<0..2^16-1>;
    } TransportParameters;
```
martinthomson: This can now be a straight typedef: `TransportParameter TransportParameters<0..2^16-1>;`

DavidSchinazi:
Added a commit which incorporate's @martinthomson and @ianswett's feedback

draft-ietf-quic-transport.md
```diff
@@ -1149,41 +1131,36 @@ expectation that it will eventually receive an Initial packet.
...
 
-When the client receives a Version Negotiation packet, it first checks that the
-Destination and Source Connection ID fields match the Source and Destination
-Connection ID fields in a packet that the client sent.  If this check fails, the
-packet MUST be discarded.
+When a client receives a Version Negotiation packet, it MUST abandon the
+current connection attempt.  Version Negotiation packets are designed to allow
+future versions of QUIC to negotiate the version in use between endpoints.
+It is therefore expected for future versions of QUIC to change how
+implementations of this version of QUIC react to Version Negotiation packets.
```
martinthomson: I don't think that this can be true.  The point here is that implementations of future versions of QUIC might change how they react to Version Negotiation packets when attempting connections with this version of QUIC.

draft-ietf-quic-transport.md
```diff
@@ -5406,6 +5280,17 @@ correct instance, it is better to send a stateless reset than wait for
...
 be influenced by an attacker.
 
+## Version Downgrade {#version-downgrade}
+
+This document defines QUIC Version Negotiation packets {{version-negotiation}},
+which can be used to negotiate the QUIC version used between two endpoints.
+However, this document does not specify how this negotiation will be performed
+between this version and subsequent future versions.  In particular, Version
+Negotiation packets do not contain any mechanism to prevent version downgrade
+attacks.  Future version of QUIC that wish to use Version Negotiation packets
```
```suggestion
attacks.  Future versions of QUIC that use Version Negotiation packets
```

draft-ietf-quic-transport.md
```diff
@@ -5406,6 +5280,17 @@ correct instance, it is better to send a stateless reset than wait for
...
+to negotiate the use of that version MUST define a mechanism that is robust
```
```suggestion
MUST define a mechanism that is robust
```

DavidSchinazi:
Applied latest comments / changes from @martinthomson

PR #2320, opened 2019-01-09
update spin-exp short header to sync with #2203

britram:
now that the first octet PR landed, spin-exp needs to be updated to reflect the change. this does that.

### DIFF
--- a/draft-ietf-quic-spin-exp.md
+++ b/draft-ietf-quic-spin-exp.md
@@ -130,31 +130,25 @@ version negotiation and connection establishment are completed.
 
 ## Proposed Short Header Format Including Spin Bit {#header}
 
-As of the current editor's version of {{QUIC-TRANSPORT}}, this proposal
-specifies using the sixth most significant bit (0x04) of the first byte in
-the short header for the spin bit.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-spin-exp.md
```diff
@@ -130,31 +130,25 @@ version negotiation and connection establishment are completed.
...
+byte in the short header for the spin bit (0x20, labeled S in
+{{fig-short-header}}). The Spin bit is set 0 or 1 depending on the stored spin
+value that is updated on packet reception as explained in {{spinbit}}.
 
+~~~
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+
-|0|K|1|1|0|S|R R|
+|0|1|S|R|R|K|P P|
```
martinthomson: Note that #2237 changes this to:

```suggestion
|0|1|S|R|K K|P P|
```

But what you have is correct for the moment.  #2237 is still under discussion.

britram: yep, watching that discussion. :)

PR #2324, opened 2019-01-09
Transport owns details of effecting keep-alive

### LINKED ISSUE #2092: More generic keep-alive text
In #2068, the direction of the discussion is that PING frames are an application-layer concept that won't necessarily be directly under the application-layer's control.  The HTTP draft currently says:
> HTTP clients are expected to use QUIC PING frames to keep connections open while there are responses outstanding for requests or server pushes. If the client is not expecting a response from the server, allowing an idle connection to time out is preferred over expending effort maintaining a connection that might not be needed. A gateway MAY use PING to maintain connections in anticipation of need rather than incur the latency cost of connection establishment to servers. Servers SHOULD NOT use PING frames to keep a connection open.

This should be reworded in terms of requesting the transport stack to keep the connection from idling out, rather than directly discussing PING frames.

MikeBishop:
Fixes #2092.


draft-ietf-quic-http.md
```diff
@@ -1202,13 +1202,14 @@ will need to open a new connection for new requests if the existing connection
...
-HTTP clients are expected to use QUIC PING frames to keep connections open while
-there are responses outstanding for requests or server pushes. If the client is
-not expecting a response from the server, allowing an idle connection to time
-out is preferred over expending effort maintaining a connection that might not
-be needed.  A gateway MAY use PING to maintain connections in anticipation of
-need rather than incur the latency cost of connection establishment to servers.
-Servers SHOULD NOT use PING frames to keep a connection open.
+HTTP clients are expected to request that the transport keep connections open
+while there are responses outstanding for requests or server pushes.  This might
+involve QUIC PING frames being generated by the transport implementation. If the
```
martinthomson: Do you want to cite Section 19.2 of transport here?  That might be better than this "This might involve" sentence, which strikes me as very passive.

PR #2325, opened 2019-01-09
Introduce HTTP_REQUEST_REJECTED

### LINKED ISSUE #2106: Echoing HTTP_REQUEST_CANCELLED from STOP_SENDING has unintended semantics
The current transport draft says:

>   A STOP_SENDING frame requests that the receiving endpoint send a RESET_STREAM frame. An  endpoint that receives a STOP_SENDING frame MUST send a RESET_STREAM frame for that stream. An endpoint SHOULD copy the error code from the STOP_SENDING frame, but MAY use 
any other application error code. The endpoint that sends a STOP_SENDING frame can ignore the 
error code carried in any RESET_STREAM frame it receives.

For HTTP, the HTTP_REQUEST_CANCELLED application error code has different semantics in each direction.  From C -> S, it means "I don't want this".  From S -> C, it means "I didn't process this".  I don't think the client can't distinguish whether the RESET_STREAM was sent before or after the STOP_SENDING was received.

Maybe it's ok to do nothing here, since the client isn't likely to attempt to retry the request it just abandoned, and the text says the sender can ignore the RESET_STREAM error code.  Or maybe we should clarify it in some way.

MikeBishop:
Fixes #2106.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1217,14 +1217,15 @@ processed.  Servers SHOULD NOT increase the QUIC MAX_STREAM_ID limit after
 sending a GOAWAY frame.
 
 Once GOAWAY is sent, the server MUST reject requests sent on streams with an
-identifier higher than the indicated last Stream ID.  Clients MUST NOT send new
-requests on the connection after receiving GOAWAY, although requests might
-already be in transit. A new connection can be established for new requests.
+identifier greater than or equal to the indicated last Stream ID.  Clients MUST
+NOT send new requests on the connection after receiving GOAWAY, although
+requests might already be in transit. A new connection can be established for
+new requests.
 
-If the client has sent requests on streams with a higher Stream ID than
-indicated in the GOAWAY frame, those requests are considered rejected
-({{request-cancellation}}).  Clients SHOULD reset any streams above this ID with
-the error code HTTP_REQUEST_REJECTED.  Servers MAY also reject requests on
+If the client has sent requests on streams with a Stream ID greater than or
+equal to that indicated in the GOAWAY frame, those requests are considered
+rejected ({{request-cancellation}}).  Clients SHOULD reset any rejected streams
+with the error code HTTP_REQUEST_REJECTED.  Servers MAY also reject requests on
 streams below the indicated ID if these requests were not processed.
 
 Requests on Stream IDs less than the Stream ID in the GOAWAY frame might have
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -963,20 +963,20 @@ for each header field.
 
 ### Request Cancellation
 
-Either client or server can cancel requests by aborting the stream (QUIC
-RESET_STREAM and/or STOP_SENDING frames, as appropriate) with an error code of
+Clients can cancel requests by aborting the stream (QUIC RESET_STREAM and/or
+STOP_SENDING frames, as appropriate) with an error code of
 HTTP_REQUEST_CANCELLED ({{http-error-codes}}).  When the client cancels a
 response, it indicates that this response is no longer of interest.
 Implementations SHOULD cancel requests by aborting both directions of a stream.
 
-When the server aborts its response stream using HTTP_REQUEST_CANCELLED, it
-indicates that no application processing was performed.  In this context,
-"processed" means that some data from the stream was passed to some higher layer
-of software that might have taken some action as a result.  The client can treat
-requests cancelled by the server as though they had never been sent at all,
-thereby allowing them to be retried later on a new connection.  Servers MUST NOT
-use the HTTP_REQUEST_CANCELLED status for requests which were partially or fully
-processed.
+When the server rejects a request without performing any application processing,
+it SHOULD abort its response stream with the error code HTTP_REQUEST_REJECTED.
+In this context, "processed" means that some data from the stream was passed to
+some higher layer of software that might have taken some action as a result.
+The client can treat requests rejected by the server as though they had never
+been sent at all, thereby allowing them to be retried later on a new connection.
+Servers MUST NOT use the HTTP_REQUEST_REJECTED error code for requests which
+were partially or fully processed.
 
 If a stream is cancelled after receiving a complete response, the client MAY
 ignore the cancellation and use the response.  However, if a stream is cancelled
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -1206,22 +1206,29 @@ function by communicating with clients.
...
 on lower stream IDs were or might be processed in this connection, while
-requests on the indicated stream ID and greater were not accepted. This enables
+requests on the indicated stream ID and greater were rejected. This enables
 client and server to agree on which requests were accepted prior to the
 connection shutdown.  This identifier MAY be lower than the stream limit
 identified by a QUIC MAX_STREAM_ID frame, and MAY be zero if no requests were
 processed.  Servers SHOULD NOT increase the QUIC MAX_STREAM_ID limit after
 sending a GOAWAY frame.
 
+<<<<<<< 64585a30a8ddc7a766080499cd44cd5d4206fe1d
```
dtikhonov: Conflict

DavidSchinazi:
This change sounds good to me, but to make sure I understand: is the only purpose of this change to split HTTP_REQUEST_CANCELLED in two; client to server stays HTTP_REQUEST_CANCELLED and server to client now becomes HTTP_REQUEST_REJECTED? And is the goal to add clarity when parsing connection logs later?

draft-ietf-quic-http.md
```diff
@@ -963,20 +963,20 @@ for each header field.
 
 ### Request Cancellation
 
-Either client or server can cancel requests by aborting the stream (QUIC
-RESET_STREAM and/or STOP_SENDING frames, as appropriate) with an error code of
+Clients can cancel requests by aborting the stream (QUIC RESET_STREAM and/or
+STOP_SENDING frames, as appropriate) with an error code of
 HTTP_REQUEST_CANCELLED ({{http-error-codes}}).  When the client cancels a
 response, it indicates that this response is no longer of interest.
 Implementations SHOULD cancel requests by aborting both directions of a stream.
```
martinthomson: Missing from either this paragraph or the next:  A server that receives a STOP_SENDING with the HTTP_REQUEST_CANCELLED error code MAY use the HTTP_REQUEST_REJECTED error code in response to indicate that no processing occurred.

draft-ietf-quic-http.md
```diff
@@ -1867,8 +1871,11 @@ FRAME_SIZE_ERROR (0x6):
 : HTTP_MALFORMED_FRAME error codes defined in {{http-error-codes}}.
 
 REFUSED_STREAM (0x7):
-: Not applicable, since QUIC handles stream management.  Would provoke a
-  STREAM_ID_ERROR from the QUIC layer.
+: When used to indicate that a request was not processed, HTTP_REQUEST_REJECTED
+  in {{http-error-codes}}.
+: When used to indicate that an active stream limit was exceeded, not applicable
```
martinthomson: I had a little trouble parsing this bit.  Maybe reorder the sentences:

> HTTP_REQUEST_REJECTED (in {{http-error-codes}}) is used to indicate when requests were not processed.  Otherwise, not applicable because QUIC handles stream management.  A STREAM_ID_ERROR at the QUIC layer is used for streams that are improperly opened.

draft-ietf-quic-http.md
```diff
@@ -1206,22 +1209,22 @@ function by communicating with clients.
...
 identifier higher than the indicated last Stream ID.  Clients MUST NOT send new
 requests on the connection after receiving GOAWAY, although requests might
 already be in transit. A new connection can be established for new requests.
 
 If the client has sent requests on streams with a higher Stream ID than
-indicated in the GOAWAY frame, those requests are considered cancelled
+indicated in the GOAWAY frame, those requests are considered rejected
 ({{request-cancellation}}).  Clients SHOULD reset any streams above this ID with
-the error code HTTP_REQUEST_CANCELLED.  Servers MAY also cancel requests on
+the error code HTTP_REQUEST_REJECTED.  Servers MAY also reject requests on
```
afrind: Why have the client send REJECTED in this case?

draft-ietf-quic-http.md
DavidSchinazi: Wouldn't that defeat the purpose of this PR? If a server receives STOP_SENDING with HTTP_REQUEST_CANCELLED, it doesn't reject the request, it acts on the client's cancellation. In that case I think it would make more sense for the server to use HTTP_REQUEST_CANCELLED.

draft-ietf-quic-http.md
MikeBishop: Because GOAWAY already tells the client that the server didn't act on these higher streams; this is effectively the client's agreement that these streams were rejected by GOAWAY.  (There's probably also a STOP_SENDING en route with the same error code.)

draft-ietf-quic-http.md
MikeBishop: From `-transport`:
> An endpoint SHOULD copy the error code from the STOP_SENDING frame, but MAY use any application error code.

_CANCELLED would be the default behavior; this text merely says that _REJECTED is also acceptable.  (And the transport draft says that the server MAY use any error code, so this is really just a restatement.)

The subtle difference would be an indication of whether the cancelled request had begun to process (and therefore might have had side-effects) before the cancellation took place.

martinthomson: Right. The trick here is that a cancelled request can't be automatically retried.  A rejected one can be.  So using _REJECTED is preferable.

DavidSchinazi: Ah, I see. I agree. Hmmm not sure how to clarify the text so others aren't confused the same way I was though.

draft-ietf-quic-http.md
afrind: This seems weird.  I think I prefer the client sending CANCELLED.

martinthomson: Oh, I missed that.  Yes.  The client CAN'T send _REJECTED.  Because if the server echoes it (as they are allowed to), then the client might conclude that nothing was processed.  We need a prohibition against clients using _REJECTED.

> Clients MUST NOT use HTTP_REQUEST_REJECTED with STOP_SENDING.

MikeBishop: It's not that the client can't, because those streams *haven't* been processed, per GOAWAY.  But yes, we should add a prohibition on its use in other cases at least.

PR #2326, opened 2019-01-09
Client can't prioritize streams it can't open

### LINKED ISSUE #2144: HTTP: Limit priority state before stream is opened
RFC 7540 [discusses priority state management](https://tools.ietf.org/html/rfc7540#section-5.3.4) at some length, especially what to do with priority information after a stream is closed.

In addition to this, in HTTP/3, we have to deal with priority information before a stream is opened. That is: just how large is the tree allowed to grow before the referenced streams' frames begin to come in?

The draft should provide some recommendations.

I can think of two ways to do limit this "front-loaded" state:
1. Open the stream when a _PRIORITY_ frame referencing it arrives on the control stream. This will invoke the stream limit checks.
2. Limit number of priority entries to number -- perhaps a function of the total number of maximum unidirectional and bidirectional streams.

MikeBishop:
Generalizes HTTP_PUSH_LIMIT_EXCEEDED to HTTP_LIMIT_EXCEEDED and mandates use of that error code where a PRIORITY frame attempts to prioritize a stream the client isn't permitted to open yet.  This has the effect of bounding priority tree state to requests the client *could* initiate (though might not have yet).

Fixes #2144.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -561,8 +561,8 @@ of a PRIORITY frame identifying a stream of any other type as a connection error
 of type HTTP_MALFORMED_FRAME.
 
 A PRIORITY frame that references a non-existent Push ID, a Placeholder ID
-greater than the server's limit, or a Stream ID the client is not permitted to
-open MUST be treated as an HTTP_LIMIT_EXCEEDED error.
+greater than the server's limit, or a Stream ID the client is not yet permitted
+to open MUST be treated as an HTTP_LIMIT_EXCEEDED error.
 
 A PRIORITY frame received on any stream other than a request or control stream
 MUST be treated as a connection error of type HTTP_WRONG_STREAM.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -1385,8 +1385,9 @@ HTTP_VERSION_FALLBACK (0x09):
 HTTP_WRONG_STREAM (0x0A):
 : A frame was received on a stream where it is not permitted.
 
-HTTP_PUSH_LIMIT_EXCEEDED (0x0B):
-: A Push ID greater than the current maximum Push ID was referenced.
+HTTP_LIMIT_EXCEEDED (0x0B):
+: An identifier greater than the current maximum for that identifier was
+  referenced.
```
martinthomson: I think that you can (and should) list the types of identifiers this applies to.  I assume that this is limited to push IDs and priority placeholders.

draft-ietf-quic-http.md
```diff
@@ -557,12 +557,12 @@ MUST be treated as a connection error of type HTTP_MALFORMED_FRAME.
...
-of PRIORITY frame with a Stream ID of any other type as a connection error of
-type HTTP_MALFORMED_FRAME.
+of a PRIORITY frame identifying a stream of any other type as a connection error
+of type HTTP_MALFORMED_FRAME.
 
-A PRIORITY frame that references a non-existent Push ID or a Placeholder ID
-greater than the server's limit MUST be treated as an HTTP_MALFORMED_FRAME
-error.
+A PRIORITY frame that references a non-existent Push ID, a Placeholder ID
+greater than the server's limit, or a Stream ID the client is not permitted to
```
martinthomson: For stream IDs, is this "not **yet** permitted"?  The previous paragraph establishes the use of HTTP_MALFORMED_FRAME for the wrong type of stream.

draft-ietf-quic-http.md
MikeBishop: And stream IDs.

PR #2346, opened 2019-01-18
Rework idle timeout definition

### LINKED ISSUE #2149: What resets idle timeout?
```
Each endpoint advertises its own idle timeout to its peer. The idle timeout
starts from the last packet received.  In order to ensure that initiating new
activity postpones an idle timeout, an endpoint restarts this timer when sending
a packet.  An endpoint does not postpone the idle timeout if another packet has
been sent containing frames other than ACK or PADDING, and that other packet has
not been acknowledged or declared lost.  Packets that contain only ACK or
PADDING frames are not acknowledged until an endpoint has other frames to send,
so they could prevent the timeout from being refreshed.
```

I find this pretty hard to read, but assuming I *am* reading it corectly,
it basically says that if you read any packet or write any packet that
has a frame other than ACK or padding, then you reset the timer. The text
about "other packet has not been acknowledged or lost" doesn't make sense
unless you have some very complicated timer scheme, because the packet
won't be acknowledge or lost until after it is sent, at which point I
have already reset the timer.

martinthomson:
Closes #2149.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2243,7 +2243,7 @@ the advertised idle timeout (see {{transport-parameter-definitions}}) is closed.
 A connection enters the draining state when the idle timeout expires.
 
 Each endpoint advertises its own idle timeout to its peer.  An enpdpoint
-restarts any timer it maintains when a packet from their peer is received and
+restarts any timer it maintains when a packet from its peer is received and
 processed successfully.  The timer is also restarted when sending a packet
 containing frames other than ACK or PADDING (an ACK-eliciting packet, see
 {{QUIC-RECOVERY}}), but only if no other ACK-eliciting packets have been sent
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2242,14 +2242,13 @@ If the idle timeout is enabled, a connection that remains idle for longer than
...
 
-Each endpoint advertises its own idle timeout to its peer. The idle timeout
-starts from the last packet received.  In order to ensure that initiating new
-activity postpones an idle timeout, an endpoint restarts this timer when sending
-a packet.  An endpoint does not postpone the idle timeout if another packet has
-been sent containing frames other than ACK or PADDING, and that other packet has
-not been acknowledged or declared lost.  Packets that contain only ACK or
-PADDING frames are not acknowledged until an endpoint has other frames to send,
-so they could prevent the timeout from being refreshed.
+Each endpoint advertises its own idle timeout to its peer.  Endpoints restart
```
dtikhonov: I'd switch to the singular number:

> _An enpdpoint restarts any timer it maintains when a packet from its peer_....

As it stands, it sounds as if several endpoints had a single peer.

PR #2349, opened 2019-01-18
Correcting polarity of SPA language

### LINKED ISSUE #2146: Preferred address text is confusing
The following text is kind of hard to track:

```A server might receive a packet addressed to its preferred IP address at any
time after it accepts a connection.  If this packet contains a PATH_CHALLENGE
frame, the server sends a PATH_RESPONSE frame as per {{migrate-validate}}.  The
server MAY send other non-probing frames from its preferred address, but MUST
continue sending all probing packets from its original IP address.

The server SHOULD also initiate path validation of the client using its
preferred address and the address from which it received the client probe.  This
helps to guard against spurious migration initiated by an attacker.
```

I'm not really following what's going on here. The server knows that it has two addresses and it wants to use its preferred address. Why does it have to probe from its preferred address. Also, I don't understand how these two paragraphs work together. The first one says you have to send probing packets from your original address but the second paragraph tells you to probe from the preferrd address.

MikeBishop:
Editorial because it now says what we originally thought it said, but design because it is changing normative language.  Fixes #2146. Does not interact with #2296.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2150,8 +2150,8 @@ A server might receive a packet addressed to its preferred IP address at any
 time after it accepts a connection.  If this packet contains a PATH_CHALLENGE
 frame, the server sends a PATH_RESPONSE frame as per {{migrate-validate}}.  The
 server MUST send other non-probing frames from its original address until it
-receives a non-probing packet from the client at its preferred address and the
-server has validated the new path.
+receives a non-probing packet from the client at its preferred address and until
+the server has validated the new path.
 
 The server MUST probe on the path toward the client from its preferred address.
 This helps to guard against spurious migration initiated by an attacker.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2149,12 +2149,11 @@ MUST continue sending all future packets to the server's original IP address.
 A server might receive a packet addressed to its preferred IP address at any
 time after it accepts a connection.  If this packet contains a PATH_CHALLENGE
 frame, the server sends a PATH_RESPONSE frame as per {{migrate-validate}}.  The
-server MAY send other non-probing frames from its preferred address, but MUST
-continue sending all probing packets from its original IP address.
+server SHOULD send other non-probing frames from its original address until it
```
```suggestion
server MUST send other non-probing frames from its original address until it
```

draft-ietf-quic-transport.md
```diff
@@ -2149,12 +2149,11 @@ MUST continue sending all future packets to the server's original IP address.
...
+receives a non-probing packet from the client at its preferred address.
```
```suggestion
receives a non-probing packet from the client at its preferred address and the server
has validated the new path.
```

PR #2350, opened 2019-01-18
No common versions means fail

### LINKED ISSUE #1917: Complete version negotiation failure
It looks like this:  Client supports A and B only.  Server supports C and D only.  Client sends initial in A (or B), server sends Version Negotiation in response.  Client does what exactly?  Might as well say.

We could get into the client maybe ignoring that because it's an attack, but we decided not to worry too much about that, so probably not worth saying.

MikeBishop:
Closes #1917.

Cherry-picking a fix to this issue that got lost in one of the defunct VN-rework PRs.


mikkelfj:
I suggest adding

"A client MAY choose to wait a while to see if a supported Version Negotiation packet arrives before failing the connection. This protects against injection attacks."

PR #2354, opened 2019-01-21
STOP_SENDING is permitted for Ready streams

### LINKED ISSUE #2345: STOP_SENDING on implicitly opened streams is error-prone

martinthomson:
...in the case where those streams are implicitly created.  There's
probably another state here if we were being entirely honest with
ourselves to distinguish between a stream that has been created and one
that has had frames sent for it, but this is already more complex than
it needs to be.

Closes #2345.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -4479,16 +4479,15 @@ Final Size:
 ## STOP_SENDING Frame {#frame-stop-sending}
 
 An endpoint uses a STOP_SENDING frame (type=0x05) to communicate that incoming
-data is being discarded on receipt at application request.  This signals a peer
-to abruptly terminate transmission on a stream.
+data is being discarded on receipt at application request.  STOP_SENDING
+requests that a peer to cease transmission on a stream.
 
-Receipt of a STOP_SENDING frame is invalid for a locally-initiated stream that
-has not yet been created or is in the "Ready" state (see
-{{stream-send-states}}). Receiving a STOP_SENDING frame for a locally-initiated
-stream that is "Ready" or not yet created MUST be treated as a connection error
-of type STREAM_STATE_ERROR.  An endpoint that receives a STOP_SENDING frame for
-a receive-only stream MUST terminate the connection with error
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -4479,16 +4479,15 @@ Final Size:
 ## STOP_SENDING Frame {#frame-stop-sending}
 
 An endpoint uses a STOP_SENDING frame (type=0x05) to communicate that incoming
-data is being discarded on receipt at application request.  This signals a peer
-to abruptly terminate transmission on a stream.
+data is being discarded on receipt at application request.  STOP_SENDING
+requests that a peer to cease transmission on a stream.
```
dtikhonov: "requests that" usually calls for the subjunctive mood:

> ...requests that a peer cease transmission...

dtikhonov:
@ianswett, does this sound normal to you?

> STOP_SENDING requests that a peer to cease transmission on a stream.

ianswett:
To clarify, I +1'd your suggestion.

More clearly: "Looks good, but I would like you to accept Dimitri's suggestion before submitting this PR."

dtikhonov:
Ah, alright.  I thought it was the other way around.

martinduke:
This LGTM. I was just about to file a similar issue when I saw this.

PR #2362, opened 2019-01-23
Define RTT

ianswett:
Explicitly state that round trip time is abbreviated as RTT


draft-ietf-quic-recovery.md
```diff
@@ -293,10 +293,10 @@ continue making forward progress.
 
 # Computing the RTT estimate
 
-RTT is calculated when an ACK frame arrives by computing the difference between
-the current time and the time the largest acked packet was sent.  An RTT sample
-MUST NOT be taken for a packet that is not newly acknowledged or not
-ack-eliciting.
+Round trip time, abbreviated RTT, is calculated when an ACK frame arrives by
```
```suggestion
Round trip time (RTT) is calculated when an ACK frame arrives by
```

Just make sure that you add the parenthetical the first time you see the acronym.  If that is prior to this section in the document, then you probably need to expand the first usage as well.

ianswett: Done and changed to round-trip time, which matches TCP RFCs

PR #2364, opened 2019-01-23
Moving text on spinbit into transport

ihlar:
Text on spinbit that is necessary for implementors moved/added to transport draft.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1760,7 +1760,9 @@ An endpoint also MUST NOT initiate connection migration if the peer sent the
 `disable_migration` transport parameter during the handshake.  An endpoint which
 has sent this transport parameter, but detects that a peer has nonetheless
 migrated to a different network MAY treat this as a connection error of type
-INVALID_MIGRATION.
+INVALID_MIGRATION.  Similarly, an endpoint MUST NOT initiate migration if its
+peer supplies a zero-length connection ID as packets without a Destination
+Connection ID cannot be attributed to a connection based on address tuple.
 
 Not all changes of peer address are intentional migrations. The peer could
 experience NAT rebinding: a change of address due to a middlebox, usually a NAT,
@@ -2000,15 +2002,19 @@ different local addresses, as discussed in {{connection-id}}.  For this to be
 effective endpoints need to ensure that connections IDs they provide cannot be
 linked by any other entity.
 
-This eliminates the use of the connection ID for linking activity from
-the same connection on different networks.  Header protection ensures
-that packet numbers cannot be used to correlate activity.  This does not prevent
-other properties of packets, such as timing and size, from being used to
+At any time, endpoints MAY change the Destination Connection ID they send to a
+value that has not been used on another path.
+
+An endpoint MUST use a new connection ID if it initiates connection migration.
+Using a new connection ID eliminates the use of the connection ID for linking
+activity from the same connection on different networks.  Header protection
+ensures that packet numbers cannot be used to correlate activity.  This does not
+prevent other properties of packets, such as timing and size, from being used to
 correlate activity.
 
-Clients MAY move to a new connection ID at any time based on
-implementation-specific concerns.  For example, after a period of network
-inactivity NAT rebinding might occur when the client begins sending data again.
+Unintentional changes in path without a change in connection ID are possible.
+For example, after a period of network inactivity, NAT rebinding might cause
+packets to be sent on a new path when the client resumes sending.
 
 A client might wish to reduce linkability by employing a new connection ID and
 source UDP port when sending traffic after a period of inactivity.  Changing the
@@ -2019,21 +2025,9 @@ genuine migrations.  Changing port number can cause a peer to reset its
 congestion state (see {{migration-cc}}), so the port SHOULD only be changed
 infrequently.
 
-Endpoints that use connection IDs with length greater than zero could have their
-activity correlated if their peers keep using the same destination connection ID
-after migration. Endpoints that receive packets with a previously unused
-Destination Connection ID SHOULD change to sending packets with a connection ID
-that has not been used on any other network path.  The goal here is to ensure
-that packets sent on different paths cannot be correlated. To fulfill this
-privacy requirement, endpoints that initiate migration and use connection IDs
-with length greater than zero SHOULD provide their peers with new connection IDs
-before migration.
-
-Caution:
-
-: If both endpoints change connection ID in response to seeing a change in
-  connection ID from their peer, then this can trigger an infinite sequence of
-  changes.
+An endpoint that exhausts available connection IDs cannot migrate.  To ensure
+that migration is possible and packets sent on different paths cannot be
+correlated, endpoints SHOULD provide new connection IDs before peers migrate.
 
 
 ## Server's Preferred Address {#preferred-address}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2765,7 +2765,8 @@ on the network path throughout the duration of a connection. The spin bit is
 only present in the short packet header, since it is possible to measure the
 initial RTT of a connection by observing the handshake. Therefore, the spin bit
 will appear after version negotiation and connection establishment are
-completed.
+completed. On-path measurement and use of the Latency Spin Bit is further 
+discussed in {{QUIC-MANAGEABILITY}}.
 
 The spin bit utilizes a single bit in the first byte of the short header. The  
 location of the bit and procedures for how to set it by clients and servers are
@@ -2779,13 +2780,13 @@ Each endpoint unilaterally decides if the spin bit is enabled or disabled for a
 connection. Implementations SHOULD allow administrators of clients and servers
 to disable the spin bit either globally or on a per-connection basis. Even when
 the spin bit is not disabled by the administrator implementations SHOULD disable
-the spin bit on a randomly chosen fraction of connections, except for
-connections that are explicitly configured to have the spin bit enabled.  
-
-The selection process SHOULD be designed such that on average the spin bit is
-disabled for at least one eighth of network paths.  The selection process should
-be externally unpredictable but consistent for any given combination of source
-and destination address/port. For instance, the implementation might have a
+the spin bit on a randomly chosen fraction of connections. However, connections 
+may be configured to explicitly enable spinning, for example in the case of
+explicit customer support and debugging.
+The random selection process SHOULD be designed such that on average the spin bit 
+is disabled for at least one eighth of network paths. The selection process 
+should be externally unpredictable but consistent for any given combination of source
+and destination address and port. For instance, the implementation might have a
 static key which it uses to key a pseudorandom function over these values and
 use the output to determine whether to send the spin bit. The selection process
 performed at the beginning of the connection SHOULD be applied for all paths
@@ -2794,11 +2795,7 @@ used by the connection.
 In case multiple connections share the same five-tuple, i.e. same source and
 destination IP address and UDP port the setting of the spin bit needs to be
 coordinated across all connections to ensure a clear signal to any on path
-measurement point, however that might not be possible.
-
-On-path measurement and use of the Latency Spin Bit is further discussed in
-{{QUIC-MANAGEABILITY}}.
-
+measurement point, however that might not be feasible.
 
 # Packetization and Reliability {#packetization}
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -851,6 +851,12 @@ limits are set in the transport parameters (see
 using MAX_STREAMS frames ({{frame-max-streams}}). Separate limits apply to
 unidirectional and bidirectional streams.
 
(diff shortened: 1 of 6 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3948,8 +3993,33 @@ Fixed Bit:
 
 Spin Bit (S):
 
-: The sixth bit (0x20) of byte 0 is the Latency Spin Bit, set as described in
-  {{!SPIN=I-D.ietf-quic-spin-exp}}.
+: The third most significant bit (0x20) of byte 0 is the Latency Spin Bit.
```
ianswett: I think this would be easier to read if either all the text was here or if this referenced the section above.  Given it's a lot of text, I tend to think the latter is better.

martinthomson:
I think that this is one editorial discussion where I would like to see some discussion.

For me, the question of document organization is secondary to a question of status.  I think that we should target the spin bit document at experimental status, not proposed standard, which would make keeping the split between documents much more appropriate.  I'm sure that others will disagree with this position, so I think that discussion is very much warranted.

Incidentally, the change from exp to std in the document is in a commit entitled "typofix", done late last year by @britram.  I was - until just now when I checked - under the impression that the goal was an experimental RFC.  I suspect that we had a discussion on this point that settled on standards track, but I'm just remembering badly.

After that, then I think we want to talk about where this slots into the current document.  I'm not sure that the place that is selected here is right, but I want to get the big ticket questions out of the way first.

britram:
haven’t had a chance to review the PR in detail yet, should do so today.

On document status, my understanding is:

- in London, we decided to adopt the spin bit as an experimental feature within the working group to generate more experience to inform an eventual adoption decision.

- in Bangkok, we achieved rough consensus to adopt the feature into the base protocol. The plan as I understood it is that documents were to be held separate until the spin-exp document reflected Bangkok consensus and the transport document was done with large-scale technical changes, after which they would be merged. This plan together the discussion about 2119 language made it clear to me that the spin bit would be proposed standard.

This PR does the merge, because the conditions (less transport change in the future, spin bit now reflects the consensus about probabilistic participation) have been met.

While general guidelines on on-path use of the signal belongs in the (informational) manageability draft, details about how exactly to generate metrics from this signal are IMO more experimental, and probably fit better over in IPPM (personal opinion, not a ruling from the chair)

larseggert:
Let's put this on the agenda for Tokyo

britram:
(FWIW, I'm not in Tokyo, and not remote -- unavoidable conflict with a final project review at the Commission)

larseggert:
@britram can anyone participate who you trust to defend your views?

larseggert:
Also, we need to start an issue on this rather than commenting on the PR

draft-ietf-quic-transport.md
```diff
@@ -2818,6 +2825,44 @@ encoded as a single byte with the value 0x01.  An endpoint MAY treat the receipt
...
+initial RTT of a connection by observing the handshake. Therefore, the spin bit
+will appear after version negotiation and connection establishment are
+completed. On-path measurement and use of the Latency Spin Bit is further
+discussed in {{QUIC-MANAGEABILITY}}.
+
+The spin bit utilizes a single bit in the first byte of the short header. The
+location of the bit and procedures for how to set it by clients and servers are
+defined in {{short-header}}.
+
+Implementations MAY select to not implement the full spin bit functionality. In
```
britram: I think I'd phrase this differently: "the spin bit is an optional feature. implementations that implement MUST implement it as specified in this document."

draft-ietf-quic-transport.md
```diff
@@ -2818,6 +2825,44 @@ encoded as a single byte with the value 0x01.  An endpoint MAY treat the receipt
...
+that case they are only REQUIRED to implement what is defined for the spin bit
+when it is disabled.
+
+Each endpoint unilaterally decides if the spin bit is enabled or disabled for a
+connection. Implementations SHOULD allow administrators of clients and servers
+to disable the spin bit either globally or on a per-connection basis. Even when
+the spin bit is not disabled by the administrator implementations SHOULD disable
```
britram: IIRC consensus here sounded like MUST probabalistically disable, since we will want to provide cover traffic in the long term, not only for non-implementing endpoints but also for flows that decide to disable it. I'd justify the MUST because it affects the privacy properties of the wire image (however minimal the actual effect is)

```suggestion
the spin bit is not disabled by the administrator implementations MUST disable
```

draft-ietf-quic-transport.md
```diff
@@ -2818,6 +2825,44 @@ encoded as a single byte with the value 0x01.  An endpoint MAY treat the receipt
...
+the spin bit on a randomly chosen fraction of connections. However, connections
+may be configured to explicitly enable spinning, for example in the case of
```
britram: in line with the SHOULD above, this isn't a may/MAY:

```suggestion
SHOULD be configurable to explicitly enable spinning, for example in the case of
```

draft-ietf-quic-transport.md
britram: +1 to this

draft-ietf-quic-transport.md
```diff
@@ -89,6 +89,13 @@ informative:
     target:
      "https://web.archive.org/web/20150315054838/http://ha.ckers.org/slowloris/"
 
+  QUIC-MANAGEABILITY:
+    title: "Manageability of the QUIC Transport Protocol"
+    author:
+      - ins: M. Kuehlewind
```
britram: add `-ins: B. Trammell` as well

britram:
okay everybody let's take the discussion over to #2369

britram:
(@larseggert probably, will check; however, if discussion on #2369 doesn't converge relatively quickly then I'll want to take it to the list beforehand as well)

draft-ietf-quic-transport.md
ihlar: Agree, that is a better phrasing

draft-ietf-quic-transport.md
ihlar: >           Also, we need to start an issue on this rather than commenting on the PR
@larseggert I'm not sure what the issue is? 
The consensus in BKK was that the bit should be included in the standard, this PR only contains the information necessary for implementors, no text on how to interpret or process the bit on path. Simply referencing a non-milestone potentially experimental document seems a bit odd.

ihlar:
@larseggert I'm not sure what the issue is?
The consensus in BKK was that the bit should be included in the standard, this PR only contains the information necessary for implementors, no text on how to interpret or process the bit on path. Simply referencing a non-milestone potentially experimental document seems a bit odd.

draft-ietf-quic-transport.md
ihlar: Not sure why this would be a SHOULD, here we describe a case where a single connection has been explicitly configured to spin (e.g some debug or tracing mode enabled). In that case, spinning shouldn't be disabled for that specific connection as a result of the random selection process.

martinthomson:
It looks like the changes Brian is suggesting are editorial, but I'm not entirely sure about the changes to normative language.  Can we make those after any merge?  (Or even before it, given that this seems to be a little controversial...)

martinthomson:
Another note, in case I miss it.  spin-exp has a contributors/acknowledgments section.  If this merges, I want to talk to Brian about who we should be acknowledging, if anyone.

britram:
IMO the changes to normative language are editorial (reflecting my understanding of consensus already reached); have withdrawn one suggestion on review.

re acks: The EC/SERI ack won't be necessary to bring over from spin-exp. I'm not sure what your eventual plan is for filling the Contributors section of the -transport document; if Contributors is all-committers, I can check to see who wrote the text from which this PR is derived in the original -spin document.

larseggert:
Please discuss on #2369 and not on this PR. (Will unlock once that has concluded.)

draft-ietf-quic-transport.md
```diff
@@ -3948,8 +3993,33 @@ Fixed Bit:
 
 Spin Bit (S):
 
-: The sixth bit (0x20) of byte 0 is the Latency Spin Bit, set as described in
-  {{!SPIN=I-D.ietf-quic-spin-exp}}.
+: The third most significant bit (0x20) of byte 0 is the Latency Spin Bit.
+
+: When the spin bit is disabled, endpoints MAY set the spin bit to any value,
```
martinthomson: Please don't have so many paragraphs nested like this.  A forward reference to a section that includes this text would be much better.

draft-ietf-quic-transport.md
```diff
@@ -3948,8 +3993,33 @@ Fixed Bit:
...
+and MUST accept any incoming value. It is RECOMMENDED that they
+set the spin bit to a random value either chosen independently for each packet,
+or chosen independently for each path and kept constant for that path.
+
+: If the spin bit is enabled for the connection the endpoint maintains a spin
```
```suggestion
: If the spin bit is enabled for the connection, the endpoint maintains a spin
```

draft-ietf-quic-transport.md
```diff
@@ -3948,8 +3993,33 @@ Fixed Bit:
...
+value, 0 or 1, and sets the spin bit in the short header to the currently stored
```
```suggestion
value and sets the spin bit in the short header to the currently stored
```

draft-ietf-quic-transport.md
```diff
@@ -3997,6 +4067,7 @@ version.  See {{QUIC-INVARIANTS}} for details on how packets from different
 versions of QUIC are interpreted.
 
 
+
```
```suggestion
```

draft-ietf-quic-transport.md
```diff
@@ -2818,6 +2825,44 @@ encoded as a single byte with the value 0x01.  An endpoint MAY treat the receipt
...
+The latency spin bit enables passive latency monitoring from observation points
+on the network path throughout the duration of a connection. The spin bit is
+only present in the short packet header, since it is possible to measure the
+initial RTT of a connection by observing the handshake. Therefore, the spin bit
+will appear after version negotiation and connection establishment are
+completed. On-path measurement and use of the Latency Spin Bit is further
+discussed in {{QUIC-MANAGEABILITY}}.
+
+The spin bit utilizes a single bit in the first byte of the short header. The
```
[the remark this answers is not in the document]
britram: +1

draft-ietf-quic-transport.md
```diff
@@ -2818,6 +2825,44 @@ encoded as a single byte with the value 0x01.  An endpoint MAY treat the receipt
 of a frame type that uses a longer encoding than necessary as a connection error
 of type PROTOCOL_VIOLATION.
 
+## Latency Spin Bit {#spin-bit}
+
+The latency spin bit enables passive latency monitoring from observation points
+on the network path throughout the duration of a connection. The spin bit is
+only present in the short packet header, since it is possible to measure the
+initial RTT of a connection by observing the handshake. Therefore, the spin bit
+will appear after version negotiation and connection establishment are
```
larseggert: will appear -> is available

draft-ietf-quic-transport.md
larseggert: Agree, too. Would use the 2119 OPTIONAL capitalization

draft-ietf-quic-transport.md
```diff
@@ -2818,6 +2825,44 @@ encoded as a single byte with the value 0x01.  An endpoint MAY treat the receipt
...
+that case they are only REQUIRED to implement what is defined for the spin bit
+when it is disabled.
+
+Each endpoint unilaterally decides if the spin bit is enabled or disabled for a
+connection. Implementations SHOULD allow administrators of clients and servers
+to disable the spin bit either globally or on a per-connection basis. Even when
+the spin bit is not disabled by the administrator implementations SHOULD disable
+the spin bit on a randomly chosen fraction of connections. However, connections
+could be configured to explicitly enable spinning, for example in the case of
+explicit customer support and debugging.
```
larseggert: I'm not sure we had consensus for that, because if an app (erroneously) enables spin for all connections using this API, there would be no cover traffic, and we required there to always be some non-spin traffic.

draft-ietf-quic-transport.md
```diff
@@ -3948,8 +3993,33 @@ Fixed Bit:
...
+set the spin bit to a random value either chosen independently for each packet,
+or chosen independently for each path and kept constant for that path.
+
+: If the spin bit is enabled for the connection, the endpoint maintains a spin
+value and sets the spin bit in the short header to the currently stored
+value when a packet with a short header is sent out. The spin value is
+initialized to 0 in the endpoint at connection start.  Each endpoint also
+remembers the highest packet number seen from its peer on the connection.
+
+: When a server receives a packet from a client, if that packet has a short
```
larseggert: Simply say "when a server receives a short-header packet"?

draft-ietf-quic-transport.md
```diff
@@ -3948,8 +3993,33 @@ Fixed Bit:
...
+header and if it increments the highest packet number seen by the server from
+the client, the server sets the spin value to the value observed in the spin bit
+in the received packet.
+
+: When a client receives a packet from a server, if the packet has a short
```
larseggert: Ditto

martinthomson:
Note to anyone who commented here, please see the updated PR.

Issue #2255, opened 2018-12-25
persistent congestion is more aggressive than RTO verified


marten-seemann:
Consider the following sequence of packets being sent

|  | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 |
| -- | -- | -- | -- | -- | -- | -- | -- | -- |
| old: | packet | packet | TLP 1 | TLP 1 | TLP 2 | TLP 2 | RTO | RTO |
| new: | packet | packet | PTO 1 | PTO 1 | PTO 2 | PTO 2 | PTO 3 | PTO 3 |

In our old loss recovery code (before unifying TLP and RTO, #2114), we used the following logic for determining if an RTO was not spurious, which would then lead to collapse of the congestion window:
```
if (rto_count > 0 && smallest_newly_acked > largest_sent_before_rto):
  OnRetransmissionTimeoutVerified(smallest_newly_acked)
```
In the example above `largest_sent_before_rto` would be packet 14. This means that to count as a verified RTO, an ACK must acknowledge 15 or 16, but not *any* packets smaller or equal than 14.

With #2114, we now collapse the congestion window when `pto_count` is larger than 2 and we receive an ACK which leads us to declare *any* packet lost. In the example above this means any ACK received after sending packet 15 which declares a packet lost would suffice. Even worse, this logic already triggers if we receive the ACK *immediately after* sending 15, i.e. when the peer didn't even have a chance to receive it.

This seems  **a lot more aggressive** than what we had before before. Is this intentional?

ianswett:
This is unintentional.  I belIeve Jana is intending to fix this in PR #2244, so please take a look at that.

marten-seemann:
If I understand correctly, #2244 is a fix for #2243 and only makes the description of the constant and the pseudo code consistent.

ianswett:
I'm removing design.  I consider this a bug fix, but still very important.

martinthomson:
I've been using `design` to mark things that will result in changes that affect implementations.  This would seem to fit that description.  Whether you consider it worth discussing is different.  If it isn't worth discussing, maybe it will just be fixed by #2293 and everything will be fine without discussion in Tokyo.

WIthout the label, this won't make the change log, and that might be a problem.

PR #2365, opened 2019-01-23
Persistent Congestion Time Threshold

### LINKED ISSUE #2356: Persistent Congestion Requires Probes to be Sent
Currently, persistent congestion is only determine when the required number of probes have been sent. But probes aren't sent if the app layer continues to send data. This likely means the persistent congestion determination should be time based instead of probe count based.

nibanks:
Updates the persistent congestion logic to be time based instead of PTO count based, to account for the cases where PTOs don't actually occur. Fixes #2356 and #2255.

This is the simplest change I could come up with that fixes the main issue, though I think a completely separate timer would still be better than piggy-backing off the `CongestionEvent`.

For instance, in the case the app is periodically sending (and therefore no PTOs occur), if the ACK eventually comes much later than when persistent congestion occurs (or should occur) the CC window only decreases when the ACK arrives; not when persistent congestion started. If you have a separate timer for persistent congestion the window could be reduced immediately instead of waiting for the ACK. For now, though, I chose to go with the simpler change/solution that still mostly gets the job done.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -44,6 +44,23 @@ normative:
         org: Mozilla
         role: editor
 
+  QUIC-TLS:
+    title: "Using TLS to Secure QUIC"
+    date: {DATE}
+    seriesinfo:
+      Internet-Draft: draft-ietf-quic-tls-latest
+    author:
+      -
+        ins: M. Thomson
+        name: Martin Thomson
+        org: Mozilla
+        role: editor
+      -
+        ins: S. Turner
+        name: Sean Turner
+        org: sn3rd
+        role: editor
+
 informative:
 
   FACK:
@@ -434,14 +451,30 @@ packet is received.  The client MAY use this value to seed the RTT estimator for
 a subsequent connection attempt to the server.
 
 
-#### Discarding Initial State {#discard-initial}
+#### Discarding Keys and Packet State {#discarding-packets}
 
-As described in Section 17.5.1 of {{QUIC-TRANSPORT}}, endpoints stop sending and
-receiving Initial packets once they start exchanging Handshake packets.  At this
-point, all loss recovery state for the Initial packet number space is also
-discarded. Packets that are in flight for the packet number space are not
-declared as either acknowledged or lost.  After discarding state, new Initial
-packets will not be sent.
+When packet protection keys are discarded (see Section 4.9 of {{QUIC-TLS}}), all
+packets that were sent with those keys can no longer be acknowledged because
+their acknowledgements cannot be processed anymore. The sender considers them no
+longer in flight. That is, the sender SHOULD discard all recovery state
+associated with those packets and MUST remove them from the count of bytes in
+flight.
+
+Endpoints stop sending and receiving Initial packets once they start exchanging
+Handshake packets (see Section 17.2.2.1 of {{QUIC-TRANSPORT}}). At this point,
+recovery state for all in-flight Initial packets is discarded.
+
+When 0-RTT is rejected, recovery state for all in-flight 0-RTT packets is
+discarded.
+
+If a server accepts 0-RTT, but does not buffer 0-RTT packets that arrive
+before Initial packets, early 0-RTT packets will be declared lost, but that
+is expected to be infrequent.
+
+It is expected that keys are discarded after packets encrypted with them would
+be acknowledged or declared lost.  Initial secrets however might be destroyed
+sooner, as soon as handshake keys are available (see Section 4.10 of
+{{QUIC-TLS}}).
 
 
 ### Probe Timeout {#pto}
@@ -952,6 +985,14 @@ The recovery period limits congestion window reduction to once per round trip.
 During recovery, the congestion window remains unchanged irrespective of new
 losses or increases in the ECN-CE counter.
 
+## Ignoring Loss of Undecryptable Packets
+
+During the handshake, some packet protection keys might not be
+available when a packet arrives. In particular, Handshake and 0-RTT packets
+cannot be processed until the Initial packets arrive, and 1-RTT packets
+cannot be processed until the handshake completes.  Endpoints MAY
+ignore the loss of Handshake, 0-RTT, and 1-RTT packets that might arrive before
+the peer has packet protection keys to process those packets.
 
 ## Probe Timeout
 
@@ -1005,23 +1046,6 @@ paces the sending of any packets in excess of the initial congestion window.
 
 A sender MAY implement alternate mechanisms to update its congestion window
 after idle periods, such as those proposed for TCP in {{?RFC7661}}.
-
-## Discarding Packet Number Space State
-
-When keys for a packet number space are discarded, any in-flight packets
-sent with those keys are removed from the count of bytes in flight.  Loss
-recovery state is also discarded, so no loss events will occur for any
-in-flight packets from that space (see {{discard-initial}}).  Note that it is
-expected that keys are discarded after those packets would be declared lost,
-but Initial secrets are destroyed earlier.
-
-When 0-RTT is rejected, all in-flight 0-RTT packets are removed from
-the count of bytes in flight.  Loss recovery state is also discarded, so no
-loss events will occur for any in-flight 0-RTT packets.
-
-If a server accepts 0-RTT, but does not buffer 0-RTT packets that arrive
-before Initial packets, early 0-RTT packets will be declared lost, but that
-is expected to be infrequent.
 
 ## Pseudocode
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1223,7 +1223,7 @@ are detected lost.
      pto = smoothed_rtt + 4 * rttvar + max_ack_delay
      return
        congestion_period >
-       pto * (2 ^ kPersistentCongestionThreshold - 1)
+         pto * (2 ^ kPersistentCongestionThreshold - 1)
 
    OnPacketsLost(lost_packets):
      // Remove lost packets from bytes_in_flight.
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -653,8 +653,12 @@ experiencing persistent congestion.  Commonly, this can be established by
 consecutive PTOs, but since the PTO timer is reset when a new ack-eliciting
 packet is sent, an explicit duration must be used to account for those cases
 where PTOs do not occur or are substantially delayed.  This duration is the
-equivalent of kPersistentCongestionThreshold consecutive PTOS: (smoothed_rtt +
-4 * rttvar + max_ack_delay) * ((2 ^ kPersistentCongestionThreshold) - 1).
+equivalent of kPersistentCongestionThreshold consecutive PTOS, and is computed
+as follows:
+~~~
+(smoothed_rtt + 4 * rttvar + max_ack_delay) * 
+    ((2 ^ kPersistentCongestionThreshold) - 1).
+~~~
 
 For example, assume:
 
@@ -682,7 +686,7 @@ When persistent congestion is established, the sender's congestion window MUST
 be reduced to the minimum congestion window (kMinimumWindow).  This response of
 collapsing the congestion window on persistent congestion is functionally
 similar to a sender's response on a Retransmission Timeout (RTO) in TCP
-{{RFC5681}}.
+{{RFC5681}} after Tail Loss Probes (TLP) {{TLP}}.
 
 ## Pacing {#pacing}
 
@@ -1311,9 +1315,8 @@ are detected lost.
 ~~~
    InPersistentCongestion(congestion_period):
      pto = smoothed_rtt + 4 * rttvar + max_ack_delay
-     return
-       congestion_period >
-         pto * (2 ^ kPersistentCongestionThreshold - 1)
+     return congestion_period >
+       pto * (2 ^ kPersistentCongestionThreshold - 1)
 
    OnPacketsLost(lost_packets):
      // Remove lost packets from bytes_in_flight.
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1009,7 +1009,7 @@ experiencing persistent congestion.  Commonly, this can be established by
 consecutive PTOs, but since the PTO timer is reset when a new ack-eliciting
 packet is sent, an explicit duration must be used to account for those cases
 where PTOs do not occur or are substantially delayed.  This duration is the
-equivalent of kPersistentCongestionThreshold consecutive PTOS, smoothed_rtt +
+equivalent of kPersistentCongestionThreshold consecutive PTOS: smoothed_rtt +
 4 * rttvar + max_ack_delay * ((2 ^ kPersistentCongestionThreshold) - 1).
 
 When persistent congestion is established, the sender's congestion window MUST
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -672,11 +672,11 @@ illustrate persistent congestion:
   t=7 | Send Pkt #4 (PTO 3)
   t=8 | Recv ACK of Pkt #4
   
-The first three packets determined to be lost when the ACK of packet 4 is
+The first three packets are determined to be lost when the ACK of packet 4 is
 received at t=8.  The congestion period is calculated as the time between the
 oldest and newest lost packets: (3 - 0) = 3.  The duration for persistent
-congestion is equal to: (1 * 3) = 3.  Because the threshold was reached, the
-network is considered to have experienced persistent congestion.
+congestion is equal to: (1 * ((2 ^ 2) - 1)) = 3.  Because the threshold was
+reached, the network is considered to have experienced persistent congestion.
 
 When persistent congestion is established, the sender's congestion window MUST
 be reduced to the minimum congestion window (kMinimumWindow).  This response of
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1003,14 +1003,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
 until an acknowledgement is received that establishes loss or delivery of
 packets.
 
-When an ACK frame is received that establishes loss of all in-flight packets
-sent prior to a threshold number of consecutive PTOs (pto_count is more than
-kPersistentCongestionThreshold, see {{cc-consts-of-interest}}), the network is
-considered to be experiencing persistent congestion, and the sender's congestion
-window MUST be reduced to the minimum congestion window (kMinimumWindow).  This
-response of collapsing the congestion window on persistent congestion is
-functionally similar to a sender's response on a Retransmission Timeout (RTO) in
-TCP {{RFC5681}}.
+## Persistent Congestion
+
+The network is considered to be experiencing persistent congestion when the
+sender goes a period of time without receiving any ACK frames for its in-flight,
+ack-eliciting packets.  Generally, this can be calculated as a number of
+consecutive PTOs (pto_count is more than  kPersistentCongestionThreshold, see
+{{cc-consts-of-interest}}), but since a PTO only occurs if no other
+ack-eliciting packets have been sent, an explicit time out must be used to
+account for those cases where PTOs don't occur.
+
+When persistent congestion is encountered the sender's congestion window MUST be
+reduced to the minimum congestion window (kMinimumWindow).  This response of
+collapsing the congestion window on persistent congestion is functionally
+similar to a sender's response on a Retransmission Timeout (RTO) in TCP
+{{RFC5681}}.
 
 
 ## Pacing
@@ -1103,6 +1110,10 @@ bytes_in_flight:
 congestion_window:
 : Maximum number of bytes-in-flight that may be sent.
 
+probe_start_time:
+: The time when QUIC started probing the network, after the most recent ACK
+  received from the peer.
+
 recovery_start_time:
 : The time when QUIC first detects a loss, causing it to enter recovery.
   When a packet sent after this time is acknowledged, QUIC exits recovery.
@@ -1120,6 +1131,7 @@ variables as follows:
 ~~~
    congestion_window = kInitialWindow
    bytes_in_flight = 0
+   probe_start_time = 0
    recovery_start_time = 0
    ssthresh = infinite
    ecn_ce_counter = 0
@@ -1132,6 +1144,8 @@ increases bytes_in_flight.
 
 ~~~
    OnPacketSentCC(bytes_sent):
+     if (bytes_in_flight == 0):
+       probe_start_time = Now()
      bytes_in_flight += bytes_sent
 ~~~
 
@@ -1147,6 +1161,8 @@ acked_packet from sent_packets.
    OnPacketAckedCC(acked_packet):
      // Remove from bytes_in_flight.
      bytes_in_flight -= acked_packet.size
+     if (bytes_in_flight != 0):
+       probe_start_time = Now()
      if (InRecovery(acked_packet.time_sent)):
        // Do not increase congestion window in recovery period.
        return
@@ -1166,6 +1182,12 @@ detected. May start a new recovery period and reduces the congestion
 window.
 
 ~~~
+   InPersistentCongestion():
+     return
+       (Now() - probe_start_time) >
+       (smoothed_rtt + 4 * rttvar + max_ack_delay) *
+       ((2 ^ kPersistentCongestionThreshold) - 1)
+
    CongestionEvent(sent_time):
      // Start a new congestion event if the sent time is larger
      // than the start time of the previous recovery epoch.
@@ -1175,7 +1197,7 @@ window.
        congestion_window = max(congestion_window, kMinimumWindow)
        ssthresh = congestion_window
        // Collapse congestion window if persistent congestion
-       if (pto_count > kPersistentCongestionThreshold):
+       if (InPersistentCongestion()):
          congestion_window = kMinimumWindow
 ~~~
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1006,10 +1006,11 @@ packets.
 When an ACK frame is received that establishes loss of all in-flight packets
 sent over a long enough period of time, the network is considered to be
 experiencing persistent congestion.  Commonly, this can be established by
-consecutive PTOs (pto_count is more than kPersistentCongestionThreshold, see
-{{cc-consts-of-interest}}), but since the PTO timer is reset when a new
-ack-eliciting packet is sent, an explicit duration must be used to account for
-those cases where PTOs do not occur or are substantially delayed.
+consecutive PTOs, but since the PTO timer is reset when a new ack-eliciting
+packet is sent, an explicit duration must be used to account for those cases
+where PTOs do not occur or are substantially delayed.  This duration is the
+equivalent of kPersistentCongestionThreshold consecutive PTOS, smoothed_rtt +
+4 * rttvar + max_ack_delay * ((2 ^ kPersistentCongestionThreshold) - 1).
 
 When persistent congestion is established, the sender's congestion window MUST
 be reduced to the minimum congestion window (kMinimumWindow).  This response of
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1219,7 +1219,7 @@ are detected lost.
 
 ~~~
    InPersistentCongestion(oldest_loss_time):
-     elapsed_time = Now() - oldest_loss_time
+     congestion_period = Now() - oldest_loss_time
      pto = smoothed_rtt + 4 * rttvar + max_ack_delay
      return
        elapsed_time >
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1242,6 +1242,23 @@ This document has no IANA actions.  Yet.
 
 Issue and pull request numbers are listed with a leading octothorp.
 
+## Since draft-ietf-quic-recovery-17
+
+- After Probe Timeout discard in-flight packets or send another (#2212, #1965)
+- Endpoints discard initial keys as soon as handshake keys are available (#1951,
+  #2045)
+- 0-RTT state is discarded when 0-RTT is rejected (#2300)
+- Loss detection timer is cancelled when ack-eliciting frames are in flight
+  (#2117, #2093)
+- Packets are declared lost if they are in flight (#2104)
+- After becoming idle, either pace packets or reset the congestion controller
+  (#2138, 2187)
+- Process ECN counts before marking packets lost (#2142)
+- Mark packets lost before resetting crypto_count and pto_count (#2208, #2209)
+- Congestion and loss recovery state are discarded when keys are discarded
+  (#2237)
+
+
 ## Since draft-ietf-quic-recovery-16
 
 - Unify TLP and RTO into a single PTO; eliminate min RTO, min TLP and min crypto
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -179,8 +179,8 @@ QUIC uses separate packet number spaces for each encryption level, except 0-RTT
 and all generations of 1-RTT keys use the same packet number space.  Separate
 packet number spaces ensures acknowledgement of packets sent with one level of
 encryption will not cause spurious retransmission of packets sent with a
-different encryption level.  Congestion control and RTT measurement are unified
-across packet number spaces.
+different encryption level.  Congestion control and round-trip time (RTT)
+measurement are unified across packet number spaces.
 
 ### Monotonically Increasing Packet Numbers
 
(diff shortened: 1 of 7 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -1003,14 +1003,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
-response of collapsing the congestion window on persistent congestion is
-functionally similar to a sender's response on a Retransmission Timeout (RTO) in
-TCP {{RFC5681}}.
+## Persistent Congestion
+
+The network is considered to be experiencing persistent congestion when the
+sender goes a period of time without receiving any ACK frames for its in-flight,
+ack-eliciting packets.  Generally, this can be calculated as a number of
+consecutive PTOs (pto_count is more than  kPersistentCongestionThreshold, see
+{{cc-consts-of-interest}}), but since a PTO only occurs if no other
```
ianswett: How about "but since the PTO timer is reset when a new ack-eliciting packet is sent,"

draft-ietf-quic-recovery.md
```diff
@@ -1132,6 +1144,8 @@ increases bytes_in_flight.
 
 ~~~
    OnPacketSentCC(bytes_sent):
+     if (bytes_in_flight == 0):
+       probe_start_time = Now()
```
ianswett: It's important not to start this if the packet isn't ack-eliciting, but I will say that requirement makes this more complex, since you have to track the first ack-eliciting packet to be sent.

draft-ietf-quic-recovery.md
```diff
@@ -1147,6 +1161,8 @@ acked_packet from sent_packets.
    OnPacketAckedCC(acked_packet):
      // Remove from bytes_in_flight.
      bytes_in_flight -= acked_packet.size
+     if (bytes_in_flight != 0):
+       probe_start_time = Now()
```
ianswett: I don't think it's necessary to check bytes_in_flight != 0 here?

draft-ietf-quic-recovery.md
```diff
@@ -1003,14 +1003,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
 
-When an ACK frame is received that establishes loss of all in-flight packets
-sent prior to a threshold number of consecutive PTOs (pto_count is more than
-kPersistentCongestionThreshold, see {{cc-consts-of-interest}}), the network is
-considered to be experiencing persistent congestion, and the sender's congestion
-window MUST be reduced to the minimum congestion window (kMinimumWindow).  This
-response of collapsing the congestion window on persistent congestion is
-functionally similar to a sender's response on a Retransmission Timeout (RTO) in
-TCP {{RFC5681}}.
+## Persistent Congestion
```
ianswett: I think this should be a sub-section of Probe Timeout?

draft-ietf-quic-recovery.md
nibanks: Yeah, you are right. I'm not sure what I was thinking when I added this. I will remove it.

draft-ietf-quic-recovery.md
nibanks: So, are you suggestion a change? Perhaps a comment?

draft-ietf-quic-recovery.md
nibanks: I'm not super opinionated either way, but I feel like coupling the two things go us into this problem in the first place. They are related but still fairly independent in my mind. Especially, if we ever go with the model of a separate timer.

draft-ietf-quic-recovery.md
nibanks: Sure.

draft-ietf-quic-recovery.md
```diff
@@ -1003,14 +1003,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
+
+The network is considered to be experiencing persistent congestion when the
+sender goes a period of time without receiving any ACK frames for its in-flight,
+ack-eliciting packets.  Generally, this can be calculated as a number of
```
mikkelfj: I'm getting used to that word ...

mikkelfj:
I'm wondering how many concurrent timers there are. Is it always at most one timer?
The probe timer is updated fairly frequently and that is actually a heavy operation if it is inserted into red-black tree (as is popular though suboptimal) when there are many concurrent connections.

Another approach would be to poll the state periodically and take action at those intervals. This could be done by not resetting the timer but inspecting state once it expires. Granted, a timeout might also be costly.

draft-ietf-quic-recovery.md
```diff
@@ -1132,6 +1144,8 @@ increases bytes_in_flight.
 
 ~~~
    OnPacketSentCC(bytes_sent):
+     if (bytes_in_flight == 0):
```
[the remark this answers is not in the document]
nibanks: That doesn't quite work. What if the first packet isn't ack_eliciting but the second is? The first would still increment the counter, so when the second comes in it wouldn't be 0 any more.

draft-ietf-quic-recovery.md
```diff
@@ -1003,14 +1003,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
-functionally similar to a sender's response on a Retransmission Timeout (RTO) in
-TCP {{RFC5681}}.
+## Persistent Congestion
+
+The network is considered to be experiencing persistent congestion when the
+sender goes a period of time without receiving any ACK frames for its in-flight,
+ack-eliciting packets.  Generally, this can be calculated as a number of
+consecutive PTOs (pto_count is more than  kPersistentCongestionThreshold, see
+{{cc-consts-of-interest}}), but since a PTO only occurs if no other
+ack-eliciting packets have been sent, an explicit time out must be used to
```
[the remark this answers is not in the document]
nibanks: Could you make these suggestions on the latest changes so I can accept them more easily?

draft-ietf-quic-recovery.md
```diff
@@ -1003,15 +1003,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
-TCP {{RFC5681}}.
-
+## Persistent Congestion
+
+The network is considered to be experiencing persistent congestion when the
+sender goes a period of time without receiving any ACK frames for its in-flight,
+ack-eliciting packets.  Generally, this can be calculated as a number of
+consecutive PTOs (pto_count is more than  kPersistentCongestionThreshold, see
+{{cc-consts-of-interest}}), but since the PTO timer is reset when a new
+ack-eliciting packet is sent, an explicit time out must be used to account for
```
```suggestion
ack-eliciting packet is sent, an explicit timeout must be used to account for
```

draft-ietf-quic-recovery.md
```diff
@@ -1003,15 +1003,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
+those cases where PTOs don't occur.
```
```suggestion
those cases where PTOs do not occur.
```

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
mikkelfj: right

draft-ietf-quic-recovery.md
```diff
@@ -1120,6 +1130,7 @@ variables as follows:
 ~~~
    congestion_window = kInitialWindow
    bytes_in_flight = 0
```
mikkelfj: `probe_pending = 0`

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
mikkelfj:
```
if (ack_eliciting && (bytes_in_flight == 0 || probe_pending)):
    probe_pending = 0
    probe_start_time = Now()
else:
   probe_pending = 1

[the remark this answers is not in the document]
Comment (mikkelfj) [reply draft-ietf-quic-recovery.md]: It's too late, I keep getting this wrong - wait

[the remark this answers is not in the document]
Comment (mikkelfj) [reply draft-ietf-quic-recovery.md]: Edited above comment, I think it is ok now

File: draft-ietf-quic-recovery.md [hunk 11]
```diff
@@ -1202,15 +1205,26 @@ Invoked by loss detection from DetectLostPackets when new packets
 are detected lost.
 
 ~~~
+   InPersistentCongestion(oldest_loss_time):
+     return
```
ianswett: This expression is hard to read, I think it's worth breaking it up some.
 elapsed_time = Now() - oldest_loss_time
 pto = smoothed_rtt + 4 * rttvar + max_ack_delay
 return elapsed_time > pto * (2 ^ kPersistentCongestionThreshold - 1)

ianswett:
I think it makes sense to fix #2255 with this PR as well?

draft-ietf-quic-recovery.md
```diff
@@ -1003,15 +1003,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
-kPersistentCongestionThreshold, see {{cc-consts-of-interest}}), the network is
-considered to be experiencing persistent congestion, and the sender's congestion
-window MUST be reduced to the minimum congestion window (kMinimumWindow).  This
-response of collapsing the congestion window on persistent congestion is
-functionally similar to a sender's response on a Retransmission Timeout (RTO) in
-TCP {{RFC5681}}.
-
+## Persistent Congestion
+
+The network is considered to be experiencing persistent congestion when the
```
ianswett: The old text was more correct than this text in the sense that it's important that not only does time pass, but all the packets are lost within that period of time.

How about:
"When an ACK frame is received that establishes loss of all in-flight packets sent over a long enough period of time, the network is considered to be experiencing persistent congestion."

draft-ietf-quic-recovery.md
```diff
@@ -1003,15 +1003,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
+sender goes a period of time without receiving any ACK frames for its in-flight,
+ack-eliciting packets.  Generally, this can be calculated as a number of
+consecutive PTOs (pto_count is more than  kPersistentCongestionThreshold, see
+{{cc-consts-of-interest}}), but since the PTO timer is reset when a new
+ack-eliciting packet is sent, an explicit timeout must be used to account for
+those cases where PTOs do not occur.
+
+When persistent congestion is encountered the sender's congestion window MUST be
```
```suggestion
When persistent congestion is established, the sender's congestion window MUST be
```

draft-ietf-quic-recovery.md
```diff
@@ -1003,15 +1003,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
-window MUST be reduced to the minimum congestion window (kMinimumWindow).  This
-response of collapsing the congestion window on persistent congestion is
-functionally similar to a sender's response on a Retransmission Timeout (RTO) in
-TCP {{RFC5681}}.
-
+## Persistent Congestion
+
+The network is considered to be experiencing persistent congestion when the
+sender goes a period of time without receiving any ACK frames for its in-flight,
+ack-eliciting packets.  Generally, this can be calculated as a number of
```
ianswett: Suggestion: replace "Generally, this can be calculated as a number of consecutive PTOs (),..." with
"Commonly, this can be established by consecutive PTOs,..."

ianswett:
#2255 should only need the pseudocode fixed up, so your current PR should be sufficient once the text is in good shape.

draft-ietf-quic-recovery.md
nibanks: For your suggestion, do you really need the word "all" in there. To me, the loss of any in-flight packets over a long enough time seems to be the requirement (at least how we have the pseudocode). It think the removal the word "all" might be better. Other than that, I am fine with making the change.

ianswett: It has to be all packets lost within that period of time(ie: total blackout).  And now that you mention it, that means the pseudocode isn't quite right either.

As #2255 details, the goal is to be consistent with the previous behavior in terms of when persistent congestion would be established.

draft-ietf-quic-recovery.md
```diff
@@ -1202,15 +1205,25 @@ Invoked by loss detection from DetectLostPackets when new packets
...
+     elapsed_time = Now() - oldest_loss_time
+     pto = smoothed_rtt + 4 * rttvar + max_ack_delay
+     return elapsed_time > pto * (2 ^ kPersistentCongestionThreshold - 1)
+
    OnPacketsLost(lost_packets):
      // Remove lost packets from bytes_in_flight.
      for (lost_packet : lost_packets):
        bytes_in_flight -= lost_packet.size
+     smallest_lost_packet = lost_packets.first()
      largest_lost_packet = lost_packets.last()
```
nibanks: Slight tangent to this PR: what do folks think about changing these variables to `oldest_lost_packet` and `newest_lost_packet`?

ianswett: I don't have a strong preference either way.

draft-ietf-quic-recovery.md
nibanks: I'm not sure how we'd update the pseudocode to indicate loss of all packets. I'd have to think a bit more about it.

nibanks:
@marten-seemann Personally, I don't think `DetectLostPackets` should be called from `OnLossDetectionTimeout` any more, but I'd like @janaiyengar to chime in. My understanding of the new PTO design is that ONLY ACK frames are supposed to trigger loss. Timers only cause additional packets to be sent, in order to trigger ACKs from the peer.

marten-seemann:
@nibanks I think the reason we call `DetectLostPackets` from `OnLossDetectionTimeout` is in order for Early Retransmit to work. We might need some more code refactoring here to achieve the desired behavior. Let's discuss this with @janaiyengar.

draft-ietf-quic-recovery.md
```diff
@@ -1001,14 +1001,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
-sent prior to a threshold number of consecutive PTOs (pto_count is more than
-kPersistentCongestionThreshold, see {{cc-consts-of-interest}}), the network is
-considered to be experiencing persistent congestion, and the sender's congestion
-window MUST be reduced to the minimum congestion window (kMinimumWindow).  This
-response of collapsing the congestion window on persistent congestion is
-functionally similar to a sender's response on a Retransmission Timeout (RTO) in
-TCP {{RFC5681}}.
+sent over a long enough period of time, the network is considered to be
+experiencing persistent congestion.  Commonly, this can be established by
+consecutive PTOs (pto_count is more than  kPersistentCongestionThreshold, see
```
```suggestion
consecutive PTOs (pto_count is more than kPersistentCongestionThreshold, see
```

draft-ietf-quic-recovery.md
```diff
@@ -1001,14 +1001,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
+{{cc-consts-of-interest}}), but since the PTO timer is reset when a new
+ack-eliciting packet is sent, an explicit timeout must be used to account for
+those cases where PTOs do not occur.
```
```suggestion
those cases where PTOs do not occur or are substantially delayed.
```

draft-ietf-quic-recovery.md
```diff
@@ -1001,14 +1001,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
-considered to be experiencing persistent congestion, and the sender's congestion
-window MUST be reduced to the minimum congestion window (kMinimumWindow).  This
-response of collapsing the congestion window on persistent congestion is
-functionally similar to a sender's response on a Retransmission Timeout (RTO) in
-TCP {{RFC5681}}.
+sent over a long enough period of time, the network is considered to be
+experiencing persistent congestion.  Commonly, this can be established by
+consecutive PTOs (pto_count is more than  kPersistentCongestionThreshold, see
+{{cc-consts-of-interest}}), but since the PTO timer is reset when a new
+ack-eliciting packet is sent, an explicit timeout must be used to account for
```
ianswett: I don't think it's really a timeout.  How about period or duration or time?

draft-ietf-quic-recovery.md
ianswett: Also, I'd remove the () here and then define what the actual period is at the end of this paragraph.

draft-ietf-quic-recovery.md
```diff
@@ -1214,15 +1218,27 @@ Invoked by loss detection from DetectLostPackets when new packets
 are detected lost.
 
 ~~~
+   InPersistentCongestion(oldest_loss_time):
+     elapsed_time = Now() - oldest_loss_time
```
```suggestion
     congestion_period = Now() - oldest_loss_time
```

draft-ietf-quic-recovery.md
nibanks: I think time or duration might fit.

draft-ietf-quic-recovery.md
```diff
@@ -1001,14 +1001,21 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
-sent prior to a threshold number of consecutive PTOs (pto_count is more than
-kPersistentCongestionThreshold, see {{cc-consts-of-interest}}), the network is
-considered to be experiencing persistent congestion, and the sender's congestion
-window MUST be reduced to the minimum congestion window (kMinimumWindow).  This
-response of collapsing the congestion window on persistent congestion is
-functionally similar to a sender's response on a Retransmission Timeout (RTO) in
-TCP {{RFC5681}}.
+sent over a long enough period of time, the network is considered to be
+experiencing persistent congestion.  Commonly, this can be established by
+consecutive PTOs (pto_count is more than kPersistentCongestionThreshold, see
```
```suggestion
consecutive PTOs
```

draft-ietf-quic-recovery.md
```diff
@@ -1001,14 +1001,22 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
-window MUST be reduced to the minimum congestion window (kMinimumWindow).  This
-response of collapsing the congestion window on persistent congestion is
-functionally similar to a sender's response on a Retransmission Timeout (RTO) in
-TCP {{RFC5681}}.
+sent over a long enough period of time, the network is considered to be
+experiencing persistent congestion.  Commonly, this can be established by
+consecutive PTOs, but since the PTO timer is reset when a new ack-eliciting
+packet is sent, an explicit duration must be used to account for those cases
+where PTOs do not occur or are substantially delayed.  This duration is the
+equivalent of kPersistentCongestionThreshold consecutive PTOS, smoothed_rtt +
```
```suggestion
equivalent of kPersistentCongestionThreshold consecutive PTOS: smoothed_rtt +
```

draft-ietf-quic-recovery.md
```diff
@@ -1214,15 +1219,27 @@ Invoked by loss detection from DetectLostPackets when new packets
...
+     oldest_lost_packet = lost_packets.first()
+     newest_lost_packet = lost_packets.last()
 
      // Start a new congestion epoch if the last lost packet
      // is past the end of the previous recovery epoch.
-     CongestionEvent(largest_lost_packet.time_sent)
+     CongestionEvent(newest_lost_packet.time_sent)
+
+     // Collapse congestion window if persistent congestion
+     if (InPersistentCongestion(oldest_lost_packet.time_sent)):
```
ianswett: I'm starting to think we should pass in congestion_period and make it oldest_lost_packet - newest_lost_packet.

I think using Now() may be incorrect.  What if one packet is lost and then a few ACKs are lost.  In this case, Now() is substantially larger than oldest_lost_packet, but in fact only one packet was lost.

draft-ietf-quic-recovery.md
```diff
@@ -1001,14 +1001,22 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
-response of collapsing the congestion window on persistent congestion is
-functionally similar to a sender's response on a Retransmission Timeout (RTO) in
-TCP {{RFC5681}}.
+sent over a long enough period of time, the network is considered to be
+experiencing persistent congestion.  Commonly, this can be established by
+consecutive PTOs, but since the PTO timer is reset when a new ack-eliciting
+packet is sent, an explicit duration must be used to account for those cases
+where PTOs do not occur or are substantially delayed.  This duration is the
+equivalent of kPersistentCongestionThreshold consecutive PTOS, smoothed_rtt +
+4 * rttvar + max_ack_delay * ((2 ^ kPersistentCongestionThreshold) - 1).
```
ianswett: Q for here and below: Should max_ack_delay be added after exponential backoff, instead of before?  Adding it after incentivizes implementations to indicate a realistic value, since an unrealistic one just adds to rttvar, which seems desirable.

nibanks: The order is the same in `SetLossDetectionTimer`:
```    // Calculate PTO duration
    timeout =
      smoothed_rtt + 4 * rttvar + max_ack_delay
    timeout = timeout * (2 ^ pto_count)```
Though, on a side issue, I feel the `kGranularity` check should be moved to the end of that calculation...

Comment (ianswett) [reply draft-ietf-quic-recovery.md]: Agreed, what you have matches the current pseudocode.  I'll open an issue for kGranularity and max_ack_delay and when they should be added.

**Comment by nibanks** [comment]
@ianswett / @janaiyengar anything else that needs to be fixed as part of this PR?

File: draft-ietf-quic-recovery.md [hunk 24]
```diff
@@ -1282,15 +1287,27 @@ Invoked by loss detection from DetectLostPackets when new packets
 are detected lost.
 
 ~~~
+   InPersistentCongestion(oldest_loss_time):
+     congestion_period = Now() - oldest_loss_time
```
ianswett: I think I concluded newest_lost_time - oldest_loss_time is the best option here.  Using Now() means that a single lost packet combined with some temporary ack loss causes this to trigger, which I do not believe is the intent.

draft-ietf-quic-recovery.md
ianswett: I'd like the opinion of @janaiyengar here.  I think this is likely fairly good, but I think passing in newest_lost_time is easy and more correct, given they're easily available in the pseudocode.

nibanks: Let me propose an example. Assume smooth_rtted = 1 sec, rttvar = 0, max_ack_delay = 0 for simplicity's sake.
```
t=0 | Send Pkt #1 (App Data)
t=1 | Send Pkt #2 (PTO 1)
t=3 | Send Pkt #3 (PTO 2)
t=7 | Send Pkt #4 (PTO 3)
t=8 | Recv ACK of Pkt #4
```
At t=8 you'd get a congestion event for packets 1 & 2 being lost. If you use Now() for the persistent congestion calculation, this would be declared as persistent congestion. If you used newest_lost_time it wouldn't. Which behavior do we want?

ianswett: If 4 is acked and the others are lost, then it should be persistent congestion.

But if 2, 3 and 4 were all acked at once(t=8), that should not be persistent congestion.

nibanks: Hmm, well then we'd have to make two changes to achieve that behavior.
1. Use `newest_lost_packet - oldest_lost_packet` for the time window.
2. Make the threshold 2 PTO instead of 3.

Without the second part above, an ACK of pkt 4 would produce a loss window of only 3 seconds (3 - 0), right?

ianswett: That sounds correct to me.  Can you add your example as an example to the text section as well?  This is clearly very confusing, so an example would help.

nibanks: Sure. Do you want me to describe the example in just words, or put almost exactly what I have above with the text?

ianswett: I think how you have it laid out is great and an effort to convert it into text will probably make it less clear.  But I'm not familiar enough with markdown to know how to get the formatting right.

nibanks: Ok. I'll give it a go.

draft-ietf-quic-recovery.md
```diff
@@ -645,14 +645,44 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
+received at t=8.  The congestion period is calculated as the time between the
+oldest and newest lost packets: (3 - 0) = 3.  The duration for persistent
+congestion is equal to: (1 * ((2 ^ 2) - 1)) = 3.  Because the threshold was
+reached, the network is considered to have experienced persistent congestion.
+
+When persistent congestion is established, the sender's congestion window MUST
+be reduced to the minimum congestion window (kMinimumWindow).  This response of
+collapsing the congestion window on persistent congestion is functionally
+similar to a sender's response on a Retransmission Timeout (RTO) in TCP
+{{RFC5681}}.
```
```suggestion
{{RFC5681}} after Tail Loss Probes (TLP) {{TLP}}.
```

draft-ietf-quic-recovery.md
```diff
@@ -1282,15 +1309,28 @@ Invoked by loss detection from DetectLostPackets when new packets
 are detected lost.
 
 ~~~
+   InPersistentCongestion(congestion_period):
+     pto = smoothed_rtt + 4 * rttvar + max_ack_delay
+     return
+       congestion_period >
+         pto * (2 ^ kPersistentCongestionThreshold - 1)
```
janaiyengar: editorial nit: this looks a bit confusing because 'return' is on a line by itself. How about:

return congestion_period >
                    pto * (2 ^ kPersistentCongestionThreshold - 1)

draft-ietf-quic-recovery.md
janaiyengar: Agreed that it seems odd to backoff on max_ack_delay. @ianswett : Let's do that separately.

draft-ietf-quic-recovery.md
janaiyengar: Yes, I agree. I had thought that Now() was best, but that does not account for pauses in sending... newest_lost_time - oldest_lost_time is quite good I think.

janaiyengar:
Looks great! Thanks for doing the work, @nibanks - I like this new form of the algorithm much better. I think it's basically correct, and the text and example are both excellent.

janaiyengar:
@ianswett: want to take a look?

draft-ietf-quic-recovery.md
```diff
@@ -645,14 +645,48 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
+
+  t=0 | Send Pkt #1 (App Data)
+  t=1 | Send Pkt #2 (PTO 1)
+  t=3 | Send Pkt #3 (PTO 2)
+  t=7 | Send Pkt #4 (PTO 3)
+  t=8 | Recv ACK of Pkt #4
+
+The first three packets are determined to be lost when the ACK of packet 4 is
+received at t=8.  The congestion period is calculated as the time between the
+oldest and newest lost packets: (3 - 0) = 3.  The duration for persistent
```
marten-seemann: This doesn't match the description in the first paragraph, does it?
> When an ACK frame is received that establishes loss of all in-flight packets, sent over a long enough period of time, the network is considered to be experiencing persistent congestion.

With this algorithm, we would detect persistent congestion even if packet 2 and 4 were acked, right?

nibanks: The text says `all in-flight packets` meaning a contiguous block of sent packets. The example doesn't call out that check for a contiguous block explicitly, but it would have to be made. So, in that case if packets 2 and 4 were acknowledged you wouldn't have that contiguous block of time that meets the requirements.

draft-ietf-quic-recovery.md
```diff
@@ -645,14 +645,48 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
+congestion is equal to: (1 * ((2 ^ 2) - 1)) = 3.  Because the threshold was
```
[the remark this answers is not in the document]
nibanks: We pretty clearly state the equation for calculating the persistent congestion window above and state that `kPersistentCongestionThreshold` is assumed to be 2 for this example. I don't really want to make this section any more verbose.

draft-ietf-quic-recovery.md
```diff
@@ -645,14 +645,48 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
-window MUST be reduced to the minimum congestion window (kMinimumWindow).  This
-response of collapsing the congestion window on persistent congestion is
-functionally similar to a sender's response on a Retransmission Timeout (RTO) in
-TCP {{RFC5681}}.
+sent over a long enough period of time, the network is considered to be
+experiencing persistent congestion.  Commonly, this can be established by
+consecutive PTOs, but since the PTO timer is reset when a new ack-eliciting
+packet is sent, an explicit duration must be used to account for those cases
+where PTOs do not occur or are substantially delayed.  This duration is the
+equivalent of kPersistentCongestionThreshold consecutive PTOS, and is computed
```
```suggestion
equivalent of kPersistentCongestionThreshold consecutive PTOs, and is computed
```

draft-ietf-quic-recovery.md
ianswett: I think this example works, but the pseudocode below is not quite right for this reason.  The easiest way to fix it would be to assume no packet number gaps and check if the number of packets lost is equal to largest - smallest + 1.

nibanks: Do we need to really spell this out with a 100% complete pseudocode example? I can imagine a couple ways for implementations to achieve this and I don't think we need to call them all out.

ianswett: The text is normative and the pseudocode is an appendix.  I'm ok with the pseudocode not being perfect, especially for cases that are fairly unlikely in practice.

marten-seemann: The point of the pseudo-code is not to tell implementations exactly what to do. You can still implement it any way you want.
I'd like to avoid inconsistencies between the text and the pseudo code. We had these before when we introduced packet number spaces, and it now turns out to be quite a lot of work to get things straight.

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
marten-seemann: Just saying, this was my experience when reading this paragraph for the first time. Would be interesting to hear what other people reading this text for the first time think.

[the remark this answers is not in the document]
nibanks: Do you have a recommendation of some specific text to add or change that would make things more clear?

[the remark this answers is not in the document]
marten-seemann: `1 * ((2 ^ kPersistentCongestionThreshold) - 1)) = 3`

draft-ietf-quic-recovery.md
```diff
@@ -645,14 +645,49 @@ packets might cause the sender's bytes in flight to exceed the congestion window
...
+  t=1 | Send Pkt #2 (PTO 1)
+  t=3 | Send Pkt #3 (PTO 2)
+  t=7 | Send Pkt #4 (PTO 3)
+  t=8 | Recv ACK of Pkt #4
+
+The first three packets are determined to be lost when the ACK of packet 4 is
+received at t=8.  The congestion period is calculated as the time between the
+oldest and newest lost packets: (3 - 0) = 3.  The duration for persistent
+congestion is equal to: (1 * ((2 ^ kPersistentCongestionThreshold) - 1)) = 3.
+Because the threshold was reached, the network is considered to have experienced
```
janaiyengar: Suggestion: "Because the threshold was reached and because none of the packets between the oldest and the newest packets are acknowledged, the network is considered ..."

draft-ietf-quic-recovery.md
```diff
@@ -1282,15 +1314,27 @@ Invoked by loss detection from DetectLostPackets when new packets
 are detected lost.
 
 ~~~
+   InPersistentCongestion(congestion_period):
```
janaiyengar: Suggested rework of this function:
```
InPersistentCongestion(newest_lost_packet):
    pto = smoothed_rtt + 4 * rttvar + max_ack_delay
    congestion_period = pto * (2 ^ kPersistentCongestionThreshold - 1)
    // Find the latest packet that was sent at least pto time before the newest lost packet
    oldest_sent_packet = FindPacketSentBefore(newest_lost_packet, congestion_period)
    // Determine if all packets in the window, including the  edges, are marked lost
    return IsWindowLost(oldest_sent_packet, newest_lost_packet)
```

draft-ietf-quic-recovery.md
```diff
@@ -1282,15 +1314,27 @@ Invoked by loss detection from DetectLostPackets when new packets
...
+     oldest_lost_packet = lost_packets.first()
+     newest_lost_packet = lost_packets.last()
 
      // Start a new congestion epoch if the last lost packet
      // is past the end of the previous recovery epoch.
-     CongestionEvent(largest_lost_packet.time_sent)
+     CongestionEvent(newest_lost_packet.time_sent)
+
+     // Collapse congestion window if persistent congestion
+     if (InPersistentCongestion(
```
janaiyengar: Change params to InPersistentCongestion to only have newest_lost_packet.

draft-ietf-quic-recovery.md
nibanks: I feel like this overly complicates things (at least for the pseudocode). What if simplify it a bit, like the following:
```
InPersistentCongestion(newest_lost_packet):
    pto = smoothed_rtt + 4 * rttvar + max_ack_delay
    congestion_period = pto * (2 ^ kPersistentCongestionThreshold - 1)
    // Determine if all packets in the window before the newest lost packet,
    // including the edges, are marked lost
    return IsWindowLost(newest_lost_packet, congestion_period)
```

janaiyengar:
@ianswett: Can you take a look? I'd like to merge this one.

draft-ietf-quic-recovery.md
```diff
@@ -1282,15 +1315,28 @@ Invoked by loss detection from DetectLostPackets when new packets
...
+     // newest lost packet, including the edges, are marked
+     // lost
+     return IsWindowLost(newest_lost_packet, congestion_period)
+
    OnPacketsLost(lost_packets):
      // Remove lost packets from bytes_in_flight.
      for (lost_packet : lost_packets):
        bytes_in_flight -= lost_packet.size
-     largest_lost_packet = lost_packets.last()
+     newest_lost_packet = lost_packets.last()
```
ianswett: Elsewhere we use largest(ie: largest_acked) and there are currently no uses of newest in the recovery draft, so I'm going to change this back to largest.  If we want to start using newest, let's do that in a separate PR.

nibanks: Fine with me. I might look at changing it later.

PR #2370, opened 2019-01-27
Stronger migration handshake

### LINKED ISSUE #2309: Migration before handshake completed is very messy
In migration interop tests, the Picoquic client was starting to simulate NAT rebinding just after obtaining the 1-RTT read and write key. That turns out to be very messy, because at that point there is no proof that the server has received the client finished message, let alone acknowledged it. As far as tests are concerned, there is an easy fix: only start the migration tests after the client has received from the server an ACK of one of its 1-RTT packet. That's the moment when the client knows for sure that the handshake is successful.

The migration before handshake complete is messy because the client will try to re-transmit handshake messages that have not been acknowledged, and has some kind of dilemma. Should these re-transmissions aims at the original DCID used during the handshake, or at one of the new CID selected during the migration? What if the continuity of the new path has not been yet verified? What if the continuity test was lost?

Do we have an actual need to support migration before the handshake is fully completed? In practice, the handshake packets create a NAT mapping, and it is rather unlikely that this mapping will just disappear before the few RTT required to complete the handshake. In fact, it is possible, maybe likely, that packets received from an alternative address during the handshake comes from attackers, not from the actual client.

I think the implementations would be simpler and more robust if the spec stated that clients should not initiate "voluntary" migration before the handshake is fully complete. This means 1-RTT ACK received, or ACK of last handshake packet from the client, whichever comes first. Similarly, servers should not try to migrate connections before receiving at least one 1-RTT packet from the client through the same addresses and ports as the Initial messages.

huitema:
State requirement that clients must wait until the server considers the handshake finished before starting migration.

Fixes #2309


draft-ietf-quic-transport.md
```diff
@@ -1847,6 +1847,11 @@ endpoint migrates to a new address.
 An endpoint MUST NOT initiate connection migration before the handshake is
 finished and the endpoint has 1-RTT keys.  The design of QUIC relies on
 endpoints retaining a stable address for the duration of the handshake.
+Clients MUST NOT initiate connection migration before they are
+certain that their peer also considers the handshake finished. This
+means that in addition to waiting for availability of 1-RTT keys,
+clients MUST wait acknowledgement by the server of one of their
+1-RTT messages before initiating connection migration.
```
mikkelfj: Suggest singular form: A client...

> A client MUST NOT initiate connection migration before it is certain
that it's peer also considers the handshake finished. This means that
in addition to waiting for availability of 1-RTT keys, a client MUST
also await acknowledgement of at least one of it's 1-RTT transmitted
packets before initiating connection migration.

draft-ietf-quic-transport.md
```diff
@@ -1847,6 +1847,11 @@ endpoint migrates to a new address.
...
 
 An endpoint also MUST NOT initiate connection migration if the peer sent the
 `disable_migration` transport parameter during the handshake.  An endpoint which
```
mikkelfj: Not in this PR, but why do we use both client, server, peer, and endpoint, when migration can only happen from the client in V1 (or did I miss something?)

erickinnear: Some of that's just a mix of different text coming in at different times, I think. 
At one point, the idea was to deliberately keep the text agnostic (for the most part, any text restricting who can initiate is clearly going to be one-sided) to endpoint, since that allowed reasoning about it as initiator and responder rather than client vs. server.

draft-ietf-quic-transport.md
```diff
@@ -1847,6 +1847,11 @@ endpoint migrates to a new address.
 An endpoint MUST NOT initiate connection migration before the handshake is
 finished and the endpoint has 1-RTT keys.  The design of QUIC relies on
 endpoints retaining a stable address for the duration of the handshake.
+Clients MUST NOT initiate connection migration before they are
```
erickinnear: Just a nit, but this appears to put stricter requirements on the statements made in the previous two sentences, is there any reason not to merge them together?

draft-ietf-quic-transport.md
huitema: The part about 1-rtt ack is actually client specific. I think I was influenced by the picotls implementation that only provides the 1-rtt read key after CFIN has been received, and did not spell out the parallel server requirement that some 1rtt packet should have been received. Will do.

mikkelfj: @huitema  > Where did we specify that only the client initiates migrations in V1?

I'm not sure, but from memory it was dropped at some point because symmetric ping pong got hairy during updates or some defence became difficult.

Maybe it didn't land? Found this:

> Caution:
If both endpoints change connection ID in response to seeing a change in connection ID from their peer, then this can trigger an infinite sequence of changes.

But there is this, which seems to be preferred address only - maybe I got it mixed up_

9.6, transport:
> Migrating a connection to a new server address mid-connection is left for future work.

draft-ietf-quic-transport.md
huitema: I think the updated version solves your issue.

draft-ietf-quic-transport.md
```diff
@@ -1844,9 +1844,15 @@ addresses (that is, IP address and/or port), such as those caused by an endpoint
...
-finished and the endpoint has 1-RTT keys.  The design of QUIC relies on
+The design of QUIC relies on
 endpoints retaining a stable address for the duration of the handshake.
+An endpoint MUST NOT initiate connection migration before the handshake is
+finished for it and its peer and the endpoint has 1-RTT keys. This means
+that the server MUST NOT initiate connection migration before it has
+received at least one 1-RTT data from the client, and the client 
+MUST NOT initiate connection migration before receiving 
+acknowledgement by the server of at least one of its
+1-RTT messages.
```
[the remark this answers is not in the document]
huitema: I really don't like the CRYPTO frames reference. There is no FIN mark on the crypto stream, and the QUIC engine proper does not know when "all crypto frames" have been received. I was careful to use signals that the transport stack understand: obtaining keys, receiving packets, etc.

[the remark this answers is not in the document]
ianswett: I'm happy with this approach, but I'd replace "before receiving acknowledgement by the server of at least one of its 1-RTT messages" with "before receiving a 1RTT ACK frame." for simplicity.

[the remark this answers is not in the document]
kazuho: @huitema 
> I really don't like the CRYPTO frames reference. There is no FIN mark on the crypto stream, and the QUIC engine proper does not know when "all crypto frames" have been received.  I was careful to use signals that the transport stack understand: obtaining keys, receiving packets, etc.

I am not sure if I agree with the observation. The requirement (i.e. "has both received and acknowledged all CRYPTO frames for that encryption level and when all CRYPTO frames for that encryption level have been acknowledged by its peer") can be implement as checking that 1-RTT read key is available and that the send buffer of CRYPTO frames in Handshake epoch is empty.

Anyways, I think using different events for starting the handshake key drop timer and permitting migration is an unnecessary complexity.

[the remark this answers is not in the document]
huitema: @kazuho _"Anyways, I think using different events for starting the handshake key drop timer and permitting migration is an unnecessary complexity."_ I agree. I think there should be a "when is handshake complete" section specifying that, and we can reference it in places like dropping key, starting migration, or starting key rotations.

martinthomson:
If people want to talk about this this week, then this needs an issue.

huitema:
@martinthomson that would be issue #2309

draft-ietf-quic-transport.md
```diff
@@ -1532,6 +1532,24 @@ be unambiguously different to ensure no confusion about their interpretation.
 One way that a new format could be introduced is to define a TLS extension with
 a different codepoint.
 
+### Handshake Completion {#handshake-completion}
+
+The connection establishment handshake concludes when both parties have
```
janaiyengar: I don't agree with some of this text, but importantly, it's more than is needed. Suggested rephrase of this paragraph, and combine this with the next paragraph: "Connection handshake is complete when version negotiation is complete and 1-RTT keys are available at both endpoints."

draft-ietf-quic-transport.md
```diff
@@ -1532,6 +1532,24 @@ be unambiguously different to ensure no confusion about their interpretation.
...
+have obtained insurance that all necessary Initial and Handshake messages
+have been properly received by their peers, as defined in {{QUIC-TLS}}.
+
+For the server, this is achieved when the TLS stack informs the
+transport that the session establishment is complete, and that the
+1-RTT receive keys can now be used, as indicated in section 4.1.3 of
+{{QUIC-TLS}}.
+
+For the client, this is achieved after the 1-RTT keys become available 
+when the server acknowledges either the client's Handshake message carrying
```
janaiyengar: Replace "the server acknowledges ... by the client" with "either all Handshake packets or a 1-RTT packet sent by the client are acknowledged"

draft-ietf-quic-transport.md
```diff
@@ -1532,6 +1532,24 @@ be unambiguously different to ensure no confusion about their interpretation.
...
+the last element of the Crypto Stream or an 1-RTT message sent by the client.
+Waiting for either of these events allows the client to verify the
+completion of the handshake even in cases where the Handshake
+acknowledgement is lost.
```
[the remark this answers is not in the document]
huitema: Jana, I temporarily removed the content of that section, since MT has taken the action to write the "handshake is complete" text, introducing the new "previous epoch done" frame, etc.

draft-ietf-quic-transport.md
huitema: Removed that text, waiting for MT.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
MikeBishop: @ianswett, a 1-RTT ACK frame might acknowledge only 0-RTT packets.  You need confirmation that the server has received and parsed at least one 1-RTT packet, no?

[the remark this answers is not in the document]
ianswett: Yes, it would have to be that a 1RTT packet was acknowledged, as you point out.

DavidSchinazi:
I like the latest text referencing handshake confirmed 👍

huitema:
@DavidSchinazi I was waiting for #2673...

DavidSchinazi:
@huitema I know :-)

draft-ietf-quic-transport.md
```diff
@@ -1794,9 +1793,10 @@ addresses (IP address and port), such as those caused by an
 endpoint migrating to a new network.  This section describes the process by
 which an endpoint migrates to a new address.
 
-An endpoint MUST NOT initiate connection migration before the handshake is
-finished and the endpoint has 1-RTT keys.  The design of QUIC relies on
+The design of QUIC relies on
```
ianswett: nit: Please reflow this paragraph so this is a full line.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
huitema: Solved in rewrite with dependency on "handshake confirmed".

PR #2373, opened 2019-01-28
STOP_SENDING to cancel a push

ianswett:
The client can't send RESET_STREAM on a push stream ID.


afrind:
I read this paragraph to apply only to how the server cancels a push.  Once the push stream is created, it would use a RESET_STREAM?

ianswett:
Good point, I misread the existing text.  I'd like this section to be a bit clearer, but I don't have any concrete suggestions at the moment.

PR #2379, opened 2019-01-28
Replace the RequireInsertCount decoding algorithm

### LINKED ISSUE #2112: Largest Reference algorithm can produce invalid values
Feed the algorithm a LargestReference of 1, TotalNumberOfInserts of 2, and MaxEntries of 10.

```python
   if LargestReference > 0:
      LargestReference -= 1  # == 0
      CurrentWrapped = TotalNumberOfInserts mod (2 * MaxEntries) # == 2

      if CurrentWrapped >= LargestReference + MaxEntries:  # == False
         # Largest Reference wrapped around 1 extra time
         LargestReference += 2 * MaxEntries
      else if CurrentWrapped + MaxEntries < LargestReference:   # == False
         # Decoder wrapped around 1 extra time

      LargestReference += TotalNumberOfInserts - CurrentWrapped # == no change
```

LargestReference is therefore 0, and invalid.

Do the same for LargestReference of 18 and you get something even better: -2.

I think that the underflow check in #1904 fixes this.

afrind:
This is @mt's alternate algorithm from #1930, and includes some of @bencebeky's language from #2249.

I renamed 'rounded' to 'MaxWrapped' - it is largest possible value for RequiredInsertCount that is 0 mod 2*MaxEntries.

Fixes #1930
Fixes #2112

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -861,21 +861,23 @@ algorithm, where TotalNumberOfInserts is the total number of inserts into the
 decoder's dynamic table:
 
 ~~~
+   FullRange = 2 * MaxEntries
    if EncodedInsertCount == 0:
       ReqInsertCount = 0
    else:
-      InsertCount = EncodedInsertCount - 1
-      CurrentWrapped = TotalNumberOfInserts mod (2 * MaxEntries)
-
-      if CurrentWrapped >= InsertCount + MaxEntries:
-         # Insert Count wrapped around 1 extra time
-         ReqInsertCount += 2 * MaxEntries
-      else if CurrentWrapped + MaxEntries < InsertCount:
-         # Decoder wrapped around 1 extra time
-         CurrentWrapped += 2 * MaxEntries
-
-      ReqInsertCount += TotalNumberOfInserts - CurrentWrapped
+      MaxValue = TotalNumberOfInserts + MaxEntries
+      # MaxWrapped is the largest possible value of
+      # ReqInsertCount that is 0 mod 2*MaxEntries
+      MaxWrapped = floor(MaxValue / FullRange) * FullRange
+      ReqInsertCount = MaxWrapped + EncodedInsertCount - 1
+      if ReqInsertCount > MaxValue:
+         # The Encoder's value wrapped one fewer time
+         if ReqInsertCount < FullRange:
+            Error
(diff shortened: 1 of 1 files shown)


afrind:
This algorithm doesn't explicitly validate that the wire value is <= 2 * MaxEntries, and if it receives a larger value, may produce a value in the future.  

I can add an explicit error check to the algorithm, or add normative text that this is an error, or both.  I'm inclined to add it to the code?

bencebeky:
I agree that an error should be generated in that case and I support adding the check to the code.  Since "the decoder reconstructs the Required Insert Count using the following algorithm", and not "this is an example algorithm to reconstruct the Required Insert Count", I consider the code to be normative, so I do not think it is necessary to spell this out in text.

BTW I'm not in love with this algorithm, mostly because it is longer and more complicated than the original without improving on clarity, but I can live with it.  I still think the one I proposed at #2248 would be the least error-prone (most implementers presumably don't care about how the algorithm works, they just copy it line by line, so the fewer lines, the better), but whatever is in the draft currently, or even the one proposed here, is just fine with me after all.

bencebeky:
> Ah, I didn't really intend that an implementation HAD to use this algorithm. An implementation can use any algorithm that produces correct results (or not decode at all, per @dtikhonov's impl that uses the wrapped value to index directly). Given that it's easy to screw up the algorithm, I like having the psuedocode for copy-and-paste/no brainer.
> 
> Therefore, let me make _that_ clear and add normative language for all error cases, but I may also add the check to the algorithm.

I understand.  In this case, I support adding normative language for checks.

Honestly, I interpret "the decoder reconstructs the Required Insert Count using the following algorithm" as indicating that the implementation must use an algorithm that has the exact same behavior as the pseudo-code.  Therefore I propose that the error check is added to the pseudo-code as well.

Maybe the text could be changed to something like "The decoder reconstructs the Required Insert Count.  If the decoder encounters an EncodedInsertCount value that could not have been produced by a conformant encoder, then the decoder MUST signal a stream error.  An example algorithm to decode the Required Insert Count is the following.".  This would indicate to me that the pseudo-code is non-normative.  But I am also just fine with the current text, as long as the requirements (on error checking and everything else) are spelled out both in the text (because you interpret that as normative) and in pseudo-code (because I interpret that as normative).  Having consistent text + pseudo-code sounds like a good idea anyway.

Issue #2145, opened 2018-12-13
Don't change CID on peer CID change


ekr:
```Endpoints that use connection IDs with length greater than zero could have their
activity correlated if their peers keep using the same destination connection ID
after migration. Endpoints that receive packets with a previously unused
Destination Connection ID SHOULD change to sending packets with a connection ID
that has not been used on any other network path.  The goal here is to ensure
that packets sent on different paths cannot be correlated. To fulfill this
privacy requirement, endpoints that initiate migration and use connection IDs
with length greater than zero SHOULD provide their peers with new connection IDs
before migration.

Caution:

: If both endpoints change connection ID in response to seeing a change in
  connection ID from their peer, then this can trigger an infinite sequence of
  changes.
```

I don't remember this being what we agreed, and it's not necessary. You only need to change when you see a new path, not after a new CID. That's how we got rid of counting to infinity.

mikkelfj:
I'm not sure it is always well-defined when you are on a new path and you risk accidental linkage. I believe this was the thinking behind this.

MikeBishop:
IIRC, @ekr is correct; we agreed to do this on peer address change.

- If you don't know you've been rebound and you reuse a CID, you've already linked yourself; the peer changing CID in response is pointless.
- If you either anticipate a rebinding or proactively change ports, and change both CID and address simultaneously, then it doesn't matter which of those triggers the peer to switch.
- If you switch CID without changing address, the address links you with fairly high probability anyway, so it doesn't really matter if the peer switches.

martinthomson:
The recommendation exists to avoid linkability across a non-migrating change in connection ID.

ekr:
I don't think that that's particularly useful, and it seems to create a bunch of new problems.

janaiyengar:
My recollection is what @MikeBishop says.  I remember discussing that changing CID on both sides allows a flow to be broken up into flowlets, but if the address is the same on both sides, that's a pretty clear signal anyways. I am fine with requiring a change only if the peer's CID and address change.

mikkelfj:
>  but if the address is the same on both sides, that's a pretty clear signal anyways. I am fine with requiring a change only if the peer's CID and address change.

That is not sufficient because the client could be migrating from domestic wifi to mobile, to train wifi, while the server always sees the IP address of the load balancer.

ekr:
In this case the LB and the server should be regarded as the same thing,
and it's the LB's job to convey that somehow.

[...]

PR #2386, opened 2019-01-30
Remove requirement for reciprocal connection ID change

### LINKED ISSUE #1795: Define a safe algorithm for changing CID in response to a change in CID
Ideally, we want an endpoint to change connection ID if its peer changes connection ID.  However, this leads to infinite changing of connection IDs if you implement the naive version of this.  We need an algorithm that is simple enough to implement, but immune to the looping problem.

This isn't critical (we can live without it), but it would be a nice to have.

Currently Section 6.11.5 is where this would live.

martinthomson:
As discussed in Tokyo, forcing your peer to change a connection ID is still possible, but we would require a change in port in order to effect that.

Closes #1795, #2145.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2110,19 +2110,10 @@ infrequently.
 
 Endpoints that use connection IDs with length greater than zero could have their
 activity correlated if their peers keep using the same destination connection ID
-after migration. Endpoints that receive packets with a previously unused
-Destination Connection ID SHOULD change to sending packets with a connection ID
-that has not been used on any other network path.  The goal here is to ensure
-that packets sent on different paths cannot be correlated. To fulfill this
-privacy requirement, endpoints that initiate migration and use connection IDs
-with length greater than zero SHOULD provide their peers with new connection IDs
-before migration.
-
-Caution:
-
-: If both endpoints change connection ID in response to seeing a change in
-  connection ID from their peer, then this can trigger an infinite sequence of
-  changes.
+after migration.  The goal here is to ensure that packets sent on different
+paths cannot be correlated. To fulfill this privacy requirement, endpoints that
+initiate migration and use connection IDs with length greater than zero SHOULD
+provide their peers with new connection IDs before migration.
 
 
 ## Server's Preferred Address {#preferred-address}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2110,7 +2110,7 @@ infrequently.
 
 Endpoints that use connection IDs with length greater than zero could have their
 activity correlated if their peers keep using the same destination connection ID
-after migration.  The goal here is to ensure that packets sent on different
+after migration.  The goal is to ensure that packets sent on different
 paths cannot be correlated. To fulfill this privacy requirement, endpoints that
 initiate migration and use connection IDs with length greater than zero SHOULD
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2110,19 +2110,10 @@ infrequently.
...
-privacy requirement, endpoints that initiate migration and use connection IDs
-with length greater than zero SHOULD provide their peers with new connection IDs
-before migration.
-
-Caution:
-
-: If both endpoints change connection ID in response to seeing a change in
-  connection ID from their peer, then this can trigger an infinite sequence of
-  changes.
+after migration.  The goal here is to ensure that packets sent on different
```
```suggestion
after migration.  The goal is to ensure that packets sent on different
```

draft-ietf-quic-transport.md
```diff
@@ -2110,19 +2110,10 @@ infrequently.
...
-privacy requirement, endpoints that initiate migration and use connection IDs
-with length greater than zero SHOULD provide their peers with new connection IDs
-before migration.
-
-Caution:
-
-: If both endpoints change connection ID in response to seeing a change in
-  connection ID from their peer, then this can trigger an infinite sequence of
-  changes.
+after migration.  The goal is to ensure that packets sent on different
```
janaiyengar: This is a bit awkward. Suggested rewrite:
"after migration. To ensure that packets sent on different paths cannot be correlated, such endpoints SHOULD provide their peers with new connection IDs before migration."

draft-ietf-quic-transport.md
```diff
@@ -2110,19 +2110,10 @@ infrequently.
 
 Endpoints that use connection IDs with length greater than zero could have their
```
janaiyengar: This is not part of your PR, but suggested change: "Endpoints that supply non-zero-length connection IDs ..."

draft-ietf-quic-transport.md
ianswett: +1 to Jana's suggestion.

draft-ietf-quic-transport.md
martinthomson: In reading this, I realized that we are assuming that this assumes that keeping the connection ID is a permissible option.  It isn't.  I'll file a follow-up.

draft-ietf-quic-transport.md
```diff
@@ -2108,21 +2108,13 @@ genuine migrations.  Changing port number can cause a peer to reset its
...
-
-Caution:
-
-: If both endpoints change connection ID in response to seeing a change in
-  connection ID from their peer, then this can trigger an infinite sequence of
-  changes.
+Endpoints that supply connection IDs with length greater than zero could have
+their activity correlated if their peers keep using the same destination
+connection ID after migration.  To ensure that packets sent on different paths
+cannot be correlated, endpoints SHOULD provide with new connection IDs before
```
```suggestion
cannot be correlated, endpoints SHOULD supply new connection IDs to their peers before the peers migrate.
```

And drop the rest of this paragraph, I don't think it adds anything.

martinthomson: #2414 might be the better PR to review then.  I've removed the duplicated text.

janaiyengar: I like the text there better, I've left a comment as well. Do you want to close this PR off in favor of that one?

martinthomson: This should close when that does.  Closing this might have side-effects.

PR #2390, opened 2019-01-30
removing stray )

seanturner:


seanturner:
Needed to fetch upstream first.

PR #2399, opened 2019-02-01
Ipv6 flow label

huitema:
Recommendation to set the IPv6 flow label in accordance to RFC 6437, as agreed in Tokyo.


draft-ietf-quic-transport.md
```diff
@@ -2203,6 +2203,21 @@ address before path validation is complete.
 A client that migrates to a new address SHOULD use a preferred address from the
 same address family for the server.
 
+## Use of IPv6 Flow-Label and migration {ipv6-flow-label}
```
```suggestion
## Use of IPv6 Flow-Label and Migration {#ipv6-flow-label}
```

draft-ietf-quic-transport.md
```diff
@@ -2203,6 +2203,21 @@ address before path validation is complete.
...
+
+Quic nodes that send data using IPv6 SHOULD document the IPv6 flow label
```
```suggestion
Endpoints that send data using IPv6 SHOULD apply an IPv6 flow label
```

draft-ietf-quic-transport.md
```diff
@@ -2203,6 +2203,21 @@ address before path validation is complete.
...
+in compliance with {{!RFC6437}}, unless the local API does not allow
+setting this parameter.
+
+The IPv6 flow label should be a pseudo-random function of the source
```
```suggestion
The IPv6 flow label SHOULD be a pseudo-random function of the source
```

draft-ietf-quic-transport.md
```diff
@@ -2203,6 +2203,21 @@ address before path validation is complete.
...
+and destination addresses, source and destination UDP ports, and the destination
+CID. The flow label generation MUST be designed to minimize the chances of
+collision with a previously used flow label, as this would enable linkability
```
```suggestion
linkability with a previously used flow label, as this would enable correlating activity on multiple paths
```

draft-ietf-quic-transport.md
```diff
@@ -2203,6 +2203,21 @@ address before path validation is complete.
...
+of multiple paths (see {{migration-linkability}}).
+
+A possible implementation is to compute the flow label as a one way
+hash function of the source and destination addresses, source and
+destination UDP ports, destination CID and a local secret.
```
```suggestion
destination UDP ports, destination CID, and a local secret.
```

draft-ietf-quic-transport.md
```diff
@@ -2203,6 +2203,21 @@ address before path validation is complete.
...
+in compliance with {{!RFC6437}}, unless the local API does not allow
+setting this parameter.
+
+The IPv6 flow label SHOULD be a pseudo-random function of the source
+and destination addresses, source and destination UDP ports, and the destination
+CID. The flow label generation MUST be designed to minimize the chances of
+linkability with a previously used flow label, as this would enable correlating
+activity on multiple paths (see {{migration-linkability}}).
+
+A possible implementation is to compute the flow label as a one way
```
```suggestion
A possible implementation is to compute the flow label as a one-way
```

draft-ietf-quic-transport.md
```diff
@@ -2203,6 +2203,21 @@ address before path validation is complete.
 A client that migrates to a new address SHOULD use a preferred address from the
 same address family for the server.
 
+## Use of IPv6 Flow-Label and Migration {#ipv6-flow-label}
+
+Endpoints that send data using IPv6 SHOULD apply an IPv6 flow label
```
larseggert:
```suggestion
Endpoints that send data using IPv6 SHOULD set an IPv6 flow label for a QUIC connection
```

draft-ietf-quic-transport.md
```diff
@@ -2203,6 +2203,21 @@ address before path validation is complete.
...
+in compliance with {{!RFC6437}}, unless the local API does not allow
+setting this parameter.
```
larseggert:
```suggestion
setting IPv6 flow labels.
```

draft-ietf-quic-transport.md
huitema: It is actually per path, not per connection, and is explained in the next paragraph.
@martinthomson suggested replacing "set" by "apply" in a previous comment.

huitema:
OK. I specified a "cryptographic hash"

becarpenter:
fwiw [RFC6437](https://tools.ietf.org/html/rfc6437) is (intentionally) non-specific about the hash function. It says "If a hash algorithm is used, as suggested in Section 3, it SHOULD include a step that makes the flow label value significantly difficult to predict [RFC4086], even with knowledge of the algorithm being used." We didn't see any real need to specify a crypto hash for this purpose, but it's certainly OK. (The example in Appendix A of 6437 stinks and I regret it.)

PR #2406, opened 2019-02-04
Close connection if peer becomes unreachable

### LINKED ISSUE #1638: Dead path timeout
#1614 makes idle timeout more precise, but still leaves open the question of when a path is considered dead.  We still need a path timeout to close paths over which an endpoint cannot reach a peer.

My suggestion here is to have N RTOs be a value that's used for detecting a dead path, for some reasonable value of N (5? 7?). An endpoint can close the connection when a path is considered dead or can wait for a new path to become available (this is extremely useful when migrating from wifi to cell).

janaiyengar:
Closes #1638.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1765,10 +1765,9 @@ NAT rebinding is not connection migration as defined in this section, though an
 endpoint SHOULD perform path validation ({{migrate-validate}}) if it detects a
 change in the IP address of its peer.
 
-When a client finds the server unreachable over any path, the client might close
-its connection to the server (see {{termination}}).  A client capable of
-connection migration MAY wait for new paths to become available before declaring
-the server unreachable.
+When an endpoint has no validated path on which to send packets, it MAY discard
+connection state.  An endpoint capable of connection migration MAY wait for a
+new path to become available before discarding connection state.
 
 This document limits migration of connections to new client addresses, except as
 described in {{preferred-address}}. Clients are responsible for initiating all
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2207,14 +2212,15 @@ same address family for the server.
...
-period of time.  A QUIC connection, once established, can be terminated in one
-of three ways:
+An established QUIC connection can be terminated in one of three ways:
 
 * idle timeout ({{idle-timeout}})
 * immediate close ({{immediate-close}})
 * stateless reset ({{stateless-reset}})
 
+In addition, an endpoint MAY drop a connection and discard all connection state
+without informing the peer if it is unable to reach the peer.
```
ianswett: How about "repeatedly unable to reach the peer."  Also, in this case, a connection close packet should still be sent in case the peer can receive packets, but packets it sends cannot reach you.

Also is 'reach the peer' obvious wording?

mikkelfj: I was wondering about this too, but assumed there was a clear definition in an earlier section. Even if you do not get a response, the peer might be able to hear a close. In that context, MAY seems fine.

draft-ietf-quic-transport.md
```diff
@@ -1862,6 +1862,11 @@ NAT rebinding is not connection migration as defined in this section, though an
 endpoint SHOULD perform path validation ({{migrate-validate}}) if it detects a
 change in the IP address of its peer.
 
+When a client finds the server unreachable over any path, the client might close
```
martinthomson: You need to say what "close" means, I think.

draft-ietf-quic-transport.md
martinthomson: Maybe instead:

> An endpoint MAY discard connection state if it does not have a validated path on which it can send packets.

martinthomson:
@janaiyengar, this is annoyingly close to done.  Would you mind?

janaiyengar:
@martinthomson : Take a look.

PR #2414, opened 2019-02-04
New connection IDs are mandatory for intentional migration 

### LINKED ISSUE #2413: Mandate using a new connection ID on migration
In fixing #2145, I discovered that we don't actually mandate this.  It's a privacy requirement that I thought we had previously agreed to, but it appears as though we missed capturing it properly.

martinthomson:
Based on #2386, which should merge soon.

Closes #2413.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2089,15 +2089,17 @@ different local addresses, as discussed in {{connection-id}}.  For this to be
 effective endpoints need to ensure that connections IDs they provide cannot be
 linked by any other entity.
 
-This eliminates the use of the connection ID for linking activity from
-the same connection on different networks.  Header protection ensures
-that packet numbers cannot be used to correlate activity.  This does not prevent
-other properties of packets, such as timing and size, from being used to
-correlate activity.
+An endpoint MUST use a new connection ID if it initiates connection migration,
+unless the peer has selected a zero-length connection ID.  This eliminates the
+use of the connection ID for linking activity from the same connection on
+different networks.  Header protection ensures that packet numbers cannot be
+used to correlate activity.  This does not prevent other properties of packets,
+such as timing and size, from being used to correlate activity.
 
 Clients MAY move to a new connection ID at any time based on
 implementation-specific concerns.  For example, after a period of network
-inactivity NAT rebinding might occur when the client begins sending data again.
+inactivity, where NAT rebinding might occur when the client begins sending data
+again.
 
 A client might wish to reduce linkability by employing a new connection ID and
 source UDP port when sending traffic after a period of inactivity.  Changing the
@@ -2108,13 +2110,12 @@ genuine migrations.  Changing port number can cause a peer to reset its
 congestion state (see {{migration-cc}}), so the port SHOULD only be changed
 infrequently.
 
-Endpoints that supply connection IDs with length greater than zero could have
-their activity correlated if their peers keep using the same destination
-connection ID after migration.  To ensure that packets sent on different paths
-cannot be correlated, endpoints SHOULD provide with new connection IDs before
-peers migrate.  To fulfill this privacy requirement, endpoints that initiate
-migration and use connection IDs with length greater than zero SHOULD provide
-their peers with new connection IDs before migration.
+An endpoint cannot migrate to a new path if it does not have a connection ID to
+use on that path.  An endpoint that exhausts available connection IDs can only
+migrate if the peer chooses to use receive packets with zero-length connection
+IDs.  To ensure that migration is possible and packets sent on different paths
+cannot be correlated, endpoints SHOULD provide new connection IDs before peers
+migrate.
 
 
 ## Server's Preferred Address {#preferred-address}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2091,7 +2091,8 @@ different local addresses, as discussed in {{connection-id}}.  For this to be
 effective endpoints need to ensure that connections IDs they provide cannot be
 linked by any other entity.
 
-Endpoints MAY move to a new connection ID at any time.
+At any time, endpoints MAY change the Destination Connection ID they send to a
+value that has not been used on another path.
 
 An endpoint MUST use a new connection ID if it initiates connection migration.
 Using a new connection ID eliminates the use of the connection ID for linking
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2113,9 +2113,10 @@ infrequently.
 An endpoint cannot migrate to a new path if it does not have a connection ID to
 use on that path.  An endpoint that exhausts available connection IDs cannot
 migrate.  If a peer chooses to receive packets with zero-length connection IDs,
-an endpoint can always migrate.  To ensure that migration is possible and
-packets sent on different paths cannot be correlated, endpoints SHOULD provide
-new connection IDs before peers migrate.
+an endpoint can always migrate - zero-length connection IDs provide no
+significant linkability.  To ensure that migration is possible and packets sent
+on different paths cannot be correlated, endpoints SHOULD provide new connection
+IDs before peers migrate.
 
 
 ## Server's Preferred Address {#preferred-address}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1853,9 +1853,9 @@ An endpoint also MUST NOT initiate connection migration if the peer sent the
 `disable_migration` transport parameter during the handshake.  An endpoint which
 has sent this transport parameter, but detects that a peer has nonetheless
 migrated to a different network MAY treat this as a connection error of type
-INVALID_MIGRATION.  Similarly, an endpoint MUST NOT migrate if its peer supplies
-a zero-length connection ID as packets without a Destination Connection ID
-cannot be attributed to a connection based on address tuple.
+INVALID_MIGRATION.  Similarly, an endpoint MUST NOT initiate migration if its
+peer supplies a zero-length connection ID as packets without a Destination
+Connection ID cannot be attributed to a connection based on address tuple.
 
 Not all changes of peer address are intentional migrations. The peer could
 experience NAT rebinding: a change of address due to a middlebox, usually a NAT,
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2089,17 +2089,19 @@ different local addresses, as discussed in {{connection-id}}.  For this to be
 effective endpoints need to ensure that connections IDs they provide cannot be
 linked by any other entity.
 
-An endpoint MUST use a new connection ID if it initiates connection migration,
-unless the peer has selected a zero-length connection ID.  This eliminates the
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2089,15 +2089,17 @@ different local addresses, as discussed in {{connection-id}}.  For this to be
...
+unless the peer has selected a zero-length connection ID.  This eliminates the
+use of the connection ID for linking activity from the same connection on
+different networks.  Header protection ensures that packet numbers cannot be
+used to correlate activity.  This does not prevent other properties of packets,
+such as timing and size, from being used to correlate activity.
 
 Clients MAY move to a new connection ID at any time based on
 implementation-specific concerns.  For example, after a period of network
-inactivity NAT rebinding might occur when the client begins sending data again.
+inactivity, where NAT rebinding might occur when the client begins sending data
```
```suggestion
inactivity, NAT rebinding may cause an unintentional change in path when the client begins sending data
```

draft-ietf-quic-transport.md
```diff
@@ -2089,15 +2089,17 @@ different local addresses, as discussed in {{connection-id}}.  For this to be
 effective endpoints need to ensure that connections IDs they provide cannot be
 linked by any other entity.
 
-This eliminates the use of the connection ID for linking activity from
-the same connection on different networks.  Header protection ensures
-that packet numbers cannot be used to correlate activity.  This does not prevent
-other properties of packets, such as timing and size, from being used to
-correlate activity.
+An endpoint MUST use a new connection ID if it initiates connection migration,
```
janaiyengar: I would move the conditional clause to the beginning, otherwise it's not clear what the "This" in the next sentence refers to.

"Unless the peer has selected a zero-length connection ID, an endpoint MUST use ..."

Also, how about adding something about what the endpoint needs to do if it has nothing from the peer? Something like: "If no new connection IDs are available, the endpoint MUST NOT migrate. The endpoint MAY close the connection, but MUST NOT send any packets on a new path with the same connection ID as that used on an earlier path."  I can't find any text that says as much. I wonder if this is a new non-editorial issue, since we currently do not have normative text about this (maybe I just can't find it yet).

draft-ietf-quic-transport.md
```diff
@@ -2108,21 +2110,12 @@ genuine migrations.  Changing port number can cause a peer to reset its
...
-
-: If both endpoints change connection ID in response to seeing a change in
-  connection ID from their peer, then this can trigger an infinite sequence of
-  changes.
+An endpoint cannot migrate to a new path if it does not have a connection ID to
+use on that path.  An endpoint that exhausts available connection IDs cannot
+migrate.  If a peer chooses to receive packets with zero-length connection IDs,
+an endpoint can always migrate.  To ensure that migration is possible and
+packets sent on different paths cannot be correlated, endpoints SHOULD provide
+new connection IDs before peers migrate.
```
janaiyengar: This text is good, but still doesn't have normatives, and I think we should have MUST NOT.

draft-ietf-quic-transport.md
```diff
@@ -2108,21 +2110,13 @@ genuine migrations.  Changing port number can cause a peer to reset its
...
-before migration.
-
-Caution:
-
-: If both endpoints change connection ID in response to seeing a change in
-  connection ID from their peer, then this can trigger an infinite sequence of
-  changes.
+An endpoint cannot migrate to a new path if it does not have a connection ID to
+use on that path.  An endpoint that exhausts available connection IDs cannot
+migrate.  If a peer chooses to receive packets with zero-length connection IDs,
```
ianswett: I think it makes sense to put this 3rd sentence first, and then say, "Otherwise, an endpoint cannot migrate..."

The current text reads in a somewhat self-contradictory way.

mikkelfj:
I think this strikes a good balance. It isn't practical to forbid migration without CID when NAT's do it behind your back, but allowing it for intentional migration would work against networks attempting to move towards safer modes of operation in the future, some which might only make sense if QUIC does indeed require new CID's on migration.

At some point with IPv6, NAT rebinding might be seen the same way as older insecure WLAN standards.

draft-ietf-quic-transport.md
```diff
@@ -1853,7 +1853,9 @@ An endpoint also MUST NOT initiate connection migration if the peer sent the
 `disable_migration` transport parameter during the handshake.  An endpoint which
 has sent this transport parameter, but detects that a peer has nonetheless
 migrated to a different network MAY treat this as a connection error of type
-INVALID_MIGRATION.
+INVALID_MIGRATION.  Similarly, an endpoint MUST NOT migrate if its peer supplies
```
ianswett: As I said on the list, I'm not a big fan of adding this 0-len CID restriction.

draft-ietf-quic-transport.md
MikeBishop: Not adding; linking to #2441 where this separate issue will be discussed.

draft-ietf-quic-transport.md
MikeBishop: Not adding; linking to #2441 where this separate issue will be discussed.

draft-ietf-quic-transport.md
```diff
@@ -2089,15 +2091,19 @@ different local addresses, as discussed in {{connection-id}}.  For this to be
 effective endpoints need to ensure that connections IDs they provide cannot be
 linked by any other entity.
 
-This eliminates the use of the connection ID for linking activity from
-the same connection on different networks.  Header protection ensures
-that packet numbers cannot be used to correlate activity.  This does not prevent
+Endpoints MAY move to a new connection ID at any time.
+
+An endpoint MUST use a new connection ID if it initiates connection migration,
+unless the peer has selected a zero-length connection ID.  Using a new
```
ianswett: Above we say you can't migrate if you use a zero-length connection ID, so I don't understand the purpose of "unless the peer has selected a zero-length connection ID."

draft-ietf-quic-transport.md
```diff
@@ -2108,21 +2114,11 @@ genuine migrations.  Changing port number can cause a peer to reset its
...
-privacy requirement, endpoints that initiate migration and use connection IDs
-with length greater than zero SHOULD provide their peers with new connection IDs
-before migration.
-
-Caution:
-
-: If both endpoints change connection ID in response to seeing a change in
-  connection ID from their peer, then this can trigger an infinite sequence of
-  changes.
+An endpoint MUST NOT intentionally migrate to a new path if it does not have a
```
ianswett: This sentence seems duplicative of the paragraph above starting with "An endpoint MUST use a new connection ID ..."

draft-ietf-quic-transport.md
```diff
@@ -2089,15 +2091,18 @@ different local addresses, as discussed in {{connection-id}}.  For this to be
 effective endpoints need to ensure that connections IDs they provide cannot be
 linked by any other entity.
 
-This eliminates the use of the connection ID for linking activity from
-the same connection on different networks.  Header protection ensures
-that packet numbers cannot be used to correlate activity.  This does not prevent
-other properties of packets, such as timing and size, from being used to
+Endpoints MAY move to a new connection ID at any time.
```
ianswett: Nit: move to may not be clear.  How about "Endpoints MAY change the Destination Connection ID they send in short header packets at any time." assuming that statement is correct.

draft-ietf-quic-transport.md
```diff
@@ -2091,7 +2091,8 @@ different local addresses, as discussed in {{connection-id}}.  For this to be
 effective endpoints need to ensure that connections IDs they provide cannot be
 linked by any other entity.
 
-Endpoints MAY move to a new connection ID at any time.
+At any time, endpoints MAY change the Destination Connection ID they send to a
+value that has not been used on another path.
```
ianswett: Much better.  Can this be "has not been used previously on this path or another path."?

martinthomson: It's OK to reuse a connection ID that has been used on the same path.  Unless the goal is to avoid temporal linkability, but that's far less of a pressing issue, I think.

One thing I want to allow is alternating connection IDs, which your suggestion would prevent.

ianswett: I'm happy with this PR, but can you explain why you'd want that?

martinthomson: I don't necessarily, but other have mentioned a desire to put multiple connection IDs on the same path.  This makes one connection look like multiple.

MikeBishop: Of course, you'd need your peer to do the same; otherwise, it's kind of pointless.

ianswett: I'd rather just have multiple connections or in v2, one could use multipath to do this.  Alternating between CIDs on a single path seems like a recipe for problems because few will exercise this option and it has dubious value in my opinion.

mikkelfj: If there is point to changing CID's on same path, then reordering can send packets with the old CID. I'm not sure there is, but it might have relevance to keep up with changes to router state. But I suppose a router will need to be able to track a CID forever if it keeps seeing it periodically.

larseggert:
This seems to have consensus, tagging as such and reassigning.

mnot:
Removing `has-consensus` tag as it needs to be canvassed on-list (and attached to an issue, not a PR).

larseggert:
I'm sorry if I misunderstood the new process, but I don't think your PR says that an issue **must** be discussed on the mailing list? Doesn't it say that the discussion can happen on the list **or** on the issue?

larseggert:
I'm sorry if I misunderstood the new process, but I don't think your PR says that an issue **must** be discussed on the mailing list? Doesn't it say that the discussion can happen on the list **or** on the issue?

mnot:
@larseggert discussion can happen in either place, but we've always checked with the list before declaring consensus (which is what `has-consensus` records). In the currently operative process, that's happening (admittedly sporadically) after the drafts are published.

janaiyengar:
@mnot @larseggert : The associated issue is #2413 and there's been a [mailing list discussion](https://mailarchive.ietf.org/arch/msg/quic/goCZG8crQZ8bM_G1-OGm130kbpM).  There seems to be some disagreement on what needs to happen next.

Specifically, some folks think this PR is an editorial one, and some want to open up the issue as a design issue for discussion.

mnot:
#2413 is a design issue. If this is just editorial and no one disputes that, you can apply it as that, but it can't close #2413.

martinthomson:
#2413 is only marked design because there is debate, therefore, I conclude that we need chairs to help us.  I requested that close to 2 weeks ago.

MikeBishop:
The requirement previously existed, but was accidentally dropped during one of the CID rewrites.  The underlying question is whether we had consensus to remove the requirement previously -- if so, then reversing ourselves requires consensus.  If not, then it's just fixing an accidental deletion.

ianswett:
If the PR that removed it doesn't have a note about removing it, I'd consider it an accidental deletion(which is what I suspect).

erickinnear:
I thought the overall requirement for changing CIDs was [still in the text](https://github.com/quicwg/base-drafts/issues/2413#issuecomment-460457572), so it doesn't seem like the issue is around adding/removing a requirement, but rather the rest of the accompanying text here? 
Otherwise, not sure how this isn't entirely editorial, since the main requirement is already in place.

PR #2417, opened 2019-02-05
update the pseudo code to include packet number spaces

### LINKED ISSUE #2405: loss recovery description mixes variables depending on the packet number space and packet numbers independent of it
[Section 6.5.2](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#rfc.section.6.5.2) lists the variables used by the loss recovery algorithm. 
Some of these variables are independent of the packet number space, e.g. `crypto_count`, `*rtt*.`
Other variables apply per packet number space, and we say so explicitly, e.g. `largest_acked_packet`.
Then there are a lot of variables that should be applied per packet number space, but we never say so explicitly, e.g. `sent_packets`, `loss_time`, etc.

The same applies to the pseudo code itself, for example `DetectLostPackets` uses both variables that only apply to one packet number space as well as variables that apply independent of the packet number space.

One way to resolve this would be to index the variables by their respective packet number space.

marten-seemann:
This is an ongoing attempt to resolve #2405.
In the current state this PR is nowhere near complete, but I hope it gets the discussion going about the necessary changes. I will update it accordingly.

So far, this PR makes the following changes:

- makes it explicit that the variables `largest_sent_packet`, `largest_acked_packet` and `sent_packets` are kept for every packet number space. By indexing these variables with their respective packet number space, we can immediately see where we need to pass a packet number space variable around,
- makes it explicit that `OnPacketAcked` and `DetectLostPackets` are packet-number-space dependent,
- I realized that `loss_time` would also apply to every packet number space, if we didn't have special treatment for crypto packets. Since we have that, we never run early retransmit for crypto packets, so `loss_time` only applies to the 1-RTT packet number space.
  - I'm not sure if is this a bug or a feature. The only way to declare crypto packets lost now is using the crypto timer. Contrary to what we say in the text, this is *less* aggressive than the logic we use for packets sent after completion of the handshake.

Edit: CI fails due to line lengths. I'll defer fixing this until we agree if this is the right way forward.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -819,10 +819,10 @@ kGranularity:
 kInitialRtt:
 : The RTT used before an RTT sample is taken. The RECOMMENDED value is 100ms.
 
-pn_space:
+kPacketNumberSpace:
 : An enum to enumerate the three packet number spaces.
 ~~~
-  enum pn_space {
+  enum kPacketNumberSpace {
     Initial,
     Handshake,
     1-RTT,
@@ -850,10 +850,10 @@ time_of_last_sent_ack_eliciting_packet:
 time_of_last_sent_crypto_packet:
 : The time the most recent crypto packet was sent.
 
-largest_sent_packet[pn_space]:
+largest_sent_packet[kPacketNumberSpace]:
 : The packet number of the most recently sent packet in the packet number space.
 
-largest_acked_packet[pn_space]:
+largest_acked_packet[kPacketNumberSpace]:
 : The largest packet number acknowledged in the packet number space so far.
 
 latest_rtt:
@@ -881,7 +881,7 @@ loss_time:
   exceeding the reordering window in time. Only applies to the 1-RTT packet
   number space.
 
-sent_packets[pn_space]:
+sent_packets[kPacketNumberSpace]:
 : An association of packet numbers in a packet number space to information
   about them.  Described in detail above in {{tracking-sent-packets}}.
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -825,7 +825,7 @@ kPacketNumberSpace:
   enum kPacketNumberSpace {
     Initial,
     Handshake,
-    Application data,
+    ApplicationData,
   }
 ~~~
 
@@ -878,7 +878,7 @@ max_ack_delay:
 
 loss_time:
 : The time at which the next packet will be considered lost based on
-  exceeding the reordering window in time. Only applies to the Application data
+  exceeding the reordering window in time. Only applies to the ApplicationData
   packet number space.
 
 sent_packets[kPacketNumberSpace]:
@@ -1047,7 +1047,7 @@ SetLossDetectionTimer():
     return
   if (loss_time != 0):
     // Time threshold loss detection.
-    // Only applies to the Application data packet number space.
+    // Only applies to the ApplicationData packet number space.
     loss_detection_timer.update(loss_time)
     return
 
@@ -1077,8 +1077,8 @@ OnLossDetectionTimeout():
     crypto_count++
   else if (loss_time != 0):
     // Time threshold loss Detection
-    // Only applies to the Application data packet number space.
-    DetectLostPackets(Application data)
+    // Only applies to the ApplicationData packet number space.
+    DetectLostPackets(ApplicationData)
   else:
     // PTO
     SendOneOrTwoPackets()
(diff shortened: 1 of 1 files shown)


marten-seemann:
@ianswett, @janaiyengar Can you please review this? Is this going in the right direction?

draft-ietf-quic-recovery.md
```diff
@@ -890,8 +891,9 @@ follows:
    min_rtt = infinite
    time_of_last_sent_ack_eliciting_packet = 0
    time_of_last_sent_crypto_packet = 0
-   largest_sent_packet = 0
-   largest_acked_packet = 0
+   for pn_space in packet number spaces:
```
ianswett: I think we should define packet number spaces more clearly above, and possibly call it kPacketNumberSpaces.  I think it makes sense to define it as a list of 3 enum values.

marten-seemann: Not sure if we need that constant, since we're only looping over the spaces once (and only for initialization), but I introduced the enum.

draft-ietf-quic-recovery.md
```diff
@@ -903,20 +915,22 @@ to OnPacketSent are described in detail above in {{sent-packets-fields}}.
...
 
 ~~~
- OnPacketSent(packet_number, ack_eliciting, in_flight,
-              is_crypto_packet, sent_bytes):
-   largest_sent_packet = packet_number
-   sent_packets[packet_number].packet_number = packet_number
-   sent_packets[packet_number].time_sent = now
-   sent_packets[packet_number].ack_eliciting = ack_eliciting
-   sent_packets[packet_number].in_flight = in_flight
+ OnPacketSent(packet_number, pn_space, ack_eliciting,
```
ianswett: Using pn_space for the variable name and the type is confusing.  I think the enum should be kPacketNumberSpace

marten-seemann: Done.

draft-ietf-quic-recovery.md
```diff
@@ -819,6 +819,16 @@ kGranularity:
 kInitialRtt:
 : The RTT used before an RTT sample is taken. The RECOMMENDED value is 100ms.
 
+kPacketNumberSpace:
+: An enum to enumerate the three packet number spaces.
+~~~
+  enum kPacketNumberSpace {
+    Initial,
+    Handshake,
+    1-RTT,
```
kazuho: I think the name for the PN space that governs 0- and 1-RTT packets is “application data”.

marten-seemann: Done.

draft-ietf-quic-recovery.md
```diff
@@ -819,6 +819,16 @@ kGranularity:
 kInitialRtt:
 : The RTT used before an RTT sample is taken. The RECOMMENDED value is 100ms.
 
+kPacketNumberSpace:
+: An enum to enumerate the three packet number spaces.
+~~~
+  enum kPacketNumberSpace {
+    Initial,
+    Handshake,
+    Application data,
```
ianswett: It's an enum, so how about ApplicationData

OT: At some point there's a hope to make this pseudocode compile in Python.

marten-seemann: Python enums don't look particularly nice though: https://docs.python.org/3/library/enum.html#creating-an-enum

marten-seemann: I renamed the variable.

ianswett:
Good point, maybe we should pick a different language then.

draft-ietf-quic-recovery.md
```diff
@@ -890,8 +901,9 @@ follows:
    min_rtt = infinite
    time_of_last_sent_ack_eliciting_packet = 0
    time_of_last_sent_crypto_packet = 0
-   largest_sent_packet = 0
-   largest_acked_packet = 0
+   for pn_space in [ Initial, Handshake, Applicaton data ]:
```
```suggestion
   for pn_space in [ Initial, Handshake, ApplicatonData ]:
```

draft-ietf-quic-recovery.md
```diff
@@ -1012,37 +1027,38 @@ timers wake up late. Timers set in the past SHOULD fire immediately.
...
+      timeout = 2 * smoothed_rtt
     timeout = max(timeout, kGranularity)
-    timeout = timeout * (2 ^ pto_count)
-
+    timeout = timeout * (2 ^ crypto_count)
     loss_detection_timer.update(
-      time_of_last_sent_ack_eliciting_packet + timeout)
+      time_of_last_sent_crypto_packet + timeout)
+    return
+  if (loss_time != 0):
```
```suggestion

  if (loss_time != 0):
```

janaiyengar:
I've fixed the nits -- merging.

PR #2419, opened 2019-02-05
Add table to show what frames go on what streams

### LINKED ISSUE #2351: Indicate which frames go on which stream types
While each frame's description indicates the stream type(s) on which it is allowed, there isn't one unified place to indicate which frames are expected on which streams.  QPACK subdivides the instruction space per stream, which has the nice property that you _can't_ put something on the wrong stream.  While that's overkill for H3 (too dramatic a departure from H2, multiple IANA registries, etc.), having a table to indicate which frames are permitted on which streams would be helpful.

LPardue:
Fixes #2351

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -396,9 +396,24 @@ implementation chooses.
 
 # HTTP Framing Layer {#http-framing-layer}
 
-As discussed above, frames are carried on QUIC streams and used on control
-streams, request streams, and push streams.  This section describes HTTP framing
-in QUIC.  For a comparison with HTTP/2 frames, see {{h2-frames}}.
+HTTP frames are carried on QUIC streams, as described in {{stream-mapping}}.
+HTTP/3 defines three stream types: control stream, request stream, and push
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -396,9 +396,24 @@ implementation chooses.
 
 # HTTP Framing Layer {#http-framing-layer}
 
-As discussed above, frames are carried on QUIC streams and used on control
-streams, request streams, and push streams.  This section describes HTTP framing
-in QUIC.  For a comparison with HTTP/2 frames, see {{h2-frames}}.
+HTTP frames are carried on QUIC streams, as described in {{stream-mapping}}.
+HTTP/3 defines three stream types: control stream, request stream, and push
+stream. This section describes HTTP/3 frame formats and the streams types which
+they are permitted on, see {{stream-frame-mapping}} for an overiew.  A
```
MikeBishop: "on which they are permitted; see ... overview"

LPardue:
So a friendly pointer to detail that could be easily misinterpreted or missed? It seems like a table footnote is required but I'm not sure how to do that in Markdown.

PR #2424, opened 2019-02-06
Reword Immediate ACK Recommendation

martinduke:
This is a reword of #2025, as discussed with @ianswett previously.

It cleans up two problems:
1) Strictly speaking, the MAY in the original text is redundant, because receivers MAY do anything except exceed the delayed ack timeout. I think it reads better now.
2) @martinthomson mentioned that the spec was a little ambiguous about how to handle additional OOO packets during the interval. I put down some language, but I'm not sure it's what @ianswett would recommend.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -138,7 +138,8 @@ Crypto Packets:
 Out-of-order Packets:
 
 : Packets that do not increase the largest received packet number for its
-cryptographic context by exactly one.
+packet number space by exactly one. Packets usually arrive out of order
+because earlier packets are lost or delayed.
 
 # Design of the QUIC Transmission Machinery
 
@@ -251,7 +252,9 @@ an ACK immediately when receiving a second ack-eliciting packet.
 In order to accelerate loss recovery and reduce timeouts, the receiver SHOULD
 send immediate ACKs for an interval after it receives an out-of-order packet.
 This interval SHOULD NOT exceed 1/8 RTT unless more out-of-order packets arrive
-during the interval.
+during the interval. Therefore, if out-of-order packets consistently arrive
+within one interval of the previous one, the receiver SHOULD continuously send
+immediate ACKs.
 
 Similarly, packets marked with the ECN Congestion Experienced (CE) codepoint in
 the IP header SHOULD be acknowledged immediately, to reduce the peer's response
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -134,7 +134,7 @@ Crypto Packets:
 
 : Packets containing CRYPTO data sent in Initial or Handshake
   packets.
-  
+
 Out-of-order Packets:
 
 : Packets that do not increase the largest received packet number for its
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -138,7 +138,7 @@ Crypto Packets:
 Out-of-order Packets:
 
 : Packets that do not increase the largest received packet number for its
-packet number space by exactly one. Packets usually arrive out of order
+  packet number space by exactly one. Packets arrive out of order
   when earlier packets are lost or delayed.
 
 # Design of the QUIC Transmission Machinery
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -139,7 +139,7 @@ Out-of-order Packets:
 
 : Packets that do not increase the largest received packet number for its
 packet number space by exactly one. Packets usually arrive out of order
-because earlier packets are lost or delayed.
+  when earlier packets are lost or delayed.
 
 # Design of the QUIC Transmission Machinery
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -134,6 +134,11 @@ Crypto Packets:
 
 : Packets containing CRYPTO data sent in Initial or Handshake
   packets.
+  
+Out-of-order Packets:
+
+: Packets that do not increase the largest received packet number for its
+cryptographic context by exactly one.
 
 # Design of the QUIC Transmission Machinery
 
@@ -244,11 +249,9 @@ ack-eliciting packet. QUIC recovery algorithms do not assume the peer sends
 an ACK immediately when receiving a second ack-eliciting packet.
 
 In order to accelerate loss recovery and reduce timeouts, the receiver SHOULD
-send an immediate ACK when it receives a new packet which is not one greater
-than the largest received packet number. A receiver MAY send immediate ACKs
-for the next few ack-eliciting packets that are received, but SHOULD NOT
-send an immediate ACK for more than 1/8 RTT after receiving an out-of-order
-packet.
+send immediate ACKs for an interval after it receives an out-of-order packet.
+This interval SHOULD NOT exceed 1/8 RTT unless more out-of-order packets arrive
+during the interval.
 
 Similarly, packets marked with the ECN Congestion Experienced (CE) codepoint in
 the IP header SHOULD be acknowledged immediately, to reduce the peer's response
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -250,7 +250,7 @@ ack-eliciting packet. QUIC recovery algorithms do not assume the peer sends
 an ACK immediately when receiving a second ack-eliciting packet.
 
 In order to accelerate loss recovery and reduce timeouts, the receiver SHOULD
-send an immediate ACK after it receives an out-of-order packet. it could send
+send an immediate ACK after it receives an out-of-order packet. It could send
 immediate ACKs for in-order packets for a period of time that SHOULD NOT exceed
 1/8 RTT unless more out-of-order packets arrive. If every packet arrives out-of-
 order, then an immediate ACK SHOULD be sent for every received packet.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -134,6 +134,11 @@ Crypto Packets:
 
 : Packets containing CRYPTO data sent in Initial or Handshake
   packets.
+  
+Out-of-order Packets:
+
+: Packets that do not increase the largest received packet number for its
+cryptographic context by exactly one.
```
martinthomson: Do you want to add: "Packets can arrive out of order as a result of packets ahead of them being lost or delayed."  I think that's accurate.

draft-ietf-quic-recovery.md
```diff
@@ -244,11 +249,9 @@ ack-eliciting packet. QUIC recovery algorithms do not assume the peer sends
...
 
 In order to accelerate loss recovery and reduce timeouts, the receiver SHOULD
-send an immediate ACK when it receives a new packet which is not one greater
-than the largest received packet number. A receiver MAY send immediate ACKs
-for the next few ack-eliciting packets that are received, but SHOULD NOT
-send an immediate ACK for more than 1/8 RTT after receiving an out-of-order
-packet.
+send immediate ACKs for an interval after it receives an out-of-order packet.
+This interval SHOULD NOT exceed 1/8 RTT unless more out-of-order packets arrive
+during the interval.
```
martinthomson: So the idea is that every out-of-order packet triggers a 1/8 RTT interval of immediate acknowledgments?

I thought that the idea was to disable the sending of immediate acknowledgments if out-of-order packets started becoming routine.  That is, if you have a highly unreliable path, then immediate acknowledgments aren't helping repair errors, they are just noise.

draft-ietf-quic-recovery.md
```diff
@@ -135,6 +135,11 @@ Crypto Packets:
 : Packets containing CRYPTO data sent in Initial or Handshake
   packets.
 
+Out-of-order Packets:
+
+: Packets that do not increase the largest received packet number for its
+cryptographic context by exactly one.
```
```suggestion
  packet number space by exactly one.
```

draft-ietf-quic-recovery.md
ianswett: I'd prefer "Packets arrive out when later packets are lost or delayed."

draft-ietf-quic-recovery.md
```diff
@@ -244,11 +249,9 @@ ack-eliciting packet. QUIC recovery algorithms do not assume the peer sends
 an ACK immediately when receiving a second ack-eliciting packet.
 
 In order to accelerate loss recovery and reduce timeouts, the receiver SHOULD
-send an immediate ACK when it receives a new packet which is not one greater
-than the largest received packet number. A receiver MAY send immediate ACKs
-for the next few ack-eliciting packets that are received, but SHOULD NOT
-send an immediate ACK for more than 1/8 RTT after receiving an out-of-order
-packet.
+send immediate ACKs for an interval after it receives an out-of-order packet.
```
ianswett: The SHOULD is intended to apply to just the first immediate ACK, as with ECN marks.

draft-ietf-quic-recovery.md
martinthomson: I think that there are two cases we care about, I only caught the first

- an earlier packet was delayed - so there is a gap before this one
- this packet was delayed - and we already received a higher-numbered one

Maybe just: "Packets arrive out of order when packets are lost or delayed."  Which doesn't really capture all the reasons, but might suffice.

draft-ietf-quic-recovery.md
martinthomson: So I'm still not clear.  If I get a continuous stream of reordered packets, a) when do I cease immediate ACKs, and b) if I do stop immediate ACKs, when can I start again?

ianswett: If every packet is reordered(ie: every other packet is missing), then you keep sending immediate ACKs.

martinthomson: So it isn't the first immediate ACK, it's every out-of-order packet that starts (or refreshes) a 1/8 RTT timer.  That helps.

martinduke:
@ianswett 

"I'd prefer "Packets arrive out when later packets are lost or delayed.""

earlier packets?

Do you want this text to replace the current definition, or in addition to the current PR?

draft-ietf-quic-recovery.md
ianswett: My suggestion should have been "Packets arrive out of order when later packets are lost or delayed.", but Martin's suggestion is slightly better because it drops 'later', which is unnecessary.

The intent was as a replacement for Martin's suggestion, in addition to the first sentence you have.

draft-ietf-quic-recovery.md
```diff
@@ -135,6 +135,12 @@ Crypto Packets:
 : Packets containing CRYPTO data sent in Initial or Handshake
   packets.
 
+Out-of-order Packets:
+
+: Packets that do not increase the largest received packet number for its
+packet number space by exactly one. Packets usually arrive out of order
```
```suggestion
  packet number space by exactly one. Packets arrive out of order
```

draft-ietf-quic-recovery.md
```diff
@@ -135,6 +135,12 @@ Crypto Packets:
...
+because earlier packets are lost or delayed.
```
```suggestion
  when earlier packets are lost or delayed.
```

draft-ietf-quic-recovery.md
```diff
@@ -244,11 +250,11 @@ ack-eliciting packet. QUIC recovery algorithms do not assume the peer sends
 an ACK immediately when receiving a second ack-eliciting packet.
 
 In order to accelerate loss recovery and reduce timeouts, the receiver SHOULD
-send an immediate ACK when it receives a new packet which is not one greater
-than the largest received packet number. A receiver MAY send immediate ACKs
-for the next few ack-eliciting packets that are received, but SHOULD NOT
-send an immediate ACK for more than 1/8 RTT after receiving an out-of-order
-packet.
+send immediate ACKs for an interval after it receives an out-of-order packet.
```
ianswett: The intent of the text is you SHOULD send one immediate ACK, but you may send more than one immediate ACK.  This new text changes that to you SHOULD send an immediate ACK for an interval, and I don't want to make that a SHOULD, only a MAY.

draft-ietf-quic-recovery.md
```diff
@@ -244,11 +250,11 @@ ack-eliciting packet. QUIC recovery algorithms do not assume the peer sends
...
+This interval SHOULD NOT exceed 1/8 RTT unless more out-of-order packets arrive
+during the interval. Therefore, if out-of-order packets consistently arrive
```
ianswett: Suggestion on an alternate clarification to Martin's question: "If every packet arrives out-of-order, then an immediate ACK will be sent for every received packet."

draft-ietf-quic-recovery.md
martinduke: The exceptions I had in mind are packet duplication and misbehaving senders.

draft-ietf-quic-recovery.md
martinduke: The text allows the interval to be zero, but I agree that's weaselly. I'll clarify more.

draft-ietf-quic-recovery.md
ianswett: Duplication doesn't count as out-of-order because the packet is ignored.  Misbehaving senders are certainly possible, but I'm not sure how important they are to this text?

draft-ietf-quic-recovery.md
```diff
@@ -244,11 +250,11 @@ ack-eliciting packet. QUIC recovery algorithms do not assume the peer sends
 an ACK immediately when receiving a second ack-eliciting packet.
 
 In order to accelerate loss recovery and reduce timeouts, the receiver SHOULD
-send an immediate ACK when it receives a new packet which is not one greater
-than the largest received packet number. A receiver MAY send immediate ACKs
-for the next few ack-eliciting packets that are received, but SHOULD NOT
-send an immediate ACK for more than 1/8 RTT after receiving an out-of-order
-packet.
+send immediate ACKs for an interval after it receives an out-of-order packet.
+This interval SHOULD be between zero (immediately acking only out-of-order
```
ianswett: I'm finding this increasingly hard to read.  Can we pop the stack and you can explain what's unclear about the existing text?

martinduke:
we've touched base offline. We understand each other, I'll rewrite to reduce bloat

draft-ietf-quic-recovery.md
```diff
@@ -244,11 +250,10 @@ ack-eliciting packet. QUIC recovery algorithms do not assume the peer sends
 an ACK immediately when receiving a second ack-eliciting packet.
 
 In order to accelerate loss recovery and reduce timeouts, the receiver SHOULD
-send an immediate ACK when it receives a new packet which is not one greater
-than the largest received packet number. A receiver MAY send immediate ACKs
-for the next few ack-eliciting packets that are received, but SHOULD NOT
-send an immediate ACK for more than 1/8 RTT after receiving an out-of-order
-packet.
+send an immediate ACK after it receives an out-of-order packet. it could send
```
```suggestion
send an immediate ACK after it receives an out-of-order packet. It could send
```

draft-ietf-quic-recovery.md
```diff
@@ -135,6 +135,12 @@ Crypto Packets:
 : Packets containing CRYPTO data sent in Initial or Handshake
   packets.
 
+Out-of-order Packets:
+
+: Packets that do not increase the largest received packet number for its
+  packet number space by exactly one. Packets arrive out of order
+  when earlier packets are lost or delayed.
```
ianswett: I'd be inclined to drop earlier because it's potentially confusing.
```suggestion
  when packets are lost or delayed.
```

nibanks: I don't know... I think it might read better with 'earlier' included. I don't think it's confusing.

PR #2431, opened 2019-02-06
Discard inconsistent packets

### LINKED ISSUE #2397: "Invariant check" for any inbound LH packet
Consensus on Slack seems to be that a server should set the version field of a Retry packet to v1 (i.e., echo the version from the client Initial ), and that a client should verify that a received Retry packet echoes the version it sent.

However, I cannot find corresponding text in the documents.

MikeBishop:
Fixes #2397 by discarding packets which are "inconsistent with the state of the connection."  Previously, this was only a failure to remove packet protection; incorrect version is added as another example inconsistency.

Note that this phrasing doesn't rule out upward-VN like #1773, since receiving an expected upgrade is not inconsistent with your local state.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1011,9 +1011,10 @@ cannot be attributed to an existing connection. A stateless reset allows a peer
 to more quickly identify when a connection becomes unusable.
 
 Packets that are matched to an existing connection are discarded if the packets
-are inconsistent with the state of that connection -- for example, if they
-indicate a different protocol version than that of the connection, or if the
-removal of packet protection is unsuccessful once the expected keys are available.
+are inconsistent with the state of that connection.  For example, packets are
+discarded if they indicate a different protocol version than that of the
+connection, or if the removal of packet protection is unsuccessful once the
+expected keys are available.
 
 Invalid packets without packet protection, such as Initial, Retry, or Version
 Negotiation, MAY be discarded.  An endpoint MUST generate a connection error if
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1013,7 +1013,7 @@ to more quickly identify when a connection becomes unusable.
 Packets that are matched to an existing connection are discarded if the packets
 are inconsistent with the state of that connection -- for example, if they
 indicate a different protocol version than that of the connection, or if the
-removal of packet protection is unsuccessful.
+removal of packet protection is unsuccessful once the expected keys are available.
 
 Invalid packets without packet protection, such as Initial, Retry, or Version
 Negotiation, MAY be discarded.  An endpoint MUST generate a connection error if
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1010,8 +1010,10 @@ Endpoints can send a Stateless Reset ({{stateless-reset}}) for any packets that
 cannot be attributed to an existing connection. A stateless reset allows a peer
 to more quickly identify when a connection becomes unusable.
 
-Packets that are matched to an existing connection, but for which the endpoint
-cannot remove packet protection, are discarded.
+Packets that are matched to an existing connection are discarded if the packets
+are inconsistent with the state of that connection -- for example, if they
+indicate a different protocol version than that of the connection, or if the
+endpoint cannot remove packet protection.
```
mikkelfj: During handshake, reordering can lead to temporarily undecryptable content which may be discarded or buffered. This is technically not inconsistent with state, but the above formulation could suggest to always drop such packets.

MikeBishop: Perhaps "if removal of packet protection fails" would exclude the case where it can't be attempted yet?

draft-ietf-quic-transport.md
```diff
@@ -1010,8 +1010,10 @@ Endpoints can send a Stateless Reset ({{stateless-reset}}) for any packets that
 cannot be attributed to an existing connection. A stateless reset allows a peer
 to more quickly identify when a connection becomes unusable.
 
-Packets that are matched to an existing connection, but for which the endpoint
-cannot remove packet protection, are discarded.
+Packets that are matched to an existing connection are discarded if the packets
+are inconsistent with the state of that connection -- for example, if they
```
martinthomson: Two sentences > 1.

draft-ietf-quic-transport.md
```diff
@@ -1010,8 +1010,10 @@ Endpoints can send a Stateless Reset ({{stateless-reset}}) for any packets that
...
+indicate a different protocol version than that of the connection, or if the
+removal of packet protection is unsuccessful.
```
```suggestion
removal of packet protection is unsuccessful once the expected keys are available.
```

mikkelfj:
I'm wondering if there is a difference between failed decryption and version number.

If the versions differ, it doesn't really make sense to try to read a subsequent coalesced packet, unless you want to do trial and error since the concepts of coalescing and length fields are not defined in the general case. Even if you understand the version, two version can not (or should not?) exist in the same datagram.

If packet protection fails on the same version, it might just be a timing issue, or the (to me) dreaded path probe with prefixed knowingly invalid handshake.

MikeBishop:
That's an interesting question.  The Length field is version-specific.  If you include it, you can have no expectation that a peer which doesn't speak that version will even be able to find anything further in the datagram -- as far as it knows, the remainder of the datagram is that connection.  But that actually has a potential use.

On receipt, if the logic first attempts to chunk up the datagram and the subsequent packet is the correct version for the connection, then acknowledging that second coalesced packet implies that you support the first packet's version.

PR #2433, opened 2019-02-07
Require RFC8470

### LINKED ISSUE #2432: HTTP/3 doesn't mandate RFC8470 compliance

MikeBishop:
Fixes #2432.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -234,6 +234,17 @@ SETTINGS frame. After the QUIC connection is established, a SETTINGS frame
 ({{frame-settings}}) MUST be sent by each endpoint as the initial frame of their
 respective HTTP control stream (see {{control-streams}}).
 
+## Early Data and HTTP/3
+
+QUIC provides a facility for early data during the first flight of packets sent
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -234,6 +234,17 @@ SETTINGS frame. After the QUIC connection is established, a SETTINGS frame
 ({{frame-settings}}) MUST be sent by each endpoint as the initial frame of their
 respective HTTP control stream (see {{control-streams}}).
 
+## Early Data and HTTP/3
+
+QUIC provides a facility for early data during the first flight of packets sent
+by the client.  Early data is not subject to the same security guarantees as
+data sent after the handshake completes.  In particular, this data can be
+replayed by an attacker able to capture the packets and the server will not know
```
martinthomson: This isn't entirely true.  Capturing and replaying packets is not sufficient, or even necessary for an attack.  I would instead point at TLS 1.3 and the discussion about what can happen there than try to repeat it.

MikeBishop:
I missed / forgot that you'd added this requirement in #2355.  I'm happy to just run with that.

PR #2438, opened 2019-02-08
QPACK table lookup noob clarifications

ghedo:
A couple of clarifications that can be useful to people who are looking at *PACK for the first time like myself. Not sure about the normative language though. I can reword it if necessary.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -362,8 +362,8 @@ addressed.
 The static table consists of a predefined static list of header fields, each of
 which has a fixed index over time.  Its entries are defined in {{static-table}}.
 
-All entries in the static table have a name and a value.  However values can be
-empty, that is, have a length of 0.
+All entries in the static table have a name and a value.  However, values can be
+empty (that is, have a length of 0).
 
 Note the QPACK static table is indexed from 0, whereas the HPACK static table
 is indexed from 1.
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -996,6 +996,9 @@ pattern.  If the entry is in the dynamic table with an absolute index greater
 than or equal to the Base, the representation starts with the '0000' four-bit
 pattern.
 
+Only the header field name stored in the static or dynamic table is used. Any
+header field value MUST be ignored.
+
 The following bit, 'N', indicates whether an intermediary is permitted to add
 this header to the dynamic header table on subsequent hops. When the 'N' bit is
 set, the encoded header MUST always be encoded with a literal representation. In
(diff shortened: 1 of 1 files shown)


dtikhonov:
Let's see how [HPACK](https://tools.ietf.org/html/rfc7541#section-6.1) handles it:

```
   Either form of header field name representation is followed by the
   header field value represented as a string literal (see Section 5.2).
```

We could simply add similar verbiage to Sections [4.5.4](https://tools.ietf.org/html/draft-ietf-quic-qpack-06#section-4.5.4) and 4.5.5. of the QPACK draft instead.

draft-ietf-quic-qpack.md
```diff
@@ -963,6 +963,9 @@ followed by the `S` bit indicating whether the reference is into the static
 field is represented as an integer with a 6-bit prefix (see Section 5.1 of
 [RFC7541]).
 
+If the entry in the static or dynamic table does not provide any header field
+value (as opposed to a name), a zero-length string MUST be used.
+
```
MikeBishop: Setting aside the legacy HPACK table that had nothing at index 0 (and my implementation did indeed return `(String.Empty,String.Empty)`), is this even possible?  If there's a reference to an index which isn't valid, that's an error.  If it is valid, then something has been inserted there -- which might have been a pair of empty strings, but that's still the contents.

LPardue: I think the "problem case" is referencing a static table item with empty value. We discussed how to build the header list and felt that it it the job of the H3 endpoint to validate the header list, not QPACKs. This change makes it clear what to do, and what H3 could check in validation.

martinthomson: I think that it's easiest to say that entries in the static table all have a value, but that some values are empty.  This text seems to confer some sort of special status on "absent" values.

ghedo: @martinthomson I reworded that part and moved it to the "Static Table" section.

ghedo:
@dtikhonov I can add something like that, but TBH I don't think that the HPACK wording would have helped me, so I'd rather keep my clarification (that says "ignore value") as well.

draft-ietf-quic-qpack.md
```diff
@@ -362,6 +362,9 @@ addressed.
 The static table consists of a predefined static list of header fields, each of
 which has a fixed index over time.  Its entries are defined in {{static-table}}.
 
+All entries in the static table have a name and a value.  However values can be
```
MikeBishop:
```suggestion
All entries in the static table have a name and a value.  However, values can be
```

draft-ietf-quic-qpack.md
```diff
@@ -362,6 +362,9 @@ addressed.
...
+empty, that is, have a length of 0.
```
MikeBishop:
```suggestion
empty (that is, have a length of 0).
```

Issue #2339, opened 2019-01-16
QPACK: instructions and stream description is suboptimal


LPardue:
Did the section hierarchy get messed up a little in a recent refactor?

The concept of QPACK instructions are first introduced in section 2, along with encoder and decoder streams. There is no forward reference to the later definitions.

The real problem I have is with section 4.2. It contains a bulleted list that is almost identical to section 2, but subtly different. The text following this list is also ambiguous

> There MUST be exactly one of each unidirectional stream type in each direction.

The previous list referred to request streams and push streams. These are obviously out of scope but it could get mistaken.

I think we've discussed elsewhere that implementations could decide not to use dynamic compression, and so the encoder and decoder streams are not necessary. I think this sentence could be made clearer by saying something like "if the dynamic table is being used, each endpoint MUST create exactly one pair of encoder and decoder streams."

> Receipt of a second instance of either stream type MUST be treated as a connection error of HTTP_WRONG_STREAM_COUNT.

Perhaps simplify to "receipt of additional streams which claim to be an encoder or decoder stream MUST be treated as a connection error of HTTP_WRONG_STREAM_COUNT.

> Closure of either unidirectional stream MUST be treated as a connection error of type HTTP_CLOSED_CRITICAL_STREAM.

Perhaps "The closure of encoder or decoder streams, at any point, MUST be treated as a connection error of type HTTP_CLOSED_CRITICAL_STREAM."

> This section describes the instructions which are possible on each stream type.

No it doesn't :)

While I am here, Section 4.5 is titled "Request and Push Streams" but it just talks about instructions to include in frames. 

I'm left with a feeling that there is a misplaced focus on streams rather than instructions. This could be inverted to state instructions, with a comment on where they are allowed to appear, summarised in a nice table somewhere in the document.

afrind:
The big editorial refactor intentionally reordered the sections: the goal was to describe concepts first and specific instructions second.

Regarding whether the encoder/decoder streams are required, there's a separate (design) issue open (#2100). 

Given that the instruction spaces are unique per stream type, and no instruction is allowed on more than one stream, I think it makes sense to keep the instructions organized by the stream on which they are used.  

I can clean up the ambiguous references to unidirectional streams to make it clear that we're talking about the QPACK control streams.

I'll give it another read through with your comments in mind and see what further improvements we can make.

LPardue:
I'm happy to come up with a PR containing some proposals. 

[...]

PR #2440, opened 2019-02-08
Modify QPACK steam and instruction terminology

LPardue:
This PR is more controversial than I first anticipated. It is intended to address #2339.

QPACK uses the term instruction extensively section 4 goes into great detail about all the different types. However, the instructions have been presented in terms of what streams they appear on. Which makes things a little unclear, and verbose, sometimes. I think it also conflates the idea of the true QPACK unframed encoder and decoder streams, with header block messages (sets of instructions).

So this change explicitly creates the terms encoder instruction, decoder instruction and header block instruction. It also removes a lot of the reference to request and push streams.

I'm not 100% happy with the net outcome of these changes but posting the PR for feedback.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -311,8 +311,8 @@ streams, the decoder sends an Insert Count Increment instruction (see
 ## Decoder
 
 As in HPACK, the decoder processes header blocks and emits the corresponding
-header lists. It also processes dynamic table modifications from instructions on
-the encoder stream.
+header lists. It also processes dynamic table modifications from encoder
+instructions received on the encoder stream.
 
 The decoder MUST emit header fields in the order their representations appear in
 the input header block.
@@ -320,8 +320,8 @@ the input header block.
 
 ### State Synchronization
 
-The decoder stream ({{decoder-stream}}) signals key events at the decoder that
-permit the encoder to track the decoder's state.  These events are:
+The decoder instructions ({{decoder-instructions}}) signal key events at the
+decoder that permit the encoder to track the decoder's state.  These events are:
 
 - Complete processing of a header block
 - Abandonment of a stream which might have remaining header blocks
@@ -365,8 +365,8 @@ which has a fixed index over time.  Its entries are defined in {{static-table}}.
 Note the QPACK static table is indexed from 0, whereas the HPACK static table
 is indexed from 1.
 
-When the decoder encounters an invalid static table index on a request stream or
-push stream it MUST treat this as a stream error of type
+When the decoder encounters an invalid static table index in a header block
+instruction it MUST treat this as a stream error of type
 `HTTP_QPACK_DECOMPRESSION_FAILED`.  If this index is received on the encoder
 stream, this MUST be treated as a connection error of type
 `HTTP_QPACK_ENCODER_STREAM_ERROR`.
@@ -374,8 +374,9 @@ stream, this MUST be treated as a connection error of type
 ## Dynamic Table {#table-dynamic}
 
 The dynamic table consists of a list of header fields maintained in first-in,
-first-out order.  The dynamic table is initially empty.  Entries are added by
-instructions on the encoder stream (see {{encoder-stream}}).
+first-out order. Each HTTP/3 endpoint holds a dynamic table that is initially
+empty.  Entries are added by encoder instructions received on the encoder stream
+(see {{encoder-instructions}}).
 
 The dynamic table can contain duplicate entries (i.e., entries with the same
 name and same value).  Therefore, duplicate entries MUST NOT be treated as an
@@ -446,7 +447,8 @@ rejected, the maximum table capacity is 0 until the encoder processes a SETTINGS
 frame with a non-zero value of SETTINGS_QPACK_MAX_TABLE_CAPACITY.
 
 When the maximum table capacity is 0, the encoder MUST NOT insert entries into
-the dynamic table, and MUST NOT send any instructions on the encoder stream.
+the dynamic table, and MUST NOT send any encoder instructions on the encoder
+stream.
 
 
 ### Absolute Indexing {#indexing}
@@ -463,7 +465,7 @@ The relative index begins at zero and increases in the opposite direction from
 the absolute index.  Determining which entry has a relative index of "0" depends
 on the context of the reference.
 
-On the encoder stream, a relative index of "0" always refers to the most
+In encoder instructions, a relative index of "0" always refers to the most
 recently inserted value in the dynamic table.  Note that this means the entry
 referenced by a given relative index will change while interpreting instructions
 on the encoder stream.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -713,16 +722,15 @@ references (see {{reference-tracking}}).  Changing the capacity of the dynamic
...
-## Decoder Stream
+## Decoder Instructions {#decoder-instructions}
 
-The decoder stream carries information used to ensure consistency of the dynamic
-table. Information is sent from the decoder to the encoder; that is, the server
-informs the client about the processing of the client's header blocks and table
-updates, and the client informs the server about the processing of the server's
-header blocks and table updates.
+Decoder instructions provide information used to ensure consistency of the
+dynamic table. They are sent from the decoder to the encoder on a decoder stream;
```
MikeBishop: Need to rewrap this paragraph -- your build break is an overlong line here.

draft-ietf-quic-qpack.md
```diff
@@ -580,40 +581,48 @@ prefix integer.  The remainder of the string literal is unmodified.
...
-## Stream Types
+## Instructions
 
-QPACK instructions occur in three locations, each of which uses a separate
-instruction space:
+There are three separate QPACK instruction spaces. Encoder instructions
+({{encoder-instructions}}) carry table updates, decoder instructions
+({{decoder-instructions}}) carry acknowledgments of table modifications and
+header processing, and header block instructions ({{header-block-instructions}})
+reference the QPACK table state.
```
martinthomson: Isn't the main purpose of header block instructions to cause header fields to be emitted?

draft-ietf-quic-qpack.md
```diff
@@ -580,40 +581,48 @@ prefix integer.  The remainder of the string literal is unmodified.
...
-This section describes the instructions which are possible on each stream type.
+ - A decoder stream is a unidirectional stream of type `0x68` (ASCII 'h').
+   It carries an unframed sequence of decoder instructions from decoder
+   to encoder.
 
-## Encoder Stream {#encoder-stream}
+<!-- s/exactly/no more than/  ? -->
+HTTP/3 endpoints contain a QPACK encoder and decoder. Each endpoint MUST
+initiate a single encoder stream and decoder stream. Receipt of a second
+instance of either stream type MUST treat it as a connection error of type
```
martinthomson: Check grammar: "An endpoint that receives a second" or "MUST be treated as"

draft-ietf-quic-qpack.md
```diff
@@ -811,11 +819,13 @@ An encoder cannot infer from this instruction that any updates to the dynamic
 table have been received.
 
 
-## Request and Push Streams
+## Header Block Instructions
+
+HTTP/3 endpoints convert header lists to headers blocks and exchange them inside
+HEADERS and PUSH_PROMISE frames. Header block instructions reference the static
+table, or dynamic table in a particular state without modifying it.
```
martinthomson: Same comment as before.  The primary purpose is to convey the header field values.  That they reference various tables to do so is secondary.

LPardue: Hmm. I think I agree with your position, which is why I have some struggle with the text as it currently stand. The changes here try to carry the existing QPACK position and text where possible. I'll take editor(s) cue on whether to broaden the changes in this PR.

AFAICT the QPACK approach was to discuss instructions and how they relate to the dynamic table. In that sense, this paragraph provides a clear statement that these instructions do not mutate.

draft-ietf-quic-qpack.md
LPardue: Ack

draft-ietf-quic-qpack.md
martinthomson: I'm OK with that aspect of this.  However, I would first say what these are for, then explain that they don't refer to the tables, but don't modify them.

draft-ietf-quic-qpack.md
```diff
@@ -580,40 +581,48 @@ prefix integer.  The remainder of the string literal is unmodified.
...
-## Stream Types
+## Instructions
 
-QPACK instructions occur in three locations, each of which uses a separate
-instruction space:
+There are three separate QPACK instruction spaces. Encoder instructions
+({{encoder-instructions}}) carry table updates, decoder instructions
+({{decoder-instructions}}) carry acknowledgments of table modifications and
+header processing, and header block instructions ({{header-block-instructions}})
+compress a header list by referring to the QPACK table state.
```
MikeBishop: Well, they don't really *compress* the header list, they are the compression output.  Perhaps "reconstruct" a header list?

draft-ietf-quic-qpack.md
```diff
@@ -1128,7 +1141,7 @@ The entries in the following table are registered by this document.
 
 ## Stream Type Registration
 
-This document creates two new settings in the "HTTP/3 Stream Type" registry
+This document creates two new streams in the "HTTP/3 Stream Type" registry
```
MikeBishop: Perhaps "types" or "entries"?

LPardue: The change addressed a bug in the spec. I'm happy to use entries but would do so consistently through the document (settings, errors, stream)

draft-ietf-quic-qpack.md
LPardue: Well, an encoder view of the header block is not to reconstruct the list. 

The initial text simply stated that these instructions refer to tables. Adding flavour seems to complicate the matter, unless we think spelling out verbatim is useful. 

Would something like "header block instructions convey an encoded representation of a header list by referring to the QPACK table state" suffice?

MikeBishop: WFM.

draft-ietf-quic-qpack.md
MikeBishop: Well, arguably, the other two actually are creating new settings and error types.  But yes, if you want to fix the bug and remain consistent, this should probably be "types".

draft-ietf-quic-qpack.md
```diff
@@ -580,40 +581,49 @@ prefix integer.  The remainder of the string literal is unmodified.
...
+   It carries an unframed sequence of encoder instructions from encoder
+   to decoder.
 
-This section describes the instructions which are possible on each stream type.
+ - A decoder stream is a unidirectional stream of type `0x68` (ASCII 'h').
+   It carries an unframed sequence of decoder instructions from decoder
+   to encoder.
+
+<!-- s/exactly/no more than/  ? -->
+HTTP/3 endpoints contain a QPACK encoder and decoder. Each endpoint MUST
```
afrind: I think I'll change this to MAY in a subsequent PR

draft-ietf-quic-qpack.md
afrind: I'm OK with the text here.

PR #2445, opened 2019-02-09
quic-tls nits

martinduke:
I suspect these are uncontroversial.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -14,7 +14,7 @@ pi: [toc, sortrefs, symrefs, docmapping]
 author:
   -
     ins: M. Thomson
-    name: Martin Thomsonf
+    name: Martin Thomson
     org: Mozilla
     email: mt@lowentropy.net
     role: editor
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -788,7 +788,7 @@ The value of initial_salt is a 20 byte sequence shown in the figure in
 hexadecimal notation. Future versions of QUIC SHOULD generate a new salt value,
 thus ensuring that the keys are different for each version of QUIC. This
 prevents a middlebox that only recognizes one version of QUIC from seeing or
-modifying the contents of handshake packets from future versions.
+modifying the contents of Initial packets from future versions.
```
marten-seemann: Initial is a version-dependent concept.

martinduke: but the concept of the version 1 salt is directly related to Initial packets...

marten-seemann: True, but you still can't talk about "Initial packets from future versions". Future version might not even have Initial packets.

martinthomson: @marten-seemann's point is right.  The intent here is to use a generic statement (packets that are part of the handshake process), as opposed to identifying a specific packet type in this version.

```suggestion
modifying the contents of packets from future versions.
```

martinduke:
All three Mart[ie]ns now agree, no further objections allowed.

PR #2446, opened 2019-02-09
Minor TLS draft editorial corrections

martinduke:
I am not quite as sure of these, but I believe they match the intent better.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -500,28 +500,27 @@ Client                                                    Server
 
 Get Handshake
                      Initial ------------->
-Rekey tx to 0-RTT Keys
+Install tx 0-RTT Keys
                      0-RTT --------------->
                                               Handshake Received
                                                    Get Handshake
                      <------------- Initial
-                                          Rekey rx to 0-RTT keys
-                                      Rekey tx to Handshake keys
+                                           Install rx 0-RTT keys
+                                          Install Handshake keys
                                                    Get Handshake
                      <----------- Handshake
-                                      Rekey rx to Handshake keys
-                                          Rekey tx to 1-RTT keys
+                                           Install tx 1-RTT keys
                      <--------------- 1-RTT
 Handshake Received
-Rekey tx and rx to Handshake keys
+Install tx Handshake keys
 Handshake Received
 Get Handshake
 Handshake Complete
                      Handshake ----------->
-Rekey tx and rx to 1-RTT keys
+Install 1-RTT keys
                      1-RTT --------------->
                                               Handshake Received
-                                          Rekey rx to 1-RTT keys
+                                           Install rx 1-RTT keys
                                               Handshake Complete
                                                    Get Handshake
                      <--------------- 1-RTT
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -216,10 +216,12 @@ Note that this omits the EndOfEarlyData message, which is not used in QUIC (see
 
 Data is protected using a number of encryption levels:
 
-- Plaintext
-- Early Data (0-RTT) Keys
-- Handshake Keys
-- Application Data (1-RTT) Keys
+- Initial Keys can be derived by any observer, and so they do not
+  provide cryptographic protection or authentication.
+- Early Data (0-RTT) Keys. These keys are not forward-secure and must protect
+  only idempotent data.
+- Handshake Keys do not authenticate either endpoint.
+- Application Data (1-RTT) Keys provide full authentication and encryption.
 
 Application data may appear only in the early data and application data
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -216,10 +216,12 @@ Note that this omits the EndOfEarlyData message, which is not used in QUIC (see
...
-- Plaintext
-- Early Data (0-RTT) Keys
-- Handshake Keys
-- Application Data (1-RTT) Keys
+- Initial Keys can be derived by any observer, and so they do not
+  provide cryptographic protection or authentication.
+- Early Data (0-RTT) Keys. These keys are not forward-secure and must protect
+  only idempotent data.
+- Handshake Keys do not authenticate either endpoint.
+- Application Data (1-RTT) Keys provide full authentication and encryption.
```
mikkelfj: I think this looses to overview of what levels exist. This bullet list now replaces an overview with an explanation. Perhaps keep the original list, and add an explanation below.

draft-ietf-quic-tls.md
```diff
@@ -269,7 +271,7 @@ At a high level, there are two main interactions between the TLS and QUIC
 components:
 
 * The TLS component sends and receives messages via the QUIC component, with
-  QUIC providing a reliable stream abstraction to TLS.
+  QUIC providing a reliable stream and record abstraction to TLS.
```
mikkelfj: I'm not sure QUIC provides the record abstraction, at least not entirely, it only handles the record protection. TLS still does its own internal record framing as I understand it, but don't take my word for it.

martinduke: This is the big draft-14 change, is it not?

mikkelfj: Well, not sure about 14, possibly, but the big change was to remove TLS encryption layer piped through already encrypted QUIC. But as I understand it, TLS still maintains its own record structure and framing within the QUIC crypto stream. So in a sense it both is and isn't TLS records.

martinthomson: The separation of CRYPTO into different spaces doesn't really result in a record-like abstraction.

draft-ietf-quic-tls.md
```diff
@@ -505,26 +508,24 @@ Rekey tx to 0-RTT Keys
...
 Handshake Received
 Get Handshake
 Handshake Complete
                      Handshake ----------->
-Rekey tx to 1-RTT keys
+Rekey tx and rx to 1-RTT keys
                      1-RTT --------------->
                                               Handshake Received
                                           Rekey rx to 1-RTT keys
-                                                   Get Handshake
```
martinthomson: This bit (and the next) need to remain for NewSessionTicket.

Do you think that it would help to identify which TLS messages are being sent and received here?

draft-ietf-quic-tls.md
martinthomson: I tend to agree with @mikkelfj here.  The short names are easier to get at without the extra text.

draft-ietf-quic-tls.md
martinduke: I would have said no, except for the bit about the NST. I might file another PR about that. The kind of ugly thing about this diagram is rekeying the rx from 0RTT to Handshake. There's no actual trigger for this; in practice we're actually keeping both keys, of course.

draft-ietf-quic-tls.md
martinduke: OK, this is going away, except for s/Plaintext/Initial

draft-ietf-quic-tls.md
martinduke: It's not the separate spaces, it's the crypto frame itself. We're stripping out the TLS record layer and replacing it with the frame. I'm not sure how you can compare the figures at the start of Sections 2 and 3 and come to a different conclusion.

draft-ietf-quic-tls.md
MikeBishop: Perhaps it would be more appropriate to say "Install Handshake rx keys," then?  We have the separate PR forthcoming about when keys from previous levels are dropped; perhaps in this diagram we could exclusively talk about installing them.

draft-ietf-quic-tls.md
martinduke: I don't want to hold up this PR forever on this point, but do @mt and @mikkelfj still disagree that QUIC is providing a record abstraction?

martinthomson: I think that this would be a mistake.  TLS only requires a stream abstraction.  QUIC has its own need for a record-like abstraction (we call them packets), but that is not relevant here.

If you revert this, I can merge it.

(@martinduke, be careful who you are tagging...)

martinthomson:
Ahh, the sweet taste of surrender.  Thanks Martin.

PR #2449, opened 2019-02-11
remove unused variable largest_sent_packet

marten-seemann:


martinthomson:
Oops, I just meant to approve that.  Need more sleep.  Sorry @ianswett, @janaiyengar; back out if you think this is a bad idea.

ianswett:
I think it's ok to remove this.  It might have been used for early retransmit before, but now that's part of time threshold loss detection.

Issue #2398, opened 2019-01-31
Specify DCID of 0-RTT packets


mikkelfj:
This has been discussed before, but after reorg. of text, I can't find any clear statement in Transport Draft about how to create DCID for 0-RTT. The DCID field is not mentioned for 0-RTT packets in transport draft.

https://github.com/quicwg/base-drafts/issues/1513
https://github.com/quicwg/base-drafts/issues/1721

Note: I got the think about this in the context of replay attack prevention which would be simpler of the 0-RTT DCID is not random. But that discussion was already closed in the above issue.

martinthomson:
We previously discussed this and the conclusion was to change to the server-selected connection ID once a packet has been received from the server.  Until then, the random value would be used.  If this is missing, then we just need some text in the 0-RTT packet section to cover this.  This probably just needs to reference [Section 7.2](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#negotiating-connection-ids), which says:

>  Upon receiving a packet, each endpoint sets the Destination Connection ID it sends to match the value of the Source Connection ID that they receive.

mikkelfj:
Based on proposed 0-RTT routing to single instance for replay protection, discussed on the mailing list, it would be helpful if the server can assign a DCID for use in 0-RTT that is potentially longer than normal CID's. This is because CID routing lifetime may shorter than 0-RTT session tickets, and a longer DCID could make it easier for a server farm to route 0-RTT packets correctly.

MikeBishop:
Are you proposing that the client should get a CID along with the NewSessionTicket (or the NEW_TOKEN)?  That's been discussed before, and while it makes sense to me, it means that load balancers need to be able to differentiate the two types (and potentially sizes) of CID.

On the whole, I think this is a promising direction, but perhaps not needed for v1.

mikkelfj:
Yes, along with the session ticket. The ticket transfer is a bit of a black box so I can only propose the high level concept. (I'm wondering if Retry and 0-RTT could be made more similar, but whatever).

> That's been discussed before, and while it makes sense to me, it means that load balancers need to be able to differentiate the two types (and potentially sizes) of CID.

At least such that they have an option to do so if they are so inclined. I.e. there is no immediate downside since you can also just provide a standard CID if that is better for a particular infrastructure.

[...]

PR #2450, opened 2019-02-11
Connection IDs for 0-RTT packets

martinthomson:
For #2398 and @mikkelfj.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1355,6 +1355,9 @@ long enough to fulfill the minimum size for every QUIC version it supports.
 The client populates the Source Connection ID field with a value of its choosing
 and sets the SCIL field to match.
 
+The same Destination and Source Connection ID values are used for any 0-RTT
+packets the client sends.
+
 The Destination Connection ID field in the server's Initial packet contains a
 connection ID that is chosen by the recipient of the packet (i.e., the client);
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1355,6 +1355,9 @@ long enough to fulfill the minimum size for every QUIC version it supports.
 The client populates the Source Connection ID field with a value of its choosing
 and sets the SCIL field to match.
 
+The same Destination and Source Connection ID values are used for any 0-RTT
+packets the client sends.
+
```
mikkelfj: This could be read as if the ID is the same in both directions.

mikkelfj:
Thanks

draft-ietf-quic-transport.md
ianswett: Agreed, this is not completely clear.  How about: "The first flight of 0-RTT packets use the same Destination and Source Connection ID values as the client's first Initial."?

Also, on 1356, "sets the SCIL field to match." is not very clear.  I'd either drop the clause or say "and sets the SCIL field to indicate the length."

mikkelfj:
Note that section 17.2.3. 0-RTT does not mention 0-RTT CID but the following section 7.2.4. Handshake Packet does mention Handshake CID's.

PR #2451, opened 2019-02-11
Add a loss_time per packet number space

### LINKED ISSUE #2435: loss recovery of crypto packets is less aggressive, not more aggressive
We claim that 
> Data in CRYPTO frames is critical to QUIC transport and crypto negotiation, so a more aggressive timeout is used to retransmit it.

As pointed out in #2417, the only mechanisms we have for retransmitting crypto packets is Fast Retransmit (i.e. when an ACK declares a packet lost) and the Crypto Retransmission timer. Specifically, we don't have an Early Retransmit timer for crypto packets. This means that if a tail packet in one packet number space is lost, we wait 2 RTTs before retransmitting it, whereas with Early Retransmit the retransmission would have happened after 9/8 RTT.

ianswett:
Fixes #2435 

I failed to fully think this through when #2417 landed, but the intent was always that time threshold loss detection apply to Initial and Handshake packet number spaces, as well as ApplicationData.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1040,7 +1040,7 @@ SetLossDetectionTimer():
     loss_detection_timer.cancel()
     return
 
-  [loss_time, pn_space] = GetEarliestLossTimer()
+  [loss_time, pn_space] = GetEarliestLossTime()
   if (loss_time != 0):
     // Time threshold loss detection.
     loss_detection_timer.update(loss_time)
@@ -1078,7 +1078,7 @@ Pseudocode for OnLossDetectionTimeout follows:
 
 ~~~
 OnLossDetectionTimeout():
-  [loss_time, pn_space] = GetEarliestLossTimer()
+  [loss_time, pn_space] = GetEarliestLossTime()
   if (loss_time != 0):
     // Time threshold loss Detection
     DetectLostPackets(pn_space)
@@ -1130,7 +1130,7 @@ DetectLostPackets(pn_space):
       sent_packets.remove(unacked.packet_number)
       if (unacked.in_flight):
         lost_packets.insert(unacked)
-    else if (pn_space == ApplicationData):
+    else:
       if (loss_time[pn_space] == 0):
         loss_time[pn_space] = unacked.time_sent + loss_delay
       else:
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1023,7 +1023,7 @@ Pseudocode for SetLossDetectionTimer follows:
 ~~~
 // Returns the earliest loss_time and the packet number
 // space it's from.  Returns 0 if all times are 0.
-GetEarliestLossTimer():
+GetEarliestLossTime():
   time = loss_time[Initial]
   space = Initial
   for pn_space in [ Handshake, ApplicatonData ]:
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1040,7 +1040,7 @@ SetLossDetectionTimer():
     loss_detection_timer.cancel()
     return
 
-  [loss_time, pn_space] = GetEarliestLossTime()
+  loss_time, _ = GetEarliestLossTime()
   if (loss_time != 0):
     // Time threshold loss detection.
     loss_detection_timer.update(loss_time)
@@ -1078,7 +1078,7 @@ Pseudocode for OnLossDetectionTimeout follows:
 
 ~~~
 OnLossDetectionTimeout():
-  [loss_time, pn_space] = GetEarliestLossTime()
+  loss_time, pn_space = GetEarliestLossTime()
   if (loss_time != 0):
     // Time threshold loss Detection
     DetectLostPackets(pn_space)
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -873,7 +873,7 @@ max_ack_delay:
   received ACK frame may be larger due to late timers, reordering,
   or lost ACKs.
 
-loss_times[kPacketNumberSpace]:
+loss_time[kPacketNumberSpace]:
 : The time at which the next packet in that packet number space will be
   considered lost based on exceeding the reordering window in time.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -1029,6 +1028,12 @@ SetLossDetectionTimer():
     loss_detection_timer.cancel()
     return
 
+  for (loss_time : loss_times) {
```
marten-seemann: I don't think this guarantees that you're picking the earliest timer. We need something like 
```
min_loss_time = min_non_zero(loss_time[Initial], loss_time[Handshake], loss_time[ApplicationData])
if (min_loss_time != 0):
    loss_detection_timer.update(min_loss_time)
    return
```

draft-ietf-quic-recovery.md
```diff
@@ -1117,11 +1115,11 @@ DetectLostPackets(pn_space):
       if (unacked.in_flight):
         lost_packets.insert(unacked)
     else if (pn_space == ApplicationData):
```
marten-seemann: Just `else`, no `else if`.

draft-ietf-quic-recovery.md
```diff
@@ -873,10 +873,9 @@ max_ack_delay:
   received ACK frame may be larger due to late timers, reordering,
   or lost ACKs.
 
-loss_time:
-: The time at which the next packet will be considered lost based on
-  exceeding the reordering window in time. Only applies to the ApplicationData
-  packet number space.
+loss_times[kPacketNumberSpace]:
```
marten-seemann: Nit: `loss_time` would be more consistent, since we also use `largest_acked_packet[kPacketNumberSpace]`, not `largest_acked_packets[kPacketNumberSpace]`.

draft-ietf-quic-recovery.md
```diff
@@ -1067,14 +1066,18 @@ Pseudocode for OnLossDetectionTimeout follows:
 
 ~~~
 OnLossDetectionTimeout():
-  if (crypto packets are in flight):
+  packets_lost = false
+  for pn_space in [ Initial, Handshake, ApplicatonData ]:
+    if (loss_times[pn_space] != 0):
+      // Time threshold loss Detection
+      DetectLostPackets(pn_space)
```
marten-seemann: It seems a bit wasteful to run this for every packet number space that has a non-zero `loss_time`, since we'll only detect lost packets in the space that has the earliest `loss_time`.

draft-ietf-quic-recovery.md
```diff
@@ -1067,14 +1066,18 @@ Pseudocode for OnLossDetectionTimeout follows:
...
+      packets_lost = true
+  // Don't retransmit all crypto data if a packet was just lost.
+  if (!packets_lost &&
```
marten-seemann: If `packets_lost = true`, we'd send PTO packets. I don't think this is the intention. We should return early in that case.

ianswett: I added GetEarliestLossTime, which I think makes this more efficient and correct.

draft-ietf-quic-recovery.md
ianswett: Added GetEarliestLossTime(), PTAL

draft-ietf-quic-recovery.md
```diff
@@ -1022,13 +1021,31 @@ timers wake up late. Timers set in the past SHOULD fire immediately.
...
+// Returns the earliest loss_time and the packet number
+// space it's from.  Returns 0 if all times are 0.
+GetEarliestLossTime():
+  time = loss_time[Initial]
+  space = Initial
+  for pn_space in [ Handshake, ApplicatonData ]:
+    if loss_time[pn_space] != 0 &&
+       (time == 0 || loss_time[pn_space] < time):
+      time = loss_time[pn_space];
+      space = Initial
```
marten-seemann: `space = pn_space`

draft-ietf-quic-recovery.md
```diff
@@ -1022,13 +1021,31 @@ timers wake up late. Timers set in the past SHOULD fire immediately.
...
+  return [time, space]
+
 SetLossDetectionTimer():
   // Don't arm timer if there are no ack-eliciting packets
   // in flight.
   if (no ack-eliciting packets in flight):
     loss_detection_timer.cancel()
     return
 
+  [loss_time, pn_space] = GetEarliestLossTime()
```
marten-seemann: In Python, you don't need square brackets to receive multiple return values.

And since we're not using the the `pn_space` here, we could write: `loss_time, _ = GetEarliestLossTime()`

draft-ietf-quic-recovery.md
```diff
@@ -1067,14 +1078,15 @@ Pseudocode for OnLossDetectionTimeout follows:
 
 ~~~
 OnLossDetectionTimeout():
-  if (crypto packets are in flight):
+  [loss_time, pn_space] = GetEarliestLossTime()
```
marten-seemann: Same here.

draft-ietf-quic-recovery.md
```diff
@@ -1067,14 +1078,15 @@ Pseudocode for OnLossDetectionTimeout follows:
...
+  if (loss_time != 0):
+    // Time threshold loss Detection
+    DetectLostPackets(pn_space)
+  // Don't retransmit all crypto data if a packet was just lost.
```
marten-seemann: This comment only makes sense for someone who followed the evolution of this algorithm. We should instead have a comment that describes when to retransmit all crypto data.

draft-ietf-quic-recovery.md
```diff
@@ -1022,13 +1021,31 @@ timers wake up late. Timers set in the past SHOULD fire immediately.
...
+// space it's from.  Returns 0 if all times are 0.
+GetEarliestLossTime():
+  time = loss_time[Initial]
+  space = Initial
+  for pn_space in [ Handshake, ApplicatonData ]:
+    if loss_time[pn_space] != 0 &&
+       (time == 0 || loss_time[pn_space] < time):
+      time = loss_time[pn_space];
+      space = pn_space
+  return [time, space]
```
marten-seemann: In Python you don’t need the brackets here. Sorry for missing this on my earlier review.

ianswett: Thanks.

PR #2455, opened 2019-02-12
describe what happens when the PTO timer expires

marten-seemann:


draft-ietf-quic-recovery.md
```diff
@@ -506,9 +506,10 @@ delay sending an acknowledgement.
 The PTO value MUST be set to at least kGranularity, to avoid the timer expiring
 immediately.
 
-When a PTO timer expires, the PTO period MUST be set to twice its current value.
-This exponential reduction in the sender's rate is important because the PTOs
-might be caused by loss of packets or acknowledgements due to severe congestion.
+When a PTO timer expires, the sender sends a probe packet as described in the
+next section. The PTO period MUST be set to twice its current value. This
```
```suggestion
next section. The PTO period then MUST be set to twice its current value. This
```

draft-ietf-quic-recovery.md
```diff
@@ -506,9 +506,10 @@ delay sending an acknowledgement.
 The PTO value MUST be set to at least kGranularity, to avoid the timer expiring
 immediately.
 
-When a PTO timer expires, the PTO period MUST be set to twice its current value.
-This exponential reduction in the sender's rate is important because the PTOs
-might be caused by loss of packets or acknowledgements due to severe congestion.
+When a PTO timer expires, the sender sends a probe packet as described in the
```
ianswett: It could be two probe packets, so maybe "sends one or two probe packets"?

nibanks: Or simply "probes the network"?

martinthomson: Or simply nothing.  A probe packet is sent if two packets are sent, and this is not normative text.

marten-seemann: I like @nibanks's suggestion. It's independent of how the probing is actually done, which we describe in detail later.

draft-ietf-quic-recovery.md
```diff
@@ -506,9 +506,10 @@ delay sending an acknowledgement.
 The PTO value MUST be set to at least kGranularity, to avoid the timer expiring
 immediately.
 
-When a PTO timer expires, the PTO period MUST be set to twice its current value.
-This exponential reduction in the sender's rate is important because the PTOs
-might be caused by loss of packets or acknowledgements due to severe congestion.
+When a PTO timer expires, the sender probes the network as described in the next
+section. The PTO period MUST be set to twice its current value. This exponential
+reduction in the sender's rate is important because the PTOs might be caused by
```
```suggestion
backoff in the sender's rate is important because the PTOs might be caused by
```

marten-seemann: Out of scope for this PR.

PR #2456, opened 2019-02-12
the crypto timeout is based on the last crypto packet sent

marten-seemann:
From reading the text, it was not clear that the timer is set based on the last (and not the first) crypto packet sent.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -412,9 +412,10 @@ is available, or if the network changes, the initial RTT SHOULD be set to 100ms.
 When an acknowledgement is received, a new RTT is computed and the timer
 SHOULD be set for twice the newly computed smoothed RTT.
 
-Every time a crypto packet is sent, the sender MUST set a timer for the crypto
-timeout period.  Upon timeout, the sender MUST retransmit all unacknowledged
-CRYPTO data if possible.
+When a crypto packet is sent, the sender MUST set a timer for the crypto
+timeout period.  This timer MUST be updated when a new crypto packet is sent.
+Upon timeout, the sender MUST retransmit all unacknowledged CRYPTO data if
+possible.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -412,7 +412,7 @@ is available, or if the network changes, the initial RTT SHOULD be set to 100ms.
 When an acknowledgement is received, a new RTT is computed and the timer
 SHOULD be set for twice the newly computed smoothed RTT.
 
-When crypto packets are sent, the sender MUST set a timer for the crypto
+Every time a crypto packets is sent, the sender MUST set a timer for the crypto
```
ianswett: I'm not sure this is significantly clearer, but I do understand what you're trying to address.

```suggestion
Every time a crypto packet is sent, the sender MUST set a timer for the crypto
```

draft-ietf-quic-recovery.md
```diff
@@ -412,7 +412,7 @@ is available, or if the network changes, the initial RTT SHOULD be set to 100ms.
 When an acknowledgement is received, a new RTT is computed and the timer
 SHOULD be set for twice the newly computed smoothed RTT.
 
-When crypto packets are sent, the sender MUST set a timer for the crypto
+Every time a crypto packet is sent, the sender MUST set a timer for the crypto
```
ianswett: How about "A timer MUST be updated to expire after the crypto timeout period after a new crypto packet is sent"?

mikkelfj: > If no previous RTT is available,
-->
If no previous RTT, or a-priori estimate, is available,

ianswett: @mikkelfj is this suggestion about a different paragraph?

mikkelfj: yes, I should have been more clear. It is existing text in the preceding paragraph, or thereabout.

PR #2457, opened 2019-02-12
Rework GOAWAY text (again)

### LINKED ISSUE #2226: Why do I have to explicitly cancel after GOAWAY
Why do I have to explicitly cancel after GOAWAY?

```
Once GOAWAY is sent, the server MUST cancel requests sent on streams with an
identifier higher than the indicated last Stream ID.  Clients MUST NOT send new
requests on the connection after receiving GOAWAY, although requests might
already be in transit. A new connection can be established for new requests.

If the client has sent requests on streams with a higher Stream ID than
indicated in the GOAWAY frame, those requests are considered cancelled
({{request-cancellation}}).  Clients SHOULD reset any streams above this ID with
the error code HTTP_REQUEST_CANCELLED.  Servers MAY also cancel requests on
streams below the indicated ID if these requests were not processed.
```

Given that GOAWAY already tells the client this, what's the point?

MikeBishop:
Another attempt to group related concepts in this section which hopefully fixes #2226.

(This is technically design, because previously-required server behavior is now only recommended.)

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1259,8 +1259,8 @@ identified by a QUIC MAX_STREAM_ID frame, and MAY be zero if no requests were
 processed.  Servers SHOULD NOT increase the QUIC MAX_STREAM_ID limit after
 sending a GOAWAY frame.
 
-Clients MUST NOT send new requests on the connection after receiving GOAWAY,
-although a new connection MAY be established to send additional requests.
+Clients MUST NOT send new requests on the connection after receiving GOAWAY;
+a new connection MAY be established to send additional requests.
 
 Some requests might already be in transit. If the client has already sent
 requests on streams with a Stream ID greater than or equal to that indicated in
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1270,10 +1270,10 @@ client.  It is RECOMMENDED that the server explicitly reject such requests (see
 streams.  The client MAY also cancel requests which will not be processed.
 
 Requests on Stream IDs less than the Stream ID in the GOAWAY frame might have
-been processed; their status cannot be known until they are completed
-successfully, reset individually, or the connection terminates.  Servers MAY
-reject individual requests on streams below the indicated ID if these requests
-were not processed.
+been processed; their status cannot be known until a response is received, the
+stream is reset individually, or the connection terminates.  Servers MAY reject
+individual requests on streams below the indicated ID if these requests were not
+processed.
 
 Servers SHOULD send a GOAWAY frame when the closing of a connection is known
 in advance, even if the advance notice is small, so that the remote peer can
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -1259,21 +1259,21 @@ identified by a QUIC MAX_STREAM_ID frame, and MAY be zero if no requests were
...
-requests might already be in transit. A new connection can be established for
-new requests.
-
-If the client has sent requests on streams with a Stream ID greater than or
-equal to that indicated in the GOAWAY frame, those requests are considered
-rejected ({{request-cancellation}}).  Clients SHOULD cancel any requests on
-streams above this ID.  Servers MAY also reject requests on streams below the
-indicated ID if these requests were not processed.
+Clients MUST NOT send new requests on the connection after receiving GOAWAY,
+although a new connection MAY be established to send additional requests.
```
martinthomson: I don't think that this is an "although" case.  Maybe a separate sentence.

draft-ietf-quic-http.md
```diff
@@ -1270,10 +1270,10 @@ client.  It is RECOMMENDED that the server explicitly reject such requests (see
 streams.  The client MAY also cancel requests which will not be processed.
 
 Requests on Stream IDs less than the Stream ID in the GOAWAY frame might have
-been processed; their status cannot be known until they are completed
-successfully, reset individually, or the connection terminates.  Servers MAY
-reject individual requests on streams below the indicated ID if these requests
-were not processed.
+been processed; their status cannot be known until a response is received, the
+stream is reset individually, or the connection terminates.  Servers MAY reject
```
ianswett: Does reject == reset here?

MikeBishop: Reject and cancel are defined (cross-ref in previous paragraph) as reset with particular error codes.  If that's not clear, I'm open to proposed wordsmithing.

draft-ietf-quic-http.md
```diff
@@ -1259,21 +1259,21 @@ identified by a QUIC MAX_STREAM_ID frame, and MAY be zero if no requests were
...
-equal to that indicated in the GOAWAY frame, those requests are considered
-rejected ({{request-cancellation}}).  Clients SHOULD cancel any requests on
-streams above this ID.  Servers MAY also reject requests on streams below the
-indicated ID if these requests were not processed.
+Clients MUST NOT send new requests on the connection after receiving GOAWAY;
+a new connection MAY be established to send additional requests.
+
+Some requests might already be in transit. If the client has already sent
+requests on streams with a Stream ID greater than or equal to that indicated in
+the GOAWAY frame, those requests will not be processed and can be retried by the
```
```suggestion
the GOAWAY frame, those requests will not be processed and MAY be retried by the client on a new connection
```

draft-ietf-quic-http.md
```diff
@@ -1259,21 +1259,21 @@ identified by a QUIC MAX_STREAM_ID frame, and MAY be zero if no requests were
...
+client.  It is RECOMMENDED that the server explicitly reject such requests (see
+{{request-cancellation}}) in order to clean up transport state for the affected
+streams.  The client MAY also cancel requests which will not be processed.
```
janaiyengar: Suggestion: Swap the order of these last two sentences, so that client behavior is described first and then server. Also, the following:
```suggestion
streams.  The client MAY also cancel these requests.
```

PR #2463, opened 2019-02-13
Fix packet and stream numbers in example

martinduke:
The choices were odd for a "typical" example


martinthomson:
Ugh, this is going to conflict with my other PR now..

PR #2467, opened 2019-02-14
Refine Discussion of 0-RTT Transport Parameters

### LINKED ISSUE #2464: Remember Fewer Transport Parameters for 0-RTT
Sec 7.3.1 of quic-transport says: 
A client that attempts to send 0-RTT data MUST remember the transport parameters used by the server. The transport parameters that the server advertises during connection establishment apply to all connections that are resumed using the keying material established during that handshake. Remembered transport parameters apply to the new connection until the handshake completes and new transport parameters from the server can be provided.

Later on, the section makes it clear that this is really about the flow control parameters, which makes a lot of sense, but here I am storing all this data to be compliant.

I propose that we change the MUST above to refer to only the 6 FC parameters. This reduces the stuff that clients have to store.

going through each:
original_connection_id: clearly useless
idle_timeout: useless in handshake
stateless_reset_token: clearly useless
max_packet_size: could be useful, but meh
ack_delay_exponent: not useful before we get the server Handshake
max_ack_delay: same
disable_migration: same
preferred_address: same

martinduke:
Fixes #2464, but there are some value judgments here. I believe the new layout is much clearer for implementers.

I made max_packet_size optional for reasons that I think are self-evident. I thought saving max_ack_delay might be useful for clients setting loss timers on their 0-RTT stuff. Everything else is required or forbidden.

Per @MikeBishop's suggestion, I put all the new TPs in the required category.

I also specified behavior when the EE msg omits a TP. I thought it would be simpler for servers that spit out a standard set of TPs to do things that way.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1378,28 +1378,38 @@ A server MUST include the original_connection_id transport parameter
 ({{transport-parameter-definitions}}) if it sent a Retry packet to enable
 validation of the Retry, as described in {{packet-retry}}.
 
-
 ### Values of Transport Parameters for 0-RTT {#zerortt-parameters}
 
-A client that attempts to send 0-RTT data MUST remember the transport parameters
-used by the server.  The transport parameters that the server advertises during
-connection establishment apply to all connections that are resumed using the
-keying material established during that handshake.  Remembered transport
+The transport parameters that the server advertises during connection
+establishment generally apply to all connections that are resumed using
+the keying material established during that handshake.  Remembered transport
 parameters apply to the new connection until the handshake completes and new
 transport parameters from the server can be provided.
 
-A server can remember the transport parameters that it advertised, or store an
-integrity-protected copy of the values in the ticket and recover the information
-when accepting 0-RTT data.  A server uses the transport parameters in
-determining whether to accept 0-RTT data.
+The value of the server's previous original_connection_id, preferred_address,
+stateless_reset_token, and ack_delay_exponent MUST NOT be used when
+establishing a new connection; rather, the client should wait to observe the
+server's new values in the handshake.
 
-A server MAY accept 0-RTT and subsequently provide different values for
-transport parameters for use in the new connection.  If 0-RTT data is accepted
-by the server, the server MUST NOT reduce any limits or alter any values that
-might be violated by the client with its 0-RTT data.  In particular, a server
-that accepts 0-RTT data MUST NOT set values for the following parameters
-({{transport-parameter-definitions}}) that are smaller
-than the remembered value of those parameters.
+The client MAY store the server's original max_ack_delay and max_packet_size
+transport parameters. The server MAY accept 0-RTT and subsequently provide
+different values for these transport parameters for use in the new connection.
+The client MUST use these values after acknowledging them. If the server does
+not provide a new value, the client MUST use the default value.
+
+A client that attempts to send 0-RTT data MUST remember all other transport
+parameters used by the server, including new transport parameters
+({{new-transport-parameters}}) that it is able to process. The server can
+remember these transport parameters, or store an integrity-protected copy of
+the values in the ticket and recover the information when accepting 0-RTT
+data. A server uses the transport parameters in determining whether to accept
+0-RTT data.
+
+If 0-RTT data is accepted by the server, the server MUST NOT reduce any
+limits or alter any values that might be violated by the client with its
+0-RTT data.  In particular, a server that accepts 0-RTT data MUST NOT set
+values for the following parameters ({{transport-parameter-definitions}})
+that are smaller than the remembered value of those parameters.
 
 * initial_max_data
 * initial_max_stream_data_bidi_local
@@ -1415,15 +1425,10 @@ values for 0-RTT.  This includes initial_max_data and either
 initial_max_streams_bidi and initial_max_stream_data_bidi_remote, or
 initial_max_streams_uni and initial_max_stream_data_uni.
 
-The value of the server's previous preferred_address MUST NOT be used when
-establishing a new connection; rather, the client should wait to observe the
-server's new preferred_address value in the handshake.
-
 A server MUST either reject 0-RTT data or abort a handshake if the implied
 values for transport parameters cannot be supported.
 
-
-### New Transport Parameters
+### New Transport Parameters {#new-transport-parameters}
 
 New transport parameters can be used to negotiate new protocol behavior.  An
 endpoint MUST ignore transport parameters that it does not support.  Absence of


draft-ietf-quic-transport.md
```diff
@@ -1378,28 +1378,38 @@ A server MUST include the original_connection_id transport parameter
...
-than the remembered value of those parameters.
+The value of the server's previous original_connection_id, preferred_address,
+stateless_reset_token, and ack_delay_exponent MUST NOT be used when
+establishing a new connection; rather, the client should wait to observe the
+server's new values in the handshake.
+
+The client MAY store the server's original max_ack_delay and max_packet_size
+transport parameters. The server MAY accept 0-RTT and subsequently provide
+different values for these transport parameters for use in the new connection.
+The client MUST use these values after acknowledging them. If the server does
```
```suggestion
The client MUST use these values after the handshake completes. If the server does
```
...modulo any rewrapping needed from line-length changes.

draft-ietf-quic-transport.md
```diff
@@ -1378,28 +1378,38 @@ A server MUST include the original_connection_id transport parameter
...
+not provide a new value, the client MUST use the default value.
+
+A client that attempts to send 0-RTT data MUST remember all other transport
+parameters used by the server, including new transport parameters
+({{new-transport-parameters}}) that it is able to process. The server can
+remember these transport parameters, or store an integrity-protected copy of
+the values in the ticket and recover the information when accepting 0-RTT
+data. A server uses the transport parameters in determining whether to accept
+0-RTT data.
```
MikeBishop: I think this comment about the server remembering needs to be paired with the prohibition on the server reducing values.  Since you've split out the "in particular" ones already, it might be reasonable to say that the server must not reduce limits / alter values in the transport parameters the client is required to remember.

This is especially true if we go with Ian's suggestion to require future transport parameter definitions to specify which remembering category they fall in.

draft-ietf-quic-transport.md
```diff
@@ -1415,15 +1425,10 @@ values for 0-RTT.  This includes initial_max_data and either
...
-The value of the server's previous preferred_address MUST NOT be used when
-establishing a new connection; rather, the client should wait to observe the
-server's new preferred_address value in the handshake.
-
 A server MUST either reject 0-RTT data or abort a handshake if the implied
 values for transport parameters cannot be supported.
 
-
-### New Transport Parameters
+### New Transport Parameters {#new-transport-parameters}
```
MikeBishop: Technically unnecessary, since that's the auto-generated anchor anyway.  But it doesn't hurt anything.

draft-ietf-quic-transport.md
```diff
@@ -1378,28 +1378,38 @@ A server MUST include the original_connection_id transport parameter
...
 
-
 ### Values of Transport Parameters for 0-RTT {#zerortt-parameters}
 
-A client that attempts to send 0-RTT data MUST remember the transport parameters
-used by the server.  The transport parameters that the server advertises during
-connection establishment apply to all connections that are resumed using the
-keying material established during that handshake.  Remembered transport
+The transport parameters that the server advertises during connection
+establishment generally apply to all connections that are resumed using
```
martinthomson: "generally" is a little to squishy a word to use here.  I'd prefer to be crisp:

> The value of transport parameters from a connection are remembered and apply to any 0-RTT packets that are sent in subsequent connections that are resumed from that connection, except for those transport parameters that are explicitly excluded.  Remembered transport parameters apply to the new connection until the handshake completes and the client starts sending 1-RTT packets.  Once the handshake completes, the client uses the transport parameters established in the handshake.
>
> Any transport parameter that does not or cannot apply to 0-RTT packets does not need to be remembered.  This includes max_ack_delay, <list>.  Any transport parameter that applies to the connection as a whole does not need to be remembered.  This includes disable_migration, <list>.  Endpoints are not obligated to remember and use unknown transport parameters.

Then I think we can tabulate transport parameters and add a "0-RTT" column, indicating whether it applies to 0-RTT or not.

draft-ietf-quic-transport.md
```diff
@@ -1378,28 +1378,38 @@ A server MUST include the original_connection_id transport parameter
...
-might be violated by the client with its 0-RTT data.  In particular, a server
-that accepts 0-RTT data MUST NOT set values for the following parameters
-({{transport-parameter-definitions}}) that are smaller
-than the remembered value of those parameters.
+The value of the server's previous original_connection_id, preferred_address,
+stateless_reset_token, and ack_delay_exponent MUST NOT be used when
+establishing a new connection; rather, the client should wait to observe the
+server's new values in the handshake.
+
+The client MAY store the server's original max_ack_delay and max_packet_size
```
martinthomson: max_packet_size is important.  That is the one in addition to the flow control stuff that needs to be respected.

martinthomson: max_ack_delay is not needed for 0-RTT, though it can help in loss recovery, it's a value that can change.

martinduke: I figure the client always has the option of using 1200B packets until the handshake's done.

martinduke: The client may find it useful to retain this information to set its 0-RTT loss timer properly, rather than having to wait for the handshake and reset it properly. Am I not thinking clearly on how this is done?

draft-ietf-quic-transport.md
martinduke: I like this wording. I'll try go get to this in the next few days.

draft-ietf-quic-transport.md
martinthomson: The client can do that for Handshake and Initial, but not 1-RTT.  0-RTT should follow the same rules though.

martinthomson: No, you are right.  I think that what you have is fine for that one.

martinthomson: I realize now what bothered me about max_ack_delay.  This is not something you use in 0-RTT, but it is something you might remember in the same way you might remember the RTT.  It's not guaranteed to be good, but it is certainly better than the defaults for return connections.  You should remove this from this section and maybe add it where we talk about remembering the RTT (if we do that, it's late and I couldn't find it).

martinduke: Why don't you use it in 0RTT? You have to set the loss detection timer, which involves using peer max_ack_delay. If I read the recovery draft correctly?

I made it optional because I suppose someone could update the timer based on the new TPs. The more I think about it, better to just save it.

martinduke: I don't understand your comment. The safe thing to do is to send 0-RTT at 1280 until a new probe can be sent. I can also understand the opposite choice, which is why this is a MAY.

martinthomson: The point of max_packet_size is to limit the size of packets.  I realize that if you are making a new connection, the limit applies only to 1-RTT packets because Initial and Handshake packets are sent without knowing what the limits are (the server could comply with a request to send smaller Handshake packets, and maybe the client could also, but it's not something we can guarantee).  However, when sending 0-RTT, you know what the limit was, so why not respect it?

draft-ietf-quic-transport.md
```diff
@@ -1378,28 +1378,39 @@ A server MUST include the original_connection_id transport parameter
...
-might be violated by the client with its 0-RTT data.  In particular, a server
-that accepts 0-RTT data MUST NOT set values for the following parameters
-({{transport-parameter-definitions}}) that are smaller
-than the remembered value of those parameters.
+Clients remember the value of transport parameters from a server and apply
+them to any 0-RTT packets that are sent in subsequent connections, except
+for transport parameters that are explicitly excluded. Remembered transport
+parameters apply to the new connection until the handshake completes and the
+client starts sending 1-RTT packets. Once the handshake completes, the client
+uses the transport parameters established in thehandshake.
```
```suggestion
uses the transport parameters established in the handshake.
```

draft-ietf-quic-transport.md
```diff
@@ -1378,28 +1378,39 @@ A server MUST include the original_connection_id transport parameter
...
+
+The value of the server's previous original_connection_id, preferred_address,
+stateless_reset_token, and ack_delay_exponent MUST NOT be used when
+establishing a new connection; rather, the client should wait to observe the
+server's new values in the handshake. If the server does not provide a new
+value, the client MUST use the default value.
+
+The client MAY store the server's original max_packet_size transport
+parameter.
```
```suggestion
```

draft-ietf-quic-transport.md
```diff
@@ -1378,28 +1378,39 @@ A server MUST include the original_connection_id transport parameter
...
+client starts sending 1-RTT packets. Once the handshake completes, the client
+uses the transport parameters established in thehandshake.
+
+The value of the server's previous original_connection_id, preferred_address,
+stateless_reset_token, and ack_delay_exponent MUST NOT be used when
+establishing a new connection; rather, the client should wait to observe the
+server's new values in the handshake. If the server does not provide a new
+value, the client MUST use the default value.
+
+The client MAY store the server's original max_packet_size transport
```
```suggestion
```

draft-ietf-quic-transport.md
```diff
@@ -1378,28 +1378,39 @@ A server MUST include the original_connection_id transport parameter
...
+parameters used by the server. The server can remember these transport
+parameters, or store an integrity-protected copy of the values in the ticket
+and recover the information when accepting 0-RTT data. A server uses the
+transport parameters in determining whether to accept 0-RTT data.
+
+If 0-RTT data is accepted by the server, the server MUST NOT reduce any
+limits or alter any values that might be violated by the client with its
+0-RTT data.  In particular, a server that accepts 0-RTT data MUST NOT set
+values for the following parameters ({{transport-parameter-definitions}})
+that are smaller than the remembered value of those parameters.
```
```suggestion
that are smaller than the remembered value of those parameters:
```

draft-ietf-quic-transport.md
```diff
@@ -1378,28 +1378,39 @@ A server MUST include the original_connection_id transport parameter
...
-({{transport-parameter-definitions}}) that are smaller
-than the remembered value of those parameters.
+Clients remember the value of transport parameters from a server and apply
+them to any 0-RTT packets that are sent in subsequent connections, except
+for transport parameters that are explicitly excluded. Remembered transport
+parameters apply to the new connection until the handshake completes and the
+client starts sending 1-RTT packets. Once the handshake completes, the client
+uses the transport parameters established in thehandshake.
+
+The value of the server's previous original_connection_id, preferred_address,
```
martinthomson: I think that maybe this paragraph can be moved after the one starting with "The definition of"

janaiyengar: Editorial rewrite suggestion:  "A client MUST NOT use remembered values for the following parameters: original_connection_id, preferred_address, stateless_reset_token, ack_delay_exponent.  The client MUST use the server's new values in the handshake instead, and absent new values from the server, the default value."

janaiyengar:
@martinduke : Are you planning to incorporate feedback into the PR?

martinduke:
Yes, sorry for the delay.

draft-ietf-quic-transport.md
martinduke: OK, I'll make it a MUST

draft-ietf-quic-transport.md
martinduke: I took the first paragraph wholesale. I think "applies to the connection as a whole" is a confusing construction (all params apply affect the whole connection). Let me know what you think of the rewording.

martinthomson:
Thanks for doing this Martin.  I will raise this on the list so that people are clear about this.  Technically, there isn't a change here, but it certainly looks like it if you don't understand what is going on.

PR #2468, opened 2019-02-14
Address Validation Nits

martinduke:
Although not super-important, the biggest decision in here is that the packet number of the response to a RETRY is 1. Based on interop experience, I believe this is the semi-consensus of implementers.


DavidSchinazi:
Should there be normative text mandating that the packet number be increased?

mikkelfj:
> Should there be normative text mandating that the packet number be increased?

I think we need a new issue on that. The information about this is scattered around issuing CID's, path validation, retransmission of Initial. It would be good with a clear statement of when the packet number starts counting. I also believe that there is somewhere text to the effect that a CID cannot have the same PN twice, but that does not apply to Retry.

PR #2469, opened 2019-02-15
Fix inconsistent leading zeros

LPardue:
Simply making this consistent through the document by doing the minimal change and removing leading 0s. Happy to add them if we prefer.


MikeBishop:
Works for me.

PR #2476, opened 2019-02-15
Packetization Nits

martinduke:
The deleted parenthetical is not accurate, and is unnecessary: 0-RTT and 1-RTT with various key phases share a PN space without sharing keys.

I prefer the MUST NOT in coalescing, but if there are objections I can revert it.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2594,7 +2594,7 @@ MUST attempt to process the remaining packets.
 
 Retry packets ({{packet-retry}}), Version Negotiation packets
 ({{packet-version}}), and packets with a short header ({{short-header}}) do not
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2594,7 +2594,7 @@ MUST attempt to process the remaining packets.
 
 Retry packets ({{packet-retry}}), Version Negotiation packets
 ({{packet-version}}), and packets with a short header ({{short-header}}) do not
-contain a Length field and so cannot be followed by other packets in the same
+contain a Length field and so MUST NOT be followed by other packets in the same
```
martinthomson: I prefer cannot. It isn't an interoperability requirement that these not be followed by anything else, it is an outright impossibility.

martinthomson:
Thanks Martin.

PR #2483, opened 2019-02-20
Clarify what to do when ALPN negotiation fails

ghedo:
Just so you don't have to read the whole RFC7301.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1216,9 +1216,9 @@ negotiation.  TLS uses Application Layer Protocol Negotiation (ALPN)
 {{!RFC7301}} to select an application protocol.  Unless another mechanism is
 used for agreeing on an application protocol, endpoints MUST use ALPN for this
 purpose.  When using ALPN, endpoints MUST abort a connection if an application
-protocol is not negotiated with a no_application_protocol alert. While
-{{!RFC7301}} only specifies that servers use this alert, QUIC clients MUST also
-use it to terminate a connection when ALPN negotiation fails.
+protocol is not negotiated with a no_application_protocol alert (error code
+0x178). While {{!RFC7301}} only specifies that servers use this alert, QUIC
+clients MUST also use it to terminate a connection when ALPN negotiation fails.
 
 An application-layer protocol MAY restrict the QUIC versions that it can operate
 over.  Servers MUST select an application protocol compatible with the QUIC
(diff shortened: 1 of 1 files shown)


ghedo:
Yeah I was thinking about that. I wonder if it would make sense for the client to simply send PROTOCOL_VIOLATION instead of a TLS alert.

ghedo:
@martinthomson I added a clarification to allow clients to send no_application_negotiated as well. Should I send an email to the mailing list for discussion now that the new process is in effect?

martinthomson:
Unfortunately yes.  This is a design change so it will need an issue as well.

I like it and we should do it, but can you add the error code we would use in CONNECTION_CLOSE to the text?

ghedo:
@martinthomson ok, added error code, created issue (#2495) and sent email to the list.

draft-ietf-quic-tls.md
```diff
@@ -1216,7 +1216,9 @@ negotiation.  TLS uses Application Layer Protocol Negotiation (ALPN)
 {{!RFC7301}} to select an application protocol.  Unless another mechanism is
 used for agreeing on an application protocol, endpoints MUST use ALPN for this
 purpose.  When using ALPN, endpoints MUST abort a connection if an application
```
janaiyengar: This isn't within your change, but what  does "abort" mean. Should this be "drop QUIC connection state"?

martinthomson: I think that an immediate close is what was intended.

draft-ietf-quic-tls.md
```diff
@@ -1216,7 +1216,9 @@ negotiation.  TLS uses Application Layer Protocol Negotiation (ALPN)
...
-protocol is not negotiated.
+protocol is not negotiated with a no_application_protocol alert (error code
+0x178). While {{!RFC7301}} only specifies that servers use this alert, QUIC
```
```suggestion
0x178, see {{tls-errors}}).  While {{!RFC7301}} only specifies that servers use this alert, QUIC
```

draft-ietf-quic-tls.md
```diff
@@ -1216,7 +1216,9 @@ negotiation.  TLS uses Application Layer Protocol Negotiation (ALPN)
 {{!RFC7301}} to select an application protocol.  Unless another mechanism is
 used for agreeing on an application protocol, endpoints MUST use ALPN for this
 purpose.  When using ALPN, endpoints MUST abort a connection if an application
-protocol is not negotiated.
+protocol is not negotiated with a no_application_protocol alert (error code
```
```suggestion
protocol is not negotiated with a no_application_protocol TLS alert (QUIC error code
```

draft-ietf-quic-tls.md
janaiyengar:
```suggestion
purpose.  When using ALPN, endpoints MUST immediately close a connection (see Section 10.3 in {{QUIC-TRANSPORT}}) if an application
```

ghedo:
@janaiyengar thanks, suggestions accepted.

PR #2486, opened 2019-02-21
Don't close the connection if MAX_STREAMS is too large

### LINKED ISSUE #2487: Specify what occurs when the maximum Stream Id exceeds 2^62 - 1 in MAX_STREAMS
Currently, stream IDs are as high as 2^62-1.  The MAX_STREAMS frame specifies a limit as a varint that has a max value of 2^62-1 that translates to a Stream ID of 4 * max_streams + initial_stream_id_for_type.  This is larger than the max expressible stream ID, so we should specify what happens in this case.

I'm suggesting that we allow all streams to be opened and MUST NOT close the connection in PR #2486

ianswett:
Fixes #2487

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -849,9 +849,9 @@ initial_stream_id_for_type) can be opened (see {{long-packet-types}}).  Initial
 limits are set in the transport parameters (see
 {{transport-parameter-definitions}}) and subsequently limits are advertised
 using MAX_STREAMS frames ({{frame-max-streams}}). Separate limits apply to
-unidirectional and bidirectional streams.  If the maximum stream ID that can
-be opened is greater than 2^62 - 1, then the receiver MUST interpret it as no
-stream limit; this MUST NOT be treated as an error.
+unidirectional and bidirectional streams.  When a MAX_STREAMS frame permits
+more streams than can be expressed by the stream ID (i.e. greater than 2^60 - 1),
+all valid Stream IDs of that type can be used.
 
 Endpoints MUST NOT exceed the limit set by their peer.  An endpoint that
 receives a STREAM frame with a stream ID exceeding the limit it has sent MUST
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -849,7 +849,9 @@ initial_stream_id_for_type) can be opened (see {{long-packet-types}}).  Initial
 limits are set in the transport parameters (see
 {{transport-parameter-definitions}}) and subsequently limits are advertised
 using MAX_STREAMS frames ({{frame-max-streams}}). Separate limits apply to
-unidirectional and bidirectional streams.
+unidirectional and bidirectional streams.  If the maximum stream ID that can
+be opened is greater than 2^62 - 1, then the receiver MUST interpret it no
```
```suggestion
be opened is greater than 2^62 - 1, then the receiver MUST interpret it as no
```

draft-ietf-quic-transport.md
```diff
@@ -849,7 +849,9 @@ initial_stream_id_for_type) can be opened (see {{long-packet-types}}).  Initial
...
+stream limit and the receiver MUST NOT close the connection.
```
MikeBishop: I don't think this says quite what you mean -- I can send you a MAX_STREAMS frame with a very large value and prohibit you from closing the connection thereafter?  Of course, the only way to violate that prohibition is to close the connection, and the only punishment is to close the connection, so this would be entirely toothless even if it were what you meant.

```suggestion
stream limit; this MUST NOT be treated as an error.
```

ianswett:
Thanks Mike, I took your suggested language.

janaiyengar:
I think we're leaning towards closing with error in #2487 (and including @martinthomson's comment above). @MikeBishop : I'm think this should be close with error error simply because this shouldn't be happening at the peer.

mikkelfj:
Yes, after thinking about it some more, an error is better. Otherwise it is a can if worms. Usually we hit peer with a hammer, so why not here?

ianswett:
I sent out #2499 as the alternative fix for this issue.

ianswett:
Closing this in favor of #2499 which seems more popular.

PR #2489, opened 2019-02-28
Fix PTO calculation in pseudo-code

ghedo:
Maybe I'm missing something, but it seems #2480 forgot to update the pseudo-code where PTO is used.

Also, should the PTO calculated in `InPersistentCongestion()` also use `kGranularity`?


ianswett:
Thanks for the fix.  And yes, I believe InPersistentCongestion() should use kGranularity as well.

PR #2497, opened 2019-03-05
Change error code for DUPLICATE_PUSH

martinduke:
I believe UNEXPECTED_FRAME is the preferred type.


LPardue:
I actually prefer the wording in duplicate push section and wonder if we should port that to PUSH_PROMISE too

PR #2501, opened 2019-03-05
Define Error when Empty Bits != 0

martinduke:
in the PRIORITY frame


MikeBishop:
I agree, this doesn't seem to mesh with the discussion on the issue.  If nothing else, it introduces the inconsistency @LPardue pointed out, but the gist of the issue seems to be that a receiver MAY check that it's zero.

martinduke:
@MikeBishop Are you in favor of changing MUST to MAY, should I discard the PR, or have I misunderstood your comment completely?

MikeBishop:
MUST => MAY

PR #2506, opened 2019-03-08
Clarify RTT calculations in recovery

### LINKED ISSUE #2137: ACK Delay interactions with RTT calculation
>  Limiting ack_delay to max_ack_delay ensures a peer specifying an extremely small max_ack_delay doesn’t cause more spurious timeouts than a peer that correctly specifies max_ack_delay. 

Does this mean that you cap the ACK Delay field to the value from the max_ack_delay transport parameter with min()?  That's what the pseudocode says, but "limit" is sufficiently vague that I wasn't sure.  Especially since that has the effect of feeding massively inflated RTT samples into the estimator.  ACK Delay will be large if the packet contains retransmissions.

Shouldn't we be throwing out these samples rather than distorting them?  I mean, that was my understanding of the outcome of #1438, and it seems like this case is rare enough that it won't starve the RTT estimator too badly.

>  If the result is smaller than the min_rtt, the RTT should be used, but the ack delay field should be ignored.

I think that this means that the RTT calculation should be done with an ACK Delay of 0.

ianswett:
Also clarifies the implications of ignoring ack_delay when it would result in an RTT sample less than min_rtt.

Fixes #2137

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -309,15 +309,15 @@ acked packet was sent.  An RTT sample MUST NOT be taken for a packet that
 is not newly acknowledged or not ack-eliciting.
 
 When RTT is calculated, the ack delay field from the ACK frame SHOULD be limited
-to the max_ack_delay specified by the peer.  Limiting ack_delay to max_ack_delay
-ensures a peer specifying an extremely small max_ack_delay doesn't cause more
-spurious timeouts than a peer that correctly specifies max_ack_delay. It SHOULD
-be subtracted from the RTT as long as the result is larger than the min_rtt.
-If the result is smaller than the min_rtt, the RTT should be used, but the
-ack delay field should be ignored.
-
-A sender calculates both smoothed RTT (SRTT) and RTT variance (RTTVAR) similar
-to those specified in {{?RFC6298}}, see {{on-ack-received}}.
+to the max_ack_delay specified by the peer by using a minimum of the two values.
+Limiting ack_delay to max_ack_delay ensures a peer specifying an extremely small
+max_ack_delay doesn't cause more spurious timeouts than a peer that correctly
+specifies max_ack_delay. It SHOULD be subtracted from the RTT as long as the
+result is larger than the min_rtt. If the result is smaller than the min_rtt,
+the RTT should be used, but the ack delay field should be ignored.  Ignoring
+ack_delay when it would cause an RTT sample smaller than the current min_rtt
+prevents a peer from manipulating RTT samples below min_rtt at the cost of slightly
+increasing RTT samples in the presence of persistent delayed acks.
 
 A sender takes an RTT sample when an ACK frame is received that acknowledges a
 larger packet number than before (see {{on-ack-received}}).  A sender will take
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -330,6 +330,31 @@ min_rtt is the minimum RTT measured over the connection, prior to adjusting by
...
+
+On every newly acknowledged ack-eliciting largest acked:
+~~~
+latest_rtt = ack_time - send_time_of_largest_acked
+~~~
+
+First RTT sample:
+~~~
+min_rtt = latest_rtt
+smoothed_rtt = latest_rtt
```
martinthomson: So no ack delay here at all?  That probably needs explanation.

draft-ietf-quic-recovery.md
```diff
@@ -330,6 +330,31 @@ min_rtt is the minimum RTT measured over the connection, prior to adjusting by
 ack delay.  Ignoring ack delay for min RTT prevents intentional or unintentional
 underestimation of min RTT, which in turn prevents underestimating smoothed RTT.
```
martinthomson: I find the structure of this section to be a little hard to follow.  I think that you want to lay it out differently.

The baseline concepts here are:

1. the observed RTT (latest_rtt in the code), which includes any processing delays on the receiver end

2. minimum RTT, which is set to the minimum observed RTT for the path.  this doesn't include any processing delays on the receiver end, which might inflate the value a little, but we assume that receivers will immediately acknowledge a packet at some point (usually this is a good idea for a new path, and handshaking more or less depends on that)

3. ack delay, which captures the processing time and any forced delays on the receiver.  This is limited to the maximum value the receiver advertises.  This value is self-reported, so any error needs to result in using less network capacity, not more, or it could be used to increase network load inappropriately.  The goal here is to have an RTT estimate that is larger.  Thus, the design is such that if ACK delay is under-reported, or the maximum is set lower than the reality, this will result in an inflated RTT estimate.  If this is over-reported, the maximum will cap the value and again cause the RTT estimate to be driven up also.

4. The final RTT calculation.  This is calculated based on the observed RTT and the adjusted ack delay.  The value is smoothed so that some history is retained.

This structure matches the code, and I think

draft-ietf-quic-recovery.md
janaiyengar: For the first sample, using ack_delay will reduce the smoothed_rtt to below min_rtt. I think adding a sentence up in the para above might make sense. How about "Note that smoothed_rtt computation on the first RTT sample does not use ack_delay, because doing so will lead to a smoothed_rtt that is smaller than the min_rtt."

draft-ietf-quic-recovery.md
```diff
@@ -309,27 +309,54 @@ acked packet was sent.  An RTT sample MUST NOT be taken for a packet that
...
+samples are generated within an RTT, the smoothed RTT and RTT variance could
+retain inadequate history, as suggested in {{?RFC6298}}. Changing these
+computations is currently an open research question.
 
 min_rtt is the minimum RTT measured over the connection, prior to adjusting by
 ack delay.  Ignoring ack delay for min RTT prevents intentional or unintentional
 underestimation of min RTT, which in turn prevents underestimating smoothed RTT.
 
+A sender calculates both smoothed RTT (SRTT) and RTT variance (RTTVAR) similar
+to those specified in {{?RFC6298}}.  Note that smoothed_rtt does for the first
```
```suggestion
to those specified in {{?RFC6298}}.  Note that computing smoothed_rtt for the first
```

PR #2521, opened 2019-03-14
QPACK: Use the term "HTTP/3" in the title

omasanori:
It seems appropriate given that the term is used elsewhere in the text.
However, I am sorry if there is a specific reason for keeping "HTTP over QUIC" here.


afrind:
This matches HPACK (Header Compression for HTTP/2), so sure.  I guess if we use vanilla QPACK for HTTP/4+ people will figure it out.

circleci failed but the error appears unrelated.

omasanori:
Thanks.

Issue #2526, opened 2019-03-17
PUSH_ID as a frame


lnicco:
the push id that links a push stream to a push promise is now an unframed varint that comes right after the stream preface. 

this is right now the only unframed data (except made for the unidirectional stream prefaces) in the specs and I am proposing to make that a frame as well for consistency and to simplify implementation and allow code reuse. 

motivation: the specs already define frames that only carry a single varint (mostly push related)
  * max push id
  * duplicate push 
  * cancel push
  * and goaway!

here instead we decided to leave it unframed just because we can since it is at the beginning of the stream

cost: two extra bytes per push stream

the text will need to specify that on a push stream this new frame (PUSH_ID ?) MUST be the first one, otherwise it is a protocol error. 

I am happy to put up a PR, in case we reach consensus

martinthomson:
The varint type and length of a frame is also unframed.  So you are asking to trade one unframed varint for two.  Is your assertion that this is unique and difficult code to write, or that the unframed content is somehow difficult to manage?

LPardue:
This seems to be going back to some of the early discussion we had on a thread I can't find. I'm not against discussing things again because we have more experience. However, I don't see a compelling reason for the change. My implementation needs to parse unframed data because of the stream header and for QPACK streams. 

I think this change might make my implementation more complex. Reading multiple varints requires more work: more checks to see if the stream has enough data, more checks to ensure only 1 frame sent at a specific time (this is the same as SETTINGS)

lnicco:
> The varint type and length of a frame is also unframed. 

@martinthomson those are in the frame header -- so they are part of the frame. I am not sure what your point is here then.

Here I am assuming that an implementation has generic code to parse frames, and that parsing anything that is not part of a frame requires special handling. 
While this is implementation specific and there can definitely be counter examples, from an abstraction and consistency point of view I see no reason for doing it the way we are doing it now. 

I believe there should be two type of streams, framed (control, request, and push streams) and unframed (QPACK, and potentially extension streams). The push stream right now is a hybrid, that only becomes a framed stream after receiving that first PUSH_ID. If I am recalling correctly when we discussed potentially having DATA frames with zero-length to transition to unframed data, it was also considered a bit unusual. Here I am arguing that we are sort of doing the reverse here where we have an unframed stream that then transitions to framed. I think this generates complexity. And really the benefit we get is saving two bytes and one frame type.

[...]

PR #2527, opened 2019-03-18
Push stream data and PUSH_PROMISE reordering clarification

rmarx:
As discussed on Slack, currently the text makes no mention of the case where the PUSH_PROMISE arrives behind (some of) the data on the push stream itself. However, it does mention this situation for the DUPLICATE_PUSH frame. 

This proposed change recommends buffering pushed data for an unknown PUSH_ID and using flow control to prevent this from being a problem. 

I am unsure if more text is needed (e.g., what happens if the PUSH_ID is never resolved (which would mean a misbehaving server)), but I think the proposed change should suffice. 

This is also related to issue #2526.
Thanks to @LPardue for his help.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1207,7 +1207,14 @@ the requests if the requested resource is already being pushed.
 When a server later fulfills a promise, the server push response is conveyed on
 a push stream (see {{push-streams}}). The push stream identifies the Push ID of
 the promise that it fulfills, then contains a response to the promised request
-using the same format described for responses in {{request-response}}.
+using the same format described for responses in {{request-response}}. Due to
+reordering, data on a push stream can arrive before the corresponding
+PUSH_PROMISE, in which case both the associated client request and the pushed
+request headers are unknown. Clients which receive a new push stream with an
+as-yet-unknown Push ID can buffer the stream data in expectation of the matching
+PUSH_PROMISE. A client can use stream flow control (see section 4.1 of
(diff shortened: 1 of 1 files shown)


LPardue:
LGTM. Non-normative language seems to be the best form of guidance here

draft-ietf-quic-http.md
```diff
@@ -1207,7 +1207,14 @@ the requests if the requested resource is already being pushed.
 When a server later fulfills a promise, the server push response is conveyed on
 a push stream (see {{push-streams}}). The push stream identifies the Push ID of
 the promise that it fulfills, then contains a response to the promised request
-using the same format described for responses in {{request-response}}.
+using the same format described for responses in {{request-response}}. Due to
+reordering, data on a push stream can arrive before the corresponding
+PUSH_PROMISE, in which case both the associated client request and the pushed
+request headers are unknown. Clients which receive a new push stream with an
```
martinthomson: CMOS change:
```suggestion
request headers are unknown. Clients that receive a new push stream with an
```

rmarx: That was based on identical text for DUPLICATE_PUSH a few lines up. I've now also changed that to keep consistency.

MikeBishop: You haven't changed it _here_, though.

rmarx: Sorry, wasn't sure if I should do this or the editors before merge. Should be accepted now.

PR #2535, opened 2019-03-20
disallow port changes on disable_migration

martinduke:
Some infrastructures are likely to load balance on port as well as address.


martinthomson:
"Address" has been taken to mean IP and port in other contexts, but no harm in being extra precise here.

PR #2537, opened 2019-03-20
Clarify meaning of RESET_STREAM

martinduke:
Saying it "terminates a stream" dangerously makes it sound like it has the same meaning as HTTP/2, which is false.


martinthomson:
Thanks for all these little patches Martin, they are a real help.

Issue #2471, opened 2019-02-15
Stateless Reset Lacks Normative Text


martinduke:
While there are some prohibitions, and an explicit MAY for responding to Long Headers, there is no normative statement for the usual use case for Stateless Reset.

I believe it's a SHOULD, but maybe someone thinks it's a MUST.

Relatedly, in 10.4.1, MUST we attempt to detect Stateless Reset, or MAY endpoints just drop after decrypt failure?

At the end of 10.4.2, MUST we not reuse connection IDs if there's a static key? (I think the answer is yes)

PR #2548, opened 2019-03-24
Clarify that stateless_reset_token is not mandatory

### LINKED ISSUE #2547: Clarify that stateless_reset_token is not mandatory
The language could be clearer, and there was some confusion today during the hackathon.

/cc @marten-seemann

ghedo:
Fixes #2547


martinthomson:
I don't see any need to make this change (especially the parenthetical).

ghedo:
As per discussion in #2547, #2471 already covers this, so I'll just wait to see what comes out of that.

Issue #2555, opened 2019-03-26
Define idle period for congestion control


pravb:
Currently the recovery draft 19 requires the sender to either reducing the cwnd to initial cwnd OR pace the send after idle. Idle is defined as "it ceases to send data and has no bytes in
   flight" . This will mean that just a very small delay in the sending application thread will cause QUIC to reset its cwnd when not pacing. IMO we need to define an interval for this which could be a function of kGranularity or as PTO. FWIW in TCP the idle period for cwnd restart is defined as one retransmission timeout (reference https://tools.ietf.org/html/rfc5681 Section 4.1).

ianswett:
However, this still has the problem that if I don't pace and I have a window of 100 packets, I can burst 100 packets into the network.  This seems suboptimal from both the sender and the network's perspective.

Now that I think about it, I think the concept of limiting the size of a burst is the key here, not so much idleness, and I think defining idle more accurately is both not necessary and likely to be difficult.  Ideally you would send no more than max(10, BW * kTimerGranularity) packets into the network at once.

pravb:
@ianswett pacing or limiting burst is the right idea. TCP RFC limits the burst to IW but it says do that after an idle period of an RTO. Presumably the authors and community considered that period to be small enough that network capacity / conditions hadn't changed drastically. 

Implementation wise how would we do this without a timer? If not pacing, when would one decide to cut cwnd back to restart cwnd (say IW)?

ianswett:
Your interpretation of RFC5661 looks correct and is different from what's in the QUIC recovery doc.  But the implications are that it's ok to send an unlimited sized burst into the network as long as you only drained the pipe very recently.  I think that recommendation is a poor one unless the window sizes are fairly small(in which case they'd be close to IW10 anyway).

The sender is bursting packets into the network, which previously did not cause loss because they were ack-clocked, but one or more buffers on the network doesn't have enough space for the full CWND.  It seems likely this could cause losses for other flows sharing the link as well.  This could happen frequently in HTTP applications such as DASH video playback.

NewCWV is already referenced for decaying CWND after idle.  I could change the text to say nothing about reducing CWND to IW upon full idle and instead say you shouldn't burst more than IW into the network at once ever, and senders may use pacing, ACK-clocking, and/or timers to accomplish that?

https://tools.ietf.org/html/rfc7661 may also be relevant to this topic?

pravb:
[...]

PR #2557, opened 2019-03-26
Simplify kPersistentThreshold

janaiyengar:
This simplifies the definition of kPersistentThreshold from an exponent to a PTO multiplier.


janaiyengar:
I think let's keep that separate. I think #2555 might end up with different outcomes anyways.

yangchi:
Curious, why wasn't this listed under draft-20 C.1? https://tools.ietf.org/html/draft-ietf-quic-recovery-20#appendix-C.1

janaiyengar:
Oversight. Fixed now in
https://github.com/quicwg/base-drafts/commit/05e9d183140a3b9474eae24a9c013a9b9aa9773e
.
Thanks!

PR #2575, opened 2019-04-01
Normative text for 1-RTT before handshake completion

### LINKED ISSUE #2572: ClientFinished is HoL blocking, but has no normative text
Previous discussions have established the CFIN is HoL blocking, partially because even though it's not strictly necessary, if client auth is in use, it's critical to receive the client's entire Handshake flight before the server processes 1RTT packets.  I don't think the current TLS text is sufficiently clear(and normative) about that.

Section 4.1.1 has:
    "Important:  Until the handshake is reported as complete, the
      connection and key exchange are not properly authenticated at the
      server.  Even though 1-RTT keys are available to a server after
      receiving the first handshake messages from a client, the server
      cannot consider the client to be authenticated until it receives
      and validates the client's Finished message.

      The requirement for the server to wait for the client Finished
      message creates a dependency on that message being delivered.  A
      client can avoid the potential for head-of-line blocking that this
      implies by sending a copy of the CRYPTO frame that carries the
      Finished message in multiple packets.  This enables immediate
      server processing for those packets."

https://tools.ietf.org/html/draft-ietf-quic-tls-19#section-4.1.1

martinthomson:
Closes #2572.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -437,7 +437,9 @@ Important:
   are not properly authenticated at the server.  Even though 1-RTT keys are
   available to a server after receiving the first handshake messages from a
   client, the server cannot consider the client to be authenticated until it
-  receives and validates the client's Finished message.
+  receives and validates the client's Finished message.  A server MUST NOT
+  process 1-RTT packets until the handshake is complete.  A server MAY buffer or
+  discard 1-RTT packets that it cannot read.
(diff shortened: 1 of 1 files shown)


mikkelfj:
I realise this may be a bit out of context for this PR, but I can't find any clear definition of when the handshake is complete other than this being magically spawned from the inner workings of TLS.

Issue #1964, opened 2018-11-05
Bad Edge Case in Recovery Draft


martinduke:
In OnLossDetectionTimeout():

if (crypto packets are outstanding):
       // Crypto retransmission timeout.
       RetransmitUnackedCryptoData()
       crypto_count++

The literal interpretation is a little loopy. In the case where crypto data is retransmitted and then the acks arrive for the original packets, we have outstanding crypto packets but no unacked crypto data. This pseudocode will result in sending nothing, and then (not shown) starting the Loss Detection timer even though there is nothing in flight, which seems bad.

Two improvements to make here:
1) If the there is no unacked crypto data, it would be better to send any available 1-RTT data.
2) We should only start the loss detection timer if there is something to send.

ianswett:
Thanks for writing this up, I think this points out that the crypto timer should be enabled when there's crypto data to send, not when crypto packets are in flight.

However, there's an exception to that, which is if there's no crypto data to send on the client side because all Initial packets have been acked, but the client didn't receive the server's Handshake, then you can hit the Anti-amplification attack limit, as you discovered in New York.

This might be awkward to writeup, but I'll give it a try, unless you'd like to?

ianswett:
This was fixed by #2590

PR #2590, opened 2019-04-05
Don't arm the handshake timer if there's no data

ianswett:
Rewrites a substantial portion of the Crypto Retransmission Timeout section and specifies that the handshake timer is not armed if the time threshold loss detection timer is set.

Fixes the portion of #1964 not fixed in #2281

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -442,8 +442,7 @@ connections over the same network SHOULD use the previous connection's final
 smoothed RTT value as the resumed connection's initial RTT.  If no previous RTT
 is available, or if the network changes, the initial RTT SHOULD be set to 500ms,
 resulting in a 1 second initial handshake timeout as recommended in
-{{?RFC6298}}. When the first acknowledgement is received, an RTT is computed and
-the timer SHOULD be set for twice the newly computed RTT.
+{{?RFC6298}}.
 
 When a crypto packet is sent, the sender MUST set a timer for twice the smoothed
 RTT.  This timer MUST be updated when a new crypto packet is sent and when
@@ -451,7 +450,7 @@ an acknowledgement is received which computes a new RTT sample. Upon timeout,
 the sender MUST retransmit all unacknowledged CRYPTO data if possible.
 
 On each consecutive expiration of the crypto timer without receiving an
-acknowledgement for a new packet, the sender SHOULD double the crypto
+acknowledgement for a new packet, the sender MUST double the crypto
 retransmission timeout and set a timer for this period.
 
 Until the server has validated the client's address on the path, the amount of
@@ -462,12 +461,14 @@ sent, then no alarm should be armed until data has been received from the
 client.
 
 Because the server could be blocked until more packets are received, the client
-MUST ensure the crypto retransmission timer is set if there is unacknowledged
-crypto data and MUST ensure the timer is set until it has 1-RTT keys.
-If the timer expires and the client has no CRYPTO data to retransmit and does
-not have Handshake keys, it MUST send an Initial packet in a UDP datagram of
-at least 1200 bytes.  If the client has Handshake keys, it MUST send a
-Handshake packet.
+MUST ensure that the crypto retransmission timer is set if there is
+unacknowledged crypto data or if the client does not yet have 1-RTT keys.
+If the crypto retransmission timer expires before the client has 1-RTT keys,
+it is possible that the client may not have any crypto data to retransmit.
+The client MUST send a new packet to allow the server to continue sending
+data. If Handshake keys are available to the client, it MUST send a
+Handshake packet, and otherwise it MUST send an Initial packet in a UDP
+datagram of at least 1200 bytes.
 
 The crypto retransmission timer is not set if the time threshold
 {{time-threshold}} loss detection timer is set.  When the crypto
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1121,7 +1121,7 @@ SetLossDetectionTimer():
     loss_detection_timer.update(loss_time)
     return
 
-  if (crypto packets are in flight
+  if (has unacknowledged crypto data
       || endpoint is client without 1-RTT keys):
     // Crypto retransmission timer.
     if (smoothed_rtt == 0):
@@ -1159,7 +1159,7 @@ OnLossDetectionTimeout():
     DetectLostPackets(pn_space)
   // Retransmit crypto data if no packets were lost
   // and there is crypto data to retransmit.
-  else if (has unacked crypto data):
+  else if (has unacknowledged crypto data):
     // Crypto retransmission timeout.
     RetransmitUnackedCryptoData()
     crypto_count++
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -448,7 +448,7 @@ When a crypto packet is sent, the sender MUST set a timer for twice the smoothed
 RTT.  This timer MUST be updated when a new crypto packet is sent and when
 an acknowledgement is received which computes a new RTT sample. Upon timeout,
 the sender MUST retransmit all unacknowledged CRYPTO data if possible.  The
-sender MUST NOT declare in flight crypto packets as lost when the crypto timer
+sender MUST NOT declare in-flight crypto packets as lost when the crypto timer
 expires.
 
 On each consecutive expiration of the crypto timer without receiving an
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -461,8 +461,8 @@ client.
 
 Because the server could be blocked until more packets are received, the client
 MUST ensure the crypto retransmission timer is set if there is unacknowledged
-crypto data and MUST ensure the the timer is set until it has 1RTT keys.
-If the timer expires and the client has no  CRYPTO data to retransmit and does
+crypto data and MUST ensure  the timer is set until it has 1RTT keys.
+If the timer expires and the client has no CRYPTO data to retransmit and does
 not have Handshake keys, it MUST send an Initial packet in a UDP datagram of
 at least 1200 bytes.  If the client has Handshake keys, it MUST send a
 Handshake packet.
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -434,8 +434,12 @@ and larger thresholds increase loss detection delay.
 Data in CRYPTO frames is critical to QUIC transport and crypto negotiation, so a
 more aggressive timeout is used to retransmit it.
 
+MUST start the crypto retransmission timer if there is outstanding CRYPTO data. It MUST also start the crypto retransmission timer until it has 1-RTT keys,
+
 The initial crypto retransmission timeout SHOULD be set to twice the initial
-RTT.
+RTT.  On each consecutive expiration of the crypto timer without receiving an
+acknowledgement for a new packet, the sender SHOULD double the crypto
+retransmission timeout and set a timer for this period.
 
 At the beginning, there are no prior RTT samples within a connection.  Resumed
 connections over the same network SHOULD use the previous connection's final
@@ -458,17 +462,15 @@ sent, then no alarm should be armed until data has been received from the
 client.
 
 Because the server could be blocked until more packets are received, the client
-MUST start the crypto retransmission timer until it has 1RTT keys, even if there
-is no unacknowledged CRYPTO data.  If the timer expires and the client has no
-CRYPTO data to retransmit and does not have Handshake keys, it MUST send an
-Initial packet in a UDP datagram of at least 1200 bytes.  If the client has
-Handshake keys, it MUST send a Handshake packet.
+MUST ensure the crypto retransmission timer is set if there is unacknowledged
+crypto data and MUST ensure the the timer is set until it has 1RTT keys.
+If the timer expires and the client has no  CRYPTO data to retransmit and does
+not have Handshake keys, it MUST send an Initial packet in a UDP datagram of
+at least 1200 bytes.  If the client has Handshake keys, it MUST send a
+Handshake packet.
 
-On each consecutive expiration of the crypto timer without receiving an
-acknowledgement for a new packet, the sender SHOULD double the crypto
-retransmission timeout and set a timer for this period.
-
-When crypto packets are in flight, the probe timer ({{pto}}) is not active.
+When the crypto retransmission timer is active, the probe timer ({{pto}})
+is not active.
 
 
 ### Retry and Version Negotiation
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1118,8 +1118,8 @@ SetLossDetectionTimer():
     loss_detection_timer.update(loss_time)
     return
 
-  if (crypto packets are in flight &&
-      crypto data to send):
+  if (crypto packets are in flight
+      || endpoint is client without 1-RTT keys):
     // Crypto retransmission timer.
     if (smoothed_rtt == 0):
       timeout = 2 * kInitialRtt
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -458,11 +458,11 @@ sent, then no alarm should be armed until data has been received from the
 client.
 
 Because the server could be blocked until more packets are received, the client
-MUST start the crypto retransmission timer even if there is no unacknowledged
-CRYPTO data.  If the timer expires and the client has no CRYPTO data to
-retransmit and does not have Handshake keys, it SHOULD send an Initial packet in
-a UDP datagram of at least 1200 bytes.  If the client has Handshake keys, it
-SHOULD send a Handshake packet.
+MUST start the crypto retransmission timer until it has 1RTT keys, even if there
```
marten-seemann: How would a lost CFIN be retransmitted?

draft-ietf-quic-recovery.md
```diff
@@ -1151,7 +1152,8 @@ OnLossDetectionTimeout():
     DetectLostPackets(pn_space)
   // Retransmit crypto data if no packets were lost
   // and there are still crypto packets in flight.
-  else if (crypto packets are in flight):
+  else if (crypto packets are in flight &&
+           crypto data to send):
     // Crypto retransmission timeout.
     RetransmitUnackedCryptoData()
     crypto_count++
```
martinduke: else if (client && (no 1-RTT keys)
    send a 1200B packet
or something to that effect

draft-ietf-quic-recovery.md
```diff
@@ -1114,7 +1114,8 @@ SetLossDetectionTimer():
     loss_detection_timer.update(loss_time)
     return
 
-  if (crypto packets are in flight):
+  if (crypto packets are in flight &&
+      crypto data to send):
```
martinduke: || (client && (no 1-RTT keys yet))

draft-ietf-quic-recovery.md
martinduke: MUST start the crypto retransmission timer upon timeout....

In line 448: ack-eliciting crypto packet?

LIne 451: "if possible, and restart the crypto retransmission timer if there is unacknowledged data"

draft-ietf-quic-recovery.md
```diff
@@ -458,11 +458,11 @@ sent, then no alarm should be armed until data has been received from the
...
 
 Because the server could be blocked until more packets are received, the client
-MUST start the crypto retransmission timer even if there is no unacknowledged
-CRYPTO data.  If the timer expires and the client has no CRYPTO data to
-retransmit and does not have Handshake keys, it SHOULD send an Initial packet in
-a UDP datagram of at least 1200 bytes.  If the client has Handshake keys, it
-SHOULD send a Handshake packet.
+MUST start the crypto retransmission timer until it has 1RTT keys, even if there
+is no unacknowledged CRYPTO data.  If the timer expires and the client has no
+CRYPTO data to retransmit and does not have Handshake keys, it SHOULD send an
```
martinduke: SHOULD -> MUST? If we don't do this, the connection could hang in a fairly common case.

draft-ietf-quic-recovery.md
ianswett: It's not just timeout(ie: it needs to be set after an ACK is received), but I'll try to clarify.

line 448: The definition of crypto packet is that it has a CRYPTO frame in it, so it's always ack-eliciting.

Retransmitting the crypto data would naturally update the timer(see previous sentence) and there would be unacknowledged crypto data in that case, so I'm not sure your extra clause is clarifying or potentially confusing?

martinduke: There would be outstanding CRYPTO data, so it would restart the timer and then send.

draft-ietf-quic-recovery.md
martinduke: ah you beat me to it.

draft-ietf-quic-recovery.md
ianswett: That's in the the other PR: #2281

draft-ietf-quic-recovery.md
ianswett: Also in the other PR.

draft-ietf-quic-recovery.md
marten-seemann: You're right, but it's kind of hard to understand.
What about the following structure: "MUST start the crypto retransmission timer if there is outstanding CRYPTO data. It MUST also start the crypto retransmission timer until it has 1-RTT keys,..."

draft-ietf-quic-recovery.md
```diff
@@ -435,7 +435,9 @@ Data in CRYPTO frames is critical to QUIC transport and crypto negotiation, so a
 more aggressive timeout is used to retransmit it.
 
 The initial crypto retransmission timeout SHOULD be set to twice the initial
-RTT.
+RTT.  On each consecutive expiration of the crypto timer without receiving an
+acknowledgement for a new packet, the sender SHOULD double the crypto
+retransmission timeout and set a timer for this period.
```
martinduke: Not quite right. If we got some acknowledgments, we still need to restart the timer. And note the pseudocode comments below.

ianswett: This was just a reorg.  This text is about doubling the timer after expiry when no ACKs are received.

I think the pseudocode comments are all addressed in the other PR.

draft-ietf-quic-recovery.md
martinduke: You're right, please disregard.

draft-ietf-quic-recovery.md
```diff
@@ -458,17 +460,15 @@ sent, then no alarm should be armed until data has been received from the
...
-CRYPTO data.  If the timer expires and the client has no CRYPTO data to
-retransmit and does not have Handshake keys, it SHOULD send an Initial packet in
-a UDP datagram of at least 1200 bytes.  If the client has Handshake keys, it
-SHOULD send a Handshake packet.
-
-On each consecutive expiration of the crypto timer without receiving an
-acknowledgement for a new packet, the sender SHOULD double the crypto
-retransmission timeout and set a timer for this period.
+MUST ensure the crypto retransmission timer is set if there is unacknowledged
+crypto data and MUST ensure  the timer is set until it has 1RTT keys.
```
martinduke: double space

draft-ietf-quic-recovery.md
martinduke: Ah, OK. Line 448 probably should read "When an acknowledgment is received, a new RTT is computed. The endpoint MUST restart the timer and SHOULD set it for twice the newly computed smoothed RTT."

draft-ietf-quic-recovery.md
```diff
@@ -458,17 +462,16 @@ sent, then no alarm should be armed until data has been received from the
...
-SHOULD send a Handshake packet.
-
-On each consecutive expiration of the crypto timer without receiving an
-acknowledgement for a new packet, the sender SHOULD double the crypto
-retransmission timeout and set a timer for this period.
+MUST ensure the crypto retransmission timer is set if there is unacknowledged
+crypto data and MUST ensure the timer is set until it has 1-RTT keys.
+If the timer expires and the client has no CRYPTO data to retransmit and does
+not have Handshake keys, it MUST send an Initial packet in a UDP datagram of
+at least 1200 bytes.  If the client has Handshake keys, it MUST send a
```
marten-seemann: Why would the client send a Handshake packet if there’s no CRYPTO data to retransmit? If there’s no Handshake CRYPTO data to retransmit, this means that the server already received the client‘s TLS messages sent with Handshake keys, and won’t be blocked by the 3x Limit any more, doesn’t it?

martinduke:
If the server first flight does not include finished and the client
handshake ack is lost, the client has no unacked data but the server is
flow control limited.

draft-ietf-quic-recovery.md
ianswett: I believe the particular concern is that the client could have Handshake keys because they receive the server's Initial, but not receive the server's Handshake packet, so have nothing to send in Handshake encryption.

If the client ever had CRYPTO data to send in Handshake, then this would not be a concern, as you state.

draft-ietf-quic-recovery.md
```diff
@@ -442,13 +442,17 @@ connections over the same network SHOULD use the previous connection's final
 smoothed RTT value as the resumed connection's initial RTT.  If no previous RTT
 is available, or if the network changes, the initial RTT SHOULD be set to 500ms,
 resulting in a 1 second initial handshake timeout as recommended in
-{{?RFC6298}}. When an acknowledgement is received, a new RTT is computed and the
-timer SHOULD be set for twice the newly computed smoothed RTT.
+{{?RFC6298}}. When the first acknowledgement is received, an RTT is computed and
+the timer SHOULD be set for twice the newly computed RTT.
```
janaiyengar: How about "When the first acknowledgement is received, the timer SHOULD be set for twice the value of the resulting RTT sample"? Now that I see further below, why is this necessary at all? I would remove this last sentence. The beginning of the next para says exactly this, since the smoothed_rtt is set to the RTT sample on the first ack.

draft-ietf-quic-recovery.md
```diff
@@ -442,13 +442,17 @@ connections over the same network SHOULD use the previous connection's final
...
 
-When a crypto packet is sent, the sender MUST set a timer for the crypto
-timeout period.  This timer MUST be updated when a new crypto packet is sent.
-Upon timeout, the sender MUST retransmit all unacknowledged CRYPTO data if
-possible.
+When a crypto packet is sent, the sender MUST set a timer for twice the smoothed
+RTT.  This timer MUST be updated when a new crypto packet is sent and when
+an acknowledgement is received which computes a new RTT sample. Upon timeout,
+the sender MUST retransmit all unacknowledged CRYPTO data if possible.
```
janaiyengar: Why "if possible"?

draft-ietf-quic-recovery.md
```diff
@@ -442,13 +442,17 @@ connections over the same network SHOULD use the previous connection's final
...
+
+On each consecutive expiration of the crypto timer without receiving an
+acknowledgement for a new packet, the sender SHOULD double the crypto
+retransmission timeout and set a timer for this period.
```
janaiyengar: I'd make this a MUST.

draft-ietf-quic-recovery.md
```diff
@@ -458,17 +462,16 @@ sent, then no alarm should be armed until data has been received from the
...
-CRYPTO data.  If the timer expires and the client has no CRYPTO data to
-retransmit and does not have Handshake keys, it SHOULD send an Initial packet in
-a UDP datagram of at least 1200 bytes.  If the client has Handshake keys, it
-SHOULD send a Handshake packet.
-
-On each consecutive expiration of the crypto timer without receiving an
-acknowledgement for a new packet, the sender SHOULD double the crypto
-retransmission timeout and set a timer for this period.
+MUST ensure the crypto retransmission timer is set if there is unacknowledged
+crypto data and MUST ensure the timer is set until it has 1-RTT keys.
```
janaiyengar: This needs rephrasing. How about "MUST ensure that the crypto retransmission timer is set if there is unacknowledged crypto data or if the client does not yet have 1-RTT keys."

draft-ietf-quic-recovery.md
```diff
@@ -458,17 +462,16 @@ sent, then no alarm should be armed until data has been received from the
...
+If the timer expires and the client has no CRYPTO data to retransmit and does
+not have Handshake keys, it MUST send an Initial packet in a UDP datagram of
+at least 1200 bytes.  If the client has Handshake keys, it MUST send a
+Handshake packet.
```
janaiyengar: Suggested reword: "If the crypto retransmission timer expires before the client has 1-RTT keys, it is possible that the client may not have any crypto data to retransmit. The client MUST still send a new packet to allow the server to continue sending data (see Section XX in {{QUIC-TRANSPORT}}).  If Handshake keys are available to the client, it MUST send a Handshake packet, and otherwise, an Initial packet in a UDP datagram of at least 1200 bytes."

draft-ietf-quic-recovery.md
```diff
@@ -1114,7 +1117,8 @@ SetLossDetectionTimer():
     loss_detection_timer.update(loss_time)
     return
 
-  if (crypto packets are in flight):
+  if (crypto packets are in flight &&
+      crypto data to send):
```
janaiyengar: what does this predicate mean? That there's new data to send?

draft-ietf-quic-recovery.md
ianswett: SG, removed.

draft-ietf-quic-recovery.md
ianswett: It can't if the anti-amplification limit limits it.

draft-ietf-quic-recovery.md
ianswett: It's supposed to mean that there's unacknowledged crypto data.  Should I say "unacknowledged crypto data" instead?

draft-ietf-quic-recovery.md
ianswett: Done, with some minor editing.

draft-ietf-quic-recovery.md
ianswett: I removed this from here due to the other PR, but left it in below.  Feel free to suggest alternate phrasing.

draft-ietf-quic-recovery.md
```diff
@@ -458,17 +463,18 @@ sent, then no alarm should be armed until data has been received from the
...
-SHOULD send a Handshake packet.
-
-On each consecutive expiration of the crypto timer without receiving an
-acknowledgement for a new packet, the sender SHOULD double the crypto
-retransmission timeout and set a timer for this period.
+MUST ensure that the crypto retransmission timer is set if there is
+unacknowledged crypto data or if the client does not yet have 1-RTT keys.
+If the crypto retransmission timer expires before the client has 1-RTT keys,
+it is possible that the client may not have any crypto data to retransmit.
+The client MUST send a new packet to allow the server to continue sending
```
```suggestion
However, the client MUST send a new packet, containing only PING or PADDDING frames if necessary, to allow the server to continue sending
```

draft-ietf-quic-recovery.md
```diff
@@ -442,13 +442,18 @@ connections over the same network SHOULD use the previous connection's final
...
 
-When a crypto packet is sent, the sender MUST set a timer for the crypto
-timeout period.  This timer MUST be updated when a new crypto packet is sent.
-Upon timeout, the sender MUST retransmit all unacknowledged CRYPTO data if
-possible.
+When a crypto packet is sent, the sender MUST set a timer for twice the smoothed
+RTT.  This timer MUST be updated when a new crypto packet is sent and when
+an acknowledgement is received which computes a new RTT sample. Upon timeout,
+the sender MUST retransmit all unacknowledged CRYPTO data if possible.  The
+sender MUST NOT declare in flight crypto packets as lost when the crypto timer
```
```suggestion
sender MUST NOT declare in-flight crypto packets as lost when the crypto timer
```

draft-ietf-quic-recovery.md
```diff
@@ -1151,7 +1158,8 @@ OnLossDetectionTimeout():
     DetectLostPackets(pn_space)
   // Retransmit crypto data if no packets were lost
   // and there are still crypto packets in flight.
-  else if (crypto packets are in flight):
+  else if (crypto packets are in flight &&
+           crypto data to send):
```
janaiyengar: I still don't get this. If crypto packets are in flight, doesn't that mean there is unacked crypto data to send? Why is this second predicate useful?

ianswett: Crypto packets could be in flight because of a timer based retransmission, but the original was ACked.

ianswett: See issue #1964

janaiyengar: Ah, ok. How about just "unacked crypto data exists"? Is that enough to capture it?

janaiyengar: I meant using just this one condition instead of the two that you have.

Ralith: I agree that this wording is confusing; in particular, it's not obvious that "to send" means "not yet acknowledged," except perhaps by process of elimination.

ianswett: Sure, I took Jana's suggestion.

PR #2601, opened 2019-04-09
Frames of unknown type cause connection close

janaiyengar:
This was always intended (and also written in the description of FRAME_ENCODING_ERROR), but didn't show up in the text somehow.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2701,10 +2701,10 @@ frames. -->
 {: #packet-frames title="QUIC Payload"}
 
 QUIC payloads MUST contain at least one frame, and MAY contain multiple frames
-and multiple frame types.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2701,10 +2701,10 @@ frames. -->
 {: #packet-frames title="QUIC Payload"}
 
 QUIC payloads MUST contain at least one frame, and MAY contain multiple frames
-and multiple frame types.
+and multiple frame types.  Frames MUST fit within a single QUIC packet and MUST
```
ianswett: This last bit is true, but it's not just a MUST, it seems impossible given the framing we have.  I'm wondering if a can not is more appropriate, given it's essentially impossible?

But I guess this is pre-existing normative text, so in the interest in keeping this PR editorial, this is fine by me.

PR #2605, opened 2019-04-10
Reference RFC7661 for application limited

### LINKED ISSUE #2554: Congestion control during application limited state
The text in recovery draft 19 says "The congestion window should not be increased in slow start or
   congestion avoidance when it is not fully utilized". This is a deviation from how congestion control works in deployed TCP implementation. We need a crisp definition of what is meant by "full utilization". I also have concern that this will make QUICs performance for bulk transfers poorer than TCP because application I/O patterns and scheduling of the sending thread will affect cwnd growth. After making the definition crisp and consensus that we need to include such a provision, the pseudo code will need to be updated to handle this as well.

ianswett:
Combines the section on sending after idle and application limited because they share a lot in common.

Praveen, I think this is better, but tell me if you'd like to add/change something else?

Fixes #2554 and #2555

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -742,25 +742,13 @@ As an example of a well-known and publicly available implementation of a flow
 pacer, implementers are referred to the Fair Queue packet scheduler (fq qdisc)
 in Linux (3.11 onwards).
 
-
-## Sending data after an idle period
-
-A sender becomes idle if it ceases to send data and has no bytes in flight.  A
-sender's congestion window MUST NOT increase while it is idle.
-
-When sending data after becoming idle, a sender MUST reset its congestion window
-to the initial congestion window (see Section 4.1 of {{?RFC5681}}), unless it
-paces the sending of packets. A sender MAY retain its congestion window if it
-paces the sending of any packets in excess of the initial congestion window.
-
-A sender MAY implement alternate mechanisms to update its congestion window
-after idle periods, such as those proposed for TCP in {{?RFC7661}}.
-
 ## Application Limited Sending
 
-The congestion window should not be increased in slow start or congestion
-avoidance when it is not sufficiently utilized.  The congestion window could
-be under-utilized due to insufficient application data or flow control credit.
+A sender's congestion window MUST NOT increase when no packets are newly
+acknowledged, such as when the connection is idle. The congestion window
+SHOULD NOT be increased in slow start or congestion avoidance when it is not
+sufficiently utilized.  The congestion window could be under-utilized due to
+insufficient application data or flow control credit.
 
 A sender MAY use the pipeACK method described in section 4.3 of {{?RFC7661}}
 to determine if the congestion window is sufficiently utilized.
@@ -770,6 +758,12 @@ and not fully utilize the congestion window due to this delay. A sender
 should not consider itself application limited if it would have fully
 utilized the congestion window without pacing delay.
 
+Sending more than intial window into the network at once may cause losses.
+Implemementations SHOULD use pacing, ACK-clocking, a reduction in CWND and/or
+timers to minimize bursts.
+
+A sender MAY implement alternate mechanisms to update its congestion window
+after idle periods, such as those proposed for TCP in {{?RFC7661}}.
 
 
 # Security Considerations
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -742,31 +742,28 @@ As an example of a well-known and publicly available implementation of a flow
...
 ## Application Limited Sending
 
-The congestion window should not be increased in slow start or congestion
-avoidance when it is not fully utilized.  The congestion window could be
-under-utilized due to insufficient application data or flow control credit.
+A sender's congestion window MUST NOT increase when no packets are newly
+acknowledged, such as when the connection is idle. The congestion window
+SHOULD NOT be increased in slow start or congestion avoidance when it is not
+sufficiently utilized.  The congestion window could be under-utilized due to
+insufficient application data or flow control credit.
```
janaiyengar: Editorial suggestion, after removing the first sentence: "A congestion window that is under-utilized SHOULD NOT be increased in either slow start or congestion avoidance.  This can happen due to insufficient application data or flow control credit."

I don't think saying "idle" is necessary, this text captures it in "insufficient app data", IMO.

draft-ietf-quic-recovery.md
```diff
@@ -742,31 +742,28 @@ As an example of a well-known and publicly available implementation of a flow
...
+
+A sender MAY use the pipeACK method described in section 4.3 of {{?RFC7661}}
+to determine if the congestion window is sufficiently utilized.
```
janaiyengar: I'm not sure how to read this. We are saying that the sender SHOULD NOT increase the cwnd if it's under-utilized, but we are not quite defining what under-utilized means and instead pointing to pipeACK for one way of defining it (for TCP). This doesn't seem great to me.

For the purpose of this PR, how about you leave it like this and open a new issue to summarize a definition of an under-utilized cwnd?

draft-ietf-quic-recovery.md
```diff
@@ -742,31 +742,28 @@ As an example of a well-known and publicly available implementation of a flow
...
 
 A sender that paces packets (see {{pacing}}) might delay sending packets
 and not fully utilize the congestion window due to this delay. A sender
 should not consider itself application limited if it would have fully
 utilized the congestion window without pacing delay.
 
+Sending more than intial window into the network at once may cause losses.
```
```suggestion
Bursting more than an intial window's worth of data into the network might cause short-term congestion and losses.
```

draft-ietf-quic-recovery.md
```diff
@@ -742,31 +742,28 @@ As an example of a well-known and publicly available implementation of a flow
...
+Implemementations SHOULD use pacing, ACK-clocking, a reduction in CWND and/or
```
```suggestion
Implemementations SHOULD either use pacing or reduce their congestion window to limit such bursts.
```
Reducing the cwnd basically means using ack-clocking to increase the cwnd.

draft-ietf-quic-recovery.md
```diff
@@ -742,31 +742,28 @@ As an example of a well-known and publicly available implementation of a flow
...
+timers to minimize bursts.
+
+A sender MAY implement alternate mechanisms to update its congestion window
+after idle periods, such as those proposed for TCP in {{?RFC7661}}.
```
```suggestion
after periods of under-utilization, such as those proposed for TCP in {{?RFC7661}}.
```

draft-ietf-quic-recovery.md
```diff
@@ -742,31 +742,28 @@ As an example of a well-known and publicly available implementation of a flow
...
-
-When sending data after becoming idle, a sender MUST reset its congestion window
-to the initial congestion window (see Section 4.1 of {{?RFC5681}}), unless it
-paces the sending of packets. A sender MAY retain its congestion window if it
-paces the sending of any packets in excess of the initial congestion window.
-
-A sender MAY implement alternate mechanisms to update its congestion window
-after idle periods, such as those proposed for TCP in {{?RFC7661}}.
-
 ## Application Limited Sending
```
janaiyengar: How about "Under-utilizing the Congestion Window"?

draft-ietf-quic-recovery.md
```diff
@@ -742,31 +742,28 @@ As an example of a well-known and publicly available implementation of a flow
...
 
-The congestion window should not be increased in slow start or congestion
-avoidance when it is not fully utilized.  The congestion window could be
-under-utilized due to insufficient application data or flow control credit.
+A sender's congestion window MUST NOT increase when no packets are newly
+acknowledged, such as when the connection is idle. The congestion window
```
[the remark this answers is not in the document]
ianswett: This was written to specifically forbid the cubic bug, but I realize it's not even tight enough to do that, so I'll remove it.

draft-ietf-quic-recovery.md

janaiyengar:
I'll take this in, and open an issue for defining "under-utilized cwnd"

PR #2609, opened 2019-04-11
consistent citing of internet drafts

reschke:


martinthomson:
Thanks Julian, but I think that a different form of consistency is better.

reschke:
Don't let IDNITS catch you :-)

Issue #2607, opened 2019-04-10
RTT Initial measurement


goelvidhi:
On receiving the INITIAL packet, the server passes the CRYPTO data to TLS which then does a bunch of things like computing handshake key, generate INITIAL data etc and then passes it to QUIC. From measurements, it takes ~10ms before the server sends out the INITIAL packet and ACK for client's INITIAL packet.

Now, the first RTT measurement doesn't include ack delay. For a very low RTT connection (two machines connected back to back via ethernet), this causes the initial RTT ~ 10ms, which is too high for such a connection. Even after subsequent RTT samples of 0ms, would take a long time to normalize the high initial value.

Is there a reason why we don't subtract ack_delay from measured RTT for the INITIAL packet?

ianswett:
The reason to not subtract ack_delay on the first ACK is it could allow the peer to manipulate the SRTT, which may be particularly dangerous at the beginning of a connection.  As a mitigation, the current text in 4.1 of Recovery says: 
  "In order to quickly complete the handshake and avoid spurious
   retransmissions due to crypto retransmission timeouts, crypto packets
   SHOULD use a very short ack delay, such as the local timer
   granularity.  ACK frames MAY be sent immediately when the crypto
   stack indicates all data for that packet number space has been
   received."
https://tools.ietf.org/html/draft-ietf-quic-recovery-19#section-4.1

We could add some text indicating that sending an ACK immediately improves the first RTT sample if that seems helpful?

ianswett:
PR #2592 is a substantial rewrite of the RTT estimation section, so it may be worth reading that to see if it provides the rationale you were looking for.

janaiyengar:
This is an interesting point. In addition to having an inflated RTT,
there's also the risk of the crypto timer firing for long RTTs.

In quicly, we don't delay acks during the handshake to avoid these issues.
I wonder if this is worth mentioning in the draft.

marten-seemann:
@janaiyengar, we already have [text about this](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#crypto-handshake-data):
> In order to quickly complete the handshake and avoid spurious retransmissions due to crypto retransmission timeouts, crypto packets SHOULD use a very short ack delay, such as the local timer granularity. ACK frames MAY be sent immediately when the crypto stack indicates all data for that packet number space has been received.

martinthomson:
[...]

PR #2610, opened 2019-04-12
Simplify UpdateRtt pseudocode

janaiyengar:


draft-ietf-quic-recovery.md
```diff
@@ -1101,22 +1101,24 @@ OnAckReceived(ack, pn_space):
...
   // min_rtt ignores ack delay.
   min_rtt = min(min_rtt, latest_rtt)
   // Limit ack_delay by max_ack_delay
   ack_delay = min(ack_delay, max_ack_delay)
-  // Adjust for ack delay if it's plausible.
+  // Adjust for ack delay if plausible.
   adjusted_rtt = latest_rtt
-  if (latest_rtt - min_rtt > ack_delay):
+  if (latest_rtt > min_rtt + ack_delay):
     adjusted_rtt = latest_rtt - ack_delay
```
larseggert: You can simply do `latest_rtt -= ack_delay` and then use `latest_rtt` instead of `adjusted_rtt` below, since `latest_rtt` is not used anywhere anymore

draft-ietf-quic-recovery.md
```diff
@@ -1101,22 +1101,24 @@ OnAckReceived(ack, pn_space):
 
 
 UpdateRtt(latest_rtt, ack_delay):
```
larseggert: Why is `latest_rtt` passed in as a parameter? It's already a global. Does this parameter differ from the global?

draft-ietf-quic-recovery.md
larseggert: (But if you want to use the global here, maybe that adjustment to the global is wrong?)

draft-ietf-quic-recovery.md
```diff
@@ -1101,22 +1101,24 @@ OnAckReceived(ack, pn_space):
...
+  if (smoothed_rtt == 0):
+    // First RTT sample.
```
larseggert: Shouldn't you adjust by `ack_delay` here?

draft-ietf-quic-recovery.md
martinthomson: latest_rtt probably shouldn't be global.

draft-ietf-quic-recovery.md
martinthomson: I don't think so.  #2607 covers this though.

draft-ietf-quic-recovery.md
martinthomson: The prose relies on latest_rtt in descriptions.  Adjusting that value directly would confuse things.

draft-ietf-quic-recovery.md
marten-seemann: It's used in `DetectLostPackets`, so it needs to be global.
To avoid confusion with scoping rules, the function parameter should have a different name than the global variable though.

PR #2613, opened 2019-04-12
Remove old spin bit references

LPardue:
It just so happens that the old, independent Spin Bit document expires **today**. AFAIK the necessary pieces of text have been merged into the Transport document, making this I-D outdated and confusing to people randomly browsing the core docs repo.

This change removes all references to the old spin bit doc.


larseggert:
@mirjak @britram : any reason not to merge this? i think we have all the things we need in -transport now, as @LPardue says

mirjak:
yes, should be fine.

PR #2614, opened 2019-04-13
Send ACK immediately when all data in PN space received

### LINKED ISSUE #2607: RTT Initial measurement

janaiyengar:
Changes MAY to SHOULD.
Closes #2607.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -258,7 +258,7 @@ incoming packets.
 
 In order to quickly complete the handshake and avoid spurious retransmissions
 due to crypto retransmission timeouts, crypto packets SHOULD use a very short
-ack delay, such as the local timer granularity.  ACK frames MAY be sent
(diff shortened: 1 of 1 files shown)


martinthomson:
Note that stacks generally don't indicate that all data is available, they just provide newer keys, or provide data that should be protected with those newer keys.  But the spirit of this is fine.

PR #2615, opened 2019-04-13
Make latest_rtt global

janaiyengar:
It's being sent as a parameter right now in UpdateRttI() and is global otherwise.
Also since min_rtt is now set explicitly for the first sample, initialize min_rtt to 0 for consistency.


ianswett:
Does this fix an issue?

marten-seemann:
@larseggert raised this in https://github.com/quicwg/base-drafts/pull/2610#discussion_r274765966.

PR #2617, opened 2019-04-13
Set loss delay at least kGranularity

### LINKED ISSUE #2616: latest_rtt can be zero
PTO works out fine, but loss_delay becomes weird. We should probably handle that.

janaiyengar:
Closes #2616.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -488,10 +488,11 @@ Using max(SRTT, latest_RTT) protects from the two following cases:
 
 An endpoint might consistently record RTT samples as 0 in extremely low latency
 networks, leading to a smoothed_rtt of 0.  Consequently, the endpoint could
-declare all packets as lost immediately upon receiving an acknowledgement for a
-later packet.  That is, the endpoint would not provide any reordering tolerance.
-To avoid declaring packets as lost too early, the time threshold MUST be set to
-at least kGranularity (defined in {{ld-consts-of-interest}}).
+declare all earlier packets as lost immediately upon receiving an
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -486,6 +486,12 @@ Using max(SRTT, latest_RTT) protects from the two following cases:
 * the latest RTT sample is higher than the SRTT, perhaps due to a sustained
   increase in the actual RTT, but the smoothed SRTT has not yet caught up.
 
+An endpoint might consistently record RTT samples as 0 in extremely low latency
+networks, leading to a smoothed_rtt of 0.  Consequently, the endpoint could
+declare all packets as lost immediately upon sending them.  To avoid declaring
```
ianswett: This part isn't true.  There would still have to be reordering, since you need a later sent packet to be ACKed.

janaiyengar: Good point. I've changed the text now to say that srtt of 0 means there's no reordering tolerance.

draft-ietf-quic-recovery.md
```diff
@@ -486,6 +486,13 @@ Using max(SRTT, latest_RTT) protects from the two following cases:
 * the latest RTT sample is higher than the SRTT, perhaps due to a sustained
   increase in the actual RTT, but the smoothed SRTT has not yet caught up.
 
+An endpoint might consistently record RTT samples as 0 in extremely low latency
+networks, leading to a smoothed_rtt of 0.  Consequently, the endpoint could
+declare all packets as lost immediately upon receiving an acknowledgement for a
```
```suggestion
declare all earlier packets as lost immediately upon receiving an acknowledgement for a
```

Issue #2565, opened 2019-03-29
Crypto retx timeout should have higher priority than time based loss detection


yangchi:
In both SetLossDetectionTimer() and OnLossDetectionTimeout(), the loss_time based timer setup and loss detection happens before the crypto loss recovery logics. That seems at the cross road with the idea that crypto packets having higher priority in loss recovery. In fact in OnLossDetectionTimerou(), if loss_time is present and we go into DetectLostPackets(), then we will only decect loss of one packet number space between Initial and Handshake. Plus the loss_time used to set the timer is the earliest among all possible values, which means even if only one of Initial and Handshake space has inflight packets, we may still not retransmit all the inflight packets when this timer fires. So this current code really doesn't follow the idea that when crypto is inflight and loss timer fires, all inflight crypto needs to be retransmitted.

If we change the code so that crypto is handled first, we can also reduce the states need to be maintained and simplify some of the functions. In current draft there are 3 loss_times, one per packet number space. By always handling crypto first, we will need such loss_time for AppData space only, and no need to have the helper function to GetEarliestLossTime().

(I'm more than happy to cook up a PR if people think this makes sense.)

yangchi:
It's been a few days. I sent PR #2619.

marten-seemann:
This was introduced as a fix to https://github.com/quicwg/base-drafts/issues/2435, and fixed in https://github.com/quicwg/base-drafts/pull/2451. I don't think that reverting that fix is a good idea.

ianswett:
+1 to @marten-seemann's comment.  Timers are necessary when you have no information about what might be lost, but if you do have information, there's no reason not to use that instead.

ianswett:
I sent out an editorial PR to clarify why time based loss detection has higher priority.

PR #2619, opened 2019-04-15
Handle crypto timer before time threshold timer

yangchi:
This is for issue 2565 (https://github.com/quicwg/base-drafts/issues/2565). This is the behavior until draft-17 i think. Crypto always had higher priority than other packets thus other type of loss detection (reordering threshold, time threshold and early retransmit timeout). Then in draft-17 the reordering threshold and time threshold were merged together. Somehow, loss_time also got a higher priority than crypto timer. With such change it's possible that loss_time fires, there are crypto packets in flight, but not all of them are eligible to retransmit, depending on their sending time and the loss_time value, so we may only retransmit some of them.

This diff brings the old behavior back. Crypto timer will have higher priority than timer threshold loss detection, and when crypto timer fires, if there are outstanding handshake packets, all of them are retransmitted.

This also simplifies the loss_time. There will be no need to have one loss_timer per packet number space, only ApplicationData space will need a loss_time.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -909,6 +909,9 @@ processing only applies to a single space.
 packet_number:
 : The packet number of the sent packet.
 
+pn_space:
+: The packet number space of the sent packet.
+
 ack_eliciting:
 : A boolean that indicates whether a packet is ack-eliciting.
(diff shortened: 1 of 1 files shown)


yangchi:
close. #2620 clarifies this.

PR #2620, opened 2019-04-15
Clarify time threshold vs crypto timer

ianswett:
Adds some explanation about why it's better to arm/fire the time threshold alarm before the crypto retransmission timeout.

Hopefully clarifies comments from #2565

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -539,7 +539,7 @@ and otherwise it MUST send an Initial packet in a UDP datagram of at least
 The crypto retransmission timer is not set if the time threshold
 {{time-threshold}} loss detection timer is set.  The time threshold loss
 detection timer is expected to both expire earlier than the crypto
-retransmission timeout and be much less likely to spuriously retransmit data.
+retransmission timeout and be less likely to spuriously retransmit data.
 The Initial and Handshake packet number spaces typically have a small number
 of packets in them, so time threshold loss detection will typically declare
 packets lost before packet threshold.
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -540,10 +540,12 @@ The crypto retransmission timer is not set if the time threshold
 {{time-threshold}} loss detection timer is set.  The time threshold loss
 detection timer is expected to both expire earlier than the crypto
 retransmission timeout and be much less likely to spuriously retransmit data.
-The Initial and Handshake packet number spaces are typically going to have a
-small number of packets in them, so it's expected packet threshold loss
-detection will rarely declare packets lost before time threshold. When the
-crypto retransmission timer is active, the probe timer ({{pto}}) is not active.
+The Initial and Handshake packet number spaces typically have a small number
+of packets in them, so time threshold loss detection will typically declare
+packets lost before packet threshold.
+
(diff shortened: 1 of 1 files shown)


yangchi:
Thanks for the explanation. It's much clearer.
Can i ask for one more clarification? In DetectLostPackets, shouldn't the for-loop have been something like this:
`foreach unacked in sent_packets[pn_space]:`

ianswett:
Thanks for catching the pseudocode error in DetectLostPackets.  PTAL.

yangchi:
lgtm. thank you!

draft-ietf-quic-recovery.md
```diff
@@ -537,8 +537,13 @@ and otherwise it MUST send an Initial packet in a UDP datagram of at least
 1200 bytes.
 
 The crypto retransmission timer is not set if the time threshold
-{{time-threshold}} loss detection timer is set.  When the crypto
-retransmission timer is active, the probe timer ({{pto}}) is not active.
+{{time-threshold}} loss detection timer is set.  The time threshold loss
+detection timer is both expected to expire earlier than the crypto
```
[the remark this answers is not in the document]
janaiyengar: Can you explain what this expectation is based on?

draft-ietf-quic-recovery.md
```diff
@@ -537,8 +537,13 @@ and otherwise it MUST send an Initial packet in a UDP datagram of at least
...
+retransmission timeout and be much less likely to spuriously retransmit data.
+The Initial and Handshake packet number spaces are typically going to have a
```
janaiyengar: I'm not sure what the number of packets in these spaces has to do with the description here about crypto timeout. Why is this sentence here?

ianswett: If the number of packets is small, it's unlikely you'll hit the packet threshold loss detection, making time threshold much more important.

ianswett: I tightened up the text, but feel free to suggest further clarification.

draft-ietf-quic-recovery.md
```diff
@@ -537,8 +537,15 @@ and otherwise it MUST send an Initial packet in a UDP datagram of at least
...
 
 The crypto retransmission timer is not set if the time threshold
-{{time-threshold}} loss detection timer is set.  When the crypto
-retransmission timer is active, the probe timer ({{pto}}) is not active.
+{{time-threshold}} loss detection timer is set.  The time threshold loss
+detection timer is expected to both expire earlier than the crypto
+retransmission timeout and be less likely to spuriously retransmit data.
+The Initial and Handshake packet number spaces typically have a small number
+of packets in them, so time threshold loss detection will typically declare
+packets lost before packet threshold.
```
janaiyengar: I think what you want to say is that time-threshold is more likely to be used for loss detection than packet-threshold, since there just might not be enough packets. So how about just this, given that you've just talked about time-threshold: "The Initial and Handshake packet number spaces will typically carry a small number of packets, so losses are less likely to be detected using packet-threshold detection."

That said, I still am not sure that it's worth mentioning the packet-threshold here at all.  But my opinion is not strong, so your call.

ianswett: Thanks, suggestion taken.  I think why you'd arm the loss-detection timer is non-obvious, so I thought more explanatory text was better than less, and I think the small number of packets argument is relevant.

Issue #2493, opened 2019-03-01
Awkward text in "Sending Probe Packets"


larseggert:
As discussed with @janaiyengar, 6.2.2.2 is awkward, because in the first paragraph it says
```
   When a PTO timer expires, the sender MUST send one ack-eliciting
   packet as a probe.
```
And then the very last paragraph changes that by saying
```
   When a PTO timer expires, new or previously-sent data may not be
   available to send and packets may still be in flight.  A sender can
   be blocked from sending new data in the future if packets are left in
   flight.  Under these conditions, a sender SHOULD mark any packets
   still in flight as lost.
```
What is super unclear is that the final paragraph apparently means to imply that when there is no new or unACKed data to send, *nothing* should be send, i.e., it invalidates the SHOULD in the first paragraph.

Related to that, the pseudo code still uses `SendTwoPackets()`, which is also inaccurate, since it's really "at least one (unless you don't have new or unACKed data) but can also send two". Good luck finding a good name for that function :-)

ianswett:
The pseudocode now uses SendOneOrTwoPackets(), but the last paragraph pretty clearly violates the previous MUST, which is very bad.

The simplest fix would be to change the MUST to a "SHOULD unless otherwise specified", but I'm increasingly thinking we should stick with the MUST and if you get into a situation you have nothing to send, require sending something ack-eliciting to ensure in-flight packets are removed from flight.

Thoughts?

PR #2622, opened 2019-04-15
MUST send if there is something to send

ianswett:
Editorial fix for #2493

This seems like the minimal fix for the existing text, though I wonder if it's the correct one.


janaiyengar:
I think this fix is fine, but it's inadequate. I'll take this in since it's useful.

ianswett:
Can you clarify why it's inadequate on the still open issue?

janaiyengar:
@ianswett: primarily because it does not correct the pseudocode, but I've also made some textual adjustments in #2624

PR #2623, opened 2019-04-15
My packet numbers go to -1

ianswett:
Alternate fix for #2611

Alternate PR is #2621


janaiyengar:
I don't think we should introduce negative numbers here. I understand that it makes the pseudo code slightly simpler, but it makes implementing the pseudo code much harder for those using uints.

marten-seemann:
@janaiyengar Why do you want to use uints? The varint encoding imposes a limit smaller than MAX_INT64 anyway.

PR #2624, opened 2019-04-15
Add details of marking packets lost on PTO

### LINKED ISSUE #2493: Awkward text in "Sending Probe Packets"

janaiyengar:
This PR adds text and pseudo-code to clarify what an endpoint should do when PTO fires and there is no data or other ack-eliciting frames to send.

Closes #2493.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -624,17 +624,20 @@ time.
 ### Sending Probe Packets
 
 When a PTO timer expires, a sender MUST send at least one ack-eliciting packet
-as a probe, unless there are no ack-eliciting frames to send.  An endpoint MAY
-send up to two ack-eliciting packets, to avoid an expensive consecutive PTO
-expiration due to a single packet loss.
+as a probe, unless there is no data available to send.  An endpoint MAY send up
+to two ack-eliciting packets, to avoid an expensive consecutive PTO expiration
+due to a single packet loss.
 
-It is possible that the sender has no new or previously-sent data or other
-ack-eliciting frames to send.  For example, this can happen when there is no new
-application data to send and all streams with data in flight are reset (see
-Section 13.2 of {{QUIC-TRANSPORT}}).  Under these conditions, the sender SHOULD
-mark any packets still in flight as lost.  The sender MAY send an ack-eliciting
-packet and re-arm the PTO timer instead, to avoid aggressively marking packets
-as lost (with the resulting congestion controller reaction).
+It is possible that the sender has no new or previously-sent data to send.  For
+example, this can happen when there is no new application data to send and all
+streams with data in flight are reset (see Section 13.2 of {{QUIC-TRANSPORT}}).
+Under these conditions, the sender SHOULD send a PING or another ack-eliciting
+frame, and re-arm the PTO timer.
+
+When there is no data to send, the sender MAY mark any packets still in flight
+as lost instead of sending an additional ack-eliciting packet.  Doing so might
+cause packets currently in flight to be aggressively marked as lost, with the
+resulting congestion controller reaction.
 
 Consecutive PTO periods increase exponentially, and as a result, connection
 recovery latency increases exponentially as packets continue to be dropped in
@@ -1245,17 +1248,9 @@ OnLossDetectionTimeout():
   else:
     // PTO
     pto_count++
-    if (any ack-eliciting frames available to send):
-      SendOneOrTwoPackets()
-    else:
-      // Mark all packets in flight as lost
-      foreach unacked in sent_packets:
-        sent_packets.remove(unacked.packet_number)
-        if (unacked.in_flight):
-          lost_packets.insert(unacked)
-      // Inform the congestion controller
-      if (!lost_packets.empty()):
-        OnPacketsLost(lost_packets)
+    // Send data if new data is available, else retransmit old data.
+    // If neither is available, send PING frames.
+    SendOneOrTwoPackets()
 
   SetLossDetectionTimer()
 ~~~
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1246,11 +1246,10 @@ OnLossDetectionTimeout():
        SendOnePaddedInitialPacket()
     crypto_count++
   else:
-    // PTO
-    pto_count++
-    // Send data if new data is available, else retransmit old data.
+    // PTO. Send new data if available, else retransmit old data.
     // If neither is available, send PING frames.
     SendOneOrTwoPackets()
+    pto_count++
 
   SetLossDetectionTimer()
 ~~~
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1251,7 +1251,7 @@ OnLossDetectionTimeout():
     crypto_count++
   else:
     // PTO. Send new data if available, else retransmit old data.
-    // If neither is available, send a PING frame.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -592,7 +592,7 @@ TCP's retransmission timeout period {{?RFC6298}}.
 
 ### Computing PTO
 
-When an ack-eliciting packet is transmitted, the sender schedules a timer for
```
ianswett: Can you undo the editorializing about sender/endpoint for this PR?  I prefer sender in this context, but I'd prefer to separate that question out.

draft-ietf-quic-recovery.md
```diff
@@ -611,22 +611,31 @@ delay sending an acknowledgement.
...
-probe, unless there is nothing to send. A sender MAY send up to two
-ack-eliciting packets, to avoid an expensive consecutive PTO expiration due
-to a single packet loss.
+When a PTO timer expires, an endpoint MUST send at least one ack-eliciting
+packet as a probe, unless there are no ack-eliciting frames to send.  An
+endpoint MAY send up to two ack-eliciting packets, to avoid an expensive
+consecutive PTO expiration due to a single packet loss.
+
+It is possible that the endpoint has no new or previously-sent data or other
+ack-eliciting frames to send.  This can happen for example when there is no new
```
```suggestion
ack-eliciting frames to send.  For example, this can happen when there is no new
```

draft-ietf-quic-recovery.md
```diff
@@ -1243,8 +1245,18 @@ OnLossDetectionTimeout():
     crypto_count++
   else:
     // PTO
-    SendOneOrTwoPackets()
     pto_count++
+    if (any ack-eliciting frames available to send):
+      SendOneOrTwoPackets()
+    else:
+      // Mark all packets in flight as lost
```
ianswett: This pseudocode makes an excellent argument that we should recommend sending a PING to establish loss.

janaiyengar: Why? Because you'd have to mark packets as lost?

janaiyengar:
@ianswett: I've done s/endpoint/sender. Let's deal with this separately. On the SHOULD, thinking about this, yes, I think I agree with you. I'll modify the PR, let's see how that looks.

janaiyengar:
Ok, modified, take a look

draft-ietf-quic-recovery.md
```diff
@@ -623,10 +623,21 @@ time.
...
-ack-eliciting packets, to avoid an expensive consecutive PTO expiration due
-to a single packet loss.
+When a PTO timer expires, a sender MUST send at least one ack-eliciting packet
+as a probe, unless there is no data available to send.  An endpoint MAY send up
+to two ack-eliciting packets, to avoid an expensive consecutive PTO expiration
+due to a single packet loss.
+
+It is possible that the sender has no new or previously-sent data to send.  For
+example, this can happen when there is no new application data to send and all
+streams with data in flight are reset (see Section 13.2 of {{QUIC-TRANSPORT}}).
```
ianswett: As Nick Banks has repeatedly pointed out, when you reset a stream, you send a STREAM_RESET frame, which is ack-eliciting, so this example doesn't really hold up that well.  A MESSAGE/DATAGRAM frame is the easiest real-world example because it's ack-eliciting and unreliable, but it's not in the v1 draft :(

nibanks: Maybe the best example is if the PTO timer expires and the only thing left outstanding was a spurious retransmission. For instance, packet 1 sent a stream frame, then you thought it was lost, then you sent packet 2 but then an ACK came for packet 1. Then PTO fires. You can mark that packet as possibly lost, but it wouldn't cause anything to be queued for transmit.

draft-ietf-quic-recovery.md
```diff
@@ -623,10 +623,21 @@ time.
...
+Under these conditions, the sender SHOULD send PING or other ack-eliciting
```
```suggestion
Under these conditions, the sender SHOULD send a PING or other ack-eliciting
```

draft-ietf-quic-recovery.md
```diff
@@ -623,10 +623,21 @@ time.
...
+frames in up to two packets, re-arming the PTO timer.
```
ianswett: Why send two packets in the case when you're just trying to establish loss?  I'd specifically say 1 packet for this case.

draft-ietf-quic-recovery.md
```diff
@@ -623,10 +623,21 @@ time.
...
+
+When there is no data to send, the sender MAY mark any packets still in flight
```
ianswett: I'd like to be clear that this is an alternative, so maybe say "Alternatively, when there is..."?

draft-ietf-quic-recovery.md
```diff
@@ -1242,7 +1246,8 @@ OnLossDetectionTimeout():
        SendOnePaddedInitialPacket()
     crypto_count++
   else:
-    // PTO
+    // PTO. Send new data if available, else retransmit old data.
+    // If neither is available, send PING frames.
```
```suggestion
    // If neither is available, send a PING frame.
```

draft-ietf-quic-recovery.md
```diff
@@ -1455,8 +1460,7 @@ Invoked when an ACK frame with an ECN section is received from the peer.
 
 ## On Packets Lost
 
-Invoked by loss detection from DetectLostPackets when new packets
-are detected lost.
+Invoked when packets are deemed lost.
```
ianswett: Is it worth specifying "from DetectLostPackets" still?

draft-ietf-quic-recovery.md
ianswett: Because of the reset issue, I think another example, such as Nick's, would be best.

ianswett:
As written now, this swaps a SHOULD and a MAY, so I'm changing it to design.

draft-ietf-quic-recovery.md
janaiyengar: Yeah, good point. Thanks for the alternate example, @nibanks.

draft-ietf-quic-recovery.md
```diff
@@ -623,10 +623,25 @@ time.
...
+an example, consider the following sequence of events: new application data is
+sent in a STREAM frame, deemed lost, then retransmitted in a new packet, and
+then the original transmission is acknowledged.  In the absence of any new
+application data, a PTO timer expiration now would find the sender with no new
+or previously-sent data to send.
+
+When there is no data to send, the sender SHOULD send a PING or other
+ack-eliciting frame in a single packet, re-arming the PTO timer.
+
+Alternatively, instead of sending the ack-eliciting packet, the sender MAY mark
```
```suggestion
Alternatively, instead of sending an ack-eliciting packet, the sender MAY mark
```

PR #2564, opened 2019-03-28
Updated spinbit text

ihlar:
#2364  got messed up with my latest update. 
Suggesting closing that PR and discussing it here instead.

PR #2628, opened 2019-04-17
Recommend of per-connection ID randomness for spin bit

mirjak:
This was discussed on the original PR. While it is perfectly okay for an endpoint to set a per-packet random value (because there is also nothing we could even do to avoid that), I strongly think that this document should RECOMMEND per-flow/conn-ID randomness.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3975,8 +3975,7 @@ measurement points.
 
 When the spin bit is disabled, endpoints MAY set the spin bit to any value, and
 MUST ignore any incoming value. It is RECOMMENDED that endpoints set the spin
-bit to a random value either chosen independently for each packet or chosen
-independently for each connection ID.
+bit to a random value chosen independently for each connection ID.
 
 If the spin bit is enabled for the connection, the endpoint maintains a spin
 value and sets the spin bit in the short header to the currently stored


janaiyengar:
@mirjak: I don't think this is editorial.  Can you file an issue?

mirjak:
If you don't think it's editorial, I don't think it was okay to merge this sentence without consensus.

MikeBishop:
@mnot, @larseggert, the proposed change here was discussed [previously](https://github.com/quicwg/base-drafts/pull/2564#discussion_r270914240) on #2564, but wasn't acted upon.  I read the previous comment as saying she disagrees that there was consensus to merge #2564 without addressing that discussion.  That makes this a larger issue than a single PR, and belongs to the chairs.

The proposed change is *technically* editorial, because it's guidance rather than a hard protocol requirement.  However, given the contentious nature of everything related to the spin bit, I'm inclined to think that this is worth at least running by the list and tracking with an issue.

martinthomson:
This seems more like a question of policy rather than protocol engineering.  Is there a specific issue that this addresses?

mnot:
@mirjak if you'd like the WG to consider this, please file an issue (preferred), or bring it up on the WG mailing list. While the editors do sometimes take purely editorial PRs, it's not how we discuss anything that's not purely editorial.

Editors, please do not label PRs with `design`.

martinthomson:
@mirjak please see the request in https://github.com/quicwg/base-drafts/pull/2628#issuecomment-486007420

PR #2634, opened 2019-04-18
Fix stale reference and edge case in GOAWAY text

MikeBishop:
Makes two minor fixes in the GOAWAY text, both editorial:
- Correctly references the MAX_STREAMS frame instead of the MAX_STREAM_ID frame
- Notes that servers need not (and cannot) send a GOAWAY if the client has used up all the possible stream IDs; the client simply won't be making any more requests.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1318,6 +1318,10 @@ connection to become idle, or MAY initiate an immediate closure of the
 connection.  An endpoint that completes a graceful shutdown SHOULD use the
 HTTP_NO_ERROR code when closing the connection.
 
+If a client has consumed all available bidirectional stream IDs with requests,
+the server need not send a GOAWAY frame; the client is unable to make further
+requests.
+
 ## Immediate Application Closure
 
 An HTTP/3 implementation can immediately close the QUIC connection at any time.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -1318,6 +1318,10 @@ connection to become idle, or MAY initiate an immediate closure of the
 connection.  An endpoint that completes a graceful shutdown SHOULD use the
 HTTP_NO_ERROR code when closing the connection.
 
+If a client has consumed all available bidirectional stream IDs with requests,
```
ianswett: This is true, but would the HTTP layer know this?

MikeBishop: It has to be tracking the greatest stream ID it has received a request on in order to put that value in a GOAWAY frame.  So, yes.  Another implementation might be to retrieve this number and then find that it's out-of-bounds, but it amounts to the same thing.

draft-ietf-quic-http.md
```diff
@@ -1318,6 +1318,10 @@ connection to become idle, or MAY initiate an immediate closure of the
...
+the server need not send a GOAWAY frame; the client is unable to make further
```
```suggestion
the server need not send a GOAWAY frame, since the client is unable to make further
```

PR #2635, opened 2019-04-19
Targeted attacks in Security Considerations

### LINKED ISSUE #2026: Can Initial/0-RTT CIDs safely be used for routing?

MikeBishop:
I went to fix #2026 and found that the offending text was already removed, so we're most of the way there.  I added a note to the Security Considerations about not letting clients dictate your routing decisions.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -5385,11 +5385,11 @@ define a mechanism that is robust against version downgrade attacks.
 ## Targeted Attacks by Routing
 
 Deployments should limit the ability of an attacker to target a new connection
-to a particular server instance.  This means that client-controlled fields, such as the
-initial Destination Connection ID used on Initial and 0-RTT packets SHOULD NOT
-be used by themselves to make routing decisions.  Ideally, routing decisions are made
-independently of client-selected values; a Source Connection ID can be selected
-to route later packets to the same server.
+to a particular server instance.  This means that client-controlled fields, such
+as the initial Destination Connection ID used on Initial and 0-RTT packets
+SHOULD NOT be used by themselves to make routing decisions.  Ideally, routing
+decisions are made independently of client-selected values; a Source Connection
+ID can be selected to route later packets to the same server.
 
 # IANA Considerations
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -5382,6 +5382,14 @@ Negotiation packets do not contain any mechanism to prevent version downgrade
 attacks.  Future versions of QUIC that use Version Negotiation packets MUST
 define a mechanism that is robust against version downgrade attacks.
 
+## Targeted Attacks by Routing
+
+Deployments should limit the ability of an attacker to target a new connection
+to a particular instance.  This means that client-controlled fields, such as the
```
```suggestion
to a particular server instance.  This means that client-controlled fields, such as the
```

draft-ietf-quic-transport.md
```diff
@@ -5382,6 +5382,14 @@ Negotiation packets do not contain any mechanism to prevent version downgrade
...
+initial Destination Connection ID used on Initial and 0-RTT packets SHOULD NOT
+be used to make routing decisions.  Ideally, routing decisions are made
```
```suggestion
be used by themselves to make routing decisions.  Ideally, routing decisions are made
```

draft-ietf-quic-transport.md
```diff
@@ -5382,6 +5382,14 @@ Negotiation packets do not contain any mechanism to prevent version downgrade
...
+independently of client-selected values; a Source Connection ID can be selected
+to route subsequent packets to the same server.
```
```suggestion
to route later packets to the same server.
```

PR #2642, opened 2019-04-23
Don't send PING in Initial or Handshake

### LINKED ISSUE #2640: client should send PING in Initial or Handshake, but PINGs are forbidden in these packets
The [text about the crypto timer](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#crypto-retransmission-timeout) says:
> If the crypto retransmission timer expires before the client has 1-RTT keys, it is possible that the client may not have any crypto data to retransmit. However, the client MUST send a new packet, containing only PING or PADDDING frames if necessary, to allow the server to continue sending data. If Handshake keys are available to the client, it MUST send a Handshake packet, and otherwise it MUST send an Initial packet in a UDP datagram of at least 1200 bytes.

PING frames are forbidden in both Initial and Handshake packets.

ianswett:
Fixes #2640

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -538,11 +538,10 @@ MUST ensure that the crypto retransmission timer is set if there is
 unacknowledged crypto data or if the client does not yet have 1-RTT keys.
 If the crypto retransmission timer expires before the client has 1-RTT keys,
 it is possible that the client may not have any crypto data to retransmit.
-However, the client MUST send a new packet, containing only PING or PADDING
-frames if necessary, to allow the server to continue sending data. If
-Handshake keys are available to the client, it MUST send a Handshake packet,
-and otherwise it MUST send an Initial packet in a UDP datagram of at least
-1200 bytes.
+However, the client MUST send a new packet, containing only PADDING frames if
+necessary, to allow the server to continue sending data. If Handshake keys
+are available to the client, it MUST send a Handshake packet, and otherwise
+it MUST send an Initial packet in a UDP datagram of at least 1200 bytes.
 
 The crypto retransmission timer is not set if the time threshold
 {{time-threshold}} loss detection timer is set.  The time threshold loss
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -543,6 +543,10 @@ necessary, to allow the server to continue sending data. If Handshake keys
 are available to the client, it MUST send a Handshake packet, and otherwise
 it MUST send an Initial packet in a UDP datagram of at least 1200 bytes.
 
+Because packets only containing PADDING do not elicit an acknowledgement,
+they may never be acknowledged, but they are removed from bytes in flight
+when the client gets Handshake keys and the Initial keys are discarded.
+
 The crypto retransmission timer is not set if the time threshold
 {{time-threshold}} loss detection timer is set.  The time threshold loss
 detection timer is expected to both expire earlier than the crypto


martinthomson:
It's probably worth acknowledging the edge case and just saying that this protocol doesn't do anything about that.  Note that the case we're concerned about should involve the server sending tons of data, which gives it plenty of opportunity to provide the client with an ACK that will unstick it.

janaiyengar:
LGTM, once you've added the note about congestion.

martinthomson:
Still holding on a note about congestion.  I will merge and open another issue if that doesn't happen soon-ish.

ianswett:
I'll do that today.

ianswett:
I added  description of why I believe this is not a problem.  Conveniently, the logic is simpler than I previously stated, because of the fast Initial -> Handshake transition when the client gets Handshake keys.

ianswett:
Now that I've started a PR to combine the crypto handshake timer and PTO, I'm concerned I chose the wrong fix for #2640 in this PR, and it would be better to allow PING in Initial.

In the existing PTO case when there are bytes in flight, but nothing to send, we say "When there is no data to send, the sender SHOULD send a PING or other ack-eliciting frame in a single packet, re-arming the PTO timer."

If we allow PING in Initial, that text will work fine with a combined timer, since all Initials must be in padded datagrams.  Otherwise, combining them adds a slightly odd special case.

@janaiyengar I'd like your thoughts on this in particular.

draft-ietf-quic-recovery.md
```diff
@@ -538,11 +538,14 @@ MUST ensure that the crypto retransmission timer is set if there is
...
-and otherwise it MUST send an Initial packet in a UDP datagram of at least
-1200 bytes.
+However, the client MUST send a new packet, containing only PADDING frames if
+necessary, to allow the server to continue sending data. If Handshake keys
+are available to the client, it MUST send a Handshake packet, and otherwise
+it MUST send an Initial packet in a UDP datagram of at least 1200 bytes.
+
+Because packets only containing PADDING do not elicit an acknowledgement,
+they may never be acknowledged, but they are removed from bytes in flight
+when the client gets Handshake keys and the Initial keys are discarded.
```
janaiyengar: Suggested rewording:
"Because the server will eventually discard Initial and Handshake keys, and because packets containing only PADDING frames do not elicit an acknowledgement, it is possible that no acknowledgement is ever received for these additional packets.  They are however eventually removed from the count of bytes in flight when the client discards the corresponding keys (see {{discarding-keys}})."

janaiyengar:
@ianswett: Good catch! It's useful to go back to the principle we have in place: if we want a PTO running, we need to have an ack-eliciting packet outstanding. I don't think there's a strong enough argument to create an exception here, so I agree that we should change this to always sending a PING when the timer goes off and there's nothing to send.

Since the title here is quite clearly the opposite of that, in the interest of transparency, I propose that you close this PR in favor of a new one that says send PING (and stack this new PR on top of #2655).

janaiyengar:
Um. Ok. This is attached to an issue that is labeled `transport` and now has consensus declared on it with this PR. I'm fried right now, but let's discuss this before doing anything.

mikkelfj:
In my view the point of not sending PING early is to reduce the complexity of the early packet number spaces and allow special code to handle these. The fewer types of packets and ACK's here the better. However, if PING ends up simplifying this logic, then certainly there is an argument for that. The question then is: would PING' complicate ACK logic in early packet number spaces? That would only really be the case if there were no ACK's otherwise since PING's are easy to handle. 

OTOH - if it is only a matter of cleaning up two paragraphs in the spec it may not be worthwhile. An exception may be the natural thing if the PTO logic is in fact separate anyway in the different PN spaces.

martinthomson:
Note that I didn't include this in the changelog for -21.  It seems entirely editorial to my reading.

PR #2655, opened 2019-04-27
Merge crypto timeout into PTO

### LINKED ISSUE #2648: Crypto timeout value should include rttvar
Not sure why we don't already.

### LINKED ISSUE #2650: Recovery would be simpler if the Handshake timer was unified with the PTO timer
The main optimization the handshake timer makes is the assumption that max_ack_delay does not need to be waited for before firing.  However, it also uses a completely different equation for a timeout (2 * SRTT vs SRTT + 4*RTTVar).  We could fix the max_ack_delay difference by defaulting max_ack_delay to 0(#2646), like almost all other transport parameters.

If there are no RTT measurements or estimates, we could use 1 second.  If there are no RTT measurements, but there is an estimate, we could either use the PTO formula and assume RTTVar is 1/2 RTT(=> 3*RTT), or we could recommend estimating RTTVar as well as RTT.  The first approach seems more practical.

Another notable difference is the advice to retransmit all the handshake packets instead of 1 or 2, but given the current 3x amplification limit, if a server sends 2 packets in response to a ClientHello, they can't even retransmit both packets, so advising that they retransmit all handshake packets is not indicative of real-world behavior.  If the server only has 1 handshake packet, then there's no change.  The client should always be able to fit its CRYPTO data into a single packet, unless there are large client certs.  If there are large client certs, one would expect time or packet threshold loss recovery to be faster and more effective than timer based retransmissions.  And at that point, the client should have at least one RTT sample.

The goal of this issue is to make the recovery mechanisms simpler, as well as including RTTVar in the handshake timer, as #2648 suggests.

ianswett:
Makes the crypto timeout(including deadlock prevention) a special case of PTO.

Fixes #2650, Fixes #2648 

I expect some cleanup afterwards, or we can try to ensure consistency in this PR.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1173,10 +1173,10 @@ SetLossDetectionTimer():
     // Time threshold loss detection.
     loss_detection_timer.update(loss_time)
     return
-
+    
   if (has unacknowledged crypto data
       || endpoint is client without 1-RTT keys):
-    // Crypto retransmission timer.
+      // Crypto retransmission deadlock timer.
     if (smoothed_rtt == 0):
       timeout = 2 * kInitialRtt
     else:
@@ -1185,12 +1185,6 @@ SetLossDetectionTimer():
     timeout = timeout * (2 ^ crypto_count)
     loss_detection_timer.update(
       time_of_last_sent_crypto_packet + timeout)
-    return
-
-  // Don't arm timer if there are no ack-eliciting packets
-  // in flight.
-  if (no ack-eliciting packets in flight):
-    loss_detection_timer.cancel()
     return
 
   // Calculate PTO duration
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -500,102 +500,7 @@ thresholds reduce reordering resilience and increase spurious retransmissions,
 and larger thresholds increase loss detection delay.
 
 
-## Crypto Retransmission Timeout
-
-Data in CRYPTO frames is critical to QUIC transport and crypto negotiation, so a
-more aggressive timeout is used to retransmit it.
-
-The initial crypto retransmission timeout SHOULD be set to twice the initial
-RTT.
-
-At the beginning, there are no prior RTT samples within a connection.  Resumed
-connections over the same network SHOULD use the previous connection's final
-smoothed RTT value as the resumed connection's initial RTT.  If no previous RTT
-is available, or if the network changes, the initial RTT SHOULD be set to 500ms,
-resulting in a 1 second initial handshake timeout as recommended in
-{{?RFC6298}}.
-
-When a crypto packet is sent, the sender MUST set a timer for twice the smoothed
-RTT.  This timer MUST be updated when a new crypto packet is sent and when
-an acknowledgement is received which computes a new RTT sample. Upon timeout,
-the sender MUST retransmit all unacknowledged CRYPTO data if possible.  The
-sender MUST NOT declare in-flight crypto packets as lost when the crypto timer
-expires.
-
-On each consecutive expiration of the crypto timer without receiving an
-acknowledgement for a new packet, the sender MUST double the crypto
-retransmission timeout and set a timer for this period.
-
-Until the server has validated the client's address on the path, the amount of
-data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -689,6 +597,69 @@ is received that newly acknowledges packets, loss detection proceeds as
...
+CRYPTO data sent in Initial packets should be retransmitted.  If no data can be
+sent, then no alarm should be armed until data has been received from the
+client.
+
+Because the server could be blocked until more packets are received, the client
+MUST ensure that the crypto retransmission timer is set if there is
+unacknowledged crypto data or if the client does not yet have 1-RTT keys.
+If the crypto retransmission timer expires before the client has 1-RTT keys,
+it is possible that the client may not have any crypto data to retransmit.
+However, the client MUST send a new packet, containing only PING or PADDING
```
```suggestion
However, the client MUST send a new packet, containing only PADDING
```
PING is not valid nor necessary here, no?

draft-ietf-quic-recovery.md
```diff
@@ -689,6 +597,69 @@ is received that newly acknowledges packets, loss detection proceeds as
...
+When 0-RTT is rejected, recovery state for all in-flight 0-RTT packets is
+discarded.
+
+If a server accepts 0-RTT, but does not buffer 0-RTT packets that arrive
+before Initial packets, early 0-RTT packets will be declared lost, but that
+is expected to be infrequent.
+
+It is expected that keys are discarded after packets encrypted with them would
+be acknowledged or declared lost.  Initial secrets however might be destroyed
+sooner, as soon as handshake keys are available (see Section 4.10 of
```
mikkelfj: Discarding keys based on being declared lost seems odd to me. Either you abandon the handshake, or you discard keys when you are ready to advance to the next PN space, however that is defined. That readiness might involve overlapped PN spaces, waiting for older packets to ACK or to declare lost, but loss alone cannot be the criteria.

draft-ietf-quic-recovery.md
```diff
@@ -689,6 +597,69 @@ is received that newly acknowledges packets, loss detection proceeds as
...
+The initial PTO SHOULD be set to twice the initial RTT or 1 second if no
+initial RTT is available as recommended in {{?RFC6298}}.
+
+Until the server has validated the client's address on the path, the amount of
+data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
+If not all unacknowledged CRYPTO data can be sent, then all unacknowledged
+CRYPTO data sent in Initial packets should be retransmitted.  If no data can be
+sent, then no alarm should be armed until data has been received from the
+client.
+
```
mikkelfj: If no alarm is armed, the connection is at the mercy of the client. I supposed connection timeout is not armed yet, and also the wrong metric, but the server needs to be able to time out the connection attempt somehow.

On a related note, text below on Retry/VNEG says they do not ack an Initial, but I suppose they could be used to time out connection state server side?

draft-ietf-quic-recovery.md
janaiyengar: Fair point. Yeah, I think only PADDING is fine.

draft-ietf-quic-recovery.md
janaiyengar: This is pre-existing text, not new to this PR. The point however here isn't that you discard keys based on declaring packets lost, but that it is generally expected to be the case that packets are already declared lost (or acked) by the time keys are discarded.

draft-ietf-quic-recovery.md
janaiyengar: Yes, I meant to come to this -- thanks for catching it. This is not specific to this PR, so I've filed #2656 for it.

draft-ietf-quic-recovery.md
```diff
@@ -687,6 +620,69 @@ be acknowledged or declared lost.  Initial secrets however might be destroyed
...
+packet indicates that the Initial was received but not processed.  Neither
+packet can be treated as an acknowledgment for the Initial.
+
+The client MAY however compute an RTT estimate to the server as the time period
+from when the first Initial was sent to when a Retry or a Version Negotiation
+packet is received.  The client MAY use this value to seed the RTT estimator for
+a subsequent connection attempt to the server.
+
+
+## Discarding Keys and Packet State {#discarding-packets}
```
martinthomson: Does this really belong under a section titled "Probe Timeout"?

janaiyengar: Agreed, this section should be promoted.

PR #2665, opened 2019-05-02
Reference 8.2 of RFC7540 to explain Server Push

ianswett:
The transition to "Push Streams" has no preceding mention of server push.  This adds a reference to the correct section of RFC7540.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -944,6 +944,10 @@ able to send stream data first after the cryptographic handshake completes.
 
 ### Push Streams
 
+Server push is an optional feature introduced in HTTP/2 that allows a server to
+initiate a response before a request has been made.  See section 8.2 of
+{{!RFC7540}} for more details.
+
 A push stream is indicated by a stream type of `0x01`, followed by the Push ID
 of the promise that it fulfills, encoded as a variable-length integer. The
 remaining data on this stream consists of HTTP/3 frames, as defined in
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -948,8 +948,8 @@ able to send stream data first after the cryptographic handshake completes.
 ### Push Streams
 
 Server push is an optional feature introduced in HTTP/2 that allows a server to
-initiate a response before a request has been made.  See Section 8.2 of
-{{!RFC7540}} for more details.
+initiate a response before a request has been made.  See {{server-push}} for
+more details.
 
 A push stream is indicated by a stream type of `0x01`, followed by the Push ID
 of the promise that it fulfills, encoded as a variable-length integer. The
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -948,7 +948,7 @@ able to send stream data first after the cryptographic handshake completes.
 ### Push Streams
 
 Server push is an optional feature introduced in HTTP/2 that allows a server to
-initiate a response before a request has been made.  See section 8.2 of
+initiate a response before a request has been made.  See Section 8.2 of
 {{!RFC7540}} for more details.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -370,6 +370,10 @@ able to send stream data first after the cryptographic handshake completes.
 
 ### Push Streams
 
+Server push is an optional feature introduced in HTTP/2 that allows a server to
+initiate a response before a request has been made.  See section 8.2 of
+{{!RFC7540}} for more details.
```
janaiyengar: That's not quite accurate... since the response has to be associated with a previous request. I don't think that level of detail is necessary though. How about "The Server Push feature of HTTP/2 (see Section 8.2 of {{!RFC7540}}) is facilitated in HTTP/3 via push streams."

LPardue:
If mention of server push is desired before mention of push stream, would it be simpler to move the final sentence of the paragraph to the start?

> Server push and Push IDs are described in Section 5.4.

draft-ietf-quic-http.md
mikkelfj: I'm not too fond of relying on HTTP/2 to understand HTTP/3 beyond - "see this list header and verbs".

MikeBishop: Understandably -- it's a line we have to walk.  (Note that HTTP/2 defines no verbs and pseudo-no headers.)

My principle has been that I'm trying to weed out assumptions that you've read RFC7540 and understand a given concept already.  I'm happy to save redundant and possibly-conflicting text by pointing to a specific section of RFC7540, however.

ianswett:
Maybe this would be better if section 4 was moved after section 6?  Then the concepts would be explained before going into the framing.

I'd also like to reorder/renumber all the push frame types so they're all adjacent to one another and in a sensible order.  Is there a reason they're numbered as they are?  If not, I'll open an issue to renumber them.

MikeBishop:
There's already an issue to reorder some of the sections, so I'm happy to consider this as part of the same discussion.  As to the numbering, the primary reason is maintaining the same frame IDs for things that stayed over from HTTP/2 -- PUSH_PROMISE has the same ID, CANCEL_PUSH has the ID that HTTP/2 used for RST_STREAM, and DUPLICATE_PUSH is brand new and so pushes into unallocated space.

MikeBishop:
Given that #2701 is in now, do you still see this as a problem?

ianswett:
I think the Server Push section still lacks a simple explanation of what the push does.  A sentence like what I added in this PR at the beginning would help that.  Currently the Server Push section goes directly into mechanics.

https://github.com/quicwg/base-drafts/blob/master/draft-ietf-quic-http.md#server-push

"HTTP/3 server push is similar to what is described in HTTP/2 {{!HTTP2}}, but uses different mechanisms.

Each server push is identified by a unique Push ID. ..."

I'd add more to the first paragraph in case people don't want to read the H2 draft.

MikeBishop:
I think I'd prefer for the reference to be to 4.4 of the current draft instead, since that does point to RFC7540 and has more context as well.

MikeBishop:
Whoops -- which also doesn't have the two-sentence explanation of what push is.  That's up in the intro.  Let me copy that over to 4.4.

draft-ietf-quic-http.md
```diff
@@ -944,6 +947,10 @@ able to send stream data first after the cryptographic handshake completes.
 
 ### Push Streams
 
+Server push is an optional feature introduced in HTTP/2 that allows a server to
+initiate a response before a request has been made.  See section 8.2 of
```
```suggestion
initiate a response before a request has been made.  See Section 8.2 of
```

ianswett:
Thanks for the added text Mike, it looks great.  I'm happy with this, so it's up to you to decide when this is ready to merge.

PR #2667, opened 2019-05-04
Recovery Editorial Nits

ianswett:
From a full read en route to Prague.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -219,7 +219,7 @@ not available.
 QUIC ends a loss epoch when a packet sent after loss is declared is
 acknowledged. TCP waits for the gap in the sequence number space to be filled,
 and so if a segment is lost multiple times in a row, the loss epoch may not
-end. Because both reduce CWND only once per epoch,  QUIC may reduce
+end for several round trips. Because both reduce CWND only once per epoch,  QUIC will reduce its
 congestion window correctly once for every round trip that experiences loss, while TCP may only reduce
 the congestion window once.
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -219,9 +219,9 @@ not available.
 QUIC ends a loss epoch when a packet sent after loss is declared is
 acknowledged. TCP waits for the gap in the sequence number space to be filled,
 and so if a segment is lost multiple times in a row, the loss epoch may not
-end for several round trips. Because both reduce CWND only once per epoch,
-QUIC will reduce its congestion window correctly once for every round trip that
-experiences loss, while TCP may only reduce the congestion window once.
+end for several round trips. Because both should reduce their congestion windows
+only once per epoch, QUIC will do it correctly once for every round trip that
+experiences loss, while TCP may only do it once across multiple round trips.
 
 ### No Reneging
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -217,9 +217,9 @@ not available.
 ### Clearer end to a loss epoch
 
 QUIC may reduce cwnd several times on successive losses whereas TCP will keep
-cwnd constant until (longer) recovery exit or an RTO occurs, because QUIC uses
-packet numbers to determine when a loss epoch has ended and TCP uses packet
-sequence numbers.
+cwnd constant until forward progress is made or an RTO occurs. This difference
+is because QUIC uses packet numbers to determine when a loss epoch has ended
+and TCP uses packet sequence numbers.
 
 ### No Reneging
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -118,8 +118,8 @@ ACK-only:
 In-flight:
 
 : Packets are considered in-flight when they have been sent
-  and neither acknowledged nor declared lost, and they are not
-  ACK-only.
+  and neither acknowledged nor declared lost nor abandoned due
+  to keys being discarded, and they are not ACK-only.
```
janaiyengar: suggestion: "Packets are considered in-flight when they have been sent and are not ACK-only, and they are not acknowledged, declared lost, or abandoned along with old keys."

draft-ietf-quic-recovery.md
```diff
@@ -214,6 +214,13 @@ Most TCP mechanisms implicitly attempt to infer transmission ordering based on
 TCP sequence numbers - a non-trivial task, especially when TCP timestamps are
 not available.
 
+### Clearer end to a loss epoch
```
```suggestion
### Clearer Loss Epoch
```

draft-ietf-quic-recovery.md
```diff
@@ -214,6 +214,13 @@ Most TCP mechanisms implicitly attempt to infer transmission ordering based on
...
+
+QUIC may reduce cwnd several times on successive losses whereas TCP will keep
```
janaiyengar: s/cwnd/congestion window

Also, this doesn't match the title of this section. I would not talk about the cwnd reduction here, instead focus on what a loss epoch is -- that no congestion window action should occur more than once per roundtrip time, since a controller only sees the response to any action it takes about a roundtrip time later. TCP and QUIC use packet numbers to figure out when this roundtrip ends, but QUIC does it more accurately than TCP, especially when multiple losses occur within the same roundtrip time.

draft-ietf-quic-recovery.md
```diff
@@ -469,15 +476,16 @@ as TCP-NCR {{?RFC4653}}, to improve QUIC's reordering resilience.
 
 ### Time Threshold {#time-threshold}
 
-Once a later packet has been acknowledged, an endpoint SHOULD declare an earlier
-packet lost if it was sent a threshold amount of time in the past. The time
-threshold is computed as kTimeThreshold * max(SRTT, latest_RTT).
+Once a later packet has been acknowledged in the same packet number space, an
```
```suggestion
Once a later packet within the same packet number space has been acknowledged, an
```

draft-ietf-quic-recovery.md
```diff
@@ -469,15 +476,16 @@ as TCP-NCR {{?RFC4653}}, to improve QUIC's reordering resilience.
...
+endpoint SHOULD declare an earlier packet lost if it was sent a threshold amount
+of time in the past. To avoid declaring packets as lost too early, the time
```
```suggestion
of time in the past. To avoid declaring packets as lost too early, this time
```

draft-ietf-quic-recovery.md
```diff
@@ -915,7 +919,7 @@ It is expected that implementations will be able to access this information by
 packet number and crypto context and store the per-packet fields
 ({{sent-packets-fields}}) for loss recovery and congestion control.
 
-After a packet is declared lost, it SHOULD be tracked for an amount of time
+After a packet is declared lost, it could be tracked for an amount of time
```
```suggestion
After a packet is declared lost, the endpoint can track it for an amount of time
```

draft-ietf-quic-recovery.md
ianswett: Thanks, PTAL.

ianswett:
Thanks Jana, suggestions taken and "Clearer Loss Epoch" rewritten.

draft-ietf-quic-recovery.md
```diff
@@ -214,6 +214,15 @@ Most TCP mechanisms implicitly attempt to infer transmission ordering based on
 TCP sequence numbers - a non-trivial task, especially when TCP timestamps are
 not available.
 
+### Clearer Loss Epoch
+
+QUIC ends a loss epoch when a packet sent after loss is declared is
+acknowledged. TCP waits for the gap in the sequence number space to be filled,
+and so if a segment is lost multiple times in a row, the loss epoch does not
```
```suggestion
and so if a segment is lost multiple times in a row, the loss epoch may not
```

draft-ietf-quic-recovery.md
```diff
@@ -214,6 +214,15 @@ Most TCP mechanisms implicitly attempt to infer transmission ordering based on
...
+end. Because both reduce CWND only once per epoch,  QUIC may reduce
```
```suggestion
end for several round trips. Because both reduce CWND only once per epoch,  QUIC will reduce its
```

draft-ietf-quic-recovery.md
```diff
@@ -214,6 +214,15 @@ Most TCP mechanisms implicitly attempt to infer transmission ordering based on
...
+congestion window multiple times in the same circumstances TCP only reduces
```
```suggestion
congestion window correctly once for every round trip that experiences loss, while TCP may only reduce
```

PR #2668, opened 2019-05-06
QPACK might be part of the exchange, but not

### LINKED ISSUE #2657: HTTP/3 message exchange ignores QPACK table updates
This arises from a different discussion I had about the message exchange definition in section 5.1. The description is correct but glosses over the detail of QPACK, which is mentioned very briefly later in section 5.1.1.

For complete accuracy, I would argue that a single message exchange may also contain one or more QPACK encoder instructions sent on the encoder stream.

MikeBishop:
Fixes #2657.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -942,9 +942,9 @@ A server MAY interleave one or more PUSH_PROMISE frames (see
 PUSH_PROMISE frames are not part of the response; see {{server-push}} for more
 details.
 
-The HEADERS and PUSH_PROMISE frames might reference concurrent updates to the
-QPACK table.  While these updates are not directly part of the message exchange,
-they must be received and processed before the message can be consumed.  See
+The HEADERS and PUSH_PROMISE frames might reference updates to the QPACK table.
+While these updates are not directly part of the message exchange, they must be
+received and processed before the message can be consumed.  See
 {{header-formatting}} for more details.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -942,6 +942,11 @@ A server MAY interleave one or more PUSH_PROMISE frames (see
 PUSH_PROMISE frames are not part of the response; see {{server-push}} for more
 details.
 
+The HEADERS and PUSH_PROMISE frames might reference updates to the QPACK table
+which are sent at the same time.  While these updates are not directly part
```
LPardue: Maybe the first sentence could say something like

"The HEADERS and PUSH_PROMISE frames might referenced synchronised updates to the QPACK dynamic table."

martinthomson: or "HEADERS and PUSH_PROMISE frames might reference[!] concurrent updates to the QPACK dynamic table"

draft-ietf-quic-http.md
```diff
@@ -942,6 +942,11 @@ A server MAY interleave one or more PUSH_PROMISE frames (see
 PUSH_PROMISE frames are not part of the response; see {{server-push}} for more
 details.
 
+The HEADERS and PUSH_PROMISE frames might reference concurrent updates to the
+QPACK table.  While these updates are not directly part of the message exchange,
```
afrind: I agree with Martin that we should say 'QPACK dynamic table' here.  

I'm not crazy about 'concurrent'.  What if we just said 'might reference updates'?

PR #2674, opened 2019-05-08
Define abandonment

### LINKED ISSUE #2671: Clarify "abandon" in the context of version negotiation
The draft currently reads:
"When a client receives a Version Negotiation packet, it MUST abandon the current connection attempt."
However, it does not define what it is to abandon a connection attempt. Is it an abrupt termination without writing a connection close? That probably makes the most sense since the server is meant to be stateless at this stage.

martinthomson:
Just in case it wasn't perfectly clear from the choice of words.

Closes #2671.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1130,6 +1130,8 @@ This system allows a server to process packets with unsupported versions without
 retaining state.  Though either the Initial packet or the Version Negotiation
 packet that is sent in response could be lost, the client will send new packets
 until it successfully receives a response or it abandons the connection attempt.
+As a result, the client discards all state for the connection and does not send
+any more packets.
 
 A server MAY limit the number of Version Negotiation packets it sends.  For
 instance, a server that is able to recognize packets as 0-RTT might choose not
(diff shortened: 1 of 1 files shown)


martinthomson:
I know how you feel.  We'll be defining "the" eventually.  I don't know where the line is, but I think that it helps in this case to be doubly explicit.

marten-seemann:
I'm wondering how one could possibly misinterpret "abandoning" here. Furthermore, I don't think it's relevant for the server here if the client actually discarded the connection state, as long as it doesn't any more packets.

draft-ietf-quic-transport.md
```diff
@@ -1130,6 +1130,8 @@ This system allows a server to process packets with unsupported versions without
 retaining state.  Though either the Initial packet or the Version Negotiation
 packet that is sent in response could be lost, the client will send new packets
 until it successfully receives a response or it abandons the connection attempt.
+As a result, the client discards all state for the connection and does not send
+any more packets.
```
```suggestion
any more packets on the connection.
```

janaiyengar:
I don't think we need to spend a lot of energy here. I agree that this doesn't seem necessary, but I'm not going to fight it either.

PR #2675, opened 2019-05-08
Define under-utilizing the congestion window

### LINKED ISSUE #2630: Define "under-utilization" of cwnd
#2605 changed the discussion in recovery from idle behavior and app limited periods to  periods of under-utilized cwnd. This is great, but a precise definition of when the cwnd goes under-utilized is still missing.

### LINKED ISSUE #2686: Pacing when under-utilizing the Congestion Window
"Bursting more than an intial window’s worth of data into the network might cause short-term congestion and losses."
• Is that really "more than...", I think the evidence was that a larger IW (as in QUIC) can lead to congestion.
• We could explain burst size and scenarios, or could simply say something like:
"Bursting an initial window’s worth of data into the network can cause short-term congestion and losses for some path conditions."
(This also fixes /intial/initial)?

ianswett:
Fixes #2630, Fixes #2686 

I believe this also fixes the remaining portion of #2555


draft-ietf-quic-recovery.md
```diff
@@ -836,9 +836,11 @@ in Linux (3.11 onwards).
 
 ## Under-utilizing the Congestion Window
 
-A congestion window that is under-utilized SHOULD NOT be increased in either
-slow start or congestion avoidance. This can happen due to insufficient
-application data or flow control credit.
+When bytes in flight is smaller than the congestion window and sending is not
+pacing limited, the congestion window is under-utilized.  When this occurs,
```
marten-seemann: Where does the pacing limitation come from?

ianswett: There are congestion controllers now, and hopefully more in the future, that operate primarily pacing limited, not CWND limited, so this was an effort to capture that, but it's certainly imperfect.  Any suggestions?

draft-ietf-quic-recovery.md
```diff
@@ -848,9 +850,10 @@ and not fully utilize the congestion window due to this delay. A sender
 should not consider itself application limited if it would have fully
 utilized the congestion window without pacing delay.
 
-Bursting more than an intial window's worth of data into the network might
-cause short-term congestion and losses. Implemementations SHOULD either use
-pacing or reduce their congestion window to limit such bursts.
+Sending multiple packets into the network without any delay between them
+creates a burst of load that might cause short-term congestion and losses.
```
```suggestion
creates a packet burst that might cause short-term congestion and losses.
```

draft-ietf-quic-recovery.md
```diff
@@ -848,9 +850,10 @@ and not fully utilize the congestion window due to this delay. A sender
...
+Implemementations SHOULD either use pacing or reduce their congestion window
```
```suggestion
Implementations SHOULD either use pacing or reduce their congestion window
```

draft-ietf-quic-recovery.md
janaiyengar: I don't think this description of burst is necessary though. It doesn't help with #2686, if that's what you were trying to address. I would drop this, but I realized that the existing statement is potentially misleading in what it says. (more discussion on this in #2686)

ianswett: I added the definition of burst, because I realize it might not be obvious?

I think the core of #2686 is that the existing text is unclear.  The existing text says you SHOULD not burst more than an initial window into the network(ever), with nothing specific to coming out of quiescence.

janaiyengar: >  The existing text says you SHOULD not burst more than an initial window into the network(ever), with nothing specific to coming out of quiescence.

Yes, I think this is the problem. I believe the intent here was to say that a sender SHOULD NOT do more than IW as an upper limit on bursts without recommending that as the correct limit for any bursting.

We can either specify that _all_ bursts should be limited to IW, irrespective of how much data is in the network already, or that bursts coming out of quiescence should be limited to IW. While I'm not sure how to change this text here, I'm trying to steer clear of saying something overly broad. (I'll suggest some text).

mikkelfj: If not out of quiescence, an IW burst would probably be bad. You could just wait a millisecond or two and repeat in badly designed pacing logic.

ianswett: I'd rather have it apply always, because I don't want to have one packet sent, and then a sender can burst the full window.  Making it specific to quiescence adds unnecessary complexity to the rules and may not accomplish the goal of mitigating the congestion caused to other flows of bursty sending.

I thought my existing text does that, but people are still confused, so better text would be greatly appreciated.

mikkelfj: The problem is that it is not clear when the next burst can be sent.

draft-ietf-quic-recovery.md
junhochoi: I think BBR is considered here for pacing limited case? But there is no description what is pacing limited.

janaiyengar: @junhochoi : The point here is that if you're unable to utilize the full cwnd because you are limited by a pacer, then it doesn't count as app limited.

PR #2677, opened 2019-05-08
Trigger unhappy

LPardue:
IMO the trigger is the peer's behaviour, and the reaction by the recipient is to send a STOP_SENDING in response. I.e. we don't state the reaction is to force the remote peer to send a STOP_SENDING

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -449,7 +449,7 @@ HTTP_INCOMPLETE_REQUEST.
 A server can send a complete response prior to the client sending an entire
 request if the response does not depend on any portion of the request that has
 not been sent and received. When this is true, a server MAY request that the
-client abort transmission of a request without error by triggering a QUIC
+client abort transmission of a request without error by send a QUIC
 STOP_SENDING frame with error code HTTP_EARLY_RESPONSE, sending a complete
 response, and cleanly closing its stream. Clients MUST NOT discard complete
 responses as a result of having their request terminated abruptly, though
@@ -908,7 +908,7 @@ transport parameter `initial_max_stream_data_uni`.
 
 If the stream header indicates a stream type which is not supported by the
 recipient, the remainder of the stream cannot be consumed as the semantics are
-unknown. Recipients of unknown stream types MAY trigger a QUIC STOP_SENDING
+unknown. Recipients of unknown stream types MAY send a QUIC STOP_SENDING
 frame with an error code of HTTP_UNKNOWN_STREAM_TYPE, but MUST NOT consider such
 streams to be a connection error of any kind.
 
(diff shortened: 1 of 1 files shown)


MikeBishop:
The reason for this language is because an HTTP implementation cannot send a QUIC frame -- it can merely ask/tell the transport implementation to do so.  Feel free to suggest alternative verbs, but we have "trigger" because "send" already got nixed.

LPardue:
> The reason for this language is because an HTTP implementation cannot send a QUIC frame

This rule is not applied universally in the document, probably because I didn't know about it :)

Examples:
* > If the push stream has been opened by the server, the server SHOULD send a QUIC RESET_STREAM
* > When a client sends a STOP_SENDING with HTTP_REQUEST_CANCELLED, a
   server MAY send the error code HTTP_REQUEST_REJECTED in the
   corresponding RESET_STREAM if no processing was performed.
* > an HTTP/3 client SHOULD send non-zero values for the QUIC transport
   parameters "initial_max_stream_data_bidi_local".  An HTTP/3 server
   SHOULD send non-zero values for the QUIC transport parameters
   "initial_max_stream_data_bidi_remote" and "initial_max_bidi_streams".

Suggestion welcome on how to address this consistently.

MikeBishop:
Well, of those three, I was the last person to touch two of them and @martinthomson the other.  So not just you -- I can be sloppy all on my own.  Thanks for finding them!

The language used elsewhere for sending a STOP_SENDING or RESET_STREAM is "aborting the stream," an action that the HTTP implementation would take on the transport API.  That covers two of them; for the settings, we could say that the HTTP implementation "provides" or "sets" values for them.

janaiyengar:
"cause a XXX frame to be sent"?

LPardue:
It feels to me like we might benefit from hoisting the definition of aborting an HTTP/3 stream (including whether that results in a emitting a STOP_SENDING or RESET_STREAM frame) to one place. This makes it explicitly clear and referable throughout the document.

ianswett:
Trigger is used in exactly two spots, where are both fixed in this PR. Unless we're sure what to do going forward, how about landing this and deciding how to better name this later?

Also, I'm not a big fan of trigger as a term.  Cause and send are both preferable options in my opinion.

mikkelfj:
trigger -> instigate ?

ianswett:
If we want to be more correct going forward, I think HTTP should avoid talking about transport frames entirely.

Better would be "Close the connection with application error code FOO", etc.  Because we don't only mean "Send a connection close frame", we actually mean "Close the connection and as part of that, send a connection close frame."

MikeBishop:
In principle, I like that:  Note the use of "abort the stream" instead of RESET_STREAM in most cases.  However, somewhere there has to be a mapping of those terms into what it means for the transport, just so everyone's clear.

LPardue:
Closing this PR in favour of #2857

PR #2679, opened 2019-05-09
change the ack delay exponent to an ack delay multiplier

### LINKED ISSUE #2670: Remove ack_delay_exponent TP
I asked the question on Slack and got quite a bit of feedback so I figured I'd open an issue.

Right now, it seems all but one implementation (quicly) is using the default ack_delay_exponent value of 8. Quicly is using an even higher value of 10 because they don't care about more than millisecond timer resolution.

This leads me to believe that having this be a configurable value (in the QUIC TP) is an over design, and we should simplify this. Pick a value (8 or 10, I personally don't care which) and just leave it at that.

The only argument I've heard for keeping it configurable is that intra-DC environments might want to have a smaller exponent to have more fine grain ACK delays. I'm not sure that's a great reason, personally. At that small of time granularity, most OS timers aren't going to be very accurate. So, even if you can calculate an extremely accurrate RTT, it's not going to help your loss recovery timers any.

marten-seemann:
Since several people expressed interested in changing the `ack_delay_exponent` to a multiplier in #2670, I decided to write down a PR with that change.
We will probably have a discussion about the default value of this TP. For now, I chose 10 microseconds, which is close to the value we had so far.

Closes #2670.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1406,7 +1406,7 @@ not store a transport parameter it cannot process.
 
 A client MUST NOT use remembered values for the following parameters:
 original_connection_id, preferred_address, stateless_reset_token,
-ack_delay_exponent and active_connection_id_limit. The client MUST use the
+ack_delay_multiplier and active_connection_id_limit. The client MUST use the
 server's new values in the handshake instead, and absent new values from the
 server, the default value.
 
@@ -4068,7 +4068,7 @@ language from Section 3 of {{!TLS13=RFC8446}}.
       initial_max_stream_data_uni(7),
       initial_max_streams_bidi(8),
       initial_max_streams_uni(9),
-      ack_delay_exponent(10),
+      ack_delay_multiplier(10),
       max_ack_delay(11),
       disable_migration(12),
       preferred_address(13),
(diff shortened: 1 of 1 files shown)


martinthomson:
Let's pretend that this addresses/closes #2670.

draft-ietf-quic-transport.md
```diff
@@ -4186,12 +4186,13 @@ initial_max_streams_uni (0x0009):
...
 
-: The ACK delay exponent is an integer value indicating an
-  exponent used to decode the ACK Delay field in the ACK frame ({{frame-ack}}).
-  If this value is absent, a default value of 3 is assumed (indicating a
-  multiplier of 8). Values above 20 are invalid.
+: The ACK delay multiplier is an integer value indicating a multiplier used to
+  decode the ACK Delay field in the ACK frame ({{frame-ack}}). If this value is
+  absent, a default value of 10 (corresponding to a resolution of 10
+  microseconds)is assumed. Values above 1,000,000 (corresponding to a resolution
+  of 1 second) are invalid.
```
martinthomson: It might be better to leave the multiplier at 1 rather than pick a value.

Also, I assume that we want to prohibit 0.

mikkelfj:
Is this necessary? Imposing a multiplier on peer might be OK, but if we ever get to a situation where division is forced, an exponent is vastly better - so this concern is about precefence.

And I don’t really see what a multiplier gains over an exponent.

marten-seemann:
@mikkelfj Where is a division necessary? The PR only requires multiplication.

mikkelfj:
As I say, this sets a precedence if we need to do more scales in the future. In the current form the sender decides what division is necessary. So I think there should be strong win for multiplication to prefer that.

That said, it is already messy with base 10 units, but constant mult/div by 1000 is manageable.

Another complication is overflow detection. With an exponent you can easily construct a mask. With multiplication you must either add another max var, or have another multiplication to compute it on the fly. Otherwise this will probably be an attack vector on naive implementations.

martinthomson:
I don't think we need to worry about optimization of these corner cases.

martinthomson:
The working group decided not to make a change here.

PR #2682, opened 2019-05-09
Allow AES_128_CCM_8 in QUIC

### LINKED ISSUE #2581: Bring back AEAD_AES_128_CCM_8 now that we pad the plaintext
As discussed in #2019, we did not originally have a good way to make AEAD_AES_128_CCM_8 work in QUIC, because it needs a 16 byte sample for header protection but only produces an 8 byte auth tag. However, since then we have merged #2030, and it's therefore easy to pad short packets by 8 extra bytes when using CCM_8. There might be QUIC use-cases that could benefit from the shorter tags, and it would be nice to have QUIC support all the original TLS 1.3 cipher suites.

DavidSchinazi:
This resolves #2581. Now that we pad the payload to ensure sufficient header protection sample size, QUIC can use all of the cipher suites defined in the TLS 1.3 spec. Note that this PR also removes AEAD_AES_256_CCM because that is not defined in TLS 1.3 and I believe it was mentioned here by mistake.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -824,13 +824,7 @@ Connection ID in the client's first Initial packet (see {{initial-secrets}}).
 This provides protection against off-path attackers and robustness against QUIC
 version unaware middleboxes, but not against on-path attackers.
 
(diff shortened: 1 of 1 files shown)


martinthomson:
This should have been closed with #2743.

PR #2683, opened 2019-05-09
Pull some H2 content into the introduction

### LINKED ISSUE #1601: Discuss HTTP/2 more in the introduction
Per side-conversation feedback.  Maybe new text, maybe move some from HTTP/2 transition section.

MikeBishop:
Some edits to the beginning of the draft based on an RFC7540 read-through.

Closes #1601.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -196,6 +196,10 @@ This document uses the variable-length integer encoding from
 
 The following terms are used:
 
+abort:
+: An abrupt termination of a connection or stream, possibly due to an error
+  condition.
+
 client:
 : The endpoint that initiates an HTTP/3 connection.  Clients send HTTP requests
   and receive HTTP responses.
@@ -241,11 +245,11 @@ stream:
 stream error:
 : An error on the individual HTTP/3 stream.
 
+The term "payload body" is defined in Section 3.3 of {{!RFC7230}}.
+
 Finally, the terms "gateway", "intermediary", "proxy", and "tunnel" are defined
 in Section 2.3 of {{!RFC7230}}.  Intermediaries act as both client and server at
 different times.
-
-The term "payload body" is defined in Section 3.3 of {{!RFC7230}}.
 
 
 # Connection Setup and Management {#connection-setup}
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -88,22 +88,39 @@ code and issues list for this draft can be found at
 
 HTTP semantics are used for a broad range of services on the Internet. These
 semantics have commonly been used with two different TCP mappings, HTTP/1.1 and
-HTTP/2.  HTTP/2 introduced a framing and multiplexing layer to improve latency
-without modifying the transport layer.  However, TCP's lack of visibility into
-parallel requests in both mappings limited the possible performance gains.
+HTTP/2.  HTTP/3 supports the same semantics over a new transport protocol, QUIC.
+
+## Prior versions of HTTP
+
+HTTP/1.1 was a TCP mapping which used whitespace-delimited text fields to convey
+HTTP messages.  While these exchanges were human-readable, the whitespace led to
+parsing difficulties and workarounds to be tolerant of variant behavior.
+Because each connection could be used only for a single HTTP request and
+response at a time, multiple parallel TCP connections were used, reducing the
+ability of the congestion controller to accurately manage traffic between
+endpoints.
+
+HTTP/2 introduced a binary framing and multiplexing layer to improve latency
+without modifying the transport layer.  However, because the parallel nature
+of HTTP/2's multiplexing was not visible to TCP's loss recovery mechanisms,
+a lost or reordered packet caused all active transactions to experience a stall
+regardless of whether that transaction was impacted by the lost packet.
+
+## Delegation to QUIC
 
 The QUIC transport protocol incorporates stream multiplexing and per-stream flow
 control, similar to that provided by the HTTP/2 framing layer. By providing
 reliability at the stream level and congestion control across the entire
 connection, it has the capability to improve the performance of HTTP compared to
 a TCP mapping.  QUIC also incorporates TLS 1.3 at the transport layer, offering
-comparable security to running TLS over TCP, but with improved connection setup
-latency (unless TCP Fast Open {{?RFC7413}}} is used).
+comparable security to running TLS over TCP, with the improved connection setup
+latency of TCP Fast Open {{?RFC7413}}}.
 
 This document defines a mapping of HTTP semantics over the QUIC transport
-protocol, drawing heavily on the design of HTTP/2. This document identifies
-HTTP/2 features that are subsumed by QUIC, and describes how the other features
-can be implemented atop QUIC.
+protocol, drawing heavily on the design of HTTP/2.  While delegating stream
+management and flow control issues to QUIC, a similar binary framing is used on
+each stream. Some HTTP/2 features are subsumed by QUIC, while other features are
+implemented atop QUIC.
 
 QUIC is described in {{QUIC-TRANSPORT}}.  For a full description of HTTP/2, see
 {{!RFC7540}}.
@@ -922,10 +939,10 @@ implementation chooses.
 
 ## HTTP Message Exchanges {#request-response}
 
-A client sends an HTTP request on a client-initiated bidirectional QUIC
-stream. A client MUST send only a single request on a given stream.
-A server sends one or more HTTP responses on the same stream as the request,
-as detailed below.
+A client sends an HTTP request on a client-initiated bidirectional QUIC stream.
+A client MUST send only a single request on a given stream. A server sends zero
+or more non-final HTTP responses on the same stream as the request, followed by
+a single final HTTP response, as detailed below.
 
 An HTTP message (request or response) consists of:
 
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -93,12 +93,12 @@ HTTP/2.  HTTP/3 supports the same semantics over a new transport protocol, QUIC.
 ## Prior versions of HTTP
 
 HTTP/1.1 is a TCP mapping which uses whitespace-delimited text fields to convey
-HTTP messages.  While these exchanges are human-readable, the whitespace leads
-to parsing difficulties and workarounds to be tolerant of variant behavior.
-Because each connection can transfer only a single HTTP request or response at a
-time in each direction, multiple parallel TCP connections are used, reducing the
-ability of the congestion controller to accurately manage traffic between
-endpoints.
+HTTP messages.  While these exchanges are human-readable, using whitespace for
+message formatting leads to parsing difficulties and workarounds to be tolerant
+of variant behavior. Because each connection can transfer only a single HTTP
+request or response at a time in each direction, multiple parallel TCP
+connections are often used, reducing the ability of the congestion controller to
+accurately manage traffic between endpoints.
 
 HTTP/2 introduced a binary framing and multiplexing layer to improve latency
 without modifying the transport layer.  However, because the parallel nature of
@@ -130,10 +130,11 @@ QUIC is described in {{QUIC-TRANSPORT}}.  For a full description of HTTP/2, see
 HTTP/3 provides a transport for HTTP semantics using the QUIC transport protocol
 and an internal framing layer similar to HTTP/2.
 
-An HTTP/3 endpoint can be discovered using HTTP Alternative Services; this
-process is described in greater detail in {{discovery}}.  Once a client knows
-that an HTTP/3 server exists at a certain endpoint, it opens a QUIC connection.
-QUIC provides protocol negotiation, stream-based multiplexing, and flow control.
+Once a client knows that an HTTP/3 server exists at a certain endpoint, it opens
+a QUIC connection. QUIC provides protocol negotiation, stream-based
+multiplexing, and flow control. An HTTP/3 endpoint can be discovered using HTTP
+Alternative Services; this process is described in greater detail in
+{{discovery}}.
 
 Within each stream, the basic unit of HTTP/3 communication is a frame
 ({{frames}}).  Each frame type serves a different purpose.  For example, HEADERS
@@ -1876,11 +1877,12 @@ section describes the approach taken to design HTTP/3, points out important
 differences from HTTP/2, and describes how to map HTTP/2 extensions into HTTP/3.
 
 HTTP/3 begins from the premise that similarity to HTTP/2 is preferable, but not
-a hard requirement.  HTTP/3 departs from HTTP/2 where QUIC's behavior
-differences from TCP (lack of ordering, support for streams) are either
-beneficial or need to be accommodated.  HTTP/3 attempts to avoid gratuitous
-changes which make it difficult or impossible to build extensions with the same
-semantics applicable to both protocols at once.
+a hard requirement.  HTTP/3 departs from HTTP/2 where QUIC differs from TCP,
+either to take advantage of QUIC features (like streams) or to accommodate
+important shortcomings (such as a lack of total ordering). These differences
+make HTTP/3 similar to HTTP/2 in key aspects, such as the relationship of
+requests and responses to streams. However, the details of the HTTP/3 design are
+substantially different than HTTP/2.
 
 These departures are noted in this section.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -88,28 +88,100 @@ code and issues list for this draft can be found at
...
 HTTP semantics are used for a broad range of services on the Internet. These
 semantics have commonly been used with two different TCP mappings, HTTP/1.1 and
-HTTP/2.  HTTP/2 introduced a framing and multiplexing layer to improve latency
-without modifying the transport layer.  However, TCP's lack of visibility into
-parallel requests in both mappings limited the possible performance gains.
+HTTP/2.  HTTP/3 supports the same semantics over a new transport protocol, QUIC.
+
+## Prior versions of HTTP
+
+HTTP/1.1 was a TCP mapping which used whitespace-delimited text fields to convey
```
mikkelfj: Past tense may not be appropriate as it is still in active use in a few places.

draft-ietf-quic-http.md
```diff
@@ -88,28 +88,100 @@ code and issues list for this draft can be found at
...
+HTTP messages.  While these exchanges were human-readable, the whitespace led to
+parsing difficulties and workarounds to be tolerant of variant behavior.
+Because each connection could be used only for a single HTTP request and
+response at a time, multiple parallel TCP connections were used, reducing the
```
mikkelfj: That is not true. Pipelining is a thing.

draft-ietf-quic-http.md
```diff
@@ -88,28 +88,100 @@ code and issues list for this draft can be found at
...
-
-## Notational Conventions
+{{!HTTP2=RFC7540}}.
+
+# HTTP/3 Protocol Overview
+
+HTTP/3 provides a transport for HTTP semantics using the QUIC transport protocol
+and an internal framing layer similar to HTTP/2.
+
+An HTTP/3 endpoint is discovered using HTTP Alternative Services.  Once a client
```
mikkelfj: Can be

draft-ietf-quic-http.md
```diff
@@ -88,28 +88,100 @@ code and issues list for this draft can be found at
...
+Within each stream, the basic unit of HTTP/3 communication is a frame
+({{frames}}).  Each frame type serves a different purpose.  For example, HEADERS
+and DATA frames form the frames form the basis of HTTP requests and responses
+({{request-response}}).  Other frame types like SETTINGS, PRIORITY, and GOAWAY
+are used to managed the overall connection.
+
+Multiplexing of requests is performed using the QUIC stream abstraction, with
+each request and response consuming a single QUIC stream.  Streams are largely
+independent of each other, so one stream that is blocked or suffers packet loss
+does not prevent progress on other streams.
```
mikkelfj: But progress of other streams can when flow credits are exhausted.

draft-ietf-quic-http.md
```diff
@@ -88,28 +88,100 @@ code and issues list for this draft can be found at
...
+  semantics are expressed using frames.
+- Connection Closure ({{connection-closure}}) describes how connections are
+  terminated, either gracefully or abruptly.
+- Extensions to HTTP/3 ({{extensions}}) describes how new capabilities can be
+  added in future documents.
+- Error Handling ({{errors}}) describes how error conditions are handled and
+  expressed, either on a particular stream or for the connection as a whole.
+
+Readers familiar with HTTP/2 will find a more detailed comparison with that
+protocol in {{h2-considerations}}.
```
mikkelfj: With that protocol? HTTP/1.1?

draft-ietf-quic-http.md
```diff
@@ -122,13 +194,65 @@ Field definitions are given in Augmented Backus-Naur Form (ABNF), as defined in
...
+connection error:
+: An error that affects the entire HTTP/3 connection.
+
+endpoint:
+: Either the client or server of the connection.
+
+frame:
+: The smallest unit of communication within an HTTP/3 connection, consisting of
+  a header and a variable-length sequence of octets structured according to the
+  frame type.
```
mikkelfj: It may be worth noting that frames can span mutiple packets unlike QUIC frames.

draft-ietf-quic-http.md
```diff
@@ -1742,7 +1866,7 @@ assigned by IANA.
 
 --- back
 
-# Considerations for Transitioning from HTTP/2
+# Considerations for Transitioning from HTTP/2 {#h2-considerations}
 
 HTTP/3 is strongly informed by HTTP/2, and bears many similarities.  This
 section describes the approach taken to design HTTP/3, points out important
```
ianswett: Below "HTTP/3 departs from HTTP/2 primarily where necessary to
accommodate the differences in behavior between QUIC and TCP (lack of ordering,	accommodate the 
support for streams)."

I'd tweak this to be more along the lines of "We changed it when we thought a new design was better, with particular focus on taking advantage of QUIC's lack of head of line blocking" to reflect the state we're in today.  Don't take my suggestion verbatim, but I think that's the concept?

draft-ietf-quic-http.md
```diff
@@ -88,28 +88,100 @@ code and issues list for this draft can be found at
...
-latency (unless TCP Fast Open {{?RFC7413}}} is used).
+comparable security to running TLS over TCP, with the improved connection setup
+latency of TCP Fast Open {{?RFC7413}}}.
 
 This document defines a mapping of HTTP semantics over the QUIC transport
-protocol, drawing heavily on the design of HTTP/2. This document identifies
-HTTP/2 features that are subsumed by QUIC, and describes how the other features
-can be implemented atop QUIC.
+protocol, drawing heavily on the design of HTTP/2.  While delegating stream
+management and flow control issues to QUIC, a similar binary framing is used on
```
LPardue: what does the term stream management mean in this case? In some perspectives, QUIC doesn't manage much other than ensuring the stream arrives and is presented in-order at the other end (as explained elsewhere in this doc).

draft-ietf-quic-http.md
```diff
@@ -88,28 +88,100 @@ code and issues list for this draft can be found at
...
+and an internal framing layer similar to HTTP/2.
+
+An HTTP/3 endpoint is discovered using HTTP Alternative Services.  Once a client
+knows that an HTTP/3 server exists at a certain endpoint, it opens a QUIC
+connection. QUIC provides protocol negotiation, stream-based multiplexing, and
+flow control.
+
+Within each stream, the basic unit of HTTP/3 communication is a frame
+({{frames}}).  Each frame type serves a different purpose.  For example, HEADERS
+and DATA frames form the frames form the basis of HTTP requests and responses
```
```suggestion
and DATA frames form the basis of HTTP requests and responses
```

draft-ietf-quic-http.md
```diff
@@ -122,13 +194,65 @@ Field definitions are given in Augmented Backus-Naur Form (ABNF), as defined in
...
+  transport protocol.
+
+connection error:
+: An error that affects the entire HTTP/3 connection.
+
+endpoint:
+: Either the client or server of the connection.
+
+frame:
+: The smallest unit of communication within an HTTP/3 connection, consisting of
```
LPardue: I know this comes from HTTP/2 but one could argue that frames are not the smallest unit in HTTP/3.

draft-ietf-quic-http.md
```diff
@@ -122,13 +194,65 @@ Field definitions are given in Augmented Backus-Naur Form (ABNF), as defined in
...
+: An endpoint.  When discussing a particular endpoint, "peer" refers to the
+  endpoint that is remote to the primary subject of discussion.
+
+receiver:
+: An endpoint that is receiving frames.
+
+sender:
+: An endpoint that is transmitting frames.
+
+server:
```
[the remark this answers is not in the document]
LPardue: disagree, the order is lexical and mirrors the ordering of terms in RFC 7540

draft-ietf-quic-http.md
```diff
@@ -922,10 +1046,10 @@ implementation chooses.
...
 ## HTTP Message Exchanges {#request-response}
 
-A client sends an HTTP request on a client-initiated bidirectional QUIC
-stream. A client MUST send only a single request on a given stream.
-A server sends one or more HTTP responses on the same stream as the request,
-as detailed below.
+A client sends an HTTP request on a client-initiated bidirectional QUIC stream.
+A client MUST send only a single request on a given stream. A server sends zero
+or more non-final HTTP responses on the same stream as the request, followed by
+a single final HTTP response, as detailed below.
```
LPardue: :+1:

draft-ietf-quic-http.md
LPardue: Based on the evidence, the design changes seem to be motivated because we lost something we had, not because we are taking advantage of some new ability.

MikeBishop: We've gotten a little stretchy in places, but we've been explicitly trying to stay out of the "improve on HTTP/2 in ways unrelated to QUIC" game.  (Some fixes to the more egregious gaps in prioritization are the primary exception.)  I'm trying to reflect that aspiration, though you're correct that we might want to reflect the stretchiness a bit as well.

draft-ietf-quic-http.md
mikkelfj: I'm doing that, processing multiple requests concurrently on different backends and streaming the responses back in order.

draft-ietf-quic-http.md
janaiyengar: @mikkelfj : I suspect you're missing that comparison "of HTTP/3" is assumed (which is reasonable, IMO)

draft-ietf-quic-http.md
janaiyengar: How about "HTTP/3 departs from HTTP/2 where QUIC's behavior differences from TCP are either beneficial or need to be accommodated (absence of linear ordering, support for framing)."

draft-ietf-quic-http.md
mikkelfj: Well, understand the text, but it reads as if was written immediately after the first paragraph and then a lot of text was thrown in, end then "that protocol" stands dangling becuse it might mean either HTTP/1.1 or QUIC until you realize it cannot mean QUIC since QUIC didn't exist when HTTP/2 was written. So I merely suggest typing out HTTP/1.1 explicitly.

Also, my feedback have come across a bit grumpy, that was not the intention. I really do appreciate the extra clarifying text on this topic.

mikkelfj: Actually I am missing that. I was assuming HTTP/1.1 but now it is HTTP/3. So yes, I think a clarification is needed.

draft-ietf-quic-http.md
MikeBishop: Doesn't exhausted flow control credits fall under "is blocked"?

draft-ietf-quic-http.md
MikeBishop: That's a good suggestion, but I'm not sure I want to overload what is essentially a glossary with that.  I'll add that to the description of frames in general.

draft-ietf-quic-http.md
```diff
@@ -88,28 +88,101 @@ code and issues list for this draft can be found at
...
 semantics have commonly been used with two different TCP mappings, HTTP/1.1 and
-HTTP/2.  HTTP/2 introduced a framing and multiplexing layer to improve latency
-without modifying the transport layer.  However, TCP's lack of visibility into
-parallel requests in both mappings limited the possible performance gains.
+HTTP/2.  HTTP/3 supports the same semantics over a new transport protocol, QUIC.
+
+## Prior versions of HTTP
+
+HTTP/1.1 is a TCP mapping which uses whitespace-delimited text fields to convey
+HTTP messages.  While these exchanges are human-readable, the whitespace leads
```
martinthomson: It isn't just the whitespace, but the use of whitespace for delineation makes some aspects of parsing and processing difficult.  You should be careful not to talk about shortcomings that this version doesn't address.

draft-ietf-quic-http.md
```diff
@@ -88,28 +88,101 @@ code and issues list for this draft can be found at
...
+to parsing difficulties and workarounds to be tolerant of variant behavior.
+Because each connection can transfer only a single HTTP request or response at a
+time in each direction, multiple parallel TCP connections are used, reducing the
```
martinthomson: "are used" or "are often used"

draft-ietf-quic-http.md
```diff
@@ -88,28 +88,101 @@ code and issues list for this draft can be found at
...
+ability of the congestion controller to accurately manage traffic between
+endpoints.
```
martinthomson: Only because people haven't worked out how to run the same congestion controller for multiple connections.  I mean, there is no reason you can't do that... in theory.

draft-ietf-quic-http.md
```diff
@@ -88,28 +88,101 @@ code and issues list for this draft can be found at
...
+# HTTP/3 Protocol Overview
 
-## Notational Conventions
+HTTP/3 provides a transport for HTTP semantics using the QUIC transport protocol
+and an internal framing layer similar to HTTP/2.
+
+An HTTP/3 endpoint can be discovered using HTTP Alternative Services; this
+process is described in greater detail in {{discovery}}.  Once a client knows
+that an HTTP/3 server exists at a certain endpoint, it opens a QUIC connection.
+QUIC provides protocol negotiation, stream-based multiplexing, and flow control.
```
martinthomson: I'd invert this paragraph.  The key points are in the second sentence.  The alternative services thing is really secondary here.

draft-ietf-quic-http.md
```diff
@@ -122,13 +195,65 @@ Field definitions are given in Augmented Backus-Naur Form (ABNF), as defined in
...
+
+connection:
+: A transport-layer connection between two endpoints, using QUIC as the
+  transport protocol.
+
+connection error:
+: An error that affects the entire HTTP/3 connection.
 
+endpoint:
+: Either the client or server of the connection.
```
martinthomson: Love the circular definition here.

draft-ietf-quic-http.md
```diff
@@ -1742,18 +1869,18 @@ assigned by IANA.
...
 differences from HTTP/2, and describes how to map HTTP/2 extensions into HTTP/3.
 
 HTTP/3 begins from the premise that similarity to HTTP/2 is preferable, but not
-a hard requirement.  HTTP/3 departs from HTTP/2 primarily where necessary to
-accommodate the differences in behavior between QUIC and TCP (lack of ordering,
-support for streams).  We intend to avoid gratuitous changes which make it
-difficult or impossible to build extensions with the same semantics applicable
-to both protocols at once.
+a hard requirement.  HTTP/3 departs from HTTP/2 where QUIC's behavior
+differences from TCP (lack of ordering, support for streams) are either
```
```suggestion
where QUIC differs from TCP, either to take advantage of QUIC features (like streams) or to accommodate important shortcomings (such as a lack of total ordering).
```

draft-ietf-quic-http.md
```diff
@@ -1742,18 +1869,18 @@ assigned by IANA.
...
+beneficial or need to be accommodated.  HTTP/3 attempts to avoid gratuitous
+changes which make it difficult or impossible to build extensions with the same
+semantics applicable to both protocols at once.
```
martinthomson: I think that we can lose this sentence now.  It might be time to face up to reality:

> These differences make HTTP/3 similar to HTTP/2 in key aspects, such as the relationship of requests and responses to streams.  However, the details of the HTTP/3 design are substantially different than HTTP/2.

draft-ietf-quic-http.md
mikkelfj: Agree with @martinthomson 
Technically, whitespace is very easy to parse, and can be more efficient than variable length encoded frames. The real problem is the vague definition of what constitutes whitespace, when and where to expect it, and the different kinds of whitespace in existence. JSON has no such issues with whitespace.

draft-ietf-quic-http.md
MikeBishop: Everyone loves [their own baby](https://tools.ietf.org/html/rfc7540#section-2.2). 😉

draft-ietf-quic-http.md
MikeBishop: I heard that proposal again in Prague, as a matter of fact.  I wish them all success.

draft-ietf-quic-http.md
```diff
@@ -1774,7 +1898,7 @@ removed. Because stream termination is handled by QUIC, an END_STREAM flag is
 not required.  This permits the removal of the Flags field from the generic
 frame layout.
 
-Frame payloads are largely drawn from {{!RFC7540}}. However, QUIC includes many
+Frame payloads are largely drawn from {{!HTTP2}}. However, QUIC includes many
```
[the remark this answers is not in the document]
MikeBishop: I'd prefer to handle that separately, since that's distinct from what I'm doing here and potentially sizable.  I've opened #2698 with a first pass at that change.

Issue #2684, opened 2019-05-10
Mixture of transport and congestion control functions


gorryfair:
Section 9.4 contains a mixture of transport (handling migration, setup, etc) and congestion control functions (choice of path parameters - RTO, window, etc). 

The abstract suggests that these are a part of QUIC Recovery ID, which seems correct to me. So, I would like to discuss if these are intentionally a part of the transport spec, rather than the Loss Detection & Congestion Control Spec.? 

Specifically, para 1,2 : "... SHOULD NOT contribute to congestion control or RTT estimation for the new path." and "...  SHOULD immediately reset ...."  - Where the ID could note whatever transport implementors need to be aware of, but point normatively to the QUIC Recovery ID for RFC2119 language.

Para 3: "An endpoint MUST NOT return to the send rate..." seems a reasonable thing to note, but also, I think, should point to QUIC Recovery ID.

larseggert:
Discussed in London: @gorryfair to work with @janaiyengar to propose an editorial PR

ianswett:
The target is to fix this issue with the same PR as #2909

martinthomson:
Discussed today with editors and we believe that these issues have been rectified in the numerous edits that section has received since this issue was opened.

Issue #2367, opened 2019-01-23
Text banning post-handshake client auth is unclear


DavidSchinazi:
draft-tls section 4.4 (Peer Authentication) states:
> A server MUST NOT use post-handshake client authentication (see Section 4.6.2 of [TLS13]).

This sounds like there are security implications, but what they are is unclear. As a reader of the document, this makes me doubt my understanding of the protocol.

If we don't have a good justification for banning this, we shouldn't be banning it.

(This is me reopening #2294 since that was closed for being a question.)

martinthomson:
Is the text unclear, or are you asking to re-enable post-handshake authentication?

mikkelfj:
At least I am curious about the reasoning behind this, not that I am against it.

martinthomson:
@mikkelfj, The best way to understand the rationale is to try to work through how a particular protocol might use this feature.  Read [HTTP/2 secondary certificate authentication](https://httpwg.org/http-extensions/draft-ietf-httpbis-http2-secondary-certs.html), look at the design and discussion and see the sorts of problems it tries to address.  In that case, some of the problems are with server-authentication, which you can safely ignore, but the general class of problems related to concurrency and how that relates to the status of authentication produce a fairly complex design.

Now I might concede that it is possible to build a protocol that uses this TLS feature well, but I would still hold that [exported authenticators](https://tlswg.github.io/tls-exported-authenticator/draft-ietf-tls-exported-authenticator.html) are a far superior design.

mikkelfj:
@martinthomson Thanks for the explantion and pointers. On exported authenticators I guess there is some overlap with OAuth2, Fido, and temporary web service tokens once you get past the handshake.

DavidSchinazi:
@martinthomson if we have a good reason (which it sounds like we do), then this issue tracks adding that reason to the draft.

DavidSchinazi:
Now that @davidben has clearly written this up, perhaps we can resolve this issue by adding a reference to https://tools.ietf.org/html/draft-ietf-httpbis-http2-tls13 in draft-ietf-quic-tls. If that sounds good to the editors I can write a PR.

mikkelfj:
I'd prefer more direct text. You get references to HTTP/2 doc and that in turn references TLS 1.2 with compression rules etc. that needs to be filtered out.

janaiyengar:
@DavidSchinazi : that sounds like a good way forward.

PR #2687, opened 2019-05-10
New paths may seed initial_rtt

### LINKED ISSUE #2644: RTT measurement using PATH_* frames
Should the recovery draft explain how to gather RTT samples during connection migration using PATH_CHALLENGE and PATH_RESPONSE?

This is explained in:
- https://tools.ietf.org/html/draft-paulo-quic-migration-00#section-3.3

ianswett:
Using PATH_CHALLENGE/PATH_RESPONSE

I'm fairly happy with the text, but not that happy with the location.  Especially after the recently filed issues(ie: #2684) on resetting the congestion controller on path changes, I think we may need a section in recovery on that.

Fixes #2644

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -522,8 +522,8 @@ resulting in a 1 second initial handshake timeout as recommended in
 {{?RFC6298}}.
 
 A connection MAY use the delay between sending a PATH_CHALLENGE and receiving
-a PATH_RESPONSE to seed initial_rtt for a new path, but MUST not be considered
-an RTT sample.
+a PATH_RESPONSE to seed initial_rtt for a new path, but the delay MUST not be
+considered an RTT sample.
 
 When a crypto packet is sent, the sender MUST set a timer for twice the smoothed
 RTT.  This timer MUST be updated when a new crypto packet is sent and when
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -522,8 +522,8 @@ resulting in a 1 second initial handshake timeout as recommended in
 {{?RFC6298}}.
 
 A connection MAY use the delay between sending a PATH_CHALLENGE and receiving
-a PATH_RESPONSE to seed initial_rtt for a new path, but the delay MUST NOT be
-considered an RTT sample.
+a PATH_RESPONSE to seed initial_rtt for a new path, but the delay SHOULD NOT
+be considered an RTT sample.
 
 When a crypto packet is sent, the sender MUST set a timer for twice the smoothed
 RTT.  This timer MUST be updated when a new crypto packet is sent and when
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -521,6 +521,10 @@ is available, or if the network changes, the initial RTT SHOULD be set to 500ms,
 resulting in a 1 second initial handshake timeout as recommended in
 {{?RFC6298}}.
 
+A new path on an existing connection MAY use the delay between sending a
+PATH_CHALLENGE and receiving a PATH_RESPONSE to seed initial_rtt for a new path,
+but MUST not be considered an RTT sample.
+
 When a crypto packet is sent, the sender MUST set a timer for twice the smoothed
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -521,6 +521,10 @@ is available, or if the network changes, the initial RTT SHOULD be set to 500ms,
 resulting in a 1 second initial handshake timeout as recommended in
 {{?RFC6298}}.
 
+A new path on an existing connection MAY use the delay between sending a
```
janaiyengar: can you fix this sentence? the path does not use the delay

ianswett: Better?

draft-ietf-quic-recovery.md
```diff
@@ -521,6 +521,10 @@ is available, or if the network changes, the initial RTT SHOULD be set to 500ms,
 resulting in a 1 second initial handshake timeout as recommended in
 {{?RFC6298}}.
 
+A connection MAY use the delay between sending a PATH_CHALLENGE and receiving
+a PATH_RESPONSE to seed initial_rtt for a new path, but MUST not be considered
```
dtikhonov: Consider adding "the delay" -- "..., but _the delay_ MUST not be considered..."

In the current form, the sentence suggests that the connection must not be considered an RTT sample.

ianswett: Thanks, done.

martinthomson:
1. Do we really want to keep the requirement to respond with a PATH_RESPONSE *immediately*?

2. This doesn't talk about paths.  Now, we might infer from this that the RTT to a peer isn't likely to dramatically change when paths change, but we'd want to be saying that directly.

3. How does this sit with the change in #2637?

mikkelfj:
On 2.: if you roam from 5G to 3G you could see significant latency spikes. Phasing out 3G networks and keeping 2G could move traffic to 2G EDGE in areas of thin coverage. Traffic could also move to narrow band IoT (if it is possible to run QUIC on these at all) with MUCH higher latency. Endpoints could choose stay on narrow band most of the time to save battery and perhaps online costs, but use 5G when a device wakes up for real (I could see a use case for that). Though of course might be served with separate connections for such extreme transitions.

ianswett:
1. I think this is OK as long as we limit how many path responses we send immediately?  Realistically, it's not that different from a generate_204, but it's something to be aware of.
2. I think we need to add a whole section on paths and connection migration in recovery. See #2684   I can do that in this PR if people want? 
3. #2367 decreases the value of this, since you may only see one half of the path.  It might still be better than nothing?  Initial RTT doesn't need to be correct, but getting it within a factor of 2 is helpful.

DavidSchinazi:
@ianswett I don't think you meant 2367

ianswett:
Thanks David, fixed.

janaiyengar:
I'm split on this. While I understand the urge to optimize here, I'm not sure that documenting it here is necessary. 

Saying that an endpoint MAY use this signal gives permission, but does so without noting the potential issues. At a minimum, you would need to talk about the problems that are lurking ... especially in light of #2637, as I've noted in [my comment on #2644](https://github.com/quicwg/base-drafts/issues/2644#issuecomment-490721850).

Which leads me to wondering if it's really worth documenting this here without more experimentation and data.

ianswett:
I'm a fan of forbidding it's use as an actual RTT measurement, which would be one reason to say something?

rpaulo:
I'd prefer that it's NOT forbidden.  I'm fine with waiting for experimentation.

draft-ietf-quic-recovery.md
```diff
@@ -521,6 +521,10 @@ is available, or if the network changes, the initial RTT SHOULD be set to 500ms,
 resulting in a 1 second initial handshake timeout as recommended in
 {{?RFC6298}}.
 
+A connection MAY use the delay between sending a PATH_CHALLENGE and receiving
+a PATH_RESPONSE to seed initial_rtt for a new path, but the delay MUST not be
```
```suggestion
a PATH_RESPONSE to seed initial_rtt for a new path, but the delay MUST NOT be
```

ianswett:
I changed this to SHOULD NOT from MUST NOT to allow some experimentation.

PR #2688, opened 2019-05-10
Clarify server CONNECTION_CLOSE with Handshake

### LINKED ISSUE #2151: Where can you send CONNECTION_CLOSE
```
If the connection has been successfully established, endpoints MUST send any
CONNECTION_CLOSE frames in a 1-RTT packet.  Prior to connection establishment a
peer might not have 1-RTT keys, so endpoints SHOULD send CONNECTION_CLOSE frames
in a Handshake packet.  If the endpoint does not have Handshake keys, or it is
not certain that the peer has Handshake keys, it MAY send CONNECTION_CLOSE
frames in an Initial packet.  If multiple packets are sent, they can be
coalesced (see {{packet-coalesce}}) to facilitate retransmission.
```

Isn't the rule here that you always send it in the highest level?

ianswett:
Clarifies that if the server wants the client to process a connection close, it SHOULD send one in Initial and Handshake.

It seemed odd to me to say MAY send a CONNECTION_CLOSE in Initial, so I changed that to a SHOULD.  I can revert that and the other SHOULD, but I think the current PR makes more sense.

Came up in discussion on #2541 and I believe may be necessary to fully address #2541.  If not, I can open another issue.

Closes #2151

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2307,10 +2307,14 @@ signal closure.
 If the connection has been successfully established, endpoints MUST send any
 CONNECTION_CLOSE frames in a 1-RTT packet.  Prior to connection establishment a
 peer might not have 1-RTT keys, so endpoints SHOULD send CONNECTION_CLOSE frames
-in a Handshake packet.  If the endpoint does not have Handshake keys, or it is
-not certain that the peer has Handshake keys, it MAY send CONNECTION_CLOSE
-frames in an Initial packet.  If multiple packets are sent, they can be
-coalesced (see {{packet-coalesce}}) to facilitate retransmission.
+in a Handshake packet.If the endpoint does not have Handshake keys, it MAY
+send CONNECTION_CLOSE frames in an Initial packet.
+
+The server can have Handshake keys but not know that the client has Handshake
+keys.  In order to guarantee a CONNECTION_CLOSE is processed, it has to send
+a CONNECTION_CLOSE in both Handshake and Initial, because the client discards
+Initial keys as soon as it has Handshake keys. If multiple packets are sent,
+they can be coalesced (see {{packet-coalesce}}) to facilitate retransmission.
 
 
 ## Stateless Reset {#stateless-reset}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2307,7 +2307,7 @@ signal closure.
 If the connection has been successfully established, endpoints MUST send any
 CONNECTION_CLOSE frames in a 1-RTT packet.  Prior to connection establishment a
 peer might not have 1-RTT keys, so endpoints SHOULD send CONNECTION_CLOSE frames
-in a Handshake packet.If the endpoint does not have Handshake keys, it MAY
+in a Handshake packet.  If the endpoint does not have Handshake keys, it MAY
 send CONNECTION_CLOSE frames in an Initial packet.
 
 The server can have Handshake keys but not know that the client has Handshake
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2311,7 +2311,7 @@ the handshake, it is possible that more advanced packet protection keys are not
 available to the peer, so the frame MAY be replicated in a packet that uses a
 lower packet protection level.
 
-If the handshake is confirmed, endpoints MUST send any CONNECTION_CLOSE frames
+When the handshake is confirmed, an endpoint MUST send any CONNECTION_CLOSE frames
 in a 1-RTT packet.  Prior to handshake confirmation, the peer might not have
 1-RTT keys, so the endpoint SHOULD send CONNECTION_CLOSE frames in a Handshake
 packet.  If the endpoint does not have Handshake keys, it SHOULD send
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2307,14 +2307,14 @@ signal closure.
 If the connection has been successfully established, endpoints MUST send any
 CONNECTION_CLOSE frames in a 1-RTT packet.  Prior to connection establishment a
 peer might not have 1-RTT keys, so endpoints SHOULD send CONNECTION_CLOSE frames
-in a Handshake packet.  If the endpoint does not have Handshake keys, it MAY
+in a Handshake packet.  If the endpoint does not have Handshake keys, it SHOULD
 send CONNECTION_CLOSE frames in an Initial packet.
 
-The server can have Handshake keys but not know that the client has Handshake
-keys.  In order to guarantee a CONNECTION_CLOSE is processed, it has to send
-a CONNECTION_CLOSE in both Handshake and Initial, because the client discards
-Initial keys as soon as it has Handshake keys. If multiple packets are sent,
-they can be coalesced (see {{packet-coalesce}}) to facilitate retransmission.
+The server may not know whether the client has Handshake keys.  In order to
+guarantee a CONNECTION_CLOSE is processed, it SHOULD send a CONNECTION_CLOSE
+in both Handshake and Initial, because the client discards Initial keys as soon
+as it has Handshake keys. If multiple packets are sent, they can be coalesced
+(see {{packet-coalesce}}).
 
 
 ## Stateless Reset {#stateless-reset}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2308,7 +2308,7 @@ When sending CONNECTION_CLOSE, the goal is to ensure that the peer will process
 the frame.  Generally, this means sending the frame in a packet with the highest
 level of packet protection to avoid the packet being discarded.  However, during
 the handshake, it is possible that more advanced packet protection keys are not
-available to a peer, so the frame MAY be replicated in a packet that uses a
+available to the peer, so the frame MAY be replicated in a packet that uses a
 lower packet protection level.
 
 If the handshake is confirmed, endpoints MUST send any CONNECTION_CLOSE frames
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2311,10 +2311,10 @@ the handshake, it is possible that more advanced packet protection keys are not
 available to the peer, so the frame MAY be replicated in a packet that uses a
 lower packet protection level.
 
-When the handshake is confirmed, an endpoint MUST send any CONNECTION_CLOSE frames
-in a 1-RTT packet.  Prior to handshake confirmation, the peer might not have
-1-RTT keys, so the endpoint SHOULD send CONNECTION_CLOSE frames in a Handshake
-packet.  If the endpoint does not have Handshake keys, it SHOULD send
+After the handshake is confirmed, an endpoint MUST send any CONNECTION_CLOSE
+frames in a 1-RTT packet.  Prior to handshake confirmation, the peer might not
+have 1-RTT keys, so the endpoint SHOULD send CONNECTION_CLOSE frames in a
+Handshake packet.  If the endpoint does not have Handshake keys, it SHOULD send
 CONNECTION_CLOSE frames in an Initial packet.
 
 A client will always know whether the server has Handshake keys
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2321,8 +2321,8 @@ A client will always know whether the server has Handshake keys
 (see {{discard-initial}}), but it is possible that a server does not know
 whether the client has Handshake keys.  Under these circumstances, a server
 SHOULD send a CONNECTION_CLOSE frame in both Handshake and Initial packets
-to ensure that at least one of them is processable by the client. These packets
-can be coalesced into a single UDP datagram (see {{packet-coalesce}}).
+to ensure that at least one of them is processable by the client.  These
+packets can be coalesced into a single UDP datagram (see {{packet-coalesce}}).
 
 
 ## Stateless Reset {#stateless-reset}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2317,7 +2317,7 @@ in a 1-RTT packet.  Prior to handshake confirmation, a peer might not have
 packet.  If the endpoint does not have Handshake keys, it SHOULD send
 CONNECTION_CLOSE frames in an Initial packet.
 
-The server may not know whether the client has Handshake keys.  In order to
+A client will always know whether the server has Handshake keys (see {{discard-initial}}), but it is possible that a server does not know whether the client has Handshake keys.
 guarantee a CONNECTION_CLOSE is processed, it SHOULD send a CONNECTION_CLOSE
 in both Handshake and Initial, because the client discards Initial keys as soon
 as it has Handshake keys. If multiple packets are sent, they can be coalesced
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2317,11 +2317,12 @@ in a 1-RTT packet.  Prior to handshake confirmation, a peer might not have
 packet.  If the endpoint does not have Handshake keys, it SHOULD send
 CONNECTION_CLOSE frames in an Initial packet.
 
-A client will always know whether the server has Handshake keys (see {{discard-initial}}), but it is possible that a server does not know whether the client has Handshake keys.
-guarantee a CONNECTION_CLOSE is processed, it SHOULD send a CONNECTION_CLOSE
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2307,10 +2307,14 @@ signal closure.
 If the connection has been successfully established, endpoints MUST send any
 CONNECTION_CLOSE frames in a 1-RTT packet.  Prior to connection establishment a
 peer might not have 1-RTT keys, so endpoints SHOULD send CONNECTION_CLOSE frames
-in a Handshake packet.  If the endpoint does not have Handshake keys, or it is
-not certain that the peer has Handshake keys, it MAY send CONNECTION_CLOSE
-frames in an Initial packet.  If multiple packets are sent, they can be
-coalesced (see {{packet-coalesce}}) to facilitate retransmission.
+in a Handshake packet.If the endpoint does not have Handshake keys, it MAY
```
```suggestion
in a Handshake packet.  If the endpoint does not have Handshake keys, it MAY
```

draft-ietf-quic-transport.md
```diff
@@ -2307,10 +2307,14 @@ signal closure.
...
+send CONNECTION_CLOSE frames in an Initial packet.
+
+The server can have Handshake keys but not know that the client has Handshake
+keys.  In order to guarantee a CONNECTION_CLOSE is processed, it has to send
+a CONNECTION_CLOSE in both Handshake and Initial, because the client discards
+Initial keys as soon as it has Handshake keys. If multiple packets are sent,
+they can be coalesced (see {{packet-coalesce}}) to facilitate retransmission.
```
```suggestion
they can be coalesced (see {{packet-coalesce}}).
```

draft-ietf-quic-transport.md
```diff
@@ -2307,10 +2307,14 @@ signal closure.
...
-not certain that the peer has Handshake keys, it MAY send CONNECTION_CLOSE
-frames in an Initial packet.  If multiple packets are sent, they can be
-coalesced (see {{packet-coalesce}}) to facilitate retransmission.
+in a Handshake packet.If the endpoint does not have Handshake keys, it MAY
+send CONNECTION_CLOSE frames in an Initial packet.
+
+The server can have Handshake keys but not know that the client has Handshake
+keys.  In order to guarantee a CONNECTION_CLOSE is processed, it has to send
+a CONNECTION_CLOSE in both Handshake and Initial, because the client discards
+Initial keys as soon as it has Handshake keys. If multiple packets are sent,
```
```suggestion
Initial keys as soon as it has Handshake keys.  If multiple packets are sent,
```

draft-ietf-quic-transport.md
```diff
@@ -2307,10 +2307,14 @@ signal closure.
...
 peer might not have 1-RTT keys, so endpoints SHOULD send CONNECTION_CLOSE frames
-in a Handshake packet.  If the endpoint does not have Handshake keys, or it is
-not certain that the peer has Handshake keys, it MAY send CONNECTION_CLOSE
-frames in an Initial packet.  If multiple packets are sent, they can be
-coalesced (see {{packet-coalesce}}) to facilitate retransmission.
+in a Handshake packet.If the endpoint does not have Handshake keys, it MAY
+send CONNECTION_CLOSE frames in an Initial packet.
+
+The server can have Handshake keys but not know that the client has Handshake
+keys.  In order to guarantee a CONNECTION_CLOSE is processed, it has to send
```
nibanks: Should `has`  be `SHOULD`? If you're going to send a CONNECTION_CLOSE then you want it to be received. If the server isn't certain the client has the handshake keys, it really should send both, IMO.

draft-ietf-quic-transport.md
```diff
@@ -2307,10 +2307,14 @@ signal closure.
...
 CONNECTION_CLOSE frames in a 1-RTT packet.  Prior to connection establishment a
 peer might not have 1-RTT keys, so endpoints SHOULD send CONNECTION_CLOSE frames
-in a Handshake packet.  If the endpoint does not have Handshake keys, or it is
-not certain that the peer has Handshake keys, it MAY send CONNECTION_CLOSE
-frames in an Initial packet.  If multiple packets are sent, they can be
-coalesced (see {{packet-coalesce}}) to facilitate retransmission.
+in a Handshake packet.If the endpoint does not have Handshake keys, it MAY
+send CONNECTION_CLOSE frames in an Initial packet.
+
+The server can have Handshake keys but not know that the client has Handshake
```
nibanks: The `but not know` reads a little weird to me for some reason. Perhaps something along the lines of `but not be certain the client does.`?

draft-ietf-quic-transport.md
ianswett: Well, it both minimizes the number of datagrams and makes retransmission easier.  I'll update it with more verbose text about why this can facilitate retransmission.

draft-ietf-quic-transport.md
ianswett: I believe we've been using one space in this doc?

draft-ietf-quic-transport.md
ianswett: I tend to agree, but I was attempting to write non-normative text.  I'll change that and see how it reads.

draft-ietf-quic-transport.md
ianswett: Rewritten

draft-ietf-quic-transport.md
```diff
@@ -2307,10 +2307,14 @@ signal closure.
 If the connection has been successfully established, endpoints MUST send any
 CONNECTION_CLOSE frames in a 1-RTT packet.  Prior to connection establishment a
 peer might not have 1-RTT keys, so endpoints SHOULD send CONNECTION_CLOSE frames
-in a Handshake packet.  If the endpoint does not have Handshake keys, or it is
-not certain that the peer has Handshake keys, it MAY send CONNECTION_CLOSE
-frames in an Initial packet.  If multiple packets are sent, they can be
-coalesced (see {{packet-coalesce}}) to facilitate retransmission.
+in a Handshake packet.  If the endpoint does not have Handshake keys, it SHOULD
+send CONNECTION_CLOSE frames in an Initial packet.
```
janaiyengar: This needs to be only a statement of fact that describes the exception to the preceding SHOULD. "If the endpoint does not have Handshake keys, it sends CONNECTION_CLOSE frames in an Initial packet"

draft-ietf-quic-transport.md
```diff
@@ -2307,10 +2307,14 @@ signal closure.
...
+
+The server may not know whether the client has Handshake keys.  In order to
```
```suggestion
It is possible that an endpoint does not know whether the peer has Handshake keys.  In order to
```

draft-ietf-quic-transport.md
```diff
@@ -2307,10 +2307,14 @@ signal closure.
...
+guarantee a CONNECTION_CLOSE is processed, it SHOULD send a CONNECTION_CLOSE
```
janaiyengar: The server can't guarantee this. (see proposed text below)

draft-ietf-quic-transport.md
```diff
@@ -2307,10 +2307,14 @@ signal closure.
...
+in both Handshake and Initial, because the client discards Initial keys as soon
+as it has Handshake keys. If multiple packets are sent, they can be coalesced
+(see {{packet-coalesce}}).
```
janaiyengar: Suggested rewording of text after the first sentence:
"Under these circumstances, an endpoint SHOULD send a CONNECTION_CLOSE frame in both Handshake and Initial packets to increase the likelihood of a CONNECTION_CLOSE frame being processed by the peer.  These packets can be coalesced into a single UDP datagram (see {{packet-coalesce}})."

draft-ietf-quic-transport.md
ianswett: I'm fairly sure server to client is the only case two connection closes may need to be sent, which is why it's written with server an client.

draft-ietf-quic-transport.md
ianswett: A server can't guarantee the packet arrives, but I think sending both guarantees that if it does arrive at the client, the client can be decrypted.  Am I missing something?

draft-ietf-quic-transport.md
ianswett: I'm confused why we have a MUST and a SHOULD for the other two cases, but you're advising no normative text for the Initial case?  Am I missing something?

draft-ietf-quic-transport.md
martinthomson: A bigger point here is that the server also isn't sure when the client gets 1-RTT keys.  We should say the same about that transition.  I believe that we can rely on handshake confirmation for this transition.  Endpoints might choose to send CONNECTION_CLOSE in Handshake packets prior to the handshake being confirmed.

draft-ietf-quic-transport.md
ianswett: I have in mind 17.2.2.1 in Transport: https://tools.ietf.org/html/draft-ietf-quic-transport-20#section-17.2.2.1

"A client stops both sending and processing Initial packets when it sends its first Handshake packet."

This clearly indicates that a client will almost never be processing Initial and Handshake packets at once, so if you only send one connection close, some clients will drop it.  In all other cases with #2673 landed, the client and server should both still have Handshake keys even if they also have 1-RTT keys?

Is there something about TLS I'm implicitly assuming as well?  I can imagine I am, but I don't know what it is.

draft-ietf-quic-transport.md
ianswett: Agreed, another spot we're relying on #2673?

ianswett: I updated the text about sending 1-RTT connection closes to use Handshake confirmed, which lines up with #2673.  I think that's a good fix?

draft-ietf-quic-transport.md
martinthomson: Ahh, the two docs blur together in my mind.  This is OK.  I would suggest that you need to address both transitions here though.  The transition to Handshake and the transition to 1-RTT have the same sort of uncertainty.

draft-ietf-quic-transport.md
martinthomson: Don't sound so uncertain :)  I think that the fix is good.

And we'd (collectively) be crazy if we didn't take #2673.

MikeBishop: We'd also individually go crazy if we don't take it.  😉

draft-ietf-quic-transport.md
ianswett: The handshake to 1-RTT transition has an overlap where the peer has both sets of keys, unlike Initial to Handshake, so I think using handshake confirmed as the signal here provides what we want?

draft-ietf-quic-transport.md
```diff
@@ -2304,13 +2304,24 @@ the application requests that the connection be closed.  The application
...
+available to a peer, so the frame MAY be replicated in a packet that uses a
+lower packet protection level.
+
+If the handshake is confirmed, endpoints MUST send any CONNECTION_CLOSE frames
+in a 1-RTT packet.  Prior to handshake confirmation, a peer might not have
+1-RTT keys, so endpoints SHOULD send CONNECTION_CLOSE frames in a Handshake
+packet.  If the endpoint does not have Handshake keys, it SHOULD send
+CONNECTION_CLOSE frames in an Initial packet.
+
+The server may not know whether the client has Handshake keys.  In order to
```
```suggestion
A client will always know whether the server has Handshake keys (see {{discard-initial}}), but it is possible that a server does not know whether the client has Handshake keys.
```

draft-ietf-quic-transport.md
janaiyengar: "Under these circumstances, a server SHOULD send a CONNECTION_CLOSE frame in both Handshake and Initial packets to ensure that at least one of them is processable by the client. These packets can be coalesced into a single UDP datagram (see {{packet-coalesce}})."

draft-ietf-quic-transport.md
```diff
@@ -2304,13 +2304,25 @@ the application requests that the connection be closed.  The application
...
-in a Handshake packet.  If the endpoint does not have Handshake keys, or it is
-not certain that the peer has Handshake keys, it MAY send CONNECTION_CLOSE
-frames in an Initial packet.  If multiple packets are sent, they can be
-coalesced (see {{packet-coalesce}}) to facilitate retransmission.
+When sending CONNECTION_CLOSE, the goal is to ensure that a peer will process
+the frame.  Generally, this means sending the frame in a packet with the highest
+level of packet protection to avoid the packet being discarded.  However, during
+the handshake, it is possible that more advanced packet protection keys are not
+available to a peer, so the frame MAY be replicated in a packet that uses a
+lower packet protection level.
```
mikkelfj: This paragraph doesn't really say anything that the next paragraph doesn't cover. Also, `a peer` sounds like an endpoint can have multiple peers.

draft-ietf-quic-transport.md
```diff
@@ -2304,13 +2304,25 @@ the application requests that the connection be closed.  The application
...
+
+If the handshake is confirmed, endpoints MUST send any CONNECTION_CLOSE frames
```
```suggestion
When the handshake is confirmed, an endpoint MUST send any CONNECTION_CLOSE frames
```

draft-ietf-quic-transport.md
```diff
@@ -2304,13 +2304,25 @@ the application requests that the connection be closed.  The application
...
+in a 1-RTT packet.  Prior to handshake confirmation, a peer might not have
+1-RTT keys, so endpoints SHOULD send CONNECTION_CLOSE frames in a Handshake
```
```suggestion
1-RTT keys, so the endpoint SHOULD send CONNECTION_CLOSE frames in a Handshake
```

draft-ietf-quic-transport.md
```diff
@@ -2304,13 +2304,25 @@ the application requests that the connection be closed.  The application
...
-coalesced (see {{packet-coalesce}}) to facilitate retransmission.
+When sending CONNECTION_CLOSE, the goal is to ensure that a peer will process
+the frame.  Generally, this means sending the frame in a packet with the highest
+level of packet protection to avoid the packet being discarded.  However, during
+the handshake, it is possible that more advanced packet protection keys are not
+available to a peer, so the frame MAY be replicated in a packet that uses a
+lower packet protection level.
+
+If the handshake is confirmed, endpoints MUST send any CONNECTION_CLOSE frames
+in a 1-RTT packet.  Prior to handshake confirmation, a peer might not have
```
```suggestion
in a 1-RTT packet.  Prior to handshake confirmation, the peer might not have
```

draft-ietf-quic-transport.md
ianswett: That makes it sound a bit too much like one should close the connection upon completing the handshake.  I think If works better, though "After" could also work.

draft-ietf-quic-transport.md
```diff
@@ -2304,13 +2304,25 @@ the application requests that the connection be closed.  The application
...
 signal closure.
 
-If the connection has been successfully established, endpoints MUST send any
-CONNECTION_CLOSE frames in a 1-RTT packet.  Prior to connection establishment a
-peer might not have 1-RTT keys, so endpoints SHOULD send CONNECTION_CLOSE frames
-in a Handshake packet.  If the endpoint does not have Handshake keys, or it is
-not certain that the peer has Handshake keys, it MAY send CONNECTION_CLOSE
-frames in an Initial packet.  If multiple packets are sent, they can be
-coalesced (see {{packet-coalesce}}) to facilitate retransmission.
+When sending CONNECTION_CLOSE, the goal is to ensure that a peer will process
```
```suggestion
When sending CONNECTION_CLOSE, the goal is to ensure that the peer will process
```

draft-ietf-quic-transport.md
```diff
@@ -2304,13 +2304,25 @@ the application requests that the connection be closed.  The application
...
+the frame.  Generally, this means sending the frame in a packet with the highest
+level of packet protection to avoid the packet being discarded.  However, during
+the handshake, it is possible that more advanced packet protection keys are not
+available to a peer, so the frame MAY be replicated in a packet that uses a
```
```suggestion
available to the peer, so the frame MAY be replicated in a packet that uses a
```

draft-ietf-quic-transport.md
janaiyengar: I missed this earlier, but I agree. I'm fine with dropping this paragraph entirely. It seems like it's simply repeating what's in the next para, but in more general terms.

ianswett: I don't have a strong opinion, but I see your point, this was @martinthomson addition, so I'll let him comment.

martinthomson: The point was to articulate the principles behind the design.  It might be that the mechanics are sufficiently precise that that is not necessary, but we did have #2151.  @ianswett's changes didn't really directly address that.

martinthomson: If people want to fix this, please open a followup issue.

martinthomson:
We might just delete this tomorrow, but this is strictly better than what we have now.

PR #2695, opened 2019-05-13
Push ID limit checking in DUPLICATE_PUSH

### LINKED ISSUE #2681: DUPLICATE_PUSH push ID validation
Two questions: 
1. Should DUPLICATE_PUSH respect max push ID limit?
  PUSH_PROMISE says the following:

 > A server MUST NOT use a Push ID that is larger than the client has
   provided in a MAX_PUSH_ID frame (Section 4.2.8) and MUST NOT use the
   same Push ID in multiple PUSH_PROMISE frames

2. What should a client do if it sees a DUPLICATE_PUSH for an ID of a PUSH_PROMISE that it never received?

This could be caused by a race condition, so it seems sensible to allow this to clients to support this without error. We should probably document that like we did for other racy things. E.g.

> If the client receives a CANCEL_PUSH frame, that frame might identify
   a Push ID that has not yet been mentioned by a PUSH_PROMISE frame.

MikeBishop:
Fixes #2681.

Skirts the editorial/design line.  Previously, DUPLICATE_PUSH referred to a Push ID in a previously-sent PUSH_PROMISE frame; that frame would have caused a connection error on arrival.  This says that the error should be sent regardless of which frame arrives first.

May need reconciling with #2662.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -890,7 +890,11 @@ of a DUPLICATE_PUSH frame as a connection error of type HTTP_UNEXPECTED_FRAME.
 
 The DUPLICATE_PUSH frame carries a single variable-length integer that
 identifies the Push ID of a resource that the server has previously promised
-(see {{frame-push-promise}}).
+(see {{frame-push-promise}}), though that promise might not be received before
+this frame.  A server MUST NOT use a Push ID that is larger than the client has
+provided in a MAX_PUSH_ID frame ({{frame-max-push-id}}).  A client MUST treat
+receipt of a PUSH_PROMISE that contains a larger Push ID than the client has
+advertised as a connection error of type HTTP_MALFORMED_FRAME.
 
 This frame allows the server to use the same server push in response to multiple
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -890,7 +890,11 @@ of a DUPLICATE_PUSH frame as a connection error of type HTTP_UNEXPECTED_FRAME.
 
 The DUPLICATE_PUSH frame carries a single variable-length integer that
 identifies the Push ID of a resource that the server has previously promised
-(see {{frame-push-promise}}).
+(see {{frame-push-promise}}), though that promise might not be received before
+this frame.  A server MUST NOT use a Push ID that is larger than the client has
+provided in a MAX_PUSH_ID frame ({{frame-max-push-id}}).  A client MUST treat
+receipt of a PUSH_PROMISE that contains a larger Push ID than the client has
```
martinthomson: or DUPLICATE_PUSH?

MikeBishop: Derp, yes.  Copied the text and didn't get this changed.

PR #2703, opened 2019-05-15
CONNECT error:  MUST close, SHOULD RST

### LINKED ISSUE #2228: How do I generate an RST?
S 5.2.

```
A TCP connection error is signaled with QUIC RESET_STREAM frame. A proxy treats
any error in the TCP connection, which includes receiving a TCP segment with the
RST bit set, as a stream error of type HTTP_CONNECT_ERROR
({{http-error-codes}}).  Correspondingly, a proxy MUST send a TCP segment with
the RST bit set if it detects an error with the stream or the QUIC connection.
```

I don't believe this is possible on UNIX. What system call do I use to
generate RST? Neither close() nor shutdown() does it.

MikeBishop:
Fixes #2228.


ekr:
LGTM

PR #2710, opened 2019-05-16
Split SETTINGS handling error text

### LINKED ISSUE #2706: Second SETTINGS frame on any stream other than the control stream error
> SETTINGS frames always apply to a connection, never a single stream. A SETTINGS frame MUST be sent as the first frame of each control stream (see Section 4.2.1) by each peer, and MUST NOT be sent subsequently or on any other stream. If an endpoint receives a SETTINGS frame on a different stream, the endpoint MUST respond with a connection error of type HTTP_WRONG_STREAM. If an endpoint receives a second SETTINGS frame, the endpoint MUST respond with a connection error of type HTTP_UNEXPECTED_FRAME.

It's unclear which error should be reported if an endpoint receives a second SETTINGS frame on any stream that's not a control stream. Perhaps this should be explicitly defined in the draft?

LPardue:
Fixes #2706 by clearly splitting the two different error cases.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1248,11 +1248,13 @@ identifier" and a "setting value".
 
 SETTINGS frames always apply to a connection, never a single stream.  A SETTINGS
 frame MUST be sent as the first frame of each control stream (see
-{{control-streams}}) by each peer, and MUST NOT be sent subsequently or on any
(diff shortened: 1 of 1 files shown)


MikeBishop:
Moves existing text to be clearer, but doesn't change the stated requirements.  Editorial, and fine by me.

PR #2701, opened 2019-05-15
Reorder sections

MikeBishop:
Fixes #2220; no text changes outside the Document Organization section.

PR #2712, opened 2019-05-16
Specify error on receipt a DATA frame before a HEADERS frame on a request stream

DaanDeMeyer:
The current draft doesn't mention what to do when receiving a DATA frame before a HEADERS frame on a request stream. This pr explicitly requires recipients to respond with a connection error of type HTTP_UNEXPECTED_FRAME.

I'm not entirely sure the DATA frame definition section is the best place for this addition (if it's a valid addition). Another possibility would be the HTTP Message Exchanges section. Any input on this is appreciated.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1248,11 +1248,13 @@ identifier" and a "setting value".
 
 SETTINGS frames always apply to a connection, never a single stream.  A SETTINGS
 frame MUST be sent as the first frame of each control stream (see
-{{control-streams}}) by each peer, and MUST NOT be sent subsequently or on any
-other stream. If an endpoint receives a SETTINGS frame on a different stream,
-the endpoint MUST respond with a connection error of type HTTP_WRONG_STREAM. If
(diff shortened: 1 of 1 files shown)


LPardue:
I think this is good text to add. I'll defer _exactly where_ to Mike.

Note that this problem also affects push streams, not just request streams. That is, the pushed DATA must not appear before the pushed HEADERS. 

So it seems to me that the Message Exchange section might help us cover all cases. But maybe not.

MikeBishop:
I tend to lean toward the Message Exchange section -- this is a question of getting frames in an order that can't possibly carry a valid HTTP message.

MikeBishop:
Ack, I was fixing the fact that #2701 broke everything and messed this up.  Once the PR closed, I lost access to @DaanDeMeyer's branch to un-mess it.  Moving to a quicwg branch.

MikeBishop:
Bah, and it won't let me make this PR point to the new branch, either.  Grr, GitHub.  Grr.  This is now #2715.

PR #2727, opened 2019-05-20
Clarify the text recommending replacing CIDs when consumed

### LINKED ISSUE #2725: We lost the text recommending replacement of used CIDs
We previously had text: 
```
The endpoint SHOULD do this by supplying a new connection ID when a connection
ID is retired by its peer or when the endpoint receives a packet with a
previously unused connection ID.
```

In adding text for the TP (active_connection_id_limit), we kept the same idea, but lost a little bit of the text: 
```
An endpoint SHOULD NOT provide more connection IDs than the peer's limit.  If an
endpoint has provided its peer with the maximum number of connection IDs, it
SHOULD only provide a new connection ID when the peer retires one.
```

We still need the SHOULD replace used CIDs in the event that we *haven't* provided the peer with the maximum number of CIDs.

erickinnear:
Reword the text to be a bit more clear about what's expected from a CID issuer. 

Closes #2725.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -951,7 +951,7 @@ supplied connection ID is 1.
 
 Additional connection IDs are communicated to the peer using NEW_CONNECTION_ID
 frames ({{frame-new-connection-id}}).  The sequence number on each newly-issued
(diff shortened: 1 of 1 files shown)


martinthomson:
This inverts the form of the recommendation, but doesn't really change the design.

PR #2731, opened 2019-05-20
Transport editorial suggestions

ianswett:
Qualifies a MUST with an 'otherwise', since otherwise it would appear to directly contradict previous normative text.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1327,8 +1327,8 @@ Initial or Retry packet from the server, it populates the Destination Connection
 ID field with an unpredictable value.  This MUST be at least 8 bytes in length.
 Until a packet is received from the server, the client MUST use the same value
 unless it abandons the connection attempt and starts a new one. The initial
-Destination Connection ID is used to determine packet protection keys for Initial
-packets.
+Destination Connection ID is used to determine packet protection keys for
+Initial packets.
 
 The client populates the Source Connection ID field with a value of its choosing
 and sets the SCIL field to indicate the length.  The first flight of 0-RTT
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1336,11 +1335,7 @@ and sets the SCIL field to indicate the length.
 The first flight of 0-RTT packets use the same Destination and Source Connection
 ID values as the client's first Initial.
 
-The Destination Connection ID field in the server's Initial packet contains a
-connection ID that is chosen by the recipient of the packet (i.e., the client);
-the Source Connection ID includes the connection ID that the sender of the
-packet wishes to use (see {{connection-id}}). The server MUST use consistent
-Source Connection IDs during the handshake.
+The server MUST use consistent Source Connection IDs during the handshake.
```
martinthomson: This implies that the server cannot change its Source Connection ID between sending a Retry and sending an Initial.  I hope that we agree that this is an implausible requirement.  The consistency applies only after sending an Initial.

ianswett: Weirdly, this text was already there, see lines 1342/1343.

But I agree, this is clearly incorrect or depends upon the definition of consistent, which I don't believe is worth defining.  I'm going to remove this paragraph.

draft-ietf-quic-transport.md
```diff
@@ -1323,26 +1322,20 @@ Destination Connection ID field of packets being sent to them.  Upon receiving a
 packet, each endpoint sets the Destination Connection ID it sends to match the
 value of the Source Connection ID that they receive.
 
-When an Initial packet is sent by a client which has not previously received a
-Retry packet from the server, it populates the Destination Connection ID field
-with an unpredictable value.  This MUST be at least 8 bytes in length. Until a
-packet is received from the server, the client MUST use the same value unless it
-abandons the connection attempt and starts a new one. The initial Destination
-Connection ID is used to determine packet protection keys for Initial packets.
+When an Initial packet is sent by a client which has not previously received an
```
```suggestion
When an Initial packet is sent by a client that has not previously received an
```

PR #2733, opened 2019-05-21
Allow reuse of stateless reset tokens

### LINKED ISSUE #2732: Reusing SRTs
We hit this during interop, where one implementation would re-issue the same SRT(Stateless Reset Token) it included in the TPs in each new NEW_CONNECTION_ID frame. The question was raised whether that was allowed or not, and @martinthomson decided that "design text was needed."

martinthomson:
This came up as the result of an incomplete implementation, but there
are - maybe - good reasons to have the same stateless reset token being
used for multiple connection IDs on the same connection.

There are drawbacks however.  Reuse of a stateless reset token means
that you have to remember (and consistently route) connection IDs, even
when they are retired.

This text makes a note of the possibility and explains the drawbacks.

Closes #2732.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2477,13 +2477,15 @@ that is reset by revealing the Stateless Reset Token MUST NOT be reused for new
 connections at nodes that share a static key.
 
 The same Stateless Reset Token MAY be used for multiple connection IDs on the
-same connection.  An endpoint that reuses a Stateless Reset Token MUST ensure
-that packets with Destination Connection ID field values that correspond to a
-reused Stateless Reset Token are attributed to the same connection, even when
-the connection ID has been retired.  Otherwise, an attacker might be able to
-send a packet with a retired connection ID and cause the endpoint to produce a
-Stateless Reset that it can use to disrupt the connection; just as with the
-attacks in {{reset-oracle}}.
+same connection.  However, reuse of a Stateless Reset Token might expose an
+endpoint to denial of service if associated connection IDs are forgotten while
+the associated token is still active at a peer.  An endpoint MUST ensure that
+packets with Destination Connection ID field values that correspond to a reused
+Stateless Reset Token are attributed to the same connection as long as the
+Stateless Reset Token is still usable, even when the connection ID has been
+retired.  Otherwise, an attacker might be able to send a packet with a retired
+connection ID and cause the endpoint to produce a Stateless Reset that it can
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2476,6 +2476,14 @@ the same static key (see {{reset-oracle}}).  A connection ID from a connection
...
 
+The same Stateless Reset Token MAY be used for multiple connection IDs on the
+same connection.  An endpoint that reuses a Stateless Reset Token MUST ensure
+that any connection ID used on the connection is matched to the active
+connection, including any routing performed at load balancers, even when a
+connection ID has been retired.  Otherwise, an attacker might be able to send a
+packet with a retired connection ID and cause the endpoint to produce a
+Stateless Reset.
+
 Note that Stateless Reset packets do not have any cryptographic protection.
```
mikkelfj: I don't understand this text. I can see the issuer of the reset can and should coordinate with load balancers, but the text reads as if the sender is responsible for the receiver matching up the connection ID, or that an active connection should be matched sender side when the reset is transmitted to cancel an unknown connection, which by definition isn't possible. In part it is unclear what is mean by connection ID since there are two kinds.

mnot: @mikkelfj can you clarify what this comment is in relation to?

martinthomson: I've tweaked the text (there was a grammatical error), but I think that this was related to the change.  It was a minor editorial thing.  It's really hard to get this right.

mikkelfj: The text has been changed since I wrote the above, but largely:

> An endpoint that reuses a Stateless Reset Token MUST ensure
that any connection ID associated with the reused value is correlated with to
the active connection, even when the connection ID has been retired.

It is not entirely clear what timeframe this pertains to. There is time where the an unknown packet is received and the connection is active from the peers perspective but no longer used from the receivers perspective. And there is the time where the token is issued during connection establishment or path migration. The latter is the intent, but it is not easy to read.

It could perhaps be formulated like

A Stateless Reset Token MUST be unique to a single connection. The same token MAY be reissued for different Connection ID's on the same connection when it can be guaranteed that they all route to the same endpoint for the duration of the connection even after a Connection ID has been retired.

martinthomson: That doesn't quite work because it is OK to forget a connection ID if the associated stateless reset token is no longer in use.  Maybe I need to try again.

PR #2743, opened 2019-05-22
Better articulate principles for ciphersuites

### LINKED ISSUE #2742: Automatic applicability of new TLS cipher suites
Suppose that we had a future TLS cipher suite such as AES_OCB_SHA256. Would that work with QUIC without a new QUIC draft? The AEAD part is already clearly defined and given that AES-OCB uses AES, the same HP mechanism as we currently use with GCM can be used here. 
From my perspective, the desirable property is that if we add new TLS cipher suites with analogous cipher cores (block cipher and/or nonce-based stream cipher) to the existing spec, then no new QUIC document should be required.

An alternative design would be to just require a "QUIC applicability" paragraph in new TLS cipher suites analogous to the DTLS-OK IANA flag.

(see also #2019).

martinthomson:
For input to discussion on the related issues.  This says that we don't
do CCM_8 and makes it clear that you don't reject an attempt to use a
ciphersuite you don't understand.

This is an editorial resolution to the design issue.

Closes #2742, #2682, #2581.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -833,9 +833,9 @@ and produce an output 16 bytes larger than their input.
 
 Note:
 
-: An endpoint MUST NOT reject a ClientHello that offers a ciphersuite that
-  offers a ciphersuite that it does not support, or it would be impossible to
-  deploy a new ciphersuite.  This also applies to TLS_AES_128_CCM_8_SHA256.
+: An endpoint MUST NOT reject a ClientHello that offers a ciphersuite that it
+  does not support, or it would be impossible to deploy a new ciphersuite.  This
+  also applies to TLS_AES_128_CCM_8_SHA256.
 
 The key and IV for the packet are computed as described in {{protection-keys}}.
 The nonce, N, is formed by combining the packet protection IV with the packet
@@ -969,7 +969,7 @@ sample.
 
 To ensure that sufficient data is available for sampling, packets are padded so
 that the combined lengths of the encoded packet number and protected payload is
-at least 4 bytes longer than the sample required for header protection.  For the
+at least 4 bytes longer than the sample required for header protection.  The
 ciphersuites defined in {{?TLS13}} - other than TLS_AES_128_CCM_8_SHA256, for
 which a header protection scheme is not defined in this document - have 16-byte
 expansions and 16-byte header protection samples.  This results in needing at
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -964,10 +970,12 @@ sample.
 To ensure that sufficient data is available for sampling, packets are padded so
 that the combined lengths of the encoded packet number and protected payload is
 at least 4 bytes longer than the sample required for header protection.  For the
-AEAD functions defined in {{?TLS13}}, which have 16-byte expansions and 16-byte
-header protection samples, this results in needing at least 3 bytes of frames in
-the unprotected payload if the packet number is encoded on a single byte, or 2
-bytes of frames for a 2-byte packet number encoding.
+ciphersuites defined in {{?TLS13}} - other than TLS_AES_128_CCM_8_SHA256, for
```
nharper: Nit: Remove "For" (the sentence should start "The ciphersuites defined in").

draft-ietf-quic-tls.md
```diff
@@ -825,11 +825,17 @@ This provides protection against off-path attackers and robustness against QUIC
...
-an output 16 bytes larger than their input.
+TLS_AES_128_CCM_8_SHA256.  A ciphersuite MUST NOT be negotiated unless a header
+protection scheme is defined for the ciphersuite.  This document defines a
+header protection scheme for all ciphersuites defined in {{!TLS13}} aside from
+TLS_AES_128_CCM_8_SHA256.  These ciphersuites have a 16-byte authentication tag
+and produce an output 16 bytes larger than their input.
+
+Note:
+
+: An endpoint MUST NOT reject a ClientHello that offers a ciphersuite that
```
rpaulo: duplicate "that offers a cipher suite"

ekr:
LGTM

mikkelfj:
I think I must be entirely missing the point, but I don't understand this:

> An endpoint MUST NOT reject a ClientHello that offers a ciphersuite that ...

Can you please explain how this works? For the sake of argument, TLS_AES_128_CCM_8_SHA256 is not supported, but the peer chooses to use it, and the TLS engine behind it all decides to support it via some configuration that the endpoint is directly aware of.

How does the endpoint now deal with header protection. Is it assumed that this is a TLS black box so it just magically works if the TLS engine has accepted the cipher via some useful API call.

If it is a TLS blackbox API call, how would that work with HW offloading?

mikkelfj:
Now I understand it. Being offered by TLS is not the same as being used by any of the endpoints. Sorry for the noise.

PR #2749, opened 2019-05-22
Allow longer CIDs

### LINKED ISSUE #2736: More Connection ID Space
Although it is quite late in the day for this objection, I find that we already bumping up against the limits of the 18 byte connection ID.

Inevitably, this touches on some issues we're exploring in the quic-load-balancers draft, which isn't adopted, and many of you are no doubt happily ignoring.

There are basically three uses of the 144 bits of CID entropy in QUIC-LB:

1) Server ID encoding -- as there are only so many servers one might have, 16 bits is usually enough for this.
2) "Private" server entropy -- the load balancer doesn't care what the bits are, but they're protected by crypto. The server created and encrypted the CID, so it can use this for its local purposes, but no one else can. The amount of this depends, but if you're using a 256-bit block cipher it's something like 240 bits -- plenty
3)  "Public" entropy -- not encrypted, these bits are readable by anyone who is familiar with the encoding scheme. With a 16 B encryption block, we have only 16 bits to play with here.

QUIC-LB is already taking 2 of these last bits for config rotation, which allows server pools to incrementally deploy new CID keys or encoding schemes. So we're down to 14 bits.

The public entropy is useful for a growing pile of trusted hardware devices that are presumably not going to perform CID decryption. On f5 boxes, we have a hardware disaggregator for the many dataplane processors and threads that are operating. Today, we'd use perhaps 8 of the bits, but in future architectures even 14 might not quite be enough.

And now, having chatting with Manasi Deval from Intel after her Prague talk about hardware offload, I'm fairly sure we're going to need 4 to cheaply encode the connection ID length, at least for servers that want to use this service.

If we ever wanted a solution that allowed multiple tiers of load balancing, which is not currently in scope for QUIC-LB, we don't have nearly enough space.

While I think that existing use cases can just barely make do with what we have, it is a bad sign if something in the invariants is *already* constraining use cases.

Offline in London, we've circulated a number of possible approaches to this, but the first-order discussion is a willingness to reopen the invariants. If so, there are a couple of solutions that won't break existing load balancer deployments.

I'm happy to write a PR if this is the consensus.

martinduke:
Fixes #2736.

As a straw man, I selected a v1 limit of 48 bytes to reflect comments in London. Note that we cannot grease the unused bits of the CIL fields if we're going to make load balancers robust to version changes (which is the point in having CIDs in the invariants, IIRC).


dtikhonov:
It would be nice if an invalid length could not be expressed.  Is 48 a completely arbitrary number?

martinduke:
@dtikhonov somewhat arbitrary. @ianswett wanted it smaller than a 64B cacheline; the most extreme use case we came up with was "more than 32B".

However, I'd prefer that we not artificially limit it in the invariants, because my imagination is limited.

mikkelfj:
It's great to have CID's of length 1..3 bytes, but perhaps some guidelines are needed as there is an increased collision chance and some attack mitigations might make assumptions on the CID being at least 32 bits.

draft-ietf-quic-invariants.md
```diff
@@ -143,11 +143,13 @@ version-specific semantics are marked with an X.
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                         Version (32)                          |
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|DCIL(4)|SCIL(4)|
+|    DCIL (8)   |
```
martinthomson: I think that we should have a varint length.

draft-ietf-quic-transport.md
```diff
@@ -2345,7 +2345,7 @@ following layout:
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|0|1|               Unpredictable Bits (182..)                ...
+|0|1|               Unpredictable Bits (422..)                ...
```
martinthomson: This one remains particularly thorny.  Now there is no way to generate a stateless reset that is a reasonable size.  Yes, this is technically bounded, but now the naive stateless reset is considerably bigger than the packet that triggers it in many more cases.

martinduke: Isn't there a rule that you can't send a reset that's biger than the originating packet?

mikkelfj: eternal ping pong rule

mikkelfj: you can get around sending a reset only once per every 4 received packets on average, but consensus dislike random.

draft-ietf-quic-invariants.md
martinduke: Discussed in the room; though varints would help with greasing; it is a pain and creates serious problems for current deployments.

draft-ietf-quic-transport.md
martinduke: discussed in London; I think we're OK with this

martinduke:
@DavidSchinazi Raises a good point; for purposes of Version Negotiation, v1 servers need to accommodate longer CIDs (up to 255 bytes). I'll make the change.

dtikhonov:
> for purposes of Version Negotiation, v1 servers need to accommodate longer CIDs (up to 255 bytes).

martinduke:
@dtikhonov in v2, we might allow 255B CIDs. So to send a proper VN packet in response to v2 we need to handle that entire CID.

DavidSchinazi:
This PR looks good to me. Thanks for making an encoding that we can deploy in parallel with older versions on the same server.

martinduke:
OK, the transport drafts now acknowledges that it may have to send VN with CIL > 48 bytes.

I believe all known issues are resolved.

draft-ietf-quic-transport.md
```diff
@@ -3446,30 +3448,37 @@ Version:
...
-: The Destination Connection ID field follows the connection ID lengths and is
-  either 0 bytes in length or between 4 and 18 bytes.
-  {{negotiating-connection-ids}} describes the use of this field in more detail.
+: The Destination Connection ID field follows the DCIL and is between 0 and 48
+  bytes in length. {{negotiating-connection-ids}} describes the use of this
+  field in more detail.
+
+SCIL:
+
+: The byte following the destination Connection ID contains the length in bytes
```
```suggestion
: The byte following the Destination Connection ID contains the length in bytes
```

draft-ietf-quic-transport.md
```diff
@@ -3446,30 +3448,37 @@ Version:
...
+  of the source connection ID field that follows it.  This length is encoded as
```
```suggestion
  of the Source Connection ID field that follows it.  This length is encoded as
```

draft-ietf-quic-transport.md
```diff
@@ -3577,6 +3588,10 @@ connection IDs gives clients some assurance that the server received the packet
 and that the Version Negotiation packet was not generated by an off-path
 attacker.
 
+As future versions of QUIC may support Connection IDs larger than the version 1
+limit, Version Negotiation Packets will sometimes exceed the Connection ID
```
```suggestion
limit, Version Negotiation Packets MAY carry Connection IDs that are longer than 48 bytes.
```

draft-ietf-quic-transport.md
```diff
@@ -3577,6 +3588,10 @@ connection IDs gives clients some assurance that the server received the packet
...
+length limit of 48 bytes in version 1.
```
```suggestion

```

draft-ietf-quic-transport.md
martinthomson: s/MAY/could - it's not the place of this specification to give permission to break its own rules.

DavidSchinazi:
This CL also needs to change the encoding of the Original Destination Connection ID Length (ODCIL) in Retry packets. It is currently encoded as the four least significant bits of the first byte and that won't fit anymore. I'd suggest moving ODCIL to be a full byte between the Source Connection ID and the Original Destination Connection ID. I would then mark the four least significant bits of the first byte of Retry packets as unused similar to Version Negotiation packets.

```
+-+-+-+-+-+-+-+-+
|1|1| 3 | Unused|
|                         Version (32)                          |
|    DCIL (8)   |
|               Destination Connection ID (0..384)            ...
|    SCIL (8)   |
|                 Source Connection ID (0..384)               ...
|   ODCIL (8)   |
|          Original Destination Connection ID (0..384)        ...
|                        Retry Token (*)                      ...
```

martinduke:
@DavidSchinazi Nice catch. It should be fixed now.

draft-ietf-quic-transport.md
```diff
@@ -3814,33 +3835,35 @@ wishes to perform a retry (see {{validate-handshake}}).
...
+ addition to the long header, it contains these additional fields:
 
 ODCIL:
 
-: The four least-significant bits of the first byte of a Retry packet are not
-  protected as they are for other packets with the long header, because Retry
-  packets don't contain a protected payload.  These bits instead encode the
-  length of the Original Destination Connection ID field.  The length uses the
-  same encoding as the DCIL and SCIL fields.
+: These bits the length of the Original Destination Connection ID field.  The
```
DavidSchinazi: Typo `These bits the length`

How about mimicking the SCIL field description?
```
ODCIL:

  The byte following the Source Connection ID contains the length in bytes
  of the Original Destination Connection ID field that follows it.  This length
  is encoded as a 8-bit unsigned integer.  In QUIC version 1, this value
  MUST NOT exceed 48 bytes.
```

draft-ietf-quic-transport.md
```diff
@@ -1588,15 +1588,15 @@ A token SHOULD be constructed for the server to easily distinguish it from
 tokens that are sent in Retry packets as they are carried in the same field.
 
 The token MUST NOT include information that would allow it to be linked by an
-on-path observer to the connection on which it was issued.  For example, it
+on-path observer to the connection on which it was issued. For example, it
```
```suggestion
on-path observer to the connection on which it was issued.  For example, it
```

draft-ietf-quic-transport.md
```diff
@@ -1588,15 +1588,15 @@ A token SHOULD be constructed for the server to easily distinguish it from
...
 cannot include the connection ID or addressing information unless the values are
 encrypted.
 
 Unlike the token that is created for a Retry packet, there might be some time
 between when the token is created and when the token is subsequently used.
 Thus, a token SHOULD have an expiration time, which could be either an explicit
 expiration time or an issued timestamp that can be used to dynamically calculate
-the expiration time.  A server can store the expiration time or include it in an
+the expiration time. A server can store the expiration time or include it in an
```
```suggestion
the expiration time.  A server can store the expiration time or include it in an
```

draft-ietf-quic-transport.md
ianswett: So we're increasing the min size of a stateless reset token?  I'd really rather not, since that increases the chances of that being a DoS vector.  Keeping the ratio of incoming to outgoing bytes ~1:1 seems important for DoS.  As a server operator, I don't want to not respond to a client's PING only packet with a stateless reset just because it's too small, but I also don't want to respond to every small packet I don't have state for.

mikkelfj: I still think we should allow responding to smaller packets at some random factor rather than not responding at all. This does not mean we should not try to keep resets small though.

draft-ietf-quic-transport.md
```diff
@@ -2368,7 +2368,7 @@ of the datagram contain a Stateless Reset Token.
 
 To entities other than its intended recipient, a stateless reset will appear
 to be a packet with a short header.  For the packet to appear as valid, the
-Unpredictable Bits field needs to include at least 182 bits of data (or 23
+Unpredictable Bits field needs to include at least 422 bits of data (or 53
 bytes, less the two fixed bits).  This is intended to allow for a Destination
 Connection ID of the maximum length permitted, with a minimal packet number, and
 payload.  The Stateless Reset Token corresponds to the minimum expansion of the
```
martinthomson: A bit further down, there is a 40 and 39 that need to be changed to 70 and 69.

draft-ietf-quic-transport.md
ianswett: #2770 filed as a separate issue for this.

martinduke: OK, if you agree that this issue is non-blocking I'll proceed.

draft-ietf-quic-transport.md
```diff
@@ -3577,6 +3588,10 @@ connection IDs gives clients some assurance that the server received the packet
 and that the Version Negotiation packet was not generated by an off-path
 attacker.
 
+As future versions of QUIC may support Connection IDs larger than the version 1
+limit, Version Negotiation Packets MAY carry Connection IDs that are longer than
```
```suggestion
limit, Version Negotiation Packets could carry Connection IDs that are longer than
```
@martinthomson pointed out that this isn't the spec where we need to allow larger CIDs to happen. This merely needs to note that it could happen.

mnot:
~~~
3592: Line is 82 characters; limit is 80
limit, Version Negotiation Packets could carry Connection IDs that are longer than
Makefile:25: recipe for target 'lint' failed
~~~

@martinduke can you fix that up?

draft-ietf-quic-invariants.md
martinthomson: Maybe not for this PR, but something to consider...

"DCIL" was chosen to fit in the small space available in the diagram.  The new field is a little wider and could fit `DCID Len (8)` comfortably.

draft-ietf-quic-invariants.md
```diff
@@ -159,21 +161,13 @@ All other bits in that byte are version specific.
...
-The connection ID lengths are followed by two connection IDs.  The connection
-ID associated with the recipient of the packet (the Destination Connection ID)
-is followed by the connection ID associated with the sender of the packet (the
-Source Connection ID).
+The next byte contains the length in bytes of the destination Connection ID
+(see {{connection-id}}) that follows.  The length is encoded as an 8-bit
+unsigned integer. This is followed by the destination Connection ID itself,
+associated with the receiver of the packet.
+
+The next two fields are the Connection ID length for the source and the source
```
ianswett:
```suggestion
The next two fields are the length in bytes of the source Connection ID and the source
```

draft-ietf-quic-transport.md
```diff
@@ -3577,6 +3588,10 @@ connection IDs gives clients some assurance that the server received the packet
 and that the Version Negotiation packet was not generated by an off-path
 attacker.
 
+As future versions of QUIC may support Connection IDs larger than the version 1
+limit, Version Negotiation Packets could carry Connection IDs that are longer
```
```suggestion
limit, Version Negotiation packets could carry Connection IDs that are longer
```

draft-ietf-quic-invariants.md
martinduke: I am happy to commit this, but it's 'outdated'. Can you update?

PR #2751, opened 2019-05-22
Enable/disable/coordinate spin per path

### LINKED ISSUE #2720: Privacy of spin with multiple connections on same 5 tuple
Text says:

> In case multiple connections share the same five-tuple, that is, have the same source and destination IP address and UDP ports, endpoints should try to co-ordinate across all connections to ensure a clear signal to any on-path measurement points.

How would that work when 1/8th of the connections are not spinning? Wouldn't that either be sending a misleading signal, or violate the 1/8th rule?

martinthomson:
Closes #2720.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -4019,14 +4019,14 @@ to disable the spin bit either globally or on a per-connection basis. Even when
 the spin bit is not disabled by the administrator, implementations MUST disable
 the spin bit for a given connection with a certain likelihood. The random
 selection process SHOULD be designed such that on average the spin bit is
-disabled for at least one eighth of connections. The selection process performed
-at the beginning of the connection SHOULD be applied for all paths used by the
-connection.
+disabled for at least one eighth of network paths. The selection process
+performed at the beginning of the connection SHOULD be applied for all paths
+used by the connection.
 
-In case multiple connections share the same five-tuple, that is, have the same
-source and destination IP address and UDP ports, endpoints should try to
-co-ordinate across all connections to ensure a clear signal to any on-path
-measurement points.
+In case multiple connections share the same network path, as determined by
+having the same source and destination IP address and UDP ports, endpoints
+should try to co-ordinate across all connections to ensure a clear signal to any
+on-path measurement points.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -4019,14 +4019,14 @@ to disable the spin bit either globally or on a per-connection basis. Even when
...
 selection process SHOULD be designed such that on average the spin bit is
-disabled for at least one eighth of connections. The selection process performed
-at the beginning of the connection SHOULD be applied for all paths used by the
-connection.
-
-In case multiple connections share the same five-tuple, that is, have the same
-source and destination IP address and UDP ports, endpoints should try to
-co-ordinate across all connections to ensure a clear signal to any on-path
-measurement points.
+disabled for at least one eighth of network paths. The selection process
```
ianswett: What happens when the client thinks the path hasn't changed but the server has?  An observer could be on either side of the NAT.

mikkelfj: It doesn't really matter if the client and server disagrees in general as long as enough connections agree on enabling spin. As to privacy that is for justifying being silent so if one endpoint decides not to be, that shouldn't really matter either.

draft-ietf-quic-transport.md
```diff
@@ -4019,14 +4019,14 @@ to disable the spin bit either globally or on a per-connection basis. Even when
...
+performed at the beginning of the connection SHOULD be applied for all paths
+used by the connection.
+
+In case multiple connections share the same network path, as determined by
+having the same source and destination IP address and UDP ports, endpoints
+should try to co-ordinate across all connections to ensure a clear signal to any
+on-path measurement points.
```
ekr: This seems like a classic case of RFC 6919 language. Are any of the implementations which (1) do spin and (2) reuse host ports intending to do this?

martinthomson: I opened #2763 to explore that properly.

draft-ietf-quic-transport.md
martinthomson: Ian's question is somewhat relevant.  If the server changes their mind, then the spin state (spin, not) might change.  But that's OK.

PR #2752, opened 2019-05-22
ECN verification text

### LINKED ISSUE #2534: ECN text disables ECN too aggressively
Section 13.3.2 "ECN Verification"

> If an endpoint sets ECT codepoints on outgoing IP packets and encounters a retransmission timeout due to the absence of acknowledgments from the peer (see {{QUIC-RECOVERY}}), or if an endpoint has reason to believe that an element on the network path might be corrupting ECN codepoints, the endpoint MAY cease setting ECT codepoints...

Although I understand the intent here, retransmission timeouts, as a term, don't exist anymore in QUIC except during handshakes.

Disabling ECN after any PTO firing seems a bit extreme, so I'd like to see a better recommendation here, which covers both handshakes and short headers.

janaiyengar:
Cleans up ECN verification text.

Fixes #2534, #2205.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3114,8 +3114,8 @@ Handshake packet number space will be incremented by one and the counts for the
 ### ECN Verification {#ecn-verification}
 
 It is possible for faulty network devices on path to corrupt or erroneously drop
-all IP packets with ECN Capable Transport (ECT) codepoints set.  To provide
-robust connectivity in the presence of such devices, each endpoint independently
+all packets with ECN Capable Transport (ECT) codepoints set.  To provide robust
+connectivity in the presence of such devices, each endpoint independently
 verifies and enables use of ECN.  Even if not setting ECT codepoints on packets
 it transmits, the endpoint SHOULD provide feedback about ECN markings received
 if they are accessible.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3076,8 +3076,8 @@ rate in response, as described in {{QUIC-RECOVERY}}.
 To use ECN, QUIC endpoints first determine whether a path supports ECN marking
 and the peer is able to access the ECN codepoint in the IP header.  A network
 path does not support ECN if ECN marked packets get dropped or ECN markings are
-rewritten on the path. An endpoint verifies the path, both during connection
-establishment and when migrating to a new path (see {{migration}}).
+rewritten on the path. An endpoint validates the use of ECN on the path, both
+during connection establishment and when migrating to a new path ({{migration}}).
 
 
 ### ECN Counts
@@ -3111,38 +3111,57 @@ Handshake packet number space will be incremented by one and the counts for the
 1-RTT packet number space will be increased by two.
 
 
-### ECN Verification {#ecn-verification}
+### ECN Validation {#ecn-validation}
 
-It is possible for faulty network devices on path to corrupt or erroneously drop
-all IP packets with ECN Capable Transport (ECT) codepoints set.  To provide
-robust connectivity in the presence of such devices, each endpoint independently
-verifies and enables use of ECN.  Even if not setting ECT codepoints on packets
-it transmits, the endpoint SHOULD provide feedback about ECN markings received
-if they are accessible.
+It is possible for faulty network devices to corrupt or erroneously drop IP
+packets which have ECN markings.  To provide robust connectivity in the presence
+of such devices, each endpoint independently validates, and enables or disables
+ECN marking on packets it transmits.
 
-To verify both that a path supports ECN and that the peer can provide ECN
-feedback, an endpoint initially sets the ECT(0) codepoint in the IP header of
-all outgoing packets {{!RFC8311}}.
+ECN validation is local to a path, and MUST be independently performed for each
+path from an endpoint to its peer.  An endpoint thus independently validates ECN
+on new connection establishment, when switching to a new server preferred
+address, and on active connection migration to a new path.
 
-If all packets with ECT codepoints are dropped by a network device, endpoints
-can detect loss of the packets and MAY cease setting ECT codepoints in
-subsequent packets.  As one concrete strategy, an endpoint could disable setting
-of ECT codepoints if Initial packets with ECT codepoints set are deemed lost.
+Even if an endpoint does not use ECN markings on packets it transmits, the
+endpoint SHOULD provide feedback about ECN markings received from the peer if
+they are accessible. Not doing so will cause the peer to disable ECN marking.
 
-If a QUIC packet sent with an ECT codepoint is newly acknowledged by the peer in
-an ACK frame without ECN feedback, the endpoint stops setting ECT codepoints in
-subsequent IP packets, with the expectation that either the network path or the
-peer no longer supports ECN.
+#### Sending ECN Markings
 
-Network devices that corrupt or apply non-standard ECN markings might result in
-reduced throughput or other undesirable side-effects.  To reduce this risk, an
-endpoint uses the following steps to verify the counts it receives in an ACK
-frame.
+To start ECN validation, an endpoint SHOULD do the following when sending
+packets on a new path to a peer:
 
-* The total increase in ECT(0), ECT(1), and CE counts MUST be no smaller than
-  the total number of QUIC packets sent with an ECT codepoint that are newly
-  acknowledged in this ACK frame.  This step detects any network remarking from
-  ECT(0), ECT(1), or CE codepoints to Not-ECT.
+* Set the ECT(0) codepoint in the IP header of early outgoing packets sent on a
+  new path to the peer {{!RFC8311}}.
+
+* If all packets that were sent with the ECT(0) codepoint are eventually deemed
+  lost {{QUIC-RECOVERY}}, validation is deemed to have failed.
+
+To reduce the chances of misinterpreting congestive loss as packets dropped by a
+faulty network element, an endpoint could set the ECT(0) codepoint in the first
+ten outgoing packets on a path, or for a period of three RTTs, whichever occurs
+first.  Alternate strategies are possible.  For example, an endpoint could send
+the first ten packets interleaved: five ECT(0)-marked packets interleaved with
+five unmarked packets.  This allows the endpoint to more clearly identify
+congestive loss as such.  Implementations MAY experiment with and use other
+strategies.
+
+#### Receiving ACK Frames
+
+An endpoint that sets ECT(0) or ECT(1) codepoints on packets it transmits MUST
+use the following steps on receiving an ACK frame to validate ECN.
+
+* If this ACK frame newly acknowledges a packet that the endpoint sent with
+  either ECT(0) or ECT(1) codepoints set, and if no ECN feedback is present in
+  the ACK frame, validation fails.  This step protects against both a network
+  element that zeroes out ECN bits and a peer that is unable to access ECN
+  markings, since the peer could respond without ECN feedback in these cases.
+
+* For validation to succeed, the total increase in ECT(0), ECT(1), and CE counts
+  MUST be no smaller than the total number of QUIC packets sent with an ECT
+  codepoint that are newly acknowledged in this ACK frame.  This step detects
+  any network remarking from ECT(0), ECT(1), or CE codepoints to Not-ECT.
 
 * Any increase in either ECT(0) or ECT(1) counts, plus any increase in the CE
   count, MUST be no smaller than the number of packets sent with the
@@ -3150,29 +3169,30 @@ frame.
   This step detects any erroneous network remarking from ECT(0) to ECT(1) (or
   vice versa).
 
+Processing counts out of order can result in validation failure.  An endpoint
+SHOULD NOT perform this validation if the ACK frame is received in a packet with
+packet number lower than a previously received ACK frame.  Validating based on
+ACK frames that arrive out of order can result in disabling ECN unnecessarily.
+
 An endpoint could miss acknowledgements for a packet when ACK frames are lost.
 It is therefore possible for the total increase in ECT(0), ECT(1), and CE counts
 to be greater than the number of packets acknowledged in an ACK frame.  When
-this happens, and if verification succeeds, the local reference counts MUST be
+this happens, and if validation succeeds, the local reference counts MUST be
 increased to match the counts in the ACK frame.
 
-Processing counts out of order can result in verification failure.  An endpoint
-SHOULD NOT perform this verification if the ACK frame is received in a packet
-with packet number lower than a previously received ACK frame.  Verifying based
-on ACK frames that arrive out of order can result in disabling ECN
-unnecessarily.
+#### Validation Outcomes
 
-If verification fails, then the endpoint ceases setting ECT codepoints in
-subsequent IP packets with the expectation that either the network path or the
-peer does not support ECN.
+If validation fails, then the endpoint stops sending ECN markings in subsequent
+IP packets with the expectation that either the network path or the peer does
+not support ECN.
 
-If an ECT codepoint set is not dropped or corrupted by a network device, then a
-received packet contains either the codepoint sent by the peer or the Congestion
-Experienced (CE) codepoint set by a network device that is experiencing
-congestion.
+Upon successful validation, an endpoint can continue to set ECT codepoints in
+subsequent packets with the expectation that the path is ECN-capable.  Network
+routing and path elements can change mid-connection however; an endpoint MUST
+disable ECN if validation fails at any point in the connection.
 
-Upon successful verification, an endpoint continues to set ECT codepoints in
-subsequent packets with the expectation that the path is ECN-capable.
+Even if validation fails, an endpoint MAY re-validate ECN on the same path to
+the peer at any later time in the connection.
 
 
 # Packet Size {#packet-size}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3115,18 +3115,17 @@ Handshake packet number space will be incremented by one and the counts for the
 
 It is possible for faulty network devices to corrupt or erroneously drop packets
 with ECN markings.  To provide robust connectivity in the presence of such
-devices, each endpoint independently validates ECN counts and disables ECN
-if errors are detected.
-marking on packets it transmits.
+devices, each endpoint independently validates ECN counts and disables ECN if
+errors are detected.
 
-Endpoints validate ECN for packets sent on each network path independently.
-An endpoint thus validates ECN
-on new connection establishment, when switching to a new server preferred
-address, and on active connection migration to a new path.
+Endpoints validate ECN for packets sent on each network path independently.  An
+endpoint thus validates ECN on new connection establishment, when switching to a
+new server preferred address, and on active connection migration to a new path.
 
 Even if an endpoint does not use ECN markings on packets it transmits, the
-endpoint MUST provide feedback about ECN markings received from the peer if
-they are accessible. Failing to report ECN counts will cause the peer to disable ECN marking.
+endpoint MUST provide feedback about ECN markings received from the peer if they
+are accessible.  Failing to report ECN counts will cause the peer to disable ECN
+marking.
 
 #### Sending ECN Markings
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3053,19 +3053,21 @@ Handshake packet number space will be incremented by one and the counts for the
...
+if they are accessible.
+
+To verify both that a path supports ECN and that the peer can provide ECN
+feedback, an endpoint initially sets the ECT(0) codepoint in the IP header of
+all outgoing packets {{!RFC8311}}.
+
+If all packets with ECT codepoints are dropped by a network device, endpoints
+can detect loss of the packets and MAY cease setting ECT codepoints in
+subsequent packets.  As one concrete strategy, an endpoint could disable setting
+of ECT codepoints if Initial packets with ECT codepoints set are deemed lost.
```
ianswett: How many initial packets?  All, 1?

martinthomson: I think that it is going to need to be all.  But that leads to the question: do you let the handshake time out? Do you interleave ECT-marked and unmarked packets? Or do you start with ECT and switch the markings off after a few tries (and how many tries)?

Also, this needs to be all marked packets sent on a given path.  It is not just Initial packets, but also Handshake or 0-RTT and maybe those with short headers after a migration.

draft-ietf-quic-transport.md
```diff
@@ -3100,20 +3102,17 @@ with packet number lower than a previously received ACK frame.  Verifying based
...
 peer does not support ECN.
 
-If an endpoint sets ECT codepoints on outgoing IP packets and encounters a
-retransmission timeout due to the absence of acknowledgments from the peer (see
-{{QUIC-RECOVERY}}), or if an endpoint has reason to believe that an element on
-the network path might be corrupting ECN codepoints, the endpoint MAY cease
-setting ECT codepoints in subsequent packets.  Doing so allows the connection to
-be resilient to network elements that corrupt ECN codepoints in the IP header or
-drop packets with ECT or CE codepoints in the IP header.
+If an ECT codepoint set is not dropped or corrupted by a network device, then a
```
martinthomson: This paragraph doesn't really seem to say anything now.  You have deleted the timeout thing.  So it's not clear how you decide that a path *doesn't* support ECN now.

The point of having a timeout is to detect a black hole.  One nice thing about QUIC is that the idle timeout can be longer than any ECN validation timeout, so - at the risk of suggesting a design - maybe we could recommend that endpoints send the first packets it sends on any path marked, then they might stop marking until ECN support is confirmed.  We don't need to set a fixed rule, but it probably makes sense to recommend something.  The first N packets (where N of 10 produces pretty high confidence) or one round trip (which will be estimated based on the previous path, but that is OK) might work.  If you don't get a positive result, which might happen if the first few packets are lost for other reasons, we'll disable ECN more than is ideal, but we can allow the test to be retried.

(Maybe you could add this above the bullet list above, where I can't comment:
"A path is verified if ECT markings are either forwarded without change or the network changes the marking to Congestion Experience (ECN-CE)."
)

draft-ietf-quic-transport.md
```diff
@@ -3100,20 +3102,17 @@ with packet number lower than a previously received ACK frame.  Verifying based
...
+received packet contains either the codepoint sent by the peer or the Congestion
+Experienced (CE) codepoint set by a network device that is experiencing
+congestion.
+
+Upon successful verification, an endpoint continues to set ECT codepoints in
+subsequent packets with the expectation that the path is ECN-capable.
```
martinthomson: This is not a permanent condition.  If only one packet was marked, you might consider the path ECN-capable based on an acknowledgement of that packet.  However, that could be the result of random corruption.

I think that we can just say "until validation of ECN counts fails".  So we have the following algorithm:

For each path:

0. set ECN state to "testing"
1. after N packets or M round trips, set ECN state to "unknown"
2. on every ACK, check ECN counts
3. if ECN validation fails set ECN state to "failed".
4. if ECN validation passes and ECN state is "uncertain", set it to "capable".
5. when sending packets, mark with ECT(0) if state is "testing" or "capable"

(I know that this is unchanged text, but I just realized that we could be clearer.)

draft-ietf-quic-transport.md
gorryfair: "set" doesn't mean anything to me regarding this 2-bit code point, is this a non-zero codepoint? (i.e. not not-ECT)

draft-ietf-quic-transport.md
gloinul: Yes, to prevent spurious ECN disablement one should have a number of packets spread over time so that a single burst loss can't be the cause for the loss so that probability is low. So, the downside of the above algorithm is that to get low probability you may have several retransmission of the handshake. From an ECN perspective an implementation could mark a single packet per RTT and not give up testing until 10 RTTs has past. That would have a low probability for spurious disablement.

mirjak: Under the assumption that ECN-based black holing basically doesn't exists anymore or at at least was not measurable anymore, I would like to propose to enable ECN-marking from the beginning and simply disable ECN on time-out or if acks without ECN feedback are received. Of course ECN could be enabled anytime again if you want to try again, however, no matter if in "testing" or not if you don't receive ECN feedback (anymore) for whatever reason, you should stop using ECN no matter what.

janaiyengar:
I've taken in all feedback and reworked this section a fair bit. It's structured differently now, PTAL.

martinthomson:
The wall of commits is a little disconcerting there.

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3111,88 @@ Handshake packet number space will be incremented by one and the counts for the
...
 
-### ECN Verification {#ecn-verification}
+### ECN Validation {#ecn-validation}
 
-Each endpoint independently verifies and enables use of ECN by setting the IP
-header ECN codepoint to ECN Capable Transport (ECT) for the path from it to the
-other peer. Even if not setting ECN codepoints on packets it transmits, the
-endpoint SHOULD provide feedback about ECN markings received (if accessible).
+It is possible for faulty network devices to corrupt or erroneously drop packets
+which have ECN markings.  To provide robust connectivity in the presence of such
```
```suggestion
with ECN markings.  To provide robust connectivity in the presence of such
```
CMOS nit

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3111,88 @@ Handshake packet number space will be incremented by one and the counts for the
...
+devices, each endpoint independently validates, and enables or disables ECN
```
```suggestion
devices, each endpoint independently validates ECN counts and disables ECN
if errors are detected.
```

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3111,88 @@ Handshake packet number space will be incremented by one and the counts for the
...
+marking on packets it transmits.
 
-To verify both that a path supports ECN and the peer can provide ECN feedback,
-an endpoint sets the ECT(0) codepoint in the IP header of all outgoing
-packets {{!RFC8311}}.
+ECN validation is local to a path, and MUST be independently performed for each
```
```suggestion
Endpoints validate ECN for packets sent on each network path independently.
```

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3111,88 @@ Handshake packet number space will be incremented by one and the counts for the
...
+path from an endpoint to its peer.  An endpoint thus independently validates ECN
```
```suggestion
An endpoint thus validates ECN
```

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3111,88 @@ Handshake packet number space will be incremented by one and the counts for the
...
+on new connection establishment, when switching to a new server preferred
+address, and on active connection migration to a new path.
 
-If an ECT codepoint set in the IP header is not corrupted by a network device,
-then a received packet contains either the codepoint sent by the peer or the
-Congestion Experienced (CE) codepoint set by a network device that is
-experiencing congestion.
+Even if an endpoint does not use ECN markings on packets it transmits, the
+endpoint SHOULD provide feedback about ECN markings received from the peer if
```
martinthomson: No need to SHOULD here.

```suggestion
endpoint MUST provide feedback about ECN markings received from the peer if
```

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3111,88 @@ Handshake packet number space will be incremented by one and the counts for the
...
+they are accessible. Not doing so will cause the peer to disable ECN marking.
```
```suggestion
they are accessible. Failing to report ECN counts will cause the peer to disable ECN marking.
```

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3111,88 @@ Handshake packet number space will be incremented by one and the counts for the
...
+* If all packets that were sent with the ECT(0) codepoint are eventually deemed
+  lost {{QUIC-RECOVERY}}, validation is deemed to have failed.
+
+To reduce the chances of misinterpreting congestive loss as packets dropped by a
+faulty network element, an endpoint could set the ECT(0) codepoint in the first
+ten outgoing packets on a path, or for a period of three RTTs, whichever occurs
+first.  Alternate strategies are possible.  For example, an endpoint could send
+the first ten packets interleaved: five ECT(0)-marked packets interleaved with
+five unmarked packets.  This allows the endpoint to more clearly identify
+congestive loss as such.  Implementations MAY experiment with and use other
```
martinthomson: I'd drop the "For example" and "This allows the endpoint to more clearly identity" sentences.  Also the "Alternate strategies" as well.

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3111,88 @@ Handshake packet number space will be incremented by one and the counts for the
...
+strategies.
+
+#### Receiving ACK Frames
+
+An endpoint that sets ECT(0) or ECT(1) codepoints on packets it transmits MUST
```
martinthomson: We only mention ECT(0) up to here.  I think that we want to either mention ECT(1) in the previous section (under experimentation) or warm up to it somehow.

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3111,88 @@ Handshake packet number space will be incremented by one and the counts for the
...
+  codepoint that are newly acknowledged in this ACK frame.  This step detects
+  any network remarking from ECT(0), ECT(1), or CE codepoints to Not-ECT.
 
 * Any increase in either ECT(0) or ECT(1) counts, plus any increase in the CE
   count, MUST be no smaller than the number of packets sent with the
   corresponding ECT codepoint that are newly acknowledged in this ACK frame.
   This step detects any erroneous network remarking from ECT(0) to ECT(1) (or
   vice versa).
 
+Processing counts out of order can result in validation failure.  An endpoint
```
```suggestion
Processing ECN counts out of order can result in validation failure.  An endpoint
```

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3111,88 @@ Handshake packet number space will be incremented by one and the counts for the
...
+SHOULD NOT perform this validation if the ACK frame is received in a packet with
+packet number lower than a previously received ACK frame.  Validating based on
```
martinthomson: You still have the packet number thing that we talked about removing in favour of increased largest acknowledged...

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3111,88 @@ Handshake packet number space will be incremented by one and the counts for the
...
+IP packets with the expectation that either the network path or the peer does
+not support ECN.
+
+Upon successful validation, an endpoint can continue to set ECT codepoints in
+subsequent packets with the expectation that the path is ECN-capable.  Network
+routing and path elements can change mid-connection however; an endpoint MUST
+disable ECN if validation fails at any point in the connection.
+
+Even if validation fails, an endpoint MAY re-validate ECN on the same path to
+the peer at any later time in the connection.
```
mirjak: Maybe also add another sentence that it should revalidate on path change?

draft-ietf-quic-transport.md
mirjak: I would also like to see the easiest option spelled out: just enable ECN on all outgoing packet as dropping is rare and it can/will be disabled anytime when no ECN feedback is received.

draft-ietf-quic-transport.md
martinthomson: That would be confusing.  ECN validation is performed on a per-path basis.

mirjak: Okay, then you should say revalidation but maybe remark actually exactly this, that validation is performed on a per-path basis and has do be performed when a path change is detected.

draft-ietf-quic-transport.md
janaiyengar: Added some mentions of it above

draft-ietf-quic-transport.md
janaiyengar: That's already said above in the 2nd para of the "ECN validation" section. The point of this sentence here is that validation failure does not mean an endpoint can't try again for this path.

draft-ietf-quic-transport.md
janaiyengar: done, done, and done.

draft-ietf-quic-transport.md
janaiyengar: I think I have it now, PTAL.

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3111,87 @@ Handshake packet number space will be incremented by one and the counts for the
...
+* Set the ECT(0) codepoint in the IP header of early outgoing packets sent on a
+  new path to the peer {{!RFC8311}}.
+
+* If all packets that were sent with the ECT(0) codepoint are eventually deemed
+  lost {{QUIC-RECOVERY}}, validation is deemed to have failed.
+
+To reduce the chances of misinterpreting congestive loss as packets dropped by a
+faulty network element, an endpoint could set the ECT(0) codepoint in the first
+ten outgoing packets on a path, or for a period of three RTTs, whichever occurs
+first.  Implementations MAY experiment with and use other strategies.  An
```
martinthomson: I would make a new paragraph here that says this:

Implementations MAY experiment with and use other strategies for use of ECN.  Other methods of probing paths for ECN support are possible, as are different marking strategies, including those that use ECT(1).

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3111,87 @@ Handshake packet number space will be incremented by one and the counts for the
...
 
 * Any increase in either ECT(0) or ECT(1) counts, plus any increase in the CE
   count, MUST be no smaller than the number of packets sent with the
   corresponding ECT codepoint that are newly acknowledged in this ACK frame.
   This step detects any erroneous network remarking from ECT(0) to ECT(1) (or
   vice versa).
 
+Processing ECN counts out of order can result in validation failure.  An
+endpoint SHOULD NOT perform this validation if this ACK frame does not advance
+the largest packet number acknowledged in this connection.
```
martinthomson: Your framing makes me ask a question: Is this test connection-scoped, or just path-scoped?  I think that either works about as well as any other, but the path-scoped one might be more correct.

But then path-scoped is probably tricky to implement.  So let's not worry about that.

draft-ietf-quic-transport.md
martinthomson: Drop "to the peer".

I'm not happy about the hyphenated "re-validate", but I'm short on rephrasing suggestions, sorry.

draft-ietf-quic-transport.md
mirjak: I don't think normative language is needed here.

draft-ietf-quic-transport.md
```diff
@@ -3111,69 +3112,87 @@ Handshake packet number space will be incremented by one and the counts for the
...
+* If all packets that were sent with the ECT(0) codepoint are eventually deemed
+  lost {{QUIC-RECOVERY}}, validation is deemed to have failed.
+
+To reduce the chances of misinterpreting congestive loss as packets dropped by a
+faulty network element, an endpoint could set the ECT(0) codepoint in the first
+ten outgoing packets on a path, or for a period of three RTTs, whichever occurs
+first.  Implementations MAY experiment with and use other strategies.  An
+endpoint could send the first ten packets interleaved: five packets marked with
+ECT(0) or ECT(1) interleaved with five unmarked packets.  An endpoint could also
+simply send all packets marked with ECT(0) or ECT(1) until validation fails.
```
mirjak: Maybe say: "When the expected risk of ECN-based blackholing is low, an endpoint can also
	simply send all packets marked with ECT(0) until validation fails."
I would also like to point people to some measurement data, e.g.: "Recent studies have detected potential ECN-related connectivity problems in less than 0.3% [ECNmeasure]"

[ECNmeasure] Brian Trammell, Mirja Kühlewind, Piet De Vaere, Iain R. Learmonth, and Gorry Fairhurst. 2017. Tracking transport-layer evolution with PATHspider. In Proceedings of the Applied Networking Research Workshop (ANRW '17). ACM, New York, NY, USA, 20-26. DOI: https://doi.org/10.1145/3106328.3106336

draft-ietf-quic-transport.md
mirjak: RFC8311 does not define use of ECT(1) as interchangeable with ECT(1) anymore and therefore advises endpoints to only use ETC(0) which classic ECN. I think it's right to only talk about ECT(0) here, or if you want to mention that ECT(1) should also use for experimentation with new ECN based signal approaches, you have to point to RFC8311.

draft-ietf-quic-transport.md
martinthomson: A problem here is that the text mandates use of ECT(0), which might preclude use of something like L4S on a strict reading.  Also, we need a reason to mention ECT(1) later.

mirjak: RFC8311 says: "Protocols and senders MUST NOT use the ECT(1)
      codepoint to indicate ECT unless otherwise specified by an
      Experimental RFC in the IETF document stream."
So I don't think we should talk about use of ECT(1) without a pointer to RFC8311.

However, for this sentence I still don't see why normative language is needed.

And If think if we want to say something about ECT(1) we should rather say something like: "ECT(1) is reserved for experiments by RFC8311. If such an experiment is implemented, the described validation methods might apply as well to ETC(1)." However, that actually doesn't say much...

draft-ietf-quic-transport.md
janaiyengar: I'm not sure that we need to say anything here about the prevalence of ECN blackholing. No matter what the prevalence, this spec will have to specify methods that work when ECN blackholing is a problem. It seems to me that this text probably belongs in the applicability draft.

draft-ietf-quic-transport.md
janaiyengar: I think we only need to specify one method that can be used as a default, and suggest that implementations can use others too. I'm going with the suggestion of removing other methods entirely from this text because I don't think we need to discuss considerations of using other methods. If I'm in the minority, then I would suggest that we still keep it out of this PR, and add more methods in a separate PR.

draft-ietf-quic-transport.md
janaiyengar: I've cited 8311 for the text on sending ECT(1). This text is about what to do when receiving ECT(0) or ECT(1), and so I'm not citing 8311 here.

Unless you're suggesting that we scrub out all mentions of ECT(1) in the text here. I'm not sure that's necessary though, since the rules of validation do not change even when a future spec allows the use of ECT(1).

draft-ietf-quic-transport.md
janaiyengar: Good point. I tried to think of ways in which the path-level largest acked goes down while the connection-level largest acked goes up, but failed. I think connection-level might be a slightly stronger constraint than we want, but I _think_ it's actually correct.

draft-ietf-quic-transport.md
janaiyengar: Turns out I don't need the hyphenation.

janaiyengar:
I think this is ready. Unless someone disagrees, I'll mark the corresponding issue as proposal-ready tomorrow.

martinthomson:
I don't think that we'll find new objections.  I marked it myself.

draft-ietf-quic-transport.md
mirjak: Thanks! That's fine now. Did just reply to Martin's mail. I agree that here it should mention both codepoints.

draft-ietf-quic-transport.md
ianswett: Connection scoped works for me, at least for now.  When we have a way to ACK packets on different paths at the same time, we can revisit this question.

PR #2756, opened 2019-05-23
Remove AEAD_AES_256_CCM

DavidSchinazi:
draft-ietf-quic-tls currently mentions AEAD_AES_256_CCM, however there is no way to use that AEAD in TLS 1.3 - TLS_AES_256_CCM_SHA256 does not appear in the [list of TLS cipher suites](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4). If I understand this correctly, the presence of AEAD_AES_256_CCM in the document was an editorial oversight.


martinthomson:
Thanks.  This was extremely confusing: you meant CCM, but wrote GCM in tons of places.

PR #2759, opened 2019-05-29
QPACK editorial revision

afrind:
I made a major editorial pass on the QPACK spec, hoping to address some of the issues mentioned in #2339, and others.

I separated logical changes into individual commits in this PR, which should (hopefully) make it a bit easier to review the changes.  For example, some sections were reordered and updated, so there's a commit that just moves the text without changes.  

The "Editorial Clarifications" commit should not impact design at all, but does add new text explaining the handling of certain cases.

In addition to looking at the individual changes, giving the final product a read from start to finish would also be appreciated.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -296,9 +296,13 @@ stream becoming blocked, the encoder tracks the number of entries received by
 the decoder.  The Known Received Count tracks the total number of acknowledged
 insertions.
 
-When blocking references are permitted, the encoder uses header block
-acknowledgement to maintain the Known Received Count, as described in
-{{header-acknowledgement}}.
+When blocking references are permitted, the encoder uses Header Acknowledgement
+instructions ({{header-acknowledgement}}) to maintain the Known Received
+Count.  If a header block was potentially blocking, the acknowledgement implies
+that the decoder has received all dynamic table state necessary to process the
+header block.  If the Required Insert Count of an acknowledged header block was
+greater than the encoder's current Known Received Count, the block's Required
+Insert Count becomes the new Known Received Count.
 
 To acknowledge dynamic table entries which are not referenced by header blocks,
 for example because the encoder or the decoder have chosen not to risk blocked
@@ -340,20 +344,41 @@ MUST treat this as a connection error of type HTTP_QPACK_DECOMPRESSION_FAILED.
 
 ### State Synchronization
 
-The decoder instructions ({{decoder-instructions}}) signal key events at the
-decoder that permit the encoder to track the decoder's state.  These events are:
+The decoder signals key events by emitting decoder instructions
+({{decoder-instructions}}) on the decoder stream.
 
-- Complete processing of a header block
-- Abandonment of a stream which might have remaining header blocks
-- Receipt of new dynamic table entries
+#### Completed Processing of a Header Block
 
-Knowledge that a header block with references to the dynamic table has been
-processed permits the encoder to evict entries to which no unacknowledged
-references remain (see {{blocked-insertion}}).  When a stream is reset or
-abandoned, the indication that these header blocks will never be processed
-serves a similar function (see {{stream-cancellation}}).
+When the decoder finishes decoding a header block containing dynamic table
+references, it emits a Header Acknowledgement instruction
+({{header-acknowledgement}}).  The same Stream ID can be identified by multiple
+Header Acknowledgement instructions, as multiple header blocks can be sent on a
+single stream in the case of intermediate responses, trailers, and pushed
+requests.  Since frames carrying header blocks on each stream are received and
+processed in order, this gives the encoder precise feedback on which header
+blocks within a stream have been fully processed.
 
-The decoder chooses when to emit Insert Count Increment instructions (see
+#### Abandonment of a Stream
+
+When an endpoint receives a stream reset before the end of a stream or before
+all header blocks are processed on that stream, or when it abandons reading of a
+stream, it generates a Stream Cancellation instruction (see
+{{stream-cancellation}}).  This signals to the encoder that all references to
+the dynamic table on that stream are no longer outstanding.  A decoder with a
+maximum dynamic table capacity equal to zero (see
+{{maximum-dynamic-table-capacity}}) MAY omit sending Stream Cancellations,
+because the encoder cannot have any dynamic table references.  An encoder cannot
+infer from this instruction that any updates to the dynamic table have been
+received.
+
+The Header Acknowledgement and Stream Cancellation instructions permit the
+encoder to evict entries to which no unacknowledged references remain (see
+{{blocked-insertion}}).
+
+#### New Table Entries
+
+After receiving new table entries on the encoder stream, the decoder chooses
+when to emit Insert Count Increment instructions (see
 {{insert-count-increment}}). Emitting an instruction after adding each new
 dynamic table entry will provide the timeliest feedback to the encoder, but
 could be redundant with other decoder feedback. By delaying an Insert Count
@@ -771,24 +796,10 @@ possibly update the Known Received Count.
 ~~~~~~~~~~
 {:#fig-header-ack title="Header Acknowledgement"}
 
-The same Stream ID can be identified multiple times, as multiple header blocks
-can be sent on a single stream in the case of intermediate responses, trailers,
-and pushed requests.  Since frames carrying header blocks on each stream are
-received and processed in order, this gives the encoder precise feedback on
-which header blocks within a stream have been fully processed.
-
 If an encoder receives a Header Acknowledgement instruction referring to a
 stream on which every header block with a non-zero Required Insert Count has
 already been acknowledged, that MUST be treated as a connection error of type
 `HTTP_QPACK_DECODER_STREAM_ERROR`.
-
-When blocking references are permitted, the encoder uses acknowledgement of
-header blocks to update the Known Received Count.  If a header block was
-potentially blocking, the acknowledgement implies that the decoder has received
-all dynamic table state necessary to process the header block.  If the Required
-Insert Count of an acknowledged header block was greater than the encoder's
-current Known Received Count, the block's Required Insert Count becomes the new
-Known Received Count.
 
 
 ### Stream Cancellation
@@ -803,19 +814,6 @@ the stream ID of the affected stream encoded as a 6-bit prefix integer.
 +---+---+-----------------------+
 ~~~~~~~~~~
 {:#fig-stream-cancel title="Stream Cancellation"}
-
-A stream that is reset might have multiple outstanding header blocks with
-dynamic table references.  When an endpoint receives a stream reset before the
-end of a stream or before all header blocks are processed on that stream, or
-when it abandons reading of a stream, it generates a Stream Cancellation
-instruction.  This signals to the encoder that all references to the dynamic
-table on that stream are no longer outstanding.  A decoder with a maximum
-dynamic table capacity equal to zero (see {{maximum-dynamic-table-capacity}})
-MAY omit sending Stream Cancellations, because the encoder cannot have any
-dynamic table references.
-
-An encoder cannot infer from this instruction that any updates to the dynamic
-table have been received.
 
 ### Insert Count Increment
 
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -82,21 +82,17 @@ code and issues list for this draft can be found at
 
 # Introduction
 
-The QUIC transport protocol is designed to support HTTP semantics, and its
-design subsumes many of the features of HTTP/2.  HTTP/2 uses HPACK
-({{!RFC7541}}) for header compression.  If HPACK were used for HTTP/3, it would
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -167,15 +167,8 @@ fields (some of them with an empty value).  The dynamic table (see
...
-
-- The encoder uses a unidirectional stream to modify the state of the dynamic
-table without emitting header fields associated with any particular request.
-
-- Header blocks reference the table state without modifying it.
-
-- The decoder sends feedback to the encoder on a unidirectional stream.  This
-feedback enables the encoder to manage dynamic table state.
+QPACK defines unidirectional streams for sending instructions from encoder to
+decoder and vice-versa.
```
bencebeky: s/vice-versa/vice versa/ as per https://www.grammarly.com/blog/vice-versa/

draft-ietf-quic-qpack.md
```diff
@@ -82,25 +82,22 @@ code and issues list for this draft can be found at
...
-comes into some conflict with HPACK.  A key goal of the design of QUIC is to
-improve stream multiplexing relative to HTTP/2 by reducing head-of-line
-blocking.  If HPACK were used for HTTP/3, it would induce head-of-line
-blocking due to built-in assumptions of a total ordering across frames on all
-streams.
-
-QUIC is described in {{QUIC-TRANSPORT}}.  The HTTP/3 mapping is described in
-{{HTTP3}}. For a full description of HTTP/2, see {{?RFC7540}}. The
-description of HPACK is {{!RFC7541}}.
+The QUIC transport protocol is designed to support HTTP semantics, and its
```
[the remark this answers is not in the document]
martinthomson: A reference to h3 might help more.

draft-ietf-quic-qpack.md
```diff
@@ -327,37 +312,86 @@ instructions received on the encoder stream.
...
+requests.  Since frames carrying header blocks on each stream are received and
+processed in order, this gives the encoder precise feedback on which header
+blocks within a stream have been fully processed.
+
+#### Abandonment of a Stream
+
+When an endpoint receives a stream reset before the end of a stream or before
+all header blocks are processed on that stream, or when it abandons reading of a
+stream, it generates a Stream Cancellation instruction (see
+{{stream-cancellation}}).  This signals to the encoder that all references to
```
LPardue: is this all unacknowledged references?

LPardue: Oh is that what the following paragraph is trying to articulate?

draft-ietf-quic-qpack.md
```diff
@@ -327,37 +312,86 @@ instructions received on the encoder stream.
...
-processed permits the encoder to evict entries to which no unacknowledged
-references remain (see {{blocked-insertion}}).  When a stream is reset or
-abandoned, the indication that these header blocks will never be processed
-serves a similar function (see {{stream-cancellation}}).
+#### New Table Entries
 
-The decoder chooses when to emit Insert Count Increment instructions (see
+After receiving new table entries on the encoder stream, the decoder chooses
+when to emit Insert Count Increment instructions (see
 {{insert-count-increment}}). Emitting an instruction after adding each new
```
LPardue: should it be "this instruction" or could it be *any* instruction type.

draft-ietf-quic-qpack.md
```diff
@@ -629,23 +634,53 @@ HTTP_CLOSED_CRITICAL_STREAM.
...
 
-An addition to the header table where the header field name matches the header
-field name of an entry stored in the static table or the dynamic table starts
-with the '1' one-bit pattern.  The `S` bit indicates whether the reference is to
-the static (S=1) or dynamic (S=0) table. The 6-bit prefix integer (see Section
-5.1 of [RFC7541]) that follows is used to locate the table entry for the header
-name.  When S=1, the number represents the static table index; when S=0, the
-number is the relative index of the entry in the dynamic table.
+An encoder adds an entry to the dynamic table where the header field name
+matches the header field name of an entry stored in the static table or the
```
LPardue:
```suggestion
matches the header field name of an entry stored in the static or the
```

you omit this later anyway.

draft-ietf-quic-qpack.md
```diff
@@ -170,23 +167,18 @@ fields (some of them with an empty value).  The dynamic table (see
...
+decoder and vice-versa.
 
 ## Encoder
 
-An encoder compresses a header list by emitting either an indexed or a literal
-representation for each header field in the list.  References to the static
-table and literal representations do not require any dynamic state and never
-risk head-of-line blocking.  References to the dynamic table risk head-of-line
+An encoder compresses a header list into a header block by emitting either an
+indexed or a literal representation for each header field in the list (see
```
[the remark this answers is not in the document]
MikeBishop: "encodes," "represents"?  But regardless, this isn't new text in this change.

draft-ietf-quic-qpack.md
```diff
@@ -215,12 +218,15 @@ been referenced can still be blocking.
...
-  available.  Any encoder that uses the dynamic table has to keep track of
-  blocked entries, whereas blocked streams are optional.
+  available.  An encoder that uses the dynamic table has to keep track of
+  blocked entries.
 
 An encoder MUST NOT insert an entry into the dynamic table (or duplicate an
-existing entry) if doing so would evict a blocking entry.  In this case, the
-encoder can send literal representations of header fields.
+existing entry) if doing so would evict a blocking entry.  An encoder can always
+send literal representations of header fields when the dynamic table is full
```
MikeBishop: Perhaps just omit "when the dynamic table..."?  It's true universally, regardless of the dynamic table state.

draft-ietf-quic-qpack.md
MikeBishop: This instruction, you're right.

draft-ietf-quic-qpack.md
```diff
@@ -327,37 +312,86 @@ instructions received on the encoder stream.
...
+blocked.
+
+While blocked, header block data SHOULD remain in the blocked stream's flow
+control window.  A stream becomes unblocked when the Insert Count becomes
+greater than or equal to the Required Insert Count for all header blocks the
+decoder has started reading from the stream.
+<!-- doesn't the stream become unblocked when the encoder receives the acks? -->
+
+If the decoder encounters a header block with a Required Insert Count value
+larger than defined in {{blocked-streams}}, it MAY treat this as a connection
```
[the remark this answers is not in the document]
MikeBishop: The intent is to say that, when you actually process the block, if there was no reference to the entry the RIC said you needed, then the peer's encoder is broken.  Blow up.

draft-ietf-quic-qpack.md
```diff
@@ -327,37 +312,86 @@ instructions received on the encoder stream.
...
+error of type HTTP_QPACK_DECOMPRESSION_FAILED.  If the decoder encounters a
+header block with a Required Insert Count value smaller than defined in
+{{blocked-streams}}, it MUST treat this as a connection error of type
+HTTP_QPACK_DECOMPRESSION_FAILED as prescribed in {{invalid-references}}.
```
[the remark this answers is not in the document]
MikeBishop: Yes -- the general error case is that the RIC didn't match the actual contents of the header block; deviation in any direction is an error, even if you were in fact able to decode all the headers.

draft-ietf-quic-qpack.md
```diff
@@ -327,37 +312,86 @@ instructions received on the encoder stream.
...
+#### Completed Processing of a Header Block
+
+When the decoder finishes decoding a header block containing dynamic table
+references, it emits a Header Acknowledgement instruction
+({{header-acknowledgement}}).  The same Stream ID can be identified by multiple
+Header Acknowledgement instructions, as multiple header blocks can be sent on a
+single stream in the case of intermediate responses, trailers, and pushed
+requests.  Since frames carrying header blocks on each stream are received and
+processed in order, this gives the encoder precise feedback on which header
+blocks within a stream have been fully processed.
```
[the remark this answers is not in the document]
MikeBishop: Minor quibble with @LPardue's text:  That seems to imply that header blocks on the stream might *not* be completely processed in order.  Now, that's certainly possible to do (albeit not useful in most circumstances and not directly discussed in this doc), but acknowledgements always have to be in order here.

draft-ietf-quic-qpack.md
```diff
@@ -340,20 +344,41 @@ MUST treat this as a connection error of type HTTP_QPACK_DECOMPRESSION_FAILED.
...
+the dynamic table on that stream are no longer outstanding.  A decoder with a
+maximum dynamic table capacity equal to zero (see
+{{maximum-dynamic-table-capacity}}) MAY omit sending Stream Cancellations,
+because the encoder cannot have any dynamic table references.  An encoder cannot
+infer from this instruction that any updates to the dynamic table have been
+received.
+
+The Header Acknowledgement and Stream Cancellation instructions permit the
+encoder to evict entries to which no unacknowledged references remain (see
+{{blocked-insertion}}).
```
MikeBishop: This wording feels weird.  Entries without unacknowledged references can always be evicted.  The instructions acknowledge the references themselves.

draft-ietf-quic-qpack.md
```diff
@@ -795,8 +797,10 @@ already been acknowledged, that MUST be treated as a connection error of type
 
 ### Stream Cancellation
 
-The instruction begins with the '01' two-bit pattern. The instruction includes
-the stream ID of the affected stream encoded as a 6-bit prefix integer.
+When a stream is reset, the decoder emits a Stream Cancellation instruction. The
```
MikeBishop: ...or reading is abandoned (i.e. STOP_SENDING)

draft-ietf-quic-qpack.md
```diff
@@ -1035,17 +1038,19 @@ values that are not to be put at risk by compressing them (see Section 7.1 of
...
 {: title="Literal Header Field With Name Reference"}
 
-For entries in the static table or in the dynamic table with an absolute index
-less than the Base, the header field name is represented using the relative
-index of that entry, which is represented as an integer with a 4-bit prefix (see
-Section 5.1 of [RFC7541]). The `S` bit indicates whether the reference is to the
-static (S=1) or dynamic (S=0) table.
+The fourth (`S`) bit indicates whether the reference is to the static (S=1) or
+dynamic (S=0) table.  Finally, the header field name is represented using the
+static or relative index of that entry, which is represented as an integer with
```
MikeBishop: We define relative, absolute, and post-base indices.  We talk about "static table index" and "index into the static table".  "Static table index" is used on line 656 of the current doc.  This use of "static index" is a new variation.  I suspect a reader could reasonably infer that they all mean the same thing -- obviously we all have -- but it might be nice to align the terminology and make sure it's defined somewhere.

draft-ietf-quic-qpack.md
```diff
@@ -344,7 +337,7 @@ MUST treat this as a connection error of type HTTP_QPACK_DECOMPRESSION_FAILED.
 
 ### State Synchronization
 
-The decoder signals key events by emitting decoder instructions
+The decoder signals the following key events by emitting decoder instructions
```
MikeBishop: You can probably just replace "key" instead of adding.

draft-ietf-quic-qpack.md
MikeBishop: Maybe, but giving them a hint seems reasonable and generous.

draft-ietf-quic-qpack.md
MikeBishop: I think it's sufficient to just call them a relative index; the section defining that term describes relative to what in different contexts.

draft-ietf-quic-qpack.md
```diff
@@ -316,40 +311,39 @@ the input header block.
...
-error of type HTTP_QPACK_DECOMPRESSION_FAILED.  If the decoder encounters a
-header block with a Required Insert Count value smaller than defined in
-{{blocked-streams}}, it MUST treat this as a connection error of type
-HTTP_QPACK_DECOMPRESSION_FAILED as prescribed in {{invalid-references}}.
+When processing header blocks, the decoder expects the Required Insert Count
+value to exactly match the value defined in {{blocked-streams}. If it encounters
+a smaller value, it MUST treat this as a connection error of type
+HTTP_QPACK_DECOMPRESSION_FAILED (see {{invalid-references}}). If it encounters a
+larger value, it MAY treat this as a connection error of type
+HTTP_QPACK_DECOMPRESSION_FAILED.
```
MikeBishop: I worry slightly that people might get confused about which "value" is smaller or larger than which other "value", because that term is used to describe both.

draft-ietf-quic-qpack.md
afrind: This is covered in the "Encoder" section: 

> An encoder converts a header list into a header block by emitting either an indexed or a literal representation for each header field in the list
...
References to the static table and literal representations do not require any dynamic state and never risk head-of-line blocking

I guess it doesn't specifically state the encoder "can always send" these representations, but I'm going for less=more here.  Maybe we can add it later, if really needed.  I don't think it belongs in Avoiding Blocked Insertions, since optimistic insertions could also be blocked, and have nothing to do with literal representations.

PR #2769, opened 2019-06-04
Request to Retire Locally Issued CIDs

### LINKED ISSUE #2645: Retire My Own CID
I think the protocol should allow for the originator of a CID to retire it and once that packet has been acknowledged by the peer, remove it from its lookup table.

I have a scenario where I want to move a connection from one "partition" to another "partition" on the same machine. I use some bits from the CID to indicate the partition ID. My problem is, I've already sent out a CID to the peer indicating the old partition ID. If I change, I'm going to get a cross-partition lookup perf-hit for the rest of time on this connection.

So, I'd like to be able to pick a new partition ID, generate a new CID and send out a NEW_CONN_ID and a RETIRE_CONN_ID for the old CID. Once the peer acknowledges this, I remove the old CID from my look up.

With this design, if I need to change partitions for any reason, I only pay a perf penalty for 1 or 2 RTT.

Another option instead of supporting RETIRE for my own, could be a REPLACE_CONN_ID [orignal_sequence_number]. That might be a little less disruptive for implementations to manage. Just copy over the old one with the new one and go about your business.

nibanks:
Adds a new (optional) field, 'Retire Prior To' (placeholder name), to the NEW_CONNECTION_ID frame, which allows a sender to request that its peer immediately retire all connection IDs up to a certain sequence number. Once the peer acknowledges this packet, it has at most 3 PTO to follow through or risk the sender closing the connection.

Fixes #2645.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1001,13 +1001,11 @@ corresponding RETIRE_CONNECTION_ID frames in a timely manner.  Failing to do so
 can cause packets to be delayed, lost, or cause the original endpoint to send a
 stateless reset in response to a connection ID it can no longer route correctly.
 
-The sender of the Retire Prior To field MUST keep track of the connection IDs it
-wishes to retire until it has received a corresponding RETIRE_CONNECTION_ID
-frame, with one exception: if the sender of the Retire Prior To field has used
-distinct stateless reset tokens for all of their issued connection IDs, and 3
-times the PTO has elapsed since it received an acknowledgment for its Retire
-Prior To field, then it MAY drop state for that connection ID, and respond to
-packets with that connection ID with the corresponding stateless reset token.
+An endpoint MAY discard a connection ID for which retirement has been requested
+once an interval of no less than 3 PTO has elapsed since an acknowledgement is
+received for the NEW_CONNECTION_ID frame requesting that retirement.  Subsequent
+incoming packets using that connection ID will elicit a response with the
+corresponding stateless reset token.
 
 
 ## Matching Packets to Connections {#packet-handling}
@@ -5003,7 +5001,7 @@ Sequence Number:
 
 Retire Prior To:
 
-: A variable-length integer specifying the limit, for which all connection IDs
+: A variable-length integer specifying the limit for which all connection IDs
   assigned to the receiver that have a sequence number less than this value
   should be retired.  See {{retiring-cids}}.
 
@@ -5057,11 +5055,10 @@ The Retire Prior To field MUST be less than or equal to the Sequence Number
 field.  Values greater than the Sequence Number MUST be treated as a connection
 error of type PROTOCOL_VIOLATION.
 
-The sender cannot reduce the value of the Retire Prior To field in subsequent
-NEW_CONNECTION_ID frames.  That is, once a sender indicates a Retire Prior To
-value, it MAY indicate a smaller value, but it has no effect.  A receiver MUST
-ignore any Retire Prior To fields that do not increase.
-
+Once a sender indicates a Retire Prior To value, smaller values sent in
+subsequent NEW_CONNECTION_ID frames have no effect. A receiver MUST ignore any
+Retire Prior To fields that do not increase the largest received Retire Prior To
+value.
 
 
 ## RETIRE_CONNECTION_ID Frame {#frame-retire-connection-id}
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -994,13 +994,21 @@ packets sent from only one local address.  An endpoint that migrates away from a
 local address SHOULD retire all connection IDs used on that address once it no
 longer plans to use that address.
 
-The endpoint can explicitly require its peer to retire connection IDs by sending
-a NEW_CONNECTION_ID frame with an increased Retire Prior To field.  The peer is
-required to retire these connection IDs in a timely manner.  If the peer does
-not retire the connection IDs within 3 PTO of acknowledging the packet with the
-NEW_CONNECTION_ID frame, the endpoint MAY remove the connection IDs anyways.  If
-the peer continues to use the connection IDs they will likely receive a
-stateless reset.
+An endpoint can request that its peer retire connection IDs by sending a
+NEW_CONNECTION_ID frame with an increased Retire Prior To field.  Upon receipt,
+the peer SHOULD retire the corresponding connection IDs and send the
+corresponding RETIRE_CONNECTION_ID frame in a timely manner.  Failing to do so
+can cause packets to be delayed or lost and harm connection performance as the
+original endpoint might not route those connection IDs optimally after some
+delay.
+
+The sender of the Retire Prior To field MUST keep track of the connection IDs it
+wishes to retire until it has received a corresponding RETIRE_CONNECTION_ID
+frame, with one exception: if the sender of the Retire Prior To field has used
+distinct stateless reset tokens for all of their issued connection IDs, and 3
+times the PTO has elapsed since it received an acknowledgment for its Retire
+Prior To field, then it MAY lose track of that connection ID, and respond to
+packets with that connection ID with the corresponding stateless reset token.
 
 
 ## Matching Packets to Connections {#packet-handling}
@@ -5034,10 +5042,10 @@ sequence number, or if a sequence number is used for different connection
 IDs, the endpoint MAY treat that receipt as a connection error of type
 PROTOCOL_VIOLATION.
 
-The Retire Prior To field is a requirement for the peer to retire all connection
-IDs with a sequence number less than the specified value.  This includes the
-initial and preferred_address transport parameter connection IDs.  The peer
-SHOULD immediately retire all the connection IDs.
+The Retire Prior To field is a request for the peer to retire all connection IDs
+with a sequence number less than the specified value.  This includes the initial
+and preferred_address transport parameter connection IDs.  The peer SHOULD
+immediately retire all the connection IDs.
 
 The sender of the NEW_CONNECTION_ID frame MAY remove the connection IDs after 3
 PTO, even if the peer has not retired them with RETIRE_CONNECTION_ID yet.  The
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -994,12 +994,11 @@ packets sent from only one local address.  An endpoint that migrates away from a
 local address SHOULD retire all connection IDs used on that address once it no
 longer plans to use that address.
 
-The endpoint can explicitly request its peer to retire connection IDs by sending
+The endpoint can explicitly require its peer to retire connection IDs by sending
 a NEW_CONNECTION_ID frame with an increased Retire Prior To field.  The peer is
-required to retire these connection IDs in a timely manner.  On receipt of the
-acknowledgement for the packet that contained the NEW_CONNECTION_ID frame, the
-endpoint MAY start a 3 PTO timer.  If the timer expires before all the requested
-connection IDs are retired, the endpoint MAY close the connection with a
+required to retire these connection IDs in a timely manner.  If the peer does
+not retire the connection IDs within 3 PTO of acknowledging the packet with the
+NEW_CONNECTION_ID frame, the endpoint MAY close the connection with a
 PROTOCOL_VIOLATION.
 
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -5040,7 +5040,7 @@ initial and preferred_address transport parameter connection IDs.  The peer
 SHOULD immediately retire all the connection IDs.
 
 The sender of the NEW_CONNECTION_ID frame MAY remove the connection IDs after 3
-PTO, even if the peer has not retired them yet.  The 3 PTO timer starts on
+PTO, even if the peer has not retired them yet.  The 3-PTO timer starts on
 acknowledgement of the packet containing the NEW_CONNECTION_ID frame.  Continued
 use of the retired connection IDs after this point will likely result in a
 stateless reset being sent.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -5037,7 +5037,7 @@ PROTOCOL_VIOLATION.
 The Retire Prior To field is a request for the peer to immediately retire all
 connection IDs with a sequence number less than the specified value.  The Retire
 Prior To field MUST be less than the Sequence Number field.  Values greater than
-or equal to the Sequence Number MUST be treated as a connection error of type
+the Sequence Number MUST be treated as a connection error of type
 PROTOCOL_VIOLATION.
 
 The sender cannot reduce the value of the Retire Prior To field in subsequent
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -5035,9 +5035,10 @@ IDs, the endpoint MAY treat that receipt as a connection error of type
 PROTOCOL_VIOLATION.
 
 The Retire Prior To field is a request for the peer to immediately retire all
-connection IDs with a sequence number less than the specified value.  The Retire
-Prior To field MUST be less than the Sequence Number field.  Values greater than
-the Sequence Number MUST be treated as a connection error of type
+connection IDs with a sequence number less than the specified value.  This
+includes the initial and preferred_address transport parameter connection IDs.
+The Retire Prior To field MUST be less than the Sequence Number field.  Values
+greater than the Sequence Number MUST be treated as a connection error of type
 PROTOCOL_VIOLATION.
 
 The sender cannot reduce the value of the Retire Prior To field in subsequent
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -998,8 +998,9 @@ The endpoint can explicitly require its peer to retire connection IDs by sending
 a NEW_CONNECTION_ID frame with an increased Retire Prior To field.  The peer is
 required to retire these connection IDs in a timely manner.  If the peer does
 not retire the connection IDs within 3 PTO of acknowledging the packet with the
-NEW_CONNECTION_ID frame, the endpoint MAY close the connection with a
-PROTOCOL_VIOLATION.
+NEW_CONNECTION_ID frame, the endpoint MAY remove the connection IDs anyways.  If
+the peer continues to use the connection IDs they will likely receive a
+stateless reset.
 
 
 ## Matching Packets to Connections {#packet-handling}
@@ -5036,10 +5037,17 @@ PROTOCOL_VIOLATION.
 The Retire Prior To field is a requirement for the peer to retire all connection
 IDs with a sequence number less than the specified value.  This includes the
 initial and preferred_address transport parameter connection IDs.  The peer
-SHOULD immediately retire all the connection IDs.  The sender of the
-NEW_CONNECTION_ID frame MAY close the connection with an error of type
-PROTOCOL_VIOLATION if the peer does not retire the connection IDs within 3 PTO
-of the packet containing the NEW_CONNECTION_ID frame being acknowledged.
+SHOULD immediately retire all the connection IDs.
+
+The sender of the NEW_CONNECTION_ID frame MAY remove the connection IDs after 3
+PTO, even if the peer has not retired them yet.  The 3 PTO timer starts on
+acknowledgement of the packet containing the NEW_CONNECTION_ID frame.  Continued
+use of the retired connection IDs after this point will likely result in a
+stateless reset being sent.
+
+In order to prevent a stateless reset, sent in response to a heavily delayed
+packet using a retired connection ID, from closing the connection, the
+stateless reset token of the retired connection IDs SHOULD NOT be reused.
 
 The Retire Prior To field MUST be less than or equal to the Sequence Number
 field.  Values greater than the Sequence Number MUST be treated as a connection
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -5040,14 +5040,18 @@ initial and preferred_address transport parameter connection IDs.  The peer
 SHOULD immediately retire all the connection IDs.
 
 The sender of the NEW_CONNECTION_ID frame MAY remove the connection IDs after 3
-PTO, even if the peer has not retired them yet.  The 3-PTO timer starts on
-acknowledgement of the packet containing the NEW_CONNECTION_ID frame.  Continued
-use of the retired connection IDs after this point will likely result in a
-stateless reset being sent.
+PTO, even if the peer has not retired them with RETIRE_CONNECTION_ID yet.  The
+3-PTO timer starts on acknowledgement of the packet containing the
+NEW_CONNECTION_ID frame.  Continued use of the retired connection IDs after this
+point will likely result in a stateless reset being sent.
 
-In order to prevent a stateless reset sent in response to a heavily delayed
-packet using a retired connection ID from closing the connection, the
-stateless reset token of the retired connection IDs SHOULD NOT be reused.
+If a stateless reset is received in response to a retired connection ID, the
+reset SHOULD be ignored and the receiver SHOULD immediately switch to a
+connection ID that has not been retired if it has not done so already.
+
+In order to prevent a heavily delayed packet with a retired connection ID from
+eliciting a stateless reset that closes the connection, the stateless reset
+token of the retired connection IDs SHOULD NOT be reused.
 
 The Retire Prior To field MUST be less than or equal to the Sequence Number
 field.  Values greater than the Sequence Number MUST be treated as a connection
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -994,6 +994,14 @@ packets sent from only one local address.  An endpoint that migrates away from a
 local address SHOULD retire all connection IDs used on that address once it no
 longer plans to use that address.
 
+The endpoint can explicitly request its peer to retire connection IDs by sending
+a NEW_CONNECTION_ID frame that contains a Retire Prior To field.  The peer is
+required to retire these connection IDs in a timely manner.  On receipt of the
+acknowledgement for the packet that contained the NEW_CONNECTION_ID frame, the
+endpoint may start a 3 PTO timer, which on expiring may close the connection
+with a PROTOCOL_VIOLATION error if all the connection IDs were not retired as
+requested.
+
 
 ## Matching Packets to Connections {#packet-handling}
 
@@ -2776,11 +2784,11 @@ frames are explained in more detail in {{frame-formats}}.
 | 0x14        | DATA_BLOCKED         | {{frame-data-blocked}}         |
 | 0x15        | STREAM_DATA_BLOCKED  | {{frame-stream-data-blocked}}  |
 | 0x16 - 0x17 | STREAMS_BLOCKED      | {{frame-streams-blocked}}      |
-| 0x18        | NEW_CONNECTION_ID    | {{frame-new-connection-id}}    |
-| 0x19        | RETIRE_CONNECTION_ID | {{frame-retire-connection-id}} |
-| 0x1a        | PATH_CHALLENGE       | {{frame-path-challenge}}       |
-| 0x1b        | PATH_RESPONSE        | {{frame-path-response}}        |
-| 0x1c - 0x1d | CONNECTION_CLOSE     | {{frame-connection-close}}     |
+| 0x18 - 0x19 | NEW_CONNECTION_ID    | {{frame-new-connection-id}}    |
+| 0x1a        | RETIRE_CONNECTION_ID | {{frame-retire-connection-id}} |
+| 0x1b        | PATH_CHALLENGE       | {{frame-path-challenge}}       |
+| 0x1c        | PATH_RESPONSE        | {{frame-path-response}}        |
+| 0x1d - 0x1e | CONNECTION_CLOSE     | {{frame-connection-close}}     |
 {: #frame-types title="Frame Types"}
 
 An endpoint MUST treat the receipt of a frame of unknown type as a connection
@@ -4951,9 +4959,9 @@ Stream Limit:
 
 ## NEW_CONNECTION_ID Frame {#frame-new-connection-id}
 
-An endpoint sends a NEW_CONNECTION_ID frame (type=0x18) to provide its peer with
-alternative connection IDs that can be used to break linkability when migrating
-connections (see {{migration-linkability}}).
+An endpoint sends a NEW_CONNECTION_ID frame (type=0x18 or 0x19) to provide its
+peer with alternative connection IDs that can be used to break linkability when
+migrating connections (see {{migration-linkability}}).
 
 The NEW_CONNECTION_ID frame is as follows:
 
@@ -4962,6 +4970,8 @@ The NEW_CONNECTION_ID frame is as follows:
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                      Sequence Number (i)                    ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                    [Retire Prior To (i)]                    ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |   Length (8)  |                                               |
 +-+-+-+-+-+-+-+-+       Connection ID (32..144)                 +
@@ -4983,6 +4993,12 @@ Sequence Number:
 
 : The sequence number assigned to the connection ID by the sender.  See
   {{issue-cid}}.
+
+Retire Prior To:
+
+: A variable-length integer specifying the limit, for which all connection IDs
+  assigned to the receiver that have a sequence number less this value should be
+  retired.  See {{retiring-cids}}.  This field is only present for type=0x19.
 
 Length:
 
@@ -5018,10 +5034,16 @@ sequence number, or if a sequence number is used for different connection
 IDs, the endpoint MAY treat that receipt as a connection error of type
 PROTOCOL_VIOLATION.
 
+If the Retire Prior To field is present it is a request for the peer to
+immediately retire all connection IDs with a sequence number less than the
+specified value.  For more information see {{retiring-cids}}.  This field
+should only ever increase, though reordering and retransmission of older packets
+may make it seem to decrease.  So any decrease in this field should be ignored.
+
 
 ## RETIRE_CONNECTION_ID Frame {#frame-retire-connection-id}
 
-An endpoint sends a RETIRE_CONNECTION_ID frame (type=0x19) to indicate that it
+An endpoint sends a RETIRE_CONNECTION_ID frame (type=0x1b) to indicate that it
 will no longer use a connection ID that was issued by its peer. This may include
 the connection ID provided during the handshake.  Sending a RETIRE_CONNECTION_ID
 frame also serves as a request to the peer to send additional connection IDs for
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -994,6 +994,14 @@ packets sent from only one local address.  An endpoint that migrates away from a
...
 longer plans to use that address.
 
+The endpoint can explicitly request its peer to retire connection IDs by sending
+a NEW_CONNECTION_ID frame that contains a Retire Prior To field.  The peer is
+required to retire these connection IDs in a timely manner.  On receipt of the
+acknowledgement for the packet that contained the NEW_CONNECTION_ID frame, the
+endpoint may start a 3 PTO timer, which on expiring may close the connection
+with a PROTOCOL_VIOLATION error if all the connection IDs were not retired as
+requested.
+
```
mikkelfj: I think this is too aggressive. A single delayed packet could randomly take down a connection and attackers could speculate in that.

draft-ietf-quic-transport.md
```diff
@@ -5018,10 +5034,16 @@ sequence number, or if a sequence number is used for different connection
 IDs, the endpoint MAY treat that receipt as a connection error of type
 PROTOCOL_VIOLATION.
 
+If the Retire Prior To field is present it is a request for the peer to
+immediately retire all connection IDs with a sequence number less than the
+specified value.  For more information see {{retiring-cids}}.  This field
+should only ever increase, though reordering and retransmission of older packets
+may make it seem to decrease.  So any decrease in this field should be ignored.
+
```
mikkelfj: Retransmission of packets is not the right term here but I agree with the overall intent. To be consistent with other monotonic fields perhaps says something like the SHOULD not send a smaller value unless a previous frame is retransmitted verbatim. MUST ignore frames with a prior to value less than one previously received.

This means that the entire frame is to be ignored, not just the prior to field.

mikkelfj: Or, I'm not sure: it is a burden for the receiver to check the prior to field. The sequence number ought to be the only value ruling validity. One could state that the receiver MAY issue a protocol violation if the a sequence number is received hat contradict monotonic increase in prior to values.

draft-ietf-quic-transport.md
nibanks: How could a single delayed packet take the connection down? Best case, the same packet that acknowledges the NEW_CONNECTION_ID also contains all the RETIRE_CONNECTION_ID frames. Worst case, maybe you send them an RTT later. The timer here is 3 PTOs, not 3 RTT. If you haven't retired in that amount of time, how long of a time window should we allow?

draft-ietf-quic-transport.md
nibanks: We have the same problem with flow control values. We allow for implementations to naively retransmit packets that were lost, as is, so it's possible you have already updated the prior to field in a more recent packet, but then retransmit an older packet.

draft-ietf-quic-transport.md
nibanks: The other option here is not to kill the connection, but the sender just removes those CIDs from its lookup/routing tables. This means, if the peer continues to get them, they will get a stateless reset. I prefer the explicit behavior, because I feel like it is easier to diagnose when there is a problem; but I am willing to accept either solution.

draft-ietf-quic-transport.md
mikkelfj: I know, I just suggest a wording more close like those (don't recall the details)

draft-ietf-quic-transport.md
mikkelfj: Yes, I was thinking 3 RTT. 3 PTO may be enough, but an attacker could still delay so I prefer to drop such packets.

If you remove them from the routing table that is a deployment strategy - but most would likely keep a grace period before resetting or dropping. I'd imagine the middlebox load balancer is slow to update and only drops while backends are more aggressive because they are directly targeted. This is actually also true without explicit retirement - do we have any text on how to deal with that?

nibanks:
First, I don't see the frame type rearrange as a very large change at all. Second, I do think the atomic assurance is important, and it's just an oversight that we aren't explicitly require Sequence Number to be greater. I didn't think of that scenario.

Without the atomic assurance, I worry about the scenario where an endpoint retires all the CIDs without replacing any. It gets a lot more complicated if you have separate frames. What conditions are considered protocol violations?

In order to replace all existing CIDs, the order of the frames in a single packet starts to make a difference. You have to make sure you have the NEW_CONNECTION_ID frame followed by the RETIRE_ALL_CONNECTIONS_IDS_WITH_SEQUENCE_NUMBER_BEFORE frame.

kazuho:
I think I agree with @martinthomson that we should refrain from renumbering frames (again). One of the issues regarding the proposed renumbering is that the two CONNECTION_CLOSE frame types would not have the same MSB bits; it could be a concern for certain stacks.

OTOH, it is also my view that we do not need a new frame nor a new frame type.

How about just making the "Retire Prior To" a mandatory field of the NEW_CONNECTION_ID frame? The value can be set to zero when retirement is not necessary. We do not expect a lot of NEW_CONNECTION_ID to be inflight; spending one additional byte is IMO a non-issue.

draft-ietf-quic-transport.md
```diff
@@ -994,6 +994,14 @@ packets sent from only one local address.  An endpoint that migrates away from a
 local address SHOULD retire all connection IDs used on that address once it no
 longer plans to use that address.
 
+The endpoint can explicitly request its peer to retire connection IDs by sending
+a NEW_CONNECTION_ID frame that contains a Retire Prior To field.  The peer is
+required to retire these connection IDs in a timely manner.  On receipt of the
+acknowledgement for the packet that contained the NEW_CONNECTION_ID frame, the
+endpoint may start a 3 PTO timer, which on expiring may close the connection
```
martinthomson: MAY and MAY here

draft-ietf-quic-transport.md
```diff
@@ -4984,6 +4994,12 @@ Sequence Number:
 : The sequence number assigned to the connection ID by the sender.  See
   {{issue-cid}}.
 
+Retire Prior To:
+
+: A variable-length integer specifying the limit, for which all connection IDs
+  assigned to the receiver that have a sequence number less this value should be
+  retired.  See {{retiring-cids}}.  This field is only present for type=0x19.
```
martinthomson: This doesn't say anything about what to do with the connection ID that is negotiated during the handshake, and the one attached to a server preferred address.

draft-ietf-quic-transport.md
```diff
@@ -5018,10 +5034,16 @@ sequence number, or if a sequence number is used for different connection
 IDs, the endpoint MAY treat that receipt as a connection error of type
 PROTOCOL_VIOLATION.
 
+If the Retire Prior To field is present it is a request for the peer to
+immediately retire all connection IDs with a sequence number less than the
+specified value.  For more information see {{retiring-cids}}.  This field
+should only ever increase, though reordering and retransmission of older packets
```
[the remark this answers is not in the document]
martinthomson: There's lots of vague 2119-lite language here.  The language on flow control can be used here: 

> A receiver cannot renege on an advertisement however. That is, once a receiver advertises an offset, it MAY advertise a smaller offset, but this has no effect.
> [...]
> A sender MUST ignore any MAX_STREAM_DATA or MAX_DATA frames that do not increase flow control limits.

ianswett:
+1 to @kazuho suggestion of making it mandatory and not renumbering frame types.

draft-ietf-quic-transport.md
nibanks: I would also prefer a requirement that the peer MUST change CIDs immediately. I heard push back on that design. I would like to hear reasons why that wouldn't be a good idea, if anyone has some.

nibanks:
Thanks @ianswett @martinthomson @kazuho for your feedback. I have made some changes, making the Retire Prior To field required, so we only have one frame again, added more 2119 language, cleaned up the text on ignoring reductions in the Retire Prior To field, and made it a protocol violation to have Retire Prior To greater than or equal to Sequence Number.

@ianswett on your request to require the difference between the Retire Prior To field and Sequence Number field to be no greater than active_connection_id_limit, that would mean the sender has to keep track of both what's the highest retired sequence number retired by the peer and what it wants locally, and the max of those two values in this field. I'm not sure the extra complexity for keep track of that is worth the cost. What benefit do you see in that requirement?

@martinthomson on the topic of the handshake and preferred address CIDs, I thought we already had text that implicitly assigned them sequence numbers. Is that not the case? What text beyond that would you like me to add?

erickinnear:
Thanks for writing this up! 
Specific to this text (see additional discussion on the issue): 

I don't think that we need any requirement about the distance between RetirePriorTo/SequenceNumber active_connection_id_limit. Setting a super high value would be a bit nonsensical, so some non-normative text explaining why doing that is a complete waste of everyone's time would be good, but it doesn't actually break anything. Saying "hey this is silly, consider that limit when you set this" is probably good enough just to make implementors aware of the foot gun. 

To the point about handshake and preferred address CIDs, I think it would be nice to have some text just as a reminder along the lines of "don't forget these have sequence numbers too" to help folks avoid confusion.

draft-ietf-quic-transport.md
```diff
@@ -4984,6 +4994,12 @@ Sequence Number:
 : The sequence number assigned to the connection ID by the sender.  See
   {{issue-cid}}.
 
+Retire Prior To:
+
+: A variable-length integer specifying the limit, for which all connection IDs
+  assigned to the receiver that have a sequence number less this value should be
```
kazuho: less "than" this value?

marten-seemann:
I agree that atomicity is important here, if we want to avoid the situation where all connections IDs are retired at the same time. Adding the field to the NEW_CONNECTION_ID frame as @kazuho suggested seems to be the most straightforward way to do so.

I'm not a big fan of the 3 PTO requirement here. I don't see why it shouldn't be possible to retire the connections immediately when receiving the frame. There's a DoS risk here: if an attacker delays a packet with a connection ID that was retired afterwards, it can cause either a stateless reset or even a PROTOCOL_VIOLATION. The better way to enforce the retiring of the connection IDs is to look at packet numbers: The old connection IDs MUST NOT be used in packets with higher packet numbers than the packet that acknowledged the NEW_CONNECTION_ID frame.

kazuho:
@marten-seemann 
> The better way to enforce the retiring of the connection IDs is to look at packet numbers: The old connection IDs MUST NOT be used in packets with higher packet numbers than the packet that acknowledged the NEW_CONNECTION_ID frame.

Am I correct in understanding that this is the way key update works? Assuming that is the case, my +1 goes here.

marten-seemann:
> Am I correct in understanding that this is the way key update works? Assuming that is the case, my +1 goes here.

I think that's an implementation decision. You could certainly enforce this by requiring that `PN(packet sent with old connection IDs) < PN(first packet received with new connection ID)`. This doesn't stop a peer from acknowledging the NEW_CONNECTION_ID frame, but never acting on it though.

Another way to implement this would be to enforce that `PN(packet sent with old connection ID) < PN(ACK for NEW_CONNECTION_ID frame)`. This would be a somewhat stricter way, since it doesn't allow the peer to ignore the frame (without inducing artificial packet loss).

mikkelfj:
> The better way to enforce the retiring of the connection IDs is to look at packet numbers: The old connection IDs MUST NOT be used in packets with higher packet numbers than the packet that acknowledged the NEW_CONNECTION_ID frame.

I think this solves the problem since I really dislike that a delayed packet, 3PTO or otherwise, can take down a connection.

However, I also dislike that this is coupled strongly to ACK behaviour. I'm not sure how key updates work but assuming it does not use ACK, ACK handling can run asynchronously in a separate task, and indeed this is how I would architect a high performance implementation.

Then one could send a dedicated frame to acknowledge receiving the retire frame, but that is pointless since any packet using the new CID would have that effect.

So for the time being I don't have a good answer to this.

Altogether, if this facility opens up to DoS attacks and prevents ACK handling from running asynchronously, I think it does more harm than good and should at least be moved to v2 giving more time to consider the implications.

draft-ietf-quic-transport.md
MikeBishop: Mostly because "change immediately" is a difficult requirement to police.  Couching it in more concrete terms -- error on use of this CID in a packet with a higher packet number than the one which carried the ACK -- makes it easier to police in theory, but highlights what different parts of the system have to be coordinated to do it in practice.

Frankly, I'd even be okay with no longer recognizing those CIDs 3xPTO after sending the frame.  Fewer moving parts, easy to understand, but a small risk of stateless reset if the packet is severely delayed.

MikeBishop:
@ianswett, the problem is that the sequence can be sparse.  The client isn't required to use the CIDs in order any more, IIRC; so you could have four outstanding CIDs, but your four are sequence numbers (2, 127, 324, 324).  Nick wants to expire everything prior to 300 and he's tired of waiting for you to use and discard the two early ones.

nibanks:
@MikeBishop I'm all for making it immediate and adding stronger language. That's exactly what I wanted to do in the first place. Initially, I was getting push back though, so I relaxed it a bit. I agree it's a bit too wishy-washy in its current form. Unless anyone has reasons why they can't or shouldn't update immediately I will go ahead and strengthen the text.

MikeBishop:
I think the main reason not to be totally immediate is when you have some distributed processing generating packets.  You might not be able to guarantee that every thread/process generating packets can stop using those CIDs immediately.  However, we already went through that with the RETIRE_CID discussion -- sending RETIRE is supposed to be the signal that you've hit your sync point and *will not* use them any more.

erickinnear:
Brainstorming times when it would be valuable to not change immediately: 
- Privacy, a client that wants to avoid linkability on different links shouldn't have to change all of its CIDs regardless of link immediately, since that would trivially link those different CIDs
- If we have a server behind a load balancer and it wants to rotate keys, having every client connected to that individual server change CIDs right away could betray a bunch of information

I don't know if we need to go as far as recommending that clients wait some amount of time, but we could potentially (in QUIC LB maybe?) recommend that servers rotate their clients not all at once. And clients might want to change on different interfaces at different times.

I do like the idea of having a packet number cutoff when the client retires the CID to simplify things on the other end.

MikeBishop:
IIRC, the conclusion we reached on the CID design team was that:
- When the client sends RETIRE, that's a statement that it has purged that CID and SRT; the CID will never be used again, and the SRT won't be recognized.  The server can immediately declare that CID void upon receipt.
- If an earlier packet got delayed, it will trigger a SR, because the server doesn't recognize the CID any more.  This is also true if the client messed up and sends a later packet with the retired CID; the server will discard it.
- The client will receive the server's reset, but not recognize it as such because the SRT isn't one that it knows any more.

The important thing is that the client has already deleted the SRT before sending the RETIRE and the server can safely purge state immediately upon receipt.

In this PR, I think we have the right model by letting the client have a little time to make sure those requirements are true before sending the RETIRE.  You MUST retire the CIDs, but you're not obligated to do so instantly.

mikkelfj:
In line of @MikeBishop's comments and what I have said earlier:

Implicit retirement of "Prior To" has the issues mentioned with distributed ACK processing. But the problem is in fact worse: For almost the same reason that the sender wants to retired CID's, the receiver may want to be either migrating internally to another host, or process different parts of a connection on different servers. It can be detrimental to such deployments to have a force expiration. Therefore I think Mike is right in tying this to RETIRE_CID.

Perhaps the right model is to use RETIRE_CID as the sync point regardless of whether this is done voluntarily or motivated by a "Prior To" flush. Mike suggests the sender should defer RETIRE_CID until safe, but I think the receiver should make that call because it might have a 10 minute periodic load balancer flush or something completely different while the sender can only estimate based on RTT and the sender also wants to forget the CID ASAP while the receiver can track CID's algorithmically.

The recommendation could be to invalidate CIDs from RETIRE_CID frames no earlier than 3PTO and possibly longer if it suits the deployment. Receiving a non-zero "Prior To" field SHOULD cause the CIDs to be flushed and retired with RETIRE_CID as soon as practical. The issuer of "Prior To" MAY close a connection with the error CONNECTION_ID_TIMEOUT but deployments SHOULd avoid doing that aggressively unless critical to infrastructure since clients may not be able or willing to upgrade fast.

Note that there can be race conditions where both endpoints want to retire CID's because they are internally migrating. This can lead to problems if a CID is tied to the peers CID indirectly via source IP. We chose to have asymmetrical path migration due the complexities of this. I start to think we could some nasty corner cases here.

DavidSchinazi:
I'm OK with "SHOULD immediately" but I still think having an endpoint close the connection with protocol violation if a packet is delayed is a security risk for the reasons I detailed in the issue.

nibanks:
Personally, I don't think there is a big difference between explicitly closing the connection VS just removing the CIDs from the routing lookup, which would cause a stateless reset to be sent if the peer continues to use it.

I like the explicit nature of the close, because it will be easier to debug if there are bugs; though I do concede that if the peer did locally retire the CIDs, but somehow the packets/frames just took an extreme amount of time to get there, the connection isn't penalized.

Perhaps a vote?
Thumbs up if you prefer connection close with protocol violation.
Thumbs down if you prefer the CIDs to be removed, resulting in a stateless reset instead.

kazuho:
@nibanks 
> I don't think there is a big difference between explicitly closing the connection VS just removing the CIDs from the routing lookup, which would cause a stateless reset to be sent if the peer continues to use it.

As @DavidSchinazi has pointed out in https://github.com/quicwg/base-drafts/issues/2645#issuecomment-499470589, "having the use of a CID you've asked to retire trigger a protocol violation is a security risk. It allows an attacker to close the connection by delaying a valid packet."

If it is the case that we would have the same issue in sending a STATELESS_RESET (I assume that is what you are suggesting), I think we need to fix the issue, rather than choosing one between the two designs that both have the same security concern.

kazuho:
PS. One way of resolving the security concern would be to require the use of Stateless Reset (instead of a CONNECTION_CLOSE), and also require that Stateless Reset Tokens for CID above "Retire Prior To" to be different from the tokens below or equal to "Retire Prior To."

Note: the latter requirement is necessary because [transport draft section 10.4.2](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#rfc.section.10.4.2) states that "the same Stateless Reset Token MAY be used for multiple connection IDs on the same connection."

DavidSchinazi:
I agree with @kazuho. If we want to close the connection (which I still think we shouldn't mandate but I think we can allow) then we should use stateless resets, mandate each CID has a distinct reset token, and mandate that endpoints ignore reset tokens for CIDs they have retired. That'll prevent an attacker from closing the connection by delaying a packet.

mikkelfj:
> I agree with @kazuho. If we want to close the connection (which I still think we shouldn't mandate but I think we can allow) then we should use stateless resets, mandate each CID has a distinct reset token, and mandate that endpoints ignore reset tokens for CIDs they have retired. That'll prevent an attacker from closing the connection by delaying a packet.

That goes back to my suggestion of using epochs ("prior to" being the simplest such) because you might not want to store a reset token for every CID, only for every epoch.

marten-seemann:
I agree with @DavidSchinazi and @kazuho.

@mikkelfj In general, each CID is associated with its own reset token. This allows for a straightforward stateless implementation of a stateless reset token generator. I don't think it would be a good idea to revisit that principle.

mikkelfj:
@marten-seemann no you are probably right.

nibanks:
Thanks. All your arguments make sense. I have updated the text to say the connections IDs should just be removed after the 3 PTO, instead of closing the connection. I also added some text warning against reuse of the stateless reset tokens.

DavidSchinazi:
I'm not sure I know what "remove the connection ID" means on the receiver. I would remove any mention of 3 PTO as well. Simply state that if a sender keeps using a CID that its peer has asked to retire then packets can get misrouted or dropped.

nibanks:
@DavidSchinazi by "remove the connection ID" I meant that the connection IDs would be implicitly treated as retired. I'm not sure how best to word that section of the text. Suggestions are welcome. Additionally, I don't think many folks (though I'm not among them) would like the removal of the 3 PTO text. Several people have argued for allowing continued use of the CIDs for a short while.

mikkelfj:
I think the text needs to be more clear on whether RETIRE CID must be send for those requested to be retired, or if that is implicit.

nibanks:
@mikkelfj we have the following text already:

> The Retire Prior To field is a **requirement for the peer to retire** all connection
> IDs with a sequence number less than the specified value.  This includes the
> initial and preferred_address transport parameter connection IDs.  The **peer
> SHOULD immediately retire** all the connection IDs.

Is there something else you'd like added?

mikkelfj:
"The peer SHOULD stop using all the connection IDs as soon as possible and then retire the connection IDs by stop sending a RETIRE_CONNECTION_ID frame for each connection ID."

"retire immediately" is a bit too strong as several have commented.

MikeBishop:
> @mikkelfj In general, each CID is associated with its own reset token. This allows for a straightforward stateless implementation of a stateless reset token generator. I don't think it would be a good idea to revisit that principle.

This is true, but the text currently allows for the SRTs associated with various CIDs to be identical.  This becomes problematic if you're forcing retirement and want to stop recognizing some of the CIDs previously issued.  However, that's a problem we already have and cover in the text:

> The same Stateless Reset Token MAY be used for multiple connection IDs on the same connection. However, reuse of a Stateless Reset Token might expose an endpoint to denial of service if associated connection IDs are forgotten while the associated token is still active at a peer. An endpoint MUST ensure that packets with Destination Connection ID field values that correspond to a reused Stateless Reset Token are attributed to the same connection as long as the Stateless Reset Token is still usable, even when the connection ID has been retired. Otherwise, an attacker might be able to send a packet with a retired connection ID and cause the endpoint to produce a Stateless Reset that it can use to disrupt the connection, just as with the attacks in Section 21.8.

draft-ietf-quic-transport.md
```diff
@@ -5018,6 +5034,31 @@ sequence number, or if a sequence number is used for different connection
 IDs, the endpoint MAY treat that receipt as a connection error of type
 PROTOCOL_VIOLATION.
 
+The Retire Prior To field is a requirement for the peer to retire all connection
+IDs with a sequence number less than the specified value.  This includes the
+initial and preferred_address transport parameter connection IDs.  The peer
+SHOULD immediately retire all the connection IDs.
+
+The sender of the NEW_CONNECTION_ID frame MAY remove the connection IDs after 3
+PTO, even if the peer has not retired them yet.  The 3 PTO timer starts on
```
```suggestion
PTO, even if the peer has not retired them yet.  The 3-PTO timer starts on
```

draft-ietf-quic-transport.md
```diff
@@ -5018,6 +5034,31 @@ sequence number, or if a sequence number is used for different connection
...
+acknowledgement of the packet containing the NEW_CONNECTION_ID frame.  Continued
+use of the retired connection IDs after this point will likely result in a
+stateless reset being sent.
+
+In order to prevent a stateless reset, sent in response to a heavily delayed
```
```suggestion
In order to prevent a stateless reset sent in response to a heavily delayed
```

draft-ietf-quic-transport.md
```diff
@@ -5018,6 +5034,31 @@ sequence number, or if a sequence number is used for different connection
...
+packet using a retired connection ID, from closing the connection, the
```
```suggestion
packet using a retired connection ID from closing the connection, the
```
This parenthetical phrase is pretty critical and this sentence is pretty comma-heavy.  Consider promoting it.

draft-ietf-quic-transport.md
```diff
@@ -5018,6 +5034,31 @@ sequence number, or if a sequence number is used for different connection
 IDs, the endpoint MAY treat that receipt as a connection error of type
 PROTOCOL_VIOLATION.
 
+The Retire Prior To field is a requirement for the peer to retire all connection
+IDs with a sequence number less than the specified value.  This includes the
+initial and preferred_address transport parameter connection IDs.  The peer
+SHOULD immediately retire all the connection IDs.
+
+The sender of the NEW_CONNECTION_ID frame MAY remove the connection IDs after 3
+PTO, even if the peer has not retired them yet.  The 3-PTO timer starts on
```
ianswett: I forgot about RETIRE_CONNECTION_ID frame the first time I read this.  Can you say something like: "not retired them with RETIRE_CONNECTION_ID yet."

draft-ietf-quic-transport.md
```diff
@@ -5018,6 +5034,31 @@ sequence number, or if a sequence number is used for different connection
...
+acknowledgement of the packet containing the NEW_CONNECTION_ID frame.  Continued
+use of the retired connection IDs after this point will likely result in a
+stateless reset being sent.
+
+In order to prevent a stateless reset sent in response to a heavily delayed
```
ianswett: Rewording suggestion: "In order to prevent a heavily delayed packet with a retired connection ID from eliciting a stateless reset that closes the connection, the stateless reset token...."

draft-ietf-quic-transport.md
```diff
@@ -5018,6 +5034,31 @@ sequence number, or if a sequence number is used for different connection
...
+The Retire Prior To field is a requirement for the peer to retire all connection
+IDs with a sequence number less than the specified value.  This includes the
+initial and preferred_address transport parameter connection IDs.  The peer
+SHOULD immediately retire all the connection IDs.
+
+The sender of the NEW_CONNECTION_ID frame MAY remove the connection IDs after 3
+PTO, even if the peer has not retired them yet.  The 3-PTO timer starts on
+acknowledgement of the packet containing the NEW_CONNECTION_ID frame.  Continued
+use of the retired connection IDs after this point will likely result in a
+stateless reset being sent.
```
ianswett: It may be worth adding a paragraph for the receiver of the NEW_CONNECTION_ID frame about stateless reset. ie:
"If a stateless reset is received in response to a retired connection ID, the reset SHOULD be ignored and the receiver should immediately switch to a connection ID that has not been retired if it has not done so already."

draft-ietf-quic-transport.md
```diff
@@ -5018,6 +5042,35 @@ sequence number, or if a sequence number is used for different connection
 IDs, the endpoint MAY treat that receipt as a connection error of type
 PROTOCOL_VIOLATION.
 
+The Retire Prior To field is a request for the peer to retire all connection IDs
+with a sequence number less than the specified value.  This includes the initial
+and preferred_address transport parameter connection IDs.  The peer SHOULD
+immediately retire all the connection IDs.
```
DavidSchinazi: I would change this sentence to match the text before: `The peer SHOULD retire the corresponding connection IDs and send the corresponding RETIRE_CONNECTION_ID frame in a timely manner.` and also delete the next two paragraphs as they are redundant and contradict the "Detecting a Stateless Reset" section.

nibanks: To be clear, you suggesting I remove these two paragraphs?

> If a stateless reset is received in response to a retired connection ID, the
> reset SHOULD be ignored and the receiver SHOULD immediately switch to a
> connection ID that has not been retired if it has not done so already.
> 
> In order to prevent a heavily delayed packet with a retired connection ID from
> eliciting a stateless reset that closes the connection, the stateless reset
> token of the retired connection IDs SHOULD NOT be reused.

DavidSchinazi: Yes

draft-ietf-quic-transport.md
```diff
@@ -994,6 +994,22 @@ packets sent from only one local address.  An endpoint that migrates away from a
 local address SHOULD retire all connection IDs used on that address once it no
 longer plans to use that address.
 
+An endpoint can request that its peer retire connection IDs by sending a
+NEW_CONNECTION_ID frame with an increased Retire Prior To field.  Upon receipt,
```
ianswett: New_CONNECTION_ID frame(s)?

draft-ietf-quic-transport.md
```diff
@@ -994,6 +994,22 @@ packets sent from only one local address.  An endpoint that migrates away from a
...
+the peer SHOULD retire the corresponding connection IDs and send the
+corresponding RETIRE_CONNECTION_ID frame in a timely manner.  Failing to do so
+can cause packets to be delayed or lost and harm connection performance as the
```
ianswett: I'm a bit vague on "delayed or lost and harm connection performance as the original endpoint might not route those connection IDs optimally after some delay."  Instead, how about "delayed, lost, or cause the original endpoint to send a stateless reset in response to a connection ID it can no longer route correctly."?

draft-ietf-quic-transport.md
```diff
@@ -994,6 +994,22 @@ packets sent from only one local address.  An endpoint that migrates away from a
...
+original endpoint might not route those connection IDs optimally after some
+delay.
+
+The sender of the Retire Prior To field MUST keep track of the connection IDs it
```
ianswett: I still think this MUST is unenforceable and a bit odd.  If my load balancer restarts in 5 seconds and the peer hasn't retired the old CIDs by then, they're still getting retired.

draft-ietf-quic-transport.md
```diff
@@ -994,6 +994,22 @@ packets sent from only one local address.  An endpoint that migrates away from a
...
+wishes to retire until it has received a corresponding RETIRE_CONNECTION_ID
+frame, with one exception: if the sender of the Retire Prior To field has used
+distinct stateless reset tokens for all of their issued connection IDs, and 3
+times the PTO has elapsed since it received an acknowledgment for its Retire
+Prior To field, then it MAY lose track of that connection ID, and respond to
```
```suggestion
Prior To field, then it MAY drop state for that connection ID, and respond to
```

draft-ietf-quic-transport.md
```diff
@@ -5018,6 +5042,28 @@ sequence number, or if a sequence number is used for different connection
 IDs, the endpoint MAY treat that receipt as a connection error of type
 PROTOCOL_VIOLATION.
 
+The Retire Prior To field is a request for the peer to retire all connection IDs
+with a sequence number less than the specified value.  This includes the initial
+and preferred_address transport parameter connection IDs.  The peer SHOULD
+retire the corresponding connection IDs and send the corresponding
+RETIRE_CONNECTION_ID frame in a timely manner
```
ianswett: frame(s)

draft-ietf-quic-transport.md
nibanks: Change it to a `SHOULD`?

draft-ietf-quic-transport.md
```diff
@@ -994,6 +994,22 @@ packets sent from only one local address.  An endpoint that migrates away from a
...
+original endpoint might not route those connection IDs optimally after some
+delay.
+
+The sender of the Retire Prior To field MUST keep track of the connection IDs it
+wishes to retire until it has received a corresponding RETIRE_CONNECTION_ID
+frame, with one exception: if the sender of the Retire Prior To field has used
+distinct stateless reset tokens for all of their issued connection IDs, and 3
+times the PTO has elapsed since it received an acknowledgment for its Retire
+Prior To field, then it MAY lose track of that connection ID, and respond to
+packets with that connection ID with the corresponding stateless reset token.
```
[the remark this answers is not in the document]
nibanks: @erickinnear can you elaborate on this? We already have the requirement that `Retire Prior To` is <= than `Sequence Number` so it cannot retire all CIDs. I'm not sure if we can require replacement of ALL previously existing CIDs.

draft-ietf-quic-transport.md
```diff
@@ -5018,6 +5042,28 @@ sequence number, or if a sequence number is used for different connection
...
+with a sequence number less than the specified value.  This includes the initial
+and preferred_address transport parameter connection IDs.  The peer SHOULD
+retire the corresponding connection IDs and send the corresponding
+RETIRE_CONNECTION_ID frame in a timely manner
+
+The sender of the NEW_CONNECTION_ID frame MAY remove the connection IDs after 3
+PTO, even if the peer has not retired them with RETIRE_CONNECTION_ID yet.  The
+3-PTO timer starts on acknowledgement of the packet containing the
+NEW_CONNECTION_ID frame.  Continued use of the retired connection IDs after this
+point will likely result in a stateless reset being sent.
```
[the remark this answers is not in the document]
nibanks: I'm fine with removing it if you feel having it above is enough.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
ianswett: There's no requirement to supply as many as the peer wants, so there can't be a requirement to replace all it wants or as many as are retired either.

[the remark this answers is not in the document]
erickinnear: We had previously discussed introducing such a requirement, but I think we might already be covered here, since we're using the existing RETIRE_CID frame for the retirement, by the "SHOULD replace when retired" that's over in that part of the text.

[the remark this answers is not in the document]
erickinnear: (So you have 3 that get retired, for each that you retire, you get back a NEW_CID with the same RetirePriorTo as the original one)

draft-ietf-quic-transport.md
MikeBishop: It's currently a MUST to persist all the CIDs with the same SRT.  The way to be compliant would be to advance RPT enough to expire all the CIDs, even if there are some in the middle that could otherwise be kept.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
erickinnear: Yeah, at least personally it feels like there's guidance on how to do that up above already, I'd probably just skip it. :)

draft-ietf-quic-transport.md
```diff
@@ -994,6 +994,19 @@ packets sent from only one local address.  An endpoint that migrates away from a
 local address SHOULD retire all connection IDs used on that address once it no
 longer plans to use that address.
 
+An endpoint can request that its peer retire connection IDs by sending a
+NEW_CONNECTION_ID frames with an increased Retire Prior To field.  Upon receipt,
```
erickinnear: Sorry, typo: 
```suggestion
NEW_CONNECTION_ID frame with an increased Retire Prior To field.  Upon receipt,
```

erickinnear:
This is looking good (modulo the one extra paragraph at the end)!

draft-ietf-quic-transport.md
```diff
@@ -4984,6 +4999,12 @@ Sequence Number:
 : The sequence number assigned to the connection ID by the sender.  See
   {{issue-cid}}.
 
+Retire Prior To:
+
+: A variable-length integer specifying the limit for which all connection IDs
+  assigned to the receiver that have a sequence number less than this value
+  should be retired.  See {{retiring-cids}}.
```
marten-seemann: This sentence is kind of difficult to parse.

We should also specify that this value MUST be smaller or equal to the Sequence Number carried by this frame.

draft-ietf-quic-transport.md
```diff
@@ -5018,6 +5039,27 @@ sequence number, or if a sequence number is used for different connection
 IDs, the endpoint MAY treat that receipt as a connection error of type
 PROTOCOL_VIOLATION.
 
+The Retire Prior To field is a request for the peer to retire all connection IDs
+with a sequence number less than the specified value.  This includes the initial
+and preferred_address transport parameter connection IDs.  The peer SHOULD
+retire the corresponding connection IDs and send the corresponding
+RETIRE_CONNECTION_ID frames in a timely manner
```
marten-seemann: This feels duplicate to the first diff of this PR. Maybe we can move the sentence about the initial SPA connection ID up there?

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
nibanks: I prefer the text I have already.

draft-ietf-quic-transport.md
nibanks: The text below already species this must be <=. As improving the text, I agree it's a bit wordy. Do you have any specific suggestions for improving it?

MikeBishop: Given that the specifics of the limit are explained elsewhere, keep the field description brief.  "A variable-length integer indicating which connection IDs should be retired."

draft-ietf-quic-transport.md
```diff
@@ -994,6 +994,19 @@ packets sent from only one local address.  An endpoint that migrates away from a
...
+NEW_CONNECTION_ID frame with an increased Retire Prior To field.  Upon receipt,
+the peer SHOULD retire the corresponding connection IDs and send the
+corresponding RETIRE_CONNECTION_ID frames in a timely manner.  Failing to do so
+can cause packets to be delayed, lost, or cause the original endpoint to send a
+stateless reset in response to a connection ID it can no longer route correctly.
+
+An endpoint MAY discard a connection ID for which retirement has been requested
+once an interval of no less than 3 PTO has elapsed since an acknowledgement is
+received for the NEW_CONNECTION_ID frame requesting that retirement.  Subsequent
+incoming packets using that connection ID will elicit a response with the
```
igorlord: You want to say "Subsequent incoming packets using that connection ID **MAY** elicit a response with the corresponding stateless reset token."

Note that both behaviors are possible: MAY drop packets silently (misrouted, for example), and/or MAY elicit a stateless reset.

martinthomson: If by "MAY" you mean "could", that makes sense.  "MAY" in the sense of permitting something, is not really in question here.

igorlord: Yes, that's right.

PR #2802, opened 2019-06-18
remove misleading sentence about incompatible QUIC versions and ALPN

marten-seemann:
The server doesn't pick a version. The server only picks the application protocol from what the client offered, and it might do that based on the QUIC version that the client picked.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1279,8 +1279,7 @@ An application-layer protocol MAY restrict the QUIC versions that it can operate
 over.  Servers MUST select an application protocol compatible with the QUIC
 version that the client has selected.  If the server cannot select a compatible
 combination of application protocol and QUIC version, it MUST abort the
-connection. A client MUST abort a connection if the server picks an incompatible
-combination of QUIC version and ALPN identifier.
+connection.
 
 
 ## QUIC Transport Parameters Extension {#quic_parameters}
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -1279,8 +1279,7 @@ An application-layer protocol MAY restrict the QUIC versions that it can operate
 over.  Servers MUST select an application protocol compatible with the QUIC
```
mikkelfj: Plural singular mix: Servers vs a client

marten-seemann: Not part of this PR.

marten-seemann:
@martinthomson That depends on how you read RFC 7301, doesn't it?

Section 3.1 says:
> Servers that receive a ClientHello containing the "application_layer_protocol_negotiation" extension MAY return a suitable protocol selection response to the client.

I interpret "suitable" as "one of the options that the client proposed", but unfortunately that RFC is not very specific.

martinthomson:
That is confusing, but it doesn't mean that the server doesn't pick an ALPN.  We do require that it select an ALPN value, and even if it picks one that the client doesn't offer (whether legitimate or not), that doesn't invalidate this text.

draft-ietf-quic-tls.md
```diff
@@ -1279,8 +1279,7 @@ An application-layer protocol MAY restrict the QUIC versions that it can operate
...
 version that the client has selected.  If the server cannot select a compatible
 combination of application protocol and QUIC version, it MUST abort the
-connection. A client MUST abort a connection if the server picks an incompatible
-combination of QUIC version and ALPN identifier.
```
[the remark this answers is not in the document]
marten-seemann: I like that one. Just updated the PR.

PR #2806, opened 2019-06-19
Merge Crypto timeout and PTO

### LINKED ISSUE #2648: Crypto timeout value should include rttvar

### LINKED ISSUE #2650: Recovery would be simpler if the Handshake timer was unified with the PTO timer

### LINKED ISSUE #2886: Crypto Retransmission Timeout interacts poorly with Congestion Control
The crypto retransmission timeout works similar to the PTO timeout. On each timer expiration, you are supposed to send some data, and do an exponential backoff.
However, there's an important difference: PTO packets are not blocked by the congestion controller, whereas crypto retransmissions are. 

This can lead to a deadlock: Consider the case where the CFIN is lost, and the client sends all the available window of application data. When the crypto retransmission timer fires, the client would be supposed to retransmit the CFIN, but the congestion controller will not allow that. Furthermore, since the CFIN was lost, the server won't be able to read any of the 1-RTT packets, so cwnd will never be cleared.

ianswett:
This merges the crypto timeout into the PTO timeout.  Both have exponential backoff, so this is mostly a simplification, though this allows the crypto timeout to take RTT variance into account if possible(#2648) and I believe is simpler to understand.

A second attempt at PR #2655 

Fixes #2650, Fixes #2648
I believe this also fixes #2886

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -585,7 +585,7 @@ client.
 Because the server could be blocked until more packets are received, the client
 MUST ensure that the retransmission timer is set if there is
 unacknowledged crypto data or if the client does not yet have 1-RTT keys.
-If the retransmission timer expires before the client has 1-RTT keys,
+If the probe timer expires before the client has 1-RTT keys,
 it is possible that the client may not have any crypto data to retransmit.
 However, the client MUST send a new packet, containing only PADDING frames if
 necessary, to allow the server to continue sending data. If Handshake keys
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -165,7 +165,7 @@ of frames contained in a packet affect recovery and congestion control logic:
 
 * Long header packets that contain CRYPTO frames are critical to the
   performance of the QUIC handshake and use shorter timers for
-  acknowledgement and retransmission.
+  acknowledgement.
 
 * Packets that contain only ACK frames do not count toward congestion control
   limits and are not considered in-flight.
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1221,7 +1221,10 @@ OnLossDetectionTimeout():
   if (loss_time != 0):
     // Time threshold loss Detection
     DetectLostPackets(pn_space)
-  else if (endpoint is client without 1-RTT keys):
+    SetLossDetectionTimer()
+    return
+
+  if (endpoint is client without 1-RTT keys):
     // Client sends an anti-deadlock packet: Initial is padded
     // to earn more anti-amplification credit,
     // a Handshake packet proves address ownership.
@@ -1229,13 +1232,12 @@ OnLossDetectionTimeout():
       SendOneHandshakePacket()
     else:
       SendOnePaddedInitialPacket()
-    pto_count++
   else:
     // PTO. Send new data if available, else retransmit old data.
     // If neither is available, send a single PING frame.
     SendOneOrTwoPackets()
-    pto_count++
 
+  pto_count++
   SetLossDetectionTimer()
 ~~~
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -564,13 +564,11 @@ packet-threshold loss detection.
 ## Handshakes and new paths
 
 The initial probe timeout for a new connection or new path SHOULD be
-set to twice the initial RTT.
-
-Resumed connections over the same network SHOULD use the previous connection's
-final smoothed RTT value as the resumed connection's initial RTT.  If no
-previous RTT is available, or if the network changes, the initial RTT SHOULD
-be set to 500ms, resulting in a 1 second initial timeout as recommended in
-{{?RFC6298}}.
+set to twice the initial RTT.  Resumed connections over the same network 
+SHOULD use the previous connection's final smoothed RTT value as the resumed
+connection's initial RTT.  If no previous RTT is available, the initial RTT
+SHOULD be set to 500ms, resulting in a 1 second initial timeout as recommended
+in {{?RFC6298}}.
 
 A connection MAY use the delay between sending a PATH_CHALLENGE and receiving
 a PATH_RESPONSE to seed initial_rtt for a new path, but the delay SHOULD NOT
@@ -579,9 +577,9 @@ be considered an RTT sample.
 Until the server has validated the client's address on the path, the amount of
 data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
 If not all data can be retransmitted, then data at Initial and Handshake
-encryption should be retransmitted before any ApplicationData. If no data can be
-sent, then the PTO alarm should not be armed until data has been received from
-the client.
+encryption should be retransmitted before any ApplicationData data. If no data
+can be sent, then the PTO alarm should not be armed until data has been
+received from the client.
 
 Because the server could be blocked until more packets are received, the client
 MUST ensure that the retransmission timer is set if the client does not yet
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -580,8 +580,8 @@ Until the server has validated the client's address on the path, the amount of
 data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
 If not all data can be retransmitted, then data at Initial and Handshake
 encryption should be retransmitted before any ApplicationData. If no data can be
-sent, then no alarm should be armed until data has been received from the
-client.
+sent, then the PTO alarm should not be armed until data has been received from
+the client.
 
 Because the server could be blocked until more packets are received, the client
 MUST ensure that the retransmission timer is set if the client does not yet
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -553,8 +553,6 @@ sender might choose to optimize this by setting the timer fewer times if it
 knows that more ack-eliciting packets will be sent within a short period of
 time.
 
-
-
 The retransmission timer is not set if the time threshold
 {{time-threshold}} loss detection timer is set.  The time threshold loss
 detection timer is expected to both expire earlier than the crypto
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -583,9 +583,8 @@ sent, then no alarm should be armed until data has been received from the
 client.
 
 Because the server could be blocked until more packets are received, the client
-MUST ensure that the retransmission timer is set if there is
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -616,6 +553,50 @@ sender might choose to optimize this by setting the timer fewer times if it
...
+{{?RFC6298}}.
+
+A connection MAY use the delay between sending a PATH_CHALLENGE and receiving
+a PATH_RESPONSE to seed initial_rtt for a new path, but the delay SHOULD NOT
+be considered an RTT sample.
+
+Until the server has validated the client's address on the path, the amount of
+data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
+If not all unacknowledged CRYPTO data can be sent, then all unacknowledged
+CRYPTO data sent in Initial packets should be retransmitted.  If no data can be
```
```suggestion
CRYPTO data sent in Initial packets should be retransmitted if possible.  If no data can be
```

draft-ietf-quic-recovery.md
```diff
@@ -616,6 +553,50 @@ sender might choose to optimize this by setting the timer fewer times if it
...
+sent, then no alarm should be armed until data has been received from the
+client.
```
mikkelfj: How about idle timeout / connection abort?

marten-seemann: The idle timeout is a transport concept, it has nothing to do with loss recovery.

mikkelfj: Maybe say "no recovery alarm" or "no retransmission alarm"

draft-ietf-quic-recovery.md
```diff
@@ -616,6 +553,50 @@ sender might choose to optimize this by setting the timer fewer times if it
 knows that more ack-eliciting packets will be sent within a short period of
 time.
 
+The retransmission timer is not set if the time threshold
```
maolson-msft:
```suggestion
The probe timer is not set if the time threshold
```

draft-ietf-quic-recovery.md
```diff
@@ -616,6 +553,50 @@ sender might choose to optimize this by setting the timer fewer times if it
...
+data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
+If not all unacknowledged CRYPTO data can be sent, then all unacknowledged
+CRYPTO data sent in Initial packets should be retransmitted.  If no data can be
+sent, then no alarm should be armed until data has been received from the
+client.
+
+Because the server could be blocked until more packets are received, the client
+MUST ensure that the retransmission timer is set if there is
+unacknowledged crypto data or if the client does not yet have 1-RTT keys.
+If the retransmission timer expires before the client has 1-RTT keys,
```
maolson-msft:
```suggestion
If the probe timer expires before the client has 1-RTT keys,
```

ianswett:
Thanks for your patience, I believe this is now ready for a full review.

draft-ietf-quic-recovery.md
```diff
@@ -606,16 +544,57 @@ delay sending an acknowledgement.
...
 time.
 
+The probe timer is not set if the time threshold {{time-threshold}} loss
+detection timer is set.  The time threshold loss detection timer is expected
+to both expire earlier than the PTO and be less likely to spuriously retransmit
+data. The Initial and Handshake packet number spaces will typically contain a
+small number of packets, so losses are less likely to be detected using
+packet-threshold loss detection.
+
+## Handshakes and new paths
```
```suggestion
## Handshakes and New Paths
```

draft-ietf-quic-recovery.md
```diff
@@ -1283,7 +1237,7 @@ OnLossDetectionTimeout():
        SendOneHandshakePacket()
      else:
        SendOnePaddedInitialPacket()
-    crypto_count++
+    pto_count++
```
```suggestion
     pto_count++
```
The alignment looks off here.

draft-ietf-quic-recovery.md
```diff
@@ -543,49 +544,46 @@ delay sending an acknowledgement.
...
 immediately.
 
-When a PTO timer expires, the sender probes the network as described in the next
-section. The PTO period MUST be set to twice its current value. This exponential
-reduction in the sender's rate is important because the PTOs might be caused by
-loss of packets or acknowledgements due to severe congestion.
+When a PTO timer expires, the PTO period MUST be set to twice its current
+value. This exponential reduction in the sender's rate is important because
+the PTOs might be caused by loss of packets or acknowledgements due to severe
+congestion.
```
mikkelfj: Would we want some upper limit or this? What happens if idle timeout is not set - just indefinite inactivity?

marten-seemann: Why would there be no idle timeout?

mikkelfj: There should be, and maybe that is the answer. But I think a change was made to allow idle timeout to be indefinite. So I wonder how that will affect PTO.

marten-seemann: I don't remember any change like that. Every peer decides on its idle timeout. So even if a peer decides to have an infinite idle timeout (why would you do that to yourself?), once you reach a PTO duration on the order of what you might consider a typical idle timeout (let's say 30 seconds), exponential backoff should take care of your concern.

mikkelfj: Thats what I am asking - should the PTO just grow, or would you clamp it to 30 seconds or something?

ianswett: TCP clamps to a value(typically 30 seconds), but I think it's better to recommend a dead path timeout, such as a limit on the number of successive PTOs(I think some TCP impls default to 7?).

draft-ietf-quic-recovery.md
ianswett: You're correct, the else is off.

draft-ietf-quic-recovery.md
```diff
@@ -606,16 +544,57 @@ delay sending an acknowledgement.
...
+
+A connection MAY use the delay between sending a PATH_CHALLENGE and receiving
+a PATH_RESPONSE to seed initial_rtt for a new path, but the delay SHOULD NOT
+be considered an RTT sample.
+
+Until the server has validated the client's address on the path, the amount of
+data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
+If not all data can be retransmitted, then data at Initial and Handshake
+encryption should be retransmitted before any ApplicationData data. If no data
+can be sent, then the PTO alarm should not be armed until data has been
```
```suggestion
can be sent, then the PTO alarm MUST NOT be armed until data has been
```

draft-ietf-quic-recovery.md
janaiyengar: Maybe it's worth adding a sentence here to say that "The life of a connection that is experiencing consecutive PTOs is limited by the endpoint's idle timeout."

draft-ietf-quic-recovery.md
```diff
@@ -606,16 +544,56 @@ delay sending an acknowledgement.
...
+Because the server could be blocked until more packets are received, the client
+MUST ensure that the retransmission timer is set if the client does not yet
+have 1-RTT keys.  If the probe timer expires before the client has 1-RTT keys,
+it is possible that the client may not have any crypto data to retransmit.
+However, the client MUST send a new packet, containing only PADDING frames if
+necessary, to allow the server to continue sending data. If Handshake keys
+are available to the client, it MUST send a Handshake packet, and otherwise
+it MUST send an Initial packet in a UDP datagram of at least 1200 bytes.
+
+Because Initial packets only containing PADDING do not elicit an
```
RyanAtGoogle: nit: "only containing" => "containing only"

draft-ietf-quic-recovery.md
```diff
@@ -606,16 +544,56 @@ delay sending an acknowledgement.
...
+acknowledgement, they may never be acknowledged, but they are removed from
+bytes in flight when the client gets Handshake keys and the Initial keys are
+discarded.
```
RyanAtGoogle: Because of this, I wonder if it might make sense to include a PING along with an otherwise PADDING only packet?

draft-ietf-quic-recovery.md
```diff
@@ -512,76 +512,14 @@ thresholds reduce reordering resilience and increase spurious retransmissions,
...
-
-
 ## Probe Timeout {#pto}
 
-A Probe Timeout (PTO) triggers a probe packet when ack-eliciting data is in
-flight but an acknowledgement is not received within the expected period of
-time.  A PTO enables a connection to recover from loss of tail packets or acks.
-The PTO algorithm used in QUIC implements the reliability functions of Tail Loss
-Probe {{?TLP=I-D.dukkipati-tcpm-tcp-loss-probe}} {{?RACK}}, RTO {{?RFC5681}} and
+A Probe Timeout (PTO) triggers sending one or two probe packets when
```
RyanAtGoogle: This is probably obvious but it doesn't seem to explicitly say what causes 1 vs 2 packet to be sent. Perhaps this is clear elsewhere? I wonder if it might make sense to say something like "two packets, if there is sufficient data available, and one otherwise". But maybe that's not really helpful.

ianswett: It's 1 or 2 because some impls only do 1 packet PTO(TLP).  I could recommend two packets during the handshake, in order to speed handshake completion?

It actually should be datagrams, but I'm inclined to change that in a follow-up, rather that burying it in this change?

draft-ietf-quic-recovery.md
ianswett: True, we recommend that elsewhere.  I think we should change that in a follow-up though, since it's not directly related to this change?

ianswett: #2895 filed

PR #2811, opened 2019-06-19
Remove SETTINGS_NUM_PLACEHOLDERS

LPardue:
Addresses #2810


LPardue:
I think we prefer #2814 over this PR but leaving open for reference.

PR #2826, opened 2019-06-21
Make exceeding stream limits a connection error

### LINKED ISSUE #2825: Exceeding Max STREAM ID should be a connection error
In most places in QUIC, if an error is clearly avoidable(ie: only a buggy implementation would do so), then we advise closing the connection with an error, not just closing the stream with an error.

In section 4.5 of transport(https://tools.ietf.org/html/draft-ietf-quic-transport-20#section-4.5) it says:
"Endpoints MUST NOT exceed the limit set by their peer.  An endpoint
 that receives a STREAM frame with a stream ID exceeding the limit it
 has sent MUST treat this as a stream error of type STREAM_LIMIT_ERROR"

Came up in discussion of #2730

ianswett:
Fixes #2825

This aligns section 4.5 with section 19.11's text:
 "An endpoint MUST terminate a connection
  with a STREAM_LIMIT_ERROR error if a peer opens more streams than was
  permitted."

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -861,8 +861,8 @@ If either is received, the connection MUST be closed immediately with a
 connection error of type STREAM_LIMIT_ERROR (see {{immediate-close}}).
 
 Endpoints MUST NOT exceed the limit set by their peer.  An endpoint that
-receives a STREAM frame with a stream ID exceeding the limit it has sent MUST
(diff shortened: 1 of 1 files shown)


martinthomson:
I'm going to recommend calling this editorial on the basis that it reconciles this text with another MUST-level requirement.  But I will leave the issue open for the chairs to make a design/editorial assessment and any resulting consensus call.

PR #2831, opened 2019-06-23
fix effect of Retry on 0-RTT packets

marten-seemann:
As of #2746, the client MUST use the same handshake message.


marten-seemann:
@MikeBishop That's quite complicated, and I'm not sure if we've chosen the right design here.
I'd like to keep this PR editorial though. I opened #2842 to discuss this.

PR #2832, opened 2019-06-23
remove misleading sentence about connection IDs after Retry

marten-seemann:
Retry always changes the connection ID.


ianswett:
@nibanks The paragraph 2 before this one says exactly that.

PR #2846, opened 2019-06-26
fix overspecific coalescing requirement for Initials

marten-seemann:
It doesn't matter what an Initial is coalesced with. It could be a 0-RTT packet, it could be an (invalid) Handshake packet, or the bytes after the Initial packet could be completely random. The only thing that matters is that the resulting UDP datagram has at least 1200 bytes.


marten-seemann:
CI failure isn't my fault. The faulty commit seems to be bbe2b80d490c39c1d5900a5f00b4bbe11a29c4df.

PR #2853, opened 2019-06-27
transport: fix dangling ref to recovery

reschke:


draft-ietf-quic-transport.md
```diff
@@ -2924,7 +2924,7 @@ messages are delayed or lost.  Note that the same limitation applies to other
 data sent by the server protected by the 1-RTT keys.
 
 Endpoints SHOULD send acknowledgments for packets containing CRYPTO frames with
-a reduced delay; see Section 6.2.1 of {{QUIC-RECOVERY}}.
+a reduced delay; see Section 6.4 of {{QUIC-RECOVERY}}.
```
```suggestion
a reduced delay; see Section 6.2 of {{QUIC-RECOVERY}}.
```

reschke: No, I think the change is correct. See <https://tools.ietf.org/html/draft-ietf-quic-transport-20#section-13.1.2> and <https://tools.ietf.org/html/draft-ietf-quic-recovery-20#section-6.2.1>.

(FWIW, it would be good if these checks could be made part of the toolset used for the build)

martinthomson: There are changes in the editor's drafts. [Section 6.2](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#crypto-retransmission-timeout) deals with crypto timeout now.

I would love to have time to build a means of cross referencing when using markdown. It is just a frightening prospect. I find ruby and kramdown-rfc2629 both really difficult to work in. I haven't worked out how to do it architecturally, either. If you have any suggestions (other than using XML as the source format ...) I would be interested, though that might still be insufficient from a time perspective.

martinthomson:
Ian is fixing this for us in #2916.

PR #2857, opened 2019-06-27
Define terms for application actions

### LINKED ISSUE #2805: Rationalize terminology around transport/application interactions

MikeBishop:
This consists of two related commits:

- Defines the minimal set of operations which a QUIC implementation MUST expose to the application, introducing terms for them which don't rely on the specific wire formatting
- Changes HTTP/3 to use these terms instead of referring to specific QUIC frames

Note that this doesn't remove all references to QUIC frames, but the remaining references are explanatory -- they describe what will happen at the transport layer after taking a particular action, rather than being part of the normative text.  (But if you find one of these you think needs to go, please include that in this review.)

This is a design, because exposing this functionality is mandatory.  I'm not sure how you would previously have created a usable QUIC implementation without these operations, but you weren't previously required to have them all.

Fixes #2805, closes #2677.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -452,12 +452,12 @@ HTTP_INCOMPLETE_REQUEST.
 
 A server can send a complete response prior to the client sending an entire
 request if the response does not depend on any portion of the request that has
-not been sent and received. When this is true, a server MAY request that the
-client abort transmission of a request without error by triggering a QUIC
-STOP_SENDING frame with error code HTTP_EARLY_RESPONSE, sending a complete
-response, and cleanly closing its stream. Clients MUST NOT discard complete
-responses as a result of having their request terminated abruptly, though
-clients can always discard responses at their discretion for other reasons.
+not been sent and received. When this is true, a server MAY abort reading the
+receiving part of the request stream with error code HTTP_EARLY_RESPONSE,
+sending a complete response, and cleanly closing the sending part of the stream.
+Clients MUST NOT discard complete responses as a result of having their request
+terminated abruptly, though clients can always discard responses at their
+discretion for other reasons.
 
 
 ### Header Formatting and Compression {#header-formatting}
@@ -505,11 +505,11 @@ this limit are not guaranteed to be accepted.
 
 ### Request Cancellation and Rejection {#request-cancellation}
 
-Clients can cancel requests by aborting the stream (QUIC RESET_STREAM and/or
-STOP_SENDING frames, as appropriate) with an error code of
-HTTP_REQUEST_CANCELLED ({{http-error-codes}}).  When the client cancels a
-response, it indicates that this response is no longer of interest.
-Implementations SHOULD cancel requests by aborting both directions of a stream.
+Clients can cancel requests by abruptly terminating the sending part of the
+request stream with an error code of HTTP_REQUEST_CANCELLED
+({{http-error-codes}}).  When the client aborts reading a response, it indicates
+that this response is no longer of interest. Implementations SHOULD cancel
+requests by terminating both directions of a stream.
 
 When the server rejects a request without performing any application processing,
 it SHOULD abort its response stream with the error code HTTP_REQUEST_REJECTED.
@@ -522,11 +522,11 @@ were partially or fully processed.  When a server abandons a response after
 partial processing, it SHOULD abort its response stream with the error code
 HTTP_REQUEST_CANCELLED.
 
-When a client sends a STOP_SENDING with HTTP_REQUEST_CANCELLED, a server MAY
-send the error code HTTP_REQUEST_REJECTED in the corresponding RESET_STREAM
-if no processing was performed.  Clients MUST NOT reset streams with the
-HTTP_REQUEST_REJECTED error code except in response to a QUIC STOP_SENDING
-frame that contains the same code.
+When a client abruptly terminates a request with the error code
+HTTP_REQUEST_CANCELLED, a server MAY abruptly terminate the response using the
+error code HTTP_REQUEST_REJECTED if no processing was performed.  Clients MUST
+NOT use the HTTP_REQUEST_REJECTED error code, except when a server has requested
+closure of the request stream with this error code.
 
 If a stream is cancelled after receiving a complete response, the client MAY
 ignore the cancellation and use the response.  However, if a stream is cancelled
@@ -599,9 +599,9 @@ a single direction are not invalid, but are often handled poorly by servers, so
 clients SHOULD NOT close a stream for sending while they still expect to receive
 data from the target of the CONNECT.
 
-A TCP connection error is signaled with QUIC RESET_STREAM frame. A proxy treats
-any error in the TCP connection, which includes receiving a TCP segment with the
-RST bit set, as a stream error of type HTTP_CONNECT_ERROR
+A TCP connection error is signaled by abruptly terminating the stream. A proxy
+treats any error in the TCP connection, which includes receiving a TCP segment
+with the RST bit set, as a stream error of type HTTP_CONNECT_ERROR
 ({{http-error-codes}}).  Correspondingly, if a proxy detects an error with the
 stream or the QUIC connection, it MUST close the TCP connection.  If the
 underlying TCP implementation permits it, the proxy SHOULD send a TCP segment
@@ -785,9 +785,9 @@ amount of data a server may commit to the pushed stream.
 
 If a promised server push is not needed by the client, the client SHOULD send a
 CANCEL_PUSH frame. If the push stream is already open or opens after sending the
-CANCEL_PUSH frame, a QUIC STOP_SENDING frame with an error code of
-HTTP_REQUEST_CANCELLED can be used. This asks the server not to transfer
-additional data and indicates that it will be discarded upon receipt.
+CANCEL_PUSH frame, the client can abort reading the stream with an error code of
+HTTP_REQUEST_CANCELLED. This asks the server not to transfer additional data and
+indicates that it will be discarded upon receipt.
 
 # Connection Closure
 
@@ -827,9 +827,9 @@ Servers initiate the shutdown of a connection by sending a GOAWAY frame
 on lower stream IDs were or might be processed in this connection, while
 requests on the indicated stream ID and greater were rejected. This enables
 client and server to agree on which requests were accepted prior to the
-connection shutdown.  This identifier MAY be zero if no requests were
-processed.  Servers SHOULD NOT increase the QUIC MAX_STREAMS limit after
-sending a GOAWAY frame.
+connection shutdown.  This identifier MAY be zero if no requests were processed.
+Servers SHOULD NOT permit additional transport streams after sending a GOAWAY
+frame.
 
 Clients MUST NOT send new requests on the connection after receiving GOAWAY;
 a new connection MAY be established to send additional requests.
@@ -927,12 +927,10 @@ All client-initiated bidirectional streams are used for HTTP requests and
 responses.  A bidirectional stream ensures that the response can be readily
 correlated with the request. This means that the client's first request occurs
 on QUIC stream 0, with subsequent requests on stream 4, 8, and so on. In order
-to permit these streams to open, an HTTP/3 client SHOULD send non-zero values
-for the QUIC transport parameters `initial_max_stream_data_bidi_local`. An
-HTTP/3 server SHOULD send non-zero values for the QUIC transport parameters
-`initial_max_stream_data_bidi_remote` and `initial_max_bidi_streams`. It is
-RECOMMENDED that `initial_max_bidi_streams` be no smaller than 100, so as to not
-unnecessarily limit parallelism.
+to permit these streams to open, an HTTP/3 server SHOULD configure non-zero
+minimum values for the number of permitted streams and the initial flow control
+window for each stream. It is RECOMMENDED that at least 100 requests be
+permitted at a time, so as to not unnecessarily limit parallelism.
 
 HTTP/3 does not use server-initiated bidirectional streams, though an extension
 could define a use for these streams.  Clients MUST treat receipt of a
@@ -962,18 +960,16 @@ see {{extensions}} for more details.
 
 The performance of HTTP/3 connections in the early phase of their lifetime is
 sensitive to the creation and exchange of data on unidirectional streams.
-Endpoints that set low values for the QUIC transport parameters
-`initial_max_uni_streams` and `initial_max_stream_data_uni` will increase the
-chance that the remote peer reaches the limit early and becomes blocked. In
-particular, the value chosen for `initial_max_uni_streams` should consider that
-remote peers may wish to exercise reserved stream behavior ({{stream-grease}}).
-To avoid blocking, both clients and servers MUST allow the peer to create at
-least one unidirectional stream for the HTTP control stream plus the number of
-unidirectional streams required by mandatory extensions (such as QPACK) by
-setting an appropriate value for the QUIC transport parameter
-`initial_max_uni_streams` (three being the minimum value required for the base
-HTTP/3 protocol and QPACK), and SHOULD use a value of 1,024 or greater for the
-QUIC transport parameter `initial_max_stream_data_uni`.
+Endpoints that excessively restrict the number or flow control window of these
+streams will increase the chance that the remote peer reaches the limit early
+and becomes blocked. In particular, implementations should consider that remote
+peers may wish to exercise reserved stream behavior ({{stream-grease}}) with
+some of the unidirectional streams they are permitted to use. To avoid blocking,
+both clients and servers MUST allow the peer to create at least one
+unidirectional stream for the HTTP control stream plus the number of
+unidirectional streams required by mandatory extensions (three being the minimum
+number required for the base HTTP/3 protocol and QPACK), and SHOULD provide at
+least 1,024 bytes of flow control credit to each stream.
 
 Note that an endpoint is not required to grant additional credits to create more
 unidirectional streams if its peer consumes all the initial credits before
@@ -984,9 +980,9 @@ create additional streams as allowed by their peer.
 
 If the stream header indicates a stream type which is not supported by the
 recipient, the remainder of the stream cannot be consumed as the semantics are
-unknown. Recipients of unknown stream types MAY trigger a QUIC STOP_SENDING
-frame with an error code of HTTP_STREAM_CREATION_ERROR, but MUST NOT consider
-such streams to be a connection error of any kind.
+unknown. Recipients of unknown stream types MAY abort reading of the stream with
+an error code of HTTP_STREAM_CREATION_ERROR, but MUST NOT consider such streams
+to be a connection error of any kind.
 
 Implementations MAY send stream types before knowing whether the peer supports
 them.  However, stream types which could modify the state or semantics of
@@ -1277,14 +1273,12 @@ When a server receives this frame, it aborts sending the response for the
 identified server push.  If the server has not yet started to send the server
 push, it can use the receipt of a CANCEL_PUSH frame to avoid opening a push
 stream.  If the push stream has been opened by the server, the server SHOULD
-send a QUIC RESET_STREAM frame on that stream and cease transmission of the
-response.
+abruptly terminate that stream.
 
 A server can send the CANCEL_PUSH frame to indicate that it will not be
 fulfilling a promise prior to creation of a push stream.  Once the push stream
 has been created, sending CANCEL_PUSH has no effect on the state of the push
-stream.  A QUIC RESET_STREAM frame SHOULD be used instead to abort transmission
-of the server push response.
+stream.  The server SHOULD abruptly terminate the push stream instead.
 
 A CANCEL_PUSH frame is sent on the control stream.  Receiving a CANCEL_PUSH
 frame on a stream other than the control stream MUST be treated as a connection
@@ -1493,7 +1487,7 @@ The MAX_PUSH_ID frame (type=0xD) is used by clients to control the number of
 server pushes that the server can initiate.  This sets the maximum value for a
 Push ID that the server can use in a PUSH_PROMISE frame.  Consequently, this
 also limits the number of push streams that the server can initiate in addition
-to the limit set by the QUIC MAX_STREAMS frame.
+to the limit maintained by the QUIC transport.
 
 The MAX_PUSH_ID frame is always sent on the control stream.  Receipt of a
 MAX_PUSH_ID frame on any other stream MUST be treated as a connection error of
@@ -1590,8 +1584,9 @@ the cause of a connection or stream error.
 
 ## HTTP/3 Error Codes {#http-error-codes}
 
-The following error codes are defined for use in QUIC RESET_STREAM frames,
-STOP_SENDING frames, and CONNECTION_CLOSE frames when using HTTP/3.
+The following error codes are defined for use when abruptly terminating the
+sending part of streams, aborting reading of the receiving part of streams, or
+immediately closing connections.
 
 HTTP_NO_ERROR (0x00):
 : No error.  This is used when the connection or stream needs to be closed, but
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-http.md
```diff
@@ -452,12 +452,12 @@ HTTP_INCOMPLETE_REQUEST.
...
 A server can send a complete response prior to the client sending an entire
 request if the response does not depend on any portion of the request that has
-not been sent and received. When this is true, a server MAY request that the
-client abort transmission of a request without error by triggering a QUIC
-STOP_SENDING frame with error code HTTP_EARLY_RESPONSE, sending a complete
-response, and cleanly closing its stream. Clients MUST NOT discard complete
-responses as a result of having their request terminated abruptly, though
-clients can always discard responses at their discretion for other reasons.
+not been sent and received. When this is true, a server MAY abort reading the
+receiving part of the request stream with error code HTTP_EARLY_RESPONSE, send a
```
martinthomson: Do you think that we can just say "abort reading the request stream" and drop the  "receiving part of" piece on the basis that "abort reading" can only apply to the receiving part?

draft-ietf-quic-http.md
```diff
@@ -505,11 +505,11 @@ this limit are not guaranteed to be accepted.
 
 ### Request Cancellation and Rejection {#request-cancellation}
 
-Clients can cancel requests by aborting the stream (QUIC RESET_STREAM and/or
-STOP_SENDING frames, as appropriate) with an error code of
-HTTP_REQUEST_CANCELLED ({{http-error-codes}}).  When the client cancels a
-response, it indicates that this response is no longer of interest.
-Implementations SHOULD cancel requests by aborting both directions of a stream.
+Clients can cancel requests by abruptly terminating the sending part of the
```
martinthomson: "abruptly terminating" or just "resetting" ?

draft-ietf-quic-http.md
```diff
@@ -1590,8 +1584,9 @@ the cause of a connection or stream error.
 
 ## HTTP/3 Error Codes {#http-error-codes}
 
-The following error codes are defined for use in QUIC RESET_STREAM frames,
-STOP_SENDING frames, and CONNECTION_CLOSE frames when using HTTP/3.
+The following error codes are defined for use when abruptly terminating the
+sending part of streams, aborting reading of the receiving part of streams, or
```
martinthomson: As above regarding reference to "parts".

draft-ietf-quic-http.md
```diff
@@ -1718,8 +1713,8 @@ extension is disabled if the setting is omitted.
 The security considerations of HTTP/3 should be comparable to those of HTTP/2
 with TLS.  Note that where HTTP/2 employs PADDING frames and Padding fields in
 other frames to make a connection more resistant to traffic analysis, HTTP/3 can
-rely on QUIC PADDING frames or employ the reserved frame and stream types
-discussed in {{frame-grease}} and {{stream-grease}}.
+either rely on transport-layer padding or employ the reserved frame and stream
+types discussed in {{frame-grease}} and {{stream-grease}}.
```
martinthomson: Aside: we should probably have a conversation about the trade-offs with padding.  Transport-layer padding is in some ways superior, like in how it can be granular to a single byte, which reserved streams and frames cannot do.  But PADDING frames at the transport aren't retransmitted when lost, which could lead to some interesting attacks if packets can be suppressed by an attacker.  If retransmissions aren't padded, boom.

draft-ietf-quic-transport.md
```diff
@@ -345,6 +345,31 @@ the relative priority of streams.  When deciding which streams to dedicate
...
+## Required Operations on Streams
+
+There are certain operations which an application MUST be able to perform when
+interacting with QUIC streams.  This document does not specify an API, but
+any implementation of this version of QUIC MUST expose the ability to perform
+the operations described in this section on a QUIC stream.
+
+On the sending part of a stream:
+
+- Attempt to write data, understanding when stream flow control credit
```
martinthomson: List construction here implies continuation of a sentence, but the capital letter doesn't fit and there is no joiner.

I think that you want:

> On the sending part of a stream, application protocols need to be able to:
> * write data, understanding when .... ({{frame-stop-sending}});
> * end the stream, ....; and
> * reset the stream ... was already ended.

draft-ietf-quic-transport.md
```diff
@@ -345,6 +345,31 @@ the relative priority of streams.  When deciding which streams to dedicate
...
+  ({{data-flow-control}}) has successfully been reserved to send the written
+  data or possibly discovering that the stream has been closed because the peer
```
martinthomson: I think that you might remove this "possibly discovering" clause.  Then add a new separate paragraph after the lists that says that application protocols will want to be informed about stream state changes and events, including when data is available, when the stream ends, is reset, or reading has been aborted by a peer, and when new streams are created by a peer.

As a general rule, I don't like having sections end on a list, so that would help with my little compulsion too.

draft-ietf-quic-transport.md
```diff
@@ -1101,6 +1126,38 @@ suggested structure:
...
 
 
+## Required Operations on Connections
+
+There are certain operations which an application MUST be able to perform when
+interacting with the QUIC transport.  This document does not specify an API, but
+any implementation of this version of QUIC MUST expose the ability to perform
+the operations described in this section on a QUIC connection.
+
+When implementing the client role:
```
martinthomson: Same list construction comment as above.

draft-ietf-quic-transport.md
```diff
@@ -1101,6 +1126,38 @@ suggested structure:
...
+
+- Open a connection, which begins the exchange described in {{handshake}}
```
martinthomson: Add: to enable 0-RTT, and to be informed when 0-RTT has been either accepted or rejected by a server.

draft-ietf-quic-transport.md
```diff
@@ -1101,6 +1126,38 @@ suggested structure:
...
+
+When implementing the server role:
+
+- Listen for incoming connections, which prepares for the exchange described in
+  {{handshake}}
+- If Early Data is supported, embed application-controlled data in the TLS
+  resumption ticket sent to the client
+- If Early Data is supported, retrieve application-controlled data from the
+  client's resumption ticket and approve/veto accepting Early Data
```
martinthomson: and enable rejection of 0-RTT based on that information

I don't like "veto" much here.  And we don't need to have an active acceptance involved.

draft-ietf-quic-transport.md
```diff
@@ -1101,6 +1126,38 @@ suggested structure:
...
+
+In either role:
+
+- Configure minimum values for the initial number of permitted streams of each
+  type, as communicated in the transport parameters ({{transport-parameters}})
```
martinthomson: I think that there is a more general point, which basically means controlling resource allocation of all types, including streams and flow control.

The point you are looking at here - setting a minimum on the number of streams - is something HTTP needs, so I would add rather than replace.

draft-ietf-quic-transport.md
```diff
@@ -1101,6 +1126,38 @@ suggested structure:
...
+- Keep a connection from silently closing, either by generating PING frames
+  ({{frame-ping}}) or by requesting that the transport send additional frames
+  before the idle timeout expires ({{idle-timeout}})
+- Cease issuing credit for new streams via MAX_STREAMS frames
```
martinthomson: See the above point about controlling resource allocation.

draft-ietf-quic-transport.md
```diff
@@ -345,6 +345,31 @@ the relative priority of streams.  When deciding which streams to dedicate
...
+- Abruptly terminate the stream, resulting in a RESET_STREAM frame
+  ({{frame-reset-stream}}), even if the stream was cleanly terminated previously
+
+On the receiving part of a stream:
+
+- Attempt to read data, possibly discovering that the peer has terminated the
+  stream either cleanly or abruptly
+- Abort reading of the stream and request closure, resulting in a STOP_SENDING
+  frame ({{frame-stop-sending}})
+
```
mikkelfj: This could be read as if aborting a stream should always result in a STOP_SENDING frame, but I don't think that that is necessarily true. I understand that the ability to do so is important though.

draft-ietf-quic-transport.md
mikkelfj: Cease and resume

MikeBishop: The only place where this is required by HTTP is to support GOAWAY; I'd argue that the ability to stop is required, and the ability to resume is something an implementation could choose to expose.

draft-ietf-quic-transport.md
MikeBishop: It's always true if the recipient hasn't already received through the end of the stream.  But yes, I'll try to expand on it.  It might be worth reframing some things in the transport doc to work in terms of acting on these requests from the application.

draft-ietf-quic-http.md
```diff
@@ -962,18 +960,16 @@ see {{extensions}} for more details.
...
-particular, the value chosen for `initial_max_uni_streams` should consider that
-remote peers may wish to exercise reserved stream behavior ({{stream-grease}}).
-To avoid blocking, both clients and servers MUST allow the peer to create at
-least one unidirectional stream for the HTTP control stream plus the number of
-unidirectional streams required by mandatory extensions (such as QPACK) by
-setting an appropriate value for the QUIC transport parameter
-`initial_max_uni_streams` (three being the minimum value required for the base
-HTTP/3 protocol and QPACK), and SHOULD use a value of 1,024 or greater for the
-QUIC transport parameter `initial_max_stream_data_uni`.
+Endpoints that excessively restrict the number or flow control window of these
```
LPardue: This reads weird to me, my brain keeps thinking "or" is a typo for "of", which is untrue. I think it could read better to say

"Endpoints that excessively restrict the number of these streams or the size of each stream's flow control window will increase the chance that the remote peer reaches the limit early and becomes blocked."

draft-ietf-quic-http.md
```diff
@@ -962,18 +960,16 @@ see {{extensions}} for more details.
...
+streams will increase the chance that the remote peer reaches the limit early
+and becomes blocked. In particular, implementations should consider that remote
+peers may wish to exercise reserved stream behavior ({{stream-grease}}) with
+some of the unidirectional streams they are permitted to use. To avoid blocking,
+both clients and servers MUST allow the peer to create at least one
+unidirectional stream for the HTTP control stream plus the number of
+unidirectional streams required by mandatory extensions (three being the minimum
+number required for the base HTTP/3 protocol and QPACK), and SHOULD provide at
+least 1,024 bytes of flow control credit to each stream.
```
LPardue: observation: the change to remove references to Transport Parameters in this paragraph could lead to some interesting unintended issues. Previously it was clear that the problems mentioned were related to small values of initial limits (imposed by TPs). One way to interpet the new text is that it is ok to advertise low values in TPs as long as you send MAX_STREAMS or MAX_STREAM_DATA.

Just wondering if that was the intent of this change, given some of the problems we've seen with these values during interop.

draft-ietf-quic-transport.md
```diff
@@ -345,6 +345,34 @@ the relative priority of streams.  When deciding which streams to dedicate
...
+On the receiving part of a stream, application protocols need to be able to:
+
+- read data
+- abort reading of the stream and request closure, possibly resulting in a
+  STOP_SENDING frame ({{frame-stop-sending}})
+
+Applications also need to be informed of state changes on streams, including
+when the peer has initiated, reset, or aborted reading on a stream; when new
+data is available; and when data on the stream is blocked due to flow
```
ianswett: Should this be flow or congestion control?  And should there be a signal when the stream is unblocked so you can resume sending?

draft-ietf-quic-transport.md
```diff
@@ -1101,6 +1132,41 @@ suggested structure:
...
+There are certain operations which an application MUST be able to perform when
+interacting with the QUIC transport.  This document does not specify an API, but
+any implementation of this version of QUIC MUST expose the ability to perform
+the operations described in this section on a QUIC connection.
+
+When implementing the client role, applications need to be able to:
+
+- open a connection, which begins the exchange described in {{handshake}};
+- enable 0-RTT; and
+- be informed when 0-RTT has been accepted or rejected by a server.
```
ianswett: I'm not certain the application needs to know about 0-RTT accept/reject, but I believe it does need to know when it can start sending 1RTT data?

draft-ietf-quic-http.md
```diff
@@ -827,9 +827,9 @@ Servers initiate the shutdown of a connection by sending a GOAWAY frame
 on lower stream IDs were or might be processed in this connection, while
 requests on the indicated stream ID and greater were rejected. This enables
 client and server to agree on which requests were accepted prior to the
-connection shutdown.  This identifier MAY be zero if no requests were
-processed.  Servers SHOULD NOT increase the QUIC MAX_STREAMS limit after
-sending a GOAWAY frame.
+connection shutdown.  This identifier MAY be zero if no requests were processed.
+Servers SHOULD NOT permit additional transport streams after sending a GOAWAY
```
dtikhonov: "transport stream" is new term not defined elsewhere in the document.

draft-ietf-quic-transport.md
```diff
@@ -1101,6 +1132,41 @@ suggested structure:
...
+
+When implementing the server role, applications need to be able to:
+
+- listen for incoming connections, which prepares for the exchange described in
+  {{handshake}};
+- if Early Data is supported, embed application-controlled data in the TLS
+  resumption ticket sent to the client; and
+- if Early Data is supported, retrieve application-controlled data from the
+  client's resumption ticket and enable rejecting Early Data based on that
+  information.
```
ianswett: Servers need to expose whether incoming data is Early Data, correct? ie: for the Early Data header.

draft-ietf-quic-http.md
```diff
@@ -927,12 +927,10 @@ All client-initiated bidirectional streams are used for HTTP requests and
...
 on QUIC stream 0, with subsequent requests on stream 4, 8, and so on. In order
-to permit these streams to open, an HTTP/3 client SHOULD send non-zero values
-for the QUIC transport parameters `initial_max_stream_data_bidi_local`. An
-HTTP/3 server SHOULD send non-zero values for the QUIC transport parameters
-`initial_max_stream_data_bidi_remote` and `initial_max_bidi_streams`. It is
-RECOMMENDED that `initial_max_bidi_streams` be no smaller than 100, so as to not
-unnecessarily limit parallelism.
+to permit these streams to open, an HTTP/3 server SHOULD configure non-zero
+minimum values for the number of permitted streams and the initial flow control
+window for each stream. It is RECOMMENDED that at least 100 requests be
```
dtikhonov: "initial flow control window for each stream" makes it sound as if each stream's flow control window could be configured independently.

draft-ietf-quic-http.md
```diff
@@ -505,11 +505,11 @@ this limit are not guaranteed to be accepted.
 
 ### Request Cancellation and Rejection {#request-cancellation}
 
-Clients can cancel requests by aborting the stream (QUIC RESET_STREAM and/or
-STOP_SENDING frames, as appropriate) with an error code of
-HTTP_REQUEST_CANCELLED ({{http-error-codes}}).  When the client cancels a
-response, it indicates that this response is no longer of interest.
-Implementations SHOULD cancel requests by aborting both directions of a stream.
+Clients can cancel requests by resetting the request stream with an error code
+of HTTP_REQUEST_CANCELLED ({{http-error-codes}}).  When the client aborts
```
ianswett: How about "Clients can cancel requests by resetting and aborting the request stream with an error code of HTTP_REQUEST_CANCELLED"?  I think that phrasing makes it clearer that you need to reset the write side and abort the read side in order to fully cancel a request?

draft-ietf-quic-http.md
```diff
@@ -505,11 +505,11 @@ this limit are not guaranteed to be accepted.
...
+reading a response, it indicates that this response is no longer of interest.
+Implementations SHOULD cancel requests by terminating both directions of a
```
ianswett: Do we define terminating?

draft-ietf-quic-transport.md
dtikhonov: > when data on the stream is blocked due to flow control

Does this refer to the endpoint's or peer's inability to write?

draft-ietf-quic-transport.md
MikeBishop: Added to stream requirements, but now I'm not so sure that's correct.  Early-Data has to be added when the server "forwards a request prior to the completion of the TLS handshake".  That would mean the server really only needs a yes/no about the current state of the connection when the request is being forwarded.  If the handshake has completed, it doesn't matter that the request was originally early -- it's now retroactively validated as a non-replay by the successful handshake.

MikeBishop:
@janaiyengar and @martinthomson, I think this is mergeable from an HTTP standpoint.  Since it's technically a requirements change (requiring that functionality be exposed), it probably needs to be marked for a consensus call when you think it's ready.

draft-ietf-quic-transport.md
```diff
@@ -345,6 +345,33 @@ the relative priority of streams.  When deciding which streams to dedicate
...
+  ({{frame-reset-stream}}), even if the stream was already ended.
+
+On the receiving part of a stream, application protocols need to be able to:
+
+- read data
+- abort reading of the stream and request closure, possibly resulting in a
+  STOP_SENDING frame ({{frame-stop-sending}})
+
+Applications also need to be informed of state changes on streams, including
+when the peer has initiated, reset, or aborted reading on a stream; when new
```
martinthomson: List construction here implies that you can "reset reading on a stream" or that you can observe "initiation of reading on a stream".  You probably need to say "when the peer has opened or reset a stream; when a peer aborts reading a stream; when new ..."

draft-ietf-quic-transport.md
ianswett: Agreed, I think it's only yes/no about the current state of the connection.

draft-ietf-quic-transport.md
ianswett: Good point.  I still think the client needs to know when 1-RTT data can be sent, because it may want to send some data in 0-RTT but not other data.

draft-ietf-quic-transport.md
```diff
@@ -1101,6 +1131,42 @@ suggested structure:
...
+  client's resumption ticket and enable rejecting Early Data based on that
+  information.
+
+In either role, applications need to be able to:
+
+- configure minimum values for the initial number of permitted streams of each
+  type, as communicated in the transport parameters ({{transport-parameters}});
+- control resource allocation of various types, including flow control and the
+  number of permitted streams of each type;
+- identify whether the handshake has completed successfully or is still ongoing
```
ianswett: I think it'd be nice to clarify that a reason for this bullet is to know it can send data in 1-RTT.

draft-ietf-quic-transport.md
```diff
@@ -345,6 +345,33 @@ the relative priority of streams.  When deciding which streams to dedicate
...
+
+On the sending part of a stream, application protocols need to be able to:
+
+- write data, understanding when stream flow control credit
+  ({{data-flow-control}}) has successfully been reserved to send the written
+  data
+- end the stream (clean termination), resulting in a STREAM frame
+  ({{frame-stream}}) with the FIN bit set; and
+- reset the stream (abrupt termination), resulting in a RESET_STREAM frame
+  ({{frame-reset-stream}}), even if the stream was already ended.
```
mikkelfj: This is unclear - what does it mean that "the stream was already ended"? I would say it is fair for an API to issue a soft error if you try to RESET (abrupt termination) after having issued a FIN request (clean termination), or vice version. Certainly I would not want a QUIC implementation tracking a stream handle indefinitely in case some app decides to reset it eventually.

draft-ietf-quic-transport.md
mikkelfj: But why does an application need to be able to enable 0-RTT? Isn't 0-RTT optional at transport level, and more so at the app level?

draft-ietf-quic-transport.md
mikkelfj: An app should also be able to reject incoming connections, for example based on implementation specific criteria such as load or peer IP address.

draft-ietf-quic-transport.md
mikkelfj: Given that hanshake keys must now be held forever, how can it be indicated that a handshake is complete?

draft-ietf-quic-transport.md
MikeBishop: "End the stream" was the previous bullet, so I think it's reasonably clear what "already ended" should mean in this context.  If you look at the [sending stream states](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#rfc.section.3.1), this reflects the possibility to transition from "Data Sent" (stream has ended) to "Reset Sent" by sending a RESET_STREAM.

Based on the state diagram, if an implementation was already in the "Data Recvd" state, they wouldn't send a RESET_STREAM, as they're already in a terminal state. Surfacing this through the API would be a reasonable choice.  Likewise, there's no path from "Reset Sent" or "Reset Recvd" to "Data Sent"; I think attempting to write to or close a reset stream will always be an error.

draft-ietf-quic-transport.md
MikeBishop: An application needs to be able to control whether 0-RTT is offered; you're correct that this phrasing implies that 0-RTT support is mandatory, which wasn't the intention.

draft-ietf-quic-transport.md
MikeBishop: That seems like a worthwhile feature for a QUIC implementation to expose, but doesn't seem like one we should mandate every QUIC implementation must expose.

draft-ietf-quic-transport.md
MikeBishop: The phrases "handshake completes," "handshake is complete," "handshake has completed," etc. occur throughout the document.  If you believe that there can be no such concept, please open a separate issue and explain in detail.

draft-ietf-quic-transport.md
mikkelfj: So you mean to say that an application should be able to reset a stream at an arbitrary point after the stream has sent a FIN bit? If so, I don't think that is a good idea. In my view that would be an API error of invalid handle or invalid handle state. It's one thing to allow races between RESET and FIN but to require it forever is different.

draft-ietf-quic-transport.md
mikkelfj: I can agree to that.

draft-ietf-quic-transport.md
mikkelfj: I'm not objecting to this term being defined, but I do suspect there is an issue if the now infamous PR #3121 gets merged. But, as it hasn't yet, and there is pushback, I'll see where it goes. Defining exactly what is, or might be, wrong requires a deep analysis because one peer can consider a handshake complete without the other peer agreeing, and sometimes that might be fine while in other cases not so much, like path migration.

draft-ietf-quic-transport.md
MikeBishop: Rather than having this discussion on a closed PR, let's open a separate issue.

PR #2859, opened 2019-06-28
Delayed packets can't create state

### LINKED ISSUE #2148: Why would late-arriving packets create state?
```
Disposing of connection state prior to the end of the closing or draining period
could cause delayed or reordered packets to be handled poorly.  Endpoints that
have some alternative means to ensure that late-arriving packets on the
connection do not create QUIC state, such as those that are able to close the
UDP socket, MAY use an abbreviated draining period which can allow for faster
resource recovery.  Servers that retain an open socket for accepting new
connections SHOULD NOT exit the closing or draining period early.
```

I'm not following why in general this *would* create state. If the
handshake has occurred > MSL ago, and you just refuse to create state
on short packets with unknown CIDs s, why isn't that enough?

MikeBishop:
Fixes #2148.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2199,12 +2199,12 @@ the closing period would have ended.  In other words, the endpoint can use the
 same end time, but cease retransmission of the closing packet.
 
 Disposing of connection state prior to the end of the closing or draining period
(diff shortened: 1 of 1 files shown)


martinthomson:
FWIW, the state that was implied was the creation of a new connection.  That's not realistic in QUICv1 however, so this is good.

PR #2864, opened 2019-07-01
Explicitly allow discard of packets <40 bytes

### LINKED ISSUE #2153: Increase the odds of a stateless reset?
```
An endpoint can increase the odds that a packet will trigger a Stateless Reset
if it cannot be processed by padding it to at least 38 bytes.
```

Why does this increase the odds? It seems like the test here is deterministic.

martinthomson:
The stateless reset text implied a lot, but never really explicitly said
what to do when going below the minimum recommended size.

This is a design change to an issue marked editorial.  Enjoy.

Closes #2153.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2510,7 +2510,8 @@ Reset in response, which could lead to an infinite exchange.
 An endpoint MUST ensure that every Stateless Reset that it sends is smaller than
 the packet which triggered it, unless it maintains state sufficient to prevent
 looping.  In the event of a loop, this results in packets eventually being too
-small to trigger a response.
+small to trigger a response.  An endpoint MAY choose not to send a Stateless
+Reset in response to a packet that is smaller than 40 bytes.
 
 An endpoint can remember the number of Stateless Reset packets that it has sent
 and stop generating new Stateless Reset packets once a limit is reached.  Using
@@ -2524,9 +2525,6 @@ Stateless Reset.  Conversely, refusing to send a Stateless Reset in response to
 a small packet might result in Stateless Reset not being useful in detecting
 cases of broken connections where only very small packets are sent; such
 failures might only be detected by other means, such as timers.
-
-An endpoint can increase the odds that a packet will trigger a Stateless Reset
-if it cannot be processed by padding it to at least 40 bytes.
 
 
 # Error Handling {#error-handling}


draft-ietf-quic-transport.md
```diff
@@ -2525,9 +2526,6 @@ a small packet might result in Stateless Reset not being useful in detecting
 cases of broken connections where only very small packets are sent; such
 failures might only be detected by other means, such as timers.
 
-An endpoint can increase the odds that a packet will trigger a Stateless Reset
```
marten-seemann: Why did you remove this sentence?

martinthomson: See #2153.  Or were you asking for it to be retained?

marten-seemann: Yes. I think it's a good idea to always send a packet padded to > 40 bytes after quiescence, so you'd be able to quickly recover by receiving the stateless reset.

DavidSchinazi:
This PR adds a recommendation involving the number 40 without explaining where that number comes from. It would be nice to include a reference to the section that explains the math.

That said, I don't think this number should depend on the maximum connection ID length, [as discussed here](https://github.com/quicwg/base-drafts/issues/2770#issuecomment-502763574). But feel free to tell me that this PR is not the place to discuss that.

martinthomson:
If you want to talk about the idea that the connection ID might change in length, that's a separate issue, which you should feel free to open.  I am fairly sure that I can show that this isn't a meaningful insight, but happy to discuss it there.

DavidSchinazi:
@martinthomson filed #2869.

draft-ietf-quic-transport.md
```diff
@@ -2510,7 +2510,8 @@ Reset in response, which could lead to an infinite exchange.
 An endpoint MUST ensure that every Stateless Reset that it sends is smaller than
 the packet which triggered it, unless it maintains state sufficient to prevent
 looping.  In the event of a loop, this results in packets eventually being too
-small to trigger a response.
+small to trigger a response.  An endpoint MAY choose not to send a Stateless
+Reset in response to a packet that is smaller than 40 bytes.
```
ianswett: The existing text was confusing, but I don'd see how this text is relevant to looping?

MikeBishop: This provides the end condition to the loop.

draft-ietf-quic-transport.md
MikeBishop: It might be nice to explicitly mention the packet after quiescence as part of this advice, and/or advise sending at least one 40+ byte packet in each flight.

martinthomson:
We've moved on here, allowing smaller outbound packets.

Issue #2858, opened 2019-06-28
Signal error upon receipt of multiple trailing HEADERS


bencebeky:
According to https://quicwg.org/base-drafts/draft-ietf-quic-http.html#request-response, a  bidirectional HTTP stream can have at least one HEADERS frame (exactly one if the stream is client-initiated), zero or more DATA frames, then zero or one HEADERS frame, in this order.  Handling the receipt of frames that does not conform to this requirement is inconsistent:

(1) If a DATA frame is received before any other frame, that's an error:
> If a DATA frame is received before a HEADERS frame [...], the recipient MUST respond with a connection error of type HTTP_UNEXPECTED_FRAME [...].

(2) If the trailing HEADERS frame is followed by another HEADERS frame, that is to be discarded:
> Trailing header fields are carried in an additional HEADERS frame following the body. Senders MUST send only one HEADERS frame in the trailers section; receivers MUST discard any subsequent HEADERS frames.

(3) There is no guidance on how to handle a DATA frame after the trailing headers.

I suggest that every violation to this requirement should be an error.  (I do not feel strongly whether a connection error like (1) is handled right now, or a stream error.  The implementation I am contributing to will promote most stream errors to connection errors regardless.)

Note that (2) comes from the technical limitation that header blocks needed to be processed, which is not the case any longer with QPACK.  It just seems like that when the requirement to decode the header block was removed at https://github.com/quicwg/base-drafts/commit/404c5b8ee838042f7269574b608d1b15e2a25fd9, adding the requirement to signal an error was not in scope of that commit.

LPardue:
I agree with @bencebeky .

> If a DATA frame is received before a HEADERS frame [...], the recipient MUST respond with a connection error of type HTTP_UNEXPECTED_FRAME [...].

If you read that requirement a certain way, then any use of trailers will blow up the connection :) I think we could rewrite this text to be clearer about when DATA frames are expected to be received.

MikeBishop:
Good points.  Let's continue tightening down the screws.  :-)

MikeBishop:
Closed in #2867.

PR #2867, opened 2019-07-02
Tighten HEADERS and DATA frame order requirements.

bencebeky:
* Clarify HTTP message definition.
* Require the connection to be closed upon receipt of frames in
  incorrect order.
* Refer to HTTP messages in push stream section.

This addresses https://github.com/quicwg/base-drafts/issues/2858.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -415,6 +415,10 @@ before, after, or interleaved with the frames of a response message. These
 PUSH_PROMISE frames are not part of the response; see {{server-push}} for more
 details.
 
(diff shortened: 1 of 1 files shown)


bencebeky:
> This text was already somewhat lacking by not discussing the possibility of extension frames being interleaved; stating that the push stream consists of "responses" rather than "frames" extends this shortcoming to the push stream section.
> 
> I'm open to fixing this in a separate PR, if you prefer, though.

Good catch.  I added a paragraph to explicitly allow unknown frames to the message exchange section.  I also modified the wording in the push section to be more permissive of interleaved frames ("by" instead of "consists of").

LPardue:
LGTM

PR #2878, opened 2019-07-09
Initial secrets do not change after Retry

### LINKED ISSUE #2823: Do Initial secrets change after Retry packet?

ianswett:
Alternate PR to #2870 

Fixes #2823

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -805,8 +805,8 @@ packets even where the TLS versions offered do not include TLS 1.3.
 
 The secrets used for protecting Initial packets do not change during the
 connection, even after receiving a Retry.  A server that sends a Retry
-therefore needs to either remember Initial protection keys or
-save them in the Retry token.
+therefore needs to either remember Initial protection keys or save them
+in the Retry token.
 
 {{test-vectors-initial}} contains test vectors for the initial packet
 encryption.
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -805,10 +805,12 @@ The HKDF-Expand-Label function defined in TLS 1.3 MUST be used for Initial
 packets even where the TLS versions offered do not include TLS 1.3.
 
 The secrets used for protecting Initial packets do not change during the
-connection, even though the destination connection ID in client Initial packets
-changes after receiving a Retry.  A server that sends a Retry
-therefore needs to either remember the original connection ID
-or save the original connection ID in the Retry token.
+connection, even though the destination connection ID in client Initial
+packets changes after receiving a Retry.  A server that sends a Retry
+therefore needs to either remember the original connection ID or save
+the original connection ID in the Retry token.  Because the initial
+connection ID is included in the server's transport parameters, the only
+difference is when in the packet processing this connection ID is utilized.
 
 {{test-vectors-initial}} contains test vectors for the initial packet
 encryption.
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -805,7 +805,8 @@ The HKDF-Expand-Label function defined in TLS 1.3 MUST be used for Initial
 packets even where the TLS versions offered do not include TLS 1.3.
 
 The secrets used for protecting Initial packets do not change during the
-connection, even after receiving a Retry.  A server that sends a Retry
+connection, even though the destination connection ID in client Initial packets
+changes after receiving a Retry.  A server that sends a Retry
 therefore needs to either remember the original connection ID and Initial
 protection keys or save the original connection ID in the Retry token.
 
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -807,7 +807,7 @@ packets even where the TLS versions offered do not include TLS 1.3.
 The secrets used for protecting Initial packets do not change during the
 connection, even though the destination connection ID in client Initial packets
 changes after receiving a Retry.  A server that sends a Retry
-therefore needs to either remember the original connection ID and Initial
+therefore needs to either remember the original connection ID
 protection keys or save the original connection ID in the Retry token.
 
 {{test-vectors-initial}} contains test vectors for the initial packet
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -805,8 +805,8 @@ packets even where the TLS versions offered do not include TLS 1.3.
 
 The secrets used for protecting Initial packets do not change during the
 connection, even after receiving a Retry.  A server that sends a Retry
-therefore needs to either remember Initial protection keys or save them
-in the Retry token.
+therefore needs to either remember the original connection ID and Initial
+protection keys or save the original connection ID the Retry token.
 
 {{test-vectors-initial}} contains test vectors for the initial packet
 encryption.
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -809,14 +809,6 @@ connection, even after receiving a Retry.
 {{test-vectors-initial}} contains test vectors for the initial packet
 encryption.
 
-Note:
-
-: The Destination Connection ID is of arbitrary length, and it could be zero
-  length if the server sends a Retry packet with a zero-length Source Connection
-  ID field.  In this case, the Initial keys provide no assurance to the client
-  that the server received its packet; the client has to rely on the exchange
-  that included the Retry packet for that property.
-
 
 ## AEAD Usage {#aead}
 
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-tls.md
```diff
@@ -804,6 +803,9 @@ modifying the contents of packets from future versions.
 The HKDF-Expand-Label function defined in TLS 1.3 MUST be used for Initial
 packets even where the TLS versions offered do not include TLS 1.3.
 
+The secrets used for protecting Initial packets do not change during the
+connection, even after receiving a Retry.
+
 {{test-vectors-initial}} contains test vectors for the initial packet
 encryption.
```
martinthomson: Just down here (vv) there is a note that needs to be removed also.

martinthomson:
I don't think that's a concern, because the server is still required to echo the ODCID and the client should check it.  That change is entered into the handshake transcript, so it can't be spoofed.

marten-seemann:
@martinthomson This PR removes the transport parameter. Nothing is entered into the handshake transcript.

martinthomson:
Oh now I see why. Ian removed the transport parameter after I reviewed it. In an update labelled 
"Update draft-ietf-quic-transport.md".

ianswett:
Thanks for walking through that attack Marten, I was under the impression it was no longer necessary, but I'll revert that change and add the TP back.

draft-ietf-quic-tls.md
```diff
@@ -804,17 +803,12 @@ modifying the contents of packets from future versions.
 The HKDF-Expand-Label function defined in TLS 1.3 MUST be used for Initial
 packets even where the TLS versions offered do not include TLS 1.3.
 
+The secrets used for protecting Initial packets do not change during the
+connection, even after receiving a Retry.
```
martinthomson: Is it worth adding some advice on how to implement this?  It's not entirely obvious.

```suggestion
connection, even after receiving a Retry.  A server that sends a Retry
therefore needs to either remember Initial protection keys or
save them in the Retry token.
```

ianswett:
Thanks for the suggestion Martin, accepted.

draft-ietf-quic-tls.md
```diff
@@ -804,17 +803,14 @@ modifying the contents of packets from future versions.
 The HKDF-Expand-Label function defined in TLS 1.3 MUST be used for Initial
 packets even where the TLS versions offered do not include TLS 1.3.
 
+The secrets used for protecting Initial packets do not change during the
+connection, even after receiving a Retry.  A server that sends a Retry
+therefore needs to either remember Initial protection keys or save them
```
DavidSchinazi: Shouldn't this just say that the server already needs to have access to the original destination CID because of the `original_connection_id` transport parameter? I don't think the server needs to remember/encode initial keys on top of that

draft-ietf-quic-tls.md
```diff
@@ -805,8 +805,8 @@ packets even where the TLS versions offered do not include TLS 1.3.
 
 The secrets used for protecting Initial packets do not change during the
 connection, even after receiving a Retry.  A server that sends a Retry
-therefore needs to either remember Initial protection keys or save them
-in the Retry token.
+therefore needs to either remember the original connection ID and Initial
+protection keys or save the original connection ID the Retry token.
```
DavidSchinazi: typo: connection ID *in* the Retry token.

draft-ietf-quic-tls.md
```diff
@@ -771,8 +771,7 @@ TLS 1.3 (see {{initial-secrets}}).
 ## Initial Secrets {#initial-secrets}
 
 Initial packets are protected with a secret derived from the Destination
-Connection ID field from the client's first Initial packet of the
-connection. Specifically:
+Connection ID field from the client's Initial packet. Specifically:
```
MikeBishop: Why this change, when it removes the precision about *which* Initial packet?

ianswett: I can't remember why I changed that, reverted.

draft-ietf-quic-tls.md
```diff
@@ -804,17 +804,14 @@ modifying the contents of packets from future versions.
 The HKDF-Expand-Label function defined in TLS 1.3 MUST be used for Initial
 packets even where the TLS versions offered do not include TLS 1.3.
 
+The secrets used for protecting Initial packets do not change during the
+connection, even after receiving a Retry.  A server that sends a Retry
```
```suggestion
connection, even though the destination connection ID in client Initial packets
changes after receiving a Retry.  A server that sends a Retry
```

draft-ietf-quic-tls.md
```diff
@@ -804,17 +804,14 @@ modifying the contents of packets from future versions.
...
+therefore needs to either remember the original connection ID and Initial
+protection keys or save the original connection ID in the Retry token.
```
```suggestion
or save the original connection ID in the Retry token.
```

draft-ietf-quic-tls.md
```diff
@@ -804,17 +804,14 @@ modifying the contents of packets from future versions.
 The HKDF-Expand-Label function defined in TLS 1.3 MUST be used for Initial
 packets even where the TLS versions offered do not include TLS 1.3.
 
+The secrets used for protecting Initial packets do not change during the
+connection, even after receiving a Retry.  A server that sends a Retry
+therefore needs to either remember the original connection ID and Initial
```
martinthomson: No strict need to keep Initial (packet) protection keys.

```suggestion
therefore needs to either remember the original connection ID
```

MikeBishop:
#2926 implies a good point:  After this change, the "Initial DCID" and the "original DCID" are always the same.  We should collapse the terms as part of this PR or in a follow-up if we take it.

draft-ietf-quic-tls.md
ianswett: Good point, MT suggested some simplifications to the text and I added a sentence clarifying that this is not a new requirement, it just changes when the connection ID is used.

draft-ietf-quic-tls.md
```diff
@@ -805,10 +805,12 @@ The HKDF-Expand-Label function defined in TLS 1.3 MUST be used for Initial
...
-connection, even though the destination connection ID in client Initial packets
-changes after receiving a Retry.  A server that sends a Retry
-therefore needs to either remember the original connection ID
-or save the original connection ID in the Retry token.
+connection, even though the destination connection ID in client Initial
+packets changes after receiving a Retry.  A server that sends a Retry
+therefore needs to either remember the original connection ID or save
+the original connection ID in the Retry token.  Because the initial
+connection ID is included in the server's transport parameters, the only
+difference is when in the packet processing this connection ID is utilized.
```
martinthomson: This reads like justification for a change, rather than static text.  How about:

> The initial connection ID is needed by a server to reconstruct packet protection keys and so that it can produce the correct value for the original_connection_id transport parameter.

And are we using "initial" or "original" here?  I went with original on the basis that it distinguished this from Initial packets, but it's not clear to be that that distinction is necessary, or even a good idea.  After all, this connection ID is what is driving packet protection for those packets; the linkage might be useful.

MikeBishop: At the time, we were trying to avoid confusion between the _original_ DCID, which generated the keys, and the DCID in the current Initial.  It was a useful distinction in the old design.  With this change, I think that distinction is no longer useful.

PR #2880, opened 2019-07-09
Articulate error code principles

### LINKED ISSUE #2388: Articulate principles for definition of error codes
We don't really have any real principles that we agree on for deciding what error codes we are describing.

Proposal:

1. if the error carries distinct semantics (like stream rejection in HTTP), then it gets a new error code

2. if the error is frequent or particularly significant, then it gets a new error code

3. otherwise, target a more generic error code that identifies the broad area of the protocol

4. finally, if there is no more specific applicable error code, use PROTOCOL_VIOLATION

martinthomson:
This just writes down what was proposed in #2388.

Closes #2388.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2558,7 +2558,12 @@ errors can be isolated to a single stream (see {{stream-errors}}).
 
 The most appropriate error code ({{error-codes}}) SHOULD be included in the
 frame that signals the error.  Where this specification identifies error
-conditions, it also identifies the error code that is used.
+conditions, it also identifies the error code that is used; though these are
+worded as requirements, different implementation strategies might lead to
+different errors being reported.  In particular, an endpoint MAY use any
+applicable error code when it detects an error condition; a generic error code
+(such as PROTOCOL_VIOLATION or INTERNAL_ERROR) can always be used in place of
+specific error codes.
 
 A stateless reset ({{stateless-reset}}) is not suitable for any error that can
 be signaled with a CONNECTION_CLOSE or RESET_STREAM frame.  A stateless reset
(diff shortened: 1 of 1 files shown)


mikkelfj:
Looks fine, but what about some guidance on stream vs connection errors? And also the principle of nearly always choosing a hard protocol violation error, or a more granular error with same effect?

martinthomson:
Stream vs. connection is something that application protocols (like HTTP) need to worry about.  The transport only concerns itself with connection errors.  The idea that you pick the most helpful error is not something that we need to codify (in my view).

mikkelfj:
> Stream vs. connection is something that application protocols (like HTTP) need to worry about. 

Isn't that worth mentioning? Or do we already have such text?

LPardue:
> Isn't that worth mentioning? Or do we already have such text?

This is already discussed in Section 11, which makes me wonder if it would make more sense for text on this PR to be placed there too.

mikkelfj:
>  The idea that you pick the most helpful error is not something that we need to codify (in my view).

This I do not disagree with. I was merely suggesting that we explain why, as a general rule, a hard error is preferred as a design principle, not the nature of the error message which I do think should have a large degree of freedom.

martinthomson:
Maybe there is a followup needed for the HTTP draft.  If the principles are generally the same, we can copy the text and then add some more to cover the idea that we make most errors fatal to the connection rather than the stream.

ianswett:
I like Lucas suggestion of adding this text to the existing and already fairly complete text in section 11.

martinthomson:
I moved the second paragraph up to the error handling section, but the text about how codes are selected really belongs with the codes themselves.

draft-ietf-quic-transport.md
```diff
@@ -5374,6 +5379,14 @@ CRYPTO_ERROR (0x1XX):
 
 See {{iana-error-codes}} for details of registering new error codes.
 
+In defining these error codes, several principles are applied.  Error conditions
+that might require specific action on the part of a recipient are giving unique
```
LPardue: s/giving/given

martinthomson: I'm almost surprised that I can update your comment, turn it into a suggestion, then cause a commit to be created under your name.  The power is going to my head...

LPardue: I'd have made the suggestion myself but I'm on a mobile device that hides that widget. Good to know we have a workflow that makes my life easier ;)

PR #2887, opened 2019-07-09
Get test vectors in place for -23

martinthomson:
New salt, updated for invariants change.


martinthomson:
This should close #2887 as well..

PR #2888, opened 2019-07-10
Firm up language around stream termination

### LINKED ISSUE #2884: Application should be allowed to respond to STOP_SENDING instead of the endpoint
Section 11.2 of the transport draft reads:

>RESET_STREAM MUST be instigated by the protocol using QUIC, either directly or through the receipt of a STOP_SENDING frame from a peer. RESET_STREAM carries an application error code. Resetting a stream without knowledge of the application protocol could cause the protocol to enter an unrecoverable state. Application protocols might require certain streams to be reliably delivered in order to guarantee consistent state between endpoints.

This would suggest that the transport implementation itself should not be deciding to send RESET_STREAMs on its own, rather it is a matter of the application protocol using QUIC. However, it is somewhat weaselly about STOP_SENDING. My original interpretation of this text was that the implementation should expose the fact that a STOP_SENDING has been received, but it is up to the application protocol to actually reset the stream. This interpretation is not consistent with section 3.5, Solicited State Transitions: 

> A STOP_SENDING frame requests that the receiving endpoint send a RESET_STREAM frame. An endpoint that receives a STOP_SENDING frame MUST send a RESET_STREAM frame if the stream is in the Ready or Send state. If the stream is in the Data Sent state and any outstanding data is declared lost, an endpoint SHOULD send a RESET_STREAM frame in lieu of a retransmission.
I believe the only reasonable way to interpret this is that the transport itself should respond to the STOP_SENDING by issuing a RESET_STREAM.

Requiring that the QUIC implementation itself reset the stream seems to go against the philosophy that the application protocol should have complete control over resets. It's true that in the case of receiving a STOP_SENDING, the protocol on the other side is asking for a RESET_STREAM, but the entity being solicited should be its peer protocol implementation, not the QUIC implementation. I believe we can fix this by altering the normative text requiring sending a RESET_STREAM, instead prescribing that an application protocol ought to handle STOP_SENDING by eventually fulfilling the solicited reset.

martinthomson:
This was a little woolly.  Hopefully this is crisper.  Basically, the
application protocol is what asks for stream reset, either directly, or
by sending STOP_SENDING.

I added a recommendation for application protocols to define handling
for prematurely cancelled streams.  That is, I believe, just editorial,
even though it uses 2119 language, because we already do that.

Closes #2884.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -650,7 +650,7 @@ upon receipt.
 
 A STOP_SENDING frame requests that the receiving endpoint send a RESET_STREAM
 frame.  An endpoint that receives a STOP_SENDING frame MUST send a RESET_STREAM
-frame if the stream is in the Ready or Send state. If the stream is in the Data
+frame if the stream is in the Ready or Send state.  If the stream is in the Data
 Sent state and any outstanding data is declared lost, an endpoint SHOULD send a
 RESET_STREAM frame in lieu of a retransmission.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2601,12 +2601,17 @@ connection in a recoverable state, the endpoint can send a RESET_STREAM frame
...
-through the receipt of a STOP_SENDING frame from a peer.  RESET_STREAM carries
-an application error code.  Resetting a stream without knowledge of the
-application protocol could cause the protocol to enter an unrecoverable state.
-Application protocols might require certain streams to be reliably delivered in
-order to guarantee consistent state between endpoints.
+RESET_STREAM MUST be instigated by the protocol using QUIC.  RESET_STREAM
+carries an application error code.  Only the application protocol is able to
+cause a stream to be terminated.  A local instance of the application protocol
+uses a direct API call and a remote instance uses the STOP_SENDING frame, which
+triggers an automatic RESET_STREAM.
```
RyanAtGoogle: Hm. The second sentence suggests that the determination of the application error code in the RESET_STREAM is the responsibility of the application. Fair enough. But in this last sentence, it sounds like the RESET_STREAM is generated by the transport (or some such) in response to STOP_SENDING. Also fair enough. But how does the transport determine the application error code.

I think the behavior here is fine, but the text feels... wooly, as you said. I wonder if it might be easier to read if we treated the STOP_SENDING case as more of an exception, than trying to make it part of the general "only the app resets streams" behavior.  *shrug*

MikeBishop: The guidance at line 657 says:
> An endpoint SHOULD copy the error code from the STOP_SENDING frame to the
> RESET_STREAM frame it sends, but MAY use any application error code.  The
> endpoint that sends a STOP_SENDING frame MAY ignore the error code carried in
> any RESET_STREAM frame it receives.

HTTP allows in one instance for the server to use a different error code if the transport implementation supports that, but everything's fine if the transport just reflects the code before the local application knows about it.

PR #2893, opened 2019-07-12
Remove DoS vector for spoofed connection migration

### LINKED ISSUE #2342: Spoofed connection migration as a DoS vector
Section 9 "Connection Migration" has this text:

> An endpoint which has sent this [disable_migration] transport parameter, but detects that a peer has nonetheless migrated to a different network MAY treat this as a connection error of type INVALID_MIGRATION.

How can an endpoint distinguish "a peer has nonetheless migrated to a different network" vs "a man-on-a-side attacker has resent/raced a packet from a different IP"?  If this hinges on path validation, an insidious attacker can actually own the new IP and proxy packets with PATH_CHALLENGE back to the client and resend packets with PATH_CHALLENGE_RESPONSE from its IP.

### LINKED ISSUE #2389: Be clearer about the purpose of disable_migration
The mechanism for disabling migration might be clear, but it isn't clear what problems it addresses.  Some more clarity on that would be useful.

What I understand this to be intended for is server deployments that can't guarantee that packets from a new source address will be routed to the same server instance.

erickinnear:
Remove connection error response to a packets from a different source address when an endpoint has sent `disable_migration`, clarify an endpoints options when prohibited packets are received, rename `disable_migration` to `disable_active_migration`, and remove `INVALID_MIGRATION` error code.

Closes #2342, closes #2389.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1816,10 +1816,7 @@ parameter, but detects that a peer has nonetheless migrated to a different
 network MUST either drop the incoming packets on that path without generating a
 stateless reset or proceed with path validation and allow the peer to migrate.
 This helps to prevent third parties in the network causing connections to close
-due to spoofed addresses or other manipulation of observed traffic. An endpoint
-MUST NOT actively initiate migration if its peer supplies a zero-length
-connection ID as packets without a Destination Connection ID cannot be
-attributed to a connection based on address tuple.
+due to spoofed addresses or other manipulation of observed traffic.
 
 Not all changes of peer address are intentional, or active, migrations. The peer
 could experience NAT rebinding: a change of address due to a middlebox, usually
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1815,8 +1815,9 @@ parameter during the handshake. An endpoint which has sent this transport
 parameter, but detects that a peer has nonetheless migrated to a different
 network MUST either drop the incoming packets on that path without generating a
 stateless reset or proceed with path validation and allow the peer to migrate.
-This helps to prevent third parties in the network causing connections to close
-due to spoofed addresses or other manipulation of observed traffic.
+Generating a stateless reset or closing the connection would allow third parties
+in the network to cause connections to close by spoofing or otherwise
+manipulating observed traffic.
 
 Not all changes of peer address are intentional, or active, migrations. The peer
 could experience NAT rebinding: a change of address due to a middlebox, usually
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -4303,10 +4303,11 @@ max_ack_delay (0x000b):
 disable_active_migration (0x000c):
 
 : The disable active migration transport parameter is included if the endpoint
-  does not support active connection migration ({{migration}}). Peers of an endpoint that sets
-  this transport parameter MUST NOT send any packets, including probing packets
-  ({{probing}}), from a local address or port other than that used to perform
-  the handshake.  This parameter is a zero-length value.
+  does not support active connection migration ({{migration}}). Peers of an
+  endpoint that sets this transport parameter MUST NOT send any packets,
+  including probing packets ({{probing}}), from a local address or port other
+  than that used to perform the handshake.  This parameter is a zero-length
+  value.
 
 preferred_address (0x000d):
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1809,19 +1809,24 @@ The design of QUIC relies on endpoints retaining a stable address for the
...
+initiating migration, if the peer sent the `disable_active_migration` transport
+parameter during the handshake. An endpoint which has sent this transport
+parameter, but detects that a peer has nonetheless migrated to a different
+network MUST either drop the incoming packets on that path without generating a
+stateless reset or proceed with path validation and allow the peer to migrate.
+This helps to prevent third parties in the network causing connections to close
+due to spoofed addresses or other manipulation of observed traffic. An endpoint
+MUST NOT actively initiate migration if its peer supplies a zero-length
+connection ID as packets without a Destination Connection ID cannot be
+attributed to a connection based on address tuple.
```
martinthomson: With the change in #2851, we don't need this last sentence.  The point there is that source addresses aren't allowed to be used in routing of packets, so we don't need to worry about forbidding migration in that case.

erickinnear: Good point! We cover this area much more thoroughly elsewhere, I'll remove this.

erickinnear: Removed in 5a7293d

erickinnear:
Re: The value of this TP, I tend to agree with @martinthomson, pretty much all of the machinery must be implemented whether or not the TP is sent. But agreed - let's discuss elsewhere, since I think this particular change is in pretty good shape now.

draft-ietf-quic-transport.md
```diff
@@ -1809,19 +1809,21 @@ The design of QUIC relies on endpoints retaining a stable address for the
...
-allocating a new outgoing port or even a new outgoing IP address for a flow.  An
-endpoint MUST perform path validation ({{migrate-validate}}) if it detects any
-change to a peer's address, unless it has previously validated that address.
+An endpoint also MUST NOT send packets from a different local address, actively
+initiating migration, if the peer sent the `disable_active_migration` transport
+parameter during the handshake. An endpoint which has sent this transport
+parameter, but detects that a peer has nonetheless migrated to a different
+network MUST either drop the incoming packets on that path without generating a
+stateless reset or proceed with path validation and allow the peer to migrate.
+This helps to prevent third parties in the network causing connections to close
```
ianswett: This = Not generating a stateless reset?  If so, how about "Generating a stateless reset would allow third parties in the network to cause a connection to close by spoofing or other manipulation of observed traffic."?

erickinnear: Changed in e7d2297, thanks!

igorlord:
I think PR is fine, right I would had added an explicit explanation of when a server may consider sending thing TP (other paths are likely to either not work or be very inefficient) and what a client that needs to migrate to a new path should do (recheck DNS from the new path and reconnect).

pravb:
The primary use case for disable_active_migration is that load balancing infrastructure cannot support any changes to the 4-tuple, intentional or otherwise. So in this case if a peer still changes its tuples, the packets may hit a a new server endpoint that doesn't have any existing state, and will generate a stateless reset? 

I'd like to hold consensus on this issue until this is clarified in the text.

erickinnear:
@pravb I think we got to the point where either (a) the packets miraculously get to the right place and then the endpoint can choose to either honor them or drop them or (b) they don't, in which case that's sad, but that's what happens when your infrastructure doesn't handle NAT rebinding or any other changes to the 4-tuple.

In neither case do you want to close the connection with an error, in case someone else has injected a packet with a different source address.

Note https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#rfc.section.21.8
>21.8. Stateless Reset Oracle
>Stateless resets create a possible denial of service attack analogous to a TCP reset injection. This attack is possible if an attacker is able to cause a stateless reset token to be generated for a connection with a selected connection ID. An attacker that can cause this token to be generated can reset an active connection with the same connection ID.
>
>If a packet can be routed to different instances that share a static key, for example by changing an IP address or port, then an attacker can cause the server to send a stateless reset. To defend against this style of denial service, endpoints that share a static key for stateless reset (see Section 10.4.2) MUST be arranged so that packets with a given connection ID always arrive at an instance that has connection state, unless that connection is no longer active.

Should you be in a situation where the client will accept your stateless reset for a connection that's active if someone spoofs a source address and the packet ends up on a server that has no state? If that can happen, that seems like we should open a new issue, since that's something that can happen at any time (although seems to me to be pretty much covered by the text in 21.8).

The point of this change is to be clear that the purpose of the TP is to instruct the endpoint sending the packets to not intentionally change its address or port, but that the server receiving the packets (if for whatever reason they still get there) needs to not close the connection if it sees that happening. It can drop the offending packets, or process them, but it's a problem if it does the previous behavior which was to close the connection.

pravb:
@erickinnear thanks for the detailed explanation. This looks good to me and I have no objections to consensus.

PR #2896, opened 2019-07-15
Move wasteful packet stuff to -transport

### LINKED ISSUE #2890: HTTP/2 in Peer Denial of Service?
Under the section _Peer Denial of Service_ of TLS draft, a referenced was made to HTTP/2.

https://github.com/quicwg/base-drafts/blame/master/draft-ietf-quic-tls.md#L1393

I was wonder if this is in fact referring to HTTP/3 or HTTP/2. Also curious why the section is labelled `{#useless}`

martinthomson:
Issue #2890 observes that this text mentions HTTP/2, which isn't
necessary.  In reviewing it, I noticed that this talks about messages
that do nothing, but not really about messages that do little (or that
can be reversed to have net zero effect).  So I expanded the text a
little.

Closes #2890.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1388,20 +1388,6 @@ attacker cannot forge them.  Put together, these defenses limit the level of
 amplification.
 
 
-## Peer Denial of Service {#useless}
-
-QUIC, TLS, and HTTP/2 all contain messages that have legitimate uses in some
-contexts, but that can be abused to cause a peer to expend processing resources
-without having any observable impact on the state of the connection.  If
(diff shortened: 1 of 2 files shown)


zz85:
Thanks for improving the section's clarity.

As a follow up question, when implementations decide that excessive packets is an attack, is there a recommended actionable (eg. sending reset, dropping the connection, do nothing) or would that be up for implementer's discretion?

martinthomson:
Maybe this needs to be stated explicitly, but there is no point in continuing discussion with an attacker.  So while it might be considered polite to send a connection error, it is also acceptable to just drop any further incoming packets.  I'll add that here.

PR #2898, opened 2019-07-15
HTTP Upgrade in H3 is silliness

### LINKED ISSUE #2889: HTTP/3 support for status 101(Switching Protocols)
HTTP/3 makes no explicit comment about this. HTTP/2 prohibits it. So I think that we might want to add equivalent text. https://tools.ietf.org/html/rfc7540#section-8.1.1

> HTTP/2 removes support for the 101 (Switching Protocols)
   informational status code ([RFC7231], Section 6.2.2).

   The semantics of 101 (Switching Protocols) aren't applicable to a
   multiplexed protocol.

LPardue:
Fixes #2889


MikeBishop:
Is it worth adding a reference to RFC 8441?  That's an alternative that covers what people looking for this functionality were probably trying to do.

LPardue:
That's a good point, let me craft something up and I'll update the PR soon.

LPardue:
I'll get an update out tomorrow

LPardue:
I'm actually not convinced that referencing 8441 will help clarify things here because the extension is H2 specific. No H3-equivalent exists (or at least, no H3 setting to enable the equivalent CONNECT behaviour exists).

bagder:
I agree. I think referencing 8441 might just add to the confusion - even if something *like* 8441 certainly can be introduced.

LPardue:
Some others have shown interest for defining/framing this capability in HTTP/3. I think that activity should consider ore broadly what text makes sense in the core document.

PR #2903, opened 2019-07-17
Remove Incorrect Sentence in Coalescing Section

### LINKED ISSUE #2900: Small clarification in 12.2 Coalescing Packets
In some testing with WinQuic, we noticed a small inconsistency in Section 12.2 of the transport spec that needs some clarification.
Section 12.2 says that if decryption fails, an implementation MUST attempt to process remaining packets in a coalesced UDP datagram. However, consider the following scenario:

1. An endpoint received a UDP datagram with a single short header packet, the first of the 1-RTT encryption level. This implicitly ACKs the end of the handshake and the receiver discards the handshake keys.
2. The sender then sends a coalesced datagram with a single long-header packet (an ACK for the handshake), and then a short header packet containing more 1-RTT data.

Since the receiver has thrown away the handshake keys, they are unable to decrypt the long header packet in the second datagram, and without being able to decrypt the header protection on the long header packet, they cannot ascertain the length, and are unable to find the short header packet that follows. As a result, the whole datagram is thrown away.

Can the "MUST" referenced above be clarified for which scenarios are/aren't supported?

nibanks:
Fixes #2900.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2661,8 +2661,7 @@ Initial ({{packet-initial}}), 0-RTT ({{packet-0rtt}}), and Handshake
 ({{packet-handshake}}) packets contain a Length field, which determines the end
 of the packet.  The length includes both the Packet Number and Payload
 fields, both of which are confidentiality protected and initially of unknown
-length. The length of the Payload field is learned once header protection is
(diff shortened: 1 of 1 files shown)


martinthomson:
I don't want to take this change. See the issue.

PR #2915, opened 2019-07-21
Up to two full-sized datagrams, not packets

ianswett:
Now that the PTO retransmits Initial and Handshake data, it's possible and likely desirable to bundle multiple QUIC packets in a single datagram.  This change makes it clear that is allowed.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -517,8 +517,8 @@ and larger thresholds increase loss detection delay.
 A Probe Timeout (PTO) triggers sending one or two probe datagrams when
 ack-eliciting packets are not acknowledged within the expected period of
 time or the handshake has not been completed.  A PTO enables a connection to
-recover from loss of tail packets or acknowledgements. The PTO algorithm used in QUIC
-implements the reliability functions of Tail Loss Probe
+recover from loss of tail packets or acknowledgements. The PTO algorithm used
+in QUIC implements the reliability functions of Tail Loss Probe
 {{?TLP=I-D.dukkipati-tcpm-tcp-loss-probe}} {{?RACK}}, RTO {{?RFC5681}} and
 F-RTO algorithms for TCP {{?RFC5682}}, and the timeout computation is based on
 TCP's retransmission timeout period {{?RFC6298}}.
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -517,7 +517,7 @@ and larger thresholds increase loss detection delay.
 A Probe Timeout (PTO) triggers sending one or two probe datagrams when
 ack-eliciting packets are not acknowledged within the expected period of
 time or the handshake has not been completed.  A PTO enables a connection to
-recover from loss of tail packets or acks. The PTO algorithm used in QUIC
+recover from loss of tail packets or ACKs. The PTO algorithm used in QUIC
 implements the reliability functions of Tail Loss Probe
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -514,7 +514,7 @@ and larger thresholds increase loss detection delay.
 
 ## Probe Timeout {#pto}
 
-A Probe Timeout (PTO) triggers sending one or two probe packets when
+A Probe Timeout (PTO) triggers sending one or two probe datagrams when
 ack-eliciting packets are not acknowledged within the expected period of
 time or the handshake has not been completed.  A PTO enables a connection to
 recover from loss of tail packets or acks. The PTO algorithm used in QUIC
```
```suggestion
recover from loss of tail packets or ACKs. The PTO algorithm used in QUIC
```

draft-ietf-quic-recovery.md
```diff
@@ -598,8 +598,8 @@ discarded.
 
 When a PTO timer expires, a sender MUST send at least one ack-eliciting packet
 as a probe, unless there is no data available to send.  An endpoint MAY send up
-to two ack-eliciting packets, to avoid an expensive consecutive PTO expiration
-due to a single packet loss.
+to two full sized datagrams containing ack-eliciting packets, to avoid an
+expensive consecutive PTO expiration due to a single lost datagram.
```
mikkelfj: Where is that limit of two specified? Here? Why?

ianswett: If you send 1, it might be lost.  Ideally, fewer is better, so the question of 1 or 2 was left to implementers.

mikkelfj: I mean, there is no limit to coalescing AFAIK, so the limit of two here seems arbitrary, albeit probably very sensible.
https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#rfc.section.12.2

draft-ietf-quic-recovery.md
```diff
@@ -514,10 +514,10 @@ and larger thresholds increase loss detection delay.
 
 ## Probe Timeout {#pto}
 
-A Probe Timeout (PTO) triggers sending one or two probe packets when
+A Probe Timeout (PTO) triggers sending one or two probe datagrams when
 ack-eliciting packets are not acknowledged within the expected period of
 time or the handshake has not been completed.  A PTO enables a connection to
-recover from loss of tail packets or acks. The PTO algorithm used in QUIC
+recover from loss of tail packets or ACKs. The PTO algorithm used in QUIC
```
```suggestion
recover from loss of tail packets or acknowledgements. The PTO algorithm used in QUIC
```

draft-ietf-quic-recovery.md
```diff
@@ -598,8 +598,8 @@ discarded.
 
 When a PTO timer expires, a sender MUST send at least one ack-eliciting packet
 as a probe, unless there is no data available to send.  An endpoint MAY send up
-to two ack-eliciting packets, to avoid an expensive consecutive PTO expiration
-due to a single packet loss.
+to two full sized datagrams containing ack-eliciting packets, to avoid an
```
```suggestion
to two full-sized datagrams containing ack-eliciting packets, to avoid an
```

draft-ietf-quic-recovery.md
ianswett: There's no limit of 2 packets in a datagram here and I don't think there's one elsewhere?

mikkelfj: > An endpoint MAY send up to two full-sized datagrams containing ack-eliciting packets,

I read that as it MAY NOT send more than two, but it MAY send less.

mikkelfj: Ah, I am confusing UDP datagrams and packets.

PR #2916, opened 2019-07-21
Move Generating Acknowledgements to Transport

### LINKED ISSUE #1860: ACK-only feedback loop in recovery instead of transport
Currently, it's the transport draft that mentions ACK-sending behaviour for ACK-only packets:

> Implementations MUST NOT generate packets that only contain ACK
   frames in response to packets which only contain ACK and PADDING
   frames.

and

> To avoid creating an indefinite
   feedback loop, an endpoint MUST NOT send an ACK frame in response to
   a packet containing only ACK or PADDING frames, even if there are
   packet gaps which precede the received packet.

and

> Because ACK frames are not sent in response to ACK-only packets, a
   receiver that is only sending ACK frames will only receive
   acknowledgements for its packets if the sender includes them in
   packets with non-ACK frames.  A sender SHOULD bundle ACK frames with
   other frames when possible.

This feels like it should be moved or at least mentioned again in the recovery document (e.g., in the current [4.4 Generating Acknowledgements](https://tools.ietf.org/html/draft-ietf-quic-recovery-15#section-4.4)).

ianswett:
Fixes #1860 

Removed text about a faster ACK in response to crypto packets because it's obsolete with #2646


draft-ietf-quic-transport.md
```diff
@@ -2880,6 +2880,28 @@ packet.
...
 
+
+## Generating Acknowledgements {#generating-acks}
+
+An acknowledgement SHOULD be sent immediately upon receipt of a second
+ack-eliciting packet. QUIC recovery algorithms do not assume the peer sends
+an ACK immediately when receiving a second ack-eliciting packet.
+
+In order to accelerate loss recovery and reduce timeouts, the receiver SHOULD
+send an immediate ACK after it receives an out-of-order packet. It could send
```
nibanks: What's the exact definition for `out-of-order`? If the last packet number received is `X`, are received packet numbers of `X+2` and `X-1` both considered out of order?

ianswett: Yes, both would be considered out of order.  I thought we already said to send an immediate ACK when the PN is not 1 larger than the largest processed packet number, but I can't find it.

draft-ietf-quic-recovery.md
```diff
@@ -376,26 +298,25 @@ min_rtt is set to the latest_rtt on the first sample in a connection, and to the
 lesser of min_rtt and latest_rtt on subsequent samples.
 
 An endpoint uses only locally observed times in computing the min_rtt and does
-not adjust for host delays reported by the peer ({{host-delay}}).  Doing so
-allows the endpoint to set a lower bound for the smoothed_rtt based entirely on
-what it observes (see {{smoothed-rtt}}), and limits potential underestimation
-due to erroneously-reported delays by the peer.
+not adjust for host delays reported by the peer.  Doing so allows the endpoint
```
janaiyengar: why remove this ref? Can you add the right ref in the transport doc?

draft-ietf-quic-transport.md
```diff
@@ -2947,6 +2997,25 @@ to unnecessarily retransmit some data.  Standard QUIC algorithms
 acknowledged.  Therefore, the receiver SHOULD repeatedly acknowledge newly
 received packets in preference to packets received in the past.
 
+### Measuring and Reporting Host Delay {#host-delay}
+
+An endpoint measures the delays intentionally introduced between when an
+ACK-eliciting packet is received and the corresponding acknowledgment is sent.
+The endpoint encodes this delay for the largest acknowledged packet in the
+Ack Delay field of an ACK frame (see Section 19.3). This allows the receiver
```
janaiyengar: change "Section 19.1" to a real reference

draft-ietf-quic-transport.md
```diff
@@ -2947,6 +2997,25 @@ to unnecessarily retransmit some data.  Standard QUIC algorithms
...
+of the ACK to adjust for any intentional delays, which is important for
+delayed acknowledgements, when estimating the path RTT. A packet might be
+held in the OS kernel or elsewhere on the host before being processed.
+An endpoint SHOULD NOT include these unintentional delays when populating
+the Ack Delay field in an ACK frame.
+
+An endpoint MUST NOT excessively delay acknowledgements of ack-eliciting
+packets.  The maximum ack delay is communicated in the max_ack_delay transport
+parameter; see Section 18.1.  max_ack_delay implies an explicit contract:
```
janaiyengar: change to a real reference

draft-ietf-quic-transport.md
```diff
@@ -2880,6 +2880,28 @@ packet.
 expectations about what implementations do with packets that have errors after
 valid frames? -->
 
+
+## Generating Acknowledgements {#generating-acks}
+
+An acknowledgement SHOULD be sent immediately upon receipt of a second
+ack-eliciting packet. QUIC recovery algorithms do not assume the peer sends
+an ACK immediately when receiving a second ack-eliciting packet.
```
martinthomson: This is an odd lead-in.  Do you want to start by saying a few more general things?

1. endpoints send acknowledgements for all packets they receive and process
2. loss recovery doesn't depend on immediate acknowledgment, but performance can be improved by acknowledging immediately in some cases
3. packets that aren't ack-eliciting are only acknowledged when sending an ACK for other reasons
4. there are recommendations to send immediate acknowledgments in a dazzling array of cases
   a. when receiving the second ack-eliciting packet
   b. when it appears as though there is loss or reordering, though not for every packet
   c. when new CRYPTO data is available to send (aka the incoming CRYPTO flight is outstanding)
   d. ...
5. When sending a packet for any reason, it's generally a good idea to send an ACK if one has not been sent recently.

draft-ietf-quic-transport.md
```diff
@@ -2924,6 +2946,34 @@ needing acknowledgement are received.  The sender can use the receiver's
...
+frame may be saved.  When a packet containing an ACK frame is acknowledged, the
+receiver can stop acknowledging packets less than or equal to the largest
+acknowledged in the sent ACK frame.
+
+In cases without ACK frame loss, this algorithm allows for a minimum of 1 RTT
+of reordering. In cases with ACK frame loss and reordering, this approach does
+not guarantee that every acknowledgement is seen by the sender before it is no
+longer included in the ACK frame. Packets could be received out of order and
+all subsequent ACK frames containing them could be lost. In this case, the
+loss recovery algorithm may cause spurious retransmits, but the sender will
```
```suggestion
loss recovery algorithm could cause spurious retransmits, but the sender will
```

draft-ietf-quic-transport.md
```suggestion
Ack Delay field of an ACK frame (see {{frame-ack}}). This allows the receiver
```

draft-ietf-quic-transport.md
```diff
@@ -2947,6 +2997,25 @@ to unnecessarily retransmit some data.  Standard QUIC algorithms
...
 received packets in preference to packets received in the past.
 
+### Measuring and Reporting Host Delay {#host-delay}
+
+An endpoint measures the delays intentionally introduced between when an
+ACK-eliciting packet is received and the corresponding acknowledgment is sent.
+The endpoint encodes this delay for the largest acknowledged packet in the
+Ack Delay field of an ACK frame (see Section 19.3). This allows the receiver
+of the ACK to adjust for any intentional delays, which is important for
+delayed acknowledgements, when estimating the path RTT. A packet might be
```
martinthomson: I think that you mean "which is important for getting a better estimate of the path RTT when acknowledgments are delayed"

draft-ietf-quic-transport.md
```diff
@@ -2947,6 +2997,25 @@ to unnecessarily retransmit some data.  Standard QUIC algorithms
...
+held in the OS kernel or elsewhere on the host before being processed.
+An endpoint SHOULD NOT include these unintentional delays when populating
```
```suggestion
An endpoint MUST NOT include delays that is does not control when populating
```

I think that we can use MUST NOT here.  Also, I think that outside of endpoint control is the key.

draft-ietf-quic-transport.md
```diff
@@ -2947,6 +2997,25 @@ to unnecessarily retransmit some data.  Standard QUIC algorithms
...
+the Ack Delay field in an ACK frame.
+
+An endpoint MUST NOT excessively delay acknowledgements of ack-eliciting
+packets.  The maximum ack delay is communicated in the max_ack_delay transport
```
```suggestion
packets.  An endpoint commits to a maximum delay using the max_ack_delay transport
```

draft-ietf-quic-transport.md
```suggestion
parameter; see {{transport-parameter-definitions}}.  max_ack_delay implies an explicit contract:
```

draft-ietf-quic-transport.md
```diff
@@ -2947,6 +2997,25 @@ to unnecessarily retransmit some data.  Standard QUIC algorithms
...
+parameter; see Section 18.1.  max_ack_delay implies an explicit contract:
+an endpoint promises to never delay acknowledgments of an ack-eliciting packet
+by more than the indicated value. If it does, any excess accrues to the RTT
+estimate and could result in spurious retransmissions from the peer.
```
martinthomson: Do you really get spurious retransmissions if your estimate of RTT is too high?  I thought that the effect would just come down to poor performance, maybe over-commitment of memory, and that sort of thing.

janaiyengar: I think this should have been "delayed retransmissions".

janaiyengar:
Ok, I've addressed comments and rebased, PTAL.

janaiyengar:
@martinthomson, PTAL.

draft-ietf-quic-transport.md
```diff
@@ -2979,10 +2979,49 @@ packet.
...
+performance can be improved by acknowledging immediately in several cases, as
+discussed below.
+
 ### Sending ACK Frames
 
-An endpoint sends ACK frames to acknowledge packets it has received and
-processed.
+An ACK frame SHOULD be sent immediately upon receipt of a second ack-eliciting
+packet. Loss detection does not assume that the peer sends an ACK immediately on
+receiving a second ack-eliciting packet.
```
martinthomson: These are both good and useful statements, but they sort of stand awkwardly next to each other as if they are strangers.

> An endpoint SHOULD immediately send an ACK frame when two ack-eliciting packets have been received without being acknowledged.  Though the algorithms in {{QUIC-RECOVERY}} do not depend on receivers following this guidance, frequent feedback improves loss and congestion response.

I tried to say something about the fact that this also provides some assurances about acknowledgement behaviour that senders could use when innovating with new algorithms, but it ended up being quite hard to do that without distracting too much from the core purpose of the section.  Also, the SHOULD got in the way: there is no real assurance, there we have to rely on the fact that this is a good recommendation that most deployments will likely respect.

draft-ietf-quic-transport.md
```diff
@@ -3020,9 +3059,34 @@ This ensures an acknowledgment is sent at least once per RTT when packets
...
+When an ACK frame is sent, one or more ranges of acknowledged packets are
+included.  Including older packets reduces the chance of spurious retransmits
+caused by losing previously sent ACK frames, at the cost of larger ACK frames.
+
+ACK frames SHOULD always acknowledge the most recently received packets, and the
+more out-of-order the packets are, the more important it is to send an updated
+ACK frame quickly, to prevent the peer from declaring a packet as lost and
+spuriously retransmitting the frames it contains.
+
+Below is one recommended approach for determining what packets to include in an
```
martinthomson: "Below" is a little imprecise.  Is this the next section, or next several sections?

```suggestion
{{ack-tracking}} and {{ack-limiting}} describe an exemplary approach for determining what packets to acknowledge in each
```

draft-ietf-quic-transport.md
```diff
@@ -3020,9 +3059,34 @@ This ensures an acknowledgment is sent at least once per RTT when packets
...
+ACK frame.
 
+### Receiver Tracking of ACK Frames
```
```suggestion
### Receiver Tracking of ACK Frames {#ack-tracking}
```

draft-ietf-quic-transport.md
```diff
@@ -3020,9 +3059,34 @@ This ensures an acknowledgment is sent at least once per RTT when packets
...
+
+In cases without ACK frame loss, this algorithm allows for a minimum of 1 RTT
+of reordering. In cases with ACK frame loss and reordering, this approach does
+not guarantee that every acknowledgement is seen by the sender before it is no
+longer included in the ACK frame. Packets could be received out of order and
+all subsequent ACK frames containing them could be lost. In this case, the
+loss recovery algorithm could cause spurious retransmits, but the sender will
+continue making forward progress.
 
 ### Limiting ACK Ranges
```
```suggestion
### Limiting ACK Ranges {#ack-limiting}
```

draft-ietf-quic-transport.md
```diff
@@ -3046,12 +3110,26 @@ to unnecessarily retransmit some data.  Standard QUIC algorithms
...
+ACK to adjust for any intentional delays, which is important for getting a
+better estimate of the path RTT when acknowledgments are delayed. A packet might
+be held in the OS kernel or elsewhere on the host before being processed.  An
+endpoint MUST NOT include delays that is does not control when populating the
+Ack Delay field in an ACK frame.
+
+An endpoint MUST NOT excessively delay acknowledgements of ack-eliciting
+packets.  An endpoint commits to a maximum delay using the max_ack_delay
+transport parameter; see {{transport-parameter-definitions}}.  max_ack_delay
+implies an explicit contract: an endpoint promises to never delay
```
martinthomson: implies explicit is odd:

```suggestion
declares an explicit contract: an endpoint promises to never delay
```

draft-ietf-quic-transport.md
ianswett: I want to do another larger pass over this text, but I'd like to keep it as is for now.

janaiyengar: I'm trying out something -- give me a few minutes

janaiyengar: Ok, a few changes, look at the last commit. I've moved some text around and I think I've captured the spirit of this more clearly.

draft-ietf-quic-transport.md
```diff
@@ -2979,21 +2979,57 @@ packet.
...
-packet (one containing only ACK and/or PADDING frames), even if there are
-packet gaps which precede the received packet. Limiting ACK frames avoids an
-infinite feedback loop of acknowledgements, which could prevent the connection
-from ever becoming idle. However, the endpoint acknowledges non-ACK-eliciting
-packets when it sends an ACK frame.
+A receiver's delayed acknowledgment timer SHOULD NOT exceed the current RTT
+estimate or the value it indicates in the `max_ack_delay` transport parameter.
+This ensures an acknowledgment is sent at least once per RTT when packets
+needing acknowledgement are received.  The sender can use the receiver's
+`max_ack_delay` value in determining timeouts for timer-based retransmission.
```
nibanks: Does this text `SHOULD NOT exceed the current RTT estimate` mean the peer could/should use a min of the max_ack_delay and the current RTT when computing probe timeouts?
```
PTO = smoothed_rtt + max(4*rttvar, kGranularity) + min(max_ack_delay, smoothed_rtt)
```

ianswett: Yes, probably.  This text is pre-existing, and maybe with max_ack_delay the 1 ack per RTT recommendation should be removed, but that's not an editorial change.

janaiyengar:
@ianswett : I think the ACK per RTT is sensible.... if the RTT is 1us, you still need to send an ACK in that timeframe if the sender is to send more packets. But yes, let's discuss that in a separate issue.

PR #2918, opened 2019-07-21
Fix language around CC on migration

### LINKED ISSUE #2685: CC state after a change of path?
I think a normal transport assumption is that a change of path implies a (potential) loss of congestion state, and that the endpoint should reset the path CC parameters. 

Transport-ID: “On confirming a peer's ownership of its new address, an endpoint
SHOULD immediately reset the congestion controller and round-trip
time estimator for the new path.”

• Does the "reset" require setting the CC to the initial window value for a connection? I think so.
• Why not MUST? 

Transport-ID:  “An endpoint MUST NOT return to the send rate used for the previous
path unless it is reasonably sure that the previous send rate is
valid for the new path. “

This could therefore be seen as a “significant” departure from conventional approaches considered safe - and extends this to “reasonably sure”. 
• Is there a rationale for extending this to “reasonably sure”, because this seems bolder than permitted for a new TCP connection. How do people quantify reasonably sure, and why not just in this case grow from the Initial window?

janaiyengar:
Cleans up some text in the transport doc around reseting (or not) CC and RTT values on migration.

Closes #2685

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2020,10 +2020,10 @@ unless it has knowledge that the previous send rate or round-trip time estimate
 is valid for the new path.  For instance, an endpoint might infer that a change
 in only the client's port number is likely indicative of a NAT rebinding,
 meaning that the new path is likely to have similar bandwidth and round-trip
-time. This determination might be a heuristic and is likely to be imperfect.
-Implementations are advised to be cautious when using previous values on the new
-path. If the determination is incorrect, the congestion controller and the RTT
-estimator are expected to adapt to the new path.
+time. This determination might be a heuristic and is likely to be imperfect.  If
+the determination is incorrect, the congestion controller and the RTT estimator
+are expected to adapt to the new path.  Generally, implementations are advised
+to be cautious when using previous values on a new path.
 
 There may be apparent reordering at the receiver when an endpoint sends data and
 probes from/to multiple addresses during the migration period, since the two
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2016,11 +2016,11 @@ estimation for the new path.
 On confirming a peer's ownership of its new address, an endpoint MUST
 immediately reset the congestion controller and round-trip time estimator for
 the new path to initial values (see Sections A.3 and B.3 in {{QUIC-RECOVERY}})
-unless it has knowledge that the previous send rate or round-trip time estimate
+unless it has knowledge that a previous send rate or round-trip time estimate
 is valid for the new path.  For instance, an endpoint might infer that a change
 in only the client's port number is likely indicative of a NAT rebinding,
 meaning that the new path is likely to have similar bandwidth and round-trip
-time. This determination might be a heuristic and is likely to be imperfect.  If
+time. This determination is likely to be imperfect.  If
 the determination is incorrect, the congestion controller and the RTT estimator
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2013,17 +2013,17 @@ The capacity available on the new path might not be the same as the old path.
...
-rebinding in a middlebox and not a complete change in path.  This determination
-likely depends on heuristics, which could be imperfect; if the new path capacity
-is significantly reduced, ultimately this relies on the congestion controller
-responding to congestion signals and reducing send rates appropriately.
+the new path to initial values (see Sections A.3 and B.3 in {{QUIC-RECOVERY}})
+unless it has knowledge that the previous send rate or round-trip time estimate
+is valid for the new path.  For instance, an endpoint might infer that a change
+in only the client's port number is likely indicative of a NAT rebinding,
+meaning that the new path is likely to have similar bandwidth and round-trip
+time. This determination might be a heuristic and is likely to be imperfect.  If
```
```suggestion
time. This determination is likely to be imperfect.  If
```

draft-ietf-quic-transport.md
```diff
@@ -2013,17 +2013,17 @@ The capacity available on the new path might not be the same as the old path.
...
-
-An endpoint MUST NOT return to the send rate used for the previous path unless
-it is reasonably sure that the previous send rate is valid for the new path.
-For instance, a change in the client's port number is likely indicative of a
-rebinding in a middlebox and not a complete change in path.  This determination
-likely depends on heuristics, which could be imperfect; if the new path capacity
-is significantly reduced, ultimately this relies on the congestion controller
-responding to congestion signals and reducing send rates appropriately.
+the new path to initial values (see Sections A.3 and B.3 in {{QUIC-RECOVERY}})
+unless it has knowledge that the previous send rate or round-trip time estimate
```
```suggestion
unless it has knowledge that a previous send rate or round-trip time estimate
```
One reason a peer might do that is that you migrated cell->Wi-Fi->cell, and it still remembers the values from when you were on cellular.  Perhaps don't restrict this to the immediate previous measurement?

draft-ietf-quic-transport.md
```diff
@@ -2013,17 +2013,17 @@ The capacity available on the new path might not be the same as the old path.
...
-rebinding in a middlebox and not a complete change in path.  This determination
-likely depends on heuristics, which could be imperfect; if the new path capacity
-is significantly reduced, ultimately this relies on the congestion controller
-responding to congestion signals and reducing send rates appropriately.
+the new path to initial values (see Sections A.3 and B.3 in {{QUIC-RECOVERY}})
+unless it has knowledge that a previous send rate or round-trip time estimate
+is valid for the new path.  For instance, an endpoint might infer that a change
+in only the client's port number is likely indicative of a NAT rebinding,
+meaning that the new path is likely to have similar bandwidth and round-trip
+time. This determination is likely to be imperfect.  If
```
tfpauly: Reading through this at an editorial level, it stands out a bit that the text says `likely` three times in the space of two sentences. I would argue that if an endpoint is inferring something, it's already clear that the state it is inferring is "likely" (but not certain) based on the fact that the state is inferred. I would suggest text like:

```
For instance, an endpoint might infer that a change
	in only the client's port number is indicative of a NAT rebinding,
	meaning that the new path is likely to have similar bandwidth and round-trip
	time. However, this determination will often be imperfect.
```

janaiyengar: Thanks for the editorializing, @tfpauly - incorporated.

PR #2919, opened 2019-07-21
MinRTT for the path

### LINKED ISSUE #2908: Why is Min RTT kept for the connection lifetime?
Section 5.2 of draft-ietf-quic-recovery:

(1)   the minimum value observed over the lifetime of the connection

- Why a minimum specified over the entire lifetime? There are paths where the initial RTT may the lower, I am unable to see why that artefact should have a long-term impact on the connection. Even if it is just a change in traffic sharing the path’s bottleneck, it still isn’t something that needs to have impact for the lifetime of a connection. Can’t we decay the value somehow, so that it reflect the recent min_RTT.

draft-ietf-quic-recovery, and here is why a smaller than current min_RTT is problematic:
(2)  MUST NOT apply the adjustment if the resulting RTT sample is
      smaller than the min_rtt.  This limits the underestimation that a
      misreporting peer can cause to the smoothed_rtt.

- How do I understand this works with paths exhibiting varying path delay?  I still can’t fathom what will happen when the lowest RTT in recent time is much larger than the min_RTT, because the RTT floor has increased (e.g. because of a change at L2). I see some rather strange behaviour in traces of QUIC when there is large long-term jitter. At present, I think this results in making inappropriate adjustments?

ianswett:
Also explains what happens to MinRTT if the path RTT changes.

Fixes #2908

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -336,7 +336,7 @@ endpoint:
   min_rtt.  This limits the underestimation that a misreporting peer can cause
   to the smoothed_rtt.
 
-On the first RTT sample for a path, the smoothed_rtt is set to the latest_rtt.
+On the first RTT sample for a network path, the smoothed_rtt is set to the latest_rtt.
 
 smoothed_rtt and rttvar are computed as follows, similar to {{?RFC6298}}.  On
 the first RTT sample for a path:
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -244,7 +244,7 @@ more accurate round-trip time estimate (see Section 13.2 of {{QUIC-TRANSPORT}}).
 At a high level, an endpoint measures the time from when a packet was sent to
 when it is acknowledged as a round-trip time (RTT) sample.  The endpoint uses
 RTT samples and peer-reported host delays (see Section 13.2 of
-{{QUIC-TRANSPORT}}) to generate a statistical description of the path's RTT.
+{{QUIC-TRANSPORT}}) to generate a statistical description of the network path's RTT.
 An endpoint computes the following three values for each path: the minimum
 value observed over the lifetime of the path (min_rtt), an
 exponentially-weighted moving average (smoothed_rtt), and the mean deviation
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -291,7 +291,7 @@ retain sufficient history is an open research question.
 
 ## Estimating min_rtt {#min-rtt}
 
-min_rtt is the minimum RTT observed for a given path.  min_rtt is set to the
+min_rtt is the minimum RTT observed for a given network path.  min_rtt is set to the
 latest_rtt on the first RTT sample, and to the lesser of min_rtt and latest_rtt
 on subsequent samples.  In this document, min_rtt is used by loss detection to
 reject implausibly small rtt samples.
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -244,9 +244,9 @@ more accurate round-trip time estimate (see Section 13.2 of {{QUIC-TRANSPORT}}).
 At a high level, an endpoint measures the time from when a packet was sent to
 when it is acknowledged as a round-trip time (RTT) sample.  The endpoint uses
 RTT samples and peer-reported host delays (see Section 13.2 of
-{{QUIC-TRANSPORT}}) to generate a statistical description of the network path's RTT.
-An endpoint computes the following three values for each path: the minimum
-value observed over the lifetime of the path (min_rtt), an
+{{QUIC-TRANSPORT}}) to generate a statistical description of the network
+path's RTT. An endpoint computes the following three values for each path:
+the minimum value observed over the lifetime of the path (min_rtt), an
 exponentially-weighted moving average (smoothed_rtt), and the mean deviation
 (referred to as "variation" in the rest of this document) in the observed RTT
 samples (rttvar).
@@ -291,10 +291,10 @@ retain sufficient history is an open research question.
 
 ## Estimating min_rtt {#min-rtt}
 
-min_rtt is the minimum RTT observed for a given network path.  min_rtt is set to the
-latest_rtt on the first RTT sample, and to the lesser of min_rtt and latest_rtt
-on subsequent samples.  In this document, min_rtt is used by loss detection to
-reject implausibly small rtt samples.
+min_rtt is the minimum RTT observed for a given network path.  min_rtt is set
+to the latest_rtt on the first RTT sample, and to the lesser of min_rtt and
+latest_rtt on subsequent samples.  In this document, min_rtt is used by loss
+detection to reject implausibly small rtt samples.
 
 An endpoint uses only locally observed times in computing the min_rtt and does
 not adjust for ACK delays reported by the peer.  Doing so allows the endpoint
@@ -302,10 +302,11 @@ to set a lower bound for the smoothed_rtt based entirely on what it observes
 (see {{smoothed-rtt}}), and limits potential underestimation due to
 erroneously-reported delays by the peer.
 
-The RTT for a network path may change over time.  If a path's actual RTT decreases,
-the min_rtt will adapt immediately on the first low sample.  If the network path's
-actual RTT increases, the min_rtt will not adapt to it, allowing future
-RTT samples that are smaller than the new RTT be included in smoothed_rtt.
+The RTT for a network path may change over time.  If a path's actual RTT
+decreases, the min_rtt will adapt immediately on the first low sample.  If
+the path's actual RTT increases, the min_rtt will not adapt to it, allowing
+future RTT samples that are smaller than the new RTT be included in
+smoothed_rtt.
 
 ## Estimating smoothed_rtt and rttvar {#smoothed-rtt}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -326,10 +326,10 @@ the peer. For Initial and Handshake packets, a max_ack_delay of 0 is used.
 At a high level, an endpoint measures the time from when a packet was sent to
 when it is acknowledged as a round-trip time (RTT) sample.  The endpoint uses
 RTT samples and peer-reported host delays ({{host-delay}}) to generate a
-statistical description of the connection's RTT.  An endpoint computes the
+statistical description of the path's RTT.  An endpoint computes the
 following three values: the minimum value observed over the lifetime of the
-connection (min_rtt), an exponentially-weighted moving average (smoothed_rtt),
-and the variance in the observed RTT samples (rttvar).
+congestion control context (min_rtt), an exponentially-weighted moving average
```
```suggestion
RTT estimator's context (min_rtt), an exponentially-weighted moving average
```

draft-ietf-quic-recovery.md
```diff
@@ -371,16 +371,23 @@ retain sufficient history is an open research question.
 
 ## Estimating min_rtt {#min-rtt}
 
-min_rtt is the minimum RTT observed over the lifetime of the connection.
-min_rtt is set to the latest_rtt on the first sample in a connection, and to the
-lesser of min_rtt and latest_rtt on subsequent samples.
+min_rtt is the minimum RTT observed for a given congestion control context.
```
janaiyengar: min_rtt is only used for loss recovery in this doc. I think it's more appropriate to say RTT estimator context in general.

draft-ietf-quic-recovery.md
```diff
@@ -371,16 +371,23 @@ retain sufficient history is an open research question.
...
+min_rtt is set to the latest_rtt on the first RTT sample, and to the lesser
+of min_rtt and latest_rtt on subsequent samples.  min_rtt is used to reject
+implausible RTT samples, but is not used directly in congestion control or
```
janaiyengar: I would drop the rest of the text after "samples" and perhaps add "which would otherwise cause the smoothed_rtt estimate to be incorrectly low.". I understand what you're trying to say, but I think saying that it is not used directly in loss recovery makes it a bit confusing I think. It is used, and I think people will have different reads on what's indirect and what's

draft-ietf-quic-recovery.md
```diff
@@ -371,16 +371,23 @@ retain sufficient history is an open research question.
...
+loss recovery.
 
 An endpoint uses only locally observed times in computing the min_rtt and does
 not adjust for host delays reported by the peer ({{host-delay}}).  Doing so
 allows the endpoint to set a lower bound for the smoothed_rtt based entirely on
 what it observes (see {{smoothed-rtt}}), and limits potential underestimation
 due to erroneously-reported delays by the peer.
 
+The RTT for the path may change over time.  If the real path RTT decreases,
```
```suggestion
The RTT for a path may change over time.  If a path's actual RTT decreases,
```

draft-ietf-quic-recovery.md
```diff
@@ -371,16 +371,23 @@ retain sufficient history is an open research question.
...
+it should be observed quickly.  If the real path RTT increases, it will not
```
```suggestion
the min_rtt will adapt immediately on the first low sample.  If the path's actual RTT increases,
```

draft-ietf-quic-recovery.md
```diff
@@ -371,16 +371,23 @@ retain sufficient history is an open research question.
...
+be observed, but the impact is limited to allowing some RTT samples that are
```
```suggestion
the min_rtt will not adapt to it, potentially causing future RTT samples that are
```

draft-ietf-quic-recovery.md
```diff
@@ -410,11 +417,11 @@ endpoint:
   min_rtt.  This limits the underestimation that a misreporting peer can cause
   to the smoothed_rtt.
 
-On the first RTT sample in a connection, the smoothed_rtt is set to the
-latest_rtt.
+On the first RTT sample in a congestion control context, the smoothed_rtt is
```
janaiyengar: RTT context

draft-ietf-quic-recovery.md
```diff
@@ -410,11 +417,11 @@ endpoint:
...
+set to the latest_rtt.
 
 smoothed_rtt and rttvar are computed as follows, similar to {{?RFC6298}}.  On
-the first RTT sample in a connection:
+the first RTT sample in a congestion control context:
```
janaiyengar: RTT context

draft-ietf-quic-recovery.md
ianswett: We never define this, but isn't the RTT estimator a subset of the congestion control context?  Also, we only use RTT estimator once in this document at the moment.

Possibly I should use path for all of these now, against @erickinnear suggestion?  Also, path is much shorter than "congestion control context"

draft-ietf-quic-recovery.md
ianswett: path?

draft-ietf-quic-recovery.md
ianswett: path?

draft-ietf-quic-recovery.md
erickinnear: Elsewhere don’t we discuss resetting a congestion control context? It seems like that often lines up with a path, but they don’t have to be the same thing. It seems like you’re saying that RTT isn’t part of congestion control (which sounds totally valid but we then need some better way to word that)?

draft-ietf-quic-recovery.md
martinthomson: When you say "congestion control context", why not "path"?

ianswett: Good question, I'll talk with Eric and Jana in person about this one and we can figure out some consistent terminology.

draft-ietf-quic-recovery.md
```diff
@@ -249,11 +249,10 @@ more accurate round-trip time estimate (see Section 13.2 of {{QUIC-TRANSPORT}}).
 At a high level, an endpoint measures the time from when a packet was sent to
 when it is acknowledged as a round-trip time (RTT) sample.  The endpoint uses
 RTT samples and peer-reported host delays (see Section 13.2 of
-{{QUIC-TRANSPORT}}) to generate a statistical description of the connection's
-RTT. An endpoint computes the following three values: the minimum value
-observed over the lifetime of the connection (min_rtt), an
-exponentially-weighted moving average (smoothed_rtt), and the variance in the
-observed RTT samples (rttvar).
+{{QUIC-TRANSPORT}}) to generate a statistical description of the path's RTT.
```
```suggestion
{{QUIC-TRANSPORT}}) to generate a statistical description of the network path's RTT.
```

draft-ietf-quic-recovery.md
```diff
@@ -295,16 +294,22 @@ retain sufficient history is an open research question.
 
 ## Estimating min_rtt {#min-rtt}
 
-min_rtt is the minimum RTT observed over the lifetime of the connection.
-min_rtt is set to the latest_rtt on the first sample in a connection, and to the
-lesser of min_rtt and latest_rtt on subsequent samples.
+min_rtt is the minimum RTT observed for a given path.  min_rtt is set to the
```
```suggestion
min_rtt is the minimum RTT observed for a given network path.  min_rtt is set to the
```

draft-ietf-quic-recovery.md
```diff
@@ -295,16 +294,22 @@ retain sufficient history is an open research question.
...
+on subsequent samples.  In this document, min_rtt is used by loss detection to
+reject implausibly small rtt samples.
 
 An endpoint uses only locally observed times in computing the min_rtt and does
 not adjust for host delays reported by the peer.  Doing so allows the endpoint
 to set a lower bound for the smoothed_rtt based entirely on what it observes
 (see {{smoothed-rtt}}), and limits potential underestimation due to
 erroneously-reported delays by the peer.
 
+The RTT for a path may change over time.  If a path's actual RTT decreases,
```
```suggestion
The RTT for a network path may change over time.  If a path's actual RTT decreases,
```

draft-ietf-quic-recovery.md
```diff
@@ -333,11 +338,10 @@ endpoint:
   min_rtt.  This limits the underestimation that a misreporting peer can cause
   to the smoothed_rtt.
 
-On the first RTT sample in a connection, the smoothed_rtt is set to the
-latest_rtt.
+On the first RTT sample for a path, the smoothed_rtt is set to the latest_rtt.
```
```suggestion
On the first RTT sample for a network path, the smoothed_rtt is set to the latest_rtt.
```

draft-ietf-quic-recovery.md
```diff
@@ -333,11 +338,10 @@ endpoint:
...
 
 smoothed_rtt and rttvar are computed as follows, similar to {{?RFC6298}}.  On
-the first RTT sample in a connection:
+the first RTT sample for a path:
```
```suggestion
the first RTT sample for a network path:
```

draft-ietf-quic-recovery.md
```diff
@@ -295,16 +294,22 @@ retain sufficient history is an open research question.
...
+reject implausibly small rtt samples.
 
 An endpoint uses only locally observed times in computing the min_rtt and does
 not adjust for host delays reported by the peer.  Doing so allows the endpoint
 to set a lower bound for the smoothed_rtt based entirely on what it observes
 (see {{smoothed-rtt}}), and limits potential underestimation due to
 erroneously-reported delays by the peer.
 
+The RTT for a path may change over time.  If a path's actual RTT decreases,
+the min_rtt will adapt immediately on the first low sample.  If the path's
```
```suggestion
the min_rtt will adapt immediately on the first low sample.  If the network path's
```

Issue #2387, opened 2019-01-30
Write up our security target against on-path attack


ekr:

mnot:
@ekr do you have a PR in progress?

larseggert:
Discussed in Montreal, will be a add-on to #2143

janaiyengar:
@mnot @larseggert @LPardue: Editors believe that #2925, which has been merged, addresses this issue. Can we close this issue alongwith #2143 ?

martinthomson:
The proposal here is to do no more than we have already done in #2925.

Issue #2143, opened 2018-12-13
Be more conservative about migration?


ekr:
I've been reading through the migration mechanism and it seems to me
that assuming that a packet from a given IP address is evidence of
migration creates a pretty easy DoS attack from an on-path attacker.
Consider the case where we have a quiescent connection between A and B
and an on-path attacker and then B sends a packet to A.

In the diagram below, the notation [A:B]: Z -> means a packet with a
source addr of A, a dest addr of B, and a payload of Z

```
A                       Attacker (Z)                  B

                                         <- [B:A]: Data
                     <- [B:A]: Data
[A:B]: ACK ->
                        [Z:B]: ACK ->

                                       X <- [B:Z]: Data
                                       X <- [B:Z]: Probe
```
                                       
As long as A doesn't try to send a data packet, B will continue to
think that A is at Z for the duration of the validation timer
(currently 3 RTO?) [0]. That's not the worst thing in the world, but
it's a pretty good cost/benefit for an on-path attacker who only has
to forge one packet; I'm not sure that that's a great design tradeoff
for a protocol where we're otherwise trying to stop on-path attacker
from interfering with the connection after the handshake completes.

The reason this works is the receipt of a packet from addr X is a
trigger for abandoning the previous path and trying migration. Another
time this can happen is at the tail of a client transmission,
though that will presumably recover faster.

At a high level, it seems like there are two major types of
migration:

1. Unknown migration (e.g., NAT re-binding)
2. Known migration (e.g., transferring from WiFi to 3G)

In the latter case, we can have the client give notice that it's
migrating, even if it doesn't know it's visible IP address, and have
the server refuse to change peer addresses unless it sees that
indication, so it's just the former case we have to deal with, and as
above. The present design does that but at the cost of creating the
situation above.

I wonder if we should be more conservative about initiating migration.
I don't have a fully worked out proposal, but here are some ideas:

- Send packets to both the original and new addresses during the
  validation period.
  
- If you receive an apparent migration that looks like a rebinding
  (e.g., same address, different port) within X seconds of a previous
  packet, then probe first (or in parallel as above).

- Have the client send some packet that indicates it thinks it's
  migrating (this seems tricky to get right)

[0] Note that if we were to allow migration on the client and the
server, even worse attacks are possible.

mikkelfj:
When you say on-path, do you mean observer? I.e. with the ability to race a packet, or to drop or replace a packet.

I suppose a race will not be successful in this attack.

[...]

PR #2925, opened 2019-07-23
Add initial threat model to security considerations

erickinnear:
This adds a section for a QUIC threat model, filling in only the migration section (for #2143, some of the rest of the threat model will come in #2387).

Some wordsmithing can be applied here, and we may want additional issues to fill out detailed descriptions of some of the worst attacks and outline heuristics that might help, but for now I'm focusing this purely on documenting what's possible/not possible for an attacker to do today.

Addresses #2143.


draft-ietf-quic-transport.md
```diff
@@ -5753,13 +5753,184 @@ DecodePacketNumber(largest_pn, truncated_pn, pn_nbits):
...
+
+TBD.
+
+## Short Headers {#short-headers-properties}
+
+TBD.
+
+## Connection Migration {#migration-properties}
+
+Connection Migration ({{migration}}) provides endpoints with the ability to
```
nibanks: I don't think this intro paragraph is really necessary.

draft-ietf-quic-transport.md
```diff
@@ -5753,13 +5753,184 @@ DecodePacketNumber(largest_pn, truncated_pn, pn_nbits):
...
+# Overview of Security Properties {#security-properties}
+
+A complete security analysis of QUIC is outside the scope of this document.  In
+this appendix, we provide an informal description of the desired security
+properties as an aid to implementors and to help guide protocol analysis.
+
+We cover properties of the handshake, general transport, and migration
+separately.
+
+## Handshake {#handshake-properties}
```
nibanks: Maybe a 0-RTT section as well?

erickinnear: Makes sense, good thinking! 
Probably others that might be interesting, too, not sure if it's worth adding a bunch of TBD sections, but definitely should have an issue covering writing up the 0-RTT section.

erickinnear:
Upcoming change to terminology here as well, just as a note.

draft-ietf-quic-transport.md
```diff
@@ -5753,13 +5753,184 @@ DecodePacketNumber(largest_pn, truncated_pn, pn_nbits):
...
 ~~~
 
+# Overview of Security Properties {#security-properties}
+
+A complete security analysis of QUIC is outside the scope of this document.  In
+this appendix, we provide an informal description of the desired security
+properties as an aid to implementors and to help guide protocol analysis.
+
+We cover properties of the handshake, general transport, and migration
+separately.
```
dtikhonov: I'd drop the use of "we."  In the whole `draft-ietf-quic-transport-22`, for instance, the word "we" is encountered only once, while RFC 7540 and 7541 contain no instances of this pronoun.

I suggest instead:
- _This appendix provides and informal description..._; and
- _Handshake, ...etc. are covered separately_.

MikeBishop: ...and perhaps file an issue for that other use of "we."

draft-ietf-quic-transport.md
```diff
@@ -5753,13 +5753,184 @@ DecodePacketNumber(largest_pn, truncated_pn, pn_nbits):
...
+
+- Inspect packets
+- Modify unencrypted packet headers
+- Inject new packets
+- Delay packets
+- Drop packets
+
+An on-path attacker cannot:
+
+- Modify encrypted packet payloads
```
MikeBishop: Well, they can, but the packets will no longer be accepted as valid, which is part of this threat model and mitigation.  Also, you don't discuss the unencrypted, but authenticated, portions of the headers.  Perhaps you should frame the capabilities in terms of what they can do to authenticated portions of the packet, and note that the payload is both encrypted *and* authenticated?

draft-ietf-quic-transport.md
```diff
@@ -5753,13 +5753,184 @@ DecodePacketNumber(largest_pn, truncated_pn, pn_nbits):
...
+which the attacker is not on-path.
+
+4. An on-path attacker can reduce the throughput of a connection by delaying
+packets or dropping them.
+
+
+### Off-Path Attacker
+
+An off-path attacker is not directly on the path between the QUIC client and
+server, but is able to obtain copies of all packets sent between the client and
```
MikeBishop: Perhaps change this to "may be able to obtain copies of some or all packets sent between the client and the server"?  Certainly there are subclasses of off-path attackers who see no traffic (inject-only) or who get only a partial view (e.g. eavesdropping on only one direction, one path, etc.).

draft-ietf-quic-transport.md
```diff
@@ -5753,13 +5753,184 @@ DecodePacketNumber(largest_pn, truncated_pn, pn_nbits):
...
+
+For the purposes of this discussion, we assume that an off-path attacker has the
+ability to observe, modify, and re-inject a packet into the network that will
+reach the destination endpoint prior to the arrival of the original packet
+observed by the attacker.  In other words, the attacker has the ability to
+consistently "win" a race with the legitimate packets between the endpoints,
+potentially causing the original packet to be ignored by the recipient.
+
+We also assume that the attacker has the resources necessary to affect NAT
+state, potentially both causing an endpoint to lose its NAT binding, and an
```
MikeBishop: Pick "the" or "an" here.  Jumping back and forth is throwing me; if there's a pattern, I'm not able to intuit it.

draft-ietf-quic-transport.md
```diff
@@ -5753,13 +5753,184 @@ DecodePacketNumber(largest_pn, truncated_pn, pn_nbits):
    return candidate_pn
 ~~~
 
+# Overview of Security Properties {#security-properties}
```
martinthomson: This is probably better placed as a "Security Considerations" section, with the current text incorporated into this as appropriate (mostly under mitigations).

draft-ietf-quic-transport.md
```diff
@@ -5753,13 +5753,184 @@ DecodePacketNumber(largest_pn, truncated_pn, pn_nbits):
...
+## Short Headers {#short-headers-properties}
+
+TBD.
+
+## Connection Migration {#migration-properties}
+
+Connection Migration ({{migration}}) provides endpoints with the ability to
+transition between IP addresses and ports on multiple paths, using one path at a
+time for sending non-probing frames.  Path validation helps to establish
+bidirectional connectivity with the peer over a new path and prevents source
```
martinthomson: The purpose of path validation is to establish that a peer is both willing and able to receive packets sent on the path.  This helps reduce the effects of address spoofing, by limiting the number of packets sent to a spoofed address, unless an attacker is able to also receive packets sent to that address.

draft-ietf-quic-transport.md
```diff
@@ -5753,13 +5753,184 @@ DecodePacketNumber(largest_pn, truncated_pn, pn_nbits):
...
+address spoofing from being used in a denial-of-service attack.  Migration is
+further restricted to be initiated only by QUIC clients.
```
martinthomson: I think that this is not a security property as much as it is a correctness or simplicity choice we made.  You can drop this sentence.

draft-ietf-quic-transport.md
martinthomson: Mike's point is excellent.  The point is that packets are authenticated.

draft-ietf-quic-transport.md
```diff
@@ -5753,13 +5753,184 @@ DecodePacketNumber(largest_pn, truncated_pn, pn_nbits):
...
+endpoint is required to send packets through this attacker to establish
+connectivity on a given path.
+
+An on-path attacker can:
+
+- Inspect packets
+- Modify unencrypted packet headers
+- Inject new packets
+- Delay packets
+- Drop packets
```
martinthomson: Another point on capabilities - an on-path attacker can split datagrams that contain multiple packets and merge packets into a single datagram.

draft-ietf-quic-transport.md
```diff
@@ -5753,13 +5753,184 @@ DecodePacketNumber(largest_pn, truncated_pn, pn_nbits):
...
+### On-Path Attacker
+
+An on-path attacker is present between the QUIC client and server, and an
+endpoint is required to send packets through this attacker to establish
+connectivity on a given path.
+
+An on-path attacker can:
+
+- Inspect packets
+- Modify unencrypted packet headers
```
```suggestion
- Modify IP and UDP packet headers
```

draft-ietf-quic-transport.md
```diff
@@ -5753,13 +5753,184 @@ DecodePacketNumber(largest_pn, truncated_pn, pn_nbits):
...
+- Drop packets
+
+An on-path attacker cannot:
+
+- Modify encrypted packet payloads
+
+In the presence of an on-path attacker, QUIC aims to provide the following
+properties.
+
+1. An on-path attacker can interrupt a QUIC connection, causing it to fail if it
```
martinthomson: The attacker doesn't interrupt the connection as much as it can prevent use of the path for that connection.

There is a special case for connection establishment that needs to be called out separately, in that an attacker can prevent connection establishment with much less interaction (i.e., it doesn't need to disrupt every packet).

draft-ietf-quic-transport.md
erickinnear: Removed all of the "we" from this diff, great suggestions as well, thanks!

draft-ietf-quic-transport.md
erickinnear: Updated this a bit, I suspect further discussion will also be in the security properties of short/long headers.

draft-ietf-quic-transport.md
erickinnear: Updated, very good point, I think here I'm focusing mostly on the worst-case type of behavior (for example, assume an attacker can _always_ win a race), but I think there would be room to cover additional situations, both in the case where that changes the considerations or simply as a note that "these people also fall into this bucket".

draft-ietf-quic-transport.md
erickinnear: I've changed some "the attacker" to "an attacker" to be more specific, we've got "an" for endpoint everywhere except one place where it's a packet between "the endpoints" that are having a conversation. 

There's a few places where we say something along the lines of 
> 2. An on-path attacker can prevent migration to a new path for which the
> attacker is also on-path by causing path validation to fail on the new path.

Which is a case where we use both "an" and "the", but otherwise it's all the same now. (And suggestions welcome for those cases, it started to sound very wrong when I switched those around.)

draft-ietf-quic-transport.md
erickinnear: Updated with that text, much nicer, thanks!

draft-ietf-quic-transport.md
erickinnear: Removed

draft-ietf-quic-transport.md
erickinnear: Good point! Hopefully we'll catch some other cases that got left out like this one -- I've added it.

draft-ietf-quic-transport.md
erickinnear: Nice, updated

draft-ietf-quic-transport.md
erickinnear: Reworded this to be a bit more clear about exactly what's getting broken (that path, not the whole thing). 

The establishment case I think I'm going to leave for the section on the handshake here.

draft-ietf-quic-transport.md
erickinnear: Two thoughts here: 
- This was an appendix since TLS 1.3 did it as an appendix. It's kind of nice that way, since we're documenting what we believe the rest of the document says, not introducing any new attacks or mitigations that shouldn't already have been covered above.
- From a purely aesthetic perspective, having 21.11.3.2. as a section means that I refer to one of these items as: 
> Section 21.11.3.2., point 3.

All that said, I've moved this to the last section in Security Considerations for the time being, and I don't have a super strong opinion on where it finally lands. :)
TLS is also a little bit different in that most of the document is a security consideration in one way or another, so some of the reasoning that resulted in structure there is not necessarily reflected in QUIC.

MikeBishop: It seems worth excerpting RFC 3552:
> While it is not a requirement that any given protocol or system be immune to all forms of attack, it is still necessary for authors to consider as many forms as possible.  Part of the purpose of the Security Considerations section is to explain what attacks are out of scope and what countermeasures can be applied to defend against them.
>
> There should be a clear description of the kinds of threats on the described protocol or technology.  This should be approached as an effort to perform "due diligence" in describing all known or foreseeable risks and threats to potential implementers and users.
>
> Authors MUST describe
>
>    1.   which attacks are out of scope (and why!)
>    2.   which attacks are in-scope
>          1.  and the protocol is susceptible to
>          2.  and the protocol protects against
>
> At least the following forms of attack MUST be considered:  eavesdropping, replay, message insertion, deletion, modification, and man-in-the-middle.  Potential denial of service attacks MUST be identified as well.  If the protocol incorporates cryptographic protection mechanisms, it should be clearly indicated which portions of the data are protected and what the protections are (i.e., integrity only, confidentiality, and/or endpoint authentication, etc.).  Some indication should also be given to what sorts of attacks the cryptographic protection is susceptible.  Data which should be held secret (keying material, random seeds, etc.) should be clearly labeled.

In other words, this reads a lot like a Security Considerations section.

ianswett:
As commented on the issue, this should fix #2387 as well.

erickinnear:
Added a few paragraphs describing general active and passive attacks in the context of RFC 3552 and more clearly defined off-path and on-path attackers to follow the definition established by that document. This will likely still take some iteration, but it should be mostly ready for that, comments welcome!

draft-ietf-quic-transport.md
erickinnear: Thanks! I've moved it up and updated the text to refer to it as such.

draft-ietf-quic-transport.md
erickinnear: Open to moving it around / trimming it down a bit. Mostly I wanted a place so if you read this you've got links and references to the appropriate sections, but without going beyond a sentence or three.

erickinnear:
@gorryfair Please take a look, would love your feedback on this!

mirjak:
I just had a quick read, and this looks fine (maybe there is an option to remove some redundancy but not sure). However, I would have expected that this would also talk about likability. Or do we already cover hat somewhere else? More generally I think it would be helpful to add some text at the beginning to not only define the attackers capabilities but also its intention, so the (goal of) the attack itself.

erickinnear:
Thanks @mirjak! There's a section, `migration-linkability`, that discusses linkability and sets out the requirements there. I wasn't originally thinking that we wanted to duplicate that in security considerations, but it could be really good to write up some of the guarantees that we think are/n't provided there (probably in a separate issue, though).

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,246 @@ decisions are made independently of client-selected values; a Source Connection
...
+destination is considered an on-path attacker.  Such an attacker generally is
+present between the QUIC client and server, and a QUIC endpoint is required to
+send packets through this attacker to establish connectivity on a given path.
+
+An on-path attacker can:
+
+- Inspect packets
+- Modify IP and UDP packet headers
+- Inject new packets
+- Delay packets
```
gorryfair: Delay (includes reordering)

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,246 @@ decisions are made independently of client-selected values; a Source Connection
...
+
+- Inspect packets
+- Inject new packets
+- Modify unencrypted packet headers
+
+A limited on-path attacker cannot:
+
+- Delay packets beyond the original packet duration
+- Drop packets
+- Modify encrypted packet payloads
```
gorryfair: he can - but they don't work

erickinnear: Changed in bda13ed

draft-ietf-quic-transport.md
erickinnear: Thanks! 
Changed in bda13ed

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+
+A spoofing attack, in which an attacker rewrites unprotected parts of a QUIC
+packet such as the source or destination address, is only effective if the
+attacker can forward packets to the original endpoint, as path validation
+({{migrate-validate}}) ensures that an endpoint's ability and willingness to
+decrypt QUIC packets is demonstrated before sending significant amounts of data
+to a new endpoint as part of an established QUIC connection.
+
+
+##### On-Path Active Attacks
```
martinthomson: I don't accept patches with five ######### in a row.  Can we bump the real stuff up a level?

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+contents of the packets to the original endpoint, since QUIC connections are
+both authenticated and encrypted.
+
+A blind attacker, one who injects packets without being able to observe valid
+packets for a QUIC connection, is unlikely to be successful, since QUIC packet
+protection ensures that valid packets are only generated by endpoints which
+possess the key material established during the handshake.  Similarly, any
+active attacker that observes QUIC packets and attempts to insert new data or
+modify existing data in those packets should not be able to generate packets
+deemed valid by the receiving endpoint.
```
martinthomson: This belongs in earlier sections. This is packet protection.

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+- Modify an authenticated and encrypted portion of a packet and cause the
+ recipient to accept that packet
+
+An on-path attacker has the opportunity to modify the packets that it observes,
+however any modifications to an authenticated portion of a packet will cause it
+to be dropped by the receiving endpoint as invalid, as QUIC payloads are both
+authenticated and encrypted.
+
+In the presence of an on-path attacker, QUIC aims to provide the following
+properties.
```
```suggestion
properties:
```

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+A limited on-path attacker can:
+
+- Inspect packets
+- Inject new packets
+- Modify unencrypted packet headers
+- Reorder packets
+
+A limited on-path attacker cannot:
+
+- Delay packets beyond the original packet duration
```
```suggestion
- Delay packets so that they arrive later than packets sent on the original path
```

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+Connection Migration ({{migration}}) provides endpoints with the ability to
+transition between IP addresses and ports on multiple paths, using one path at a
+time for transmission and receipt of non-probing frames.  Path validation
+({{migrate-validate}}) establishes that a peer is both willing and able
+to receive packets sent on a particular path.  This helps reduce the effects of
+address spoofing, by limiting the number of packets sent to a spoofed address,
+unless an attacker is able to also receive packets sent to that address.
+
+This section describes the intended security properties of connection migration
+when under various types of attack, as described in more detail by {{?RFC3552}}.
```
martinthomson: You definitely need to constrain this section to just DoS consequences of migration.  Because this doesn't cover what an attacker might learn about communications.

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+A limited on-path attacker cannot:
+
+- Delay packets beyond the original packet duration
+- Drop packets
+- Modify the authenticated and encrypted portion of a packet and cause the
+ recipient to accept that packet
+
+A limited on-path attacker can only delay packets up to the point that the
+original packets arrive before the duplicate packets, meaning that it cannot
+offer worse routing than the original path, only improved routing.  If a limited
```
```suggestion
offer routing with worse latency than the original path.  If a limited
```

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+on-path attacker drops packets, the original copy will still arrive at the
+destination endpoint.
+
+In the presence of a limited on-path attacker, QUIC aims to provide the
+following properties.
```
```suggestion
following properties:
```

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+An active attack ({{?RFC3552}}) involves writing data to the network.  An
+attacker with such a capability might be in a position to additionally prevent
+the original packets it observes from reaching their intended destination.  If
+so, they are considered to be an on-path attacker.
+
+An active attacker may also choose to rewrite the source or destination IP
+addresses of packets that it forwards or injects. Such spoofing attacks are only
+effective against a QUIC connection if the attacker can still forward the
+contents of the packets to the original endpoint, since QUIC connections are
+both authenticated and encrypted.
```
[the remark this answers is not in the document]
erickinnear: Ah, good call. This is intended to say that even if you mess with the routing of the packets by changing the addresses, the packet payloads ("contents of the packets" in the wording above) are only useful to the actual endpoints that completed the handshake.
This is likely already covered by other text, so we can probably just remove it.

draft-ietf-quic-transport.md
erickinnear: Looking at this in more detail, it looks like we can move much of the definition to the beginning of the whole section, which leaves this part to be much smaller and less nested. :)

draft-ietf-quic-transport.md
erickinnear: Fixed everywhere in 70dac1d

draft-ietf-quic-transport.md
erickinnear: Fixed in 70dac1d

draft-ietf-quic-transport.md
erickinnear: Fixed in 70dac1d

draft-ietf-quic-transport.md
erickinnear: Fixed everywhere in 70dac1d

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+
+A blind attacker, one who injects packets without being able to observe valid
+packets for a QUIC connection, is unlikely to be successful, since QUIC packet
+protection ensures that valid packets are only generated by endpoints which
```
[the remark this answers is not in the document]
erickinnear: Updated in 70dac1d

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+attackers having the capability to read packets from the network and active
+attackers having the capability to write packets into the network.
+
+Attackers are additionally categorized as either on-path attackers or off-path
+attackers (see Section 3.5 of {{?RFC3552}}); an on-path attacker is on the
+critical path for a given connection and can read, modify, or remove any packet
+it observes such that it no longer reaches its destination, while an off-path
+attacker observes the packets, but cannot prevent the original packet from
+reaching its intended destination. An off-path attacker can transmit arbitrary
+packets, and it may be able to attack the network so as to place itself on-path.
```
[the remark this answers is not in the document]
erickinnear: Done in 70dac1d

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+This section describes the intended security properties of connection migration
+when under various types of attack, as described in more detail by {{?RFC3552}}.
+
+For this purpose, attacks are divided into passive and active attacks, passive
+attackers having the capability to read packets from the network and active
+attackers having the capability to write packets into the network.
+
+Attackers are additionally categorized as either on-path attackers or off-path
+attackers (see Section 3.5 of {{?RFC3552}}); an on-path attacker is on the
+critical path for a given connection and can read, modify, or remove any packet
```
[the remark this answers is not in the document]
erickinnear: Removed in 70dac1d

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+
+This section describes the intended security properties of connection migration
+when under various types of attack, as described in more detail by {{?RFC3552}}.
+
+For this purpose, attacks are divided into passive and active attacks, passive
+attackers having the capability to read packets from the network and active
+attackers having the capability to write packets into the network.
+
+Attackers are additionally categorized as either on-path attackers or off-path
+attackers (see Section 3.5 of {{?RFC3552}}); an on-path attacker is on the
```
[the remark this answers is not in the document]
erickinnear: Changed in 70dac1d

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+
+### Connection Migration {#migration-properties}
+
+Connection Migration ({{migration}}) provides endpoints with the ability to
+transition between IP addresses and ports on multiple paths, using one path at a
+time for transmission and receipt of non-probing frames.  Path validation
+({{migrate-validate}}) establishes that a peer is both willing and able
+to receive packets sent on a particular path.  This helps reduce the effects of
+address spoofing, by limiting the number of packets sent to a spoofed address,
+unless an attacker is able to also receive packets sent to that address.
```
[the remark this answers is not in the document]
erickinnear: Done in 70dac1d

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+
+A passive attack ({{?RFC3552}}) involves observing and potentially saving
+packets being sent over the network, but does not involve sending any packets.
+However, a passive attack may involve an attacker with the ability to cause a
+routing change or other modification in the path taken by packets that comprise
+a QUIC connection.
+
+Both on-path and off-path attackers can mount a passive attack in which they
+save observed QUIC packets for an offline attack against QUIC packet protection
+at a future time; this is true for any observer of any packet on any network.
```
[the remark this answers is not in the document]
erickinnear: Moved this into short header section (and renamed that to packet protection, since technically that's not unique to SH even if this is not handshake-specific). Generally moved a bunch of text up, as well.

draft-ietf-quic-transport.md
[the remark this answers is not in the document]
erickinnear: Reworded and moved this around to hopefully be more clear

draft-ietf-quic-transport.md
erickinnear: Added text to be more clear about that

draft-ietf-quic-transport.md
erickinnear: Moved a bunch of this up into packet protection section

draft-ietf-quic-transport.md
erickinnear: Moved, trimmed out the duplication, now at the top of the entire section and provides the definitions for the whole thing. Then the sections below move up one and are how those definitions, etc. apply to migration. Much better now -- thanks!

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,252 @@ decisions are made independently of client-selected values; a Source Connection
...
+An active attacker may also choose to rewrite the source or destination IP
+addresses of packets that it forwards or injects. Such spoofing attacks are only
+effective against a QUIC connection if the attacker can still forward the
+contents of the packets to the original endpoint, since QUIC connections are
+both authenticated and encrypted.
+
+A blind attacker, one who injects packets without being able to observe valid
+packets for a QUIC connection, is unlikely to be successful, since QUIC packet
+protection ensures that valid packets are only generated by endpoints which
+possess the key material established during the handshake.  Similarly, any
```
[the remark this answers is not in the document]
erickinnear: Added

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,288 @@ decisions are made independently of client-selected values; a Source Connection
...
+address, is only effective if the attacker can forward packets to the original
+endpoint.  Packet protection ensures that the packet payloads can only be
+processed by the endpoints that completed the handshake, and invalid QUIC
+packets are ignored by those endpoints.
+
+An attacker can also modify the boundaries between QUIC packets and UDP
+datagrams, causing multiple packets to be coalesced into a single datagram, or
+splitting coalesced packets into multiple datagrams.  Such modification has no
+functional effect on a QUIC connection, although it might change the performance
+characteristics exhibited by the receiving endpoint.
```
MikeBishop: Good observation.  I remember having discussions about heuristics that assume packets sent in a single datagram arrived together.  This means those heuristics aren't reliable.  I don't recall whether we endorsed such approaches in the document (PMTU probing, maybe?), but this suggests that we might want to explicitly caution against them.  (Not necessarily here.)

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,288 @@ decisions are made independently of client-selected values; a Source Connection
...
+- Modify IP and UDP packet headers
+- Inject new packets
+- Delay packets
+- Reorder packets
+- Drop packets
+- Split and merge datagrams along packet boundaries
+
+An on-path attacker cannot:
+
+- Modify an authenticated and encrypted portion of a packet and cause the
```
```suggestion
- Modify an authenticated portion of a packet and cause the
```
The rejection of modified packets doesn't depend on the encryption, but the authentication.  The encryption itself is mostly irrelevant to this assertion.

draft-ietf-quic-transport.md
martinthomson: Of note here is that we allow Initial packets to be smaller than 1200 bytes, as long as they are coalesced with other stuff to make up a 1200 byte UDP datagram.  That too is vulnerable to this sort of composition problem.  I think that it's fine to make these assumptions and risk modification as long as they are recognized as such.  For Initial, we are doing so on the understanding that elements on the path are able to disrupt the handshake at that point anyway.  For PMTUD, we might make a different assessment as the conditions are different.

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,288 @@ decisions are made independently of client-selected values; a Source Connection
 ID can be selected to route later packets to the same server.
 
 
+## Overview of Security Properties {#security-properties}
+
+A complete security analysis of QUIC is outside the scope of this document.
+This section provides an informal description of the desired security properties
+as an aid to implementors and to help guide protocol analysis.
+
+QUIC provides protection against various types of attacks, as described in more
```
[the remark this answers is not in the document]
erickinnear: Nice! No worries, rewrapped :)

draft-ietf-quic-transport.md
```diff
@@ -5874,6 +5874,288 @@ decisions are made independently of client-selected values; a Source Connection
...
+An on-path attacker can force the QUIC handshake to fail by replacing either the
+client or server Initial messages with invalid ones.  An off-path attacker can
+also mount this attack by racing the Initials.  Once valid Initial messages have
+been exchanged, the remaining handshake messages are protected with the
+handshake keys and an on-path attacker cannot force handshake failure, though
+they can produce a handshake timeout by dropping packets.
+
+An on-path attacker can also replace the addresses of packets on either side and
+therefore cause the client or server to have an incorrect view of the remote
+addresses.
```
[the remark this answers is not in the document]
erickinnear: Yes, great idea! Added a sentence here indicating that an attack here is indistinguishable from the presence of a NAT. (We've got a similar note about some of that later, but this is a great place to start people thinking in that direction.)

draft-ietf-quic-transport.md
erickinnear: Good thinking, it might be worth a note around those rules in the Initial that such a thing is possible (and might be more likely than if you just padded the initial packet?). PMTUD is probably worth calling out a little bit more clearly -- I'll file an editorial issue for each, since that's way elsewhere.

draft-ietf-quic-transport.md
erickinnear: This seems like a nice balance with the mention of both authentication and encryption in the later paragraph, while the specific capability is only around the authenticated portion.

erickinnear:
Anything else we want to do before this is ready to land?

draft-ietf-quic-transport.md
```diff
@@ -5980,6 +5980,289 @@ decisions are made independently of client-selected values; a Source Connection
...
+A spoofing attack, in which an active attacker rewrites unprotected parts of a
+QUIC packet that it forwards or injects, such as the source or destination
+address, is only effective if the attacker can forward packets to the original
+endpoint.  Packet protection ensures that the packet payloads can only be
+processed by the endpoints that completed the handshake, and invalid QUIC
+packets are ignored by those endpoints.
+
+An attacker can also modify the boundaries between QUIC packets and UDP
+datagrams, causing multiple packets to be coalesced into a single datagram, or
+splitting coalesced packets into multiple datagrams.  Such modification has no
```
```suggestion
splitting coalesced packets into multiple datagrams.  Aside from datagrams containing Initial packets, which require padding, modification has no
```

draft-ietf-quic-transport.md
```diff
@@ -5980,6 +5980,289 @@ decisions are made independently of client-selected values; a Source Connection
...
+to receive packets sent on a particular path.  This helps reduce the effects of
+address spoofing by limiting the number of packets sent to a spoofed address.
+
+This section describes the intended security properties of connection migration
+when under various types of DoS attacks.
+
+#### On-Path Active Attacks
+
+An attacker that can cause a packet it observes to no longer reach its intended
+destination is considered an on-path attacker.  Such an attacker generally is
```
ianswett: generally is present is both a somewhat strong(implying >50% of the time) and quite vague statement.  Maybe this clause should be changed to: "When an attacker is present between the QUIC client and server, QUIC endpoints are required to send packets through the attacker to establish connectivity on a given path."?

draft-ietf-quic-transport.md
```diff
@@ -5980,6 +5980,289 @@ decisions are made independently of client-selected values; a Source Connection
...
+An on-path attacker has the opportunity to modify the packets that it observes,
+however any modifications to an authenticated portion of a packet will cause it
+to be dropped by the receiving endpoint as invalid, as QUIC payloads are both
+authenticated and encrypted.
+
+In the presence of an on-path attacker, QUIC aims to provide the following
+properties:
+
+1. An on-path attacker can prevent use of a path for a QUIC connection, causing
+it to fail if it cannot migrate to a new path that does not contain the
```
ianswett: I'm not sure it's necessary to use the word 'migrate' here.  If the handshake hasn't completed, you can't migrate yet.  If you take my suggestion, it probably makes sense to update the other bullets.

```suggestion
it to fail if it cannot use a different path that does not contain the
```

draft-ietf-quic-transport.md
```diff
@@ -5980,6 +5980,289 @@ decisions are made independently of client-selected values; a Source Connection
...
+potentially causing the original packet to be ignored by the recipient.
+
+It is also assumed that an attacker has the resources necessary to affect NAT
+state, potentially both causing an endpoint to lose its NAT binding, and an
+attacker to obtain the same port for use with its traffic.
+
+In the presence of an off-path attacker, QUIC aims to provide the following
+properties:
+
+1. An off-path attacker can race packets and attempt to become a "limited"
```
ianswett: The word race is used a few times in this section, but I don't believe it's defined and I'm not sure it's clear to everyone what it means in this context?

draft-ietf-quic-transport.md
```diff
@@ -5980,6 +5980,289 @@ decisions are made independently of client-selected values; a Source Connection
...
+on-path attacker.
+
+2. An off-path attacker can cause path validation to succeed for forwarded
+packets with the source address listed as the off-path attacker as long as it
+can provide improved connectivity between the client and the server.
+
+3. An off-path attacker cannot cause a connection to close.
```
```suggestion
3. An off-path attacker cannot cause a connection to close once the handshake has completed.
```

draft-ietf-quic-transport.md
```diff
@@ -5980,6 +5980,289 @@ decisions are made independently of client-selected values; a Source Connection
...
+A limited on-path attacker can only delay packets up to the point that the
+original packets arrive before the duplicate packets, meaning that it cannot
+offer routing with worse latency than the original path.  If a limited on-path
+attacker drops packets, the original copy will still arrive at the destination
+endpoint.
+
+In the presence of a limited on-path attacker, QUIC aims to provide the
+following properties:
+
+1. A limited on-path attacker cannot cause an active connection to close.
```
```suggestion
1. A limited on-path attacker cannot cause an established connection to close.
```

draft-ietf-quic-transport.md
```diff
@@ -5980,6 +5980,289 @@ decisions are made independently of client-selected values; a Source Connection
...
+being encrypted with per-version keys and the Handshake and later packets being
+encrypted with keys derived from the TLS key exchange.  Further, parameter
+negotiation is folded into the TLS transcript and thus provides the same
+security guarantees as ordinary TLS negotiation.  Thus, an attacker can observe
+the client's transport parameters (as long as it knows the QUIC version-specific
+salt) but cannot observe the server's transport parameters and cannot influence
+parameter negotiation.
+
+Connection IDs are unencrypted but integrity protected in all messages.  They
+are not incorporated in the TLS handshake transcript.
```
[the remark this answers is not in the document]
janaiyengar: I think this second sentence can be dropped.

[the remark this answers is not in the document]
martinthomson: Without further elaboration, I have to agree.  

Separately, I think it was Antoine who made some (good) points about authenticating connection IDs more thoroughly that we might want to consider.  Separately though (reminds self to open an issue).

draft-ietf-quic-transport.md
```diff
@@ -5980,6 +5980,289 @@ decisions are made independently of client-selected values; a Source Connection
...
+An on-path attacker can also replace the addresses of packets on either side and
+therefore cause the client or server to have an incorrect view of the remote
+addresses. Such an attack is indistinguishable from the functions performed by a
+NAT.
+
+#### Parameter Negotiation
+
+The entire handshake is cryptographically protected, with the Initial packets
+being encrypted with per-version keys and the Handshake and later packets being
+encrypted with keys derived from the TLS key exchange.  Further, parameter
```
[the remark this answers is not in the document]
martinthomson: Key exchange is correct (the key exchange depends on the message exchange, but it's the key exchange that matters here).

draft-ietf-quic-transport.md
martinthomson: It's first use would seem to work well enough to be definitional.  I'm happy with this as it is.

erickinnear:
Thanks Martin, and to all who reviewed! 
Definitely agree that further changes can be incremental editorial changes and that it's much nicer to land the main change.

draft-ietf-quic-transport.md
```diff
@@ -5980,6 +5980,289 @@ decisions are made independently of client-selected values; a Source Connection
...
+new connection establishment without incurring this cost.
+
+#### On-Path Handshake Termination
+
+An on-path attacker can force the QUIC handshake to fail by replacing either the
+client or server Initial messages with invalid messages.  An off-path attacker
+can also mount this attack by racing the Initials.  Once valid Initial messages
+have been exchanged, the remaining handshake messages are protected with the
+handshake keys and an on-path attacker cannot force handshake failure, though
+they can produce a handshake timeout by dropping packets.
```
mikkelfj: The on path attacker can also submit a MITM attack on the handshake.

erickinnear: That could be a great thing to cover, can you file an issue to discuss whatever text should be here?

mikkelfj: @erickinnear added https://github.com/quicwg/base-drafts/issues/3512

PR #2927, opened 2019-07-23
Allow Smaller Stateless Resets

### LINKED ISSUE #2869: Stateless Reset packet sizes should not depend on the maximum connection ID length
Currently, the spec recommends sending Stateless Reset packets of at least 39 bytes. The number is the minimum packet size when using the maximum connection ID length of 18:
- type byte: 1
- client connection ID: 18
- packet number + payload: 4
- auth tag: 16

39 = 1 + 18 + 4 + 16.

The intention there was to make Stateless Resets harder to distinguish from regular server-to-client traffic on a given QUIC connection.

However, when client connection IDs are in use, Stateless Resets are inherently distinguishable from regular traffic because they do not carry the previously-used client connection ID. One might say that they could be confused for a change in connection ID, which is true. But let's assume we just changed from an 18 byte client connection ID to a 4-byte connection ID (which is specifically allowed by the spec), then that new packet is still equally indistinguishable from a Stateless Reset. It is true that a client connection ID length change from 18 to 4 is unlikely, but that's not a problem. We just need it to be possible for this to be a valid packet to observers without keys.

So I propose that instead of having a minimum required size for stateless resets of 21 and a recommended minimum of 39, we can just say the minimum required size is 25. The important value here is the minimum non-zero connection ID length, not the maximum connection ID length.

At the end of the day, a Stateless Reset will stick out to observers who will notice that the client stops sending on that particular connection. The current value of 39 is harming the protocol by increasing the odds of servers not sending Stateless Resets which will cause timeouts on clients instead of failing fast. This is a tradeoff, and we should strive to avoid timeouts instead of trying to hide a packet that can't be hidden.

ianswett:
Allows smaller stateless resets to be sent(as small as 21 bytes) and advises endpoints that want to ensure a stateless reset is not detectable send a larger packet if their chosen CIDs are more than 2 bytes.

Fixes #2869, #2770


draft-ietf-quic-transport.md
```diff
@@ -2389,19 +2389,26 @@ of bytes following it that are set to unpredictable values.  The last 16 bytes
...
-Unpredictable Bits field needs to include at least 198 bits of data (or 25
-bytes, less the two fixed bits).  This is intended to allow for a Destination
-Connection ID of the maximum length permitted, with a minimal packet number, and
-payload.  The Stateless Reset Token corresponds to the minimum expansion of the
+to be a packet with a short header.  For the stateless reset to appear as a
+valid QUIC packet and be smaller than the received packet, the Unpredictable
+Bits field needs to include at least 46 bits of data (or 6 bytes, less the
+two fixed bits), because packets that are smaller than 21 bytes are never
+valid.  To ensure the stateless reset packet is not smaller than other
+packets received on the connection, the an endpoint SHOULD also ensure the
```
DavidSchinazi: typo: the an

draft-ietf-quic-transport.md
```diff
@@ -2389,19 +2389,26 @@ of bytes following it that are set to unpredictable values.  The last 16 bytes
...
+The Stateless Reset Token corresponds to the minimum expansion of the
 packet protection AEAD.  More unpredictable bytes might be necessary if the
 endpoint could have negotiated a packet protection scheme with a larger minimum
 AEAD expansion.
 
 An endpoint SHOULD NOT send a stateless reset that is significantly larger than
 the packet it receives.  Endpoints MUST discard packets that are too small to be
 valid QUIC packets.  With the set of AEAD functions defined in {{QUIC-TLS}},
-packets that are smaller than 21 bytes are never valid.
+packets that are smaller than 21 bytes are never valid.  When a packet is b
```
```suggestion
packets that are smaller than 21 bytes are never valid.
```
This looks like it was unintentional?

draft-ietf-quic-transport.md
```diff
@@ -2389,19 +2389,26 @@ of bytes following it that are set to unpredictable values.  The last 16 bytes
...
-Connection ID of the maximum length permitted, with a minimal packet number, and
-payload.  The Stateless Reset Token corresponds to the minimum expansion of the
+to be a packet with a short header.  For the stateless reset to appear as a
+valid QUIC packet and be smaller than the received packet, the Unpredictable
+Bits field needs to include at least 46 bits of data (or 6 bytes, less the
+two fixed bits), because packets that are smaller than 21 bytes are never
+valid.  To ensure the stateless reset packet is not smaller than other
+packets received on the connection, the an endpoint SHOULD also ensure the
+total packet length is at least the minimum chosen CID length + 19 bytes.
+19 bytes allows for 1 type byte, 1 data byte, the 16 byte authentication
```
DavidSchinazi: The invariant we have is `payload_length + packet_number_length >= 4` to which you add type byte (1) and auth tag (16) to get 1 + 4 + 16 = 21. I think we want 22 + chosen CID length

draft-ietf-quic-transport.md
DavidSchinazi: was this intentional?

draft-ietf-quic-transport.md

draft-ietf-quic-transport.md
```diff
@@ -2389,11 +2389,17 @@ of bytes following it that are set to unpredictable values.  The last 16 bytes
 of the datagram contain a Stateless Reset Token.
 
 To entities other than its intended recipient, a stateless reset will appear
-to be a packet with a short header.  For the packet to appear as valid, the
-Unpredictable Bits field needs to include at least 198 bits of data (or 25
-bytes, less the two fixed bits).  This is intended to allow for a Destination
-Connection ID of the maximum length permitted, with a minimal packet number, and
-payload.  The Stateless Reset Token corresponds to the minimum expansion of the
+to be a packet with a short header.  For the stateless reset to appear as a
+valid QUIC packet and be smaller than the received packet, the Unpredictable
```
MikeBishop: Needing "at least 48 bits" isn't a requirement to "be smaller than the received packet."  This addition doesn't compute.

draft-ietf-quic-transport.md
```diff
@@ -2389,11 +2389,17 @@ of bytes following it that are set to unpredictable values.  The last 16 bytes
...
+Bits field needs to include at least 48 bits of data (or 6 bytes, less the
+two fixed bits).  To ensure the stateless reset packet is not smaller than
+other packets received on the connection, an endpoint SHOULD also ensure the
+total packet length is at least the minimum chosen CID length + 22 bytes.
```
MikeBishop: Defer to @martinthomson on the styling, but perhaps spell out "plus" here.

draft-ietf-quic-transport.md
martinthomson: What Mike said.  I think that you want to say:

> A stateless reset MUST include an Unpredictable Bits field of at least 38 bits (or 5 bytes, less the two fixed bits).  However, this does not account for the length of the connection ID chosen by a peer, which is not known by the endpoint sending a stateless reset.  For a stateless reset to appear to be a valid QUIC packet, the packet needs to appear to contain one type byte, a minimum-sized connection ID, four bytes of packet number and payload, and the AEAD expansion. In order to ensure that a stateless reset is smaller than the packet that triggers it, an endpoint needs to send a packet that is one byte longer than the minimum stateless reset size.  Therefore, endpoints SHOULD always send packets that are at least 22 bytes larger than the minimum connection ID that they might select.
>
>These calculations assume a 16 byte AEAD expansion.  Smaller packets can be sent if the AEAD expansion is less than 16 bytes and larger packets need to be sent for a larger expansion.

draft-ietf-quic-transport.md
```diff
@@ -2367,7 +2367,7 @@ following layout:
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|0|1|               Unpredictable Bits (198..)                ...
+|0|1|               Unpredictable Bits (46 ..)                ...
```
```suggestion
|0|1|               Unpredictable Bits (38 ..)                ...
```

martinduke:
I don't understand what this is trying to specify. The stateless reset sender has no idea what CID length the peer has chosen (except for the 20 bytes limit), so I don't see how to guarantee that the Reset exceeds minimum size without allowing for the full 20 bytes.

Unless this PR is trying to say that it's OK to go smaller if you're willing to risk it...

martinthomson:
It's pretty much "it's OK to go smaller if you are willing to risk stateless reset messages being identifiable", but a little stronger than that.  The sender of a stateless reset is now no longer bound by the upper limit on connection ID size.  That allows for a smaller packet to be sent and still have it get a stateless reset in response.

However, the onus for ensuring that the resulting stateless reset is indistinguishable from regular packets moves to the endpoint that would receive the stateless reset.  It has to ensure that it never sends a packet small enough that a stateless reset packet would be too small to appear as valid.

For instance, take the case where a client needs at least 8 bytes of connection ID, but the server had no connection ID.  Though the client has no real need to send larger packets toward the server, it has to pad outbound packets so that the server, should it be forced to send a stateless reset, wouldn't end up sending a packet that appears to have a 7 byte connection ID (which, assuming an observer knows about the client's constraints, would be easy to identify as a stateless reset).

martinduke:
I find this text to be really unclear. Furthermore, if implementers don't pick up the SHOULD we're going to have detectable stateless resets all over the place.

And given the final version of extended connection IDs, we're talking about 2 additional bytes over the draft-20 status quo?

I don't find this change to be necessary, though I wouldn't raise a huge stink about it. However, I'd like to consider changing SHOULD to MUST. This seems like a potential security vulnerability.

At an absolute minimum, I'll propose an editorial rewrite to make these considerations clearer.

martinthomson:
In coming back to this, I can agree with Martin.  This is very unclear.

PR #2937, opened 2019-07-31
QPACK [editorial] Update text to include both criteria of "blocking entry".

bencebeky:
The definition of a blocking entry has two parts: insertion not acknowledged, or not all references acknowledged.  Update text in two places to include the part about acknowledging the insertion, lest a reader forget about it.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -229,9 +229,8 @@ table that it will emit a reference for.  As new entries are inserted, the
 encoder increases the draining index to maintain the section of the table that
 it will not reference.  If the encoder does not create new references to entries
 with an absolute index lower than the draining index, the number of
-unacknowledged references to every such entry will eventually become zero,
-allowing it to be evicted provided that its insertion has been acknowledged by
-the decoder.
+unacknowledged references to those entries will eventually become zero, allowing
+them to be evicted.
 
 ~~~~~~~~~~  drawing
    +----------+---------------------------------+--------+
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -229,8 +229,9 @@ table that it will emit a reference for.  As new entries are inserted, the
 encoder increases the draining index to maintain the section of the table that
 it will not reference.  If the encoder does not create new references to entries
 with an absolute index lower than the draining index, the number of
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -229,8 +229,9 @@ table that it will emit a reference for.  As new entries are inserted, the
 encoder increases the draining index to maintain the section of the table that
 it will not reference.  If the encoder does not create new references to entries
 with an absolute index lower than the draining index, the number of
-unacknowledged references to those entries will eventually become zero, allowing
-them to be evicted.
+unacknowledged references to every such entry will eventually become zero,
+allowing it to be evicted provided that its insertion has been acknowledged by
+the decoder.
```
afrind: This is now a pretty long sentence.  Maybe end it after 'become zero', and start a new sentence talking about eviction?

draft-ietf-quic-qpack.md
```diff
@@ -359,9 +360,9 @@ infer from this instruction that any updates to the dynamic table have been
 received.
 
 The Header Acknowledgement and Stream Cancellation instructions permit the
-encoder to remove references to entries in the dynamic table.  When entries
-have zero references they are no longer considered blocking (see
-{{blocked-insertion}}).
+encoder to remove references to entries in the dynamic table.  When an entry
+with absolute index lower than the Known Received Count has zero references,
```
afrind: Does this leave ambiguous what happens to entries with absolute indices higher than Known Received Count?

draft-ietf-quic-qpack.md
bencebeky: Ah, let me just revert the change to this paragraph.  No matter how I try to reword it, mentioning acknowledgement of the insertion always leaves it more confusing that the original text.  After all, if the number of unacknowledged references _becomes_ zero, that means it was not zero beforehand, therefore the encoder knows for sure that the insertion has been processed by the decoder.

draft-ietf-quic-qpack.md
bencebeky: I don't believe so.  It kind of implies that such an entry is considered blocking, which is true.  Note that if the entry is ever referenced in a header block and that reference is acknowledged, that guarantees that the Known Received Count is larger than the absolute index of the entry.

PR #2939, opened 2019-07-31
QPACK [design] Decoder MUST emit Header Acknowledgement.

bencebeky:
Add MUST to sentence about requirement on decoder to emit a Header Acknowledgement instruction every time it decodes a header block with non-zero Required Insert Count.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -337,7 +337,7 @@ The decoder signals the following events by emitting decoder instructions
 
 #### Completed Processing of a Header Block
 
-When the decoder finishes decoding a header block containing dynamic table
+After the decoder finishes decoding a header block containing dynamic table
 references, it MUST emit a Header Acknowledgement instruction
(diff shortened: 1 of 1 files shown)


bencebeky:
> This doesn't allow for a delay in sending an ack. That delay might be useful in reducing overheads. That might be a new issue though.

That's a good point, I haven't thought about it.  Would changing "When the decoder finishes decoding..." to "After the decoder finishes decoding..." be sufficient to allow for delayed ACKs?

martinthomson:
Works for me.

bencebeky:
> Works for me.

Done.  Thanks for raising this issue.

afrind:
You can't coalesce Header Acks like you can Insert Count Increment or transport Acks.  The only overhead savings comes from when these are sent to the peer - eg: transport overhead.  The text only says `emits...` the Ack, and doesn't really describe how quickly it has to be conveyed.  But I don't really have a problem with when vs. after either.

PR #2941, opened 2019-07-31
QPACK [editorial] Update text on instructions and representations.

bencebeky:
* Use more uniform and more specific language in description of instructions, like "follows" instead of 
  "includes" or implicit mention of a field.
* Add references to integer and string encodings where missing.
* Prefer internal references to external ones (also in Section 1.2).
* Explicitly allow a header block to not have any instructions after the mandatory prefix.
* Refer to Literal Header Field With Name Reference section in Literal Header Field Without Name Reference instead of verbatim repeating text on 'N' bit.
* Reshuffle sentences, insert references instead, remove sentences where references suffice.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -680,7 +680,7 @@ acknowledged as this instruction does not insert an entry.
 
 An encoder adds an entry to the dynamic table where the header field name
 matches the header field name of an entry stored in the static or the dynamic
-table using an instruction that begins with the '1' one-bit pattern.  The second
+table using an instruction that starts with the '1' one-bit pattern.  The second
 (`S`) bit indicates whether the reference is to the static or dynamic table. The
 6-bit prefix integer (see {{prefixed-integers}}) that follows is used to locate
 the table entry for the header name.  When S=1, the number represents the static
@@ -707,7 +707,7 @@ string literal (see {{string-literals}}).
 
 An encoder adds an entry to the dynamic table where both the header field name
 and the header field value are represented as string literals using an
-instruction that begins with the '01' two-bit pattern.
+instruction that starts with the '01' two-bit pattern.
 
 This is followed by the name represented as a 6-bit prefix string literal, and
 the value represented as an 8-bit prefix string literal (see
@@ -989,7 +989,8 @@ index of the entry in the dynamic table.
 If the entry is in the dynamic table with an absolute index greater than or
 equal to the Base, the representation starts with the '0001' 4-bit pattern,
 followed by the post-base index (see {{post-base}}) of the matching header
-field, represented as an integer with a 4-bit prefix (see {{prefixed-integers}}).
+field, represented as an integer with a 4-bit prefix (see
+{{prefixed-integers}}).
 
 ~~~~~~~~~~ drawing
   0   1   2   3   4   5   6   7
@@ -1063,8 +1064,8 @@ entry (see {{post-base}}) encoded as an integer with a 3-bit prefix (see
 The literal header field without name reference representation encodes a header
 field name and header field value as string literals (see {{primitives}}).
 
-This representation begins with the '001' three-bit pattern.  The fourth bit is the
-'N' bit as described in {{literal-name-reference}}.  The name follows,
+This representation begins with the '001' three-bit pattern.  The fourth bit is
+the 'N' bit as described in {{literal-name-reference}}.  The name follows,
 represented as a 4-bit prefix string literal, then the value, represented as an
 8-bit prefix string literal (see {{string-literals}}).
 
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -149,8 +149,8 @@ x (A)
 : Indicates that x is A bits long
 
 x (A+)
-: Indicates that x uses the prefixed integer encoding defined in Section 5.1 of
-  [RFC7541], beginning with an A-bit prefix.
+: Indicates that x uses the prefixed integer encoding defined in
+  {{prefixed-integers}}, beginning with an A-bit prefix.
 
 x ...
 : Indicates that x is variable-length and extends to the end of the region.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -798,13 +804,14 @@ as a 6-bit prefix integer.  See {{state-synchronization}}.
...
-duplications since the last Insert Count Increment or Header Acknowledgement
-that increased the Known Received Count for the dynamic table (see
-{{known-received-count}}).  The Increment field is encoded as a 6-bit prefix
-integer. The encoder uses this value to determine which table entries might
-cause a stream to become blocked, as described in {{state-synchronization}}.
+The Insert Count Increment instruction begins with the '00' two-bit pattern,
+followed by the Increment encoded as a 6-bit prefix integer.  The value of the
+Increment is the total number of dynamic table insertions and duplications
+processed by the decoder since the last time it sent a Header Acknowledgement
+instruction that increased the Known Received Count (see
```
bencebeky: I'm swapping Header Acknowledgement and Insert Count Instruction to avoid the possibility that somebody get the impression that an Insert Count Increment might not increase the Known Received Count.

draft-ietf-quic-qpack.md
```diff
@@ -821,10 +828,13 @@ connection error of type `HTTP_QPACK_DECODER_STREAM_ERROR`.
 
 ## Header Block Representations
 
-Header blocks contain compressed representations of header lists and are carried
```
bencebeky: Representations of header lists?  Representations of header fields?  Also, plural hides that blocks and lists have a one-to-one correspondance.  These are my motivations to reword.

draft-ietf-quic-qpack.md
```diff
@@ -1001,17 +1010,11 @@ If the entry is in the static table, or in the dynamic table with an absolute
 index less than the Base, this representation starts with the '01' two-bit
 pattern.
 
-Only the header field name stored in the static or dynamic table is used. Any
-header field value MUST be ignored.
-
 The following bit, 'N', indicates whether an intermediary is permitted to add
 this header to the dynamic header table on subsequent hops. When the 'N' bit is
-set, the encoded header MUST always be encoded with a literal representation. In
-particular, when a peer sends a header field that it received represented as a
```
bencebeky: I do not see any clarity or value that the "In particular.." sentence adds, it just rewords the previous sentence.  I propose to remove it.

draft-ietf-quic-qpack.md
```diff
@@ -821,10 +828,13 @@ connection error of type `HTTP_QPACK_DECODER_STREAM_ERROR`.
...
 ## Header Block Representations
 
-Header blocks contain compressed representations of header lists and are carried
-in frames on streams defined by the enclosing protocol.  These representations
-reference the static table, or dynamic table in a particular state without
-modifying it.
+A header block consists of a prefix and a possibly empty sequence of
+representations defined in this section.  Each representation corresponds to a
+single header field.  These representations can reference the static table, or
+the dynamic table in a particular state without modifying it.
```
afrind: Wondering if we still need 'without modifying it'.  True HPACK instructions both emitted header fields and modified the state, but it might be redundant in the context of this document, and it makes it read awkwardly.

draft-ietf-quic-qpack.md
```diff
@@ -968,8 +978,8 @@ decoded header list, as described in Section 3.2 of [RFC7541].
 If the entry is in the static table, or in the dynamic table with an absolute
 index less than the Base, this representation starts with the '1' 1-bit pattern,
 followed by the `S` bit indicating whether the reference is into the static or
-dynamic table.  The 6-bit prefix integer (see Section 5.1 of [RFC7541]) that
-follows is used to locate the table entry for the header name.  When S=1, the
+dynamic table.  The 6-bit prefix integer (see {{prefixed-integers}}) that
+follows is used to locate the table entry for the header field.  When S=1, the
```
afrind: nice catch!

draft-ietf-quic-qpack.md
afrind: It was copied verbatim from HPACK.  Maybe we should leave it in.

draft-ietf-quic-qpack.md
```diff
@@ -1026,20 +1030,21 @@ values that are not to be put at risk by compressing them (see Section 7.1 of
 {: title="Literal Header Field With Name Reference"}
 
 The fourth (`S`) bit indicates whether the reference is to the static or dynamic
-table.  The 4-bit prefix integer (see Section 5.1 of [RFC7541]) that follows is
+table.  The 4-bit prefix integer (see {{prefixed-integers}}) that follows is
 used to locate the table entry for the header name.  When S=1, the number
 represents the static table index; when S=0, the number is the relative index of
-the entry in the dynamic table.
+the entry in the dynamic table.  Only the header field name stored in the static
+or dynamic table is used. Any header field value MUST be ignored.
```
afrind: I feel like 'Any header field value MUST be ignored' might be confusing since it doesn't say the header field value from the static or dynamic table -- this representation also contains a value.

draft-ietf-quic-qpack.md
MikeBishop: The difference from HPACK seems worth noting, but perhaps the sentence could be reworded to flow better.  How about something like "These representations reference the static table and the dynamic table in a particular state, but do not modify the state of the dynamic table."

draft-ietf-quic-qpack.md
MikeBishop: I see what you mean about it being a restatement, but it's a ramification of "MUST always" that isn't immediately obvious.  If it's not stated, people will mess this up, I guarantee.  (They do anyway, because not all HPACK libraries tell you whether this bit was set on the way in.)

It's not my favorite feature of HPACK, but let's either keep the clarification or remove the requirement entirely.  If you want to discuss the latter, I would suggest doing so in a separate issue and PR.

draft-ietf-quic-qpack.md
MikeBishop: Or perhaps make an affirmative statement instead: "Only the header field name is taken from the static or dynamic table entry; the header field value is encoded as a string literal following the index."  There's currently no text that I see referring to the value portion of the instruction.

draft-ietf-quic-qpack.md
```diff
@@ -1026,20 +1030,21 @@ values that are not to be put at risk by compressing them (see Section 7.1 of
...
-equal to the Base, the representation starts with the '0000' four-bit
-pattern. The fifth bit is the 'N' bit as described in
-{{literal-name-reference}}.  Finally, the header field name is represented using
-the post-base index of that entry (see {{post-base}}) encoded as an integer with
-a 3-bit prefix.
+equal to the Base, the representation starts with the '0000' four-bit pattern.
+The fifth bit is the 'N' bit as described in {{literal-name-reference}}.
+Finally, the header field name is represented using the post-base index of that
+entry (see {{post-base}}) encoded as an integer with a 3-bit prefix (see
+{{prefixed-integers}}).
```
MikeBishop: Likewise, no mention here of the string literal for the value.  (Pre-existing problem, not one you introduced.)

PR #2942, opened 2019-07-31
QPACK [editorial] Misc minor editorial changes.

bencebeky:


draft-ietf-quic-qpack.md
```diff
@@ -84,7 +84,7 @@ code and issues list for this draft can be found at
 
 The QUIC transport protocol {{QUIC-TRANSPORT}} is designed to support HTTP
 semantics, and its design subsumes many of the features of HTTP/2 {{?RFC7540}}.
-HTTP/2 uses HPACK ({{!RFC7541}}) for header compression.  If HPACK were used for
```
bencebeky: No other reference in this paragraph is enclosed in parentheses.

draft-ietf-quic-qpack.md
```diff
@@ -192,7 +192,7 @@ while the decoder is relatively simple.
 
 An encoder MUST ensure that a header block which references a dynamic table
 entry is not processed by the decoder after the referenced entry has been
-evicted.  Hence the encoder needs to track information about each compressed
```
bencebeky: Once a header block is sent, this collection of information does not change.  This makes me believe "retain" is a more fitting word than "track".

draft-ietf-quic-qpack.md
```diff
@@ -272,25 +270,26 @@ An encoder can decide whether to risk having a stream become blocked. If
 permitted by the value of SETTINGS_QPACK_BLOCKED_STREAMS, compression efficiency
 can often be improved by referencing dynamic table entries that are still in
 transit, but if there is loss or reordering the stream can become blocked at the
-decoder.  An encoder avoids the risk of blocking by only referencing dynamic
```
bencebeky: "Can" better expresses that the encoder has a choice.

draft-ietf-quic-qpack.md
```diff
@@ -272,25 +270,26 @@ An encoder can decide whether to risk having a stream become blocked. If
...
+decoder.  An encoder can avoid the risk of blocking by only referencing dynamic
 table entries which have been acknowledged, but this could mean using
 literals. Since literals make the header block larger, this can result in the
 encoder becoming blocked on congestion or flow control limits.
 
 ### Known Received Count
 
-In order to identify which dynamic table entries can be safely used without a
-stream becoming blocked, the encoder tracks the number of entries received by
```
bencebeky: Reworded to include that the decoder also needs to track the Known Received Count.

draft-ietf-quic-qpack.md
```diff
@@ -272,25 +270,26 @@ An encoder can decide whether to risk having a stream become blocked. If
...
+order to be able to send Insert Count Increment instructions (see
+{{insert-count-increment}}).
 
-When blocking references are permitted, the encoder uses Header Acknowledgement
-instructions ({{header-acknowledgement}}) to maintain the Known Received
-Count.  If a header block was potentially blocking, the acknowledgement implies
-that the decoder has received all dynamic table state necessary to process the
-header block.  If the Required Insert Count of an acknowledged header block was
-greater than the encoder's current Known Received Count, the block's Required
-Insert Count becomes the new Known Received Count.
```
bencebeky: "the block's Required Insert Count becomes the new Known Received Count" always trips me: does the value of the Required Insert Count change to the value of the Known Received Count, or does the value of the Required Insert Count became to be known as the new value of the Known Received Count?

draft-ietf-quic-qpack.md
```diff
@@ -430,12 +431,9 @@ Dynamic table entries can have empty values.
 
 The size of the dynamic table is the sum of the size of its entries.
 
-The size of an entry is the sum of its name's length in bytes (as defined in
-{{string-literals}}), its value's length in bytes, and 32.
```
bencebeky: The reference to {{string-literals}} here is misleading, because length in that section can mean length after Huffman encoding.

draft-ietf-quic-qpack.md
```diff
@@ -445,19 +443,23 @@ encoder sends a Set Dynamic Table Capacity instruction
 ({{set-dynamic-capacity}}) with a non-zero capacity to begin using the dynamic
 table.
 
-Before a new entry is added to the dynamic table, entries are evicted from the
```
bencebeky: I am not aware of any reason the specs need to require an implementation to evict first, then add.  Spelling out a choice also allows the caution on use-after-free to be restricted to one of the two methods.

draft-ietf-quic-qpack.md
```diff
@@ -477,28 +479,27 @@ it can choose to use a lower dynamic table capacity (see
 
 For clients using 0-RTT data in HTTP/3, the server's maximum table capacity is
 the remembered value of the setting, or zero if the value was not previously
-sent.  When the client's 0-RTT value of the SETTING is 0, the server MAY set it
```
bencebeky: To me, zero looks better written in numbers, especially with that many occurrences of "non-zero" in the same section.

draft-ietf-quic-qpack.md
```diff
@@ -586,8 +588,10 @@ d = count of entries dropped
 ### Prefixed Integers
 
 The prefixed integer from Section 5.1 of [RFC7541] is used heavily throughout
-this document.  The format from [RFC7541] is used unmodified.  QPACK
-implementations MUST be able to decode integers up to 62 bits long.
+this document.  The format from [RFC7541] is used unmodified.  Note, however,
+that QPACK uses some prefix sizes not actually used in HPACK.
```
bencebeky: I was actually tripped by this when re-using the HPACK implementation, because that had DCHECKs on prefix length that restricted it to values that occur in HPACK.  And of course there was no test coverage for other prefixes.  In particular, (3+) and (8+) occur in the current QPACK draft but not in RFC7541.

draft-ietf-quic-qpack.md
```diff
@@ -627,14 +632,14 @@ type HTTP_WRONG_STREAM_COUNT. These streams MUST NOT be closed. Closure of
 either unidirectional stream type MUST be treated as a connection error of type
 HTTP_CLOSED_CRITICAL_STREAM.
 
-An endpoint MAY avoid creating its own encoder stream if it's not going to be
```
bencebeky: I do not have the impression that HTTP/3 defines ownership of unidirectional streams, but "create" is unambiguous anyway.

draft-ietf-quic-qpack.md
```diff
@@ -627,14 +632,14 @@ type HTTP_WRONG_STREAM_COUNT. These streams MUST NOT be closed. Closure of
...
-used (for example if the endpoint doesn't wish to use the dynamic table, or if
```
bencebeky: Each endpoint has two contexts and each context has a dynamic table, so it's better to clarify which one.

draft-ietf-quic-qpack.md
```diff
@@ -627,14 +632,14 @@ type HTTP_WRONG_STREAM_COUNT. These streams MUST NOT be closed. Closure of
...
-the maximum size of the dynamic table permitted by the peer is zero).
+An endpoint MAY avoid creating an encoder stream if it's not going to be used
+(for example if its encoder doesn't wish to use the dynamic table, or if the
+maximum size of the dynamic table permitted by the peer is zero).
 
-An endpoint MAY avoid creating its own decoder stream if the maximum size of
-its own dynamic table is zero.
```
bencebeky: I find that ownership of a dynamic table, while not necessarily ambiguous, is at least undefined.

draft-ietf-quic-qpack.md
```diff
@@ -627,14 +632,14 @@ type HTTP_WRONG_STREAM_COUNT. These streams MUST NOT be closed. Closure of
...
+An endpoint MAY avoid creating a decoder stream if its decoder sets the maximum
+capacity of the dynamic table to zero.
 
-An endpoint MUST allow its peer to create both encoder and decoder streams
```
bencebeky: "both encoder and decoder streams" might give the reader the impression that there could be multiple of each.

draft-ietf-quic-qpack.md
```diff
@@ -741,20 +746,16 @@ the existing entry is represented as an integer with a 5-bit prefix.
 ~~~~~~~~~~
 {:#fig-index-with-duplication title="Duplicate"}
 
-The existing entry is re-inserted into the dynamic table without resending
-either the name or the value. This is useful to mitigate the eviction of older
```
bencebeky: I find "mitigate the eviction" confusing.

draft-ietf-quic-qpack.md
```diff
@@ -742,19 +747,15 @@ the existing entry is represented as an integer with a 5-bit prefix.
...
-entries which are frequently referenced, both to avoid the need to resend the
-header and to avoid the entry in the table blocking the ability to insert new
-headers.
+either the name or the value. This is useful to avoid adding a reference to an
+older entry, which might block inserting new entries.
 
 
 ## Decoder Instructions {#decoder-instructions}
 
-Decoder instructions provide information used to ensure consistency of the
```
bencebeky: I propose to change this paragraph to match the first paragraph of Encoder Instructions section.

draft-ietf-quic-qpack.md
```diff
@@ -742,19 +747,15 @@ the existing entry is represented as an integer with a 5-bit prefix.
...
-dynamic table. They are sent from the decoder to the encoder on a decoder
-stream; that is, the server informs the client about the processing of the
```
bencebeky: This discussion of client and server seems out of place.  If it is really necessary, Section 2 might be a better place to talk about the two contexts and how both endpoint needs to have an encoder and a decoder.

draft-ietf-quic-qpack.md
```diff
@@ -814,7 +815,7 @@ cause a stream to become blocked, as described in {{state-synchronization}}.
 ~~~~~~~~~~
 {:#fig-size-sync title="Insert Count Increment"}
 
-An encoder that receives an Increment field equal to zero or one that increases
```
bencebeky: "zero or one" gets the reader's brain in the wrong groove.

draft-ietf-quic-qpack.md
```diff
@@ -830,10 +831,8 @@ modifying it.
 
 Each header block is prefixed with two integers.  The Required Insert Count is
 encoded as an integer with an 8-bit prefix after the encoding described in
-{{ric}}).  The Base is encoded as sign-and-modulus integer, using a single sign
```
bencebeky: A cursory web search shows that "sign-and-magnitude" is used a lot more often than "sign-and-modulus", which might conjure images of modulo arithmetic.  I propose not to put a name on this encoding, since there's an off-by-one that's not common for sign-and-modulus representations anyway.

draft-ietf-quic-qpack.md
```diff
@@ -937,11 +939,11 @@ entries, the Base will be greater than the Required Insert Count, so the delta
 will be positive and the sign bit is set to 0.
 
 An encoder that produces table updates before encoding a header block might set
-Required Insert Count and the Base to the same value.  In such case, both the
```
bencebeky: Rewording to emphasize that the encoder has very limited freedom in what to set Required Insert Count to.

draft-ietf-quic-qpack.md
```diff
@@ -993,7 +993,7 @@ field, represented as an integer with a 4-bit prefix (see Section 5.1 of
 
 ### Literal Header Field With Name Reference {#literal-name-reference}
 
-A literal header field with a name reference represents a header where the
```
bencebeky: Removing "a" to be consistent with subsection title.

draft-ietf-quic-qpack.md
```diff
@@ -1119,7 +1119,7 @@ HTTP_QPACK_DECODER_STREAM_ERROR (0x202):
 
 # Security Considerations
 
-TBD.
+TBD.  Also see Section 7.1 of [RFC7541].
```
bencebeky: This allows the reference to Section 7.1 in Literal Header Field With Name Reference section be replaced with an internal reference to this section.

draft-ietf-quic-qpack.md
```diff
@@ -272,25 +270,26 @@ An encoder can decide whether to risk having a stream become blocked. If
...
 
 ### Known Received Count
 
-In order to identify which dynamic table entries can be safely used without a
-stream becoming blocked, the encoder tracks the number of entries received by
-the decoder.  The Known Received Count tracks the total number of acknowledged
-insertions.
+The Known Received Count is the total number of dynamic table insertions and
+duplications acknowledged by the decoder.  The encoder tracks the Known Received
+Count in order to identify which dynamic table entries can be references without
```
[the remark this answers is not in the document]
bencebeky: Thank you.

draft-ietf-quic-qpack.md
```diff
@@ -445,19 +443,23 @@ encoder sends a Set Dynamic Table Capacity instruction
...
-is larger than the dynamic table capacity; the decoder MUST treat this as a
+When a new entry is added to the dynamic table, the minimum number of entries
+are evicted from the end of the dynamic table so that the final table size
+including the new entry does not exceed the table capacity.  This can be
+achieved by adding the new entry and then removing entries from the end until
+the table size is less than or equal to the table capacity.  Another option is
+to first evict entries until the size of the dynamic table is less than or equal
+to (table capacity - size of new entry), then add the new entry.  Note that a
+new entry can reference an entry in the dynamic table that will be evicted when
+adding this new entry.  Therefore implementations utilizing the latter option
```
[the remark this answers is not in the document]
bencebeky: Thank you.

draft-ietf-quic-qpack.md
```diff
@@ -830,10 +831,8 @@ modifying it.
 
 Each header block is prefixed with two integers.  The Required Insert Count is
 encoded as an integer with an 8-bit prefix after the encoding described in
-{{ric}}).  The Base is encoded as sign-and-modulus integer, using a single sign
-bit and a value with a 7-bit prefix (see {{base}}).
-
-These two values are followed by representations for compressed headers.
+{{ric}}).  The Base is encoded as a sign bit, and a Delta Base value with a
```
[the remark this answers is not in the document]
bencebeky: Thank you.

draft-ietf-quic-qpack.md
bencebeky: Just for the record, I do not feel strongly, and I'm happy to incorporate whatever @afrind suggests.

draft-ietf-quic-qpack.md
bencebeky: Yes, thank you for the suggestion.

draft-ietf-quic-qpack.md
```diff
@@ -445,19 +443,23 @@ encoder sends a Set Dynamic Table Capacity instruction
...
 
-Before a new entry is added to the dynamic table, entries are evicted from the
-end of the dynamic table until the size of the dynamic table is less than or
-equal to (table capacity - size of new entry). The encoder MUST NOT evict a
-blocking dynamic table entry (see {{blocked-insertion}}).  The entry is then
-added to the table.  It is an error if the encoder attempts to add an entry that
-is larger than the dynamic table capacity; the decoder MUST treat this as a
+When a new entry is added to the dynamic table, the minimum number of entries
+are evicted from the end of the dynamic table so that the final table size
+including the new entry does not exceed the table capacity.  This can be
```
afrind: This is an awfully long sentence without punctuation.  Maybe set off 'including the new entry' with commas?

draft-ietf-quic-qpack.md
```diff
@@ -445,19 +443,23 @@ encoder sends a Set Dynamic Table Capacity instruction
...
+achieved by adding the new entry and then removing entries from the end until
+the table size is less than or equal to the table capacity.  Another option is
+to first evict entries until the size of the dynamic table is less than or equal
+to (table capacity - size of new entry), then add the new entry.  Note that a
```
afrind: Actually, this language comes directly from HPACK.  Why is it necessary to offer the add+remove option, instead of remove+add?  There isn't normative language here, so an implementation can do it either way and still be compliant.

draft-ietf-quic-qpack.md
```diff
@@ -603,7 +607,8 @@ is used without modification.
 This document expands the definition of string literals and permits them to
 begin other than on a byte boundary.  An "N-bit prefix string literal" begins
 with the same Huffman flag, followed by the length encoded as an (N-1)-bit
-prefix integer.  The remainder of the string literal is unmodified.
+prefix integer.  The prefix size, N, can have a value between 2 and 8.  The
```
afrind: I didn't see a 2-bit prefix anywhere, should this be 3 and 8 (inclusive)?

But I see this as a bit of overkill.  If we allow 1 or 2 in QPACK2 we have to remember to change this.

draft-ietf-quic-qpack.md
```diff
@@ -937,11 +939,11 @@ entries, the Base will be greater than the Required Insert Count, so the delta
...
 
 An encoder that produces table updates before encoding a header block might set
-Required Insert Count and the Base to the same value.  In such case, both the
-sign bit and the Delta Base will be set to zero.
+Base to the value of Required Insert Count.  In such case, both the sign bit and
+the Delta Base will be set to zero.
 
 A header block that does not reference the dynamic table can use any value for
-the Base; setting Delta Base to zero is the most efficient encoding.
+the Base; setting Delta Base to zero is one of the most efficient encodings.
```
afrind: It's true there are other 2 byte encodings, but I think we should encourage 0/0.  Maybe we should even require it?

draft-ietf-quic-qpack.md
```diff
@@ -952,10 +954,8 @@ entry.
 
 ### Indexed Header Field
 
-An indexed header field representation identifies an entry in either the static
-table or the dynamic table and causes that header field to be added to the
-decoded header list, as described in Section 3.2 of [RFC7541].
-<!-- is the 7541 reference still helpful here -->
+An indexed header field instruction identifies an entry in either the static
```
afrind: Why did you change this from representation to instruction?  Though this isn't currently spelled out directly, the encoder/decoder streams contain "instructions" - which cause the receiver to take an action.  Header blocks contain representations, which emit headers fields.  The word instruction doesn't appear in this section.

The other representations don't seem to call themselves such, however.  And only this one contains the language about adding to the decoded header list.  I guess I prefer emit also, but maybe we need to make another consistency pass.

draft-ietf-quic-qpack.md
bencebeky: For comparison, RFC7541 5.1 specificly allows the prefix length to be between 1 and 8, inclusive, even though prefix lengths 1, 2, and 8 are never used in HPACK.  And that's great because now QPACK does not have to call out that the prefix can be 8 bits (for example, for Encoder Required Insert Count).  So I suggest to be as permissive here as reasonable.

If we allow N to be 1, then after the Huffman bit, the length is encoded as an 8-bit prefix integer, and not an N-1 = 0 bit prefix integer.  Is this worth adding here given that it's not used?

draft-ietf-quic-qpack.md
bencebeky: It's a good point that this section is not normative.  Let me just revert most of my changes to this paragraph for sake of simplicity.

draft-ietf-quic-qpack.md
afrind: Yeah, I guess zero pairs better with non-zero?  I don't think it makes a huge difference either way.

draft-ietf-quic-qpack.md
afrind: Ah, right, N=1 doesn't make sense.  2 and 8 inclusive?

bencebeky: Okay.  I guess it is the word "inclusive" that you are missing, so I'm adding it.

It is interesting (but irrelevant) to note that RFC7541 5.1 does not explicitly include the word "inclusive", though the next sentence implies that 8 is allowed:
"The prefix size, N, is always between 1 and 8 bits. An integer starting at an octet boundary will have an 8-bit prefix."

draft-ietf-quic-qpack.md
bencebeky: Thanks for chiming in.  In that case I'm keeping this change.

draft-ietf-quic-qpack.md
```diff
@@ -587,8 +585,10 @@ d = count of entries dropped
 ### Prefixed Integers
 
 The prefixed integer from Section 5.1 of [RFC7541] is used heavily throughout
-this document.  The format from [RFC7541] is used unmodified.  QPACK
-implementations MUST be able to decode integers up to 62 bits long.
+this document.  The format from [RFC7541] is used unmodified.  Note, however,
+that QPACK uses some prefix sizes not actually used in HPACK.
+
+QPACK implementations MUST be able to decode integers up to 62 bits long.
```
dtikhonov: As long as we are being pedantic: "up to *and including*?"

draft-ietf-quic-qpack.md
bencebeky: Created issue #3133 as the "another consistency pass".

draft-ietf-quic-qpack.md
bencebeky: Created issue #3134

PR #2948, opened 2019-08-01
Fix nit in protocol overview in TLS draft

nharper:
Mention that application data could be conveyed in other frame types
than STREAM frames. (E.g. the proposed DATAGRAM draft sends application
data in DATAGRAM frames.)


martinthomson:
This was correct at the time, because only STREAM carried application semantics, but that's too constraining, even now.  Thanks.

PR #2950, opened 2019-08-02
Remove mention of "protocol version selection" from TLS draft

nharper:


nharper:
I'm pretty sure this was obsoleted when version negotiation was removed (along with initial/negotiated/supported versions from transport parameters). If you think the TLS handshake is still validating protocol version selection, feel free to reject this PR.

PR #2954, opened 2019-08-06
Clarify out of order acking

janaiyengar:


rpaulo:
> What about the case where you receive an in order ack eliciting packet and then an out of order ack only packet? Should you still rely on delayed ack timer or send an ack immediately?

Yes, this was going to be my follow up question.

Also @janaiyengar what about the ECN paragraph just below it?

janaiyengar:
@nibanks: The current spec would not cause an ack to be triggered if non-ack-eliciting packets are received out of order. I think that's important. If the packet that was lost was another ack-only packet, this would cause unnecessary acks, and doing this at both ends could cause a minor ack storm.

@marten-seemann: out-of-order is defined in the draft above. I think it's less than ideal, but it's not ambiguous.

@rpaulo: I've rephrased the ECN section just a bit to be clearer. Does that help?

janaiyengar:
I'll merge this. If folks have any pending questions, happy to take them on in a new PR.

PR #2958, opened 2019-08-08
Move server's SETTINGS to transport parameters

### LINKED ISSUE #2790: Binding settings into session tickets
Session tickets might arrive before SETTINGS.  If so, then clients will need to understand whether those settings apply when they resume.

I think that the right answer is to say that the first (and only?) SETTINGS is bound to all session tickets.  Clients can then choose whether to remember settings with tickets and whether they wait for settings to arrive before saving tickets.

nharper:
Fixes #2790 and #2945 

Changes the outcome of #2985


nharper:
I think I've addressed the nits and updated QPACK.

Security/privacy considerations still need to be updated in HTTP and transport drafts.

nharper:
Closing this PR since #3086 is being punted to v2.

PR #2960, opened 2019-08-08
CRYPTO frames can not be sent in 0-RTT

### LINKED ISSUE #2959: Can CRYPTO frames be sent in 0-RTT packets?
In quic-transport-22 section 19.6 it says CRYPTO frames can be sent in all packet types.
> The CRYPTO frame (type=0x06) is used to transmit cryptographic handshake messages.  It can be sent in all packet types.

However in quic-tls-22 section 4 it says the CRYPTO frames can't be sent in 0-RTT packets.
> CRYPTO and CONNECTION_CLOSE frames MAY appear in packets of any encryption level except 0-RTT.

Also in quic-tls-22 section 8.3, similar statement.
> A server MUST treat receipt of a CRYPTO frame in a 0-RTT packet as a connection error of type PROTOCOL_VIOLATION.

Maybe the statement in quic-transport-22 needs to be corrected?

tandf:
Fixes #2959

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -4684,11 +4684,11 @@ Application Error Code:
 ## CRYPTO Frame {#frame-crypto}
 
 The CRYPTO frame (type=0x06) is used to transmit cryptographic handshake
-messages. It can be sent in all packet types. The CRYPTO frame offers the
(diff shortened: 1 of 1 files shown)


martinthomson:
Thanks! Beat me to it.

PR #2961, opened 2019-08-09
Repair rules for NEW_TOKEN

### LINKED ISSUE #2871: Retransmission Instructions do not include NEW_TOKEN
Sec 13.2 of QUIC Transport explains what to do for the loss of a packet for every stream type except NEW_TOKEN.

IMO it's nice but not critical to repair this loss, but I care less about how we resolve it than that we write it up.

martinthomson:
Closes #2871.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -4762,6 +4762,10 @@ Token:
 
 : An opaque blob that the client may use with a future Initial packet.
 
+An endpoint might receive multiple NEW_TOKEN frames that contain the same token
+value.  Endpoints are responsible for discarding duplicate values, which might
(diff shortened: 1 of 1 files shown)


martinthomson:
Is that not implied by the text here?

marten-seemann:
Kind of, but this text talks about the sender's perspective.
Furthermore, checking for duplicates is inherently racy if you're using tokens for new connections while the old connection is still alive.

martinthomson:
You know, I think that I put some proposed fixes for this in c016639e39c with rather than this branch.  Moved it over here.  @marten-seemann, you approved #2962, so I hope that moving that commit over here works for you.

PR #2962, opened 2019-08-09
Talk more about multiple tokens

### LINKED ISSUE #2801: describe how a client is supposed to handle multiple tokens
The spec doesn't define what is a client is supposed to do when it receives multiple NEW_TOKEN frames.

There seem to be multiple reasonable ways to handle this situation:
1. A client could always use the oldest token first. This would allow the client to establish multiple connections at the same time.
2. A client could replace any old tokens when receiving a new token. This has the advantage that the most recently issued token will (presumably) have the longest life time. Furthermore, if the server encodes connection statistics (RTT, cwnd, etc.) into the ticket, using the newest ticket would provide it with the most recent data.

From the Slack discussion, a better way might be a combination of these two:
When establishing a new connection, use the most recently obtained token first. When establishing another connection to the same server (before receiving a new NEW_TOKEN), use an older token. Client might then want to remember *n* tokens per endpoint, if they expect to establish *n* new connections simultaneously (or a few more than *n*, depending on the network conditions, to account for failed connection establishments).

martinthomson:
This addresses some questions about use of multiple tokens, as well as
cleaning up the text adjacent in minor ways.

Closes #2801.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1602,8 +1602,9 @@ one.  The client MUST NOT use the token provided in a Retry for future
 connections. Servers MAY discard any Initial packet that does not carry the
 expected token.
 
-A token SHOULD be constructed for the server to easily distinguish it from
-tokens that are sent in Retry packets as they are carried in the same field.
+A token SHOULD be constructed in a way that allows the server to distinguish it
+from tokens that are sent in Retry packets as they are carried in the same
+field.
 
 The token MUST NOT include information that would allow it to be linked by an
 on-path observer to the connection on which it was issued.  For example, it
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1632,13 +1633,22 @@ NEW_TOKEN frame.  A token obtained in a Retry packet MUST be used immediately
...
-believes that its point of network attachment has changed since the token was
-last used; that is, if there is a change in its local IP address or network
-interface.  A client needs to start the connection process over if it migrates
-prior to completing the handshake.
+A client SHOULD NOT reuse a token in different connections.  Reusing a token
+allows connections to be linked by entities on the network path; see
+{{migration-linkability}}.  A client MUST NOT reuse a token if it believes that
+its point of network attachment has changed since the token was last used; that
+is, if there is a change in its local IP address or network interface.  A client
+needs to start the connection process over if it migrates prior to completing
```
ianswett: I think it'd be preferable to avoid the word 'migrate' here, and reuse the preceding text: "if there is a change in its local IP address or network interface."?

draft-ietf-quic-transport.md
```diff
@@ -1632,13 +1633,22 @@ NEW_TOKEN frame.  A token obtained in a Retry packet MUST be used immediately
...
+the handshake.
+
+Clients might receive multiple tokens on a single connection.  Aside from
+preventing linkability, any token can be used in connection attempt.  Servers
```
```suggestion
preventing linkability, any token can be used in any connection attempt.  Servers
```

I think this is what you meant?

PR #2967, opened 2019-08-15
Preface QUIC frames with "QUIC"

tandf:


martinthomson:
This looks like the right text, but it is failing CI because it needs reflowing.

ianswett:
Was this text rewritten in #2857?  I thought the intent was to no longer reference transport frames in HTTP once that was merged.

PR #2968, opened 2019-08-15
Stateless reset too

### LINKED ISSUE #2669: Only received Stateless reset token should be checked
The current text does not explicitly say that endpoint has to check the stateless reset token received in NEW_CONNECTION_ID (and server's TP if endpoint is a client).

> The endpoint MUST
   then compare the last 16 bytes of the packet with all Stateless Reset
   Tokens that are associated with connection IDs that are currently in
   use.  This includes Stateless Reset Tokens from NEW_CONNECTION_ID
   frames and the server's transport parameters.

With this wording, I was under the impression that server can use stateless reset token in its own TP, but it seems not.

I think it would be nice to state that only received stateless reset tokens are checked.

martinthomson:
This forbids the use of duplicate stateless reset tokens.

It also cleans up some of the related language about matching of tokens.

Closes #2669, #2785.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2456,8 +2456,8 @@ the packet other than the last 16 bytes for carrying data.
 An endpoint detects a potential stateless reset when an incoming packet either
 cannot be associated with a connection, cannot be decrypted, or is marked as a
 duplicate packet.  The endpoint MUST then compare the last 16 bytes of the
-packet with all Stateless Reset Tokens that are associated with connection IDs
-that the endpoint recently used to send packets from the IP address and port on
+packet with all Stateless Reset Tokens corresponding to active connection IDs
+that the endpoint has used for sending packets to the IP address and port on
 which the datagram is received.  This includes Stateless Reset Tokens from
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -2353,12 +2353,14 @@ endpoint that wishes to communicate a fatal connection error MUST use a
...
 To support this process, a token is sent by endpoints.  The token is carried in
-the NEW_CONNECTION_ID frame sent by either peer, and servers can specify the
-stateless_reset_token transport parameter during the handshake (clients cannot
-because their transport parameters don't have confidentiality protection).  This
-value is protected by encryption, so only client and server know this value.
-Tokens are invalidated when their associated connection ID is retired via a
-RETIRE_CONNECTION_ID frame ({{frame-retire-connection-id}}).
+the Stateless Reset Token field of a NEW_CONNECTION_ID frame.  Servers can also
+specify a stateless_reset_token transport parameter during the handshake that
+applies to the connection ID that it select during the handshake; clients cannot
```
```suggestion
applies to the connection ID that it selected during the handshake; clients cannot
```

janaiyengar:
@DavidSchinazi : I expect that the chairs will put out a consensus call soon.

PR #2973, opened 2019-08-16
Require 8164 validation for non-https origins

### LINKED ISSUE #2439: Clarify use of HTTP request with http scheme over HTTP/3
It would be good if it could be clarified what will happen if someone needs/tries to send a HTTP request with an http:// uri over H3. I primarily see this to occur when you have a case where you have an HTTP proxy that receives an HTTP request and then have an HTTP/3 connection with the authorative server. If the expectation is that all requests are upgraded by the HTTP/3 sending side, then that needs to be made explicit.

MikeBishop:
Fixes #2439 in what I believe is the simplest way possible -- point to 8164 and require that `http-opportunistic` be successfully retrieved prior to sending any scheme other than `https`.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -381,6 +381,10 @@ certificate for the origin before considering it authoritative. Clients MUST NOT
 assume that an HTTP/3 endpoint is authoritative for other origins without an
 explicit signal.
 
+If the client intends to make requests for an origin containing a scheme other
+than "https", it MUST also obtain a valid `http-opportunistic` response for the
+origin as described in {{!RFC8164}} prior to making any such requests.
+
 A server that does not wish clients to reuse connections for a particular origin
 can indicate that it is not authoritative for a request by sending a 421
 (Misdirected Request) status code in response to the request (see Section 9.1.2


draft-ietf-quic-http.md
```diff
@@ -381,6 +381,10 @@ certificate for the origin before considering it authoritative. Clients MUST NOT
 assume that an HTTP/3 endpoint is authoritative for other origins without an
 explicit signal.
 
+If the client intends to make requests for an origin containing a scheme other
+than "https", it MUST also obtain a valid `http-opportunistic` response for the
```
martinthomson: I think that this requirement only applies to the "http" scheme.  It would not apply to the "ni" scheme, for instance.

MikeBishop: I'm of split opinion on that.  On the one hand, "http-opportunistic" is defined for "http"; on the other hand, you still really want to know that the connection is capable of handling a non-https URI scheme before you send it.

martinthomson: Yes, but the point of this is to point at a specific mechanism, which only applies to "http".  I'd be OK with leaving an impression that other protocols might need similar checks, that might be OK with me, but coming right out and saying that they definitely do is a bit of a big deal.

Saying that they have to use the HTTP mechanism is just wrong.  Part of the way in which that mechanism was designed was in response to HTTP-specific problems.  I'm sure that HTTP isn't unique in this regard, but it doesn't mean that you can't just start using "new-uri-scheme" without additional safeguards if that scheme was designed that way.

mikkelfj: I agree to what @martinthomson says. Unrelated and yet related I just fought (and lost) the Google Layer 7 load balancer because it insisted doing http health checks on a port that speaks websockets but does not serve http pages on the same port. In my case a tcp health check would have worked, but the point is that assuming that everything is http is not good.

MikeBishop: Try this version.

PR #2974, opened 2019-08-16
Default values in IANA

### LINKED ISSUE #2442: Put Initial Value of settings in IANA Considerations
Not a new issue but: RFC 7540 has a column for `Initial Value`. I think it makes sense that HTTP/3 does similar.

_Originally posted by @LPardue in https://github.com/quicwg/base-drafts/pull/2437/commits/f06211da71550e9a7ed03cc42d96ec5dc84dd3f0_

MikeBishop:
Fixes #2442.


draft-ietf-quic-http.md
```diff
@@ -1839,18 +1839,22 @@ Specification:
...
-| NUM_PLACEHOLDERS             |  0x9   | {{settings-parameters}}   |
-| ---------------------------- | ------ | ------------------------- |
+| ---------------------------- | ------ | ------------------------- | ------- |
+| Setting Name                 |  Code  | Specification             | Default |
+| ---------------------------- | :----: | ------------------------- | ------- |
+| Reserved                     |  0x2   | N/A                       | N/A     |
+| Reserved                     |  0x3   | N/A                       | N/A     |
+| Reserved                     |  0x4   | N/A                       | N/A     |
+| Reserved                     |  0x5   | N/A                       | N/A     |
+| MAX_HEADER_LIST_SIZE         |  0x6   | {{settings-parameters}}   | 0       |
```
LPardue: This is value is defined elsewhere to be unlimited, RFC 7540 used (infinite) but I'm not fussy.

ianswett: If you want to make SETTINGS REALLY required, making this default to 0 would certainly do it :)

ghedo: Well, not really as endpoints are allowed to ignore this value :)

But yeah https://quicwg.org/base-drafts/draft-ietf-quic-http.html#rfc.section.7.2.5.1 says the default value is unlimited, so this should be aligned to that (though I'd much rather remove this setting completely as it serves no purpose...).

MikeBishop: Bah, yes.  I was just following "SHOULD be the most restrictive possible value" on autopilot, but this is an exception (because it's advisory).

martinthomson: Would 2^62-1 be less offensive?

MikeBishop:
I went with "Default" because that and "Initial Value" now have specific, related (but distinct) usages in the context of pre-SETTINGS settings.  Swapping them seems like it would create unnecessary confusion.

LPardue:
Yeah that makes sense on a reread of section 7.2.5.2

PR #2979, opened 2019-08-20
Forbid duplicate setting identifiers.

bencebeky:
I assume the intention was to forbid the same setting identifier
occurring multiple times in a SETTINGS frame, regardless of the values.
However, since a parameter is defined as an identifier-value pair, the
current text does not forbid duplicate identifiers if the values are
different.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1340,9 +1340,9 @@ Different values for the same parameter can be advertised by each peer. For
 example, a client might be willing to consume a very large response header,
 while servers are more cautious about request size.
 
-Parameters MUST NOT occur more than once in the SETTINGS frame.  A receiver MAY
-treat the presence of the same parameter more than once as a connection error of
-type HTTP_SETTINGS_ERROR.
+Setting identifiers MUST NOT occur more than once in the SETTINGS frame.  A
+receiver MAY treat the presence of the same setting identifier more than once as
+a connection error of type HTTP_SETTINGS_ERROR.
 
 The payload of a SETTINGS frame consists of zero or more parameters.  Each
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -1340,9 +1340,9 @@ Different values for the same parameter can be advertised by each peer. For
 example, a client might be willing to consume a very large response header,
 while servers are more cautious about request size.
 
-Parameters MUST NOT occur more than once in the SETTINGS frame.  A receiver MAY
-treat the presence of the same parameter more than once as a connection error of
-type HTTP_SETTINGS_ERROR.
+Setting identifiers MUST NOT occur more than once in the SETTINGS frame.  A
```
```suggestion
The same setting identifier MUST NOT occur more than once in the SETTINGS frame.  A
```

draft-ietf-quic-http.md
```diff
@@ -1340,9 +1340,9 @@ Different values for the same parameter can be advertised by each peer. For
...
+receiver MAY treat the presence of the same setting identifier more than once as
```
```suggestion
receiver MAY treat the presence of duplicate setting identifiers as
```

draft-ietf-quic-http.md
```diff
@@ -1340,9 +1340,9 @@ Different values for the same parameter can be advertised by each peer. For
 example, a client might be willing to consume a very large response header,
 while servers are more cautious about request size.
 
-Parameters MUST NOT occur more than once in the SETTINGS frame.  A receiver MAY
-treat the presence of the same parameter more than once as a connection error of
-type HTTP_SETTINGS_ERROR.
+The same setting identifier MUST NOT occur more than once in the SETTINGS frame.  A
```
MikeBishop: Need to rewrap to fix the build break, I think.

PR #2987, opened 2019-08-23
Replace remove() with setting to null

### LINKED ISSUE #2569: RTT sample should be taken only once for each largest_acked
The pseudo code in [Section A.6](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#on-ack-received) ends up taking an RTT sample on every ack, which can result in multiple RTT samples per largest_acked. The code should check to see if the largest_acked is newly acked.

The text does say that an RTT sample MUST NOT be taken on a packet that is not newly_acked, but pseudo code needs to match.

ianswett:
Fixes #2569

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1157,7 +1157,7 @@ Pseudocode for OnPacketAcked follows:
    OnPacketAcked(acked_packet, pn_space):
      if (acked_packet.in_flight):
        OnPacketAckedCC(acked_packet)
-     sent_packets[pn_space].[acked_packet.packet_number] = null
+     sent_packets[pn_space][acked_packet.packet_number] = null
 ~~~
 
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -1157,7 +1157,7 @@ Pseudocode for OnPacketAcked follows:
    OnPacketAcked(acked_packet, pn_space):
      if (acked_packet.in_flight):
        OnPacketAckedCC(acked_packet)
-     sent_packets[pn_space].remove(acked_packet.packet_number)
+     sent_packets[pn_space].[acked_packet.packet_number] = null
```
```suggestion
     sent_packets[pn_space][acked_packet.packet_number] = null
```

mikkelfj:
It's a matter of taste, but I don't really like setting values to null when null is not defined value in a set. Removing something from a set is mathematically well defined. That Javascript happens to be weird doesn't have to affect a specification.

martinthomson:
I personally would prefer to keep remove(x) and use something like contains(x) on the test, but I trust Ian's judgment.

mikkelfj:
I suspect the issues are about operational behaviour more than formalism, and a null test is just a proposed solution. If so, I think remove and contains is a much better approach.

ianswett:
I created a 1-line PR #3015 to use contains().  I slightly prefer that one to this one.

janaiyengar:
I like #3015 better.

PR #2998, opened 2019-08-28
Error codes can't cause errors

### LINKED ISSUE #2816: Handling RESET_STREAM and STOP_SENDING with invalid error codes
The HTTP/3 spec is a little light on detail about processing an error code that should have never been sent. At @MikeBishop's suggestion, this issue is scoped to discussion of HTTP/3 stream errors:

> If a stream sees a RESET/STOP_SENDING with an error code that a) should only have originated from your side, or b) can only be a connection error, what do you do?

Example a) a client receives a STOP_SENDING with HTTP_REQUEST_CANCELLED
Example b) a server receives a RESET_STREAM with HTTP_LIMIT_EXCEEDED

MikeBishop:
Fixes #2816 by saying that, since error codes are extensible, you can't treat the receipt of an error code you don't understand or don't think you should have gotten as an error itself.  (Though it can still be an error that the stream closed, regardless of the error code.)

(Which leads to the question, should we be greasing the error code space?)

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1396,9 +1396,8 @@ connection error.
 
 Because new error codes can be defined without negotiation (see {{extensions}}),
 receipt of an unknown error code or use of an error code in an unexpected
-context MUST NOT be treated as an error.  However, the fact that a stream closed
-might itself constitute an error regardless of the error code provided -- see
-{{request-response}}.
+context MUST NOT be treated as an error.  However, closing a stream can
+constitute an error regardless of the error code -- see {{request-response}}.
 
 This section describes HTTP/3-specific error codes which can be used to express
 the cause of a connection or stream error.
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1397,7 +1397,7 @@ connection error.
 Because new error codes can be defined without negotiation (see {{extensions}}),
 receipt of an unknown error code or use of an error code in an unexpected
 context MUST NOT be treated as an error.  However, closing a stream can
-constitute an error regardless of the error code -- see {{request-response}}.
+constitute an error regardless of the error code (see {{request-response}}).
 
 This section describes HTTP/3-specific error codes which can be used to express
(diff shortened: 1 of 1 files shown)


LPardue:
> Which leads to the question, should we be greasing the error code space?)

Yes! I would create a grease stream and it could be closed (stop sending and/or reset) by peers.

martinthomson:
That seems to have very narrow applicability.  Would you only be permitted to send a greasing error code in that case?

Or to put the question differently, what would happen if you decided to randomly turn a subset of the semantic-ful error codes you generate into greasing codes?

MikeBishop:
Per the language of this PR, I think that would be fine.  Possibly with the exception of the errors that are supposed to inform behavior, like _REJECTED and perhaps EARLY_RESPONSE.

draft-ietf-quic-http.md
```diff
@@ -1394,6 +1394,12 @@ the entire connection when an error is encountered.  These are referred to as
 {{QUIC-TRANSPORT}}.  An endpoint MAY choose to treat a stream error as a
 connection error.
 
+Because new error codes can be defined without negotiation (see {{extensions}}),
+receipt of an unknown error code or use of an error code in an unexpected
+context MUST NOT be treated as an error.  However, the fact that a stream closed
```
ianswett: This last sentence is a bit odd.  How about something like "However, closing a stream can constitute an error regardless of the error code -- see..."

draft-ietf-quic-http.md
```diff
@@ -1394,6 +1394,11 @@ the entire connection when an error is encountered.  These are referred to as
 {{QUIC-TRANSPORT}}.  An endpoint MAY choose to treat a stream error as a
 connection error.
 
+Because new error codes can be defined without negotiation (see {{extensions}}),
+receipt of an unknown error code or use of an error code in an unexpected
+context MUST NOT be treated as an error.  However, closing a stream can
+constitute an error regardless of the error code -- see {{request-response}}.
```
```suggestion
constitute an error regardless of the error code; see {{request-response}}.
```

PR #3003, opened 2019-09-03
rename HTTP_UNEXPECTED_FRAME to HTTP_FRAME_UNEXPECTED and fold HTTP_W…

### LINKED ISSUE #2809: Rationalise HTTP_WRONG_STREAM and HTTP_UNEXPECTED_FRAME

LPardue:
…RONG_STREAM into it

Renaming the error to align with a suggestion from @MikeBishop made somewhere. This is a less drastic version of #2996. It too fixes #2809.

Closes #2996.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1415,7 +1415,7 @@ HTTP_CLOSED_CRITICAL_STREAM (0x104):
 : A stream required by the connection was closed or reset.
 
 HTTP_FRAME_UNEXPECTED (0x105):
-: A frame was received which was not permitted in the current state or stream.
+: A frame was received which was not permitted in the current state or on the current stream.
 
 HTTP_FRAME_ERROR (0x106):
 : A frame that fails to satisfy layout requirements or with an invalid size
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -1414,8 +1414,8 @@ HTTP_STREAM_CREATION_ERROR (0x103):
 HTTP_CLOSED_CRITICAL_STREAM (0x104):
 : A stream required by the connection was closed or reset.
 
-HTTP_UNEXPECTED_FRAME (0x105):
-: A frame was received which was not permitted in the current state.
+HTTP_FRAME_UNEXPECTED (0x105):
+: A frame was received which was not permitted in the current state or stream.
```
```suggestion
: A frame was received which was not permitted in the current state or on the current stream.
```
We've usually used "on" rather than "in" for a stream. (Though this reminds me of a great story about the choice of prepositions my former manager told once....)

MikeBishop:
Note:  I got some questions about the renumbering of the error codes; I went ahead and took it because we've had a massive renumbering since the last published draft anyway.  If we'd published the draft with the old numbering, I would have just left the hole.

PR #3007, opened 2019-09-06
Attempt to make the stateless reset text better

martinthomson:
PR #2927 had a mistake in it, and some very confusing language.  The rules for generating stateless resets was all mixed up in some complementary rules about generating packets that might trigger stateless resets.

This attempts to split out the rules for generating stateless resets (minimum size is 5 + 16 = 21), and for generating packets that might trigger a stateless reset (min_cid + 22).

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2467,11 +2467,12 @@ data (or 6 bytes, less the two fixed bits).
 
 A minimum size of 21 bytes does not guarantee that a stateless reset is
 difficult to distinguish from other packets if the recipient requires the use of
-a connection ID.  In order to ensure that a stateless reset that appears to be
-valid can be sent, all packets SHOULD be padded to at least 22 bytes longer than
-the minimum connection ID that the endpoint might use.  An endpoint that sends a
-stateless reset in response to packet that is 43 bytes or less in length SHOULD
-send a stateless reset that is one byte shorter than the packet it responds to.
+a connection ID.  To prevent a resulting stateless reset from being trivially
+distinguishable from a valid packet, all packets sent by an endpoint SHOULD be
+padded to at least 22 bytes longer than the minimum connection ID that the
+endpoint might use.  An endpoint that sends a stateless reset in response to
+packet that is 43 bytes or less in length SHOULD send a stateless reset that is
+one byte shorter than the packet it responds to.
 
 These values assume that the Stateless Reset Token is the same as the minimum
 expansion of the packet protection AEAD.  Additional unpredictable bytes are
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2465,23 +2465,26 @@ be a packet with a short header.  For the stateless reset to appear as a valid
 QUIC packet, the Unpredictable Bits field needs to include at least 38 bits of
 data (or 6 bytes, less the two fixed bits).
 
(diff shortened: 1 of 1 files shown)


mikkelfj:
Do we need discussion on the fixed QUIC header bit in the triggering packet and outgoing reset packet?

draft-ietf-quic-transport.md
```diff
@@ -2460,26 +2460,31 @@ of the packet header.  The remainder of the first byte and an arbitrary number
...
-valid QUIC packets.  With the set of AEAD functions defined in {{QUIC-TLS}},
-packets that are smaller than 21 bytes are never valid.
+To entities other than its intended recipient, a stateless reset will appear to
+be a packet with a short header.  For the stateless reset to appear as a valid
+QUIC packet, the Unpredictable Bits field needs to include at least 38 bits of
+data (or 6 bytes, less the two fixed bits).
+
+A minimum size of 21 bytes does not guarantee that a stateless reset is
+difficult to distinguish from other packets if the recipient requires the use of
+a connection ID.  In order to ensure that a stateless reset that appears to be
```
ianswett: This second sentence is hard to follow.  How about "To prevent a resulting stateless reset from being distinguished from a valid packet, all packets ..."

draft-ietf-quic-transport.md
```diff
@@ -2460,26 +2460,32 @@ of the packet header.  The remainder of the first byte and an arbitrary number
...
-AEAD expansion.
-
-An endpoint SHOULD NOT send a stateless reset that is significantly larger than
-the packet it receives.  Endpoints MUST discard packets that are too small to be
-valid QUIC packets.  With the set of AEAD functions defined in {{QUIC-TLS}},
-packets that are smaller than 21 bytes are never valid.
+To entities other than its intended recipient, a stateless reset will appear to
+be a packet with a short header.  For the stateless reset to appear as a valid
+QUIC packet, the Unpredictable Bits field needs to include at least 38 bits of
+data (or 6 bytes, less the two fixed bits).
```
martinduke: 5 bytes

draft-ietf-quic-transport.md
```diff
@@ -2460,26 +2460,32 @@ of the packet header.  The remainder of the first byte and an arbitrary number
...
+
+A minimum size of 21 bytes does not guarantee that a stateless reset is
+difficult to distinguish from other packets if the recipient requires the use of
+a connection ID.  To prevent a resulting stateless reset from being trivially
+distinguishable from a valid packet, all packets sent by an endpoint SHOULD be
+padded to at least 22 bytes longer than the minimum connection ID that the
+endpoint might use.  An endpoint that sends a stateless reset in response to
+packet that is 43 bytes or less in length SHOULD send a stateless reset that is
+one byte shorter than the packet it responds to.
```
martinduke: at least one byte?

draft-ietf-quic-transport.md
martinthomson: oops

draft-ietf-quic-transport.md
martinthomson: This is precisely one.  If the sender is following their side and sending the absolute minimum, then you can't go smaller without the resulting packet becoming obviously invalid.

martinthomson:
@martinduke if you can lay out what you think the security concerns are in another issue, that would helpful.  I realize that there are plenty of compromises in this design, which make this less than ideal.  But first I'd like to try to understand where you think this is falling short.

PR #3012, opened 2019-09-10
Add CRYPTO_BUFFER_EXCEEDED to the error code table

maskit:
`CRYPTO_BUFFER_EXCEEDED` is described in section 20 (Transport Error Codes) but it's not on the error code table in section 22.3 (QUIC Transport Error Codes Registry).


draft-ietf-quic-transport.md
```diff
@@ -5877,6 +5877,7 @@ The initial contents of this registry are shown in {{iana-error-table}}.
 | 0x7   | FRAME_ENCODING_ERROR      | Frame encoding error          | {{error-codes}} |
 | 0x8   | TRANSPORT_PARAMETER_ERROR | Error in transport parameters | {{error-codes}} |
 | 0xA   | PROTOCOL_VIOLATION        | Generic protocol violation    | {{error-codes}} |
+| 0xD   | CRYPTO_BUFFER_EXCEEDED    | CRYPTO data buffer overflowed | {{error-codes}} |
```
marten-seemann: Not an issue with this PR, but I'm wondering if we should change the number to 0x9, in order to fill the gap.

PR #3021, opened 2019-09-12
fix draft-23 test vectors

marten-seemann:


martinthomson:
Confirmed that this was a transcription error.  Thanks.

PR #3031, opened 2019-09-13
Reference "Nonces are Noticed" in the header protection analysis section

chris-wood:
This paper studies several "nonce hiding" transformations, of which the QUIC
header protection algorithm is one. This change replaces the old analysis
text with something a bit more crisp.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -74,8 +74,9 @@ informative:
     title: "Authenticated-Encryption with Associated-Data"
     author:
       - ins: P. Rogaway
-    date: 2002-09-20
-    target: "https://web.cs.ucdavis.edu/~rogaway/papers/ad.pdf"
+    date: 2002-11-22
+    target: "http://doi.acm.org/10.1145/586110.586125"
+    seriesinfo: Proceedings of the 9th ACM Conference on Computer and Communications Security, pages 98-107
 
   NAN:
     title: "Nonces are Noticed: AEAD Revisited"
@@ -84,7 +85,7 @@ informative:
       - ins: R. Ng
       - ins: B. Tackmann
     date: 2019-06-01
-    target: https://eprint.iacr.org/2019/624
+    target: "http://dx.doi.org/10.1007/978-3-030-26948-7_9"
     seriesinfo: Advances in Cryptology – CRYPTO 2019, pages 235-265
 
   QUIC-HTTP:
@@ -1441,11 +1442,12 @@ protected_field = field XOR PRF(hp_key, sample)
 ~~~
 
 As `hp_key` is distinct from the packet protection key, this construction
-(HN1) achieves AE2 security as defined in {{NAN}} and therefore guarantees privacy of `field`, the
-protected packet header. One important distinction between HN1 and the header
-protection construction in this document is that the latter uses an AEAD
-algorithm as the PRF. However, since the encrypted output of an AEAD is
-pseudorandom {{DefnAEAD}}, this achieves the properties desired from a PRF.
+(HN1) achieves AE2 security as defined in {{NAN}} and therefore guarantees
+privacy of `field`, the protected packet header. One important distinction
+between HN1 and the header protection construction in this document is that
+the latter uses an AEAD algorithm as the PRF. However, since the encrypted
+output of an AEAD is pseudorandom {{DefnAEAD}}, this achieves the properties
+desired from a PRF.
 
 Use of the same key and ciphertext sample more than once risks compromising
 header protection. Protecting two different headers with the same key and
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1440,7 +1440,7 @@ Header protection uses the output of the packet protection AEAD to derive
 protected_field = field XOR PRF(hp_key, sample)
 ~~~
 
-Assuming hp_key is distinct from the packet protection key, this construction
+As `hp_key` is distinct from the packet protection key, this construction
 (HN1) achieves AE2 security as defined in {{NAN}} and therefore guarantees privacy of `field`, the
 protected packet header. One important distinction between HN1 and the header
 protection construction in this document is that the latter uses an AEAD
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -70,14 +70,22 @@ informative:
     date: 2016-03-08
     target: "http://www.isg.rhul.ac.uk/~kp/TLS-AEbounds.pdf"
 
-  IMC:
-    title: "Introduction to Modern Cryptography, Second Edition"
+  DefnAEAD:
+    title: "Authenticated-Encryption with Associated-Data"
     author:
-      - ins: J. Katz
-      - ins: Y. Lindell
-    date: 2014-11-06
-    seriesinfo:
-      ISBN: 978-1466570269
+      - ins: P. Rogaway
+    date: 2002-09-20
+    target: "https://web.cs.ucdavis.edu/~rogaway/papers/ad.pdf"
+
+  NAN:
+    title: "Nonces are Noticed: AEAD Revisited"
+    author:
+      - ins: M. Bellare
+      - ins: R. Ng
+      - ins: B. Tackmann
+    date: 2019-06-01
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -1421,19 +1429,23 @@ amplification.
...
+`sample`, and then encrypts the header field using a pseudorandom function
+(PRF) as follows:
 
 ~~~
 protected_field = field XOR PRF(hp_key, sample)
 ~~~
 
-This construction is secure against chosen plaintext attacks (IND-CPA) {{IMC}}.
+Assuming hp_key is distinct from the packet protection key, this construction
+(HN1) achieves AE2 security and therefore guarantees privacy of `field`, the
```
mikkelfj: AE2 security is not an easily digestible term on its own. What is the objective, where is it defined?

chris-wood: It’s defined in the referenced paper. I thought it best to simply point readers in that direction if they want the details, rather try and pull it out here. (We didn’t previously define IND-CPA security, for example.)

draft-ietf-quic-tls.md
```diff
@@ -70,14 +70,22 @@ informative:
...
+
+  NAN:
+    title: "Nonces are Noticed: AEAD Revisited"
+    author:
+      - ins: M. Bellare
+      - ins: R. Ng
+      - ins: B. Tackmann
+    date: 2019-06-01
+    target: https://eprint.iacr.org/2019/624
+    seriesinfo: Advances in Cryptology – CRYPTO 2019, pages 235-265
```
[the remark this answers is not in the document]
martinthomson: I'll answer my own question: http://dx.doi.org/10.1007/978-3-030-26948-7_9

draft-ietf-quic-tls.md
mikkelfj: Even with citing, AE2 is not searchable. It is probably AE-scheme II, but I have only skimmed the paper.

martinthomson: > We  give  a  modified  syntax  for  nonce-based encryption, called NBE2, in which decryption does not get the nonce, a corresponding framework of security definitions called AE2 that guarantee nonce privacy in addition to authenticity and message privacy, and simple ways to turn NBE1 AE1-secure schemes into NBE2 AE2-secure schemes.

This is in the introduction of the paper, under "Contributions in brief."

mikkelfj: Uh sorry, I was locked on to Rogaway's paper.

draft-ietf-quic-tls.md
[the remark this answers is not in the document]
chris-wood: I chose the ePrint copy since it's more likely to stay up to date. I can swap in a reference to the CRYPTO version, though!

draft-ietf-quic-tls.md
```diff
@@ -70,14 +70,22 @@ informative:
...
+    title: "Authenticated-Encryption with Associated-Data"
     author:
-      - ins: J. Katz
-      - ins: Y. Lindell
-    date: 2014-11-06
-    seriesinfo:
-      ISBN: 978-1466570269
+      - ins: P. Rogaway
+    date: 2002-09-20
+    target: "https://web.cs.ucdavis.edu/~rogaway/papers/ad.pdf"
```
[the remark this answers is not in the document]
chris-wood: Will fix!

draft-ietf-quic-tls.md
```diff
@@ -1421,19 +1429,23 @@ amplification.
...
+Header protection uses the output of the packet protection AEAD to derive
+`sample`, and then encrypts the header field using a pseudorandom function
+(PRF) as follows:
 
 ~~~
 protected_field = field XOR PRF(hp_key, sample)
 ~~~
 
-This construction is secure against chosen plaintext attacks (IND-CPA) {{IMC}}.
+Assuming hp_key is distinct from the packet protection key, this construction
```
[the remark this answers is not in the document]
chris-wood: I used this text in the off chance that we decide to change the scheme again. Using what's currently specified is probably better though. Thanks for the suggestion!

draft-ietf-quic-tls.md
```diff
@@ -1421,19 +1429,23 @@ amplification.
...
+(HN1) achieves AE2 security and therefore guarantees privacy of `field`, the
+protected packet header. One important distinction between HN1 and the header
+protection construction in this document is that the latter uses an AEAD
+algorithm as the PRF. However, since the encrypted output of an AEAD is
```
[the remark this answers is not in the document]
chris-wood: I think the distinction is important because AEADs are (depending on one's definition) not PRFs. 

> We should probably say something about what it takes to define a new header protection scheme somewhere though.

Hmm, I'm not sure. That's probably best done elsewhere, e.g., a CFRG document, that we then cite.

draft-ietf-quic-tls.md
```diff
@@ -1421,19 +1429,23 @@ amplification.
...
+pseudorandom {{DefnAEAD}}, this achieves the properties desired from a PRF.
```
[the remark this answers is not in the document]
chris-wood: Rogaway's definition assumes the encrypted output is pseudorandom (per the Adv^PRIV requirement), and that's the property header protection relies on, so I felt it best to cite.

draft-ietf-quic-tls.md
```diff
@@ -84,7 +85,7 @@ informative:
       - ins: R. Ng
       - ins: B. Tackmann
     date: 2019-06-01
-    target: https://eprint.iacr.org/2019/624
+    target: "http://dx.doi.org/10.1007/978-3-030-26948-7_9"
```
martinthomson: With a DOI, you can just cite this as: `{{?NAN=DOI.10.1007/978-3-030-26948-7_9}}` and save all that typing.  Same for the above.

draft-ietf-quic-tls.md
```diff
@@ -1440,12 +1441,13 @@ Header protection uses the output of the packet protection AEAD to derive
...
-protection construction in this document is that the latter uses an AEAD
-algorithm as the PRF. However, since the encrypted output of an AEAD is
-pseudorandom {{DefnAEAD}}, this achieves the properties desired from a PRF.
+As `hp_key` is distinct from the packet protection key, this construction
+(HN1) achieves AE2 security as defined in {{NAN}} and therefore guarantees
+privacy of `field`, the protected packet header. One important distinction
+between HN1 and the header protection construction in this document is that
+the latter uses an AEAD algorithm as the PRF. However, since the encrypted
+output of an AEAD is pseudorandom {{DefnAEAD}}, this achieves the properties
+desired from a PRF.
```
martinthomson: This last sentence doesn't wash well for me still.

The reason that we care about the AEAD output being pseudorandom is that this is the way we select the key we use for header protection.  But that is something that the Nonces are Noticed paper should worry about in its proofs.  The important factor for us here is that we are using the HN1 construction properly.  For that, we only need to show that we are using a PRF.  To support that we just need to say that both AES-ECB and ChaCha20 are PRFs.  If that claim needs support, that requires a different paper.

draft-ietf-quic-tls.md
chris-wood: Awesome! Thanks for that tip.

draft-ietf-quic-tls.md
chris-wood: Ah, right! I was operating under the assumption that we used AEADs for the mask, but we use the primitives. Since those are PRPs and therefore PRFs, this becomes much simpler. Stay tuned for an update!

draft-ietf-quic-tls.md
mikkelfj: That link is behind a springer paywall - the former wasn't.

chris-wood: True, yet the DOI is stable. I'm sure folks won't have trouble finding the ePrint version.

martinthomson:
I'm really happy to see good research on this that confirms our intuitions about one of the most important QUIC-specific constructions.  Thanks for doing this Chris.

Are you planning to make a similar contribution to DTLS 1.3?

chris-wood:
> Thanks for doing this Chris.

My pleasure. :-)

> Are you planning to make a similar contribution to DTLS 1.3?

Yep!

PR #3033, opened 2019-09-14
Suggest sending empty CRYPTO frames in Initial/Handshake packets

ghedo:
As per @tatsuhiro-t, since PING frames are not allowed in
Initial/Handshake packets, one can use an empty CRYPTO frame when
sending a probe when there is no other ack-eliciting frame to send.

It seems to me like this is something that is fairly easy to miss, so
worth expanding the existing suggestion I think.

The alternative would be to allow PING frames in Initial/Handshake, but
I figured we are maybe too late in the process for that kind of design
change.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -533,8 +533,10 @@ then the original transmission is acknowledged.  In the absence of any new
 application data, a PTO timer expiration now would find the sender with no new
 or previously-sent data to send.
 
-When there is no data to send, the sender SHOULD send a PING or other
-ack-eliciting frame in a single packet, re-arming the PTO timer.
+When there is no data to send, the sender SHOULD send an ack-eliciting frame in
+a single packet, re-arming the PTO timer.  PING frames can be used for 0/1-RTT
+packets, while empty CRYPTO frames can be used for Initial/Handshake packets
+where PING frames are not allowed.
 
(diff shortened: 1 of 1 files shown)


marten-seemann:
I'd prefer to just allow PING in all packet types. Sending empty CRYPTO frames seems kind of hacky to me.

tatsuhiro-t:
If don't know the reasoning why PING has been prohibited in Initial or Handshake, but if there is no hard rule to exclude it from handshake packets, I think that just allowing PING in Initial/Handshake is simpler.
I just chose empty CRYPTO so that it works with draft-23 compliant implementations; literally there is no ACK eliciting frame other than that.

marten-seemann:
I brought this up a long time ago in #2307. Apparently, this issue was closed by #2355. I'm wondering why.

ghedo:
Ok, thanks.

I made https://github.com/quicwg/base-drafts/issues/3034 to discuss allowing PING in Initial/Handshake packets again and made a PR (https://github.com/quicwg/base-drafts/pull/3035). The WG can then decide if we want the editorial-only change in this PR, or the design one in the other PR.

ianswett:
I agree with @marten-seemann the intent was to allow this in #2355, thanks for sending out a different PR to allow them.

janaiyengar:
I don't recall us wanting to disallow PING frames in Initial/Handshake packets. #3035 looks good to me.

ghedo:
Closing this since the other PR was merged.

PR #3044, opened 2019-09-17
Clarify why post-handshake client auth is banned

### LINKED ISSUE #2367: Text banning post-handshake client auth is unclear

DavidSchinazi:
Add short explanation and reference to explain why QUIC-TLS bans the use of post-handshake client authentication.

Fixes #2367.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -615,9 +615,9 @@ and deployment.
 A server MUST NOT use post-handshake client authentication (as defined in
 Section 4.6.2 of {{!TLS13}}), because the multiplexing offered by QUIC prevents
 clients from correlating the certificate request with the application-level
-event which triggered it (see {{?HTTP2-TLS13=I-D.ietf-httpbis-http2-tls13}}).
+event that triggered it (see {{?HTTP2-TLS13=I-D.ietf-httpbis-http2-tls13}}).
 More specifically, servers MUST NOT send post-handshake TLS CertificateRequest
-messages, and clients MUST treat receipt of such messages as a connection error
+messages and clients MUST treat receipt of such messages as a connection error
 of type PROTOCOL_VIOLATION.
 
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-tls.md
```diff
@@ -612,8 +612,13 @@ MAY refuse a connection if the client is unable to authenticate when requested.
 The requirements for client authentication vary based on application protocol
 and deployment.
 
-A server MUST NOT use post-handshake client authentication (see Section 4.6.2 of
-{{!TLS13}}).
+A server MUST NOT use post-handshake client authentication (as defined in
+Section 4.6.2 of {{!TLS13}}), because the multiplexing offered by QUIC prevents
+clients from correlating the certificate request with the application-level
+event which triggered it (see {{?HTTP2-TLS13=I-D.ietf-httpbis-http2-tls13}}).
```
martinthomson: CMOY (cite my own work) alert.  :)

Also a CMOS tweak (Chicago Manual of Style) tweak:
```suggestion
event that triggered it (see {{?HTTP2-TLS13=I-D.ietf-httpbis-http2-tls13}}).
```

draft-ietf-quic-tls.md
```diff
@@ -612,8 +612,13 @@ MAY refuse a connection if the client is unable to authenticate when requested.
...
+More specifically, servers MUST NOT send post-handshake TLS CertificateRequest
+messages, and clients MUST treat receipt of such messages as a connection error
```
```suggestion
messages and clients MUST treat receipt of such messages as a connection error
```

draft-ietf-quic-tls.md
DavidSchinazi: I'm honored to have been confused with @davidben.

davidben: Likewise. :-)

PR #3047, opened 2019-09-17
Send after receiving an ACK

### LINKED ISSUE #2604: Pseudocode fix: Data should be sent on loss-threshold timer
(@larseggert : Thanks for catching this)

Currently, in OnLossDetectionTimeout(), we detect losses but do not sent anything on time threshold loss detection.  This means that SetLossDetectionTimer() that is called later in the function does not set a timer because inflight can be 0 (if all inflight is marked lost).

This is an oversight, since clearly we send rtxs when we mark packets as lost, and in the case of threshold detection, we send as much as the cwnd allows.

My proposal is to add the last two lines below
if (loss_time != 0):
    // Time threshold loss Detection
    DetectLostPackets(pn_space)
    SendOnePacketWithRetransmittedData()
    SendDataAsCwndAllows()

ianswett:
Fixes #2604

Clarifies that when you do send in response to an ACK, it should not be more than the initial window.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1031,7 +1031,9 @@ OnAckReceived(ack, pn_space):
     OnPacketAcked(acked_packet.packet_number, pn_space)
 
   DetectLostPackets(pn_space)
-  SendPacketsAsCongestionControllerAllows()
+  // Immediately attempt to send, but never more than the
+  // initial window in a single burst.
+  SendUpToIWPacketsAsCongestionControllerAllows()
   pto_count = 0
   SetLossDetectionTimer()
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1007,8 +1007,8 @@ OnAckReceived(ack, pn_space):
     largest_acked_packet[pn_space] =
         max(largest_acked_packet[pn_space], ack.largest_acked)
 
-  // Nothing to do if there are no newly acked packets.
   newly_acked_packets = DetermineNewlyAckedPackets(ack, pn_space)
+  // Nothing to do if there are no newly acked packets.
   if (newly_acked_packets.empty()):
     return
 
@@ -1031,9 +1031,8 @@ OnAckReceived(ack, pn_space):
     OnPacketAcked(acked_packet.packet_number, pn_space)
 
   DetectLostPackets(pn_space)
-
+  SendPacketsAsCongestionControllerAllows()
   pto_count = 0
-
   SetLossDetectionTimer()
 
 


janaiyengar:
I would be inclined to having "SendOnePacket" before "SendAsCwndAllows" as we discussed on the issue. WDYT?

ianswett:
In a world governed by congestion controllers with pacing, I'm not sure that makes sense?  It seems to pull over some TCP ideas, but I'm not sure they're relevant here?  Certainly a congestion controller could decide to always send a packet after loss, but can we leave that up to the congestion controller?

janaiyengar:
So, yes to pacing, but there are impls that won't do pacing. Even with pacing, there's now an additional delay in recovering from this packet loss -- loss recovery time is increased by pacing_delay (which will be higher now that the loss has occurred). Allowing the sender to send out 1 packet immediately minimizes the chances of loss repair delay at the receiver.

ianswett:
There is a section for the congestion controller below.  I think this is a response it can choose to have, rather than incorporating it into the loss recovery/detection pseudocode?

ianswett:
Thinking more about this, I really like having congestion control, loss detection, and what to send as separate parts of QUIC.  I'm hesitant to couple them more tightly.

draft-ietf-quic-recovery.md
```diff
@@ -1031,9 +1031,8 @@ OnAckReceived(ack, pn_space):
     OnPacketAcked(acked_packet.packet_number, pn_space)
 
   DetectLostPackets(pn_space)
```
janaiyengar: I propose we change this to be as follows, keeping with the discussion on the issue.
```
if (DetectLostPackets(pn_space)):
    SendOnePacket()
SendPacketsAsCongestionControllerAllows()
```

ianswett: Q: Why would you always send one packet?  Depending on how ACKs were grouped, this could result in an unnecessary burst of packets that could defeat the congestion controller's intent.  Based on my experience, I don't think this is a purely theoretical problem.

draft-ietf-quic-recovery.md
```diff
@@ -1031,9 +1031,8 @@ OnAckReceived(ack, pn_space):
...
-
+  SendPacketsAsCongestionControllerAllows()
   pto_count = 0
-
   SetLossDetectionTimer()
```
martinduke: we should delete this. OnPacketSent() already sets the timer, and SendPacketsAsCongestionControllerAllows may not actually send a packet.

ianswett: The ack may update the RTT, so I think we should  keep this unless we add a return value to SendPacketsAsCongestionControllerAllows() indicating whether a packet was sent?

martinduke: Ah, yes, we might have to reset the timer to reflect the new RTT. But then shouldn't the RTT update code go above this?

ianswett: It's above on line 1024.

martinduke: My mistake. I saw the UpdateRTT definition below and got confused.

draft-ietf-quic-recovery.md
janaiyengar: Detection of a lost packet at a minimum signals that one packet left the network. This is the logic by which TCP sends a packet immediately on loss detection. It allows a sender to recover from a single packet loss immediately. Do you think we should deviate from what TCP does here?

draft-ietf-quic-recovery.md
```diff
@@ -1031,9 +1031,10 @@ OnAckReceived(ack, pn_space):
     OnPacketAcked(acked_packet.packet_number, pn_space)
 
   DetectLostPackets(pn_space)
-
+  // Immediately attempt to send, but never more than the
+  // initial window in a single burst.
+  SendUpToIWPacketsAsCongestionControllerAllows()
```
janaiyengar: What's the rationale for sending a burst here? And if the congestion controller allows more, why limit to IW?

ianswett: The intent is is not to send a burst, but rather to limit any sending to the IW, as specified in the normative text: 

"Implementations MUST either use pacing or limit such bursts to the initial congestion window, which is recommended to be the minimum of 10 * max_datagram_size and max(2* max_datagram_size, 14720)), where max_datagram_size is the current maximum size of a datagram for the connection, not including UDP or IP overhead."

draft-ietf-quic-recovery.md
ianswett: I want to keep when to send(congestion control) and loss detection separate.  Stating that you always send one packet in this pseudocode commingles the two.

Also, we have no normative text that indicates sending a packet after receiving an ACK is the right thing to do here, so putting it in the pseudocode would make the pseudocode out of sync with the text.  If you think we should allow sending a packet because TCP does, then the right thing to do is file an issue and then write a PR that updates the text and the pseudocode.

draft-ietf-quic-recovery.md
```diff
@@ -1007,8 +1007,8 @@ OnAckReceived(ack, pn_space):
     largest_acked_packet[pn_space] =
         max(largest_acked_packet[pn_space], ack.largest_acked)
 
-  // Nothing to do if there are no newly acked packets.
   newly_acked_packets = DetermineNewlyAckedPackets(ack, pn_space)
+  // Nothing to do if there are no newly acked packets.
```
martinthomson: This change looks like a regression, or it is as at least unrelated: the intent before was to comment about the call to `DetermineNewlyAckedPackets()` and the use of the returned value.  I think that's still the right way to frame this.

PR #3048, opened 2019-09-17
Retransmitting ACK frames can inflate RTT

### LINKED ISSUE #2748: Retransmitting ACKs may cause misleading RTT samples
Even though a sender is encouraged to assemble frames containing up-to-date information every time it sends a packet, it is not forbidden to retransmit copies of frames from lost packets. A receiver MUST accept packets containing an outdated frame, such as a MAX_DATA frame carrying a smaller maximum data than one found in an older packet.
https://github.com/quicwg/base-drafts/pull/2312/files

If a packet containing an ACK is lost and the payload is retransmitted with a new packet number AND no ACKs are received prior to the retransmitted ACK frame arriving, the peer will calculate and RTT sample that is 2 RTTs + loss recovery delay.

As such, I'd suggest we add a warning that retransmitting ACKs can cause RTT inflation and is heavily discouraged.

There may be other issues with retransmitting ACKs, but this is the first that comes to mind.

ianswett:
Fixes #2748

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -296,7 +296,7 @@ retain sufficient history is an open research question.
 Section 13.3 of {{QUIC-TRANSPORT}} recommends sending the most recent set of
 acknowledgements in ACK frames, but if ACK frames are retransmitted, an inflated
 RTT can be measured.  Some incorrect RTT samples are prevented by only measuring
-RTT when the largest acked increases.
+RTT when the largest acknowledged increases.
 
 ## Estimating min_rtt {#min-rtt}
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -291,6 +291,13 @@ are received within an RTT.  As suggested in {{?RFC6298}}, doing so might result
 in inadequate history in smoothed_rtt and rttvar.  Ensuring that RTT estimates
 retain sufficient history is an open research question.
 
+### Retransmitted ACK frames
+
+Section 13.3 of {{QUIC-TRANSPORT}} recommends sending the most recent set of
+acknowledgements in ACK frames, but if ACK frames are retransmitted, an inflated
+RTT can be measured.  Some incorrect RTT samples are prevented by only measuring
+RTT when the largest acked increases.
+
 ## Estimating min_rtt {#min-rtt}
 
 min_rtt is the minimum RTT observed over the lifetime of the connection.


draft-ietf-quic-recovery.md
```diff
@@ -291,6 +291,13 @@ are received within an RTT.  As suggested in {{?RFC6298}}, doing so might result
 in inadequate history in smoothed_rtt and rttvar.  Ensuring that RTT estimates
 retain sufficient history is an open research question.
 
+### Retransmitted ACK frames
+
+Section 13.3 of {{QUIC-TRANSPORT}} recommends sending the most recent set of
+acknowledgements in ACK frames, but if ACK frames are retransmitted, an inflated
+RTT can be measured.  Some incorrect RTT samples are prevented by only measuring
+RTT when the largest acked increases.
```
martinthomson: Type the entire word:

```suggestion
RTT when the largest acknowledged increases.
```

draft-ietf-quic-recovery.md
```diff
@@ -291,6 +291,13 @@ are received within an RTT.  As suggested in {{?RFC6298}}, doing so might result
 in inadequate history in smoothed_rtt and rttvar.  Ensuring that RTT estimates
 retain sufficient history is an open research question.
 
+### Retransmitted ACK frames
+
+Section 13.3 of {{QUIC-TRANSPORT}} recommends sending the most recent set of
+acknowledgements in ACK frames, but if ACK frames are retransmitted, an inflated
```
```suggestion
acknowledgements in ACK frames, but if ACK frames are retransmitted without adjusting the Ack Delay, an inflated
```

janaiyengar:
I agree with @martinthomson. Perhaps it's better to add some text where we discourage retransmitting of frames (at the end of [Section 13.3](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#rfc.section.13.3)) instead of in recovery.

draft-ietf-quic-recovery.md
ianswett: I originally wrote that, but it's likely to be inflated anyway, particularly if max_ack_delay is < RTT.

ianswett:
I created #3049 as an alternative.  Or we can close this issue and move on.

mikkelfj:
I agree that ACKs should be sent up to date, but I also think it makes sense the the receiver to protect itself in this regard. Some implementations will mindlessly repeat frames and it is not enough to cause a protocol violation.

PR #3049, opened 2019-09-18
Retransmitting an ACK frame warning

### LINKED ISSUE #2748: Retransmitting ACKs may cause misleading RTT samples

ianswett:
It can result in an inflated RTT, as well as being a waste of bytes.

Fixes #2748

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3181,10 +3181,10 @@ containing that information is acknowledged.
   unless the endpoint has sent a RESET_STREAM for that stream.  Once an endpoint
   sends a RESET_STREAM frame, no further STREAM frames are needed.
 
-* The most recent set of acknowledgments are sent in ACK frames.  An ACK frame
-  SHOULD contain all unacknowledged acknowledgments, as described in
-  {{sending-ack-frames}}.  Retransmitting an ACK frame can result in an
-  inflated RTT measurement.
+* ACK frames are sent containing the most recent set of acknowledgements and
+  Ack Delay, as described in {{sending-ack-frames}}. Retransmitting an old ACK
+  frame with stale Ack Delay can cause the peer to generate an inflated RTT
+  sample.
 
 * Cancellation of stream transmission, as carried in a RESET_STREAM frame, is
   sent until acknowledged or until all stream data is acknowledged by the peer
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3181,10 +3181,11 @@ containing that information is acknowledged.
   unless the endpoint has sent a RESET_STREAM for that stream.  Once an endpoint
   sends a RESET_STREAM frame, no further STREAM frames are needed.
 
-* ACK frames are sent containing the most recent set of acknowledgements and
-  Ack Delay, as described in {{sending-ack-frames}}. Delaying the transmission
-  of packets containing ACK frames or sending old ACK frames can cause the
-  peer to generate an inflated RTT sample or unnecessarily disable ECN.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3183,7 +3183,8 @@ containing that information is acknowledged.
 
 * The most recent set of acknowledgments are sent in ACK frames.  An ACK frame
   SHOULD contain all unacknowledged acknowledgments, as described in
-  {{sending-ack-frames}}.
+  {{sending-ack-frames}}.  Retransmitting an ACK frame can result in an
```
janaiyengar: This para is not great. That second sentence totally doesn't belong here. Suggested rewrite:

"ACK frames carry the most recent set of acknowledgements and the current Ack Delay estimate. Retransmitting an old ACK frame with stale Ack Delay can cause the peer to generate an inflated RTT sample."

Then in 13.2.1 (sending-ack-frames), add this as the first paragraph:
"ACK frames SHOULD acknowledge both ack-eliciting and non-ack-eliciting packets that have been received but not yet acknowledged."

ianswett: There is already text that says "When only non-ACK-eliciting packets need to be acknowledged, an endpoint MAY wait until an ACK-eliciting packet has been received to bundle an ACK frame with outgoing frames."  I'm not sure that's sufficient, but your suggestion adds a SHOULD, which makes this change no longer editorial.  The ACK generation text needs a full rewrite, and I'd rather tackle that suggestion as part of that PR.

draft-ietf-quic-transport.md
```diff
@@ -3181,9 +3181,10 @@ containing that information is acknowledged.
   unless the endpoint has sent a RESET_STREAM for that stream.  Once an endpoint
   sends a RESET_STREAM frame, no further STREAM frames are needed.
 
-* The most recent set of acknowledgments are sent in ACK frames.  An ACK frame
-  SHOULD contain all unacknowledged acknowledgments, as described in
-  {{sending-ack-frames}}.
+* ACK frames are sent containing the most recent set of acknowledgements and
+  Ack Delay, as described in {{sending-ack-frames}}. Retransmitting an old ACK
+  frame with stale Ack Delay can cause the peer to generate an inflated RTT
+  sample.
```
martinthomson: Here's my alternative:

> ACK frames always contain an acknowledgement for the highest packet number that has been processed and an Ack Delay that is updated to the time that the packet is sent.  Note that delaying the transmission of packets containing ACK frames or sending copies of old ACK frames will cause problems for a peer, including inflated RTT estimates or unnecessary disabling of ECN.

ianswett: I took part of this, but the first statement is incorrect: ACK frames do not have to contain the highest packet number that has been processed(there was an issue about this).

draft-ietf-quic-transport.md
ianswett: I worked part of your first suggestion in.

The latter suggestion is now present(from a different PR).

PR #3055, opened 2019-09-20
Remove 1 RTT delayed ack requirement

### LINKED ISSUE #3030: Remove Requirement of an Acknowledgement per Round Trip
The spec gives no reason (that I can see) as to why an ACK frame must be sent at least once per round trip (assuming there are ack-eliciting packets to acknowledge), effectively overwriting the ACK delay logic when the RTT is very small. If the RTT is smaller than the platform's timer granularity, this isn't really even possible.

I propose to remove this requirement. If the WG instead feels that we should keep this requirement, then there should be a reasoning for it; and probably an update to the PTO calculation as well.

ianswett:
Also removes obsolete text about acknowledging CRYPTO frames quickly, now that Initial and Handshake packet spaces have a max_ack_delay of 0.

Fixes #3030

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3010,9 +3010,9 @@ An ACK frame SHOULD be generated for at least every second ack-eliciting packet.
 This recommendation is in keeping with standard practice for TCP {{?RFC5681}}.
 
 A receiver's delayed acknowledgment timer SHOULD NOT exceed the value it
-indicates in the `max_ack_delay` transport parameter. The sender uses the receiver's
-`max_ack_delay` value in determining timeouts for timer-based retransmission, as
-detailed in section 5.2.1 of {{QUIC-RECOVERY}}.
+indicates in the `max_ack_delay` transport parameter. The sender uses the
+receiver's `max_ack_delay` value in determining timeouts for timer-based
+retransmission, as detailed in section 5.2.1 of {{QUIC-RECOVERY}}.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3009,11 +3009,16 @@ guidance offered below seeks to strike this balance.
...
+An endpoint MUST NOT excessively delay acknowledgements of ack-eliciting
+packets.  An endpoint commits to a maximum delay using the max_ack_delay
+transport parameter; see {{transport-parameter-definitions}}.  max_ack_delay
+declares an explicit contract: an endpoint promises to never delay
+acknowledgments of an ack-eliciting packet by more than the indicated value. If
+it does, any excess accrues to the RTT estimate and could result in delayed
+retransmissions from the peer.  For Initial and Handshake packets, a
+max_ack_delay of 0 is used.  The sender uses the receiver's `max_ack_delay`
+value in determining timeouts for timer-based retransmission, as detailed
+in section 5.2.1 of {{QUIC-RECOVERY}}.
```
```suggestion
in Section 5.2.1 of {{QUIC-RECOVERY}}.
```

mirjak:
Sorry I'm late here but I think it would still be good to mention that congestion control usually is supposed to work on a RTT granularity and if you send less ACKs than one per RTT that can have an impact on cc. I don't think this has to be a normative requirement given you have normative requirements on max_ack_delay and a reasonable default value but it would still be good to mention this dependency as cc is done on sender and ACKing on  receiver and therefore it might not always that obvious.

PR #3057, opened 2019-09-23
PTO MUST send new data or retransmit data if possible

### LINKED ISSUE #3056: Lost client Initial is never retransmitted
As of https://github.com/quicwg/base-drafts/pull/2806, if the client's Initial is lost, there is no required behavior that leads to it being retransmitted. Because the client has received no ACKs, no `loss_time` is set, and no loss is detected, leading to the client only sending probe packets, which are not required to bear any particular data ("A probe packet **MAY** carry retransmitted unacknowledged data") so long as they are ACK-eliciting. The server, having no connection state, cannot do anything with these packets. 

Quinn probes are currently a single PING frame (plus padding, when appropriate) for simplicity, so implementing the new logic immediately broke our tests. If recovery logic now requires probe packets to bear specific data to function correctly in certain circumstances, this should be specified explicitly.

ianswett:
Fixes #3056

This is implied in some spots, but never stated with normative language.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -524,9 +524,10 @@ as a probe, unless there is no data available to send.  An endpoint MAY send up
 to two full-sized datagrams containing ack-eliciting packets, to avoid an
 expensive consecutive PTO expiration due to a single lost datagram.
 
-When the PTO timer expires, and there is new or previously sent data, it MUST
-be sent.  Data at Initial encryption MUST be sent before Handshake data and
-data at Handshake encryption MUST be sent before any ApplicationData data.
+When the PTO timer expires, and there is new or previously sent unacknowledged
+data, it MUST be sent.  Data at Initial encryption MUST be sent before
+Handshake data and data at Handshake encryption MUST be sent before any
+ApplicationData data.
 
 It is possible the sender has no new or previously-sent data to send.
 As an example, consider the following sequence of events: new application data
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -526,12 +526,11 @@ as a probe, unless there is no data available to send.  An endpoint MAY send up
 to two full-sized datagrams containing ack-eliciting packets, to avoid an
 expensive consecutive PTO expiration due to a single lost datagram.
 
-It is possible that the sender has no new or previously-sent data to send.  As
-an example, consider the following sequence of events: new application data is
-sent in a STREAM frame, deemed lost, then retransmitted in a new packet, and
-then the original transmission is acknowledged.  In the absence of any new
-application data, a PTO timer expiration now would find the sender with no new
-or previously-sent data to send.
+When the PTO timer expires, and there is new or previously sent data, it MUST
+be sent.  It is possible the sender has no new or previously-sent data to send.
+As an example, consider the following sequence of events: new application data
+is sent in a STREAM frame, deemed lost, then retransmitted in a new packet,
+and then the original transmission is acknowledged.
 
 When there is no data to send, the sender SHOULD send a PING or other
 ack-eliciting frame in a single packet, re-arming the PTO timer.
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -525,9 +525,9 @@ to two full-sized datagrams containing ack-eliciting packets, to avoid an
 expensive consecutive PTO expiration due to a single lost datagram.
 
 When the PTO timer expires, and there is new or previously sent unacknowledged
-data, it MUST be sent.  Data that was previously sent with Initial encryption MUST be sent before
-Handshake data and, similarly, data previously sent at Handshake encryption MUST be sent before any
-ApplicationData data.
+data, it MUST be sent.  Data that was previously sent with Initial encryption
+MUST be sent before Handshake data and data previously sent at Handshake
+encryption MUST be sent before any ApplicationData data.
 
 It is possible the sender has no new or previously-sent data to send.
 As an example, consider the following sequence of events: new application data
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -495,9 +495,9 @@ a PATH_RESPONSE to seed initial_rtt for a new path, but the delay SHOULD NOT
 be considered an RTT sample.
 
 Until the server has validated the client's address on the path, the amount of
-data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
-If no data can be sent, then the PTO alarm MUST NOT be armed until data has
-been received from the client.
+data it can send is limited to three times the amount of data received,
+as specified in Section 8.1 of {{QUIC-TRANSPORT}}. If no data can be sent,
+then the PTO alarm MUST NOT be armed.
 
 Since the server could be blocked until more packets are received from the
 client, it is the client's responsibility to send packets to unblock the server
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -496,10 +496,8 @@ be considered an RTT sample.
 
 Until the server has validated the client's address on the path, the amount of
 data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
-Data at Initial encryption MUST be retransmitted before Handshake data and
-data at Handshake encryption MUST be retransmitted before any ApplicationData
-data.  If no data can be sent, then the PTO alarm MUST NOT be armed until
-data has been received from the client.
+If no data can be sent, then the PTO alarm MUST NOT be armed until data has
+been received from the client.
 
 Since the server could be blocked until more packets are received from the
 client, it is the client's responsibility to send packets to unblock the server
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -526,12 +526,11 @@ as a probe, unless there is no data available to send.  An endpoint MAY send up
...
-an example, consider the following sequence of events: new application data is
-sent in a STREAM frame, deemed lost, then retransmitted in a new packet, and
-then the original transmission is acknowledged.  In the absence of any new
-application data, a PTO timer expiration now would find the sender with no new
-or previously-sent data to send.
+When the PTO timer expires, and there is new or previously sent data, it MUST
+be sent.  It is possible the sender has no new or previously-sent data to send.
+As an example, consider the following sequence of events: new application data
+is sent in a STREAM frame, deemed lost, then retransmitted in a new packet,
+and then the original transmission is acknowledged.
```
Ralith: This example only applies to implementations that can usefully process acknowledgements for packets they've already deemed lost and retransmitted the contents of, which I don't think is required. Is there a more common case?

ianswett: This is a fairly common case in our implementation, so I think it's a good example.

ianswett:
Under "Handshakes and New Paths", it says "Data at Initial encryption MUST be retransmitted before Handshake data and data at Handshake encryption MUST be retransmitted before any ApplicationData".  I moved that down to this section, since I think it fits slightly better here.

ianswett:
In terms of packet size, this is similar to what the old Handshake text recommended and what PTO should have said all along, as it aligns well with TCP's TLP and RTO behavior.

draft-ietf-quic-recovery.md
Ralith: Maybe note that this is optional behavior? I know if I was reading this for the first time, I'd try to work out where this was required and get confused.

Ralith:
The previous text read like it only described retransmission of lost data, and this now reads like it only describes retransmission in PTOs. Are both intended?

Is there a restriction anywhere that retransmitted data must not be sent at a higher encryption level than it was originally sent, or similar?

ianswett:
All of this text is in the PTO section, so only applies to PTO.

I think transport is usually the best place to talk about where data is retransmitted, but it might be worth an editorial note that data is never retransmitted at a different packet number space than the original one.

draft-ietf-quic-recovery.md
ianswett: It probably should be recommended behavior, and we should probably have some text on it.  Filed #3058 to discuss if and what to do to improve that.

draft-ietf-quic-recovery.md
```diff
@@ -526,15 +524,16 @@ as a probe, unless there is no data available to send.  An endpoint MAY send up
...
+data at Handshake encryption MUST be sent before any ApplicationData data.
 
-When there is no data to send, the sender SHOULD send a PING or other
-ack-eliciting frame in a single packet, re-arming the PTO timer.
+It is possible the sender has no new or previously-sent data to send.
+As an example, consider the following sequence of events: new application data
+is sent in a STREAM frame, deemed lost, then retransmitted in a new packet,
+and then the original transmission is acknowledged.  When there is no data to
+send, the sender SHOULD send a PING or other ack-eliciting frame in a single
+packet, re-arming the PTO timer.
```
martinthomson: I don't get this example.  In this example, you have sent something, it was received, end of story.

The new text above says that you can't arm PTO if you have nothing to send (MUST NOT even).  Which did you mean?

Ralith: In this case, the second packet is ack-eliciting and has not been acknowledged, eventually leading to a probe timeout. However, there is no unsent or sent-but-unacknowledged ack-eliciting data available to send, so a `PING` is fabricated instead.

Perhaps "there is new or previously sent data" should be more specific?

draft-ietf-quic-recovery.md
```diff
@@ -526,15 +524,16 @@ as a probe, unless there is no data available to send.  An endpoint MAY send up
 to two full-sized datagrams containing ack-eliciting packets, to avoid an
 expensive consecutive PTO expiration due to a single lost datagram.
 
-It is possible that the sender has no new or previously-sent data to send.  As
-an example, consider the following sequence of events: new application data is
-sent in a STREAM frame, deemed lost, then retransmitted in a new packet, and
-then the original transmission is acknowledged.  In the absence of any new
-application data, a PTO timer expiration now would find the sender with no new
-or previously-sent data to send.
+When the PTO timer expires, and there is new or previously sent data, it MUST
```
Ralith: Does "previously sent" here mean in-flight/unacknowledged, or literally any historical data?

Does "data" mean ack-eliciting frames specifically?

ianswett: I changed it to previously sent unacknowledged data, because that is the intent, though it's also a bit verbose.

Data is used fairly widely to mean ack-eliciting frames in this document.  I'm not sure if that is something we should codify in the definitions section?

draft-ietf-quic-recovery.md
```diff
@@ -496,10 +496,8 @@ be considered an RTT sample.
 
 Until the server has validated the client's address on the path, the amount of
 data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
-Data at Initial encryption MUST be retransmitted before Handshake data and
-data at Handshake encryption MUST be retransmitted before any ApplicationData
-data.  If no data can be sent, then the PTO alarm MUST NOT be armed until
-data has been received from the client.
+If no data can be sent, then the PTO alarm MUST NOT be armed until data has
+been received from the client.
```
martinthomson: Do you mean "until the path has been validated"?  "data" is highly non-specific.

draft-ietf-quic-recovery.md
martinthomson: Still not working for me.

I get the need for the client to send extra Initial packets again in the case where it has nothing to say, but that should be a special provision.

If p1 is acknowledged and p2 contained nothing new, then let it lie.  You don't need to know whether it was lost or not.

I get that this packet will stay outstanding unless you have something new to say, but you are only sending a PING to deal with an accounting discrepancy.  That only becomes relevant when you go to send the next packet.  Is there any way that this sort of blip can end up consuming the entire congestion window?

draft-ietf-quic-recovery.md
Ralith: On review, maybe the existing requirement that probes be ACK-eliciting is enough. Nothing wrong with retransmitting some unacknowledged ACKs alongside your PING.

draft-ietf-quic-recovery.md
ianswett: It's path validation or receiving enough data.  I updated the paragraph, PTAL.

draft-ietf-quic-recovery.md
janaiyengar: So the problem this was intended to address was exactly that accounting discrepancy. There are two ways to handle outstanding packets if we want to give up on them: call them lost and reduce the window (see text below, maybe that should be part of this paragraph), or send a PING to elicit an ACK from the receiver. The text used to recommend the calling them lost IIRC, but I don't see a clear winner here.

draft-ietf-quic-recovery.md
```diff
@@ -495,11 +495,9 @@ a PATH_RESPONSE to seed initial_rtt for a new path, but the delay SHOULD NOT
 be considered an RTT sample.
 
 Until the server has validated the client's address on the path, the amount of
-data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
-Data at Initial encryption MUST be retransmitted before Handshake data and
-data at Handshake encryption MUST be retransmitted before any ApplicationData
-data.  If no data can be sent, then the PTO alarm MUST NOT be armed until
-data has been received from the client.
+data it can send is limited to three times the amount of data received,
+as specified in Section 8.1 of {{QUIC-TRANSPORT}}. If no data can be sent,
```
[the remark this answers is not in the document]
ianswett: Are you suggesting we always arm the PTO and if we've hit the amplification factor, the server should send a PING-only packet?  That actually sounds easier than not arming it and prompts the client to send a padded Initial ACK, but I'm not sure if that's what you're suggesting?

draft-ietf-quic-recovery.md
martinduke: What is the purpose of this statement?

> Data that was previously sent with Initial encryption
> MUST be sent before Handshake data and data previously sent at Handshake
> MUST be sent before any ApplicationData data.

Doesn't this conflict with the "PTO per packet number space" PR? If I'm implementing that PR most literally, I should PTO only the data from the space that expired. For example, I can have the 0RTT/1RTT PTO timer running first with some handshake data also in flight, and I should send just 0RTT/1RTT.

If I'm coalescing, I'm not sure why this would matter.

As for "MUST send new/unacked data", is that necessary? Why is not OK to send a PING instead to reduce bytes in flight? Perhaps I'm expecting new data soon and want to preserve the window space? I'd prefer if this were a SHOULD.

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
ianswett: I created https://github.com/quicwg/base-drafts/issues/3161 to discuss whether we want to make a design change here, FWIW.

draft-ietf-quic-recovery.md
ianswett: Yes, but this is existing text and I didn't want to tie this PR to the PTO per PN space PR.

One of the reasons I'd like to get this change in soon is it's going to cause conflicts with the multiple PN spaces PR, and the other PR is more substantive.

The original issue(#3056) was that there was no MUST to retransmit the client Initial, and if the server doesn't create connection state until receiving a client Initial, the handshake fails.

That being said, the client sending a full sized ack eliciting packet that's not the Initial seems odd.  And if the server isn't going to create state, it seems like it should send a Retry to do path validation.

janaiyengar:
Thanks for opening other issues, @ianswett. I'm going to merge this, since I think it's a net betterment over what we have. We can continue the other discussions on the other issues/PRs.

PR #3059, opened 2019-09-23
http: syntax fixes for quic Alt-Svc parameter

Lekensteyn:
Use the correct ABNF grammar to express repetition and fix the Reserved
Frame Type link.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -317,11 +317,11 @@ provide version-negotiation hints to HTTP/3 clients. QUIC versions are four-byte
 sequences with no additional constraints on format. Leading zeros SHOULD be
 omitted for brevity.
 
-Syntax:
+Syntax of the "quic" parameter value:
 
 ~~~ abnf
-quic = DQUOTE version-number [ "," version-number ] * DQUOTE
-version-number = 1*8HEXDIG; hex-encoded QUIC version
+quic         = DQUOTE quic-version *( "," quic-version ) DQUOTE
+quic-version = 1*8HEXDIG      ; hex-encoded QUIC version
 ~~~
 
 Where multiple versions are listed, the order of the values reflects the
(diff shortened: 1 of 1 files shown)


LPardue:
There is some baggage with this parameter - https://github.com/quicwg/base-drafts/issues/1093

People were a bit grumpy with changing their parser 18 months ago, I think they might be even grumpier now that we have even more client implementations in the wild.

What is the benefit of this change? I think we need an issue for due process and discussion.

RyanAtGoogle:
Oh joy. This caused me to go look at our parser in Chrome, and realized we're still doing the old (pre 1097) quic=<hex>;quic=<hex> format. So Chrome has work to do no matter how we resolve this. 

That being said, I'd really rather not have two different ways to do this, so +1 to @LPardue 's comment on the motivation.

Lekensteyn:
I can remove the additional unquoted token change and keep this PR purely as editorial fix. Before the PR, the current text does not have correct ABNF grammar.

I added the unquoted token variant since it is permitted (but not required) by the grammar at https://tools.ietf.org/html/rfc7838#section-3
>    parameter     = token "=" ( token / quoted-string )

Since HTTP/3 over QUIC is not the only user of this Alt-Svc header, I can imagine that generic parsers are written such that they already understand the unquoted token value. Thus, adding support for this should not be an additional burden?

I have no strong opinion here. The additional CPU cost of handling both unquoted and quoted forms should be neglible. Disallowing the unquoted form is a restriction of the RFC 7838 grammar.

LPardue:
Let's create a new issue. 

You're correct that non-HTTP/3 implementations might need to handle this - hence the original issue which related back to structured headers (an attempt at defining some best practice in this regard -https://github.com/httpwg/http-extensions/issues/281).

Part of the problem is not CPU cost of DQUOTEs but expressing "a list of versions" and making sure that implementations know how to do that. Alt-Svc defines the parameter syntax but not how to handle repetition of parameters.

Lekensteyn:
I removed the single quote change, this should be a purely editorial fix now. I'll create a new issue for the unquoted parameter value.

PR #3064, opened 2019-09-24
Add details on how spurious losses may be detected

### LINKED ISSUE #3058: Clearer description of detecting spurious retransmissions
There are multiple mentions of spurious retransmission in the recovery draft, but no text on what detecting this may require from an implementation.  Maybe there should be?

"Spuriously declaring packets as lost leads to unnecessary retransmissions and may result in degraded performance due to the actions of the congestion controller upon detecting loss. Implementations that detect spurious retransmissions and increase the reordering threshold in packets or time MAY choose to start with smaller initial reordering thresholds to minimize recovery latency."

From comments on #3057

ianswett:
Fixes #3058

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -384,12 +384,18 @@ A packet is declared lost if it meets all the following conditions:
 The acknowledgement indicates that a packet sent later was delivered, while the
 packet and time thresholds provide some tolerance for packet reordering.
 
-Spuriously declaring packets as lost leads to unnecessary retransmissions and
-may result in degraded performance due to the actions of the congestion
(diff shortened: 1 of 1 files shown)


martinthomson:
This doesn't seem particularly actionable.  I mean, I could probably work something out, but if the goal is to provide a spec where it is relatively easy to follow the advice and produce a good implementation, then this doesn't meet that test.

Ralith:
I read this as "On acknowledgement of a packet, discard state for packets that were declared lost prior to the acknowledged packet's sending," which is at least enough to give me some ideas.

janaiyengar:
I agree with @martinthomson. I don't want to ask for a detailed mechanism for doing this, but without it, the text here is not be particularly useful. Let's close this out.

PR #3066, opened 2019-09-24
Change PTO to be per packet number space

### LINKED ISSUE #2863: unrecoverable loss pattern leads to deadlock

### LINKED ISSUE #3067: 0RTT or 0.5RTT data extends the PTO before handshake completion
In the recovery draft, when an ack-eliciting packet is sent in any packet number space, the PTO is extended.  

For example, if there is a long delay between a server sending Initial and Handshake, then that would delay retransmitting the Initial.

A larger problem may be the delay between Initial and 0.5RTT data on the server side or Initial and 0RTT data on the client side.

In both cases, prior to handshake completion, the application may write data at anytime, subject to congestion control limits and availability of ApplicationData keys.  In pathological cases(ie: an application sending tiny 'keep-alive' like pings) the Initial of the client or server may never be retransmitted until after the connection runs out of CWND, which is likely longer than the handshake timeout.

This problem was observed in gQUIC and we added an extra field to track the last sent packet containing crypto data and before the handshake was complete, armed the handshake timer from that.

The most straightforward fix to this I'm aware of is #3066, which makes PTO per PN space.

### LINKED ISSUE #3074: Probe packet section needs to suggest a crypto context
Sec 5.3.1 of quic-recovery discusses how to send loss probes. Now that this can be done at any crypto level, there ought to be some suggestions (are any MUSTs needed?) on what crypto level to send probes at.

This is probably like the CONNECTION_CLOSE suggestion; if you have multiple levels, it may be good to coalesce some packets.

I'm working through the logic and may have some suggestions soon.

ianswett:
Makes PTO consistent with loss detection, which is per packet number space.  This keeps a unified exponential backoff to avoid excess network load and doesn't arm the PTO for ApplicationData until the handshake is complete.

Fixes #3067 and fixes #3074 
Necessary for fixing #2863 without an explicit signal

Also fixes an issue where PTO was always using max_ack_delay, but there is no longer a max_ack_delay in the Initial and Handshake packet number spaces, so those timeouts were longer than necessary.

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1176,7 +1176,7 @@ Pseudocode for OnLossDetectionTimeout follows:
 ~~~
 OnLossDetectionTimeout():
   earliest_loss_time, pn_space =
-    GetEarliestTimeAndSpace(loss_times)
+    GetEarliestTimeAndSpace(loss_time)
   if (earliest_loss_time != 0):
     // Time threshold loss Detection
     DetectLostPackets(pn_space)
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -524,10 +524,10 @@ removed from bytes in flight when the Initial and Handshake keys are discarded.
 ### Sending Probe Packets
 
 When a PTO timer expires, a sender MUST send at least one ack-eliciting packet
-in the packet number space as a probe, unless there is no data available to send.
-An endpoint MAY send up to two full-sized datagrams containing ack-eliciting
-packets, to avoid an expensive consecutive PTO expiration due to a single lost
-datagram.
+in the packet number space as a probe, unless there is no data available to
+send.  An endpoint MAY send up to two full-sized datagrams containing
+ack-eliciting packets, to avoid an expensive consecutive PTO expiration due
+to a single lost datagram.
 
 In addition to sending data in the packet number space for which the timer
 expired, the sender SHOULD coalesce ack-eliciting packets from all other packet
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -438,11 +438,13 @@ and larger thresholds increase loss detection delay.
 A Probe Timeout (PTO) triggers sending one or two probe datagrams when
 ack-eliciting packets are not acknowledged within the expected period of
 time or the handshake has not been completed.  A PTO enables a connection to
-recover from loss of tail packets or acknowledgements. The PTO algorithm used
-in QUIC implements the reliability functions of Tail Loss Probe
-{{?TLP=I-D.dukkipati-tcpm-tcp-loss-probe}} {{?RACK}}, RTO {{?RFC5681}} and
-F-RTO algorithms for TCP {{?RFC5682}}, and the timeout computation is based on
-TCP's retransmission timeout period {{?RFC6298}}.
+recover from loss of tail packets or acknowledgements.  As with loss
+detection, the probe timeout is per packet number space.
+
+The PTO algorithm used in QUIC implements the reliability functions of
+Tail Loss Probe {{?TLP=I-D.dukkipati-tcpm-tcp-loss-probe}} {{?RACK}},
+RTO {{?RFC5681}} and F-RTO algorithms for TCP {{?RFC5682}}, and the timeout
+computation is based on TCP's retransmission timeout period {{?RFC6298}}.
 
 ### Computing PTO
 
@@ -471,8 +473,10 @@ the PTOs might be caused by loss of packets or acknowledgements due to severe
 congestion.  The life of a connection that is experiencing consecutive PTOs is
 limited by the endpoint's idle timeout.
 
-A sender computes its PTO timer every time an ack-eliciting packet is sent. A
-sender might choose to optimize this by setting the timer fewer times if it
+A sender computes its PTO timer every time an ack-eliciting packet is sent.
+When ack-eliciting packets are in-flight in multiple packet number spaces,
+the timer MUST be set from the packet number space with the earliest timeout.
+A sender might choose to optimize this by setting the timer fewer times if it
 knows that more ack-eliciting packets will be sent within a short period of
 time.
 
@@ -496,9 +500,7 @@ be considered an RTT sample.
 
 Until the server has validated the client's address on the path, the amount of
 data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
-Data at Initial encryption MUST be retransmitted before Handshake data and
-data at Handshake encryption MUST be retransmitted before any ApplicationData
-data.  If no data can be sent, then the PTO alarm MUST NOT be armed until
+If no data can be sent, then the PTO alarm MUST NOT be armed until
 data has been received from the client.
 
 Since the server could be blocked until more packets are received from the
@@ -522,9 +524,14 @@ removed from bytes in flight when the Initial and Handshake keys are discarded.
 ### Sending Probe Packets
 
 When a PTO timer expires, a sender MUST send at least one ack-eliciting packet
-as a probe, unless there is no data available to send.  An endpoint MAY send up
-to two full-sized datagrams containing ack-eliciting packets, to avoid an
-expensive consecutive PTO expiration due to a single lost datagram.
+in the packet number space as a probe, unless there is no data available to send.
+An endpoint MAY send up to two full-sized datagrams containing ack-eliciting
+packets, to avoid an expensive consecutive PTO expiration due to a single lost
+datagram.
+
+In addition to sending data in the packet number space for which the timer
+expired, the sender SHOULD coalesce ack-eliciting packets from all other packet
+number spaces with inflight data if sending coalesced packets is supported.
 
 It is possible that the sender has no new or previously-sent data to send.  As
 an example, consider the following sequence of events: new application data is
@@ -933,7 +940,7 @@ loss_detection_timer:
 pto_count:
 : The number of times a PTO has been sent without receiving an ack.
 
-time_of_last_sent_ack_eliciting_packet:
+time_of_last_sent_ack_eliciting_packet[kPacketNumberSpace]:
 : The time the most recent ack-eliciting packet was sent.
 
 largest_acked_packet\[kPacketNumberSpace]:
@@ -961,9 +968,9 @@ follows:
    rttvar = 0
    min_rtt = 0
    max_ack_delay = 0
-   time_of_last_sent_ack_eliciting_packet = 0
    for pn_space in [ Initial, Handshake, ApplicationData ]:
      largest_acked_packet[pn_space] = infinite
+     time_of_last_sent_ack_eliciting_packet[pn_space] = 0
      loss_time[pn_space] = 0
 ~~~
 
@@ -986,7 +993,7 @@ Pseudocode for OnPacketSent follows:
    sent_packets[pn_space][packet_number].in_flight = in_flight
    if (in_flight):
      if (ack_eliciting):
-       time_of_last_sent_ack_eliciting_packet = now
+       time_of_last_sent_ack_eliciting_packet[pn_space] = now
      OnPacketSentCC(sent_bytes)
      sent_packets[pn_space][packet_number].size = sent_bytes
      SetLossDetectionTimer()
@@ -1096,12 +1103,20 @@ Pseudocode for SetLossDetectionTimer follows:
 // Returns the earliest loss_time and the packet number
 // space it's from.  Returns 0 if all times are 0.
 GetEarliestLossTime():
-  time = loss_time[Initial]
+  return GetEarliestTimeAndSpace(loss_time)
+
+// Returns the earliest time_of_last_sent_ack_eliciting_packet
+// and the packet number space it's from.
+GetEarliestAckElicitingTime():
+  return GetEarliestTimeAndSpace(loss_time)
+
+GetEarliestTimeAndSpace(times):
+  time = times[Initial]
   space = Initial
   for pn_space in [ Handshake, ApplicationData ]:
-    if (loss_time[pn_space] != 0 &&
-        (time == 0 || loss_time[pn_space] < time)):
-      time = loss_time[pn_space];
+    if (times[pn_space] != 0 &&
+        (time == 0 || times[pn_space] < time)):
+      time = times[pn_space];
       space = pn_space
   return time, space
 
@@ -1135,8 +1150,9 @@ SetLossDetectionTimer():
       max_ack_delay
   timeout = timeout * (2 ^ pto_count)
 
-  loss_detection_timer.update(
-    time_of_last_sent_ack_eliciting_packet + timeout)
+  // Set the 
+  sent_time, _ = GetEarliestAckElicitingTime()
+  loss_detection_timer.update(sent_time + timeout)
 ~~~
 
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -501,8 +501,8 @@ be considered an RTT sample.
 
 Until the server has validated the client's address on the path, the amount of
 data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
-If no data can be sent, then the PTO alarm MUST NOT be armed for the corresponding packet number space until
-data has been received from the client.
+If no data can be sent, then the PTO alarm MUST NOT be armed until datagrams
+have been received from the client.
 
 Since the server could be blocked until more packets are received from the
 client, it is the client's responsibility to send packets to unblock the server
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -473,7 +473,7 @@ A sender computes its PTO timer every time an ack-eliciting packet is sent.
 When ack-eliciting packets are in-flight in multiple packet number spaces,
 the timer MUST be set for the packet number space with the earliest timeout,
 except for ApplicationData, which MUST be ignored until the handshake
-completes (See Section 4.1.1 of {{QUIC-TLS}}).  Not arming the PTO for
+completes; see Section 4.1.1 of {{QUIC-TLS}}.  Not arming the PTO for
 ApplicationData prioritizes completing the handshake and prevents the server
 from sending a 1-RTT PTO packet before it has the keys to process a 1-RTT
 packet.  When probe packets from two different packet number spaces are sent
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -1160,8 +1160,8 @@ Pseudocode for OnLossDetectionTimeout follows:
 
 ~~~
 OnLossDetectionTimeout():
-  loss_time, pn_space = GetEarliestTimeAndSpace(loss_times)
-  if (loss_time != 0):
+  earliest_loss_time, pn_space = GetEarliestTimeAndSpace(loss_times)
+  if (earliest_loss_time != 0):
     // Time threshold loss Detection
     DetectLostPackets(pn_space)
     SetLossDetectionTimer()
@@ -1178,7 +1178,8 @@ OnLossDetectionTimeout():
   else:
     // PTO. Send new data if available, else retransmit old data.
     // If neither is available, send a single PING frame.
-    SendOneOrTwoAckElicitingPackets()
+    _, pn_space = GetEarliestTimeAndSpace(time_of_last_sent_ack_eliciting_packet)
+    SendOneOrTwoAckElicitingPackets(pn_space)
 
   pto_count++
   SetLossDetectionTimer()
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -443,7 +443,7 @@ recover from loss of tail packets or acknowledgements.
 
 As with loss detection, the probe timeout is per packet number space.
 The PTO algorithm used in QUIC implements the reliability functions of
-Tail Loss Probe {{?RACK}}, RTO {{?RFC5681}} and F-RTO algorithms for
+Tail Loss Probe {{?RACK}}, RTO {{?RFC5681}}, and F-RTO algorithms for
 TCP {{?RFC5682}}. The timeout computation is based on TCP's retransmission
 timeout period {{?RFC6298}}.
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -951,7 +951,7 @@ time_of_last_sent_ack_eliciting_packet\[kPacketNumberSpace]:
 largest_acked_packet\[kPacketNumberSpace]:
 : The largest packet number acknowledged in the packet number space so far.
 
-loss_time\[kPacketNumberSpace]:
+loss_times\[kPacketNumberSpace]:
 : The time at which the next packet in that packet number space will be
   considered lost based on exceeding the reordering window in time.
 
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -501,7 +501,7 @@ be considered an RTT sample.
 
 Until the server has validated the client's address on the path, the amount of
 data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
-If no data can be sent, then the PTO alarm MUST NOT be armed until
+If no data can be sent, then the PTO alarm MUST NOT be armed for the corresponding packet number space until
 data has been received from the client.
 
 Since the server could be blocked until more packets are received from the
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -465,7 +465,8 @@ acknowledgement of a sent packet.  This time period includes the estimated
 network roundtrip-time (smoothed_rtt), the variance in the estimate (4*rttvar),
 and max_ack_delay, to account for the maximum time by which a receiver might
 delay sending an acknowledgement.  When the PTO is armed for Initial or
-Handshake packet number spaces, the max_ack_delay is 0.
+Handshake packet number spaces, the max_ack_delay is 0, as specified in
+13.2.5 of {{QUIC-TRANSPORT}}.
 
 The PTO value MUST be set to at least kGranularity, to avoid the timer expiring
 immediately.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-recovery.md
```diff
@@ -1135,8 +1150,9 @@ SetLossDetectionTimer():
       max_ack_delay
   timeout = timeout * (2 ^ pto_count)
 
-  loss_detection_timer.update(
-    time_of_last_sent_ack_eliciting_packet + timeout)
+  // Set the
```
DavidSchinazi: missing end of this comment?

draft-ietf-quic-recovery.md
```diff
@@ -438,11 +438,13 @@ and larger thresholds increase loss detection delay.
...
-recover from loss of tail packets or acknowledgements. The PTO algorithm used
-in QUIC implements the reliability functions of Tail Loss Probe
-{{?TLP=I-D.dukkipati-tcpm-tcp-loss-probe}} {{?RACK}}, RTO {{?RFC5681}} and
-F-RTO algorithms for TCP {{?RFC5682}}, and the timeout computation is based on
-TCP's retransmission timeout period {{?RFC6298}}.
+recover from loss of tail packets or acknowledgements.  As with loss
+detection, the probe timeout is per packet number space.
+
+The PTO algorithm used in QUIC implements the reliability functions of
+Tail Loss Probe {{?TLP=I-D.dukkipati-tcpm-tcp-loss-probe}} {{?RACK}},
```
martinthomson: Is `{{?RACK}}` another citation for Tail Loss Probe?

draft-ietf-quic-recovery.md
```diff
@@ -438,11 +438,13 @@ and larger thresholds increase loss detection delay.
...
+RTO {{?RFC5681}} and F-RTO algorithms for TCP {{?RFC5682}}, and the timeout
```
martinthomson: The commas here are odd.  I think that you could start a new sentence here:

```suggestion
RTO {{?RFC5681}}, and F-RTO  {{?RFC5682}} algorithms for TCP.  The timeout
```

Yeah, this is old text, but it's really confusing still.

draft-ietf-quic-recovery.md
```diff
@@ -496,9 +500,7 @@ be considered an RTT sample.
 
 Until the server has validated the client's address on the path, the amount of
 data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
-Data at Initial encryption MUST be retransmitted before Handshake data and
-data at Handshake encryption MUST be retransmitted before any ApplicationData
-data.  If no data can be sent, then the PTO alarm MUST NOT be armed until
+If no data can be sent, then the PTO alarm MUST NOT be armed until
```
```suggestion
If no data can be sent, then the PTO alarm MUST NOT be armed for the corresponding packet number space until
```

draft-ietf-quic-recovery.md
```diff
@@ -522,9 +524,14 @@ removed from bytes in flight when the Initial and Handshake keys are discarded.
 ### Sending Probe Packets
 
 When a PTO timer expires, a sender MUST send at least one ack-eliciting packet
-as a probe, unless there is no data available to send.  An endpoint MAY send up
-to two full-sized datagrams containing ack-eliciting packets, to avoid an
-expensive consecutive PTO expiration due to a single lost datagram.
+in the packet number space as a probe, unless there is no data available to
+send.  An endpoint MAY send up to two full-sized datagrams containing
+ack-eliciting packets, to avoid an expensive consecutive PTO expiration due
```
martinthomson: I think that we can drop the "containing ack-eliciting packets" here, it's redundant with the first sentence.

draft-ietf-quic-recovery.md
```diff
@@ -522,9 +524,14 @@ removed from bytes in flight when the Initial and Handshake keys are discarded.
...
+to a single lost datagram.
+
+In addition to sending data in the packet number space for which the timer
+expired, the sender SHOULD coalesce ack-eliciting packets from all other packet
+number spaces with inflight data if sending coalesced packets is supported.
```
martinthomson: This is important, so I think that it needs more support.  Questions that arise are: What happens if I don't do this?  Under what conditions would it be fine not to send in both spaces?

draft-ietf-quic-recovery.md
```diff
@@ -933,7 +940,7 @@ loss_detection_timer:
 pto_count:
 : The number of times a PTO has been sent without receiving an ack.
 
-time_of_last_sent_ack_eliciting_packet:
+time_of_last_sent_ack_eliciting_packet[kPacketNumberSpace]:
```
```suggestion
time_of_last_sent_ack_eliciting_packet\[kPacketNumberSpace]:
```

draft-ietf-quic-recovery.md
```diff
@@ -1096,12 +1103,20 @@ Pseudocode for SetLossDetectionTimer follows:
 // Returns the earliest loss_time and the packet number
 // space it's from.  Returns 0 if all times are 0.
 GetEarliestLossTime():
-  time = loss_time[Initial]
+  return GetEarliestTimeAndSpace(loss_time)
+
+// Returns the earliest time_of_last_sent_ack_eliciting_packet
+// and the packet number space it's from.
+GetEarliestAckElicitingTime():
+  return GetEarliestTimeAndSpace(loss_time)
```
martinthomson: You have two identical functions here that don't seem to do anything other than indirect over the global.  I don't get this change.

draft-ietf-quic-recovery.md
```diff
@@ -522,9 +524,14 @@ removed from bytes in flight when the Initial and Handshake keys are discarded.
 ### Sending Probe Packets
 
 When a PTO timer expires, a sender MUST send at least one ack-eliciting packet
-as a probe, unless there is no data available to send.  An endpoint MAY send up
-to two full-sized datagrams containing ack-eliciting packets, to avoid an
-expensive consecutive PTO expiration due to a single lost datagram.
+in the packet number space as a probe, unless there is no data available to
+send.  An endpoint MAY send up to two full-sized datagrams containing
+ack-eliciting packets, to avoid an expensive consecutive PTO expiration due
+to a single lost datagram.
```
kazuho: Confirmation question: does this mean that we'd be permitting more datagrams to be sent in effectively one PTO, when multiple packet number spaces are involved?

Consider the case where the PTO is armed for both the Handshake packet number space and the 1-RTT packet number space. Now that we have a PTO timer for each of the packet number space, we'd see both of them firing, and the text reads that an endpoint is allowed to send two datagrams for each.

mikkelfj:
I'd like to see the different packet number spaces operate independently because the initial handshake can use a less optimised, more complicated code path. While separate PTOs works for that, the requirement about SHOULD coalesce across packet number spaces goes against it. It requires upstream coordination and one PTO timer firing might affect the timeout of another PTO timer because coalesced packets are being sent. Overall this sounds rather complicated.

draft-ietf-quic-recovery.md
```diff
@@ -496,9 +500,7 @@ be considered an RTT sample.
 
 Until the server has validated the client's address on the path, the amount of
 data it can send is limited, as specified in Section 8.1 of {{QUIC-TRANSPORT}}.
-Data at Initial encryption MUST be retransmitted before Handshake data and
-data at Handshake encryption MUST be retransmitted before any ApplicationData
-data.  If no data can be sent, then the PTO alarm MUST NOT be armed until
+If no data can be sent, then the PTO alarm MUST NOT be armed until
 data has been received from the client.
```
martinduke: s/data/valid packets

"data", to me, implies payloads of CRYPTO or STREAM freams.

draft-ietf-quic-recovery.md
ianswett: It's mostly redundant, but if there are two packets, this makes it clear they have to be ack-eliciting.  This didn't change as part of this PR, FWIW.

draft-ietf-quic-recovery.md
ianswett: Good point, I updated the PR to try to limit the number of packets sent when implementations did not coalesce.

draft-ietf-quic-recovery.md
ianswett: I removed the functions, because they were overly verbose.  And as you said, I used the wrong argument in one of them.

draft-ietf-quic-recovery.md
ianswett: Changed to datagram, PTAL

draft-ietf-quic-recovery.md
```diff
@@ -471,8 +474,10 @@ the PTOs might be caused by loss of packets or acknowledgements due to severe
 congestion.  The life of a connection that is experiencing consecutive PTOs is
 limited by the endpoint's idle timeout.
 
-A sender computes its PTO timer every time an ack-eliciting packet is sent. A
-sender might choose to optimize this by setting the timer fewer times if it
+A sender computes its PTO timer every time an ack-eliciting packet is sent.
+When ack-eliciting packets are in-flight in multiple packet number spaces,
+the timer MUST be set from the packet number space with the earliest timeout.
```
```suggestion
the timer MUST be set for the packet number space with the earliest timeout.
```

draft-ietf-quic-recovery.md
```diff
@@ -471,8 +474,10 @@ the PTOs might be caused by loss of packets or acknowledgements due to severe
...
+A sender might choose to optimize this by setting the timer fewer times if it
 knows that more ack-eliciting packets will be sent within a short period of
 time.
```
janaiyengar: Remove this sentence, it's unnecessary.

draft-ietf-quic-recovery.md
```diff
@@ -522,9 +525,17 @@ removed from bytes in flight when the Initial and Handshake keys are discarded.
...
-as a probe, unless there is no data available to send.  An endpoint MAY send up
-to two full-sized datagrams containing ack-eliciting packets, to avoid an
-expensive consecutive PTO expiration due to a single lost datagram.
+in the packet number space as a probe, unless there is no data available to
+send.  An endpoint MAY send up to two full-sized datagrams containing
+ack-eliciting packets, to avoid an expensive consecutive PTO expiration due
+to a single lost datagram.
+
+In addition to sending data in the packet number space for which the timer
+expired, the sender SHOULD coalesce ack-eliciting packets from all other packet
```
janaiyengar: Is this necessarily "all" spaces?

draft-ietf-quic-recovery.md
```diff
@@ -522,9 +525,17 @@ removed from bytes in flight when the Initial and Handshake keys are discarded.
...
+number spaces with in-flight data if sending coalesced packets is supported.
+If implementations do not send coalesced packets upon timeout when multiple
+packet number spaces have in-flight data, then they MUST only send a single
```
janaiyengar: Suggestion: "If an endpoint has data in multiple packet number spaces in flight and is unable to send coalesced packet when a PTO expires, it MUST send only a single datagram on PTO expiration."

martinthomson: @janaiyengar has it right, but repeats "PTO expiration", suggest:

> If an endpoint has data in multiple packet number spaces in flight and is unable to send coalesced packet when a PTO expires, it MUST send only a single datagram.

janaiyengar:
The description should be fixed to remove "Necessary for fixing #2863". That's not true.

draft-ietf-quic-recovery.md
ianswett: 'all' removed, since coalescing 2 is sufficient.

draft-ietf-quic-recovery.md

ianswett:
I think I've addressed all the feedback, PTAL

draft-ietf-quic-recovery.md
```diff
@@ -1119,7 +1125,7 @@ PeerNotAwaitingAddressValidation():
          has received 1-RTT ACK
 
 SetLossDetectionTimer():
-  loss_time, _ = GetEarliestLossTime()
+  loss_time, _ = GetEarliestTimeAndSpace(loss_time)
```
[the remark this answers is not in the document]
ianswett: Thanks, fixed

draft-ietf-quic-recovery.md
```diff
@@ -1119,7 +1125,7 @@ PeerNotAwaitingAddressValidation():
          has received 1-RTT ACK
 
 SetLossDetectionTimer():
-  loss_time, _ = GetEarliestLossTime()
+  loss_time, _ = GetEarliestTimeAndSpace(loss_times)
```
marten-seemann: `loss_time` is a global variable, `loss_times` is not defined.
This should be 
```
earliest_loss_time, _ = GetEarliestTimeAndSpace(loss_time)
```

ianswett: Good catch, fixed.

draft-ietf-quic-recovery.md
```diff
@@ -477,6 +477,9 @@ limited by the endpoint's idle timeout.
 A sender computes its PTO timer every time an ack-eliciting packet is sent.
 When ack-eliciting packets are in-flight in multiple packet number spaces,
 the timer MUST be set for the packet number space with the earliest timeout.
+Even when there are ack-eliciting packets in-flight in multiple packet number
+spaces, the exponential increase in probe timeout occurs across all spaces to
```
martinduke: This interacts weirdly with the rest of this spec if I'm not coalescing, as I'm restricted to one packet per PTO. In the common case we are PTOing the server 1st flight, we get this behavior:
where P = srtt + 4 rttvar + max_ack_delay
 
time 0: send first flight
time P: send Initial
time 2P: send Handshake
time 4P: send 1RTT
time 9P: send Initial
time 18P: send Handshake

Maybe the lesson here is to coalesce, but we should understand what we're writing here.

ianswett: I removed the 1 packet limit when not coalescing when pointing out this existing exponential backoff behavior, so it should be:
 time 0: send first flight
 time P: send Initial and Handshake
 time 2P: Send 1RTT and Initial
 time 4P: send Handshake and 1RTT

Hopefully before 2P the server receives an ACK of something anyway, since that's seconds into the connection.

But we may want to tweak this as a result of the discard keys conclusion.

ianswett: I've updated the PR to no longer arm PTO for ApplicationData until the handshake is complete.  On the server side, it's pretty obvious you shouldn't arm it, since the server doesn't have receive keys to process 1-RTT ACKs.  I think this achieves what you originally might have hoped and solves the silliest case(server sending a 1-RTT PTO) above.

PTAL.

draft-ietf-quic-recovery.md
```diff
@@ -460,21 +463,28 @@ The PTO period is the amount of time that a sender ought to wait for an
...
-congestion.  The life of a connection that is experiencing consecutive PTOs is
-limited by the endpoint's idle timeout.
-
-A sender computes its PTO timer every time an ack-eliciting packet is sent. A
-sender might choose to optimize this by setting the timer fewer times if it
-knows that more ack-eliciting packets will be sent within a short period of
-time.
+congestion.  Even when there are ack-eliciting packets in-flight in multiple
+packet number spaces, the exponential increase in probe timeout occurs across
+all spaces to prevent excess load on the network. The life of a connection that
```
martinduke: "For example, a timeout in the Initial packet number space doubles the length of the timeout in the Handshake number space."

martinduke: There is also some conceptual subtlety that I suspect people could mess up.

If I'm coalescing my probes, then there is *never* a timeout at the higher levels -- each time, the junior level times out and a new high-level packet is sent that resets that level.

Misinterpreting this paragraph might mean that coalescing three probes could result in three bitshifts of the PTO.

This doesn't have to be in this PR, but when I reread this draft when everything's landed, hopefully it's in here somewhere.

ianswett: Hopefully it's clearer now.

draft-ietf-quic-recovery.md
```diff
@@ -460,21 +463,34 @@ The PTO period is the amount of time that a sender ought to wait for an
...
+Handshake packet number spaces, the max_ack_delay is 0.
 
 The PTO value MUST be set to at least kGranularity, to avoid the timer expiring
 immediately.
 
+A sender computes its PTO timer every time an ack-eliciting packet is sent.
+When ack-eliciting packets are in-flight in multiple packet number spaces,
+the timer MUST be set for the packet number space with the earliest timeout,
+except for ApplicationData, which MUST be ignored until the handshake
+completes (See Section 4.1.1 of {{QUIC-TLS}}).  Not arming the PTO for
```
martinthomson: How does this work with 0-RTT?  That's Application Data, but won't you want to attempt to repair on PTO?

```suggestion
completes; see Section 4.1.1 of {{QUIC-TLS}}.  Not arming the PTO for
```

draft-ietf-quic-recovery.md
janaiyengar: Did you forget to push your fix?

draft-ietf-quic-recovery.md
```diff
@@ -937,13 +956,13 @@ loss_detection_timer:
...
 
-time_of_last_sent_ack_eliciting_packet:
+time_of_last_sent_ack_eliciting_packet\[kPacketNumberSpace]:
 : The time the most recent ack-eliciting packet was sent.
 
 largest_acked_packet\[kPacketNumberSpace]:
 : The largest packet number acknowledged in the packet number space so far.
 
-loss_time\[kPacketNumberSpace]:
+loss_times\[kPacketNumberSpace]:
```
janaiyengar: Revert this change. We use singular for other arrays, which reads better to me anyways.

draft-ietf-quic-recovery.md
```diff
@@ -460,21 +463,34 @@ The PTO period is the amount of time that a sender ought to wait for an
...
 When a PTO timer expires, the PTO period MUST be set to twice its current
 value. This exponential reduction in the sender's rate is important because
 the PTOs might be caused by loss of packets or acknowledgements due to severe
-congestion.  The life of a connection that is experiencing consecutive PTOs is
-limited by the endpoint's idle timeout.
+congestion.  Even when there are ack-eliciting packets in-flight in multiple
+packet number spaces, the exponential increase in probe timeout occurs across
+all spaces to prevent excess load on the network.  For example, a timeout in
+the Initial packet number space doubles the length of the timeout in the
+Handshake packet number space.
```
junhochoi: If I understand correctly,
When Initial PTO is 10ms and Handshake PTO is 15ms:
  when Initial PTO expires, Initial PTO becomes 10ms and Handshake PTO becomes 30ms

Does it means Handshake PTO need to be re-armed at the point of Initial PTO expiration?
It may mean that Handshake PTO never be fired if Initial PTO keeps firing. There is other
text L544 saying that Initial and Handshake ack-eliciting packet should be sent at
Initial PTO expires. Then probably Handshake PTO need to be rearmed because we sent
Handshake ack-eliciting packet.
So for me it looks like now Initial PTO dominates Handshake PTO until Initial phase is complete.
Is my understanding correct?

draft-ietf-quic-recovery.md
ianswett: I fixed it by defining loss_times, but I'll revert that if you don't like it.  I prefer plurals, but I'll keep them consistent and singular for now.

draft-ietf-quic-recovery.md
ianswett: The Initial and Handshake timeouts should always be the same length, because neither uses the peer's max_ack_delay.  But they could be started from different points in time.  

As long as you send both at once, then the Initial PTO dominates.

draft-ietf-quic-recovery.md
```diff
@@ -1153,8 +1174,9 @@ Pseudocode for OnLossDetectionTimeout follows:
 
 ~~~
 OnLossDetectionTimeout():
-  loss_time, pn_space = GetEarliestLossTime()
-  if (loss_time != 0):
+  earliest_loss_time, pn_space =
+    GetEarliestTimeAndSpace(loss_times)
```
marten-seemann: This should be `loss_time` instead of `loss_times`. `loss_time` is a global variable.

ianswett: Thanks, I missed that one.

draft-ietf-quic-recovery.md
```diff
@@ -461,21 +464,34 @@ The PTO period is the amount of time that a sender ought to wait for an
...
+A sender computes its PTO timer every time an ack-eliciting packet is sent.
+When ack-eliciting packets are in-flight in multiple packet number spaces,
+the timer MUST be set for the packet number space with the earliest timeout,
+except for ApplicationData, which MUST be ignored until the handshake
+completes; see Section 4.1.1 of {{QUIC-TLS}}.  Not arming the PTO for
+ApplicationData prioritizes completing the handshake and prevents the server
+from sending a 1-RTT PTO packet before it has the keys to process a 1-RTT
+packet.  When probe packets from two different packet number spaces are sent
+when the PTO fires and both spaces have in-flight packets, this simplifies to
+setting the PTO on the lowest active packet number space.
```
janaiyengar: Can you reword this sentence? It's got two "when"s and it is a mouthful

draft-ietf-quic-recovery.md
```diff
@@ -461,21 +464,34 @@ The PTO period is the amount of time that a sender ought to wait for an
...
 When a PTO timer expires, the PTO period MUST be set to twice its current
 value. This exponential reduction in the sender's rate is important because
 the PTOs might be caused by loss of packets or acknowledgements due to severe
-congestion.  The life of a connection that is experiencing consecutive PTOs is
-limited by the endpoint's idle timeout.
+congestion.  Even when there are ack-eliciting packets in-flight in multiple
+packet number spaces, the exponential increase in probe timeout occurs across
+all spaces to prevent excess load on the network.  For example, a timeout in
+the Initial packet number space doubles the length of the timeout in the
+Handshake packet number space.
```
janaiyengar: Are all timers restarted at this point?

draft-ietf-quic-recovery.md
```diff
@@ -461,21 +464,34 @@ The PTO period is the amount of time that a sender ought to wait for an
 acknowledgement of a sent packet.  This time period includes the estimated
 network roundtrip-time (smoothed_rtt), the variance in the estimate (4*rttvar),
 and max_ack_delay, to account for the maximum time by which a receiver might
-delay sending an acknowledgement.
+delay sending an acknowledgement.  When the PTO is armed for Initial or
+Handshake packet number spaces, the max_ack_delay is 0.
```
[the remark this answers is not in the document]
ianswett: We already fixed this issue, see Issue #2638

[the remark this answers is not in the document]
ianswett: I add a citation to QUIC-TRANSPORT for this.

draft-ietf-quic-recovery.md
```diff
@@ -461,21 +464,34 @@ The PTO period is the amount of time that a sender ought to wait for an
...
 The PTO value MUST be set to at least kGranularity, to avoid the timer expiring
 immediately.
 
+A sender computes its PTO timer every time an ack-eliciting packet is sent.
+When ack-eliciting packets are in-flight in multiple packet number spaces,
+the timer MUST be set for the packet number space with the earliest timeout,
+except for ApplicationData, which MUST be ignored until the handshake
+completes; see Section 4.1.1 of {{QUIC-TLS}}.  Not arming the PTO for
+ApplicationData prioritizes completing the handshake and prevents the server
+from sending a 1-RTT PTO packet before it has the keys to process a 1-RTT
```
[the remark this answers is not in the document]
ianswett: "In response to" sounds like it's from the peer, which I find more confusing.

draft-ietf-quic-recovery.md
ianswett: Agreed, I tried to reword it.

draft-ietf-quic-recovery.md
ianswett: There's only one timer still.

marten-seemann:
What is the motivation behind not sending a PTO for the application data packet number space until the handshake is completed? If I had to guess, we're trying to not put too much load onto the network (and thereby increase the chance that PTO packets containing Initial / Handshake data make it through) until we know that we're done with the handshake.
However, this only works for the server side. The client completes the handshake after sending its cert and the CFIN, but it doesn't know if those packets were actually received.

Maybe the signal we're actually looking for here is handshake confirmed, and not handshake completed?

ianswett:
@marten-seemann The key is there's no point in arming the PTO for ApplicationData if you don't have keys to process the ACK frame if it was received.  I tried to make that clear in the text, but suggestions are welcome.

marten-seemann:
@ianswett Thanks for the explanation. Re-reading the text, I think this makes sense now. I noted that this also prevents the client from PTO-ing on 0-RTT packets, which is nice.

draft-ietf-quic-recovery.md
```diff
@@ -461,21 +464,35 @@ The PTO period is the amount of time that a sender ought to wait for an
...
+packet.  This simplifies to setting the PTO on the lowest active packet number
+space if packets from two different packet number spaces are sent when the PTO
+fires.
```
marten-seemann: I'm not sure if I understand this sentence. Consider the server that has just sent ServerHello and it's certificate, i.e. has outstanding Initial and Handshake data.
Isn't the whole point of this PR to arm the PTO timer based on both packet number spaces?

This is what I'd imagine would happen if no ACKs from the client are received:
t = 0: send Initial, immediately afterwards send Handshake. Set PTO timer for the Initial.
t = 1: PTO timer expires for the Initial packet. Send an Initial probe packet. Now the Handshake packet sent at t = 0 is the earliest outstanding packet, so set a PTO timer (with exponential backoff) for this packet.
t = 2: PTO timer expires for the Handshake packet. Send a Handshake probe packet. Now the Initial probe packet sent at t=1 is the earliest outstanding packet. The PTO timer for this packet would expire at t=5.

Is this correct?

ianswett: The Handshake PTO would expire at t=3, because of the backoff, which may have been what you meant.

The second Initial PTO would expire at t=5, as you said.

One note is that the text does recommend you send a packet in other packet number spaces when possible
"In addition to sending data in the packet number space for which the timer
expired, the sender SHOULD send ack-eliciting packets from other packet
number spaces with in-flight data, coalescing packets if possible."

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
janaiyengar: "from sending a 1-RTT packet on a PTO before ..."

We don't have a "PTO packet"

draft-ietf-quic-recovery.md
janaiyengar: @marten-seemann is right, I missed this too. The text says that the PTO is set on the "lowest active PN space", which is incorrect. I would simply drop this sentence.

draft-ietf-quic-recovery.md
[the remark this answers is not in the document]
ianswett: Done.

draft-ietf-quic-recovery.md
ianswett: Removed

marten-seemann: @ianswett No, I didn't mean t=3. The Handshake packet was sent sent at t=0, and we're doing one exponential backoff, meaning that the timer duration should be 2.
Or am I missing something here?

ianswett: I was thinking a Handshake packet was sent at 1, so it's my mistake.  Your example is correct given you're only going to send a single QUIC packet on PTO.

PR #3077, opened 2019-10-03
Remove obsolete 'Discussion' section

### LINKED ISSUE #3075: Discussion in quic-recovery of delayed ack being 25ms does not seem to relate to anything
This paragraph didn't seem to relate to anything else in the document:

> A shorter delayed ack time of 25ms was chosen because longer delayed acks can delay loss recovery and for the small number of connections where less than packet per 25ms is delivered, acking every packet is beneficial to congestion control and loss recovery.

Is it left-over from a previous version?

ianswett:
The recommendation for a 25ms delayed ack time is in transport, not recovery, so remove it from recovery.

Should this paragraph be moved to transport?

Fixes #3075


martinthomson:
I think that it's fine to just cut this.

PR #3083, opened 2019-10-08
Additional error cases for CANCEL_PUSH

### LINKED ISSUE #2819: CANCEL_PUSH edge cases
I think there are some edge cases with the CANCEL_PUSH frame that we don't explain: 

* reception of a CANCEL_PUSH frame with ID greater than the current push limit
* reception of CANCEL_PUSH with ID that has already been used

MikeBishop:
Fixes #2819.  Open to discussion about whether all of these are ID_ERROR or not.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1067,16 +1067,22 @@ push prior to the push stream being received.  The CANCEL_PUSH frame identifies
 a server push by Push ID (see {{frame-push-promise}}), encoded as a
 variable-length integer.
 
-When a server receives this frame, it aborts sending the response for the
-identified server push.  If the server has not yet started to send the server
-push, it can use the receipt of a CANCEL_PUSH frame to avoid opening a push
-stream.  If the push stream is open, the server SHOULD abruptly terminate that
-stream.  If the push stream has already closed, the server MAY take no action.
+When a client sends this frame, it is indicating that it does not wish to
+receive the promised resource.  The server SHOULD abort sending the resource,
+but the mechanism to do so depends on the state of the corresponding push
+stream.  If the server has not yet created a push stream, it does not create
+one.  If the push stream is open, the server SHOULD abruptly terminate that
+stream.  If the push stream has already closed, the server MAY abruptly
+terminate the stream (cancelling any retransmissions) or MAY take no action.
 
-When a client receives this frame, the server is indicating that it will not be
-fulfilling a promise and has not created a push stream.  Once the push stream
-has been created, sending CANCEL_PUSH would have no effect on the state of the
-push stream.  The server SHOULD abruptly terminate the push stream instead.
+When a server sends this frame, it is indicating that it will not be fulfilling
+a promise and has not created a push stream.  The client should not expect the
+corresponding promise to be fulfilled.
+
+Sending CANCEL_PUSH has no direct effect on the state of existing push streams.
+A server SHOULD NOT send a CANCEL_PUSH when it has already created a
+corresponding push stream, and a client SHOULD NOT send a CANCEL_PUSH when it
+has already received a corresponding push stream.
 
 A CANCEL_PUSH frame is sent on the control stream.  Receiving a CANCEL_PUSH
 frame on a stream other than the control stream MUST be treated as a connection
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1071,7 +1071,8 @@ When a server receives this frame, it aborts sending the response for the
 identified server push.  If the server has not yet started to send the server
 push, it can use the receipt of a CANCEL_PUSH frame to avoid opening a push
 stream.  If the push stream has been opened by the server, the server SHOULD
-abruptly terminate that stream.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -1071,7 +1071,8 @@ When a server receives this frame, it aborts sending the response for the
 identified server push.  If the server has not yet started to send the server
 push, it can use the receipt of a CANCEL_PUSH frame to avoid opening a push
 stream.  If the push stream has been opened by the server, the server SHOULD
```
ianswett: "has been opened" is now a bit odd without any qualification, since it's possible it's been opened and subsequently closed.

How about "If the push stream is open, the server SHOULD abruptly terminate that stream."

draft-ietf-quic-http.md
```diff
@@ -1093,10 +1094,15 @@ error of type HTTP_FRAME_UNEXPECTED.
 
 The CANCEL_PUSH frame carries a Push ID encoded as a variable-length integer.
 The Push ID identifies the server push that is being cancelled (see
-{{frame-push-promise}}).
+{{frame-push-promise}}).  If a CANCEL_PUSH frame is received which references a
+Push ID greater than currently allowed on the connection, this MUST be treated
```
ianswett: Should this only be enforced by the client, given it's the one limiting the number of pushes?  The next paragraph discusses the server closing the connection when the client sends a CANCEL_PUSH mentioning a push ID that's never been used by the server.

draft-ietf-quic-http.md
```diff
@@ -1071,7 +1071,8 @@ When a server receives this frame, it aborts sending the response for the
 identified server push.  If the server has not yet started to send the server
```
ianswett: How about reorganizing this section with one paragraph like this that starts with "When a server receives this frame, ..." and have another with "When a client receives this frame, ..." since the roles are completely asymmetric I believe?

draft-ietf-quic-http.md
LPardue: I think there are two effective limits - the maximum allowed push ID and the maximum used push ID. So if I understand Ian's point, there is not a situation where a client can issue a CANCEL_PUSH that violates the maximum allowed push ID without first triggering the violation of the maximum used push ID.

draft-ietf-quic-http.md
MikeBishop: The logic there is that, even if the stream is closed (FIN sent), there might still be value in sending a RESET_STREAM to abort any retransmissions.

draft-ietf-quic-http.md
MikeBishop: The server can issue a CANCEL_PUSH for a push it wasn't allowed to send, which the client should treat as an error.  And since CANCEL_PUSH and MAX_PUSH_ID are on the control stream, if the client uses a Push ID greater than it has permitted, that also seems invalid.

However, in the latter case, I can believe there are implementations where that ordering isn't maintained within the client, so perhaps we should relax that one.

ianswett: SG, I think I'm ok with this then.  I still find this section a bit of a hard read, but I don't have any specific suggestions.

MikeBishop:
I've tried to improve the readability of this section a bit; I turned Ian's suggested parallel paragraphs around to talk about why client and server would send the frame (and the corresponding responses) instead of talking about behavior on receipt.

Please take a fresh look, @LPardue and @ianswett.

draft-ietf-quic-http.md
```diff
@@ -1067,16 +1067,22 @@ push prior to the push stream being received.  The CANCEL_PUSH frame identifies
...
-stream.  The server SHOULD abruptly terminate the push stream instead.
+When a client sends this frame, it is indicating that it does not wish to
+receive the promised resource.  The server SHOULD abort sending the resource,
+but the mechanism to do so depends on the state of the corresponding push
+stream.  If the server has not yet created a push stream, it does not create
+one.  If the push stream is open, the server SHOULD abruptly terminate that
+stream.  If the push stream has already closed, the server MAY abruptly
+terminate the stream (cancelling any retransmissions) or MAY take no action.
+
+When a server sends this frame, it is indicating that it will not be fulfilling
```
```suggestion
When a server sends CANCEL_PUSH, it is indicating that it will not be fulfilling
```

draft-ietf-quic-http.md
```diff
@@ -1067,16 +1067,22 @@ push prior to the push stream being received.  The CANCEL_PUSH frame identifies
...
-fulfilling a promise prior to creation of a push stream.  Once the push stream
-has been created, sending CANCEL_PUSH has no effect on the state of the push
-stream.  The server SHOULD abruptly terminate the push stream instead.
+When a client sends this frame, it is indicating that it does not wish to
+receive the promised resource.  The server SHOULD abort sending the resource,
+but the mechanism to do so depends on the state of the corresponding push
+stream.  If the server has not yet created a push stream, it does not create
+one.  If the push stream is open, the server SHOULD abruptly terminate that
+stream.  If the push stream has already closed, the server MAY abruptly
+terminate the stream (cancelling any retransmissions) or MAY take no action.
```
```suggestion
terminate the stream or MAY take no action.
```

PR #3087, opened 2019-10-11
Make transport parameters varint

### LINKED ISSUE #3020: Transport parameter registry is too constraining for innovation
As of [draft-22](https://tools.ietf.org/html/draft-ietf-quic-transport-22#section-22.1), transport parameter identifiers are 16bit integers, with a registry where
- `[0x0000 - 0xfeff]` uses the [RFC8128 Specification Required policy](https://tools.ietf.org/html/rfc8126#section-4.6)
- `[0xff00 - 0xffff]` is reserved for experimentation

While keeping Specification Required makes sense for a good chunk of the registry space, I think there will be many experiments that are conducted before a full specification is written (some already exist today and I expect there to be many many more once the protocol is standardized).

However, those experiments only have 256 values to play with, and that is bound to cause collisions and headaches down the road.

I think the simplest solution to this problem is to use variable length integers for transport parameter identifiers. The TLS presentation language doesn't support them but since we already use VLIs for the values of several transport parameters we're no longer bound by the TLS presentation language. I think there was also support for doing this when it was discussed on #1608, but that issue was closed when the *values* were changed to VLI with #1947, without changing the identifiers.

I'm happy to write the PR if this issue gathers support.

DavidSchinazi:
This PR moves transport parameters away from TLS presentation language to instead use QUIC notation. This allows us to encode small integer transport parameters as 3 bytes instead of 5, but the main goal is to make the identifiers a 62-bit space which will allow uncoordinated experiments without risking collisions.

This fixes #3020.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3996,7 +3996,7 @@ Number Length bits.  It is used to carry "early" data from the client to the
 server as part of the first flight, prior to handshake completion. As part of
 the TLS handshake, the server can accept or reject this early data.
 
-See Section 2.3 of {{!TLS13}} for a discussion of 0-RTT data and its
+See Section 2.3 of {{!TLS13=RFC8446}} for a discussion of 0-RTT data and its
 limitations.
 
 ~~~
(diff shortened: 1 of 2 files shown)


ekr:
This doesn't seem to have consensus, based on the issue discussion

DavidSchinazi:
Agreed, the issue has not reached consensus yet. I wrote this PR to better describe what I was trying to change in the issue.

draft-ietf-quic-transport.md
```diff
@@ -4361,42 +4361,43 @@ connection.
 
 # Transport Parameter Encoding {#transport-parameter-encoding}
 
-The format of the transport parameters is the TransportParameters struct from
-{{figure-transport-parameters}}.  This is described using the presentation
-language from Section 3 of {{!TLS13=RFC8446}}.
+The `extension_data` field of the quic_transport_parameters extension defined in
+{{QUIC-TLS}} contains the QUIC transport parameters. They are encoded as a
+sequence of transport parameters, as shown in {{transport-parameter-sequence}}:
```
marten-seemann: We could consider making this an ordered list. This would simplify the check that the peer didn't send any duplicate transport parameters.

DavidSchinazi: I'm in favor of that idea (especially because it reduces fingerprinting surface), but I'd like it to be a separate issue/PR for now.

draft-ietf-quic-transport.md
```diff
@@ -4532,29 +4533,39 @@ disable_active_migration (0x000c):
...
-     opaque statelessResetToken[16];
-   } PreferredAddress;
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                       IPv4 Address (32)                       |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|         IPv4 Port (16)        |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                     IPv6 Address (128)                      ...
```
MikeBishop: This field isn't variable-length, so it shouldn't have the `...`.

draft-ietf-quic-transport.md
```diff
@@ -4532,29 +4533,39 @@ disable_active_migration (0x000c):
...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|         IPv6 Port (16)        |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| CID Length (8)|
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                      Connection ID (*)                      ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                 Stateless Reset Token (128)                 ...
```
MikeBishop: Likewise.

draft-ietf-quic-transport.md
martinthomson: Rather than this taking two lines, would you have this take 5 lines, or 8?

draft-ietf-quic-transport.md
```diff
@@ -5871,23 +5883,23 @@ merely aesthetically displeasing, or architecturally dubious).
...
+|  0x1  | idle_timeout                | {{transport-parameter-definitions}} |
+|  0x2  | stateless_reset_token       | {{transport-parameter-definitions}} |
+|  0x3  | max_packet_size             | {{transport-parameter-definitions}} |
+|  0x4  | initial_max_data            | {{transport-parameter-definitions}} |
+|  0x5  | initial_max_stream_data_bidi_local | {{transport-parameter-definitions}} |
+|  0x6  | initial_max_stream_data_bidi_remote | {{transport-parameter-definitions}} |
+|  0x7  | initial_max_stream_data_uni | {{transport-parameter-definitions}} |
+|  0x8  | initial_max_streams_bidi    | {{transport-parameter-definitions}} |
+|  0x9  | initial_max_streams_uni     | {{transport-parameter-definitions}} |
+|  0xA  | ack_delay_exponent          | {{transport-parameter-definitions}} |
```
martinthomson: HeX IS LOWeRCaSe

```suggestion
|  0xa  | ack_delay_exponent          | {{transport-parameter-definitions}} |
```

DavidSchinazi:
Closing this as it was replaced by #3108 and #3109.

PR #3089, opened 2019-10-13
Be more specific when naming bits

rmarx:
- Renamed S bit for static vs dynamic table to T bit (for Table)
- Replaced backticks with single quotes around individual bits (previously, about half used single, half used double)
- Added explicit references to bit letter indicators where they were missing (e.g., for 'H')
- Fixed broken reference to "blocked-decoding" section

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -260,7 +260,7 @@ table, the Required Insert Count is zero.
 
 When the decoder receives a header block with a Required Insert Count greater
 than its own Insert Count, the stream cannot be processed immediately, and is
-considered "blocked" (see {blocked-decoding}).
+considered "blocked" (see {{blocked-decoding}}).
 
 The decoder specifies an upper bound on the number of streams which can be
 blocked using the SETTINGS_QPACK_BLOCKED_STREAMS setting (see
@@ -311,7 +311,7 @@ instructions received on the encoder stream.
 The decoder MUST emit header fields in the order their representations appear in
 the input header block.
 
-### Blocked Decoding  {#blocked-decoding}
+### Blocked Decoding
 
 Upon receipt of a header block, the decoder examines the Required Insert Count.
 When the Required Insert Count is less than or equal to the decoder's Insert
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -311,7 +311,7 @@ instructions received on the encoder stream.
 The decoder MUST emit header fields in the order their representations appear in
 the input header block.
 
-### Blocked Decoding
+### Blocked Decoding  {#blocked-decoding}
 
 Upon receipt of a header block, the decoder examines the Required Insert Count.
 When the Required Insert Count is less than or equal to the decoder's Insert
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -311,7 +311,7 @@ instructions received on the encoder stream.
 The decoder MUST emit header fields in the order their representations appear in
 the input header block.
 
-### Blocked Decoding
+### Blocked Decoding  {#blocked-decoding}
```
LPardue: I think auto anchoring should mean this is not required (I.e. only need to define explicit links that differ from the section title)

draft-ietf-quic-qpack.md
```diff
@@ -682,10 +682,10 @@ acknowledged as this instruction does not insert an entry.
 An encoder adds an entry to the dynamic table where the header field name
 matches the header field name of an entry stored in the static or the dynamic
 table using an instruction that starts with the '1' one-bit pattern.  The second
-(`S`) bit indicates whether the reference is to the static or dynamic table. The
+('T') bit indicates whether the reference is to the static or dynamic table. The
```
LPardue: T seems odd to me because they both use a table. And then we have T=1 meaning that it's the static table.

Are we really so constrained that we can't assign a noun to bits?

draft-ietf-quic-qpack.md
rmarx: A) did not know about auto-anchoring
B) seems like it didn't work for this specific one, since draft-10 of the qpack document has an unresolved reference to the section (see https://tools.ietf.org/html/draft-ietf-quic-qpack-10#section-2.1.3). Not sure why that's failing then...

draft-ietf-quic-qpack.md
rmarx: My main gripe was that we were using S with two different meanings (Sign bit and "Static Table bit"). On slack, @afrind was ok with T, so I went with that. I see it as a table "index" personally, which makes sense to me. 

But really, I don't care which letter or noun is used, as long as the same letter/noun isn't used for different things.

draft-ietf-quic-qpack.md
MikeBishop: The reference in 2.1.3 should have `{{` instead of `{`, looks like.

rmarx: Aha, that might do the trick. Did that instead in 479c7d0

MikeBishop: Just built it locally to confirm; it works now.

draft-ietf-quic-qpack.md
MikeBishop: I'm fine with it being an index of the tables.  In my mind (thanks in part to HPACK) the static table comes "first," so it feels a little weird that it's 1 vs. 0, but that's probably better than turning something purely editorial into a needless breaking change at this point.

afrind: I'm fine with T.  T=1 -> Static is a little weird I guess, but you can just think about it as the sTatic table bit.

afrind:
As for 'B' vs `B`, transport doc just uses (B).  Maybe we should follow suit, bui I can do in a separate pass.

PR #3092, opened 2019-10-15
MUST ACK each ack-eliciting packet once

### LINKED ISSUE #3091: Require sending ACKs for any ack-eliciting packet
It appears that the spec doesn't actually require sending ACKs for every ACK-eliciting packet. This could lead to a perverse implementation just not ACKing things with obvious problems.

ianswett:
The editors assumed this was true, but I think it's worth stating this as a MUST to ensure clarity.

Fixes #3091

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3007,16 +3007,17 @@ guidance offered below seeks to strike this balance.
 
 ### Sending ACK Frames {#sending-acknowledgements}
 
-An endpoint MUST NOT excessively delay acknowledgements of ack-eliciting
-packets.  An endpoint commits to a maximum delay using the max_ack_delay
-transport parameter; see {{transport-parameter-definitions}}.  max_ack_delay
-declares an explicit contract: an endpoint promises to never delay
-acknowledgments of an ack-eliciting packet by more than the indicated value. If
-it does, any excess accrues to the RTT estimate and could result in delayed
-retransmissions from the peer.  For Initial and Handshake packets, a
-max_ack_delay of 0 is used.  The sender uses the receiver's `max_ack_delay`
-value in determining timeouts for timer-based retransmission, as detailed
-in Section 5.2.1 of {{QUIC-RECOVERY}}.
+Every packet SHOULD be acknowledged at least once, and ack-eliciting packets
+MUST be acknowledged at least once within the maximum ack delay. An endpoint
+communicates its maximum delay using the max_ack_delay transport parameter;
+see {{transport-parameter-definitions}}.  max_ack_delay declares an explicit
+contract: an endpoint promises to never intentionally delay acknowledgments
+of an ack-eliciting packet by more than the indicated value. If it does,
+any excess accrues to the RTT estimate and could result in spurious or
+delayed retransmissions from the peer. For Initial and Handshake packets,
+a max_ack_delay of 0 is used. The sender uses the receiver's `max_ack_delay`
+value in determining timeouts for timer-based retransmission, as detailed in
+Section 5.2.1 of {{QUIC-RECOVERY}}.
 
 An ACK frame SHOULD be generated for at least every second ack-eliciting packet.
 This recommendation is in keeping with standard practice for TCP {{?RFC5681}}.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3013,6 +3013,8 @@ guidance offered below seeks to strike this balance.
 
 An ACK frame SHOULD be generated for at least every second ack-eliciting packet.
 This recommendation is in keeping with standard practice for TCP {{?RFC5681}}.
+Every ack-eliciting packet MUST be acknowledged in at least one transmitted
+ACK frame, and SHOULD typically be acknoweldged in more than one ACK frame.
 
 A receiver's delayed acknowledgment timer SHOULD NOT exceed the current RTT
 estimate or the value it indicates in the `max_ack_delay` transport parameter.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -3019,14 +3019,15 @@ ACK frame, and SHOULD typically be acknowledged in more than one ACK frame.
 A receiver's delayed acknowledgment timer SHOULD NOT exceed the current RTT
 estimate or the value it indicates in the `max_ack_delay` transport parameter.
 This ensures an acknowledgment is sent at least once per RTT when packets
-needing acknowledgement are received.  The sender can use the receiver's
-`max_ack_delay` value in determining timeouts for timer-based retransmission.
+needing acknowledgement are received.  The sender uses the receiver's
+`max_ack_delay` value in determining timeouts for timer-based retransmission
+as specified in Section 5.2 of {{QUIC-RECOVERY}}.
 
 In order to assist loss detection at the sender, an endpoint SHOULD send an ACK
 frame immediately on receiving an ack-eliciting packet that is out of order. The
 endpoint MAY continue sending ACK frames immediately on each subsequently
 received packet, but the endpoint SHOULD return to acknowledging every other
-packet after a period of 1/8 x RTT, unless more ack-eliciting packets are
+packet within a period of 1/8 x RTT, unless more ack-eliciting packets are
 received out of order.  If every subsequent ack-eliciting packet arrives out of
 order, then an ACK frame SHOULD be sent immediately for every received
 ack-eliciting packet.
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -3013,6 +3013,8 @@ guidance offered below seeks to strike this balance.
 
 An ACK frame SHOULD be generated for at least every second ack-eliciting packet.
 This recommendation is in keeping with standard practice for TCP {{?RFC5681}}.
+Every ack-eliciting packet MUST be acknowledged in at least one transmitted
+ACK frame, and SHOULD typically be acknoweldged in more than one ACK frame.
```
martinthomson: I think that we want to say that every ack-eliciting packet MUST be acknowledged at least once, and ideally all packets are acknowledged until one of those acknowledgements are also acknowledged.

```suggestion
ACK frame, and SHOULD typically be acknowledged in more than one ACK frame.
```

mikkelfj:
I'm a bit worried this misses the point. Ack-eliciting packets should be acknowledged yes, but so should non-ack-eliciting packets. By stating that ack-eliciting packets should be ACK'ed at least once, it is implied that this is not the case for other packets. This may be correct but not necessarily clear. For example for long transmissions where the peer does nothing but sending ACKs, not all of those ACKs will necessarily have to be ACK'ed. But what is the guiding principle for that? And, since you in this case are sending lots of ACK eliciting packets, you can easily ACK all the peers ACK's. So when would you not want to do that? A special case that could deadlock - e.g. too many ACKs to fit in a packet?

mikkelfj:
Also, when ACKs packets don't fit in packets, this would also hold for ACK'ing ack-eliciting packets. So you could end up having to not ACK'ing some packets and induce loss. Requiring an ACK for all packets would be problematic. I'm not sure we want that. The result could be very hard to implement while the practical benefit would be small: sending multiple ACK packets with different content due to overflow.

draft-ietf-quic-transport.md
```diff
@@ -3013,18 +3013,21 @@ guidance offered below seeks to strike this balance.
 
 An ACK frame SHOULD be generated for at least every second ack-eliciting packet.
 This recommendation is in keeping with standard practice for TCP {{?RFC5681}}.
+Every ack-eliciting packet MUST be acknowledged in at least one transmitted
+ACK frame, and SHOULD typically be acknowledged in more than one ACK frame.
```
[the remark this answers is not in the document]
martinthomson: To Mikkel's comment, could we say that *every* packet SHOULD be acknowledged at least once, and that ack-eliciting packets MUST be acknowledged within the maximum ack delay.  ?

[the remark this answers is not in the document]
ianswett: Sure, that could work.

ianswett:
I just synced to master and now this seems to be an editorial change, at least to me.

That being said, PTAL at the text and tell me if there are any other changes I should make.

draft-ietf-quic-transport.md
```diff
@@ -3007,9 +3007,6 @@ guidance offered below seeks to strike this balance.
 
 ### Sending ACK Frames {#sending-acknowledgements}
 
-An ACK frame SHOULD be generated for at least every second ack-eliciting packet.
-This recommendation is in keeping with standard practice for TCP {{?RFC5681}}.
-
 An endpoint MUST NOT excessively delay acknowledgements of ack-eliciting
```
martinthomson: I think that we're still missing the "An endpoint MUST send at least one acknowledgment for every packet it receives and successfully processes."  This paragraph is mostly about max ack delay, which is fine, but it only really says that you can't excessively delay.

Sure, you could say that "never send" == "excessive delay", but you could also say that you can't delay something that never existed in the first place.

mikkelfj: What if you run out if space in the ACK frame?

martinthomson: If you send an ACK for every other received packet and drop all older acknowledgments, then you use at most 19 bytes for an ACK, probably less.  I don't think that will break any limits.

ianswett: @mikkelfj Even if you only ACK the last 2 packets per ACK, QUIC should work correctly.  It's if packets never get acknowledged, especially consistently, there are potential problems.

mikkelfj: I can see this works on average - but if you opt to max out on max delay to minimize transmissions in one direction you could find you self in trouble. A MUST would then force to send more frequently than optimal whereas SHOULD would allow for occasionally dropping the ball. Dropping the ball frequently would be be bad, of course, suggesting that ACKs are not sent often enough.

ianswett: @mikkelfj I don't understand the issue you're concerned about.  One can fit an ACK of any two packets into a datagram, which is the minimum requirement of this text.  Yes, you could set a very large max_ack_delay, which would cause enormously long PTO timeouts and be very bad for performance in most cases.

mikkelfj: I don't want to make this a hard argument, I'm just slightly concerned. Maybe it is a non-issue, or maybe potential issues are shadowed by other concerns such as PTO. Some devices might not want to fire up a transmitter radio more often than necessary. But generally I'm just concerned with a MUST that cannot be enforced or fully controlled - you can easily claim it works under usual conditions but it might not under abnormal conditions. A SHOULD would make it possible to drop packets under those a conditions. A MUST would either have to make assumptions that are hard to prove correct, or have complex code to handle edge case where ACK's queue op, or violate the MUST. I suspect most implementations would go for the latter and hope they don't violate by sending ACK's often enough.

draft-ietf-quic-transport.md
```diff
@@ -3007,25 +3007,26 @@ guidance offered below seeks to strike this balance.
 
 ### Sending ACK Frames {#sending-acknowledgements}
 
+Every packet SHOULD be acknowledged at least once, and ack-eliciting packets
+MUST be acknowledged at least once within the maximum ack delay. An endpoint
```
nibanks: I think I agree with @mikkelfj that MUST might not be the right word here. It's obviously unenforceable for one. But also, I think there are valid reasons where a packet won't be acknowledged. For one, we have a memory limit for the data structure we use to keep track of packets that need to be acknowledged. At a high enough rate, with enough (random?) packet loss you could have enough gaps to hit that limit. In that case, we start forgetting older packet numbers, and will never end up acknowledging them.

martinthomson: We have a memory limit too, but that doesn't mean that we don't send acknowledgments.  We just send packets more often as a result.

As for untestable, that's true for a bunch of things we require.  The point is to require actions that produce an interoperable, functional protocol.  In the general case, ignoring this requirement would be bad.  At insane incoming packet rates, maybe this breaks down, and that's probably not ideal, but I find it hard to believe that you can't generate enough packets outbound to acknowledge every inbound packet at least once.  Even if you only receive every other packet, you are only spending two bytes per packet to acknowledge them.  That's almost a 600:1 advantage to the receiver.

mikkelfj: I think the biggest problem is if the peer assumes a packet was not processed if it eventually concludes it was not ACK'ed. That could lead to deadlocks or double spending in some poor transactional design.

ianswett: An acknowledgement indicates a packet was received and processed.  It makes no claim that other packets are never processed.  This is true throughout the QUIC design.

mikkelfj: The MUST de facto makes that claim, although timing can make it hard to decide.

MikeBishop: No, it doesn't, because you have no guarantee that you received every ACK the peer sent.  This requires the peer to put every packet into an ACK frame sent in at least one packet; there's no requirement to ensure that it's delivered reliably to the peer.  Thus, if you haven't received an ACK, that doesn't mean the peer didn't received the packet anyway.  It could be delayed and still in transit, or the ACK might have been lost.

That said, there's no actual difference between "MUST ack every packet, but not reliably" and "SHOULD ack every packet" from the other side's point of view.  This is something that could be bench-tested on an implementation, but can't be ascertained about the peer over a lossy network.

draft-ietf-quic-transport.md
```diff
@@ -3007,25 +3007,26 @@ guidance offered below seeks to strike this balance.
...
+communicates its maximum delay using the max_ack_delay transport parameter;
+see {{transport-parameter-definitions}}.  max_ack_delay declares an explicit
```
MikeBishop: It's probably correct to start the sentence with a lower-case letter, since this is a literal name of a protocol element, but it looks weird.  Consider rephrasing to avoid the question.

draft-ietf-quic-transport.md
mikkelfj: True, I forgot that a specific PN can be present in one ACK frame on not in the next, even if there are both older and younger PNs in the next ACK frame. This makes it inconclusive.

PR #3103, opened 2019-10-16
Prohibit PADDING in response to ACK

### LINKED ISSUE #3100: ACK+PADDING in response to ACK should be illegal
quic-transport draft 13.2.1:

An endpoint MUST NOT send a packet containing
   only an ACK frame in response to a non-ACK-eliciting packet (one
   containing only ACK and/or PADDING frames),

@marten-seemann points out that this allows one to send an ACK+PADDING packet, which seems to violate the spirit of the rule. This loophole breaks #3093. I believe this may be Editorial if we all know what we meant.

martinduke:
Fixes #3100

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2880,7 +2880,7 @@ packet.
 expectations about what implementations do with packets that have errors after
 valid frames? -->
 
-### Sending ACK Frames
+### Sending ACK Frames {#sending-ack-frames}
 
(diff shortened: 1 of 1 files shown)


martinduke:
Git mistake; better PR coming

janaiyengar:
@martinduke : Did you mean to close this?

martinduke:
Yes, there was a git issue and there's a different or that is the correct
one.

Issue #3094, opened 2019-10-15
congestion window increase on every ACKed packet could result in bursty sends


goelvidhi:
When an ACK arrives that acknowledges N packets (where N could be very large), QUIC will increase the congestion window by more than 1MSS per congestion window during Congestion Avoidance causing bursty send. This could also affect the slow start.

if (congestion_window < ssthresh):
       // Slow start.
       congestion_window += acked_packet.size
     else:
       // Congestion avoidance.
       congestion_window += kMaxDatagramSize * acked_packet.size / congestion_window

One could argue that QUIC ACKs every other packet, but there ACK packet could get lost or  ACK for ACK packet could get lost.

The draft is probably assuming packet pacing but not all implementations support that.

I discussed this with Jana and he thinks that we should add something in pseudo code or some specific text in the draft.

ianswett:
PR #2675 added more specific text on avoiding bursts into the network, FYI.

goelvidhi:
Please note that PR #2675 addresses the issue of app-limited or pacing-limited cases to not increase cwnd when it is under utilized.

My concern is cwnd being inflated due to an ACK frame containing ACKs for large number of packets.

janaiyengar:
@ianswett : The problem here is that we don't talk about burstiness that can occur when pacing is not in place. PRR handles this, but we don't do PRR in the draft.

huitema:
No we need to spend lots of time on Reno as everybody appears to be doing Cubic or BBR?

janaiyengar:
Cubic has this problem. BBR cannot be assumed.

huitema:
Except Cubic opens the window based on time spent, not just ack received

huitema:
I don't get the scenario. It seems that @goelvidhi is worried about receiving an ACK for a large number of packets. This is not a congestion window problem, because it happens even if the congestion windows size remains constant. The root cause is that the ACK of N packets now reduces the number of packet in flow by N. So this is strictly about pacing, not about congestion control.

janaiyengar:
@huitema: yes, and it can and still does burst because of stretch acks

huitema:
Not just stretch Acks -- deliberate Ack compression by the network happens too. Very few alternative to pacing.

junhochoi:
https://tools.ietf.org/html/rfc3465 describes 
```This document specifies that TCP implementations MAY use L=2*SMSS
   bytes and MUST NOT use L > 2*SMSS bytes.  This choice balances
[...]

PR #3106, opened 2019-10-17
MUST pace or limit bursts to 10 packets

ianswett:
Came from the discussion of #3094

### DIFF
--- a/draft-ietf-quic-recovery.md
+++ b/draft-ietf-quic-recovery.md
@@ -753,6 +753,12 @@ delivery of ACK frames is important for efficient loss recovery. Packets
 containing only ACK frames should therefore not be paced, to avoid delaying
 their delivery to the peer.
 
+Sending multiple packets into the network without any delay between them
+creates a packet burst that might cause short-term congestion and losses.
+Implementations MUST either use pacing or limit such bursts to minimum
+of 10 * kMaxDatagramSize and max(2* kMaxDatagramSize, 14720)), the same
+as the recommended initial congestion window.
+
 As an example of a well-known and publicly available implementation of a flow
 pacer, implementers are referred to the Fair Queue packet scheduler (fq qdisc)
 in Linux (3.11 onwards).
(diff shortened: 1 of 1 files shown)


janaiyengar:
@ianswett : I realized after we did this PR yesterday that this does not fix #3094. I It reduces burstiness in general,  which is still a good thing, but it doesn't change slow start increase. I'll send a PR or ask Vidhi if she wants to.

draft-ietf-quic-recovery.md
```diff
@@ -753,6 +753,12 @@ delivery of ACK frames is important for efficient loss recovery. Packets
 containing only ACK frames should therefore not be paced, to avoid delaying
 their delivery to the peer.
 
+Sending multiple packets into the network without any delay between them
+creates a packet burst that might cause short-term congestion and losses.
+Implementations MUST either use pacing or limit such bursts to minimum
```
mikkelfj: to minimum -> to the minimum

ianswett: Done in head.

draft-ietf-quic-recovery.md
```diff
@@ -753,6 +753,12 @@ delivery of ACK frames is important for efficient loss recovery. Packets
...
+of 10 * kMaxDatagramSize and max(2* kMaxDatagramSize, 14720)), the same
+as the recommended initial congestion window.
+
```
mirjak: Sorry again late here... To we recommend a fixed value of 10 packets or do we recommend to have the burst limit the same as the initial window? E.g. if we decide in future to increase our recommendation for the initial window, should that also increase the recommendation for the burst limit? I guess to be really smart about this, you would need to say that you set the burst but initial window but decrease if you see tail loss in the initial burst (but that might be an overkill).

martinthomson: Or you could say that there are different initial window and burst limits, which is probably OK.  I don't think that we need to be that smart though.

ianswett: That's a good point.  We could swap the sentence and say "same as the initial congestion window, which is recommended to be ..." if you think it's preferable.  But I'm also not sure it matters much.

martinthomson: I think that explicitly saying "Initial congestion window" here as the primary target of the MUST requirement would be better.

mirjak: Fine for me.

ianswett: PR created, PTAL: https://github.com/quicwg/base-drafts/pull/3160

PR #3107, opened 2019-10-17
Immediately close with INVALID_TOKEN

### LINKED ISSUE #3168: Allow servers to close connections immediately when the token is corrupted
When the server receives a Token that is corrupted, it may want to close the connection quickly.  This issue suggests adding a new error code to indicate the token is corrupted, INVALID_TOKEN.

ianswett:
If the Retry token is known to be invalid by the server, then the server can close the connection with INVALID_TOKEN instead of waiting for a timeout.

Came out of discussion of Issue #3014
Fixes #3168 

Related to #3127

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1646,7 +1646,8 @@ it knows the client will not accept another Retry token.  It can either
 proceed with the handshake without verifying the token or immediately close
 ({{immediate-close}}) the connection with an connection error of
 INVALID_TOKEN to cause the handshake to fail quickly instead of waiting
-for the client to timeout.
+for the client to timeout.  When a client receives a CONNECTION_CLOSE
+with an error of INVALID_TOKEN, it MAY create a new connection.
 
 A flow showing the use of a Retry packet is shown in {{fig-retry}}.
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1643,10 +1643,12 @@ more resources available for new connections.
 
 If a server receives a client Initial that can be unprotected but contains an
 invalid Retry token, it knows the client will not accept another Retry token.
-It can either proceed with the handshake without verifying the token or
-immediately close ({{immediate-close}}) the connection with an INVALID_TOKEN
-error to cause the handshake to fail quickly instead of
-waiting for the client to timeout.
+It can proceed with the handshake without verifying the token, drop the Initial
+packet, or immediately close ({{immediate-close}}) the connection with an
+INVALID_TOKEN error to cause the handshake to fail quickly instead of waiting
+for the client to timeout. The server MAY close the connection without creating
+connection state, including not adding the connection to those in the closing
+state.
 
 A flow showing the use of a Retry packet is shown in {{fig-retry}}.
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1644,7 +1644,7 @@ more resources available for new connections.
 If a server receives a client Initial that can be unprotected but contains an
 invalid Retry token, it knows the client will not accept another Retry token.
 It can either proceed with the handshake without verifying the token or
-immediately close ({{immediate-close}}) the connection with a connection
+immediately close ({{immediate-close}}) the connection with an INVALID_TOKEN
 error of INVALID_TOKEN to cause the handshake to fail quickly instead of
 waiting for the client to timeout.
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1641,13 +1641,12 @@ of connection establishment.  By giving the client a different connection ID to
 use, a server can cause the connection to be routed to a server instance with
 more resources available for new connections.
 
-If a server receives a client Initial with a unverifiable Retry token,
+If a server receives a client Initial with an invalid Retry token,
 it knows the client will not accept another Retry token.  It can either
 proceed with the handshake without verifying the token or immediately close
 ({{immediate-close}}) the connection with an connection error of
 INVALID_TOKEN to cause the handshake to fail quickly instead of waiting
-for the client to timeout.  When a client receives a CONNECTION_CLOSE
-with an error of INVALID_TOKEN, it MAY create a new connection.
+for the client to timeout.
 
 A flow showing the use of a Retry packet is shown in {{fig-retry}}.
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -5584,7 +5584,7 @@ PROTOCOL_VIOLATION (0xA):
 
 : An endpoint detected an error with protocol compliance that was not covered by
   more specific error codes.
-  
+
 INVALID_TOKEN (0xB):
 : A server received a Retry Token in a client Initial that is invalid.
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1643,12 +1643,14 @@ more resources available for new connections.
 
 If a server receives a client Initial that can be unprotected but contains an
 invalid Retry token, it knows the client will not accept another Retry token.
-The server can drop such a packet and allow the client to time out to detect
-handshake failure, but that is a significant latency penalty to the client.
-If possible, the server SHOULD either immediately close ({{immediate-close}})
-the connection with an INVALID_TOKEN error or proceed with the handshake
-without verifying the token. The server MAY close the connection without
-creating connection state, including skipping the closing state.
+The server can discard such a packet and allow the client to time out to
+detect handshake failure, but that is a significant latency penalty to the
+client.  A server MAY proceed with the connection without verifying the token,
+though the server MUST NOT consider the client address validated.  If a server
+chooses not to proceed with the handshake, it SHOULD immediately close
+({{immediate-close}}) the connection with an INVALID_TOKEN error.  Note that a
+server has not established any state for the connection at this point and so
+does not enter the closing period.
 
 A flow showing the use of a Retry packet is shown in {{fig-retry}}.
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1645,7 +1645,7 @@ If a server receives a client Initial that can be unprotected but contains an
 invalid Retry token, it knows the client will not accept another Retry token.
 It can either proceed with the handshake without verifying the token or
 immediately close ({{immediate-close}}) the connection with an INVALID_TOKEN
-error of INVALID_TOKEN to cause the handshake to fail quickly instead of
+error to cause the handshake to fail quickly instead of
 waiting for the client to timeout.
 
 A flow showing the use of a Retry packet is shown in {{fig-retry}}.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1644,10 +1644,10 @@ more resources available for new connections.
 If a server receives a client Initial that can be unprotected but contains an
 invalid Retry token, it knows the client will not accept another Retry token.
 The server can discard such a packet and allow the client to time out to
-detect handshake failure, but that could impose a significant latency penalty on the
-client.  A server MAY proceed with the connection without verifying the token,
-though the server MUST NOT consider the client address validated.  If a server
-chooses not to proceed with the handshake, it SHOULD immediately close
+detect handshake failure, but that could impose a significant latency penalty on
+the client.  A server MAY proceed with the connection without verifying the
+token, though the server MUST NOT consider the client address validated.  If a
+server chooses not to proceed with the handshake, it SHOULD immediately close
 ({{immediate-close}}) the connection with an INVALID_TOKEN error.  Note that a
 server has not established any state for the connection at this point and so
 does not enter the closing period.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1644,7 +1644,7 @@ more resources available for new connections.
 If a server receives a client Initial that can be unprotected but contains an
 invalid Retry token, it knows the client will not accept another Retry token.
 The server can discard such a packet and allow the client to time out to
-detect handshake failure, but that is a significant latency penalty to the
+detect handshake failure, but that could impose a significant latency penalty on the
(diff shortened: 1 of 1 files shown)


nibanks:
You need to update the "Transport Error Codes" section too.

ianswett:
Thanks @nibanks I did that in a subsequent update.  Refresh and tell me if it looks correct.

nibanks:
@ianswett you're missing the section that gives some text to each error code `#error-codes`.

draft-ietf-quic-transport.md
```diff
@@ -1641,6 +1641,14 @@ of connection establishment.  By giving the client a different connection ID to
 use, a server can cause the connection to be routed to a server instance with
 more resources available for new connections.
 
+If a server receives a client Initial with a unverifiable Retry token,
```
kazuho: Maybe s/unverifiable/invalid/?

To me it seems that "unverifiable" might have the implication of the decryption process (or the integrity check) of the token failing. But a rejection of Retry token happens in other cases too, e.g., when the Retry token is outdated.

draft-ietf-quic-transport.md
```diff
@@ -1641,6 +1641,14 @@ of connection establishment.  By giving the client a different connection ID to
...
+it knows the client will not accept another Retry token.  It can either
+proceed with the handshake without verifying the token or immediately close
```
martinthomson: I don't think that we should allow the server to continue.  Why not make the INVALID_TOKEN mandatory?

draft-ietf-quic-transport.md
```diff
@@ -1641,6 +1641,14 @@ of connection establishment.  By giving the client a different connection ID to
...
+({{immediate-close}}) the connection with an connection error of
+INVALID_TOKEN to cause the handshake to fail quickly instead of waiting
+for the client to timeout.  When a client receives a CONNECTION_CLOSE
+with an error of INVALID_TOKEN, it MAY create a new connection.
```
martinthomson: This isn't a very useful "MAY", because we couldn't stop a client from doing this.  We never say anything about making other connections anywhere else.

draft-ietf-quic-transport.md
kazuho: If we are to make this a mandatory requirement, under the assumption that the servers would act like that, should we also suggest that a server should issue tokens in a way that a Retry token and a NEW_TOKEN token are distinguishable from each other?

Unless we recommend servers building a token in that way, I do not think making this a MUST has benefits.

marten-seemann: If we assume that servers only require Retries when they’re under attack, they might not bother checking the validity of a token when they’re not under attack.

martinthomson: > A token SHOULD be constructed in a way that allows the server to distinguish it from tokens that are sent in Retry packets as they are carried in the same field.

-- [Section 8.1.2](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#validate-future)

We could strengthen that to MUST if you like.  It probably should be anyway.

draft-ietf-quic-transport.md
ianswett: Done.

draft-ietf-quic-transport.md
ianswett: The server could decide it's not under heavy load and proceed with the handshake, even if the token doesn't constitute address validation.  ie: The token could expire on a shorter timeline than the session ticket/etc, so 0-RTT could succeed, but address validation would fail.

draft-ietf-quic-transport.md
ianswett: SG, sentence removed.

draft-ietf-quic-transport.md
martinthomson: I'm persuaded by Marten's comment here (which I missed at first, I think because we commented at almost the same time).

MikeBishop: I still think the requirement to make them distinguishable should be a MUST.  Otherwise, a change of token generation method could make an old NEW_TOKEN token appear to be an invalid Retry token.

ianswett:
@MikeBishop do you have a suggestion about what the client should do?  Clearly it could start a new connection, but it would't want to do that an unlimited number of times, and it may depend upon whether a TCP fallback is available.  Given that, it wasn't clear I should recommend something.

And I agree this should be a MUST, as stated in #3128

martinthomson:
I think that the best thing we can say is "the connection attempt failed".  The first version of the PR strongly suggested that making another connection attempt was the right thing to do, but that gets into lots of difficult questions about identifying why.  Given that this is an error that only happens if a Retry is spoofed, then I think that all we can say is:

it's a trap

draft-ietf-quic-transport.md
```diff
@@ -1641,6 +1641,13 @@ of connection establishment.  By giving the client a different connection ID to
 use, a server can cause the connection to be routed to a server instance with
 more resources available for new connections.
 
+If a server receives a client Initial with an invalid Retry token,
+it knows the client will not accept another Retry token.  It can either
```
janaiyengar: I don't understand this sentence. How would the server know this for sure?

kazuho: That's a keen observation (I think).

There is a chance where a Retry token is considered invalid due to packet becoming corrupt while being transmitted. A server can only say for sure that "the client will not accept another Retry token" unless either of the following conditions are met:
* The token is accompanied by a checksum (or an integrity protection). The checksum is correct, but the contents of the Retry token is invalid.
* The server successfully unprotects the Initial.

mikkelfj: The client could also connect to an unexpected endpoint or the token could have expired.

kazuho: Let me retract my [comment above](https://github.com/quicwg/base-drafts/pull/3107/files#r341481158). We do not need to care about the packet becoming corrupt on the wire, because the prerequisite for responding with a CONNECTION_CLOSE is that the Initial that has been received can be unprotected.

To avoid confusion, maybe change "If a server receives a client Initial with an invalid Retry token" to "If a server receives a client Initial that can be unprotected but contains an invalid Retry token"?

> The client could also connect to an unexpected endpoint or the token could have expired.

I'd say that that's a failure in the server deployment. We do not discuss how that should be handled.

mikkelfj: > I'd say that that's a failure in the server deployment. We do not discuss how that should be handled.

Since Retry is processed before handshake, why can't it just be the client sending to the wrong IP address?

ianswett: #3128 requires being able to distinguish between the two types of tokens.  I updated the text with Kazuho's suggestion about being able to unprotect the Initial.

draft-ietf-quic-transport.md
```diff
@@ -1641,6 +1641,13 @@ of connection establishment.  By giving the client a different connection ID to
 use, a server can cause the connection to be routed to a server instance with
 more resources available for new connections.
 
+If a server receives a client Initial that can be unprotected but contains an
+invalid Retry token, it knows the client will not accept another Retry token.
+It can either proceed with the handshake without verifying the token or
+immediately close ({{immediate-close}}) the connection with a connection
```
```suggestion
immediately close ({{immediate-close}}) the connection with an INVALID_TOKEN
```

draft-ietf-quic-transport.md
```diff
@@ -1641,6 +1641,13 @@ of connection establishment.  By giving the client a different connection ID to
...
+error of INVALID_TOKEN to cause the handshake to fail quickly instead of
```
```suggestion
error to cause the handshake to fail quickly instead of
```

draft-ietf-quic-transport.md
```diff
@@ -1641,6 +1641,13 @@ of connection establishment.  By giving the client a different connection ID to
...
+waiting for the client to timeout.
```
janaiyengar: This suggests 2 options at the server, when there are 3: it can simply drop the packet and let the client time out. You may want to call out all three options. 

Suggestion: "If a server receives a client Initial that can be unprotected but contains an invalid Retry token, it knows the client will not accept another Retry token. The server can drop such a packet and allow the client to time out to detect loss of this connection, but that is a significant latency penalty to the client. If possible, the server SHOULD either immediately close the connection with an INVALID_TOKEN error or proceed with the handshake without verifying the token."

kazuho: I agree that listing all the three options would be an improvement.

At the same time, I have a concern regarding the suggestion of doing either an immediate close or proceeding the handshake. Those two options both require the server to retain state that cannot be delayed by providing a Retry to token to the client.

I think it we might want to downgrade the SHOULD to MAY, or, suggest that a server can send a CONNECTION_CLOSE frame (with INVALID_TOKEN) without retaining state.

janaiyengar: It should be possible to detect an invalid token without per-connection state, right? 

That said, I realized that I don't think we should recommend that a server that might be getting DoSsed SHOULD send out another packet in response. How about "A server that is able to maintain enough state MAY either ..."

kazuho: By state I mean the closing and the draining state that gets built and retained once you invoke {immediate-close}. We should not recommend such behavior as it would create a DoS vector. But if it is not going to be a SHOULD but instead “can” or MAY, I think we can be precise and therefor the text we have now is fine.

ianswett: I added a sentence about not creating any state in this case.  PTAL.

draft-ietf-quic-transport.md
```diff
@@ -1641,12 +1641,14 @@ of connection establishment.  By giving the client a different connection ID to
...
-If a server receives a client Initial with an invalid Retry token,
-it knows the client will not accept another Retry token.  It can either
-proceed with the handshake without verifying the token or immediately close
-({{immediate-close}}) the connection with an connection error of
-INVALID_TOKEN to cause the handshake to fail quickly instead of waiting
-for the client to timeout.
+If a server receives a client Initial that can be unprotected but contains an
+invalid Retry token, it knows the client will not accept another Retry token.
+The server can drop such a packet and allow the client to time out to detect
+handshake failure, but that is a significant latency penalty to the client.
```
martinthomson: Maybe instead:

> The server could discard such a packet, but this could result in the client continuing to send the invalid Initial packet until it is eventually forced to abandon the connection attempt.  A server that requires address validation SHOULD immediately close ({{imm...}}) the connection with an INVALID_TOKEN error to allow the client to detect the error sooner.  Note that a server has not established any state for the connection at this point and so does not enter the closing period.  Alternatively, a server MAY proceed with the connection without verifying the token, though the server MUST NOT consider the client address to be valid.

janaiyengar: I definitely like this text better. That said, I'm not sure that we should recommend sending a connection close. Esp. when under DoS, we do not want servers doing more work than necessary.

martinthomson: What does Roberto say again?  I think that if the server is under duress, then dropping the packet is fine.  This is a recommendation that allows the server to help the client, that's all.  Do you think we need to acknowledge the possibility that the server is overloaded and add the third case?

janaiyengar: The text seems to assume that an implementer will handle the third case, where I was leaning towards being explicit. Perhaps something like: "When under duress, the server could discard such a packet. Dropping such a packet could however result in ..." That makes it a bit more explicit.

ianswett: I reworked Martin's suggestions, though I preferred to keep the sentence I had to his first sentence.  

In regards to "When under duress, the server could discard such a packet." I don't think that's adding anything on top of being clear about the consequences of discarding it, so I'd rather not add it.

draft-ietf-quic-transport.md
```diff
@@ -1641,6 +1641,17 @@ of connection establishment.  By giving the client a different connection ID to
 use, a server can cause the connection to be routed to a server instance with
 more resources available for new connections.
 
+If a server receives a client Initial that can be unprotected but contains an
+invalid Retry token, it knows the client will not accept another Retry token.
+The server can discard such a packet and allow the client to time out to
+detect handshake failure, but that is a significant latency penalty to the
```
```suggestion
detect handshake failure, but that could impose a significant latency penalty on the
```

draft-ietf-quic-transport.md
```diff
@@ -1641,6 +1641,17 @@ of connection establishment.  By giving the client a different connection ID to
...
+If a server receives a client Initial that can be unprotected but contains an
+invalid Retry token, it knows the client will not accept another Retry token.
+The server can discard such a packet and allow the client to time out to
+detect handshake failure, but that could impose a significant latency penalty on
+the client.  A server MAY proceed with the connection without verifying the
+token, though the server MUST NOT consider the client address validated.  If a
+server chooses not to proceed with the handshake, it SHOULD immediately close
+({{immediate-close}}) the connection with an INVALID_TOKEN error.  Note that a
+server has not established any state for the connection at this point and so
+does not enter the closing period.
```
kazuho: Would it make sense to change the last two sentences to something like: _If a server chooses not to proceed with the handshake, it SHOULD immediately send a CONNECTION_CLOSE frame with an INVALID_TOKEN error without creating any state. As the server does not establish any state for the connection, it does not enter the closing period ({{immediate-close}})._

Current text is confusing as the normative text suggests following the rule in {{immediate-close}}, then argues *as if it is a fact* that a server would not be retaining any state, even though what we want to do here is to *recommend* the server to send CONNECTION_CLOSE in a stateless manner.

draft-ietf-quic-transport.md
```diff
@@ -1641,6 +1641,17 @@ of connection establishment.  By giving the client a different connection ID to
 use, a server can cause the connection to be routed to a server instance with
 more resources available for new connections.
 
+If a server receives a client Initial that can be unprotected but contains an
+invalid Retry token, it knows the client will not accept another Retry token.
```
marten-seemann: This only applies if the token is a Retry token (and not a NEW_TOKEN token).
When the client uses a NEW_TOKEN token, it would exchange this token for the Retry token when it receives a Retry packet, wouldn't it?

draft-ietf-quic-transport.md
martinthomson: Discarding is mentioned up front, so we don't need to concern ourselves with repeating that.

draft-ietf-quic-transport.md
martinthomson: Note that the text explicitly says "Retry token", so I don't see how this is a problem.

PR #3109, opened 2019-10-17
Change transport parameter IANA registration policy

### LINKED ISSUE #3020: Transport parameter registry is too constraining for innovation

DavidSchinazi:
As discussed in Cupertino, this PR proposes changing the IANA policy for transport parameter identifiers.

Closes #3020.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -5830,17 +5830,18 @@ IANA \[SHALL add/has added] a registry for "QUIC Transport Parameters" under a
 "QUIC Protocol" heading.
 
 The "QUIC Transport Parameters" registry governs a 16-bit space.  This space is
-split into two spaces that are governed by different policies.  Values with the
-first byte in the range 0x00 to 0xfe (in hexadecimal) are assigned via the
-Specification Required policy {{!RFC8126}}.  Values with the first byte 0xff are
-reserved for Private Use {{!RFC8126}}.
+split into three spaces that are governed by different policies.  Values between
+0x0000 and 0x003f (in hexadecimal) are assigned via the Standards Action or IESG
+Review policies {{!RFC8126}}.  Values from 0x0040 to 0x3fff operate on the
+Specification Required policy {{!RFC8126}}.  Values from 0x4000 to 0xffff are
+assigned to Private Use {{!RFC8126}}.
 
 Registrations MUST include the following fields:
 
 Value:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -5830,17 +5830,18 @@ IANA \[SHALL add/has added] a registry for "QUIC Transport Parameters" under a
 "QUIC Protocol" heading.
 
 The "QUIC Transport Parameters" registry governs a 16-bit space.  This space is
-split into two spaces that are governed by different policies.  Values with the
-first byte in the range 0x00 to 0xfe (in hexadecimal) are assigned via the
-Specification Required policy {{!RFC8126}}.  Values with the first byte 0xff are
-reserved for Private Use {{!RFC8126}}.
+split into three spaces that are governed by different policies.  Values between
+0x0000 and 0x003f (in hexadecimal) are assigned via the Standards Action or IESG
```
janaiyengar: I don't understand these policies well enough -- why do we need a separate space for Standards Action? Isn't that covered under Specification Required?

DavidSchinazi: @janaiyengar `Standards Action` means WG-track RFC so it requires some amount of IETF consensus, whereas `Specification Required` just means someone wrote a draft. See RFC 8126 for details

MikeBishop: Not even an I-D on the Datatracker.  If you can point to a URL of a document somewhere that describes it, that's sufficient for Specification Required.

PR #3113, opened 2019-10-17
Clarify that clients don't need to remember unknown SETTINGS

### LINKED ISSUE #3110: Does a client need to remember SETTINGS it doesn't understand when doing 0-RTT resumption?
The HTTP draft currently specifies

>  If a server accepts 0-RTT, but then sends a SETTINGS frame which reduces a setting the client understands or omits a value that was previously specified to have a non-default value, this MUST be treated as a connection error of type HTTP_SETTINGS_ERROR.

Is it intentional that a client MUST remember SETTINGS values that it doesn't support (to check that the server didn't omit a setting)?

nharper:
Fixes #3110.

### DIFF
--- a/draft-ietf-quic-http.md
+++ b/draft-ietf-quic-http.md
@@ -1194,10 +1194,13 @@ A server MAY accept 0-RTT and subsequently provide different settings in its
 SETTINGS frame. If 0-RTT data is accepted by the server, its SETTINGS frame MUST
 NOT reduce any limits or alter any values that might be violated by the client
 with its 0-RTT data.  The server MUST include all settings which differ from
-their default values.  If a server accepts 0-RTT, but then sends a SETTINGS
-frame which reduces a setting the client understands or omits a value that was
-previously specified to have a non-default value, this MUST be treated as a
-connection error of type H3_SETTINGS_ERROR.
+their default values.  If a server accepts 0-RTT but then sends settings that
+are not compatible with the previously specified settings, this MUST be treated
+as a connection error of type H3_SETTINGS_ERROR. If a server accepts 0-RTT but
+then sends a SETTINGS frame that omits a setting value that the client
+understands (apart from reserved setting identifiers) that was previously
+specified to have a non-default value, this MUST be treated as a connection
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-http.md
```diff
@@ -1229,10 +1229,13 @@ A server MAY accept 0-RTT and subsequently provide different settings in its
 SETTINGS frame. If 0-RTT data is accepted by the server, its SETTINGS frame MUST
 NOT reduce any limits or alter any values that might be violated by the client
 with its 0-RTT data.  The server MUST include all settings which differ from
-their default values.  If a server accepts 0-RTT, but then sends a SETTINGS
-frame which reduces a setting the client understands or omits a value that was
-previously specified to have a non-default value, this MUST be treated as a
-connection error of type HTTP_SETTINGS_ERROR.
+their default values.  If a server accepts 0-RTT but then sends a SETTINGS
```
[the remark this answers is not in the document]
nharper: I think that's implied - if a reserved setting was specified, then it was specified to have a non-default value (since there is no default value).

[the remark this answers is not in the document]
LPardue: WFM

draft-ietf-quic-http.md
```diff
@@ -1229,10 +1229,13 @@ A server MAY accept 0-RTT and subsequently provide different settings in its
...
+frame which reduces setting the client understands, this MUST be treated as a
```
martinthomson: An interesting consequence of the "reduces" clause here is that you can't define a "supports" setting with 0=supports and 1=doesn't support.

MikeBishop: reduces "a" setting.  #2972 switched to using the term "compatible" up at lines 1215-1216, including a definition of the term; perhaps using "incompatible" instead of "reduces" here as well would help?

nharper: Thanks for catching the missing "a" in "reduces a setting" (although it's moot now that I've reworded this).

I like using the term "compatible" instead of reduces, since it avoids this odd edge case and imo is more specific.

PR #3120, opened 2019-10-17
Add retry integrity tag

### LINKED ISSUE #3014: Handling of corrupt Retry packets

### LINKED ISSUE #3274: Encrypting Retry token
In Singapore, we reached an impasse regarding the use of encryption on Retry.

From that discussion, there were two outcomes considered:

* "encrypt" using a fixed key and nonce and don't encrypt the token; put the original destination connection ID in the AAD along with the entire packet

* encrypt the token using the fixed key and a nonce that takes the first 8 bytes of the original destination connection ID; the AAD only includes the connection IDs and version

The first can be reduced to running GHASH with a fixed XOR on the final tag if you really want to save effort.  I've implemented and measured that.

The second can be implemented naively, just by building a new AES key for every record.  That's not terribly slow at 2.8x slower.  With some nasty tricks, I was able to get the AES version under 1.7x slower.  Given the relative cost of these functions, I think that this is an acceptable cost.

There's no guarantee that this is correct - I haven't even checked that these programs produce the right output, but the count of operations is approximately right, so I'm posting this issue.  My code can be found at [martinthomson/quic-retry-tag](https://github.com/martinthomson/quic-retry-tag/blob/master/bench.c) in case reproduction or validation thrills you.

Here's the output of a sample run on my old machine with about 1Mb of Retry packets fed through the different implementations, which takes about 32 seconds to complete:

```
Measuring best of 1000 iterations
ghash:          22450242
aes_gcm_slow:   60070152
aes_gcm_fast:   35240211
```

Now, obviously this is a microbenchmark.  The rest of the cost will be elsewhere, but I tried to isolate the variable.  Hopefully this helps the decision-making process.

DavidSchinazi:
As discussed in Cupertino, this PR adds an integrity tag to retry packets to prevent on-network packet corruption from breaking QUIC handshakes.

Closes #3014.
Closes #3274.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1197,6 +1197,64 @@ TLS ClientHello.  The server MAY retain these packets for later decryption in
 anticipation of receiving a ClientHello.
 
 
+## Retry Packet Integrity {#retry-integrity}
+
+Retry packets (see the Retry Packet section of {{QUIC-TRANSPORT}}) carry a
+Retry Integrity Tag that provides two properties: it allows discarding
+packets that have accidentally been corrupted by the network, and it ensures
+that valid Retry packets cannot be sent by off-path attackers.
+
+The Retry Integrity Tag is a 128-bit field that is computed as the output of
+AEAD_AES_128_GCM used with the following inputs:
+
+- The key is 128 bits all set to zero.
+- The nonce is 96 bits all set to zero.
+- The plaintext is empty.
+- The associated data is the contents of the Retry Pseudo-Packet, as described
+  in {{retry-pseudo}}:
+
+~~~
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+
+|1|1| 3 | Unused|
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                         Version (32)                          |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| DCID Len (8)  |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|               Destination Connection ID (0..160)            ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| SCID Len (8)  |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                 Source Connection ID (0..160)               ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                        Retry Token (*)                      ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| ODCID Len (8) |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|          Original Destination Connection ID (0..160)        ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+~~~
+{: #retry-pseudo title="Retry Pseudo-Packet"}
+
+The Retry Pseudo-Packet is not sent over the wire. It is computed by taking
+the transmitted Retry packet and replacing the Retry Integrity Tag with the two
+following fields:
+
+ODCID Len:
+
+: The ODCID Len contains the length in bytes of the Original Destination
+  Connection ID field that follows it, encoded as an 8-bit unsigned integer.
+
+Original Destination Connection ID:
+
+: The Original Destination Connection ID contains the value of the Destination
+  Connection ID from the Initial packet that this Retry is in response to. The
+  length of this field is given in ODCID Len. The presence of this field
+  prevents an off-path attacker from injecting a Retry packet.
+
+
 # Key Update
 
 Once the handshake is confirmed, it is possible to update the keys. The
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1201,8 +1201,8 @@ anticipation of receiving a ClientHello.
 
 Retry packets (see the Retry Packet section of {{QUIC-TRANSPORT}}) carry a
 Retry Integrity Tag that provides two properties: it allows discarding
-packets that have accidentally been corrupted by the network, and it ensures
-that valid Retry packets cannot be sent by off-path attackers.
+packets that have accidentally been corrupted by the network, and it mitigates
+off-path attackers' ability to send valid Retry packets.
 
 The Retry Integrity Tag is a 128-bit field that is computed as the output of
 AEAD_AES_128_GCM used with the following inputs:
@@ -1252,7 +1252,7 @@ Original Destination Connection ID:
 : The Original Destination Connection ID contains the value of the Destination
   Connection ID from the Initial packet that this Retry is in response to. The
   length of this field is given in ODCID Len. The presence of this field
-  prevents an off-path attacker from injecting a Retry packet.
+  mitigates an off-path attacker's ability to inject a Retry packet.
 
 
 # Key Update
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1210,8 +1210,8 @@ AEAD_AES_128_GCM {{!AEAD=RFC5116}} used with the following inputs:
 - The secret key, K, is 128 bits all set to zero.
 - The nonce, N, is 96 bits all set to zero.
 - The plaintext, P, is empty.
-- The associated data, A, is the contents of the Retry Pseudo-Packet, as illustrated
-  in {{retry-pseudo}}:
+- The associated data, A, is the contents of the Retry Pseudo-Packet, as
+  illustrated in {{retry-pseudo}}:
 
 ~~~
  0                   1                   2                   3
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1205,10 +1205,10 @@ packets that have accidentally been corrupted by the network, and it mitigates
 off-path attackers' ability to send valid Retry packets.
 
 The Retry Integrity Tag is a 128-bit field that is computed as the output of
-AEAD_AES_128_GCM used with the following inputs:
+AEAD_AES_128_GCM {{!AEAD=RFC5116}} used with the following inputs:
 
-- The key is 128 bits all set to zero.
-- The nonce is 96 bits all set to zero.
+- The secret key, K, is 128 bits all set to zero.
+- The nonce, N, is 96 bits all set to zero.
 - The plaintext is empty.
 - The associated data is the contents of the Retry Pseudo-Packet, as described
   in {{retry-pseudo}}:
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1215,7 +1215,8 @@ off-path attackers' ability to send valid Retry packets.
 The Retry Integrity Tag is a 128-bit field that is computed as the output of
 AEAD_AES_128_GCM {{!AEAD=RFC5116}} used with the following inputs:
 
-- The secret key, K, is 128 bits all set to zero.
+- The secret key, K, is 128 bits equal to the initial_salt defined in
+  {{initial-secrets}}.
 - The nonce, N, is 96 bits all set to zero.
 - The plaintext, P, is empty.
 - The associated data, A, is the contents of the Retry Pseudo-Packet, as
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1215,8 +1215,7 @@ off-path attackers' ability to send valid Retry packets.
 The Retry Integrity Tag is a 128-bit field that is computed as the output of
 AEAD_AES_128_GCM {{!AEAD=RFC5116}} used with the following inputs:
 
-- The secret key, K, is 128 bits equal to the initial_salt defined in
-  {{initial-secrets}}.
+- The secret key, K, is 128 bits equal to 0xf5ed4642e0e4c8d878bbbc8a828821c9.
 - The nonce, N, is 96 bits all set to zero.
 - The plaintext, P, is empty.
 - The associated data, A, is the contents of the Retry Pseudo-Packet, as
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1209,7 +1209,7 @@ anticipation of receiving a ClientHello.
 
 Retry packets (see the Retry Packet section of {{QUIC-TRANSPORT}}) carry a
 Retry Integrity Tag that provides two properties: it allows discarding
-packets that have accidentally been corrupted by the network, and it mitigates
+packets that have accidentally been corrupted by the network, and it diminishes
 off-path attackers' ability to send valid Retry packets.
 
 The Retry Integrity Tag is a 128-bit field that is computed as the output of
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1209,8 +1209,8 @@ AEAD_AES_128_GCM {{!AEAD=RFC5116}} used with the following inputs:
 
 - The secret key, K, is 128 bits all set to zero.
 - The nonce, N, is 96 bits all set to zero.
-- The plaintext is empty.
-- The associated data is the contents of the Retry Pseudo-Packet, as described
+- The plaintext, P, is empty.
+- The associated data, A, is the contents of the Retry Pseudo-Packet, as illustrated
   in {{retry-pseudo}}:
 
 ~~~
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -1216,7 +1216,11 @@ AEAD_AES_128_GCM used with the following inputs:
 ~~~
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
-+-+-+-+-+-+-+-+-+
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| ODCID Len (8) |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|          Original Destination Connection ID (0..160)        ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |1|1| 3 | Unused|
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                         Version (32)                          |
@@ -1231,16 +1235,12 @@ AEAD_AES_128_GCM used with the following inputs:
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                        Retry Token (*)                      ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-| ODCID Len (8) |
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|          Original Destination Connection ID (0..160)        ...
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ~~~
 {: #retry-pseudo title="Retry Pseudo-Packet"}
 
 The Retry Pseudo-Packet is not sent over the wire. It is computed by taking
-the transmitted Retry packet and replacing the Retry Integrity Tag with the two
-following fields:
+the transmitted Retry packet, removing the Retry Integrity Tag and prepending
+the two following fields:
 
 ODCID Len:
 
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -4158,8 +4158,8 @@ attempt.  After the client has received and processed an Initial or Retry packet
 from the server, it MUST discard any subsequent Retry packets that it receives.
 
 Clients MUST discard Retry packets whose Retry Integrity Tag cannot be
-validated, see the Retry Packet Integrity section of {{QUIC-TLS}}. This prevents
-an off-path attacker from injecting a Retry packet.
+validated, see the Retry Packet Integrity section of {{QUIC-TLS}}. This
+mitigates an off-path attacker's ability to inject a Retry packet.
 
 The client responds to a Retry packet with an Initial packet that includes the
 provided Retry Token to continue connection establishment.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -4159,7 +4159,7 @@ from the server, it MUST discard any subsequent Retry packets that it receives.
 
 Clients MUST discard Retry packets whose Retry Integrity Tag cannot be
 validated, see the Retry Packet Integrity section of {{QUIC-TLS}}. This
-mitigates an off-path attacker's ability to inject a Retry packet, and protects
+mitigates an off-path attacker's ability to inject a Retry packet and protects
 against accidental corruption of Retry packets.
 
 The client responds to a Retry packet with an Initial packet that includes the
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -4181,7 +4181,7 @@ from the server, it MUST discard any subsequent Retry packets that it receives.
 
 Clients MUST discard Retry packets that have a Retry Integrity Tag that cannot
 be validated, see the Retry Packet Integrity section of {{QUIC-TLS}}. This
-mitigates an off-path attacker's ability to inject a Retry packet and protects
+diminishes an off-path attacker's ability to inject a Retry packet and protects
 against accidental corruption of Retry packets.
 
 The client responds to a Retry packet with an Initial packet that includes the
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2736,7 +2736,7 @@ available.
 
 ## Protected Packets {#packet-protected}
 
-All QUIC packets except Version Negotiation and Retry packets use authenticated
+All QUIC packets except Version Negotiation packets use authenticated
 encryption with additional data (AEAD) {{!RFC5116}} to provide confidentiality
 and integrity protection. Details of packet protection are found in
 {{QUIC-TLS}}; this section includes an overview of the process.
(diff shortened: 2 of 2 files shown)


draft-ietf-quic-tls.md
```diff
@@ -1197,6 +1197,64 @@ TLS ClientHello.  The server MAY retain these packets for later decryption in
...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|               Destination Connection ID (0..160)            ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| SCID Len (8)  |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                 Source Connection ID (0..160)               ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                        Retry Token (*)                      ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| ODCID Len (8) |
```
marten-seemann: Why do we need the length here?

draft-ietf-quic-transport.md
```diff
@@ -4196,10 +4189,7 @@ processing a Retry packet; {{packet-0rtt}} contains more information on this.
 
 A server acknowledges the use of a Retry packet for a connection using the
 original_connection_id transport parameter (see
-{{transport-parameter-definitions}}).  If the server sends a Retry packet, it
-MUST include the value of the Original Destination Connection ID field of the
-Retry packet (that is, the Destination Connection ID field from the client's
-first Initial packet) in the transport parameter.
+{{transport-parameter-definitions}}).
```
marten-seemann: This could be more concrete, e.g. `by setting the original_connection_id transport parameter to the Destination Connection ID from the client's first Initial packet`.

draft-ietf-quic-tls.md
kazuho: We need the length, as if it's absent, an attacker on the path from client to server but not on the opposite direction can obtain an Initial carrying DCID=X, race an Initial carrying DCID=`prefix || X`. Assuming that the Retry in response to the spoofed Initial wins the race, the client would learn that the value of the token is `correct_token  || prefix`.

That said, I do not think having ODCID Len at this position prevents the attack, because the Retry Token field preceding this field does not have an explicit length. Rather, I think that we need to either place ODCID Len and ODCID before the Retry packet, or add the length of the Retry packet at the very front.

mikkelfj: Well, if the token is followed by an integrity tag, the would implicitly define the token length, no?

draft-ietf-quic-transport.md
nibanks: I agree with @marten-seemann here. The server MUST set the transport parameter. You're removing normative language here.

draft-ietf-quic-tls.md
kazuho: @mikkelfj This is about the format of the pseudo packet that we use to *calculate* the integrity tag.

draft-ietf-quic-transport.md
DavidSchinazi: Good point, fixed.

draft-ietf-quic-tls.md
DavidSchinazi: Good catch, moved ODCID to the start to remove the ambiguity

mikkelfj: @kazuho I would assume it had a pseudo packet length use to compute the tag, otherwise it would be a futile exercise. I'm probably still missing something. A weaker integrity check would need some padding protection but GCM should be fine.

mikkelfj: yeah, I'm missing that two fields without a length is ambigious.

draft-ietf-quic-tls.md
```diff
@@ -1197,6 +1197,64 @@ TLS ClientHello.  The server MAY retain these packets for later decryption in
 anticipation of receiving a ClientHello.
 
 
+## Retry Packet Integrity {#retry-integrity}
+
+Retry packets (see the Retry Packet section of {{QUIC-TRANSPORT}}) carry a
+Retry Integrity Tag that provides two properties: it allows discarding
+packets that have accidentally been corrupted by the network, and it ensures
+that valid Retry packets cannot be sent by off-path attackers.
```
chris-wood: nit: perhaps soften this to, "... , and it helps mitigate off-path attacker Retry packet injection"? (The word "ensures" is a little misleading to me.)

DavidSchinazi: Good point, thanks! Fixed.

draft-ietf-quic-tls.md
```diff
@@ -1197,6 +1197,64 @@ TLS ClientHello.  The server MAY retain these packets for later decryption in
 anticipation of receiving a ClientHello.
 
 
+## Retry Packet Integrity {#retry-integrity}
+
+Retry packets (see the Retry Packet section of {{QUIC-TRANSPORT}}) carry a
+Retry Integrity Tag that provides two properties: it allows discarding
+packets that have accidentally been corrupted by the network, and it mitigates
+off-path attackers' ability to send valid Retry packets.
```
ianswett: I'm unsure how this achieves the latter goal?

draft-ietf-quic-tls.md
```diff
@@ -1197,6 +1197,64 @@ TLS ClientHello.  The server MAY retain these packets for later decryption in
...
+| SCID Len (8)  |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                 Source Connection ID (0..160)               ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                        Retry Token (*)                      ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+~~~
+{: #retry-pseudo title="Retry Pseudo-Packet"}
+
+The Retry Pseudo-Packet is not sent over the wire. It is computed by taking
```
ianswett: I'm confused by the term Psuedo-Packet.  If you're intending to remove these fields in a future PR, I'd prefer to keep them in for now to make this simpler.

draft-ietf-quic-tls.md
DavidSchinazi: We prevent off-path attackers from injecting Retry packets by including the Original Destination Connection ID in the associated data here. If they didn't see the client's Initial their Retry will not validate.

draft-ietf-quic-tls.md
DavidSchinazi: This concept comes from how TCP computes its checksum to include fields in the checksum that aren't transmitted in the TCP header: https://tools.ietf.org/html/rfc793#section-3.1

ianswett: I somehow missed the removal of ODCID and ODCID len upon first review, so this makes sense now.

Q: Is there a reason this needs to be in TLS?  It seems like this doesn't interact with TLS at all.

draft-ietf-quic-tls.md
ianswett: Thanks, how about being being specific and saying "to send valid retry packets without having seen the original destination connection ID."?

That was also true before this change, so I was thinking there was some new protection I was unaware of.

draft-ietf-quic-tls.md
DavidSchinazi: This is in TLS because the transport doc currently has no dependency on AES. I'll let the editors decide where this should live.

draft-ietf-quic-tls.md
```diff
@@ -1197,6 +1197,64 @@ TLS ClientHello.  The server MAY retain these packets for later decryption in
...
 
+## Retry Packet Integrity {#retry-integrity}
+
+Retry packets (see the Retry Packet section of {{QUIC-TRANSPORT}}) carry a
+Retry Integrity Tag that provides two properties: it allows discarding
+packets that have accidentally been corrupted by the network, and it mitigates
+off-path attackers' ability to send valid Retry packets.
+
+The Retry Integrity Tag is a 128-bit field that is computed as the output of
+AEAD_AES_128_GCM used with the following inputs:
```
```suggestion
AEAD_AES_128_GCM {{!AEAD=RFC5116}} used with the following inputs:
```

draft-ietf-quic-transport.md
```diff
@@ -2736,7 +2736,7 @@ available.
 
 ## Protected Packets {#packet-protected}
 
-All QUIC packets except Version Negotiation and Retry packets use authenticated
+All QUIC packets except Version Negotiation packets use authenticated
 encryption with additional data (AEAD) {{!RFC5116}} to provide confidentiality
```
martinthomson: The confidentiality bit means that this statement still applies to Retry.

draft-ietf-quic-transport.md
```diff
@@ -4111,37 +4111,31 @@ wishes to perform a retry (see {{validate-handshake}}).
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                 Source Connection ID (0..160)               ...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-| ODCID Len (8) |
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|          Original Destination Connection ID (0..160)        ...
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 |                        Retry Token (*)                      ...
```
martinthomson: In the interests of cleaner parsing, should we add a length for the token?  That would increase the similarity of this header with the Initial header.

draft-ietf-quic-tls.md
```diff
@@ -1197,6 +1197,64 @@ TLS ClientHello.  The server MAY retain these packets for later decryption in
...
+
+- The key is 128 bits all set to zero.
```
```suggestion
- The secret key, K, is 128 bits all set to zero.
```

draft-ietf-quic-tls.md
```diff
@@ -1197,6 +1197,64 @@ TLS ClientHello.  The server MAY retain these packets for later decryption in
...
+- The nonce is 96 bits all set to zero.
```
```suggestion
- The nonce, N, is 96 bits all set to zero.
```

and continue.

draft-ietf-quic-transport.md
```diff
@@ -4111,37 +4111,31 @@ wishes to perform a retry (see {{validate-handshake}}).
...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                                                               |
++                                                               +
+|                                                               |
++                   Retry Integrity Tag (128)                   +
```
martinthomson: I wonder if we shouldn't just label this AEAD Ciphertext rather than just a tag.

draft-ietf-quic-transport.md
```diff
@@ -4163,10 +4157,9 @@ A client MUST accept and process at most one Retry packet for each connection
 attempt.  After the client has received and processed an Initial or Retry packet
 from the server, it MUST discard any subsequent Retry packets that it receives.
 
-Clients MUST discard Retry packets that contain an Original Destination
-Connection ID field that does not match the Destination Connection ID from its
-Initial packet.  This prevents an off-path attacker from injecting a Retry
-packet.
+Clients MUST discard Retry packets whose Retry Integrity Tag cannot be
+validated, see the Retry Packet Integrity section of {{QUIC-TLS}}. This
+mitigates an off-path attacker's ability to inject a Retry packet.
```
martinthomson: I think that we want to add: "and protects against accidental corruption of Retry packets."

draft-ietf-quic-transport.md
DavidSchinazi: I'm not sure whether Retry packets have less confidentiality protection than Initial packets?

draft-ietf-quic-transport.md
DavidSchinazi: Done

draft-ietf-quic-transport.md
DavidSchinazi: I initially had checksum, which is somewhat closer to our intent by not fully correct. I'd personally rather have a name that conveys intent rather than construction, but I'll let the editors decide.

draft-ietf-quic-transport.md
DavidSchinazi: We could, but then we'd have to handle the case where the length is wrong. I don't feel too strongly either way

DavidSchinazi:
@martinthomson 
> I'm ambivalent with the implicit ODCID thing. I like implicit things generally because you can't succeed if you don't have the right information, but this seems like it might still fail if the client doesn't check the tag. We could completely avoid that being an issue if we enciphered the token, but that might change the optimization options enough that it isn't worthwhile.

The ODCID being implicit has the same checking properties as an explicit ODCID - the client could ignore that field entirely before.

That said I do like the idea of making it harder to not check, by encrypting the token. @nibanks if we were to encrypt the token with a zero-key, would that still be suitable? I think the overhead is very small as long as we don't use HKDF / SHA-2 here.

draft-ietf-quic-tls.md
```diff
@@ -1197,6 +1197,64 @@ TLS ClientHello.  The server MAY retain these packets for later decryption in
...
+packets that have accidentally been corrupted by the network, and it mitigates
+off-path attackers' ability to send valid Retry packets.
+
+The Retry Integrity Tag is a 128-bit field that is computed as the output of
+AEAD_AES_128_GCM {{!AEAD=RFC5116}} used with the following inputs:
+
+- The secret key, K, is 128 bits all set to zero.
+- The nonce, N, is 96 bits all set to zero.
+- The plaintext is empty.
+- The associated data is the contents of the Retry Pseudo-Packet, as described
```
MikeBishop: I'd use a different verb than "described" when referencing the figure.  We've tried to follow the rule that illustrations aren't normative, they're just helpful illustrations of what the text says.

draft-ietf-quic-tls.md
MikeBishop: @martinthomson is the editor of both documents, so I'd say he gets to hold the baton here.

draft-ietf-quic-transport.md
MikeBishop: They're not encrypted, so kind of by definition....

draft-ietf-quic-transport.md
MikeBishop: We have a convention of having the final variable-length field have an implicit length.  It reduces the possibility of "impossible" packets.  I'm inclined to keep the PR as-is.

draft-ietf-quic-transport.md
```diff
@@ -4163,10 +4157,10 @@ A client MUST accept and process at most one Retry packet for each connection
 attempt.  After the client has received and processed an Initial or Retry packet
 from the server, it MUST discard any subsequent Retry packets that it receives.
 
-Clients MUST discard Retry packets that contain an Original Destination
-Connection ID field that does not match the Destination Connection ID from its
-Initial packet.  This prevents an off-path attacker from injecting a Retry
-packet.
+Clients MUST discard Retry packets whose Retry Integrity Tag cannot be
+validated, see the Retry Packet Integrity section of {{QUIC-TLS}}. This
+mitigates an off-path attacker's ability to inject a Retry packet, and protects
```
```suggestion
mitigates an off-path attacker's ability to inject a Retry packet and protects
```

draft-ietf-quic-tls.md
DavidSchinazi: Fine by me. Can you suggest another verb?

draft-ietf-quic-transport.md
martinthomson: We already have impossible packets.  The DCID Len field can indicate a length longer than the packet.

draft-ietf-quic-transport.md
```diff
@@ -4163,10 +4157,10 @@ A client MUST accept and process at most one Retry packet for each connection
 attempt.  After the client has received and processed an Initial or Retry packet
 from the server, it MUST discard any subsequent Retry packets that it receives.
 
-Clients MUST discard Retry packets that contain an Original Destination
-Connection ID field that does not match the Destination Connection ID from its
-Initial packet.  This prevents an off-path attacker from injecting a Retry
-packet.
+Clients MUST discard Retry packets whose Retry Integrity Tag cannot be
```
martinthomson: Unnecessary anthropomorphism (or at least implied possessive).
```suggestion
Clients MUST discard Retry packets that have a Retry Integrity Tag that cannot be
```

mikkelfj:
> There would be a performance penalty when using a software cipher (i.e. when you do not have AESNI).

Yes, this is also a concern I have but it might not be worth the complexity.

A use case goes like:

Embedded microcontroller links to hub via bluetooh and does not have an IP network of its own, but it does have a complete QUIC endpoint that the hub should not interfere with other than deliver to the proper IP remote endpoint and attach a local IP.

The device is constrained and only supports ChaCha/Poly crypto. The exception is then the Retry packet (and possibly the Initial packet, I don't recall). This requires additional software which may be a problem in limited flash storage even if the initial slowdown is otherwise acceptable.

DavidSchinazi:
@kazuho encrypting using AES-GCM with an all-zero key and all-zero nonce is just a XOR operation between a known string and the cleartext so the performance cost is very low.

@mikkelfj QUICv1 uses AES-GCM for Initial packets so all QUIC clients need AES-GCM support. You're free to build a different version of QUIC that only uses ChaChaPoly, but that won't impact the design of QUICv1.

mikkelfj:
@DavidSchinazi on AES-GCM that is not correct: The crypto part yes, but you do not encrypt you only GHASH because it is AEAD only. GHASH is a 128-bit carryless multiplication over a Galois Field which completes the checksum. If you CLMUL instructions this is efficient, otherwise not so much. That is where Poly1305 is a gain. CLMUL is not very common on misc. controllers even if they have other accelerations, at least for the time being. E.g. ESP8266.

I agree that AES-GCM is the right choice given the circumstances but I think Poly1305 would be better for initial and retry because it would work in a broader range. The benefit of AES-GCM is that highly loaded servers will have the necessary accelleration and may slow down on AES-GCM - even if this is a tiny fraction of the overall connection cost, it may be important as it fends of noise from the broader internet.

DavidSchinazi:
I'm going to have this PR match the algorithm used for Initial, and stick with AES-GCM. Feel free to file a separate issue to discuss using ChaChaPoly for both.

mikkelfj:
I think it is way too late to make such a change, which is why I wrote that given the circumstances I agree with the design choice.

kazuho:
@DavidSchinazi 
> encrypting using AES-GCM with an all-zero key and all-zero nonce is just a XOR operation between a known string and the cleartext so the performance cost is very low.

Oh I see what you mean. That's a keen observation. Though, I think that observation also shows the reason some of us thought use of encryption to be beneficial to be pointless.

IIUC, https://github.com/quicwg/base-drafts/pull/3120#pullrequestreview-304931396 suggested using encryption, because it would prevent implementations from using the token without checking the tag. The assumption behind is that AEAD-decryption is going to be an atomic operation.

But as you point out, in this particular case, there is a interesting optimization where endpoints can have a pre-built XOR vector, and just do GCM.

Doesn't that reduce the motivation to use encryption, as it is a complexity to apply XOR?

ianswett:
One reason to use encryption was some implementations(@larseggert ?) already had an AES-GCM implementation, and didn't want to add another hash or crypto function.

martinthomson:
So though there is a simple and easy optimization, many implementations will have AES-GCM available, but won't have an independent GHASH on tap.  And getting the mask requires effort.  And client's can't predict how many bytes of mask they need, so they would have to get quite a few.  I'm speculating here, but I tend to think that this puts a thumb on the scale in the right way.

kazuho:
@martinthomson Yeah, my train of thought as a server stack developer that happens to also work as a client was that I might be interested in having a pre-built XOR mask on the server-side. Then, I might expand that constant vector to the 1,200 bytes, and use it for decoding the Retry packets too.

That said, I do agree that not all stack developers would (think and) actually implement that way. So I agree that there's some value in encrypting the token rather than just applying GMAC, at the cost of as small as requiring XOR on the server-side.

draft-ietf-quic-tls.md
```diff
@@ -1197,6 +1197,64 @@ TLS ClientHello.  The server MAY retain these packets for later decryption in
...
+: The ODCID Len contains the length in bytes of the Original Destination
+  Connection ID field that follows it, encoded as an 8-bit unsigned integer.
+
+Original Destination Connection ID:
+
+: The Original Destination Connection ID contains the value of the Destination
+  Connection ID from the Initial packet that this Retry is in response to. The
+  length of this field is given in ODCID Len. The presence of this field
+  mitigates an off-path attacker's ability to inject a Retry packet.
+
```
huitema: Overselling! If the off-path attacker has a copy of the original Initial packet sent by the client, then it knows the ODCID. I don't see how appending a checksum will change that.

DavidSchinazi: By "off-path attacker" I meant an attacker who does not have a copy of any packets previously sent on-path. Can you suggest better wording?

DavidSchinazi: By "off-path attacker" I meant an attacker who does not have a copy of any packets previously sent on-path. Can you suggest better wording?

DavidSchinazi: By "off-path attacker" I meant an attacker who does not have a copy of any packets previously sent on-path. Can you suggest better wording?

huitema: If the off-path attacker has not seen the client's initial packet, it cannot set the DCID of the Retry to the SCID of the client's Initial, which means the Retry would be rejected by the client already.

The additional protection is only effective if the client uses predictable CID, e.g. zero length -- but that's a truly bad idea for a variety of reasons. Such clients would be exposed to other attacks, e.g. injecting random bytes in the crypto-stream.

So in practice, the retry attack is only effective if the attacker is "in the middle" or "on the side". And the checksum will not protect against that.

huitema: But I would shut up if you replaced "mitigates" by "diminishes".

DavidSchinazi: In my mind mitigate and diminish have very similar meanings in this context, so I've switched the to text to use the word diminish

ekr:
I would suggest two revisions to this PR.

1. Let's encrypt the Retry Token.
2. Instead of having a zero key, let's use a per-version key (analagous to the per-version salt). This has the same anti-ossification benefits.

DavidSchinazi:
@ekr
1. encrypting the token has been discussed in this PR and does not have consensus
2. the zero key can be per-version (by which I mean that QUICv1 uses a zero key, and QUICv2 can decide to use a different key)

The discussion on this PR and related issue #3014 indicates that there is interest in the Retry Integrity Tag. On top of that, the PR as-is is currently covered by the chairs' latest consensus call. Therefore I'd suggest you please file a separate issue to discuss encrypting the token, and we can discuss that once this PR has landed. You can also file a second issue to discuss zero key vs non-zero.

kazuho:
@DavidSchinazi 
> the zero key can be per-version (by which I mean that QUICv1 uses a zero key, and QUICv2 can decide to use a different key)

I think we should change the Retry key for every draft version (or every time we change the Initial salt), as that would prevent ossification assuming that we would be running certain number of draft versions concurrently.

I also prefer having a non-zero salt for the upcoming draft version, as it would be an indication to the implementors that the key is a "constant", rather than zeros that endpoints might synthetically generate (e.g., by using `bzero`).

ekr:
> The discussion on this PR and related issue #3014 indicates that there is interest in the Retry Integrity Tag. On top of that, the PR as-is is currently covered by the chairs' latest consensus call.

Yes,  and I'm responding to that saying we ought to encrypt.

huitema:
Can you elaborate why? Is this about security, or ossification?

-- Christian Huitema

ekr:
Ossification. Same reason we encrypt Initial

ekr:
To recap my proposal from this morning.

1. We should have a separate fixed key derived by HKDF-ing the salt. This is just good key separation practice and we can publish the fixed value in the spec if we want. I think there was consensus on this and I can send text for this. @DavidSchinazi ?

2. We should encrypt the token using that fixed key and use the top min(12,dcid_len) bytes of the DCID. This will provide some obfuscation at the cost of just the AES operation but not the KDF.

DavidSchinazi:
I've made this a separate random key to avoid the HKDF. @ekr I hope this addresses point (1).

Regarding point (2) I don't care either way but if we were to use a connection ID as the nonce I would use the original destination connection ID since that's not sent over the wire but required to be known by endpoints.

ekr:
Yes a random key is fine

marten-seemann:
Can we add a test vector to the appendix of the TLS document?

martinthomson:
> Can we add a test vector to the appendix of the TLS document?

A pull request would be appreciated.  One that creates a .js file - so that the value can be automatically generated - would be ideal.

huitema:
In Singapore, we were debating whether to just do checksum, versus encrypt and use an initial vector based on the CID. I wanted to compare the two variants using a benchmark designed for Picotls, but I ran into a little bit of trouble because of the structure of the picotls encryption API, which looks like:

1) Initialize an encryption context with the key context, the IV, and the authenticated data
2) Encrypt the message
3) Compute the AEAD checksum

With that API, there is no option to not use an IV,. Using a constant IV would have exactly the same cost as using a variable IV. I went on to compare the time to encrypt 64 bytes versus the time to just authenticate 64 bytes. Results are below for two implementations of AES128 GCM, one from OpenSSL crypto library and another from Windows Bcrypt library -- both measured using 64 bit code on my Windows laptop:


   | encrypt | decrypt | authenticate | verify
--  | -- | -- | -- | --
bcrypt  | 0.35751 | 0.22602 | 0.24687 | 0.20189
openssl 1.1.1c | 0.30211 | 0.23956 | 0.24296 | 0.23926

The times are measured here in microseconds -- these are the average on 100,000 operations. We see that the times are fractions of microseconds in both cases. The encryption operation is more expensive than a simple authentication, 45% more with brcrypt, 24% more with openssl. The decrypt operation is 12% more expensive than the simple authenticate with brcypt, almost the same with openssl.

janaiyengar:
@huitema : Let's have general discussion on #3014, not here.

huitema:
On 1/16/2020 7:35 PM, Jana Iyengar wrote:

> Merged #3120 <https://github.com/quicwg/base-drafts/pull/3120> into
> master.
>

I am busy implementing that, and I have a doubt.

In the TLS spec, AEAD keys and vectors are derived from a master secret.
Something like:

   [sender <https://tools.ietf.org/html/rfc8446#ref-sender>]_write_key = HKDF-Expand-Label(Secret, "key", "", key_length)
   [sender <https://tools.ietf.org/html/rfc8446#ref-sender>]_write_iv  = HKDF-Expand-Label(Secret, "iv", "", iv_length)

Then we use the sequence number as a "per packet nonce", and encrypt
using "IV^per-packet_nonce"

But the text of the TLS draft says "The secret key, K, is 128 bits equal
to 0xf5ed4642e0e4c8d878bbbc8a828821c9. The nonce, N, is 96 bits all set
to zero."

Should I read that as equivalent to:

    retry_key = 0xf5ed4642e0e4c8d878bbbc8a828821c9

    retry_iv = 0x00000000000000000000000000

    per-packet-nonce= 0x00000000000000000000000000

Or is it:

     retry_key = HKDF-Expand-Label( 0xf5ed4642e0e4c8d878bbbc8a828821c9,
"key", "", 16)

    retry_iv = HKDF-Expand-Label(0xf5ed4642e0e4c8d878bbbc8a828821c9,
"iv", "", 12)

    per-packet-nonce= 0x00000000000000000000000000

-- Christian Huitema

kazuho:
I'd argue that it is the former. The new text refers to RFC5116 (AES-GCM) directly, and that's natural, because we are hardcoding the use of AES-GCM in this particular case, and because, unlike Initial packets, there is no reason to have a generalized design here.

kazuho:
PS. Though I understand your concern that not every TLS stack exposes direct access to AES-GCM.

If it is the case that the API we can assume our TLS stack to provide is a function that takes (AEAD algorithm, hash algorithm, secret, label prefix) as the input, then we should better specify those rather than specifying the AES key and IV directly.

PR #3121, opened 2019-10-17
Forbid Handshake key discarding

### LINKED ISSUE #2863: unrecoverable loss pattern leads to deadlock

martinthomson:
After many iterations of solutions to this problem, we have decided to
first forbid discard.  We would like to have a concrete discard point,
but we couldn't agree on one that was provably safe.

If we can come up with a mechanism for discarding earlier, or a signal
that comes earlier than the end of the connection, we will first
carefully look at it, but we might accept further changes.

Closes #2863.

### DIFF
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -761,7 +761,8 @@ and ignoring any outstanding Initial packets.
 ### Discarding Handshake Keys
 
 An endpoint MUST NOT discard its handshake keys.  Discarding Handshake keys too
-early can lead to deadlock conditions.
+early can lead to a deadlock condition where one endpoint never receives
+acknowledgments for Handshake packets.
 
 
 ### Discarding 0-RTT Keys
--- a/draft-ietf-quic-tls.md
+++ b/draft-ietf-quic-tls.md
@@ -760,14 +760,8 @@ and ignoring any outstanding Initial packets.
 
 ### Discarding Handshake Keys
 
-An endpoint MUST NOT discard its handshake keys until the TLS handshake is
-confirmed ({{handshake-confirmed}}).  An endpoint SHOULD discard its handshake
-keys as soon as it has confirmed the handshake.  Most application protocols
-will send data after the handshake, resulting in acknowledgements that allow
-both endpoints to discard their handshake keys promptly.  Endpoints that do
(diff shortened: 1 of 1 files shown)


mikkelfj:
It would be helpful with a bit of context on the deadlock problem in the document. Sitting on handshake keys without making use of them will not fix the problem.

mikkelfj:
More specifically: since we have multiple packet number spaces, it is highly practical to replace the early handshake handler with a handler that only deals with post handshake packets. But such a handler will not work if it must still deal with handshake packets. That is a big architectural problem.

Me complaining won't make the deadlock go away, obviously, but the current situation is not very agreeable.

nibanks:
I also find it hard to understand why we must keep the handshake keys (and therefor the corresponding packet space) forever, and that there is no other way to solve this problem.

draft-ietf-quic-tls.md
```diff
@@ -760,14 +760,8 @@ and ignoring any outstanding Initial packets.
...
-An endpoint MUST NOT discard its handshake keys until the TLS handshake is
-confirmed ({{handshake-confirmed}}).  An endpoint SHOULD discard its handshake
-keys as soon as it has confirmed the handshake.  Most application protocols
-will send data after the handshake, resulting in acknowledgements that allow
-both endpoints to discard their handshake keys promptly.  Endpoints that do
-not have reason to send immediately after completing the handshake MAY send
-ack-eliciting frames, such as PING, which will cause the handshake to be
-confirmed when they are acknowledged.
+An endpoint MUST NOT discard its handshake keys.  Discarding Handshake keys too
+early can lead to deadlock conditions.
```
```suggestion
early can lead to the peer never receiving an acknowledgement for Handshake data.
```

martinthomson:
I would very much like to replace this text with something better.  But we were given an impossible set of requirements.  A strong preference was expressed for not having an explicit mechanism for driving toward a state where both peers can agree to drop keys.  All our efforts to find implicit signals have been shown to fail.  As a result, while we can drive handshakes to a completed state, because we rely on lazy signals to drive them to a confirmed state, we don't get to drop handshake keys.

MikeBishop:
While I share everyone's dismay, discussion about whether this is the right technical decision belongs on the issue, not the PR.  See you on #2863.

martinthomson:
OBE

PR #3128, opened 2019-10-22
Mandatory distinction between token types

### LINKED ISSUE #3127: NEW_TOKEN and Retry tokens must be distinguishable
We currently say that these tokens SHOULD be distinguishable by a server on the basis that an error in a NEW_TOKEN token is recoverable, but an error in a Retry token is fatal.  I don't see why this can't be MUST.  If we make it "MUST", then we can rely on this property in the design.

martinthomson:
Closes #3127.

Question for reviewers: do you think it necessary to point out that this distinction doesn't need to be hidden from other entities?  That's almost obvious based on the way that this results in different behaviour from servers, but should it be written down?

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1630,12 +1630,6 @@ that allows the server to identity how it was provided to a client.  These
 tokens are carried in the same field, but require different handling from
 servers.
 
-A token MUST NOT include information that would allow values to be linked by an
-on-path observer to the connection on which it was issued.  For example, it
-cannot include the connection ID or addressing information unless the values are
-encrypted.  Information that allows the server to distinguish between tokens
-from Retry and NEW_TOKEN MAY be accessible to entities other than the server.
-
 
 ### Address Validation using Retry Packets {#validate-retry}
 
@@ -1695,6 +1689,13 @@ Thus, a token SHOULD have an expiration time, which could be either an explicit
 expiration time or an issued timestamp that can be used to dynamically calculate
 the expiration time.  A server can store the expiration time or include it in an
 encrypted form in the token.
+
+A token issued with NEW_TOKEN MUST NOT include information that would allow
+values to be linked by an on-path observer to the connection on which it was
+issued.  For example, it cannot include the connection ID or addressing
+information unless the values are encrypted.  Information that allows the server
+to distinguish between tokens from Retry and NEW_TOKEN MAY be accessible to
+entities other than the server.
 
 It is unlikely that the client port number is the same on two different
 connections; validating the port is therefore unlikely to be successful.
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1625,8 +1625,8 @@ controller.  Clients are only constrained by the congestion controller.
 
 ### Token Construction
 
-Tokens sent in NEW_TOKEN frames or Retry packets MUST be constructed in a way
-that allows the server to identity how it was provided to a client.  These
+A token sent in a NEW_TOKEN frames or a Retry packet MUST be constructed in a
+way that allows the server to identity how it was provided to a client.  These
 tokens are carried in the same field, but require different handling from
 servers.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-transport.md
```diff
@@ -1675,9 +1675,10 @@ one.  The client MUST NOT use the token provided in a Retry for future
 connections. Servers MAY discard any Initial packet that does not carry the
 expected token.
 
-A token SHOULD be constructed in a way that allows the server to distinguish it
-from tokens that are sent in Retry packets as they are carried in the same
-field.
+A token send in NEW_TOKEN frames MUST be constructed in a way that allows the
```
kazuho: s/A token send/A token being sent/?

draft-ietf-quic-transport.md
```diff
@@ -1623,6 +1623,20 @@ also constrained in what they can send by the limits set by the congestion
...
 
 
+### Token Construction
+
+Tokens sent in NEW_TOKEN frames or Retry packets MUST be constructed in a way
+that allows the server to identity how it was provided to a client.  These
+tokens are carried in the same field, but require different handling from
+servers.
+
+A token MUST NOT include information that would allow values to be linked by an
```
kazuho: I think this requirement is specific to tokens used for validating future connections?

Therefore, something like "A token issued using NEW_TOKEN frames MUST NOT ..."

martinthomson: Yeah, I moved too much up.

draft-ietf-quic-transport.md
```diff
@@ -1675,22 +1683,20 @@ one.  The client MUST NOT use the token provided in a Retry for future
...
 between when the token is created and when the token is subsequently used.
 Thus, a token SHOULD have an expiration time, which could be either an explicit
 expiration time or an issued timestamp that can be used to dynamically calculate
 the expiration time.  A server can store the expiration time or include it in an
 encrypted form in the token.
 
+A token issued with NEW_TOKEN MUST NOT include information that would allow
+values to be linked by an on-path observer to the connection on which it was
+issued.  For example, it cannot include the previous connection ID or addressing
+information unless the values are encrypted.  Information that allows the server
```
janaiyengar: Move this "unless" clause up to the first sentence. The unless applies to the MUST NOT.

draft-ietf-quic-transport.md
```diff
@@ -1675,22 +1683,20 @@ one.  The client MUST NOT use the token provided in a Retry for future
...
-field.
-
-The token MUST NOT include information that would allow it to be linked by an
-on-path observer to the connection on which it was issued.  For example, it
-cannot include the connection ID or addressing information unless the values are
-encrypted.
-
 Unlike the token that is created for a Retry packet, there might be some time
 between when the token is created and when the token is subsequently used.
 Thus, a token SHOULD have an expiration time, which could be either an explicit
```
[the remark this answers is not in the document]
janaiyengar: ping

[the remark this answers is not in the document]
martinthomson: Problems with existing text can be taken up on a new issue.

draft-ietf-quic-transport.md
```diff
@@ -1623,6 +1623,14 @@ also constrained in what they can send by the limits set by the congestion
 controller.  Clients are only constrained by the congestion controller.
 
 
+### Token Construction
+
+A token sent in a NEW_TOKEN frames or a Retry packet MUST be constructed in a
+way that allows the server to identity how it was provided to a client.  These
```
mikkelfj: "server to identity" -> "server to identify"

draft-ietf-quic-transport.md
```diff
@@ -1675,22 +1683,20 @@ one.  The client MUST NOT use the token provided in a Retry for future
...
-
 Unlike the token that is created for a Retry packet, there might be some time
 between when the token is created and when the token is subsequently used.
 Thus, a token SHOULD have an expiration time, which could be either an explicit
 expiration time or an issued timestamp that can be used to dynamically calculate
 the expiration time.  A server can store the expiration time or include it in an
 encrypted form in the token.
 
+A token issued with NEW_TOKEN MUST NOT include information that would allow
+values to be linked by an on-path observer to the connection on which it was
```
mikkelfj: Note that we have inconsistent terminology: and "off-path" attack can still read packets in the threat model, but here an "on-path" observer can see packets presumably as opposed to "off-path".

PR #3131, opened 2019-10-22
Improve language on evictable entries.

bencebeky:
Add a note that references on the encoder stream do not block an entry
from eviction.  Trivial as it is, I overlooked this at first, and added quite
some unnecessary complexity that was eventually removed at
https://quiche.googlesource.com/quiche/+/b09d44167c543b80b2c27408dd1f9df3b07c9741.
My motivation here is to help future implementers to avoid this pitfall.

Merge redundant Reference Tracking section into Limits on Dynamic Table
Insertions section without relaxing requirements on the encoder.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -205,9 +205,9 @@ If the dynamic table does not contain enough room for a new entry without
 evicting other entries, and the entries which would be evicted are not
 evictable, the encoder MUST NOT insert that entry into the dynamic table
 (including duplicates of existing entries). In order to avoid this, an encoder
-that uses the dynamic table has to keep track of each dynamic table referenced
-by each header block until that header block is acknowledged by the decoder (see
-{{header-acknowledgement}}).
+that uses the dynamic table has to keep track of each dynamic table entry
+referenced by each header block until that header block is acknowledged by the
+decoder (see {{header-acknowledgement}}).
 
 #### Avoiding Prohibited Insertions
 
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -188,28 +188,25 @@ input header list.
 QPACK is designed to contain the more complex state tracking to the encoder,
 while the decoder is relatively simple.
 
-### Reference Tracking
-
-An encoder MUST ensure that a header block which references a dynamic table
-entry is not processed by the decoder after the referenced entry has been
-evicted.  Hence the encoder needs to retain information about each compressed
-header block that references the dynamic table until that header block is
-acknowledged by the decoder (see {{header-acknowledgement}}).
-
 ### Blocked Dynamic Table Insertions {#blocked-insertion}
 
-A dynamic table entry is considered blocking and cannot be evicted until its
-insertion has been acknowledged and there are no outstanding unacknowledged
-references to the entry.  In particular, a dynamic table entry that has never
-been referenced can still be blocking.
+A dynamic table entry is blocking if its absolute index is larger than or equal
+to the Known Received Count, or if it is referenced by an unacknowledged header
+block.  In particular, a dynamic table entry that has never been referenced can
+still be blocking.
+
+Note that references on the encoder stream do not make an entry blocking,
+because those are guaranteed to be processed before the instruction that evicts
+the entry.
 
 An encoder MUST NOT insert an entry into the dynamic table (or duplicate an
 existing entry) if doing so would evict a blocking entry.  In order to avoid
-this, an encoder that uses the dynamic table has to keep track of blocking
-entries.
+this, the encoder needs to retain information about each header block that
+references the dynamic table until that header block is acknowledged by the
+decoder (see {{header-acknowledgement}}).
 
-Note:
-: A blocking entry is unrelated to a blocked stream, see {{blocked-streams}}.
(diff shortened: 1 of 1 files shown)


dtikhonov:
I would not tie the definition of what it means for an entry to be acknowledged to an implementation.  Keeping a count of entries that have been inserted via the `Known Received Count` is just one way to do it.

The phrase "until its [the entry's] insertion has been acknowledged" is not ambiguous.  I don't agree that there is confusion to begin with.

bencebeky:
> This removes the text that you have to track references. Of course you _do_ to accomplish what this section describes. Do we want a normative term somewhere?
> 

Sure, let me get back to this tomorrow.

> Separately, discussion started on Slack -- maybe instead of improving the language about the term overlap, we could just change the term used for one of the "blocking"s.
>

I'm fine with finding a new term.

bencebeky:
> I would not tie the definition of what it means for an entry to be acknowledged to an implementation. Keeping a count of entries that have been inserted via the `Known Received Count` is just one way to do it.
> 

Both endpoints must keep track of Known Received Count anyway, so this does not feel too implementation-specific to me.

dtikhonov:
> Both endpoints must keep track of Known Received Count anyway

That is incorrect.  An alternative way to do this is to use modulo arithmetic.

bencebeky:
> 
> The phrase "until its [the entry's] insertion has been acknowledged" is not ambiguous. I don't agree that there is confusion to begin with.
>

After re-reading the PR, I agree that this phrase is not ambigous.  I'm reverting this part.

bencebeky:
> This removes the text that you have to track references. Of course you _do_ to accomplish what this section describes. Do we want a normative term somewhere?
> 

The following sentence from the old text that I propose to remove:

> An encoder MUST ensure that a header block which references a dynamic table	
> entry is not processed by the decoder after the referenced entry has been	
> evicted.

is redundant with the following sentence that I am leaving untouched:

> An encoder MUST NOT insert an entry into the dynamic table (or duplicate an
> existing entry) if doing so would evict a blocking entry.

(Redundant in the sense that the second sentence implies the first.)  I do not believe that removing the first sentence while leaving the second one relaxes normative requirements on the encoder, nor do I believe that keeping both sentences makes the draft clearer.

afrind:
Did #3172 address your concerns here?  If not, you may need to rebase and resolve the conflicts.

MikeBishop:
> There were
> already four sentences starting with "Note that" throughout the draft
> before this PR (now there are six)....

Thanks for pointing out one of my main writing tics.  😉  The other is overuse of em-dashes.

> Did #3172 address your concerns here? If not, you may need to rebase and resolve the conflicts.

The Note Bence improved is unnecessary and gone now, but I think Bence's other changes are orthogonal to #3172 and worth keeping.

bencebeky:
> > There were
> > already four sentences starting with "Note that" throughout the draft
> > before this PR (now there are six)....
> 
> Thanks for pointing out one of my main writing tics.  The other is overuse of em-dashes.
> 
> > Did #3172 address your concerns here? If not, you may need to rebase and resolve the conflicts.
> 
> The Note Bence improved is unnecessary and gone now, but I think Bence's other changes are orthogonal to #3172 and worth keeping.

Thanks.  I rebased and rephrased as necessary.

draft-ietf-quic-qpack.md
```diff
@@ -188,30 +188,26 @@ input header list.
...
 Inserting entries into the dynamic table might not be possible if the table
 contains entries which cannot be evicted.
 
 A dynamic table entry cannot be evicted immediately after insertion, even if it
 has never been referenced. Once the insertion of a dynamic table entry has been
-acknowledged and there are no outstanding unacknowledged references to the
-entry, the entry becomes evictable.
+acknowledged and there are no outstanding references to the entry in
+unacknowledged header blocks, the entry becomes evictable.  Note that
+unacknowledged references on the encoder stream do not preclude the eviction of
```
afrind: I don't know that we need to say this, since the previous sentence explicitly says 'references to the entry in unacknowledged header blocks' and header blocks do not appear on the encoder stream.

Or maybe the sentence should just indicate that headers on the encoder stream that reference other headers do not count as a reference.

dtikhonov: I don't agree with introducing the concept of "an unacknowledged reference on the encoder stream."  In the current version of the draft (12), the term "unacknowledged reference" always implies a reference from the header block.  In addition, only insertions and header blocks can be acknowledged.

bencebeky: I have a slight preference of adding an explicit note that encoder stream references do not count, to make sure no one unnecessarily implements tracking for them (like I foolishly did).  But I'm flexible as far as wording is concerned, and do not insist on adding this note if you don't believe it adds any value.  Let me know what you think.

bencebeky: I agree with your observation, thank you.  I removed "unacknowledged" from this sentence.  Does it read better?

MikeBishop: This might be a (very old!) holdover from when inserts weren't always on a single stream.  In that case, you might have an unacknowledged insert that depended on another entry for the field name.  But as they *are* on a single stream, the insert that causes an eviction can only arrive after the insert that has the dependency has already been processed.

Issue #3082, opened 2019-10-08
Insert Count Increment instruction description has race condition


bencebeky:
Assume a decoder receives the following data:

A: the first dynamic table insertion,
B: the second dynamic table insertion,
C: the third dynamic table insertion,
X: a header block with required insert count of one.

A, B, and C are guaranteed to be received in this order, because they are all on
the encoder stream.  X arrives on a request stream, so its order relative to the
insertions is not guaranteed.

Consider the orders A, B, X, C, and A, X, B, C.  Suppose the decoder sends a
Header Acknowledgement of the header block as soon as it is received (all
necessary entries are available by that time).  This Header Acknowledgement
instruction increases the encoder's Known Received Count by one.  Suppose that
after receiving C, the decoder decides to send an Insert Count Increment
instruction.

According to
https://quicwg.org/base-drafts/draft-ietf-quic-qpack.html#insert-count-increment,
"The value of the Increment is the total number of dynamic table insertions and
duplications processed by the decoder since the last time it sent a Header
Acknowledgement instruction that increased the Known Received Count (see Section
2.1.4) or an Insert Count Increment instruction."

In the first case, one insertion (C) was processed since the last Header
Acknowledgment instruction that increased the Known Received Count was sent.  In
the second case, two insertions (B and C).

I'm ready to address this issue with a PR but waiting for review of issue 2942 which would conflict.

afrind:
Ah, it needs to explicitly reference the Required Insert Count from the header block that was acked, rather than "since the last time".  I think this is editorial, since that's the intent of the text.

bencebeky:
I agree that it's editorial.

In #2942 I'm already proposing to mandate both the encoder and the decoder to keep track of the Known Received Count.  Then Insert Count Increment could simply increase Known Received Count, without any references to previously acked header blocks if any etc.

afrind:
Fixed by #3132

PR #3132, opened 2019-10-22
Fix bug in Insert Count Increment description.

bencebeky:
Fixes issue #3082.

This PR fixes the race condition and clarifies that both endpoints must
keep track of the Known Received Count.

Also remove "If a header block was potentially blocking" phrase from Known Received Count section, because it is redundant with the clear instructions on whether to increase the Known Received Count in the next sentence.

Also remove "for example because the encoder or the decoder have chosen not to risk blocked	
streams" and defer to new-table-entries instead, so that all guidance on emitting Insert Count Increment instructions is in one place.  Note that the decoder has no way to know if the encoder chose not to risk blocked streams, so I find this phrase confusing to start with.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -279,27 +279,25 @@ table entries which have been acknowledged, but this could mean using
 literals. Since literals make the header block larger, this can result in the
 encoder becoming blocked on congestion or flow control limits.
 
+
 ### Known Received Count
 
-The Known Received Count is the total number of dynamic table insertions and
-duplications acknowledged by the decoder.  The encoder tracks the Known Received
-Count in order to identify which dynamic table entries can be referenced without
-potentially blocking a stream.  The decoder tracks the Known Received Count in
-order to be able to send Insert Count Increment instructions (see
-{{insert-count-increment}}).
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -279,27 +279,25 @@ table entries which have been acknowledged, but this could mean using
...
+### Known Received Count
 
-If a header block was potentially blocking, the encoder infers from receiving a
-Header Acknowledgement instruction ({{header-acknowledgement}}) that the decoder
-has received all dynamic table state necessary to process that header block.  If
-the Required Insert Count of the acknowledged header block is greater than the
-current Known Received Count, the encoder updates the Known Received Count to
-the value of the Required Insert Count of the acknowledged header block.
+The Known Received Count is the total number of acknowledged dynamic table
+insertions and duplications.  It is updated using Header Acknowledgement
```
afrind: I'm not sure it was wrong before to say it's the number acknowledged by the decoder.   The decoder updates as it acknowledges, the encoder updates as it receives acknowledgements.  Without it, you need to read further to clarify that this is not the insert+dup count as sent by the encoder.

PR #3133, opened 2019-10-22
Make language of representations consistent.

bencebeky:
This was inspired by the comment by afrind at
https://github.com/quicwg/base-drafts/pull/2942#discussion_r335222617.

For each representation, uniformly
* use the word "representation" in the first sentence;
* use the word "identifies" if it's a verbatim entry; "encodes" otherwise;
* identify relative versus post-base indexing in the first sentence;
* move drawing right after first sentence;
* use consistent language describing binary format.

### DIFF
--- a/draft-ietf-quic-qpack.md
+++ b/draft-ietf-quic-qpack.md
@@ -965,9 +965,8 @@ entry.
 
 ### Indexed Header Field
 
-An indexed header field representation identifies an entry in either the static
-table or the dynamic table and causes that header field to be added to the
-decoded header list.
+An indexed header field representation identifies an entry in the static table,
+or an entry in the dynamic table with an absolute index less than the Base.
 
 ~~~~~~~~~~ drawing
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qpack.md
```diff
@@ -965,9 +965,8 @@ entry.
 
 ### Indexed Header Field
 
-An indexed header field representation identifies an entry in either the static
-table or the dynamic table and causes that header field to be added to the
-decoded header list.
+An indexed header field representation identifies an entry in the static table,
+or an entry in the dynamic table with an absolute index less than the Base.
```
afrind: I now find the comparison of absolute index and "Base" (which I assert is a count) to be awkward.  Does it read better with 

> less than the *value of* the Base

afrind: We can tackle this in another PR.  This is all goodness so I'm merging now :)

PR #3138, opened 2019-10-23
Applications need to disable or selectively reject 0-RTT

### LINKED ISSUE #3136: 0-RTT is still optional
In [Section 5.4](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#rfc.section.5.4), we say that one of the required operations on a connection is "enable 0-RTT".

@mikkelfj [asks](https://github.com/quicwg/base-drafts/pull/2857/files_):
> But why does an application need to be able to enable 0-RTT? Isn't 0-RTT optional at transport level, and more so at the app level?

The current text implies that implementations are required to support 0-RTT, which is not the case.  It would be more correct to say that applications must be able to prevent early data from being offered on a connection, even if the implementation supports it.

martinthomson:
Closes #3136.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1159,7 +1159,7 @@ the operations described in this section on a QUIC connection.
 When implementing the client role, applications need to be able to:
 
 - open a connection, which begins the exchange described in {{handshake}};
-- disable or selectively reject 0-RTT; and
+- enable 0-RTT when available; and
 - be informed when 0-RTT has been accepted or rejected by a server.
 
 When implementing the server role, applications need to be able to:
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -1159,7 +1159,7 @@ the operations described in this section on a QUIC connection.
 When implementing the client role, applications need to be able to:
 
 - open a connection, which begins the exchange described in {{handshake}};
-- enable 0-RTT; and
+- disable or selectively reject 0-RTT; and
 - be informed when 0-RTT has been accepted or rejected by a server.
 
 When implementing the server role, applications need to be able to:


draft-ietf-quic-transport.md
```diff
@@ -1159,7 +1159,7 @@ the operations described in this section on a QUIC connection.
 When implementing the client role, applications need to be able to:
 
 - open a connection, which begins the exchange described in {{handshake}};
-- enable 0-RTT; and
+- disable or selectively reject 0-RTT; and
```
ianswett: A client doesn't receive 0-RTT, so how would it reject it?

The way I think of it is:
"be informed when 0-RTT keys are available."
"be informed when 1-RTT keys are available."

I don't think the client application needs to know if 0-RTT is rejected or accepted, as long as the handshake succeeds?  And applications like HTTP/3 don't want to enable 0-RTT, but only send some data with 0-RTT keys.

martinthomson: Whoops.  Let me fix that.

martinthomson: That was silly.  I think that "enable 0-RTT *when available*" suffices.  We don't need to go into detail about 0-RTT keys or any of that mess, that's an internal concern of the transport.

For our stack, we enable 0-RTT always (you can't turn it off), and rely on applications using the handshake state to determine whether what is being sent is 0-RTT (or 0.5-RTT) or not.  If there is no 0-RTT, then attempts to create streams will fail.

ianswett: Your implementation sound similar to ours.  But in both cases, it seems like the application doesn't care of 0-RTT succeeds or fails.  It only cares to know when 1-RTT keys are available.  But the existing text indicates the application needs to know about 0-RTT success/failure.

martinthomson: The application (protocol) definitely need to know when 0-RTT is rejected because it has to start over completely for anything that it might have already sent (that's the very next interface requirement).  However, for that to happen, it makes sense for the application to know when it was attempting 0-RTT.

Our API would satisfy this simply by allowing stream creation and writing prior to handshake completion: the application just needs to understand that writing before handshake completion means that you are attempting 0-RTT.  Our TLS stack has a different API, because of backward compatibility: you have to explicitly turn on 0-RTT.  Either approach works fine, except that QUIC doesn't need to concern itself as much with backward compatibility.

ianswett: Makes sense, I forgot 0-RTT wasn't automatically re-sent in IETF QUIC.

PR #3139, opened 2019-10-23
Packets on one path must not adjust values for a different path

### LINKED ISSUE #2909: Path Migration makes unjustified assumptions about a new path.
This is awkward, because the issue of assumptions about the path appears spread over multiple places.

Section 9.4 of draft-ietf-quic-transport  :
   The capacity available on the new path might not be the same as the
   old path.  Packets sent on the old path SHOULD NOT contribute to
   congestion control or RTT estimation for the new path.
GF: That would be fine **IF** the new path were a separate CC context, which I am not yet clear whether it is, read on:
—
Section 9.4 of draft-ietf-quic-transport  :
   On confirming a peer's ownership of its new address, an endpoint
   SHOULD immediately reset the congestion controller and round-trip
   time estimator for the new path.
GF: Why not MUST? - because I think this is a new path, I think it can only be should if something can confirm this is the same set of network-devices and classifiers on the path. I don't know how you would know that, and I don't like the guess.
—
draft-ietf-quic-transport :
   For instance, a change in the client's port
   number is likely indicative of a rebinding in a middlebox and not a
   complete change in path. 
GF: So. maybe this is the guess, and I don't believe it! In a world of ECMP and other methods, this needs to be treated as a potential change of path.
—
Section 9.4 of draft-ietf-quic-transport:
   While multiple paths might be used during connection migration, a
   single congestion control context and a single loss recovery context
   (as described in [QUIC-RECOVERY]) may be adequate.  
GF: Well it could if the paths **WERE** known to be identical, but how would an endpoint guarantee that? Section 9.2 of QUIC transport seems to suggest the CC is reset for the new path, which I would agree with, but which part of QUIC recovery says that?
—
draft-ietf-quic-recovery:
   For instance, an
   endpoint might delay switching to a new congestion control context
   until it is confirmed that an old path is no longer needed (such as
   the case in Section 9.3.3).
GF:  This seems underspecified and novel to me. I think it may b quite OK to continue using the old path and the old CC, for a while. Same question: is this implying the two paths share a single CC and is there some real mechanism to detect that they are really the same set of network devices on path?

erickinnear:
The intent here was that either (a) you only ever have a singe CC context and you reset it for each path, or (b) you have a CC context for each path, in which case each context is strongly tied to a path and you don't need to reset anything (except when it's been a while and the previous values may no longer be valid, etc.). We require that everyone does at least (a), while not preventing folks who want to do extra work for a potential performance gain from doing (b).

Packets *can* contribute to the CC/RTT estimation for the old path if it's still around and you're willing to maintain that state, which is why this (mistakenly) was a SHOULD.

Closes #2909.

### DIFF
--- a/draft-ietf-quic-transport.md
+++ b/draft-ietf-quic-transport.md
@@ -2096,7 +2096,7 @@ more likely to indicate an intentional migration rather than an attack.
 ## Loss Detection and Congestion Control {#migration-cc}
 
 The capacity available on the new path might not be the same as the old path.
-Packets sent on the old path SHOULD NOT contribute to congestion control or RTT
+Packets sent on the old path MUST NOT contribute to congestion control or RTT
 estimation for the new path.
 
 On confirming a peer's ownership of its new address, an endpoint MUST


martinthomson:
This looks good.  I'm calling it editorial.  However, it changes normative language, so I want an opinion on that from at least @ianswett or @janaiyengar before merging (other opinions are also highly valued).

draft-ietf-quic-transport.md
```diff
@@ -2096,7 +2096,7 @@ more likely to indicate an intentional migration rather than an attack.
 ## Loss Detection and Congestion Control {#migration-cc}
 
 The capacity available on the new path might not be the same as the old path.
-Packets sent on the old path SHOULD NOT contribute to congestion control or RTT
+Packets sent on the old path MUST NOT contribute to congestion control or RTT
 estimation for the new path.
```
mikkelfj: What about the initial RTT estimate on the new path?

ianswett: That's a fair point.  In practice, it's more likely the new path is similar to the old path RTT than it is to a default like 1 second, given these are the same two endpoints.

But I'm not sure that optimization is worth considering.  If you're validating the new path before migrating, which is probably the typical case, you expect to get an RTT sample in 1 round trip anyway.

mirjak: Yes, good question. So I think it is correct that it is important to make sure that when you migrate you need to use a new cc context and not just keep using the old one. But I guess this was a SHOULD because using the old estimate as the new initial could make sense e.g. if you have very small RTTs or only a few ms. On the other hand, as Ian said, you should get a new sample with the validation anyway and therefore it should not have an impact. Do we maybe need more text to actually explain that?

erickinnear: I'd tend to agree with Ian that it's probably not necessary to add text covering that -- you're doing the safest thing if you treat them as completely independent, and you're likely to have new data from the new path just from validation or within a few RTTs.

(That said, I'd be quite interested in the results of experiments in this area, and there's no enforcement here that prevents such experimentation.)

janaiyengar:
+1 that we should require this behavior. We know that implementers will try out tricks here if they want to, and there's no enforcement to stop that from happening.

erickinnear:
I agree that this seems generally editorial, but either way I think this is ready to go?

PR #3140, opened 2019-10-23
Delete reference to crypto timeout

martinduke:
this is editorial, I believe.


martinthomson:
Thanks.

ianswett:
Thanks for the fix!

